Issue #288, opened 2021-04-26
De gebruiker kan inloggen met DigiD, eHerkenning en eIDAS.


joeribekker:
Eis: Authenticatie En Basisgegevens 1.
Zie ook NvI: 173

SilviaAmAm:
- [ ] Digid
- [ ] EHerkenning
- [ ] eIDAS

SilviaAmAm:
Related PRs:
- Open-Forms branch: https://github.com/open-formulieren/open-forms/pull/476
- Open-forms-deployment branch: https://github.com/maykinmedia/open-forms-deployment/pull/5

sergei-maertens:
eHerkenning PRs are merged, not deployed yet.

PR #466, opened 2021-07-01
Feature #323 - Privacy policy page

SilviaAmAm:
Fixes a small part of #323 

For issue #288 (implementing log in with eHerkenning), we need a URL that points to the privacy policy page. This URL is needed because it has to be present in the metadata file that we use to request the test environment. This is the reason for this PR.

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -148,6 +148,7 @@ INSTALLED_APPS = [
     "openforms.prefill.contrib.kvk.apps.KVKPrefillApp",
     "openforms.prefill.contrib.stufbg.apps.StufBgApp",
     "openforms.prefill.contrib.haalcentraal.apps.HaalCentraalApp",
+    "openforms.privacy_policy",
     "openforms.authentication",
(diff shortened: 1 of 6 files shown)


joeribekker (changes requested):
Hi Silvia.

I'm leaving a general review of the code because this is not what I intended. To be clear: No privacy policy content should be uploaded in Open Forms!

The summary should include something like:

image

I'll describe what needs to be done here:

- [ ] Add non-required field in te config model: `privacy_policy_link`
- [ ] Add required field in te config model:  `privacy_policy_label` (default should be: "Ja, ik heb kennis genomen van het {{ privacybeleid }} en geef uitdrukkelijk toestemming voor het verwerken van de door mij opgegeven gegevens.")
- [ ] The summary page shows a checkbox like the image above (but can probably be a one-liner) where the label is taken from the `privacy_policy_label`.
- [ ] The word between curly braces can be replaced with a hyperlink to the `privacy_policy_link` (to keep it somewhat in line with other variables used in OF). If no link, then just remove the curly braces.
- [ ] Only allow submission if the checkbox is checked.

Converted to new ticket.

PR #477, opened 2021-07-08
Add security features (two factor authentication) to admin

### LINKED ISSUE #473: As user I want a secure admin
*Splitted from #323 *

* [x] Add 2FA to admin (You can use our default project as example since it was added there after we started this project.)
* [x] Enforce strong passwords in admin (Django has some settings I think)
* [x] Password reset needs to be present (should also be in default project).

shea-maykinmedia:
Fixes #473

For the second point `Enforce strong passwords in admin` I didn't do anything since `AUTH_PASSWORD_VALIDATORS` was already in the default project but I did test that it was working as expected. 

I tested this both with creating users through the `createsuperuser` command and also making a new user/gebruiker through the admin screen.

**Screenshots**

Logging in

Screenshot 2021-07-08 at 12 17 35

Screenshot 2021-07-08 at 12 18 51

Changing password validation.  Nothing was changed about this but the ticket mentioned it so I tested to double check.

Screenshot 2021-07-08 at 12 24 51

New user setting up two factor authentication for the first time.  Maybe we want to change 'Aanmelden' to something else? 'Volgende' is what is was before.

Screenshot 2021-07-08 at 12 34 02

Screenshot 2021-07-08 at 12 34 17

Screenshot 2021-07-08 at 12 35 31

Screenshot 2021-07-08 at 12 35 40


joeribekker:
Package was released here: https://pypi.org/project/maykin-django-two-factor-auth/

sergei-maertens:
Closing in favour of #568

PR #491, opened 2021-07-14
Fixes #181 -- add option to configure analytics providers

### LINKED ISSUE #181: Een gebruiker van de toepassing kan zonder tussenkomst van de leverancier een trackingscode toevoegen van een ...
...statistiekprogramma. Dit zijn minimaal Siteimprove, Matomo, Piwik en Google Analytics. Daarbij kan de cookiemelding worden in- of uitgeschakeld. Als de melding wordt uiteschakeld, wordt de data alsnog direct verzameld.

Eis: Algemeen 7.
Zie ook NvI: 134

**Tasks**

- [x] Add option to enter various tracking codes
- [x] Add templates for each provider, included on the form detail page
- [x] cookiemelding?

**Verduidelijking van Gemeente nodig**

- [x] Hoe moet het frame er een beetje uit zien van een "blanco" gemeente pagina waarin het formulier komt?
- [x] Welke melding moet in de cookiemelding?
- [ ] Graag voorbeelden geven van snippets voor SiteImprove, Matomo, Piwik

sergei-maertens:
Fixes #181

TODO:
- [x] Support various providers
    - [x] Google Analytics / Tag Manager
    - [x] Matomo
    - [x] Piwik
    - [x] SiteImprove
- [x] tests
- [x] cookie notice?


src/openforms/config/models.py
```diff
@@ -136,8 +136,83 @@ class GlobalConfiguration(SingletonModel):
...
+        max_length=10,
+        blank=True,
+        help_text=_(
+            "Your SiteImprove ID - you can find this from the embed snippet example, "
+            "which should contain a URL like '//siteimproveanalytics.com/js/siteanalyze_XXXXX.js'. "
+            "The XXXXX is your ID."
+        ),
+    )
+
+    analytics_cookie_consent_group = models.ForeignKey(
```
joeribekker, line 191: Is this intentionally missing from the admin page?

src/openforms/fixtures/cookie_consent.json
```diff
@@ -0,0 +1,88 @@
+[
```
joeribekker, line 1: Is this file supposed to be loaded for every new installation, or just for test purposes?

If the first, then we should document this or make it part of some hook or playbook?
If the latter, then it should be in `fixtures/test/` imo.

src/openforms/templates/cookie_consent/cookiegroup_list.html
```diff
@@ -0,0 +1,24 @@
...
+            <ul class="list">
+                {% for cookie_group in object_list  %}
+                    <li class="list__item">
+                        {% include "cookie_consent/_cookie_group.html" %}
+                    </li>
+                {% endfor %}
+            </ul>
+        {% endblock %}
+    </div>
+{% endblock %}
```
joeribekker, line 24: I think this page needs a "Close"-button that will take the user back to the previous page. After selecting Accept/Decline, there's not web-way to return to where I was when I clicked the cookiebanner message.

src/openforms/templates/master.html
```diff
@@ -9,17 +14,28 @@
...
+    {% include "includes/cookie-notice.html" %}
 
-<script src="{% static 'bundles/openforms-js.js' %}"></script>
-{% block extra_js %}{% endblock %}
+    <script src="{% static 'bundles/openforms-js.js' %}"></script>
+    {% block extra_js %}{% endblock %}
+    {% if enable_analytics %}
+        {# Load the various supported analytics provider snippets #}
+        {% include "includes/analytics/all_bottom.html" %}
+    {% endif %}
```
joeribekker, line 38: Although this is a temporary page, we should include a link to the cookie setting page at all times (footer stuff prolly). Something not to forget when we create the subdomain page.

joeribekker (changes requested):

src/openforms/config/models.py
sergei-maertens: nope, oversight!

src/openforms/fixtures/cookie_consent.json
sergei-maertens: It's both - intended as a starting point to have *something* for new installations, and also used in unit tests because it's convenient.

I'll add a data migration that loads the fixture, so it's loaded for new fixtures & can be edited if needed (so we don't overwrite changes made by customers which would happen with a hook or playbook).

PR #493, opened 2021-07-15
[#223] Implemented file upload component

Bartvaderkin:
Also includes #488 (email registration) #224 (file size) #226 (image resize) #314 (file name)

SDK: https://github.com/open-formulieren/open-forms-sdk/pull/37


src/openforms/submissions/attachments.py
```diff
@@ -0,0 +1,211 @@
...
+                upload.file_name, base_name, i, len(uploads)
+            )
+
+            attachment, created = SubmissionFileAttachment.objects.create_from_upload(
+                submission_step, key, upload, file_name=file_name
+            )
+            result.append((attachment, created))
+
+            if created and resize_apply and resize_size:
+                resize_submission_attachment.delay(attachment.id, resize_size)
```
sergei-maertens, line 102: There's a possible race condition here where the image resizing has not completed before the submission is completed (and registration kicks off). This could lead to the original images without resizing being uploaded into a zaaksysteem/documenten API.

Given the user interaction required I don't think this will occur, but it's technically possible. I've created #507 so that it's at least on the backlog. Please update the code with a comment pointing to the possible problem and include a link to #507.

src/openforms/submissions/templates/admin/submissions/submission/change_form.html
```diff
@@ -21,5 +21,24 @@
...
         </div>
+        <div class="form-row field-attachments">
+            <div>
+                <label>{% trans "Attachments:" %}</label>
+                <div class="readonly">
+                    <ul>
+                        {% for key, files in attachments.items %}
+                            <li>{{ key }}:
+                                {% for file in files %}
+                                    <a href="{{ file.content.url }}">
```
sergei-maertens, line 32: I suspect this will be a URL that doesn't work because of the private media field thing, unless that magically uses the `PrivateMediaView`.

src/openforms/submissions/tests/test_submission_attachment.py
```diff
@@ -0,0 +1,347 @@
...
+
+@temp_private_root()
+class SubmissionAttachmentTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.test_image_path = os.path.join(
+            os.path.dirname(__file__), "files", "image-256x256.png"
+        )
+
+    # TODO private storage
```
sergei-maertens, line 34: what's this TODO about?

sergei-maertens (changes requested):
Good work!

The permission classes for GET/DELETE of a temporary upload needs fixing, the rest are rather minor remarks.

src/openforms/submissions/api/serializers.py
```diff
@@ -243,3 +244,43 @@ class ReportStatusSerializer(serializers.Serializer):
...
+        name: 'The_Name_Of_The_File.doc',
+        size: 1000
+    }
+    """
+
+    file = serializers.FileField(write_only=True, required=True, use_url=False)
+
+    url = serializers.SerializerMethodField(
+        label=_("Url"), source="get_url", read_only=True
+    )
```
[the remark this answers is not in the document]
Bartvaderkin: Sounds familiar but manual doesn't mention it and this works so I guess it is fine.

src/openforms/submissions/attachments.py
Bartvaderkin: We'll probaly have to re-evaluate the async/completion handling as the payment feature will add complications to it as well.

src/openforms/submissions/templates/admin/submissions/submission/change_form.html
Bartvaderkin: I swapped the admin-side file views for custom views.

src/openforms/submissions/tests/test_submission_attachment.py
Bartvaderkin: Something that has been fixed, so I removed the todo.

src/openforms/submissions/api/serializers.py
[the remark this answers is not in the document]
sergei-maertens: Checked it - it's the first positional argument that's validated and not the `source` keyword argument. The `source` kwarg is overwritten too, so you can leave that off:

```py
    def __init__(self, method_name=None, **kwargs):
        self.method_name = method_name
        kwargs['source'] = '*'
        kwargs['read_only'] = True
        super().__init__(**kwargs)
```

sergei-maertens (commented):
With the new models, the `default_admin_index` fixture also needs updating still :-)

Bartvaderkin:
@sergei-maertens I pushed all feedback of today.

sergei-maertens (approved):

PR #501, opened 2021-07-19
:sparkles: [#278] Add Objects API registration backend

stevenbal:
fixes: #278


src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -0,0 +1,42 @@
...
+
+        created_object = objects_client.create(
+            "object",
+            {
+                "type": options["objecttype"],
+                "record": {
+                    "typeVersion": options["objecttype_version"],
+                    "data": {
+                        "data": submission.get_merged_data(),
+                        "type": "terugbelnotitie",
```
alextreme, line 35: I think the 'type' has to be configurable :)

alextreme (approved):
Dank! Prima wat mij betreft, wacht wel met mergen tot Sergei's review

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    global configuration and defaults
+    """
+
+    objects_service = models.OneToOneField(
+        "zgw_consumers.Service",
+        verbose_name=_("Objects API"),
+        on_delete=models.PROTECT,
+        limit_choices_to={"api_type": APITypes.orc},
+        related_name="objects_objects_config",
```
sergei-maertens, line 18: `related_name` does not make any sense with `SingletonModel` since there's only ever one record in the database.

```suggestion
        related_name="+",
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -0,0 +1,42 @@
...
+        objects_client = config.objects_service.build_client()
+
+        created_object = objects_client.create(
+            "object",
+            {
+                "type": options["objecttype"],
+                "record": {
+                    "typeVersion": options["objecttype_version"],
+                    "data": {
+                        "data": submission.get_merged_data(),
```
sergei-maertens, line 34: question: how do we deal with file uploads/attachments?

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -0,0 +1,221 @@
...
+                "startAt": date.today().isoformat(),
+                "endAt": date.today().isoformat(),
+                "registrationAt": date.today().isoformat(),
+                "correctionFor": 0,
+                "correctedBy": "",
+            },
+        }
+
+        m.post(
+            # FIXME zgw-consumers seems to duplicate /api/v2
```
sergei-maertens, line 96: so that effectively makes this PR broken, right?

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -0,0 +1,221 @@
...
+            json=expected_result,
+        )
+
+        plugin = ObjectsAPIRegistration("objects_api")
+        result = plugin.register_submission(submission, {})
+
+        # Result is simply the created object
+        self.assertEqual(result, expected_result)
+
+        # self.assertEqual(len(m.request_history), 1)
```
sergei-maertens, line 202: would be the calls to fetch the API spec and the calls to do the creation, I'd like an explicit assertion here

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/plugin.py
joeribekker: I think we need to:

1) Allow a Documents API to be configured to upload attachment to; And relate the object to these documents (array of urls?)
2) Allow ZGW API's to be configured to either save the submission as PDF in the Documents API (as is now) **OR** save the submission in the Objects API. This also solves attachments because... ZGW.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
stevenbal: For Objects API v2 yes (though v2 is not released yet). The spec of objects v2 has `/api/v2/` in the paths (which won't work with zgw-consumers) like this:

```yaml
/api/v2/objects:
    get:
      operationId: object_list
      description: Retrieve a list of OBJECTs and their actual RECORD. The actual
        record is defined as if the query parameter `date=<today>` was given.
```

Is this an issue in the objects v2 API spec or should this be addressed in zgw-consumers?

I guess for this PR I could switch to objects v1?

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -22,6 +24,12 @@ def register_submission(
         config = ObjectsAPIConfig.get_solo()
         config.apply_defaults_to(options)
 
+        zgw_config = ZgwConfig.get_solo()
```
stevenbal, line 27: Not sure if this should use the `ZGWConfig` or whether the DRC service, `informatieobjecttype` and `vertrouwelijkheidaanduiding` should be added to `ObjectsAPIConfig`.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
stevenbal: I asked Anna and she said Silvia made an issue about this in objects API (https://github.com/maykinmedia/objects-api/issues/226). This API spec change was introduced in 1.1.1: https://github.com/maykinmedia/objects-api/blob/1.1.1/src/openapi.yaml#L71, so the Objects API registration should still work with Objects API <= 1.1.0. For the tests, I changed the Objects OAS to 1.1.0

Is there anything I can do/should do on the openforms/zgw-consumers side about this?

annashamray: The changes to remove `/api/v[0-9]` are merged to the Objects API master branch (and are available through 'latest' tag I believe), so it should not block you anymore

sergei-maertens: I'm not sure what this means, but let's target the latest version of Objects API (`latest` docker image tag).

You can pin the API spec in Open Forms to the commit that is now `master` and please at a comment referring back to this discussion.

Once v2.0 is out, we'll update the tags/images etc.

src/openforms/registrations/contrib/objects_api/plugin.py
sergei-maertens: Let's separate them, you'd expect plugins and their configuration to be standalone, which currently isn't the case.

src/openforms/registrations/contrib/objects_api/plugin.py
sergei-maertens: Uploading to Documents API & storing the resulting URLs in a field seems the simplest way to go for the time being, while still allowing us to make it more nice and expand later.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -0,0 +1,326 @@
...
+            informatieobjecttype="https://catalogi.nl/api/v1/informatieobjecttypen/1",
+            organisatie_rsin="000000000",
+        )
+
+    def tearDown(self):
+        super().tearDown()
+
+        # Reset the schema fetcher cache, to verify that
+        # the OAS is being retrieved on every submission
+        schema_fetcher.cache = {}
```
sergei-maertens, line 44: ZGW consumers installs its django based cache system: https://github.com/maykinmedia/zgw-consumers/blob/master/zgw_consumers/cache.py

I'd move this ordeal to the `setUp` method:

```
def setUp(self):
    super().setUp()

    # ensure the schema cache is cleared before and after each test
    schema_fetcher.cache.clear()
    self.addCleanup(schema_fetcher.cache.clear)
```

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
stevenbal: Mentioned in https://github.com/open-formulieren/open-forms/pull/501/commits/c297718ced70d98a520cb4b64afb9082f7009716

stevenbal:
@sergei-maertens it seems that something goes wrong with the objecten schema in zds_client:
https://github.com/open-formulieren/open-forms/pull/501/checks?check_run_id=3220627009#step:10:1465

It appears that the `Accept-Crs` header in the objecten schema does not have the property `required` (while the headers `Content-Crs` and `Content-Type` do): https://github.com/open-formulieren/open-forms/blob/e08b742c3b05c714cde68efdfaa722a1ba043296/src/openforms/registrations/contrib/objects_api/tests/files/objecten.yaml#L177-L185

Should `required` always be present in the headers in the OAS? Or is this an issue with zds_client?

src/openforms/registrations/contrib/objects_api/plugin.py
stevenbal: changed in https://github.com/open-formulieren/open-forms/pull/501/commits/8ccd9d8f683680328b0ecad1d50ef2d99009c8ee

src/openforms/registrations/contrib/objects_api/plugin.py
stevenbal: Added this in https://github.com/open-formulieren/open-forms/pull/501/commits/1cbdb3b95445cf5cf32ecb116158b77d3dea3b61.

@sergei-maertens I tried some stuff locally and I can't seem to trigger the creation of a `SubmissionFileAttachment`, does this depend on the storage backend that is being used?

sergei-maertens:
@stevenbal that's a bug in zds-client, which was fixed by Silvia in the past few weeks! Upgrading to the latest gemma-zds-client should fix that.

src/openforms/registrations/contrib/objects_api/plugin.py
sergei-maertens: The submission needs to be completed, and I believe it's part of the Celery task workflow to actually process the submissions.

sergei-maertens:
@stevenbal what's the status of this? Ready for review or not? If it is, can you click the "Ready for review" button?

I also see the build is failing, and the TODO item in the PR desription is now possible since the dependency PR has been merged.

stevenbal:
@sergei-maertens it's not ready for review yet, I'll try to get to it asap, but https://github.com/open-formulieren/open-forms/pull/592 got a higher priority by Den Haag

src/openforms/registrations/contrib/zgw_apis/service.py
```diff
@@ -35,9 +36,14 @@ def create_zaak(options: dict) -> dict:
...
 def create_document(
-    name: str, submission_report: SubmissionReport, options: dict
+    name: str,
+    submission_report: SubmissionReport,
+    options: dict,
+    config: Optional[SingletonModel] = None,
 ) -> dict:
-    config = ZgwConfig.get_solo()
+    if not config:
+        config = ZgwConfig.get_solo()
```
sergei-maertens, line 45: I don't like how this implicitly requires the model fields to have the same name. Leaving it in for now, but please pay attention to keeping a clean Python interface, it's already all complicated enough as it is.

sergei-maertens (approved):

src/openforms/registrations/contrib/zgw_apis/service.py
sergei-maertens: refactored it into a getter function

PR #503, opened 2021-07-19
Feature #220 - Form versions

### LINKED ISSUE #220: Eerdere versies van een formulier inclusief alle elementen en instellingen kunnen door de gebruiker worden ...
...teruggeplaatst.

Eis: Elementen 10.
Zie ook NvI: 48, 73, 149

**Taken**

~Zie technische tickets: #148, #149, #150~

- [x] Remove Reversion (as done in #148)
- [x] Add FormVersion model that contains the reference to a Form and a complete form export blob, date of creation 
- [x] Add API endpoint to create a FormVersion (=export) of the current form.
- [x] Call this endpoint when the "save"-button is pressed (and wait for all the React calls to finish)
- [x] Add history page (behind the history button on the form detail page)
- [x] Show a list of form versions. and their date, and a button to restore it (=import form).

### LINKED ISSUE #512: Admin CRUD problems if FormDefinition creation fails on first submission
Currently, in the `onSubmit` function, the component checks whether it should make a POST or a PUT depending on `state.newForm`. 

The order in which the `onSubmit` creates/updates is Form > FormDefinitions > FormSteps > FormVersions.

When creating a new form, if an error occurs in the creation of the Form Definitions, then the form has already been created but `state.newForm` is still `true`. This means that when trying to resubmit after the corrections to the form definitions, a POST will be done for the form instead of a PUT (which will give an error since a form with the slug already exists).

SilviaAmAm:
Fixes #220 
Fixes #512


src/openforms/forms/admin/form.py
```diff
@@ -70,16 +69,18 @@ def render_change_form(
         context.update({"use_react": self.use_react(request)})
         return super().render_change_form(request, context, add, change, form_url, obj)
 
+    def changelist_view(self, request, extra_context=None):
+        context = {
+            "has_change_permission": self.has_change_permission(request),
+        }
+        context.update(extra_context or {})
+        return super().changelist_view(request, context)
```
sergei-maertens, line 77: Why is this needed?

src/openforms/forms/api/viewsets.py
```diff
@@ -294,7 +293,94 @@ def export(self, request, *args, **kwargs):
...
+        parameters=[
+            OpenApiParameter(
+                name="form_uuid_or_slug",
+                location=OpenApiParameter.PATH,
+                type=str,
+                description=_("Either a UUID4 or a slug identifying the form."),
+            )
+        ],
+        request=FormVersionSerializer,
+        responses={status.HTTP_200_OK: FormVersionSerializer(many=True)},
```
sergei-maertens, line 345: these are the defaults and should not be specified explicitly, right?

src/openforms/forms/api/viewsets.py
```diff
@@ -294,7 +293,94 @@ def export(self, request, *args, **kwargs):
...
+                description=_("Either a UUID4 or a slug identifying the form."),
+            ),
+            OpenApiParameter(
+                name="uuid",
+                location=OpenApiParameter.PATH,
+                type=str,
+                description=_("The uuid of the form version"),
+            ),
+        ],
+        responses={status.HTTP_201_CREATED: ""},
```
sergei-maertens, line 331: Shouldn't this be HTTP 204 instead?

src/openforms/forms/api/viewsets.py
```diff
@@ -294,7 +293,94 @@ def export(self, request, *args, **kwargs):
...
+            type=str,
+            description=_("Either a UUID4 or a slug identifying the form."),
+        )
+    ]
+)
+@extend_schema_view(
+    create=extend_schema(
+        summary=_("Save form version"),
+        tags=["forms"],
+        responses={status.HTTP_201_CREATED: FormVersionSerializer},
```
sergei-maertens, line 312: This is the default serializer + HTTP status code, so shouldn't need to be specified.

src/openforms/forms/api/viewsets.py
```diff
@@ -294,7 +293,94 @@ def export(self, request, *args, **kwargs):
...
+
+        return Response(status=status.HTTP_201_CREATED)
+
+    def list(self, request, *args, **kwargs):
+        """List all the versions of a form"""
+        form = get_object_or_404(Form, uuid=self.kwargs["form_uuid_or_slug"])
+
+        form_version_serializer = FormVersionSerializer(
+            instance=FormVersion.objects.filter(form=form), many=True
+        )
```
sergei-maertens, line 382: I feel like this should just be using the `ListModelMixin` combined with filtering the queryset from the `NestedViewSetMixin`.

src/openforms/forms/models/form.py
```diff
@@ -182,3 +184,72 @@ def get_keys_for_email_confirmation(self) -> List[str]:
     def iter_components(self, recursive=True):
         for form_step in self.formstep_set.select_related("form_definition"):
             yield from form_step.iter_components(recursive=recursive)
+
+    def all_versions(self):
+        from .form_version import FormVersion
+
+        return FormVersion.objects.filter(form=self)
```
sergei-maertens, line 191: This seems to be unused, and the better approach would be to use `form.formversion_set.all()` instead anyway.

src/openforms/forms/models/form_version.py
```diff
@@ -0,0 +1,34 @@
...
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from openforms.utils.fields import StringUUIDField
+
+
+class FormVersion(models.Model):
+    uuid = StringUUIDField(_("UUID"), unique=True, default=_uuid.uuid4)
+
+    date_creation = models.DateTimeField(
```
sergei-maertens, line 13: using `date` in a field name that's actually a datetime always makes me feel a bit uneasy as the name gives the wrong impression.

```suggestion
    created = models.DateTimeField(
```

src/openforms/forms/static/forms/scss/components/admin/_form-object-tools.scss
```diff
@@ -0,0 +1,13 @@
...
+    margin: 0;
+  }
+
+  &__loading {
+    a {
+      opacity: 0.65;
+      cursor: not-allowed;
+    }
+  }
+}
```
sergei-maertens, line 13: Don't target HTML elements, only target class names. This allows us to change the markup without having to update styling.

src/openforms/forms/templates/admin/forms/form/change_form.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'admin/change_form.html' %}
-{% load i18n static openforms %}
+{% load i18n static openforms admin_urls%}
```
sergei-maertens, line 2:
```suggestion
{% load i18n static openforms admin_urls %}
```

src/openforms/forms/templates/admin/forms/form/object_history.html
```diff
@@ -0,0 +1,22 @@
...
+    >{# React managed #}</div>
+
+    </div>
+</div>
+
+<script src="{% static 'bundles/core-js.js' %}"></script>
+{% endblock %}
+
+
+
```
sergei-maertens, line 22:
```suggestion
```

src/openforms/forms/tests/admin/test_form_version.py
```diff
@@ -0,0 +1,17 @@
+from django.urls import reverse
+
+from django_webtest import WebTest
+
+from openforms.accounts.tests.factories import UserFactory
+
+
+class FormVersionAdminImportExportTests(WebTest):
+    def setUp(self):
+        self.user = UserFactory.create(is_superuser=True, is_staff=True)
```
sergei-maertens, line 10: I think we have a `SuperUserFactory` now that you can use.

src/openforms/forms/tests/factories.py
```diff
@@ -35,3 +36,17 @@ class Meta:
...
+        model = FormVersion
+
+    form = factory.SubFactory(FormFactory)
+    export_blob = {}
+
+    @factory.post_generation
+    def post(obj, create, extracted, **kwargs):
+        json_form = form_to_json(obj.form.id)
+        obj.export_blob = json_form
+        obj.save()
```
sergei-maertens, line 52: Please follow the django model structure:

1. fields
2. Meta
3. methods

```suggestion
class FormVersionFactory(factory.django.DjangoModelFactory):
    form = factory.SubFactory(FormFactory)
    export_blob = {}

    class Meta:
        model = FormVersion

    @factory.post_generation
    def post(obj, create, extracted, **kwargs):
        json_form = form_to_json(obj.form.id)
        obj.export_blob = json_form
        obj.save()
```

src/openforms/js/components/admin/form_versions/FormVersionsTable.js
```diff
@@ -0,0 +1,78 @@
...
+        await apiCall(
+            `${FORM_ENDPOINT}/${formUuid}/versions/${versionUuid}/restore`,
+            {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json',
+                    'X-CSRFToken': csrftoken,
+                },
+            }
+        );
```
sergei-maertens, line 24: Is there a reason you're not using the `post` function from `utils/fetch`?

src/openforms/js/components/admin/form_versions/FormVersionsTable.js
```diff
@@ -0,0 +1,78 @@
...
+    };
+
+    useAsync(async () => {await getFormVersions(formUuid);}, []);
+
+    const rows = formVersions.map((version, index) => {
+        const dateCreation = new Date(version.dateCreation);
+        return (
+           <tr key={index}>
+                <th>{ dateCreation.toDateString() } { dateCreation.toLocaleTimeString() }</th>
+                <td><a href="#" onClick={() => restoreVersion(csrftoken, formUuid, version.uuid, formAdminUrl)}>Herstellen</a></td>
```
sergei-maertens, line 43: I think you need to capture the `event` here too and call `event.preventDefault()` otherwise the browser will jump to the top of the page because of the `#` href.

sergei-maertens (changes requested):

src/openforms/forms/admin/form.py
SilviaAmAm: This was part of the `VersionAdmin` https://github.com/etianen/django-reversion/blob/master/reversion/admin.py#L229, and without it the `import form` button in the admin is no longer visible.

src/openforms/forms/api/viewsets.py
SilviaAmAm: Yes!

src/openforms/forms/tests/factories.py
SilviaAmAm: I was trying to be consistent with the factories defined above. Should I change the other factories too?

src/openforms/js/components/admin/form_versions/FormVersionsTable.js
SilviaAmAm: I didn't want the second part of the fetch, since no body is returned: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/utils/fetch.js#L43

src/openforms/forms/tests/factories.py
sergei-maertens: Feel free to do so!

src/openforms/js/components/admin/form_versions/FormVersionsTable.js
sergei-maertens: You can adapt the `fetch` call to only get the JSON if there is a response body and otherwise return `null`. That's a good fix/cleanup.

sergei-maertens (approved):

PR #514, opened 2021-07-23
Session timeouts

### LINKED ISSUE #308: Als beheerder kan ik een maximale sessietijd instellen voor ingelogde gebruikers, zodat de site beter beveiligd is.
Eis: Authenticatie En Basisgegevens 6c.
Zie ook NvI: 97

Requirements:
- [x] Maximum session time needed for both for the SDK and the Admin
- [x] Both times should be adjustable in the Admin screen and do not need to be the same (eg. two different fields)

Note: We will **not** be showing a warning to the user before expiring their session.  This will be done in #509

shea-maykinmedia:
Fixes #308 

Related SDK PR: https://github.com/open-formulieren/open-forms-sdk/pull/44

I'm not sure if we want form specific overrides like the button texts.  If we do this won't be hard to add but it felt like something that isn't needed.

I don't know if there is any good way to unit test this but I did manually test it to verify it works.

**Screenshots**

Screenshot 2021-07-29 at 11 11 34


src/openforms/middleware.py
```diff
@@ -18,3 +20,18 @@ def __call__(self, request):
...
+class SessionTimeoutMiddleware:
+    """
+    Allows us to set the expiry time of the session based on what
+    is configured in our GlobalConfiguration
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
```
joeribekker, line 34: This call should ALSO handle form_session_timeout. I think something like this can solve it:

```
if request.user.is_staff:
  # set expiry(admin timeout)
else:
  # set expiry(user timeout)
```

src/openforms/middleware.py
```diff
@@ -18,3 +20,18 @@ def __call__(self, request):
...
+        # https://docs.djangoproject.com/en/2.2/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase.set_expiry
+        request.session.set_expiry(GlobalConfiguration.get_solo().admin_session_timeout)
```
joeribekker, line 36: if minutes this would become something like `timedelta(minutes=timeout)`

joeribekker (changes requested):

joeribekker:
Discussed briefly with Sergei but we don't want a timer in the SDK. Timeout really needs to be invalided server side.

src/openforms/middleware.py
shea-maykinmedia: Ya, it would be 
```
>>> timedelta(minutes=5).seconds
300
```
though we could also just multiply this number by `60`.  Not sure if one is preferred over the other?

sergei-maertens: you'd want `total_seconds` if you convert that, which takes into account possible days/hours/minutes etc.

src/openforms/conf/base.py
```diff
@@ -171,6 +171,7 @@
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "axes.middleware.AxesMiddleware",
+    "openforms.middleware.SessionTimeoutMiddleware",
```
sergei-maertens, line 174: I would expect this to be higher up, right after Django's session middleware

src/openforms/middleware.py
```diff
@@ -18,3 +22,23 @@ def __call__(self, request):
...
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.user.is_staff:
+            timeout = GlobalConfiguration.get_solo().admin_session_timeout
+        else:
+            timeout = GlobalConfiguration.get_solo().form_session_timeout
```
sergei-maertens, line 40:
```suggestion
        config = GlobalConfiguration.get_solo()
        timeout = config.admin_session_timeout if request.user.is_staff else config.form_session_timeout
```

sergei-maertens (changes requested):
Where are the tests?

Also note this line from the [documentation](https://docs.djangoproject.com/en/2.2/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase.set_expiry)

> Reading a session is not considered activity for expiration purposes. Session expiration is computed from the last time the session was modified.

I think this doesn't actually work for the intended purpose - the session is modified when you start a form/submission (either by logging in and setting the BSN in the session, but definitely by storing the submission ID in the session). However, when you're filling out the form with submission steps, that no longer modifies the session. So someone can be filling out a form with multiple steps and take longer than the session timeout, and I think that it will still allow them to continue.

shea-maykinmedia:
> Where are the tests?

Good question ;) I'm not really sure how I can write useful unit tests for this functionality.  I'll think about it a bit more but if you have any suggestions I'm happy to hear them.

> However, when you're filling out the form with submission steps, that no longer modifies the session.

I did just double check/test this and since we're calling `set_expiry` this is modifying the session and so the "timer" resets on every request.  In practice this would mean if the time limit is 5 minutes then as long as each individual step is completed within 5 minutes you won't see a timeout.

src/openforms/conf/base.py
```diff
@@ -168,6 +168,7 @@
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
+    "openforms.middleware.SessionTimeoutMiddleware",
```
shea-maykinmedia, line 171: I put the middleware here since it needs to come after the `AuthenticationMiddleware` since that is what adds the user to the request so we can do `request.user`.

Ref: https://github.com/django/django/blob/main/django/contrib/auth/middleware.py#L25

src/openforms/middleware.py
```diff
@@ -18,3 +22,24 @@ def __call__(self, request):
...
+
+    def __call__(self, request):
+        config = GlobalConfiguration.get_solo()
+        timeout = (
+            config.admin_session_timeout
+            if request.user.is_staff
+            else config.form_session_timeout
+        )
+        # https://docs.djangoproject.com/en/2.2/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase.set_expiry
+        request.session.set_expiry(timedelta(minutes=timeout).seconds)
```
sergei-maertens, line 44: I'll leave this here :-)

```
>>> from datetime import timedelta
>>> timedelta(weeks=1, seconds=5).seconds
5
>>> timedelta(weeks=1, seconds=5).total_seconds()
604805.0
```

sergei-maertens (changes requested):

src/openforms/middleware.py
sergei-maertens: FYI - this doesn't affect us really because we pass `minutes` which is converted into seconds, however it's definitely more robust to use `total_seconds()` in case the call ever changes

src/openforms/tests/test_session_expiry.py
```diff
@@ -0,0 +1,211 @@
...
+                )
+                self.assertEqual(session.get_expiry_date(), expected_expiry)
+
+        # make another request to the admin to validate the session has expired
+        #  and the user will be redirected to the admin login page
+        with self.subTest(part="check expired and redirection to login page"):
+            with freeze_time("2021-07-29T14:10:00Z"):
+                response = self.client.get(self.url, user=self.superuser)
+
+                self.assertEqual(response.status_code, status.HTTP_302_FOUND)
```
sergei-maertens, line 210: This should rather use `self.assertRedirects`

sergei-maertens (approved):

src/openforms/tests/test_session_expiry.py
shea-maykinmedia: Okay, good to know to next time :)

PR #516, opened 2021-07-23
Mark submission registration status as complete when form has no registration backend

### LINKED ISSUE #515: Form submission with no backend should have registration status "succes"
Form submission with no backend selected should just be considered to be saved in Open Forms itself

**Tasks**

- [ ] Set registration backend status to "succes" if the no registration backend is selected and the form was succesfully submitted. This should happen at the same time as the completion date is set.

shea-maykinmedia:
Fixes #515


sergei-maertens (approved):

shea-maykinmedia:
Just double checked this.  First one (pending) was done on the master branch, second one (success) was done on this branch.

Screenshot 2021-07-26 at 09 14 43

PR #517, opened 2021-07-24
Update translations

joeribekker:
* Changed some translated strings in the code to be more accurate and/or in line with other strings, and let this be reflected in migrations.
* Fixed a potential ordering issue with file uploads in `test_submission_attachment.py` (happened locally and needed tests to pass).
* Added BAG to admin index (encountered when checking translations in the admin).
* Updated all translations ofcourse.


sergei-maertens (approved):

joeribekker:
Hmm, not sure why migrations check fails. Locally it seems not to find any changes.

sergei-maertens:
> Hmm, not sure why migrations check fails. Locally it seems not to find any changes.

Was because you ran `compilemessages` locally and the translations were in the migrations because of the `gettext` call instead of `gettext_lazy`. I've updated the models.py now.

joeribekker:
Thanks, missed that. Nice catch on gettext

PR #525, opened 2021-07-26
Add Bag configuration documentation

shea-maykinmedia:
There wasn't a related ticket for this though I can create one if we want.

I added some documentation about the `BAG configuration` to the `docs` folder so it'll be shown in read-the-docs.

**Screenshot**

Screenshot 2021-07-28 at 11 20 34


shea-maykinmedia:
The build is failing because an example url gives a 400 response.  If this is otherwise good I'll fix just this but for now I'm going to wait to see if there are other changes that should be made and if there are I'll make the url change as part of the other changes.

docs/installation/configurations.rst
```diff
@@ -0,0 +1,27 @@
+.. _configurations:
+
+=======================
+Configurations
+=======================
+
+Various configurations can be configured within Open Forms to add data and validation to the forms.
```
sergei-maertens, line 7:
```suggestion
The various modules and plugins in Open Forms require some configuration before they can be used.
```

docs/installation/configurations.rst
```diff
@@ -0,0 +1,27 @@
...
+
+BAG configuration
+-----------------
```
sergei-maertens, line 10: Please stick to the same type of heading markers as used in the rest of the documentation.

```suggestion
BAG configuration
=================
```

docs/installation/configurations.rst
```diff
@@ -0,0 +1,27 @@
...
+
+The BAG configuration contains a single service to be configured.
+To add a service you need to fill in the following information
+
+* ``Label``: What you want to call the service.  Eg. ``Bag Service``
```
sergei-maertens, line 15:
```suggestion
* ``Label``: What you want to call the service.  Eg. ``BAG (Kadaster)``
```

docs/installation/configurations.rst
```diff
@@ -0,0 +1,27 @@
...
+
+* ``Type``: The type of service this service is.  Currently it is limited to ``ORC (Overige)``
```
sergei-maertens, line 17:
```suggestion
* ``Type``: The type of API.  Currently it is limited to ``ORC (Overige)``
```

docs/installation/configurations.rst
```diff
@@ -0,0 +1,27 @@
...
+
+* ``Api root url``: The root of the url this service will communicate with. An example root url would be https://api.bag.kadaster.nl/lvbag/individuelebevragingen/v2/
+
+* ``Authorization type``: The type of authorization the Api uses.  Will likely be ``API key``
+
+* ``Header key``: The header key the Api will use to read the Api key.  Will likely be ``X-Api-Key``
```
sergei-maertens, line 23:
```suggestion
* ``Header key``: The HTTP header name to use for the API-key authorization. Will likely be ``X-Api-Key``.
```

docs/installation/configurations.rst
```diff
@@ -0,0 +1,27 @@
...
+
+* ``Header value``: The Api key.  An Api key can be requested from https://www.kadaster.nl/
+
+* ``OAS``: A url where the OpenAPI 3 yaml file can be retrieved.  An example file would be https://raw.githubusercontent.com/lvbag/BAG-API/master/Technische%20specificatie/Yaml's/BAG%20API%20Individuele%20Bevragingen/resolved/individuelebevragingen/v2/adressen.yaml
```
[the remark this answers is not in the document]
sergei-maertens: Please also pin the API spec to a git tag or commit rather than the master branch.

docs/installation/index.rst
```diff
@@ -11,3 +11,4 @@ Installation
    quickstart
    deployment
    self_signed
+   configurations
```
sergei-maertens, line 14:
```suggestion
   configuration
```

sergei-maertens (changes requested):

sergei-maertens:
> The build is failing because an example url gives a 400 response. If this is otherwise good I'll fix just this but for now I'm going to wait to see if there are other changes that should be made and if there are I'll make the url change as part of the other changes.

The link checks config (in `docs/conf.py`) allows for URL patterns to be excluded, but often if the link does not necessarily have to be clickable in the docs, you should just put it between code back-ticks to format it as code rather than a link.

sergei-maertens (approved):

Issue #532, opened 2021-07-29
Implement logging out functionality


SilviaAmAm:
- [x]  Backend: an endpoint that when called deletes the current session. The call should be `DELETE /api/v1/authentication/session`. There should thus be an `api` package in the `authentication` package/module.
- [x] Frontend: a logout button that is visible from within every form, if the user is authenticated. Clicking this button should call the endpoint mentioned above. The button should be under the line with previous/next page. https://github.com/open-formulieren/open-forms-sdk/pull/49

sergei-maertens:
SDK PR open-formulieren/open-forms-sdk#49 was merged

PR #533, opened 2021-07-29
Feature #532 - Logout endpoint

SilviaAmAm:
Part of #532

### DIFF
--- a/src/openforms/authentication/api/views.py
+++ b/src/openforms/authentication/api/views.py
@@ -1,7 +1,8 @@
+from django.http import HttpResponse
 from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.utils import extend_schema, extend_schema_view
-from rest_framework import authentication, permissions
+from rest_framework import authentication, permissions, serializers, status
 from rest_framework.views import APIView
 
 from ...utils.api.views import ListMixin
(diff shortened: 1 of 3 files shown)


src/openforms/authentication/api/views.py
```diff
@@ -28,3 +30,20 @@ class PluginListView(ListMixin, APIView):
 
     def get_objects(self):
         return list(register)
+
+
+class AuthenticationLogoutView(GenericAPIView):
```
sergei-maertens, line 35: I think this should rather be a `rest_framework.views.APIView`

src/openforms/authentication/api/views.py
```diff
@@ -28,3 +30,20 @@ class PluginListView(ListMixin, APIView):
...
+    )  # just to shut up some warnings in drf-spectacular
+
+    @extend_schema(
+        summary=_("Delete session"),
+        description=_("Calling this endpoint will clear the current user session."),
+        responses={(204, "application/json"): ""},
+    )
+    def delete(self, request, *args, **kwargs):
+        request.session.flush()
+        return HttpResponse(status=status.HTTP_204_NO_CONTENT)
```
sergei-maertens, line 49:
```suggestion
    @extend_schema(
        summary=_("Delete session"),
        description=_("Calling this endpoint will clear the current user session and delete the session cookie."),
    )
    def delete(self, request, *args, **kwargs):
        request.session.flush()
        return HttpResponse(status=status.HTTP_204_NO_CONTENT)
```
For a `delete` call, the HTTP 204 is the default and doesn't need to be specified explicitly.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #539, opened 2021-07-30
[#273] Implemented payments plugins, with Ogone

Bartvaderkin:
Related to #273


src/openforms/payments/views.py
```diff
@@ -0,0 +1,306 @@
...
+            plugin_id,
+            # TODO pass amount from form
+            Decimal(10),
+            form_url,
+        )
+
+        info = plugin.start_payment(request, payment)
+        return Response(
+            self.serializer_class(instance=info, context={"request": request}).data
+        )
```
sergei-maertens, line 127: Since these are DRF views, please raise the appropriate exceptions from DRF instead of the bare-bones Django ones.



For example, `rest_framework.exceptions.ParseError(code="invalid-plugin", detail=_("Unknown plugin specified")`

src/openforms/payments/views.py
```diff
@@ -0,0 +1,306 @@
...
+        except KeyError:
+            return HttpResponseBadRequest("unknown plugin")
+
+        if not submission.form.payment_required:
+            return HttpResponseBadRequest("payment not required")
+
+        if plugin_id != submission.form.payment_backend:
+            return HttpResponseBadRequest("plugin not allowed")
+
+        form_url = request.GET.get("next")
```
sergei-maertens, line 105:
```suggestion
        form_url = request.query_params.get("next")
```
You're in the `def post` handler trying to access `.GET` - it'll work, but it's weird. DRF request objects have a `query_params` attribute to deal with this.

src/openforms/payments/views.py
```diff
@@ -0,0 +1,306 @@
...
+                "blocked payment start with non-allowed redirect to '%(form_url)s'",
+                {"form_url": form_url},
+            )
+            return HttpResponseBadRequest("redirect not allowed")
+
+        payment = SubmissionPayment.objects.create_for(
+            submission,
+            plugin_id,
+            # TODO pass amount from form
+            Decimal(10),
```
sergei-maertens, line 120: As a minimal implementation, allow people to specify the price as a decimal field on the form itself.

src/openforms/payments/views.py
```diff
@@ -0,0 +1,306 @@
...
+            form_url,
+        )
+
+        info = plugin.start_payment(request, payment)
+        return Response(
+            self.serializer_class(instance=info, context={"request": request}).data
```
sergei-maertens, line 126: since you're using `GenericAPIView`, I believe that `self.get_serializer(...)` and `self.get_serializer_context` are available which you can call.

src/openforms/payments/views.py
```diff
@@ -0,0 +1,306 @@
...
+    lookup_url_kwarg = "uuid"
+    queryset = SubmissionPayment.objects.all()
+
+    parser_classes = (FormParser, MultiPartParser)
+
+    def _handle_return(self, request, uuid: str):
+        """
+        Handle the return flow after the user provided payment credentials.
+
+        This can be either directly to us, or via an payment plugin. This
```
sergei-maertens, line 189:
```suggestion
        This can be either directly to us, or via a payment plugin. This
```

sergei-maertens (changes requested):
* `src/openforms/plugins/plugin.py` is an empty file, probably can be deleted?
* `src/openforms/payments/contrib/ogone/api/` -> seems to be unused and candidate for deletion.

src/openforms/payments/contrib/ogone/data.py
```diff
@@ -0,0 +1,91 @@
...
+
+    SHASIGN: str = ""
+
+    @classmethod
+    def from_dict(cls, value_dict):
+        kws = dict()
+        for f in dataclasses.fields(cls):
+            if f.name in value_dict:
+                kws[f.name] = value_dict[f.name]
+        return cls(**kws)
```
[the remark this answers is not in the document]
Bartvaderkin: Yes, this is to make sure we don't get errors if unmapped parameters show up.

src/openforms/payments/contrib/ogone/plugin.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@register("ogone-legacy")
+class OgoneLegacyPaymentPlugin(BasePlugin):
+    verbose_name = _("Ogone Legacy")
+    configuration_options = OgoneOptionsSerializer
+
+    def start_payment(self, request, payment: SubmissionPayment):
+        # decimal to cents
+        amount_cents = round(payment.amount * 100)
```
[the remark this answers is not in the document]
Bartvaderkin: Ogone needs an integer amount of cents, so we got to multiply our Decimal and convert to int. 

I changed it to use Decimal methods: `(payment.amount * 100).to_integral_exact()`

src/openforms/payments/contrib/ogone/plugin.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def apply_status(self, payment, ogone_status) -> None:
+        if payment.status in PaymentStatus.is_final:
+            # shouldn't happen or race-condition
+            return
+
+        new_status = OgoneStatus.as_payment_status(ogone_status)
+        if payment.status != new_status:
+            payment.status = new_status
+            payment.save()
```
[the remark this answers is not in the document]
Bartvaderkin: I rewrote it as an atomic queryset update() that includes the preconditions.

src/openforms/payments/models.py
```diff
@@ -0,0 +1,74 @@
...
+    def options(self):
+        # TODO we should store a copy of this
+        return self.submission.form.payment_backend_options
+
+    @property
+    def form(self):
+        return self.submission.form
+
+    class Meta:
+        unique_together = (("submission", "plugin_id", "order_id"),)
```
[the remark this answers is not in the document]
Bartvaderkin: This unique_together exists because the original order-ID was the submission UUID, and then I needed to add numbers to it to make them unique for Ogone when restarting the payment flow.

As we're going to use simpler order-ID's (unrelated to the submission UUID) we might actually want to make them globally unique.

sergei-maertens (approved):

PR #540, opened 2021-08-02
Feature #231 - Hide submission button

SilviaAmAm:
First part of #231


src/openforms/submissions/api/serializers.py
```diff
@@ -85,6 +85,12 @@ class SubmissionSerializer(serializers.HyperlinkedModelSerializer):
             "submission_uuid": "submission__uuid",
         },
     )
+    can_submit = serializers.BooleanField(
+        label=_("can submit"),
+        source="form.can_submit",
+        help_text=_("Whether the user is allowed to submit this form."),
+        required=False,
+    )
```
sergei-maertens, line 93: Can we make this field read-only? I'm worried now that if you POST to the submissions endpoint with `canSubmit: true|false`, that it will actually overwrite the value set on the form (because of the `form.can_submit` source).

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #529, opened 2021-07-27
KvK API Search got deprecated, we need to switch to KvK API Zoeken...


joeribekker:
See: https://developers.kvk.nl/nl/apis/zoeken

Bartvaderkin:
Note this API has a lot less information then the old one:

It is easy to see in the diff of the response mock: https://github.com/open-formulieren/open-forms/pull/543/files#diff-96760b963f93e9dad9c81bdfaa5b4b1e04f2d1a8bb1f339a3ce5642cd56a39d3

PR #543, opened 2021-08-05
[#529] Switched KvK validator & prefill to different API

Bartvaderkin:


src/openforms/prefill/contrib/kvk/plugin.py
```diff
@@ -46,13 +46,12 @@ def get_prefill_values(
...
+        items = results["resultaten"]
         if not items:
             return {}
 
         if len(items) > 1:
-            logger.warning("multiple results for")
-            return {}
+            logger.warning("multiple results for KvK search")
 
         data = items[0]
```
sergei-maertens, line 56: can we filter the items based on `type: hoofdvestiging` rather than just taking the first one?

sergei-maertens (changes requested):

src/openforms/prefill/contrib/kvk/plugin.py
Bartvaderkin: Sure, makes sense.

sergei-maertens (approved):

Issue #487, opened 2021-07-12
Add file-upload support to StufZDS registration backend


Bartvaderkin:
When registering the zaak also send `my_submission.get_attachments()` to the backend.
 
Based on #430 / #480

PR #548, opened 2021-08-06
[#487] Added support for attachments to Stuff ZDS registration backend

Bartvaderkin:


src/openforms/registrations/contrib/stuf_zds/client.py
```diff
@@ -287,6 +287,43 @@ def create_zaak_document(
...
+                "zaak_identificatie": zaak_id,
+                "document_identificatie": doc_id,
+                "titel": "bijlage",
+                "auteur": "open-forms",
+                "taal": "nld",
+                "inhoud": base64_body,
+                "status": "definitief",
+                "bestandsnaam": submission_attachment.get_display_name(),
+                "formaat": submission_attachment.get_format(),
+                "beschrijving": "Bijgevoegd document",
```
sergei-maertens, line 314: Fine for now, but I expect in the future there will be meta-data on the form field that shall serve as input for the attributes here, such as the label of the form field etc.

src/openforms/submissions/models.py
```diff
@@ -560,3 +560,6 @@ def delete(self, using=None, keep_parents=False):
 
     def get_display_name(self):
         return self.file_name or self.original_name
+
+    def get_format(self):
+        return os.path.splitext(self.get_display_name())[1].lstrip(".")
```
sergei-maertens, line 565: Shouldn't this be a mime type? e.g. `application/pdf`

sergei-maertens (commented):
One question about the format

src/openforms/submissions/models.py
Bartvaderkin: All examples are extensions but the spec indeed prefers mime type so I modified the PR.

sergei-maertens (approved):

Issue #486, opened 2021-07-12
Add file-upload support to ZGW registration backend


Bartvaderkin:
When registering the zaak also send `my_submission.get_attachments()` to the backend.

Based on #430 / #480

PR #549, opened 2021-08-06
[#486] Added support for attachments to ZGW registration backend

Bartvaderkin:


Bartvaderkin:
I tested this on test.openzaak.nl

sergei-maertens (approved):

PR #555, opened 2021-08-11
Feature #231 - Form logic rules (triggers + actions)

SilviaAmAm:
Related to #231


src/openforms/submissions/form_logic.py
```diff
@@ -0,0 +1,59 @@
...
+        return configuration
+
+    rules = FormLogic.objects.filter(form_step=step.form_step)
+    for rule in rules:
+        trigger = jsonLogic(rule.json_logic_trigger, data)
+        for action in rule.actions:
+            if action["type"] == "value":
+                configuration = set_property_value(
+                    configuration, rule.component, "value", trigger
+                )
```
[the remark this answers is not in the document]
sergei-maertens: Decided to separate trigger & action - the value action itself can be specified with JSON logic but we'll decouple that from the trigger JSON logic.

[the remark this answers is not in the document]
SilviaAmAm: After discussion: we decided that also for the action type `value` the trigger should give a `true` or `false` result, so that it is consistent with the other action types.

The actions of type `value` will contain some jsonLogic field that can be used to extract the value of another component *if* the trigger evaluates to `true`.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -238,6 +253,44 @@ function reducer(draft, action) {
...
             break;
         }
+        /**
+         * Form Logic rules actions
+         */
+        case 'LOADED_RULES': {
+            draft.rules.rules = action.payload.length ? action.payload : [{
+                ...EMPTY_RULE,
+                form: draft.form.url
+            }];
```
sergei-maertens, line 263: The empty-rule-by-default was mostly a debug thing so I had one less button to click. I don't think we should start with an empty rule, as saving a new form without focusing the logic tab would then lead to invalid rules being created and thus frustration for the end-user. Let's just keep an empty list as default.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -307,11 +360,88 @@ const getFormData = async (formUuid, dispatch) => {
...
+    try {
+        const response = await get(PREFILL_PLUGINS_ENDPOINT);
+        dispatch({
+            type: 'PREFILL_PLUGINS_LOADED',
+            payload: response.data
+        });
+    } catch (e) {
+        dispatch({type: 'SET_FETCH_ERRORS', payload: {loadingErrors: e.message}});
+    }
+};
```
sergei-maertens, line 428: This was refactored in the master branch, please check the current `useAsync` implementation with `loadPlugins`.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -307,11 +360,88 @@ const getFormData = async (formUuid, dispatch) => {
...
+
+const loadExistingRules = async (formUuid, dispatch) => {
```
sergei-maertens, line 430: This should also use the new data-loading format which is more declarative

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -442,6 +576,64 @@ const FormCreationForm = ({csrftoken, formUuid, formHistoryUrl }) => {
...
+    const onRuleDelete = (index) => {
+        dispatch({
+            type: 'DELETED_RULE',
+            payload: {index: index}
+        });
+    };
+
+    const onRuleAdd = () => {
+        dispatch({type: 'ADD_RULE'});
+    };
```
sergei-maertens, line 596: You can simplify this and inline it in the prop where it's used rather than defining another local function. That reduces visual clutter a bit.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -442,6 +576,64 @@ const FormCreationForm = ({csrftoken, formUuid, formHistoryUrl }) => {
...
+
+    const submitRules = async () => {
```
sergei-maertens, line 598: Can we move this into `data.js` and let it take some variables, so that you can call `submitRules(state.rules)` from the component and catch possible exceptions, to then relay all that to the dispatch? I really want to make this component less complex.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -442,6 +576,64 @@ const FormCreationForm = ({csrftoken, formUuid, formHistoryUrl }) => {
...
+        if (state.rules.rulesToDelete.length) {
+            for (const ruleUuid of state.rules.rulesToDelete) {
+                // Rules that were added but are not saved in the backend yet don't have a UUID.
+                if (!ruleUuid) return;
+
+                var response = await apiDelete(`${LOGICS_ENDPOINT}/${ruleUuid}`, csrftoken);
+                if (!response.ok) {
+                    throw new Error('An error occurred while deleting logic rules.');
+                }
+            }
```
sergei-maertens, line 633: A good candidate for `await Promise.all(...)` to do the HTTP requests in parallel rather than sequentially, improving the UX of end-users.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -615,7 +817,7 @@ const FormCreationForm = ({csrftoken, formUuid, formHistoryUrl }) => {
                 : null
             }
 
-            <Tabs>
+            <Tabs defaultIndex={5}>
```
sergei-maertens, line 820:
```suggestion
            <Tabs>
```
was only needed for development/debugging

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -638,6 +840,9 @@ const FormCreationForm = ({csrftoken, formUuid, formHistoryUrl }) => {
                     <Tab>
                         <FormattedMessage defaultMessage="Payment" description="Payment tab title" />
                     </Tab>
+                    <Tab>
+                        <FormattedMessage defaultMessage="Logics" description="Form logic tab title" />
```
sergei-maertens, line 844:
```suggestion
                        <FormattedMessage defaultMessage="Logic" description="Form logic tab title" />
```

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -731,6 +936,17 @@ const FormCreationForm = ({csrftoken, formUuid, formHistoryUrl }) => {
                         onChange={onFieldChange}
                     />
                 </TabPanel>
+
+                <TabPanel>
+                    <Fieldset title="Logica">
```
sergei-maertens, line 941: `title` Needs to be translatable

src/openforms/js/components/admin/form_design/logic/Action.js
```diff
@@ -0,0 +1,116 @@
+import React, {useContext} from 'react';
+import Select from '../../forms/Select';
+import {ACTION_TYPES, ACTIONS_WITH_OPTIONS, MODIFIABLE_PROPERTIES, PROPERTY_VALUES} from './constants';
+import ComponentSelection from './ComponentSelection';
+import LiteralValueInput from './LiteralValueInput';
+import {ComponentsContext} from './Context';
+import OperandTypeSelection from './OperandTypeSelection';
+import FAIcon from "../../FAIcon";
```
sergei-maertens, line 8:
```suggestion
import FAIcon from '../../FAIcon';
```
argh, we really should fix them all so that we can copy-paste without worry :)

src/openforms/js/components/admin/form_design/logic/Action.js
```diff
@@ -0,0 +1,116 @@
...
+            type: action.actionType,
+            property: {value: action.componentProperty},
+            // The data in 'value' needs to be valid jsonLogic
+            value: action.componentLiteralValue || {var: action.componentVariableValue},
+            state: action.componentPropertyValue,
+        }
+    };
+
+    const confirmDelete = (index) => {
+        if(window.confirm('Are you sure you want to delete this action?')){
```
sergei-maertens, line 27: should be translatable

src/openforms/js/components/admin/form_design/logic/Action.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+    };
+
+
+    return (
+        <div className="action">
+            <div className="actions">
+                <FAIcon icon="trash" extraClassname="icon icon--danger actions__action" title="Delete" onClick={confirmDelete} />
+            </div>
+            Then,
```
sergei-maertens, line 38: Should translate these too

src/openforms/js/components/admin/form_design/logic/Action.js
```diff
@@ -0,0 +1,116 @@
...
+            </div>
+            Action:
+            <div style={{background: '#eee', border: 'dashed 1px #ccc', 'marginTop': '1em'}}>
+                <pre>{JSON.stringify(jsonAction, null, 2)}</pre>
+            </div>
+        </div>
+    );
+};
+
+export {Action};
```
sergei-maertens, line 116:
```suggestion
export default Action;
```
The idea is that every JS file is a single component, which makes it the default export by definition.

This follows the Create React App boilerplate pattern.

src/openforms/js/components/admin/form_design/logic/ActionSet.js
```diff
@@ -0,0 +1,177 @@
+import React, {useEffect} from 'react';
+import {Action} from './Action';
+import {useImmerReducer} from 'use-immer';
+import usePrevious from 'react-use/esm/usePrevious';
+import isEqual from 'lodash/isEqual';
+import PropTypes from "prop-types";
```
sergei-maertens, line 6:
```suggestion
import PropTypes from 'prop-types';
```

src/openforms/js/components/admin/form_design/logic/ActionSet.js
```diff
@@ -0,0 +1,177 @@
...
+import Trigger from "./Trigger";
```
sergei-maertens, line 7: I wish we had `isort` for JS

Please group third-party dependencies at the top, follow by a blank line and then own-component imports.

src/openforms/js/components/admin/form_design/logic/ActionSet.js
```diff
@@ -0,0 +1,177 @@
...
+                value: action.componentLiteralValue || {var: action.componentVariableValue},
+                state: action.componentPropertyValue,
+                // Selecting if the new component value should come from a literal or another component doesn't
+                // change anything in the JSON, only entering the literal or picking the component does.
+                // So, this 'source' attribute is here only to keep track of the value for the OperandTypeSelection
+                // dropdown, but is not used in the backend
+                source: action.componentValueSource
+        }
+    };
+};
```
sergei-maertens, line 83: Indentation is a bit messed up here :)

src/openforms/js/components/admin/form_design/logic/ActionSet.js
```diff
@@ -0,0 +1,177 @@
...
+
+const parseJsonAction = (jsonAction) => {
+    let componentValueSource, componentLiteralValue, componentVariableValue;
+    if (jsonAction.action.value.var === undefined) {
```
sergei-maertens, line 87:
```suggestion
    if (jsonAction.action.value.var == null) {
```
this is a weird javascript thing where it will also match `undefined` (if you use the non-strict comparison), but a typical pattern. It catches both `undefined` and `null`.

src/openforms/js/components/admin/form_design/logic/ActionSet.js
```diff
@@ -0,0 +1,177 @@
...
+            {state.actions.map((action, index) => (
+                <Action
+                    key={index}
+                    action={action}
+                    onChange={onActionChange.bind(null, index)}
+                    onDelete={onActionDelete.bind(null, index)}
+                />
+            ))}
+            <button type="button" onClick={ () => dispatch({type: 'ACTION_ADDED'}) }>
+                Add action
```
sergei-maertens, line 165: Should be translatable, but probably we need to look at other places where similar buttons/actions are used to have  a consistent UI

src/openforms/js/components/admin/form_design/logic/ActionSet.js
```diff
@@ -0,0 +1,177 @@
...
+    );
+};
+
+ActionSet.propTypes = {
+    name: PropTypes.string.isRequired,
+    actions: PropTypes.arrayOf(PropTypes.object).isRequired,
+    onChange: PropTypes.func.isRequired,
+};
+
+export {ActionSet};
```
sergei-maertens, line 177:
```suggestion
export default ActionSet;
```

src/openforms/js/components/admin/form_design/logic/constants.js
```diff
@@ -0,0 +1,76 @@
...
+const MODIFIABLE_PROPERTIES = [
+    ['required', 'required'],
+    ['hidden', 'hidden'],
+    ['disabled', 'disabled']
+];
+
+
+const PROPERTY_VALUES = [
+  ['true', 'Yes'],
+  ['false', 'No']
```
sergei-maertens, line 65: Can we use actual booleans here so that they're properly serialized in JSON too?

sergei-maertens (changes requested):
Can you please pay attention to test structure? An explicit block structure like the following is recommended:

```
# test setup - factories, logging in...
data = Factory.create()

# calling the thing you're testing
result = my_functionality_i_am_testing()

# making assertions
assert result is not None
assert len(result) == 2
```

Try to avoid blank lines between assertions and data-prep lines, and ensure the blocks are visually separated. That makes it easier to see what is happening in a test.

sergei-maertens:
The rest of feedback will be processed in another PR

PR #564, opened 2021-08-16
:sparkles: fix-login-page-using-new-2fa-style

### LINKED ISSUE #473: As user I want a secure admin

azertyValentin:
Fixes #473 and replaces #477

fix-login-page-using-new-2fa-style


src/openforms/conf/base.py
```diff
@@ -794,3 +794,6 @@
 OPENFORMS_LOCATION_CLIENT = config(
     "OPENFORMS_LOCATION_CLIENT", "openforms.contrib.bag.client.BAGClient"
 )
+
+
+ADMIN_DISPLAY_DROP_DOWN_MENU_CONDITION_FUNCTION = "openforms.utils.django-two-factors.should_display_dropdown_menu"
```
sergei-maertens, line 799: Please group this setting with the other admin index setting(s).

Also, in admin index, the prefix should be `ADMIN_INDEX` and not just `ADMIN`, so please update that PR as well.

sergei-maertens: `django-two-factors` is not a valid python identifier -> use underscores

And rather put this in `src/openforms/utils/django_two_factor_auth.py`

src/openforms/utils/django-two-factors.py
```diff
@@ -0,0 +1,5 @@
+from django.utils.module_loading import import_string
+
+
+def should_display_dropdown_menu(request):
+    return import_string("django_admin_index.utils.should_display_dropdown_menu")(request) and request.user.is_verified()
```
sergei-maertens, line 5: This code is barely readable and there's no reason to do dynamic imports here

```suggestion
from django_admin_index.utils import should_display_dropdown_menu as default_should_display_dropdown_menu

def should_display_dropdown_menu(request) -> bool:
    return default_should_display_dropdown_menu(request) and request.user.is_verified()
```

sergei-maertens (changes requested):
I'm missing the dependency update for django-admin-index, which has the `ADMIN_DISPLAY_DROP_DOWN_MENU_CONDITION_FUNCTION` setting

requirements/base.txt
```diff
@@ -189,7 +189,8 @@ lxml==4.6.3
     # via
     #   -r requirements/base.in
     #   xmlsec
-maykin-django-two-factor-auth[phonenumbers]==2.0.2
+#maykin-django-two-factor-auth[phonenumbers]==2.0.2
+git+https://github.com/maykinmedia/django-two-factor-auth/@task-334-fix-login-page-style#egg=django-two-factor-auth
```
sergei-maertens, line 193: Never update `base.txt` directly, you should be using the `pip-compile` tooling. See the header of this file:

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    ./bin/compile_dependencies.sh
#
```

sergei-maertens (changes requested):

src/openforms/conf/base.py
azertyValentin: done

src/openforms/utils/django-two-factors.py
azertyValentin: done

requirements/base.txt
azertyValentin: done

sergei-maertens (approved):

sergei-maertens:
Closing in favour of #568

PR #571, opened 2021-08-19
Ensure voornamen attribute is in request soap when it's an attribute

### LINKED ISSUE #567: Voornaam komt niet in StUF-BG scope ondanks het configureren van Voornaam prefill
Het voornaam attribuut wordt niet in `<scope>` in het XML bericht opgenomen.

```
            <ns:scope>
                <ns:object StUF:entiteittype="NPS">
                    <ns:inp.bsn xsi:nil="true"/>
                    <ns:geslachtsnaam xsi:nil="true"/>
                    <ns:verblijfsadres>
                        <ns:gor.straatnaam xsi:nil="true"/>
                    </ns:verblijfsadres>
                </ns:object>
            </ns:scope>
```

shea-maykinmedia:
Fixes #567 

The cause of the bug was that the xml template used when making request had the conditional `{% if voornaam %}` while the attribute was named `voornamen` so even when voornamen was an attribute the conditional would always be false and so it was not added.

**Screenshot** 

Added every possible Stuf-BG attribute and they all prefilled.  This address does not have a house letter or house number addition which is why these fields are blank.

Screenshot 2021-08-19 at 12 02 21

### DIFF
--- a/src/openforms/prefill/contrib/stufbg/plugin.py
+++ b/src/openforms/prefill/contrib/stufbg/plugin.py
@@ -69,8 +69,8 @@ class StufBgPrefill(BasePlugin):
 
         response_dict = {}
         for attribute in attributes:
-            response_dict[attribute] = glom(
-                data, ATTRIBUTES_TO_STUF_BG_MAPPING[attribute], default=None
-            )
+            value = glom(data, ATTRIBUTES_TO_STUF_BG_MAPPING[attribute], default=None)
+            if "@noValue" not in value:
+                response_dict[attribute] = value
 
         return response_dict
--- a/src/stuf/stuf_bg/tests/test_client.py
+++ b/src/stuf/stuf_bg/tests/test_client.py
@@ -121,3 +121,6 @@ class StufBGConfigTests(TestCase):
                             f'Attributes "{subset}" produces an invalid StUF-BG. '
                             f"Error: {xmlschema.error_log.last_error.message}"
                         )
+                    for attribute in subset:
+                        with self.subTest(subset=subset, attribute=attribute):
+                            self.assertIn(attribute, data)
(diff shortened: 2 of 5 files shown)


src/openforms/prefill/contrib/stufbg/plugin.py
```diff
@@ -69,8 +70,8 @@ def get_prefill_values(
 
         response_dict = {}
         for attribute in attributes:
-            response_dict[attribute] = glom(
-                data, ATTRIBUTES_TO_STUF_BG_MAPPING[attribute], default=None
-            )
+            value = glom(data, ATTRIBUTES_TO_STUF_BG_MAPPING[attribute], default=None)
+            if "@noValue" not in value:
```
shea-maykinmedia, line 74: This change is unrelated to the ticket but is another bug I found so I wanted to fix it while I was looking at this code.  If a request attribute does not have a value then stuf will return something like

```
<ns:aoa.huisletter StUF:noValue="geenWaarde" xsi:nil="true" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"/>
```

which at this point will be transform into a dictionary

```
OrderedDict([('@noValue', 'geenWaarde'), ('@http://www.w3.org/2001/XMLSchema-instance:nil', 'true'), ('@xmlns', OrderedDict([('xsi', 'http://www.w3.org/2001/XMLSchema-instance')]))])
```

and without this check it would be returned to the SDK which would then display `[object Object]` in the input field.

Note that even numeric attributes like house number will be strings (eg. '2' and not 2) at this point so this will still work with those properly.

src/stuf/stuf_bg/tests/test_client.py
```diff
@@ -121,3 +121,6 @@ def test_getting_request_data_returns_valid_data(self):
                             f'Attributes "{subset}" produces an invalid StUF-BG. '
                             f"Error: {xmlschema.error_log.last_error.message}"
                         )
+                    for attribute in subset:
+                        with self.subTest(subset=subset, attribute=attribute):
+                            self.assertIn(attribute, data)
```
shea-maykinmedia, line 126: This seemed like the best way to unit test against a bug like this happening.  The reason this test didn't fail before is the `voornamen` attribute would not be added but since it was still valid against the xsd the test would pass.

This will at least ensure the attribute is in the xml we will be using when making the request.

sergei-maertens (approved):

joeribekker:
Nice work @shea-maykinmedia

PR #573, opened 2021-08-19
JCC-afspraken client/plugin

### LINKED ISSUE #276: Koppeling met JCC Afspraken t.b.v. afspraak maken
Eis: Koppelingen 5g.

**Acties van gemeenten**

* [ ] Aanpak bespreken (via servicebus achteraf, via formulier direct)
* [ ] Voorstel om afspraak formulier te laten bestaan uit vaste stappen goedkeuren
* [x] Testomgeving nodig
* [x] Documentatie nodig

**Verduidelijking**

Example can be seen here: https://www.horstaandemaas.nl/afspraak-maken/formulier?product_code=1

**Tasks**

* [x] Implement JCC Afspraken client
* [x] In the admin, allow JCC-afspraken client to be configured.
* [x] In the admin, add appointment tab to admin form view, which allows it to be enabled, backend to be selected, product and language settings.
* [x] Form steps should be added to render the form.

joeribekker:
Fixes #276 

Note some improvements that will be made in the future https://github.com/open-formulieren/open-forms/issues/597

SDK pull request https://github.com/open-formulieren/open-forms-sdk/pull/55

**Screenshots**

Screenshot 2021-08-30 at 11 16 42
Screenshot 2021-08-30 at 11 26 22
Screenshot 2021-08-30 at 11 14 31
Screenshot 2021-08-30 at 11 14 21
Screenshot 2021-08-30 at 11 12 54
Screenshot 2021-08-30 at 11 12 43
Screenshot 2021-08-30 at 11 12 28


src/openforms/appointments/api/tests/test_endpoints_with_jcc.py
```diff
@@ -0,0 +1,211 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.json(),
+            ["2021-08-19", "2021-08-20", "2021-08-23"],
+        )
+
+    def test_get_dates_returns_400_when_missing_query_params(self):
+        for query_param in ["", "?product_id=79", "?location_id=1"]:
+            with self.subTest(query_param=query_param):
+                response = self.client.get(self.endpoint)
```
sergei-maertens, line 154: For the future, please use the dict-structure for query params in tests:

```py
self.client.get(self.endpoint, {"product_id": 79})
```

and the test client will format the querystring itself.

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+            type: 'select',
+            input: true,
+            label: 'Select product component for locations',
+            key: 'productForLocations',
+            dataSrc: 'custom',
+            tooltip: 'Choose the product component and we wil prefill locations for that product in this component',
+            valueProperty: 'value',
+            data: {
+                custom(context) {
+                    return Utils.getContextComponents(context);
```
sergei-maertens, line 60: can we filter this down to only `type: select` components?

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+            type: 'select',
+            input: true,
+            label: 'Select product component for dates',
+            key: 'productForDates',
+            dataSrc: 'custom',
+            tooltip: 'Choose the product component and we wil prefill dates for that product in this component',
+            valueProperty: 'value',
+            data: {
+                custom(context) {
+                    return Utils.getContextComponents(context);
```
sergei-maertens, line 80: can we filter this down to only `type: select` components?

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+            type: 'select',
+            input: true,
+            label: 'Select location component for dates',
+            key: 'locationForDates',
+            dataSrc: 'custom',
+            tooltip: 'Choose the location component and we wil prefill dates for the location and product in this component',
+            valueProperty: 'value',
+            data: {
+                custom(context) {
+                    return Utils.getContextComponents(context);
```
sergei-maertens, line 94: can we filter this down to only `type: select` components?

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+            type: 'select',
+            input: true,
+            label: 'Select product component for times',
+            key: 'productForTimes',
+            dataSrc: 'custom',
+            tooltip: 'Choose the product component and we wil prefill times for that product in this component',
+            valueProperty: 'value',
+            data: {
+                custom(context) {
+                    return Utils.getContextComponents(context);
```
sergei-maertens, line 114: can we filter this down to only `type: select` components?

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+            type: 'select',
+            input: true,
+            label: 'Select location component for dates',
+            key: 'locationForTimes',
+            dataSrc: 'custom',
+            tooltip: 'Choose the location component and we wil prefill times for that location in this component',
+            valueProperty: 'value',
+            data: {
+                custom(context) {
+                    return Utils.getContextComponents(context);
```
sergei-maertens, line 128: can we filter this down to only `type: select` components?

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+            type: 'select',
+            input: true,
+            label: 'Select location component for dates',
+            key: 'dateForTimes',
+            dataSrc: 'custom',
+            tooltip: 'Choose the date component and we wil prefill times for that date in this component',
+            valueProperty: 'value',
+            data: {
+                custom(context) {
+                    return Utils.getContextComponents(context);
```
sergei-maertens, line 142: can we filter this down to only `type: select` components?

src/openforms/js/components/form/select.js
```diff
@@ -1,5 +1,175 @@
...
+}];
+
+
+const APPOINTMENT = {
+    key: 'appointment',
+    label: 'Appointment',
+    components: [
+        {
+            type: 'checkbox',
+            key: 'showProducts',
```
sergei-maertens, line 40: I'd like to see all the appointment configuration namespaced in a nested object to prevent (future) key collisions

```suggestion
            key: 'appointments.showProducts',
```
Of course this applies for the other components too below this one :-)

src/openforms/js/components/form/textfield.js
```diff
@@ -2,6 +2,20 @@ import {Formio} from 'react-formio';
 
 import {DEFAULT_TEXT_TABS, PREFILL} from './edit/tabs';
 
+const APPOINTMENT = {
+    key: 'appointment',
+    label: 'Appointment',
+    components: [
+        {
+            type: 'checkbox',
+            key: 'appointmentLastName',
```
sergei-maertens, line 11: would also namespace this as with the `Select` component

src/openforms/js/formio_module.js
```diff
@@ -8,6 +8,7 @@ import PhoneNumberField from './components/form/phoneNumber';
 import BsnField from './components/form/bsn';
 import PostcodeField from "./components/form/postcode";
 import FileField from "./components/form/file";
+import SelectField from "./components/form/select";
```
sergei-maertens, line 11:
```suggestion
import SelectField from './components/form/select';
```

src/openforms/submissions/api/viewsets.py
```diff
@@ -139,6 +140,8 @@ def on_submission_commit():
 
         transaction.on_commit(on_submission_commit)
 
+        book_appointment_for_submission(submission)
```
sergei-maertens, line 143: Does this *need* to run in the request response cycle? Or could we off-load this to a celery task (in the above `on_submission_commit`?

It looks like the error-handling is already done as part of this function, so even if there is not a successful appointment registration, the end-user flow will not be interrupted now.

src/openforms/submissions/models.py
```diff
@@ -265,6 +272,41 @@ def get_merged_data_with_component_type(self) -> dict:
...
+            "appointmentLastName": "clientLastName",
+            "appointmentBirthDate": "clientDateOfBirth",
+        }
+
+        for step in self.submissionstep_set.exclude(data=None).select_related(
+            "form_step"
+        ):
+            components = step.form_step.form_definition.configuration["components"]
+            flattened_components = get_flattened_components(components)
+            for component in flattened_components:
```
sergei-maertens, line 291: doesn't it make more sense to use the `form.iter_components()` here instead, so you don't really have to concern yourself with the model relations?

src/openforms/submissions/models.py
```diff
@@ -265,6 +272,41 @@ def get_merged_data_with_component_type(self) -> dict:
 
         return merged_data
 
+    def get_merged_appointment_data(self) -> dict:
```
sergei-maertens, line 275: I'd prefer to see detailed type hints in the form of `Dict[str, ???]` so I know what types the key/values have.

sergei-maertens (changes requested):
* I am missing the access control tests -> tests that verify that the endpoints are not allowed if you don't have an active submission.
* #612 is created to refactor some UI for the appointments

src/openforms/appointments/api/views.py
```diff
@@ -0,0 +1,165 @@
...
+        OpenApiParameter(
+            "location_id",
+            OpenApiTypes.STR,
+            OpenApiParameter.QUERY,
+            description=_("Id of the location"),
+            required=True,
+        ),
+    ],
+)
+class DatesListView(APIView):
```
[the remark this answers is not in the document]
shea-maykinmedia: > Shouldn't this also have the ListMixin?

Since this is returning a list of dates (as strings) I don't think that using a serializer and ListMixin is necessary.  We _could_ do this so that the response is something like `[{'date': '2021-08-23'}, {'date': '2021-08-24'}, ...]` but I think the current response of `['2021-08-23', '2021-08-24', ...]` makes more sense.

I will update the API schema though.

Same with the TimesListView.

src/openforms/appointments/utils.py
```diff
@@ -1,8 +1,86 @@
...
+    try:
+        client = get_client()
+        appointment_id = client.create_appointment(
+            [product], location, start_at, appointment_client
+        )
+        submission.appointment_information = f"Appointment Id: {appointment_id}"
+    except AppointmentCreateFailed as e:
+        submission.appointment_information = "Failed to make appointment"
+    finally:
+        submission.save()
```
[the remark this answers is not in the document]
shea-maykinmedia: I did think about this and maybe this makes more sense but it also feels a bit like overkill.

Either `appointment_id` or `error_information` will be filled in and not both.  Also if there is an `appointment_id` we know it's successful and if there isn't then we know it's a failure so the `status` wouldn't really be needed so you'd really just have a model with two fields but only one or the other would be filled in.  Which is why I felt it may be overkill and we could just put the relevant info into a single field.

I am fine though if you think a one to one model is still a better solution but I'm curious what your thoughts are after reading the above?

src/openforms/js/formio_module.js
shea-maykinmedia: Since we're using both single and double quotes in this file I wasn't sure which was the correct :p. I think it's worth changing all of them to single quotes in this file so they're all consistent.

src/openforms/submissions/api/viewsets.py
shea-maykinmedia: > Does this need to run in the request response cycle? Or could we off-load this to a celery task (in the above on_submission_commit?

Good question :) The reason I did it like this is that in the next couple of stories related to the appointments we need to pass the appointment identifier (if there is one) so that we can add information about the appointment to the confirmation email.  This means this function needs to run before `send_confirmation_email`.  

We could off-load this to a celery task but then we'll need to do the same with the confirmation email.  If we do want to off-load both of these then we can though I wonder if it's better to do as part of this pull request or add this to one of the "improvement" tickets related to this.

src/openforms/js/components/form/select.js
shea-maykinmedia: Not as far as I can tell.  I believe this is the function being called https://github.com/formio/formio.js/blob/c94992cacc50924d277132045dee0245a23e44b0/src/utils/utils.js#L1113-L1126 and it doesn't look like there is a way to either specify the type in the context or filter the values this function returns since it returns a list of objects with just a `label` and `value`.

If there is a way though it would be good to do :)

sergei-maertens (approved):

sergei-maertens:
We'll get this merged when the build passes, and for #614 the necessary refactor will be done in a separate branch. Shea and I have coordinated the actions for this.

PR #583, opened 2021-08-20
Issue #575 - Select boxes rendering

### LINKED ISSUE #575: Rendering of Select Boxes in PDF summary
Currently, the select boxes appear like a python dictionary in the submission confirmation PDF summary. 
It also seems that if the table is too wide, it runs off the page.
Screenshot from 2021-08-20 12-50-04
More details can be found in issue 15 in Taiga

SilviaAmAm:
Fixes #575 

In the same way that the files attachment have a special treatment for rendering them in the report, now also the select boxes have a special treatment.
I also updated the submission report template to prevent the table with the form data to run outside the page.
Note: If the label/value from a field are an extremely long word, this will still render in an ugly way. It is due to this bug in weasyprint: https://github.com/Kozea/WeasyPrint/issues/36

### DIFF
--- a/src/openforms/submissions/models.py
+++ b/src/openforms/submissions/models.py
@@ -4,7 +4,7 @@ import uuid
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import date, timedelta
-from typing import Any, List, Mapping, Optional, Tuple
+from typing import Any, Dict, List, Mapping, Optional, Tuple
 
 from django.contrib.postgres.fields import JSONField
 from django.core.files.base import ContentFile, File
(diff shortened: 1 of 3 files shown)


src/openforms/submissions/models.py
```diff
@@ -253,6 +255,30 @@ def get_merged_data(self) -> dict:
 
         return merged_data
 
+    def get_printable_data(self) -> Dict[str, str]:
```
sergei-maertens, line 258: Fine for now, but if we keep expanding on this, we might have to just register a display/print function which takes the `info` argument and have a mapping of `type: print_function` so that can be tested/maintained in a bit more isolated way.

sergei-maertens (approved):

Issue #576, opened 2021-08-20
Title of checkboxes cut off


SilviaAmAm:
When the user is filling in the form, if the title of a field is too long it is cut off.
![afbreken tekst radiobutton](https://user-images.githubusercontent.com/19154114/130224265-8b15e341-e092-4aa5-9df8-2ded5bf6a12f.png)

PR #584, opened 2021-08-20
Style the form logic editor

sergei-maertens:
Relates to #231

Turned the developer styling into end-user styling.

Screenshot:
image


src/openforms/js/components/admin/form_design/logic/Trigger.js
```diff
@@ -197,64 +202,62 @@ const Trigger = ({ name, logic, onChange }) => {
...
-                            selectedComponent={triggerComponent}
-                            operator={operator}
+        <div className="logic-trigger">
+            <div className="logic-trigger__editor">
+                <div className="dsl-editor">
+                    <div className="dsl-editor__node">
+                        <FormattedMessage description="Logic trigger prefix" defaultMessage="When" />
+                    </div>
+                    <div className="dsl-editor__node">
+                        <ComponentSelection
```
SilviaAmAm, line 220: I noticed that if we have components with a long label, the dropdown becomes very long (I had one of these components because I was looking at bug https://github.com/open-formulieren/open-forms/issues/576 :smiley: )
Screenshot from 2021-08-23 10-20-55

SilviaAmAm (changes requested):

PR #591, opened 2021-08-23
Feature #294 - DOB checks

### LINKED ISSUE #294: Als minderjarige kan ik geen verhuizing doorgeven op basis van een uitgevoerde leeftijdscheck.
Eis: Authenticatie En Basisgegevens 3d.
Zie ook NvI: 27, 176

This also depends on "stopping" a form, as described in #233 

The idea is that a logic rule can be defined that says: if the date of birth (from prefill) is greater than `$currentDate - 18 years` -> then block the submit button.

**Tasks**
- [ ] Ensure we can use nested jsonLogic triggers
- [x] Add jsonlogic operator to get the current date
- [x] Add jsonlogic operator to compare two dates

SilviaAmAm:
Fixes #294 

In the end I changed the way the jsonLogic is validated. I left it as it was before (i.e. `jsonLogic(expression)` is evaluated and if it raises an error then the jsonLogic is considered invalid), but modified what happens if the operands evaluate to `None` ( as in the case where no data/incomplete data is passed along with the expression). Now `jsonLogic(expression)` will return None if the operands of some operators are None: https://github.com/maykinmedia/json-logic-py/pull/4

### DIFF
--- a/src/openforms/utils/tests/test_json_logic.py
+++ b/src/openforms/utils/tests/test_json_logic.py
@@ -1,5 +1,7 @@
 from django.test import SimpleTestCase
 
+from json_logic import jsonLogic
+
 from ..json_logic import JsonLogicTest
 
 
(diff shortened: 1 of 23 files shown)


src/openforms/utils/tests/test_json_logic.py
```diff
@@ -1,5 +1,7 @@
 from django.test import SimpleTestCase
 
+from json_logic import jsonLogic
```
sergei-maertens, line 3: is this import used or not?

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #602, opened 2021-08-30
Feature #183 - More form logic

### LINKED ISSUE #183: Een gebruiker van de toepassing kan zonder tussenkomst van de leverancier logica toekennen aan een element of ...
...veld op basis van data en getallen. Dat betekent per veld of selectie van velden: een minimale en / of maximale negatieve en positieve waarde in getallen tot 2 cijfers achter de komma, een bereik in data (bijvoorbeeld x periode in dagen of weken vooruit of achteruit in de tijd en tijdstippen die aan elkaar gerelateerd kunnen worden. Indien niet wordt voldaan aan de logica verschijnt een voor de inwoner of ondernemer duidelijke melding in beeld waarin de verwachtte waarde wordt weergegeven.

Eis: Algemeen 9.
Zie ook NvI: 9, 47, 66

**Verduidelijking**

Heeft te maken met... vergunning aanvragen, dan moet aanvraagdatum bijv. tussen bepaalde data liggen. Ik zoek een paar voorbeelden van wat mogelijk moet zijn (velden die met elkaar praten). Datum velden die aan elkaar gekoppeld zijn.

Bespreek met Heda na 14 juli.

**Tasks**

- [x] Add "(n.v.t.)" to the step title whenever a step is not needed due to decisions/logic (progress indicator in SDK, split off from #179)
* [x] Datum voor/na/op datum
* [x] Datum x is maximaal +y van datum z
* [x] Tekst bevat/is

SilviaAmAm:
Fixes #183 

Todo: 
- [x] Update json-logic-py to deal with empty string dates and dates that are actually datetimes: https://github.com/maykinmedia/json-logic-py/pull/5
- [x] Figure out how the logic check on one step can affect another step. For example: If a user enters the date of birth and they are younger than 18, how can you directly mark a later step on 'Drivers license' as not applicable?
- [x] Modify SDK so that if a step is marked as 'not applicable' there is '(n.v.t)' in the title.


src/openforms/forms/api/serializers.py
```diff
@@ -392,6 +393,15 @@ class LogicComponentActionSerializer(serializers.Serializer):
             "optional if the action type is `{action_type}`, otherwise required."
         ).format(action_type=LogicActionTypes.disable_next),
     )
+    form_step = serializers.CharField(
```
sergei-maertens, line 396: Shouldn't this rather be a `HyperlinkedRelatedField` which takes a URL reference and is resolved to the form step instance?

Mostly reasoning from consistency with other endpoints/serializers where URL references are used throughout the API.

src/openforms/js/components/admin/form_design/logic/OperandTypeSelection.js
```diff
@@ -8,33 +8,19 @@ import Select from '../../forms/Select';
...
+const OperandTypeSelection = ({name, operandType, onChange, filter}) => {
     const intl = useIntl();
-    const choices = Object.entries(COMPONENT_TYPE_TO_OPERAND_TYPE[componentType]);
+    const choices = getTranslatedChoices(intl, OPERAND_TYPES);
+
     return (
         <Select
             name={name}
-            choices={getTranslatedChoices(intl, choices)}
+            choices={filter ? choices.filter(filter) : choices}
```
sergei-maertens, line 23: I'd specify a default value for `filter` which is `allowAny` (`const allowAny = () => true;`) so that you can _always_ call the `filter` function provided.

Usually it's best practice to provide a default value for props that are not required.

src/openforms/js/components/admin/form_design/logic/StepSelection.js
```diff
@@ -0,0 +1,28 @@
+import React, {useContext} from 'react';
+import PropTypes from 'prop-types';
+
+import Select from "../../forms/Select";
```
sergei-maertens, line 4:
```suggestion
import Select from '../../forms/Select';
```

src/openforms/js/components/admin/form_design/logic/Today.js
```diff
@@ -9,29 +9,35 @@ import {NumberInput} from '../../forms/Inputs';
 import {getTranslatedChoices} from '../../../../utils/i18n';
 
 
+const EMTPTY_RELATIVE_DELTA = [0, 0, 0];
```
sergei-maertens, line 12:
```suggestion
const EMPTY_RELATIVE_DELTA = [0, 0, 0];
```

src/openforms/js/components/admin/form_design/logic/Today.js
```diff
@@ -9,29 +9,35 @@ import {NumberInput} from '../../forms/Inputs';
...
+
+
 const Today = ({name, value, onChange}) => {
     const sign = value ? jsonLogic.get_operator(value) : '+';
-    const years = value ? value[sign][1]['years'] : 0;
+    var rdelta = value && value[sign][1]?.rdelta ? value[sign][1].rdelta : EMTPTY_RELATIVE_DELTA;
+    if (rdelta.length < 3) {
+        rdelta = [...rdelta].concat(EMTPTY_RELATIVE_DELTA).slice(0, 3);
```
sergei-maertens, line 19: This sort of assumes implicitly that `EMPTY_RELATIVE_DELTA` consists of zeroes - which is the case but it's weird to use part of an array. Maybe I'd rather use [`Array.fill`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) instead of this concat + slice setup.

Also - you could do `[...rdelta, ...EMPTY_RELATIVE_DELTA]` for the same concat results, FYI

src/openforms/submissions/api/viewsets.py
```diff
@@ -288,8 +293,18 @@ def update(self, request, *args, **kwargs):
...
+        description=_("Apply/check the logic rules specified on the form."),
         request=FormDataSerializer,
+        responses={
+            200: inline_serializer(
+                "SubmissionStateLogicSerializer",
+                {
+                    "submission": SubmissionSerializer(),
+                    "step": SubmissionStepSerializer(),
+                },
+            )
```
sergei-maertens, line 306: Please define an explicit serializer for this. Can just be a subclass of `serializers.Serializer` and you pass it a dict as instance, that should work.

src/openforms/submissions/api/viewsets.py
```diff
@@ -304,7 +319,18 @@ def logic_check(self, request, *args, **kwargs):
...
+            evaluate_form_logic(
+                submission_step.submission, submission_step, merged_data
+            )
 
-        serializer = self.get_serializer(instance=submission_step)
-        return Response(serializer.data)
+        submission_serializer = SubmissionSerializer(
+            instance=submission_step.submission, context={"request": request}
+        )
+        submission_step_serializer = self.get_serializer(instance=submission_step)
```
sergei-maertens, line 329: `get_serializer_class` can be implement to look at `self.action` and return the appropriate serializer class based on that.

src/openforms/submissions/models.py
```diff
@@ -62,10 +61,20 @@ def get_next_step(self) -> Optional["SubmissionStep"]:
         candidates = (
             step
             for step in self.submission_steps[offset:]
-            if not step.completed and step.available
+            if not step.completed and step.is_applicable
         )
         return next(candidates, None)
 
+    def get_submission_step(self, form_step_uuid: str) -> "SubmissionStep":
```
sergei-maertens, line 68:
```suggestion
    def get_submission_step(self, form_step_uuid: str) -> Optional["SubmissionStep"]:
```
if there's a chance that `None` is returned, type hint should indicate this.

sergei-maertens (changes requested):

src/openforms/forms/api/serializers.py
SilviaAmAm: I think that with a `HyperlinkedRelatedField` the form steps resolves to the `FormStep` object and then since the action is a `JSONField`  we get an error that it is not serializable. Maybe I should make it a URL field with a validate method that checks that it is an existing formstep?

src/openforms/js/components/admin/form_design/logic/StepSelection.js
SilviaAmAm: aaargh quotes!!!

src/openforms/submissions/api/viewsets.py
SilviaAmAm: I couldn't get it to work with the dict as instance, so I added a dataclass instead

src/openforms/submissions/api/viewsets.py
SilviaAmAm: This part of the code is no longer present because of the explicit serializer (see comment https://github.com/open-formulieren/open-forms/pull/602#discussion_r702941366)

src/openforms/forms/api/serializers.py
sergei-maertens: cool, yeah, I figured that would've been the problem

sergei-maertens (approved):

PR #609, opened 2021-08-31
Feature/appointment changing and cancelling

### LINKED ISSUE #263: De gebruiker kan een afspraak annuleren of wijzigen via een url in een bevestigings e-mail.
Eis: Afspraken 3.
Zie ook NvI: 60, 92

**Tasks**

* [x] Add API endpoint to cancel appointment.
* [ ] ~Add API endpoint to change the appointment.~
* [x] The API endpoints should take an appointment ID and an emailaddress as validation. The emailaddress should match the mail address where the confirmation mail is sent.
* [x] Add page (view/template/js/css-whatever) to cancel the appointment (see design below) + confirmation page (Afspraak succesvol geannuleerd).
* [ ] To change an appointment, we can use the link that was used to create the appointment in the first place. So, point the user to the "appointment form", keep track that it was an existing appointment, create the appointment afterwards and cancel the old one.
* [ ] Improvement over the previous bullet: Allow the pre-selection of a product using a query param: `/form/my-form?product=XYZ` and use this in the change appointment link (see below).
* [x] Add links to these pages in the confirmation email: 
  **Afspraak wijzigen of annuleren**
  o [Ik wil mijn afspraak annuleren]()
  o [Ik wil mijn afspraak wijzigen]()

**Design / mock**

image

shea-maykinmedia:
Fixes #263 

~~Relies on https://github.com/open-formulieren/open-forms/pull/573 being merged first~~

**Screenshot**

I'm unsure how accurate this styling is to how the real email will look.  I can do a modification on the styling if this does not look correct.

Screenshot 2021-09-06 at 09 14 23


src/openforms/appointments/base.py
```diff
@@ -220,3 +224,16 @@ def get_appointment_details_html(self, identifier: str) -> str:
...
             "appointments/appointment_details.html", {"appointment": details}
         )
+
+    def get_appointment_links_html(self, submission: Submission) -> str:
+
+        token = submission_appointment_token_generator.make_token(submission)
+        cancel_url = reverse(
+            "api:appointments-verify-cancel-appointment-link",
+            kwargs={"token": token, "submission_id": submission.id},
+        )
```
sergei-maertens, line 234: _please_ base64 encode the submission ID to not plainly expose database IDs. I'd even say to just base64 the submission UUID instead of using the database PK. See also the django password reset URLs: https://github.com/django/django/blob/3921b1c6d24c9d5a60e5f5f83c9a394104089c21/django/contrib/auth/views.py#L270

src/openforms/appointments/base.py
```diff
@@ -220,3 +224,16 @@ def get_appointment_details_html(self, identifier: str) -> str:
...
+
+        return render_to_string(
+            "appointments/appointment_links.html",
+            {"cancel_url": cancel_url},
+        )
```
sergei-maertens, line 239: Let's decouple getting the links from the rendering in the template.

This method should just return a dict with the links & any templates should call this method instead and use the links.

I want the option to easily include this information in other places, like and API endpoint used by the SDK without being constrained by the template engine.

src/openforms/appointments/exceptions.py
```diff
@@ -8,3 +14,9 @@ class AppointmentDeleteFailed(AppointmentException):
 
 class AppointmentCreateFailed(AppointmentException):
     pass
+
+
+class CancelAppointmentFailed(APIException):
+    status_code = status.HTTP_400_BAD_REQUEST
```
sergei-maertens, line 20: It doesn't seem like this is a _client_ error, which 4xx indicates. The client did nothing wrong.

I would rather expect HTTP 503 here or perhaps HTTP 502

@joeribekker any preference/input?

src/openforms/submissions/utils.py
```diff
@@ -85,13 +85,12 @@ def send_confirmation_email(submission: Submission):
...
 
-    try:
+    if hasattr(submission, "appointment_info"):
         client = get_client()
         content += client.get_appointment_details_html(
             submission.appointment_info.appointment_id
         )
-    except (AppointmentInfo.DoesNotExist, ValueError):
-        pass
+        content += client.get_appointment_links_html(submission)
```
sergei-maertens, line 93: Let's stop hardcoding these things and just piling on top of content, and instead implement a custom template tag such as `{% appointment_information %}` and `{% get_appointment_links as links %}` so that these can be templated out by the end user.

The end result should be that end end-user can do something like this:

```django
{# other content such as summary #}

{% appointment_information %}

{% get_appointment_links as links %}

U kan deze afspraak <a href="{{ links.cancel_url }}">annuleren</a>
```

& then of course document these options with the rest of the templates documentation.

src/openforms/config/models.py
```diff
@@ -134,6 +134,11 @@ class GlobalConfiguration(SingletonModel):
             "URL to the main website. Used for the 'back to municipality website' link."
         ),
     )
+    sdk_website = models.URLField(
+        _("sdk website link"),
+        blank=True,
+        help_text=_("URL to the SDK."),
+    )
```
sergei-maertens, line 141: Please rename this specifically to "cancel_appointment_page". I'm also expecting that a global option will not be sufficient and there will be pages specific per form, so this may get a form-specific override in the future. Let's hold off on that for now though. Either way, forms will be embedded in many different CMS pages and there is no "single sdk website".

Help text should also be something like "URL to the page where the appointment cancellation should  happen."

The URL to the SDK is confusing since there's an SDK URL setting which points to which javascript bundle to load.

src/openforms/submissions/tokens.py
```diff
@@ -104,4 +104,95 @@ def _num_days(self, dt) -> int:
         return (dt - date(2001, 1, 1)).days
 
 
+class SubmissionTokenGenerator:
```
sergei-maertens, line 107: Let's put this one in the `appointments` app - since we require tokens that are valid until the appointment date.

So also the `get_expiry_days` based on a setting is wrong - we can have an exact validity of a token based on the appointment date, let's use that information instead of having a token that's valid even if the appointment has passed or not valid long enough.

src/openforms/submissions/tokens.py
```diff
@@ -104,4 +104,95 @@ def _num_days(self, dt) -> int:
...
+        will no longer validate.
+        Failing that, eventually settings.SUBMISSION_REPORT_URL_TOKEN_TIMEOUT_DAYS will
+        invalidate the token.
+        """
+        submission_report_attributes = ("id", "uuid")
+        submission_report_bits = [
+            str(getattr(submission, attribute) or "")
+            for attribute in submission_report_attributes
+        ]
+        return "".join(submission_report_bits) + str(timestamp)
```
sergei-maertens, line 188: * submission report is not relevant for appointments, at all.
* the attributes also do not _change_ after the appointment has changed, and that's the whole premise of a token system like this.

After cancellation, the appointmentinfo is either deleted or updated -> _that_ should be included in the token attributes & other items that may change (appointment date/time comes to mind immediately).

This means that once an appointment is cancelled or modified, the links are now invalidated. See again the Django built in password reset token logic - tokens are invalidated after the password hash has changed there, which makes the link single-use.

src/openforms/submissions/tokens.py
```diff
@@ -104,4 +104,95 @@ def _num_days(self, dt) -> int:
...
+
+    def _num_days(self, dt) -> int:
+        """
+        Return the number of days between 01-01-2001 and today
+        """
+        return (dt - date(2001, 1, 1)).days
+
+
 token_generator = SubmissionReportTokenGenerator()
+submission_token_generator = SubmissionTokenGenerator()
```
sergei-maertens, line 198: bad variable naming

src/openforms/conf/base.py
```diff
@@ -456,6 +456,7 @@
 # Base URL of where the SDK is hosted.
 SDK_BASE_URL = config("SDK_BASE_URL", "https://open-forms.test.maykin.opengem.nl/sdk")
 
+SUBMISSION_TOKEN_TIMEOUT_DAYS = config("SUBMISSION_TOKEN_TIMEOUT_DAYS", default=1)
```
sergei-maertens, line 459: this setting should not exist in the context of appointments

src/openforms/conf/base.py
```diff
@@ -191,7 +191,7 @@
     # 'django.middleware.locale.LocaleMiddleware',
     "corsheaders.middleware.CorsMiddleware",
     "django.middleware.common.CommonMiddleware",
-    "django.middleware.csrf.CsrfViewMiddleware",
+    # "django.middleware.csrf.CsrfViewMiddleware",
```
sergei-maertens, line 194: this is a **massive** security risk, I'm hoping it's an accident and not intentional

src/openforms/appointments/tokens.py
```diff
@@ -0,0 +1,19 @@
+from django.utils import timezone
+
+from openforms.appointments.models import AppointmentInfo
+from openforms.submissions.tokens import SubmissionTokenGenerator
+
+
+class SubmissionAppointmentTokenGenerator(SubmissionTokenGenerator):
```
sergei-maertens, line 7: Let's implement this entire generator here rather than subclassing - there's too much stuff in motion at the moment that will cause confusion.

src/openforms/appointments/tokens.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Used to generate and check tokens for submission appointments.
+    """
+
+    def get_expiry_days(self, submission):
+        try:
+            return (submission.appointment_info.start_time - timezone.now()).days
+        except (AppointmentInfo.DoesNotExist, TypeError):
+            return super().get_expiry_days(submission)
```
sergei-maertens, line 16: I do not understand the logic here, why should these exceptions not cause problems? Why should it fall back to some weird behaviour? This goes against the Zen of Python "Refuse the temptation to guess"

src/openforms/appointments/tests/test_models.py
```diff
@@ -1,9 +1,10 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.test import TestCase
 
-from openforms.appointments.tests.factories import AppointmentInfoFactory
 from openforms.submissions.tests.factories import SubmissionFactory
 
+from ..tests.factories import AppointmentInfoFactory
```
sergei-maertens, line 6:
```suggestion
from .factories import AppointmentInfoFactory
```

sergei-maertens (changes requested):
`src/openforms/utils/helpers.py` is an empty file, please remove it

src/openforms/conf/base.py
shea-maykinmedia: Yes, this was me doing a bit of debugging and accidentally not reverted.

src/openforms/appointments/exceptions.py
shea-maykinmedia: For the moment I've made this a 502.

docs/manual/templates.rst
```diff
@@ -119,11 +119,13 @@ in the template. Typically, these vary per form!
...
-Element              Description
-===================  ===========================================================================
-``{% summary %}``    A full summary of all elements marked to show in the email.
-===================  ===========================================================================
+===================================  ===========================================================================
+Element                              Description
+===================================  ===========================================================================
+``{% summary %}``                    A full summary of all elements marked to show in the email.
+``{% appointment_information %}``    The information about the appointment to show in the email.
+``{% get_appointment_links %}``      Retrieves relevant links about the appointment.
```
shea-maykinmedia, line 127: Not sure if we want more documentation on this since this essentially adds information that _can_ be used but doesn't display any information itself which makes it different than `appointment_information` or `summary`.

src/openforms/appointments/base.py
```diff
@@ -220,3 +226,20 @@ def get_appointment_details_html(self, identifier: str) -> str:
...
+        cancel_uri = reverse(
+            "appointments:appointments-verify-cancel-appointment-link",
+            kwargs={
+                "token": token,
+                "base64_submission_uuid": urlsafe_base64_encode(
+                    str(submission.uuid).encode()
+                ),
+            },
+        )
+        cancel_url = f"{settings.BASE_URL}{cancel_uri}"
```
shea-maykinmedia, line 243: Is this the best way to build the full url?  Since we don't have access to the request here I was hoping we can just use this settings variable.

src/openforms/appointments/base.py
```diff
@@ -220,3 +225,18 @@ def get_appointment_details_html(self, identifier: str) -> str:
...
+        token = submission_appointment_token_generator.make_token(submission)
+
+        cancel_path = reverse(
+            "appointments:appointments-verify-cancel-appointment-link",
+            kwargs={
+                "token": token,
+                "submission_uuid": submission.uuid,
+            },
+        )
+        cancel_url = f"{settings.BASE_URL}{cancel_path}"
```
shea-maykinmedia, line 240: Is this the best way to build the full url?  We don't have access to the request so we can't do `request.build_absolute_url()`(or whatever the exact function is) and even if we did I'm not sure that would be the best way to do it?

src/openforms/appointments/tests/test_endpoints.py
```diff
@@ -0,0 +1,88 @@
...
+                "submission_uuid": submission.uuid,
+            },
+        )
+
+        response = self.client.get(endpoint)
+
+        # Assert submission is stored in session
+        self.assertEqual(
+            str(submission.uuid), self.client.session[SUBMISSIONS_SESSION_KEY][0]
+        )
```
shea-maykinmedia, line 48: Is this the correct way to assert this?  At least for this unit test I don't think there should ever be a second submission in the session.

src/openforms/appointments/api/tests/test_endpoints_with_jcc.py
```diff
@@ -228,3 +239,123 @@ def test_get_times_returns_403_when_no_active_sessions(self):
...
+        self.submission.form = form
+        self.submission.save()
+
+        SubmissionStepFactory.create(
+            submission=self.submission,
+            data={
+                "email": "maykin@media.nl",
+            },
+            form_step=form_step,
+        )
```
sergei-maertens, line 339: See earlier remark about test setup

src/openforms/appointments/api/tests/test_endpoints_with_qmatic.py
```diff
@@ -199,3 +213,115 @@ def test_get_times_returns_403_when_no_active_sessions(self):
...
+        self.submission.form = form
+        self.submission.save()
+
+        SubmissionStepFactory.create(
+            submission=self.submission,
+            data={
+                "email": "maykin@media.nl",
+            },
+            form_step=form_step,
+        )
```
sergei-maertens, line 265: See earlier remark about the test setup

src/openforms/appointments/api/urls.py
```diff
@@ -1,10 +1,21 @@
...
+    TimesListView,
+)
 
 urlpatterns = [
     path("products", ProductsListView.as_view(), name="appointments-products-list"),
     path("locations", LocationsListView.as_view(), name="appointments-locations-list"),
     path("dates", DatesListView.as_view(), name="appointments-dates-list"),
     path("times", TimesListView.as_view(), name="appointments-times-list"),
+    path(
+        "<str:submission_uuid>/cancel",
```
sergei-maertens, line 17:
```suggestion
        "<uuid:submission_uuid>/cancel",
```
https://docs.djangoproject.com/en/2.2/topics/http/urls/#path-converters there is a `uuid` converter

src/openforms/appointments/api/views.py
```diff
@@ -167,3 +181,46 @@ def get_objects(self):
...
+    def post(self, request, *args, **kwargs):
+        submission = self.get_object()
+
+        serializer = CancelAppointmentInputSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        emails = submission.get_email_confirmation_recipients(submission.data)
+
+        if serializer.validated_data["email"] not in emails:
+            raise PermissionDenied
```
sergei-maertens, line 216: A comment describing the business logic/requirement from the ticket here would provide context that is now completely absent.

src/openforms/appointments/base.py
sergei-maertens: Yeah I think it is, for now at least. It does mean that we use the `BASE_URL` setting for more than DigiD, so it should be documented as such.

An alternative is using `django.contrib.sites` but I really dislike that and brings its own problems, so let's improve on documentation here.

However, don't join URLs together like that, as the `BASE_URL` setting may have a trailing slash. To be safe, you should use `urllib.parse.urljoin`.

src/openforms/appointments/tests/test_endpoints.py
sergei-maertens: Just use `self.assertIn(str(submission.uuid), self.client.session[SUBMISSIONS_SESSION_KEY])`

src/openforms/appointments/tests/test_endpoints.py
```diff
@@ -0,0 +1,88 @@
...
+
+        expected_redirect_url = (
+            f"http://maykinmedia.nl/afspraak-annuleren?"
+            f"time=2021-07-21T12%3A00%3A00+00%3A00&submission_uuid={str(submission.uuid)}"
+        )
```
sergei-maertens, line 53: Please use the `furl` dependency (https://pypi.org/project/furl/) that's already in this project to properly mutate/build/parse URLs.

src/openforms/appointments/tokens.py
```diff
@@ -0,0 +1,92 @@
...
+
+from django.conf import settings
+from django.utils import timezone
+from django.utils.crypto import constant_time_compare, salted_hmac
+from django.utils.http import base36_to_int, int_to_base36
+
+from openforms.submissions.models import Submission
+
+
+class SubmissionAppointmentTokenGenerator:
```
sergei-maertens, line 11: #655 is still pending, but it has a refactor for token generators in general. Depending on which PR gets merged first, the other will have to adapt to make use of this.

src/openforms/appointments/views.py
```diff
@@ -0,0 +1,58 @@
...
+        403: OpenApiResponse(
+            response=ExceptionSerializer,
+            description=_("Unable to verify token."),
+        ),
+    },
+)
+class VerifyCancelAppointmentLinkView(RedirectView):
+
+    permission_classes = ()
+    authentication_classes = ()
```
sergei-maertens, line 33: these are not relevant for `RedirectView`

src/openforms/appointments/views.py
```diff
@@ -0,0 +1,58 @@
...
+@extend_schema(
+    summary=_("Verify the appointment cancel link."),
+    responses={
+        302: None,
+        403: OpenApiResponse(
+            response=ExceptionSerializer,
+            description=_("Unable to verify token."),
+        ),
+    },
+)
```
sergei-maertens, line 29: it's not an API endpoint, so we don't document it as such

src/openforms/appointments/views.py
```diff
@@ -0,0 +1,58 @@
...
+class VerifyCancelAppointmentLinkView(RedirectView):
+
+    permission_classes = ()
+    authentication_classes = ()
+
+    def get_redirect_url(self, submission_uuid: int, token: str, *args, **kwargs):
+        try:
+            submission = Submission.objects.get(uuid=submission_uuid)
+        except ObjectDoesNotExist:
```
sergei-maertens, line 38:
```suggestion
        except Submission.DoesNotExist:
```

src/openforms/appointments/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+        add_submmission_to_session(submission, self.request.session)
+
+        config = GlobalConfiguration.get_solo()
+
+        return (
+            f"{config.cancel_appointment_page}"
+            f"?time={submission.appointment_info.start_time.isoformat()}"
+            f"&submission_uuid={str(submission_uuid)}"
+        )
```
sergei-maertens, line 58: This is not how you build URLs safely and there is plenty of tooling at your hands to do it the right way:

1. `furl`, mentioned earlier
2. python's stdlib `urlib` package
3. Django's own utils: https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.http.urlencode

Also - don't go crazy with f-strings, the fact that you can evaluate about anything between the curly braces doesn't mean you should. It should only really be used for drop-in variables, so bring those variables/method calls outside of the f-string.

besides that, doing `f"{str(uuid)}"` is literally the same effect as doing the more readable `f"{uuid}"`, as the string representation will be used by the f-string feature.

src/openforms/appointments/views.py
```diff
@@ -0,0 +1,58 @@
...
+        valid = submission_appointment_token_generator.check_token(submission, token)
+        if not valid:
+            logger.debug("Called endpoint with an invalid token: %s", token)
+            raise PermissionDenied("Cancel url is not valid")
+
+        add_submmission_to_session(submission, self.request.session)
+
+        config = GlobalConfiguration.get_solo()
+
+        return (
```
sergei-maertens, line 54: This should `raise RunTimeError("No appointment page configured")` if the appointment cancel page URL is empty (the model field is `blank=True` and has no default).

src/openforms/emails/templatetags/appointments.py
```diff
@@ -0,0 +1,17 @@
...
+
+from openforms.appointments.utils import get_client
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def appointment_information(context):
+    client = get_client()
+    return client.get_appointment_details_html(context["_appointment_id"])
```
sergei-maertens, line 11: I suspect this will crash if there's no appointment ID in the template context.

sergei-maertens (changes requested):

docs/manual/templates.rst
joeribekker: can we also keep it a little consistant, and rename `get_appointment_links` to `appointment_links` in the template?

joeribekker (changes requested):

docs/manual/templates.rst
sergei-maertens: it's meant to be used as `{% get_appointment_links as links %} {{ links.cancel_url|urlize }}` for example, so I think the `get_` is more semantic. In the future, this will also include the `change_url`.

src/openforms/appointments/api/tests/test_endpoints_with_jcc.py
```diff
@@ -228,3 +239,123 @@ def test_get_times_returns_403_when_no_active_sessions(self):
...
+                "components": [
+                    {"key": "email", "label": "Email", "confirmationRecipient": True},
+                ],
+            }
+        )
+
+        form_step = FormStepFactory.create(form=form, form_definition=form_def)
+
+        self.submission.form = form
+        self.submission.save()
```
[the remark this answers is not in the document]
shea-maykinmedia: The reason was that all the tests require a submission but only some of the tests require the form to be set up in such a way.

That said I can update the tests to use `from_components` now that we have this.

src/openforms/appointments/tokens.py
shea-maykinmedia: Okay good to know.  I'm guessing you mean once #655 is merged that this will need to inherit from `BaseTokenGenerator`?

src/openforms/appointments/api/urls.py
shea-maykinmedia: The reason for using `str` is that the UUIDs of the submissions are stored in the session as strings, so `ActiveSubmissionPermission.has_object_permission` does `submission_id in active_submissions` but if `submission_id` at this point is a UUID then this check will fail and the user would get a 403.

I could update https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/api/permissions.py#L36 to be `return str(submission_id) in active_submissions` if you think that's a better solution?  Or maybe there is a better way solution?

src/openforms/emails/templatetags/appointments.py
shea-maykinmedia: It will.  This function requires an appointment id and if there is no appointment id (either because one was not created or because it just wanted added to the context) then this shouldn't be called so I though it was better to crash than fail silently.

docs/manual/templates.rst
```diff
@@ -119,11 +119,32 @@ in the template. Typically, these vary per form!
...
+         {% get_appointment_links as links %}
+         Cancel Appointment: {{ links.cancel_url|urlize }}
+
+
+   .. tab:: Rendered
+
+      .. code:: text
+
+         Cancel Appointment: http://fake.nl/api/v1/submission-uuid/token/verify/
+
```
shea-maykinmedia, line 147: Is this okay for an example?

Screenshot 2021-09-14 at 15 51 19

Screenshot 2021-09-14 at 15 51 25

joeribekker: Yes, I like

src/openforms/emails/templatetags/appointments.py
sergei-maertens: Considering that the content editors/form designers are not experts on the DTL and the forgiveness typically found in the DTL (empty variable -> just output empty string), I think we should give some leeway here. If the context var is not present, just return an empty string.

src/openforms/appointments/api/tests/test_endpoints_with_jcc.py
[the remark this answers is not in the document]
sergei-maertens: In any case you want test-data to look realistic, Bart has already been troubled with updating/extending tests that have a very weird, semantically incorrect test setup that then caused problems.

src/openforms/appointments/api/urls.py
sergei-maertens: Yes, go ahead with updating the permission class.

And if we were to keep the `str` pattern, _add a comment explaining the non-obvious choice_. We really can't guess from the code itself why the code was written that way, so document design decisions.

src/openforms/appointments/tokens.py
sergei-maertens: correct, and it's merged now :)

src/openforms/appointments/tests/test_base.py
```diff
@@ -45,3 +51,23 @@ def test_get_appointment_details_html(self):
         self.assertIn("Remarks", result)
         self.assertIn("Some", result)
         self.assertIn("<h1>Data</h1>", result)
+
+    @patch("openforms.appointments.base.submission_appointment_token_generator")
```
sergei-maertens, line 55: Remove mocking here and use real token generator

src/openforms/appointments/tests/test_base.py
```diff
@@ -45,3 +51,23 @@ def test_get_appointment_details_html(self):
...
+    def test_get_appointment_links(self, token_generator_mock):
+        submission = SubmissionFactory.create()
```
sergei-maertens, line 57: TODO: set up AppointmentInfo -> is needed for token generation

src/openforms/appointments/tests/test_base.py
```diff
@@ -45,3 +51,23 @@ def test_get_appointment_details_html(self):
...
+        result = self.plugin.get_appointment_links(submission)
+
+        cancel_path = reverse(
+            "appointments:appointments-verify-cancel-appointment-link",
+            kwargs={
+                "token": fake_token,
+                "submission_uuid": submission.uuid,
+            },
+        )
+        cancel_url = urljoin(settings.BASE_URL, cancel_path)
```
sergei-maertens, line 71: use `@override_settings(BASE_URL=...)` instead and verify that the link is indeed the right fully qualified URL.

src/openforms/appointments/tests/test_endpoints.py
```diff
@@ -0,0 +1,123 @@
...
+
+from openforms.config.models import GlobalConfiguration
+from openforms.submissions.constants import SUBMISSIONS_SESSION_KEY
+from openforms.submissions.tests.factories import SubmissionFactory
+
+from .factories import AppointmentInfoFactory
+
+
+class VerifyCancelAppointmentLinkViewTests(TestCase):
+    @patch(
```
sergei-maertens, line 18: use the real token generator instead of mocking

src/openforms/appointments/tests/test_endpoints.py
```diff
@@ -0,0 +1,123 @@
...
+
+        response = self.client.get(endpoint)
+
+        # Assert submission is stored in session
+        self.assertIn(
+            str(submission.uuid), self.client.session[SUBMISSIONS_SESSION_KEY]
+        )
+
+        expected_redirect_url = (
+            furl(config.cancel_appointment_page)
```
sergei-maertens, line 51: use string value from earlier set up to make it more explicit - in general you want to avoid "best practices" in tests and keep them as dumb and straight forward as possible.

src/openforms/appointments/tests/test_endpoints.py
```diff
@@ -0,0 +1,123 @@
...
+                }
+            )
+            .url
+        )
+
+        self.assertRedirects(
+            response, expected_redirect_url, fetch_redirect_response=False
+        )
+
+    @patch(
```
sergei-maertens, line 65: Remove mock

src/openforms/appointments/tests/test_endpoints.py
```diff
@@ -0,0 +1,123 @@
...
+                "token": "mocked",
+                "submission_uuid": uuid.uuid4(),
+            },
+        )
+
+        response = self.client.get(endpoint)
+
+        self.assertEqual(response.status_code, 403)
+
+    @patch(
```
sergei-maertens, line 106: Avoid mocking

src/openforms/appointments/tokens.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    key_salt = "openforms.appointments.tokens.SubmissionAppointmentTokenGenerator"
+
+    def get_token_timeout_days(self, submission: Submission) -> int:
+        """
+        Determine how many days the token is valid for.
+
+        Defaults to :attr:`token_timeout_days`, optionally you can override this method and
+        fetch information from the object being checked.
```
sergei-maertens, line 21: Update docstring to reflect what is actually happening.

src/openforms/emails/templatetags/appointments.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@register.simple_tag(takes_context=True)
+def appointment_information(context):
+    if not context.get("_appointment_id"):
+        # Use get since _appointment_id could be an empty string
+        return ""
+
+    client = get_client()
+    return client.get_appointment_details_html(context["_appointment_id"])
```
sergei-maertens, line 15: `:=` candidate to clean up code a little bit

sergei-maertens (changes requested):
I'll pick up the changes to wrap up this PR, as discussed in the office

sergei-maertens (approved):

Issue #227, opened 2021-04-26
Een gebruiker kan een formulier tussentijds kunnen opslaan en op een later moment afronden (deze eis valt ...


joeribekker:
...buiten het MVP).

Eis: Sturing Formulier 1.
Zie ook NvI: 74, 153

**Tasks**

* [x] When clicking the "Tussentijds opslaan" button, present the user with a popup/modal to explain this feature, provide an emailaddress to sent the "continue" link to, button to confirm & logout or cancel this popup.
* [x] The user should be logged out after clicking confirm & logout
* [x] The user should recieve an email with a magic-token link to continue the form
* [x] When the link is clicked, the form & session should be restored.
* [x] The email should contain the information until the form can be continued (submissions are deleted after X days due to cleanup policies)

sergei-maertens:
Back to TODO - depends on #709

PR #617, opened 2021-09-02
Submission process tasks orchestration

### LINKED ISSUE #614: Set up async tasks orchestration
On submission complete, various tasks happen in a specific order with possible failure states.

The celery task orchestration needs to be defined and work with failure/timeouts to proceed/not-proceed. The different modules that Open Forms has will tap into this.

**Tasks**

- [x] Set up on_submission_completion handler which kicks of a celery chain
- [x] Store root task ID with submission for later checks
- [x] Organize celery tasks with possible timeout/retry behaviour
    - [x] Set up suitable decorator
    - [x] Check all tasks which *may* communicate with external systems
    - [x] Implement unique reference extraction
- [x] #651
- [x] Set up django-timeline-logger to log all events/actions (#247)
- [x] Set up status progression
- [x] Check if we can use a HMAC status check URL & freeze the download report token based on the status
- [ ] Schedule resizing of attachments BEFORE registration backend (parallel group) #507 
- [ ] #701

sergei-maertens:
Partially fixes #614

Currently this sets up the main chain of (soft-) dependent tasks. This is only the base - see the original issue for the status of all subtasks.

**Changes**

- All processing that may be doing (heavy) IO is moved to Celery tasks
- The Celery tasks are ordered in an execution chain
- The task ID of the chain is saved on the submission so that we can check the execution status asynchronously
- The beginnings of a new submission status endpoint are in place, which performs these status checks and serializes the status back to the client.
- There is now a single, unique public reference for a submission. If possible, it is sourced from the registration backend/plugin, otherwise it is generated uniquely.
- Tasks interacting with other services beyond our control can be marked as retry-in-workflow-until-timeout-is-reached. If there are still retries available, they will be executed outside of the current workflow and not block the confirmation page for the end-user.
- Cleaned up some unused parts
- Simplified a number of tests, possible because the tasks are smaller units that can be tested outside of the request-response cycle now.


docs/sub_completion.md
```diff
@@ -0,0 +1,30 @@
+# Submission completion flow
```
joeribekker, line 1: FYI I already added some flow documentation here: https://github.com/open-formulieren/open-forms/blob/master/docs/developers/submissions.rst

sergei-maertens: this will be deleted later, it was just a reminder/sketching for myself ^^

src/openforms/submissions/api/viewsets.py
```diff
@@ -108,60 +102,64 @@ def _complete(self, request, *args, **kwargs):
...
+        to the session so the user can return to the first step(s) in the UI.
         """
         submission = self.get_object()
         validate_submission_completion(submission, request=request)
-        submission.completed_on = timezone.now()
 
+        submission.completed_on = timezone.now()
         submission.save()
+
+        # TODO: implement in celery tasks in cleanup (see ./tasks/__init__.py)
```
sergei-maertens, line 130: Current train of thought is to _not_ do this in celery (it requires us to pass around the session object which is... weird), but instead keep this and re-add the submission ID to the session if the user needs to correct mistakes.

src/openforms/submissions/api/viewsets.py
```diff
@@ -108,60 +102,64 @@ def _complete(self, request, *args, **kwargs):
...
-        transaction.on_commit(on_submission_commit)
-
-        book_appointment_for_submission(submission)
+        # after committing the database transaction where the submissions completion is
+        # stored, start processing the completion.
+        transaction.on_commit(lambda: on_completion(submission.id))
 
-        if hasattr(submission.form, "confirmation_email_template"):
-            transaction.on_commit(lambda: send_confirmation_email(submission))
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
sergei-maertens, line 138: In a separate PR, I'll look at the status endpoint and ensure that the URL is included here - this will probably be token-based to not need the submission ID in the session, yet return a URL that will expire after a set time and/or internal state change.

src/openforms/submissions/api/viewsets.py
```diff
@@ -108,60 +102,64 @@ def _complete(self, request, *args, **kwargs):
...
+            200: SubmissionProcessingStatusSerializer,
+            # TODO: 403, 401, 429
+        },
+    )
+    @action(
+        detail=True,
+        methods=["get"],
+        permission_classes=(ActiveSubmissionPermission,),
+    )
+    def status(self, request, *args, **kwargs):
```
sergei-maertens, line 153: This is broken in its current state, but will be tackled in a follow up PR. The complete endpoint removes the submission ID from the session, so this permission class will always cause a 403.

Tests will be added for this endpoint in aforementioned follow up PR.

src/openforms/submissions/models.py
```diff
@@ -117,7 +117,6 @@ class Submission(models.Model):
         blank=True,
         validators=(validate_kvk,),
     )
-    current_step = models.PositiveIntegerField(_("current step"), default=0)
```
sergei-maertens, line 120: Just some clean-up - I noticed this wasn't used anywhere.

src/openforms/submissions/models.py
```diff
@@ -146,11 +158,29 @@ class Submission(models.Model):
...
 
     class Meta:
         verbose_name = _("Submission")
         verbose_name_plural = _("Submissions")
+        constraints = [
+            models.UniqueConstraint(
+                fields=("public_registration_reference",),
+                name="unique_public_registration_reference",
+                condition=~models.Q(public_registration_reference=""),
+            )
```
sergei-maertens, line 182: This conditional unique constraint allows us to enforce uniqueness without needing nullable columns, as empty reference values are excluded from the constraint. Also works well with existing data, since it's a new model field.

src/openforms/submissions/status.py
```diff
@@ -0,0 +1,65 @@
+"""
+Utility to interact with the celery task status.
+"""
```
sergei-maertens, line 3: Very much WIP still - it works, sort of, but the `payment_url` for example is not relevant and probably while implementing it in the SDK, we will discover new nuances.

src/openforms/submissions/tasks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Finalization happens _after_ the confirmation screen is shown to the end-user.
+    Showing this screen depends on all the previous registration tasks being completed,
+    so the :func:`on_completion` handler must kick this off AND have its own task ID
+    that finishes, which is checked in the submission status endpoint.
+    """
+    send_confirmation_email_task = maybe_send_confirmation_email.si(submission_id)
+    send_confirmation_email_task.delay()
+
+    # TODO: clean up session data
```
sergei-maertens, line 82: likely not here and rather keep it in the request-response cycle, but leaving the TODO here for now as a reminder.

src/openforms/submissions/tasks/registration.py
```diff
@@ -0,0 +1,119 @@
...
+        reference = generate_unique_submission_reference()
+    return reference
+
+
+def get_random_reference() -> str:
+    # 36 characters with length 6 -> 36^6 possible combinations.
+    # that's roughly 2.1 billion combinations before we run out of options.
+    # Also note that submissions are pruned after a (configurable) number of days, so
+    # used references do become available again after that time.
+    # TODO: maybe include date param?
```
sergei-maertens, line 94: as discussed this morning - holding off on that, but we keep it in the back of our mind this may be needed in the future in the event that a reference gets recycled.

src/openforms/utils/celery.py
```diff
@@ -0,0 +1,122 @@
...
+logger = logging.getLogger(__name__)
+
+ExceptionCls = type(Exception)
+
+
+def default_should_retry(exception: Exception, task: Task) -> bool:
+    return True
+
+
+def maybe_retry_in_workflow(
```
sergei-maertens, line 18: I will add some unit tests for this!

src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -695,3 +694,15 @@ def test_plugin(self, m, mock_task):
                 "//zkn:object/zkn:formaat": "application/msword",
             },
         )
+
+    def test_no_reference_can_be_extracted(self, m):
```
Bartvaderkin, line 698: Name of the test is wrong, as a reference *can* be extracted here.

sergei-maertens: nice catch, fixed!

src/openforms/utils/celery.py
sergei-maertens: done

src/openforms/submissions/api/viewsets.py
```diff
@@ -110,60 +104,64 @@ def _complete(self, request, *args, **kwargs):
...
+        * a report PDF is generated
+        * the submission is persisted to the configured backend
+        * payment is initiated if relevant
+
+        Once a submission is done processing, it's scheduled to be removed from the
+        session. This means it's no longer possible to change or read the submission
+        data (including individual steps).
 
-        The submission is persisted to the configured backend.
+        TODO: give some leeway? add endpoint to trigger cleanup immediately?
```
joeribekker, line 120: Sounds like a TODO for when we implement "Tijdelijk opslaan", as described here: https://github.com/open-formulieren/open-forms/issues/227

src/openforms/submissions/status.py
joeribekker: See comment about status/result

src/openforms/submissions/tasks/__init__.py
joeribekker: Why don't we just cleanup the session when the confirmation page is shown (aka status=done, result=success)?

This also solves the issue for adding the submission to the session again, mentioned in a TODO earlier.

If we need to block double submissions, i rather fix it with a check on the submission status.

src/openforms/submissions/tasks/appointments.py
```diff
@@ -0,0 +1,73 @@
...
+    Register an appointment for the submission IF relevant.
+
+    If the submission is for a form which is configured to create appointments,
+    ensure that the appointment is registered in the configured backend.
+
+    This can either not be needed, be succesful or fail. Either way, the result should
+    be stored in the database. If appointment registration fails, this feedback
+    should find its way back to the end-user.
+    """
+    logger.info("Registering appointment for submission %d", submission_id)
```
joeribekker, line 50: I think it's confusing to log it like this, since it implies that an appoinment registration is needed. Either we remove this log statement or amend it to indicate "if needed" / "maybe".

src/openforms/submissions/tasks/appointments.py
```diff
@@ -0,0 +1,73 @@
...
+
+    The reference may be sourced from the registration backend, or it may be sourced
+    internally because of problems with the registration backend. Either way,
+    the appointment and submission must agree on the internal reference.
+
+    This task should be scheduled after the submission backend registration was
+    attempted and after the "final" reference was obtained.
+    """
+    submission = Submission.objects.get(id=submission_id)
+    logger.info("Updating appointment for submission %d", submission_id)
```
joeribekker, line 72: Same comment as above, amend it to indicate "if needed" / "maybe".

src/openforms/submissions/tasks/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+        # any backend registration happens, as on-failure, the user should get feedback
+        # about the failure.
+        register_appointment_task,
+        # The submission report needs to already have been generated before it can be
+        # attached in the registration backend.
+        # TODO: can be in parallel with register_appointment_task and if that fails -> delete the report again
+        generate_report_task,
+        # TODO: ensure that any images that need resizing are done so before this is attempted
+        register_submission_task,
+        obtain_submission_reference_task,
```
joeribekker, line 48: I think `register_submission` (or rather the plugin being called) handles everything that has to do with sending stuff to the registration backend.

If this includes documents (attachments), it can take a while, even though the Zaak is probably already registered and the zaaknumber is technically known. Because uploading documents takes longer, an internal reference is generated which is not really needed. Maybe we should split up the "registering the submission" and "uploading the documents".

But, this sounds like another ticket...

joeribekker (changes requested):

src/openforms/appointments/service.py
```diff
@@ -0,0 +1,50 @@
...
+from openforms.submissions.models import Submission
+
+__all__ = [
+    "AppointmentRegistrationFailed",
+    "AppointmentUpdateFailed",
+    "register_appointment",
+]
+
+
+class AppointmentInteractionFailed(Exception):
```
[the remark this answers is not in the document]
sergei-maertens: missed that, will move it!

src/openforms/registrations/service.py
```diff
@@ -0,0 +1,43 @@
...
+
+    if not (result := submission.registration_result):
+        raise NoSubmissionReference("No result data saved, nothing to extract.")
+
+    # figure out which plugin to call for extraction
+    plugin = registry[backend]
+
+    try:
+        return plugin.get_reference_from_result(result)
+    except Exception as exc:
```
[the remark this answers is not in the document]
sergei-maertens: correct

src/openforms/submissions/api/serializers.py
```diff
@@ -331,3 +309,50 @@ class SubmissionStateLogicSerializer(serializers.Serializer):
 class SubmissionStateLogic:
     submission: Submission
     step: SubmissionStep
+
+
+class SubmissionProcessingStatusSerializer(serializers.Serializer):
+    state = serializers.ChoiceField(
+        label=_("background task state"),
+        choices=celery.states.ALL_STATES,
```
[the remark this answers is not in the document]
sergei-maertens: yeah, I'll incorporate that in the follow up PR dealing with the `status` endpoint. I was leaning towards that direction indeed.

src/openforms/submissions/api/views.py
```diff
@@ -17,43 +17,14 @@
...
 
 class RetrieveReportBaseView(GenericAPIView):
     queryset = SubmissionReport.objects.all()
     lookup_url_kwarg = "report_id"
 
 
-@extend_schema(
-    summary=_("Get PDF report generation status"),
-    description=_(
-        "On submission completion, a PDF report is generated with the submitted form "
```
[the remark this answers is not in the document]
sergei-maertens: better than hurting your feelings! :D

src/openforms/submissions/api/viewsets.py
sergei-maertens: bwoah, leaning more towards the timestamped HMAC'd token approach for that, similar to the download report URL

src/openforms/submissions/tasks/__init__.py
sergei-maertens: because that requires the API client to have called the status endpoint or call an endpoint-to-be-added, and we can't assume that all API clients will behave the way they should. If we keep this server-side, it's guaranteed to work the way it should.

Especially with the timestamped HMAC token approach we have fairly simple options to manage that.

Removing the submission from the session is important to make the submission data immutable, as otherwise it would be possible to mutate state and have the submission data be different from what is sent to the registration backend. I just want to block that completely - you should not have access to the step data endpoints at all anymore.

src/openforms/submissions/tasks/appointments.py
sergei-maertens: yeah, those were mostly used for my own internal debugging, but they could use some love indeed :)

src/openforms/submissions/tasks/__init__.py
sergei-maertens: Separate ticket indeed. Possibly we can use a separate celery task that is `.delay(...)`-ed as part of the plugin registration so that we get the minimum required "result" asap and can continue. But for the e-mail backend for example, we _do_ need the resized attachments before sending the e-mail, since they are added as attachment to the e-mail.

Issue #627, opened 2021-09-03
Add a management command & add it to CI to check for missing admin-index updates


sergei-maertens:
From Slack:

Sergei:
> reminder when adding models: make sure that they are added to the admin-index fixture. I can see this hasn't been done for the appointments models

Bart
> can we do a test for this? like the migration changes?

PR #631, opened 2021-09-03
[#627] Added command and CI action to check default admin-index fixture for unregistered 'Miscellaneous' models

Bartvaderkin:


Bartvaderkin:
CI fails because the code works :smile:

sergei-maertens (approved):

PR #634, opened 2021-09-06
Get booking appointment working within task orchestration

shea-maykinmedia:
I believe this is all that should be needed to get the current appointment booking to work with the new task orchestration.

Now the util function will raise the exception on a failure (after creating the Appointment Info instance) so the task will can then handle the retrying.  In the case of a success or missing information where we don't want to retry the util function simply returns which should mark the task as successful.

### DIFF
--- a/src/openforms/appointments/service.py
+++ b/src/openforms/appointments/service.py
@@ -33,12 +33,6 @@ def register_appointment(submission: Submission) -> None:
       database with the relevant context/information - the exception just signals the
       failure to the caller so that it can retry if needed.
     """
-    # TODO: implement, see PR #573
+    from .utils import book_appointment_for_submission
 
-    # DEBUGGING - simulate some time passing by external API calls
-    import time
-
-    time.sleep(2)
-
-    # SIMULATE failure
-    raise AppointmentRegistrationFailed("Missing information.", should_retry=True)
+    book_appointment_for_submission(submission)
--- a/src/openforms/appointments/utils.py
+++ b/src/openforms/appointments/utils.py
@@ -27,9 +27,9 @@ def get_client():
 def book_appointment_for_submission(submission: Submission) -> None:
     try:
         # Delete the previous appointment info if there is one since
-        #   it will cause and error creating the new one
-        # Since this will be called multiple times on a failure it's
-        #   possible this will exist
+        #   since a new one will be created
+        # This function will be called multiple times on a failure so
+        #   this is the case a previous appointment_info may exist
         submission.appointment_info.delete()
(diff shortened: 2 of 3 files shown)


src/openforms/appointments/utils.py
```diff
@@ -82,18 +92,20 @@ def book_appointment_for_submission(submission: Submission) -> None:
...
         )
-    except AppointmentCreateFailed as e:
+    except AppointmentCreateFailed:
         AppointmentInfo.objects.create(
             status=AppointmentDetailsStatus.failed,
             error_information="Failed to make appointment",
             submission=submission,
         )
+        raise AppointmentRegistrationFailed(
+            "Unable to create appointment", should_retry=True
```
shea-maykinmedia, line 107: Just to double check, this `should_retry` should be set to `True` because the task itself is what handles the timeout, meaning if this exception is raised after the task's timeout limit then the task won't actually retry, correct?

Looking at the code if this was not specified then it would default to `False` and the task would not retry even within the timeout limit.

sergei-maertens: yeah, this seems to be correct. There would be no retry if there's insufficient (=only partial) data submitted by the end-user, but if it's because JCC/Q-matic errored out, then we need to retry.

src/openforms/appointments/utils.py
```diff
@@ -82,18 +92,20 @@ def book_appointment_for_submission(submission: Submission) -> None:
...
+        )
```
sergei-maertens, line 108: please always `raise NewException from original_exception` to keep all the context in the stack.

src/openforms/appointments/service.py
```diff
@@ -33,12 +33,6 @@ def register_appointment(submission: Submission) -> None:
       database with the relevant context/information - the exception just signals the
       failure to the caller so that it can retry if needed.
     """
-    # TODO: implement, see PR #573
+    from .utils import book_appointment_for_submission
```
sergei-maertens, line 36: can this import be moved to the top of the module? I'm not immediately seeing where we have a circular dependency

sergei-maertens (changes requested):

src/openforms/appointments/service.py
shea-maykinmedia: `utils` imports `AppointmentRegistrationFailed` from this file so it can raise that exception, that's where the circular dependency is.  

We could add `from .service import AppointmentRegistrationFailed` in the `utils` file or move some of the code around to prevent the circular import but I'm not sure if those are better solutions.

sergei-maertens (approved):

Issue #597, opened 2021-08-30
Appointment improvements


joeribekker:
To prevent reloading and offer a better user experience, on the frontend:

* [x] Load products upon form step loading, instead of when clicking the pulldown.
* [x] Load locations in the dropdown upon selecting a product.
* [x] Load dates upon selecting a location.
* [x] Load times when a date is selected.
* [x] Whenever nothing is loaded, make the form elements disabled. Also make them disabled when the load is in progress.

Front end summary page

* [x] Should have human readable values in summary page.  Will need a way to save these human readable values after they have been selected.

In the admin:

* [x] Optimize the layout when designing a form, by making the checkboxes into a radio-select or dropdown (only one can be choosen anyway). Only show the additional fields when the appropriate option is selected.
* [x] Move appointment-specific configuration to separate tab #612

Submission:

* [x] ~~When an appointment cannot be made, the form should NOT be submitted. Instead the API should return an error message instead. The form then goes back to the summary page showing the error message to the user. Tracked in #614 - the SDK will poll a status endpoint which will relay the information.~~

sergei-maertens:
Was typing out an issue and then remembered this one.

---

While testing the appointments on https://open-forms.test.maykin.opengem.nl/registratie-open-bedrijven-dag, I observed the following quirks (on Brave browser, which is essentially Chromium):

- [ ] The initial load of products was slow (3.3s, I presume because of WSDL loading/fetching). Subsequent loads were reasonably fast (250ms)
- [ ] When clicking a dropdown, it looks like the options are refreshed and it closes/re-opens or re-renders. It's not a pleasant user experience
- [ ] On the overview page, the appointment details are rendered weirdly:
    - Date is displayed in `YYYY-MM-DD`, this should be localized as `DD-MM-YYYY` or `<day> <month>, <year>` in a more textual form
    - Time is displayed in ISO-8601 - this should probably be restrict to just the time in `HH:mm` format
    - Product is just the product code
    - Location is the ID (?) of the location rather than the label.

sergei-maertens:
Still working on the SDK parts

sergei-maertens:
SDK part merged, it's not great in the total result atm, but we're following up in #682

PR #645, opened 2021-09-09
Handle new format coming from SDK

shea-maykinmedia:
Changes needed for points
- Should have human readable values in summary page. Will need a way to save these human readable values after they have been selected.
- On the overview page, the appointment details are rendered weirdly

of #597 

Related SDK Pull Request: https://github.com/open-formulieren/open-forms-sdk/pull/58

The cause of the displaying of Products and Locations is that we were only saving the id and not the name/human readable value so we could only display the id back at a later point.  Now both will be stored with a dash between the id and name.

One additional positive of this is that in the submission data you will now see both the id and name instead of just the id.

### DIFF
--- a/src/openforms/appointments/tests/test_utils.py
+++ b/src/openforms/appointments/tests/test_utils.py
@@ -83,7 +83,7 @@ class BookAppointmentForSubmissionTest(TestCase):
         submission = SubmissionFactory.create(form=form)
         SubmissionStepFactory.create(
             submission=submission,
-            data={"product": "79", "time": "2021-08-25T17:00:00"},
+            data={"product": "79-Paspoort", "time": "2021-08-25T17:00:00"},
             form_step=form_step_1,
         )
         SubmissionStepFactory.create(
@@ -146,8 +146,8 @@ class BookAppointmentForSubmissionTest(TestCase):
         SubmissionStepFactory.create(
             submission=submission,
             data={
-                "product": "79",
-                "location": "1",
+                "product": "79-Paspoort",
+                "location": "1-Amsterdam",
                 "time": "2021-08-25T17:00:00+02:00",
             },
             form_step=form_step_1,
@@ -207,8 +207,8 @@ class BookAppointmentForSubmissionTest(TestCase):
         SubmissionStepFactory.create(
             submission=submission,
             data={
-                "product": "79",
-                "location": "1",
+                "product": "79-Paspoort",
+                "location": "1-Amsterdam",
                 "time": "2021-08-25T17:00:00+02:00",
             },
             form_step=form_step_1,
--- a/src/openforms/appointments/tests/test_utils.py
+++ b/src/openforms/appointments/tests/test_utils.py
@@ -83,7 +83,10 @@ class BookAppointmentForSubmissionTest(TestCase):
         submission = SubmissionFactory.create(form=form)
         SubmissionStepFactory.create(
             submission=submission,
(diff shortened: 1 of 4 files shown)


sergei-maertens (changes requested):
Can we not get a rich object from the SDK or just an ID from the SDK instead of having to parse strings? That looks so fragile, the moment the external appointments service have dashes in their IDs this collapses on itself.

shea-maykinmedia:
Just an ID is what was being done before but then we don't know the name of the product to display later in the Summary page.  I _think_ an object should work so I'll look into this.

src/openforms/appointments/tests/test_utils.py
```diff
@@ -80,7 +80,7 @@ def test_creating_appointment_with_missing_or_not_filled_in_appointment_informat
         submission = SubmissionFactory.create(form=form)
         SubmissionStepFactory.create(
             submission=submission,
-            data={"product": "79", "time": "2021-08-25T17:00:00"},
+            data={"product": "79-Paspoort", "time": "2021-08-25T17:00:00"},
```
sergei-maertens, line 83: Wrong format?

src/openforms/appointments/tests/test_utils.py
```diff
@@ -96,7 +96,8 @@ def test_creating_appointment_with_missing_or_not_filled_in_appointment_informat
             AppointmentInfo.objects.filter(
                 submission=submission,
                 status=AppointmentDetailsStatus.missing_info,
-                error_information="Missing information in form: locationID. Information not filled in by user: clientDateOfBirth. ",
+                error_information="Missing information in form: locationIDAndName. "
+                "Information not filled in by user: clientDateOfBirth. ",
```
sergei-maertens, line 100: _all_ user-facing strings must be marked translatable

src/openforms/submissions/tests/test_models.py
```diff
@@ -227,7 +227,11 @@ def test_get_merged_appointment_data(self):
         submission = SubmissionFactory.create(form=form)
         SubmissionStepFactory.create(
             submission=submission,
-            data={"product": "79", "location": "1", "time": "2021-08-25T17:00:00"},
+            data={
+                "product": "79-Paspoort",
```
sergei-maertens, line 231: old format?

sergei-maertens (changes requested):
Why aren't tests failing when some tests use the ID-NAME format and others use the object approach?

src/openforms/appointments/tests/test_utils.py
shea-maykinmedia: Yes

src/openforms/submissions/tests/test_models.py
shea-maykinmedia: Yes

shea-maykinmedia:
The tests that were using the old ID-NAME format but still passing are testing that the data being submitted in the form step is being returned in the correct key when calling the function to get the appointment related data.  So they are checking that the expected data is there but not what the expected data format should be.  Or they were testing the error flows of booking an appointment and so the data that was in the submission step didn't really matter, just that some data was there.

The tests that were testing the code around actually booking the appointment were previously updated to use the correct object format.

sergei-maertens:
I'll rebase and fix the merge conflicts after #664 is merged, as that one has touched the `AppointmentInfo` part.

sergei-maertens (approved):

PR #652, opened 2021-09-10
[#607] Renamed Form & FormDefinition.name to .public_name, added .internal_name

Bartvaderkin:
Complement: https://github.com/open-formulieren/open-forms-sdk/pull/59


joeribekker (changes requested):
Bart, please make a new branch and PR and re-do this by ONLY introducing `internal_name` and not rename name to public_name. The admin_name property is ofc a good call.

Changing 48 files, and an SDK change, and the chance that we missed a rename spot, are not needed or worth it.

Bartvaderkin:
Closing this as we went with a different approach.

PR #656, opened 2021-09-10
Logic actions bugs/improvements

SilviaAmAm:
**Problem 1**
The problem was that creating a logic action such as:

```
               {
                    "formStep": "",
                    "component": "surname",
                    "action": {
                        "type": "value",
                        "property": {},
                        "value": {"var": "name"},
                    },
                }
```

was raising a validation error because the form step is empty (while it is expected to be empty).

**Problem 2**

the function `evaluate_form_logic` was expecting form_step as a UUID but this had been changed to a URL, so was failing.

**Problem 3**

The actions of type 'value' were not changing the value of the concerned fields. This is because the configuration was being updated but not the FormIO submission data. So now the `evaluate_logic_check` sends the step data back to the front end which then also updates the submission data.

### DIFF
--- a/src/openforms/submissions/tests/form_logic/test_modify_components.py
+++ b/src/openforms/submissions/tests/form_logic/test_modify_components.py
@@ -530,6 +530,7 @@ class ComponentModificationTests(TestCase):
             ]
         }
         self.assertEqual(configuration, expected)
+        self.assertEqual({"step2_textfield1": "some value"}, submission_step_2.data)
 
     def test_evaluate_logic_with_empty_data(self):
         """
--- a/src/openforms/submissions/tests/form_logic/test_modify_components.py
+++ b/src/openforms/submissions/tests/form_logic/test_modify_components.py
@@ -708,7 +708,7 @@ class StepModificationTests(TestCase):
             },
             actions=[
                 {
-                    "form_step": str(step2.uuid),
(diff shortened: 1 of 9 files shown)


src/openforms/submissions/tests/form_logic/test_modify_components.py
```diff
@@ -708,7 +709,7 @@ def test_step_not_applicable(self):
             },
             actions=[
                 {
-                    "form_step": str(step2.uuid),
+                    "form_step": f"http://example.com{reverse('api:form-steps-detail', kwargs={'form_uuid_or_slug': form.uuid, 'uuid': step2.uuid})}",
```
sergei-maertens, line 712: Next time, please don't do the `reverse` call inside the f-string, it's very unreadable code

src/openforms/submissions/tests/form_logic/test_modify_components.py
```diff
@@ -861,7 +862,7 @@ def test_response_contains_submission(self):
             },
             actions=[
                 {
-                    "form_step": str(form_step2.uuid),
+                    "form_step": f"http://example.com{reverse('api:form-steps-detail', kwargs={'form_uuid_or_slug': form.uuid, 'uuid': form_step2.uuid})}",
```
sergei-maertens, line 865: Next time, please don't do the `reverse` call inside the f-string, it's very unreadable code

sergei-maertens (approved):

PR #659, opened 2021-09-14
[#642] Privacy policy checkbox

### LINKED ISSUE #642: As user I want to explicitely approve the use of my submitted data.
Part of #323 

The summary page should include something like:

image

I'll describe what needs to be done here:

- [x] Add option to show this checkbox on the summary page, in the global config
- [x] Add non-required field in te global config model: `privacy_policy_link`
- [x] Add required field in te global config model:  `privacy_policy_label` (default should be: "Ja, ik heb kennis genomen van het {{ privacybeleid }} en geef uitdrukkelijk toestemming voor het verwerken van de door mij opgegeven gegevens.")
- [x] The summary page shows a checkbox like the image above (but can probably be a one-liner) where the label is taken from the `privacy_policy_label`.
- [x] The word between curly braces can be replaced with a hyperlink to the `privacy_policy_link` (to keep it somewhat in line with other variables used in OF). If no link, then just remove the curly braces.
- [x] Only allow submission if the checkbox is checked.

SilviaAmAm:
Fixes #642 

Corresponding PR in SDK: https://github.com/open-formulieren/open-forms-sdk/pull/60


src/openforms/config/api/views.py
```diff
@@ -0,0 +1,32 @@
...
+    summary=_("Privacy policy info"),
+    responses={200: PrivacyPolicyInfoSerializer},
+)
+class PrivacyPolicyInfoView(APIView):
+    permission_classes = (AnyActiveSubmissionPermission,)
+
+    def get(self, request: Request) -> Response:
+        conf = GlobalConfiguration.get_solo()
+        if not conf.ask_privacy_consent:
+            return Response({"requires_privacy_consent": False})
```
joeribekker, line 24: Why not always use the serializer?

src/openforms/config/templatetags/privacy_policy.py
```diff
@@ -0,0 +1,18 @@
+from django import template
+from django.template import Context, Template
+
+from ..models import GlobalConfiguration
+
+register = template.Library()
+
+
+@register.simple_tag()
+def privacybeleid():
```
joeribekker, line 10: I think I prefer the English name as with most (all?) template tags we use now.

So, `privacy_policy`

joeribekker, line 10: I think I prefer the English name as with most (all?) template tags we use now.

So, `privacy_policy`

src/openforms/config/templatetags/privacy_policy.py
```diff
@@ -0,0 +1,18 @@
...
+    conf = GlobalConfiguration.get_solo()
+    if conf.privacy_policy_url:
+        template = '<a href="{{ privacy_policy }}">privacybeleid</a>'
+        return Template(template).render(
+            Context({"privacy_policy": conf.privacy_policy_url})
+        )
+    else:
+        return "privacybeleid"
```
joeribekker, line 18: Just make it return an empty string.

src/openforms/config/tests/test_privacy_policy.py
```diff
@@ -0,0 +1,39 @@
...
+        conf.privacy_policy_label = (
+            "Ja, ik heb kennis genomen van het {% privacybeleid %} en geef uitdrukkelijk "
+            "toestemming voor het verwerken van de door mij opgegeven gegevens."
+        )
+        conf.privacy_policy_url = ""
+        conf.save()
+
+        label = conf.render_privacy_policy_label()
+
+        self.assertEqual(
```
joeribekker, line 35: I see why you had it return this string but it's very context specific

joeribekker (changes requested):
Looks good, minor changes requested.

src/openforms/config/api/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import path
+
+from .views import PrivacyPolicyInfoView
+
+app_name = "config"
+
+urlpatterns = [
+    path(
+        "privacy_policy_info/",
```
sergei-maertens, line 9:
```suggestion
        "privacy_policy_info",
```
no trailing slashes in API endpoints

src/openforms/config/templatetags/privacy_policy.py
```diff
@@ -0,0 +1,18 @@
...
+@register.simple_tag()
+def privacy_policy():
+    conf = GlobalConfiguration.get_solo()
+    if conf.privacy_policy_url:
+        template_string = '{% load i18n %}<a href="{{ privacy_policy }}">{% trans "privacy policy" %}</a>'
+        return Template(template_string).render(
+            Context({"privacy_policy": conf.privacy_policy_url})
+        )
+
+    return ""
```
sergei-maertens, line 18: Please implement this as an [`inclusion_tag`](https://docs.djangoproject.com/en/2.2/howto/custom-template-tags/#inclusion-tags) instead.

src/openforms/config/tests/test_privacy_policy.py
```diff
@@ -0,0 +1,31 @@
...
+class PrivacyPolicyTests(TestCase):
+    def test_render_privacy_label(self):
+        conf = GlobalConfiguration.get_solo()
+        conf.privacy_policy_label = "I read the {% privacy_policy %} and agree."
+        conf.privacy_policy_url = "http://test-privacy-policy.nl"
+        conf.save()
+
+        label = conf.render_privacy_policy_label()
+
+        self.assertEqual(
```
sergei-maertens, line 15: use `self.assertHTMLEqual` (or something like that) which also parses and validates the HTML.

sergei-maertens (changes requested):

src/openforms/config/templatetags/privacy_policy.py
SilviaAmAm: I did this initially, but I couldn't figure out how to deal with the new lines that get added. So I went for a simple tag instead.

sergei-maertens: newlines should be ignored by the HTML related assertions such as https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.SimpleTestCase.assertHTMLEqual

SilviaAmAm: But even if the new lines are ignored by tests, it still looks funny to have a checkbox like this:
```
I have read the 
privacy_policy
and agree with it
```
instead of having it all in one line?

sergei-maertens: hrm it is odd that those are converted into linebreaks :thinking: let's get this merged as is then, and possibly improve it later

sergei-maertens (approved):

Issue #651, opened 2021-09-10
Implement submission status endpoint


sergei-maertens:
- [x] Auth relies on submission permission (~~session based~~ timestamped HMAC based)
- [x] ~~Remove session-cleanup from `_complete` endpoint (we need it for status endpoint checks)~~
- [x] Return root ID status: pending/errored/completed (see https://github.com/open-formulieren/open-forms/pull/617#discussion_r706061018)
- [x] Return contextual information
    - [x] (appointment) error message, if relevant
    - [x] generated PDF report download link
    - [x] ~~make payment endpoints token-based rather than session based & include links~~ not needed at this point
    - [x] confirmation page content (from template)
- [ ] Implement SDK as client/caller

joeribekker:
Copying my nice table here from earlier PR review:

Why don't we keep state cleaner:

`status`
* done (for all ready and exception states)
* in progress (for other states)

`result` (when `status` is *done*)
* retry (to have the user retry the submission)
* fail (submission failed, no point to retry the same submission)
* success (submission was received and all mandatory checks succeeded)

And, remove `processing_aborted`.

The combinations are then pretty clear:

`status` | `result` | action
--- | --- | ---
in progress | - | show spinner to user
done | retry | redirect user back to summary page, showing `error_message`
done | fail | redirect user back to start of the form, showing `error_message`
done | success | redirect/show user the confirmation page, displaying content/report/payment link

PR #664, opened 2021-09-15
Fix the celery chain handling triggerd by Submission on_completion

sergei-maertens:
Relates to #651 

Fixes the status/result interpretation from the celery task chain execution.

**Changes**

1. From the chain, we now "walk" all the nodes to obtain all the task IDs of the tasks making up the chain
2. We store all the task IDs on the submission model
3. When checking the status, the chain is "done" when either: any task fails or all tasks are finished
4. The result depends on any task failing or all tasks being a success

Verified by creating a faulty appointment submission through the management command and having a celery worker running + checking the token-based status endpoint in the browser, and it worked as expected.


src/openforms/submissions/status.py
```diff
@@ -22,29 +22,37 @@ class SubmissionProcessingStatus:
...
+        if all_success:
             return ProcessingResults.success
-        if result.state == states.REVOKED:
-            return ProcessingResults.retry
-        return ProcessingResults.failed
+        if any_failed:
+            return ProcessingResults.failed
+
+        # TODO: not sure if we can actually get to this? Maybe this should be removed.
+        return ProcessingResults.retry
```
joeribekker, line 55: As discussed, we can delete this and consider "failed" as a user recoverable error.

joeribekker (approved):

src/openforms/submissions/status.py
sergei-maertens: will tackle this while implementing the SDK aspect

PR #665, opened 2021-09-15
Improvements/appointments tab

### LINKED ISSUE #612: Move appointment configuration to separate form builder tab
... rather than having extra FormIO options defined in the different components.

After #573 is merged, we should:

**Tasks**

- [x] Remove custom FormIO component schema items (`appointmentBirthDate`, `appointmentLastName`)
- [ ] Add a field `forms.Form.appointmentconfiguration` (JSONField) where the pointers to the form definition components are stored
- [x] In the UI, present dropdowns to select the appropriate component (re-use the `ComponentSelection` component from the form logic)

shea-maykinmedia:
Fixes #612 and the last points of #597 

This removes the individual appointments tab from the components and adds a new react tab in the form view (see screenshot)

SDK PR: https://github.com/open-formulieren/open-forms-sdk/pull/61

Two things to note:
1. With this we will only be able to set up appointment configuration for the form with react admin form view.
2. I didn't do `Add a field forms.Form.appointmentconfiguration (JSONField) where the pointers to the form definition components are stored` since the current approach is more inline with how the appointments were configured before.  If we do want this point then I can implement it but it will be require more work.

**Screenshot**

Screenshot 2021-09-16 at 12 15 18


shea-maykinmedia:
I'm not a fan of a lot of the code in the `Object.enteries().map()` but I'm not sure if there is a better way to write it.  Suggestions are appreciated ;)

sergei-maertens (approved):

PR #671, opened 2021-09-15
More submission status bugfixes

sergei-maertens:


src/openforms/submissions/api/viewsets.py
```diff
@@ -92,7 +95,7 @@ def perform_create(self, serializer):
         summary=_("Complete a submission"),
         request=None,
         responses={
-            204: None,
+            200: SubmissionCompletionSerializer,
             400: CompletionValidationSerializer,
```
joeribekker, line 99: I can't put my finger on it, but I don't really like this difference in response serializers. That said, the HTTP codes are correct and the responses can/should indeed differ. This 400 response is different from other 400 responses but that's also not really a problem...

So, nevermind this comment hehe.

joeribekker (approved):

src/openforms/submissions/api/viewsets.py
sergei-maertens: that whole submission completion validation is still in its infancy anyway so I wouldn't worry about it too much yet :)

PR #676, opened 2021-09-16
Fix scss that conflicts between django admin and formio builder

### LINKED ISSUE #643: As user, I always want to see the form field menu, so I can create long form pages
**Tasks**

* [x] In the admin, keep the form field menu of Open Forms in view (let it scroll up/down with the browser view).

image

shea-maykinmedia:
Fixes #643 

This reason for it not scrolling was that `form-row` css class for the django admin has `overflow: hidden;` but this causes the `position: sticky` in the Formio css to not work properly.  If we `unset` the `overflow` property for this field then things work as expected.

I also added the `top: 100px` so that the header bar in the django admin doesn't start to overlap part of the Formio components when scrolling.

**Screenshots**

Screenshot 2021-09-16 at 16 14 52

Screenshot 2021-09-16 at 16 14 59

Screenshot 2021-09-16 at 16 15 06

### DIFF
--- a/src/openforms/scss/components/builder/_builder.scss
+++ b/src/openforms/scss/components/builder/_builder.scss
@@ -5,6 +5,11 @@
 @import '~@fortawesome/fontawesome-free/scss/solid';
 @import '~formiojs/dist/formio.full.css';
 
+
+.field-configuration {
+  overflow: unset;
+}
+
 .field-configuration,
 .form-definition {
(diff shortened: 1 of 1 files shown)


sergei-maertens:
Self-reminder: check if the modals still work properly

sergei-maertens (approved):

PR #681, opened 2021-09-17
[#244] reCAPTCHA

SilviaAmAm:
Backend part of #244 

Frontend: https://github.com/open-formulieren/open-forms-sdk/pull/65


sergei-maertens (changes requested):
Code seems okay, but why are these Django settings rather than extra configuration options in the `config.GlobalConfiguration` model? We already configure all analytics stuff there...

joeribekker:
Let's discuss this ticket first before continueing in this ticket

PR #692, opened 2021-09-21
Handle stuf bg bugs

### LINKED ISSUE #647: When StUF-BG fails the form does not render at all.
This happened after logging in with DigiD, when the first page should display some prefilled user information. This needs to be checked in the logs.

**Update**

Current behaviour is this error:

The first step loads until StUF-BG service replies (expected). StUF-BG fails leading to an error in the API response, resulting in an empty first step in the form. Thus,

`/api/v1/submissions/d9dae2ad-dd32-43d8-9705-338e4fd42d65/steps/c8b1f7c7-2fbf-472b-ba39-7914dade5745?`
returns a HTTP 502 due to a Python error: 
```
  File "/app/src/./openforms/prefill/contrib/stufbg/plugin.py", line 62, in get_prefill_values
    data = dict_response["Envelope"]["Body"]["npsLa01"]["antwoord"]["object"]
KeyError: 'npsLa01'
```

caused by this SOAP response:
```
<?xml version="1.0" encoding="UTF-8"?>
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/">
    <soapenv:Body>
        <soapenv:Fault>
            <faultcode>soapenv:Server</faultcode>
            <faultstring>Policy Falsified</faultstring>
            <faultactor>[snip: stuf-bg-service-url-here]</faultactor>
            <detail>
                <l7:policyResult status="Error in Assertion Processing" xmlns:l7="http://www.layer7tech.com/ws/policy/fault"/>
            </detail>
        </soapenv:Fault>
    </soapenv:Body>
</soapenv:Envelope>
```

**Original**

```
[...]
  File "/app/src/./openforms/prefill/contrib/stufbg/plugin.py", line 61, in get_prefill_values
    data = dict_response["Envelope"]["Body"]["npsLa01"]["antwoord"]["object"]
TypeError: 'NoneType' object is not subscriptable
```

**Tasks**

* [x] Handle Python SOAP-errors in such a way the frontend can continue. For prefill, this means the values can simply not be prefilled.
* [x] This should be logged in Django timeline logger (and in Sentry please with traceback).
* [x] Handle SOAP-Faults (see XML above) specifically to include the faultstring and details in the timeline logger (I thought this was done already but the SOAP-service is spread out a bit over the code).

shea-maykinmedia:
Fixes #647 

From what I could see the **original** error is still happening on the X test environment with the Test prefill personen form.  The error is that an empty `antwoord` response was being returned in the SOAP call and so we were getting a `TypeError` instead of a `KeyError` because `antwoord` was an empty string instead of the key not existing.

The **updated** error should have previously been caught because we are catching `KeyError`s however now the `Fault` will be logged.

I wasn't able to test this fix live because I'm not sure which SOAP environment will return these errors but I did take the SOAP response from the github issue and the **original** one from Sentry and wrote unit tests to handle these cases and double checked it by runnning locally and mocking the SOAP call with these responses.

I also didn't test how this will look in Sentry because I didn't want to send stuff from my local environment into our open-forms sentry but if you do want to do this let me know.

If there are suggestions for testing this further I'm happy hear :)


sergei-maertens (approved):

sergei-maertens:
Let's just test this out

PR #699, opened 2021-09-22
Improve user groups

### LINKED ISSUE #257: Als beheerder wil ik dat de beveiliging van de webapplicaties van de oplossing voldoet aan de NCSC ICT-beveiligingsrichtlijnen ...
...voor webapplicaties.

Eis: Beveiliging En Privacy 3.

**Invulling**

Deze staan hier beschreven: https://www.ncsc.nl/documenten/publicaties/2019/mei/01/ict-beveiligingsrichtlijnen-voor-webapplicaties

Een deel hiervan wordt getackled door ISO 2700x:

image

Maatregelen genoemd in bovenstaande documenten worden opgevolgd. Ten behoeve hiervan zijn nog enkele taken nodig:

**Tasks**

* [x] Vaste groepen aanmaken voor gebruikers: Redacteurs (mogen formulieren maken), Functioneel beheerders (mogen configuratie aanpassen en accounts aanmaken), Beheerders (mogen alles, geen superuser), Behandelaars (mogen inzendingen inzien) (Beveiligingsrichtlijn B.02)
* [x] Toon groepen en superuser flags in User list view zodat de autorisatie matrix helder is.
* [x] Nginx headers zoveel mogelijk beperken (vraag Alex).
* [x] Stel voor alle cookies de flags ‘secure’ en ‘HttpOnly’ in (zie Django docs). Getackled [hier](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/conf/base.py#L425) en [hier](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/conf/base.py#L418).

shea-maykinmedia:
Fixes #257 

For the first point `Vaste groepen aanmaken voor gebruikers` I made a management command because I noticed on https://open-forms.test.maykin.opengem.nl/ we already had some groups set up and I'm guessing other environments already have groups set up so this seemed like the best way to accomplish this point while still working with the current environments that are already set up.

The `Redacteur` and `Functional beheer` groups were already created on https://open-forms.test.maykin.opengem.nl/ so I follow the permissions that were created there.  For the `Behandelaar` group I believe they only need the single permission. 

The last two points of the ticket were already done so this PR does not contain anything related to those two points.


sergei-maertens (changes requested):
This test environment is just a test environment. Joeri discussed this that it should be a fixture. Remove all this management code and replace it with a simple fixture.

Additionally, this is one of the fixtures that should automatically be loaded like the admin index fixture, which is hooked up to the `post_migrate` signal I think.

src/openforms/accounts/models.py
```diff
@@ -63,3 +63,8 @@ def get_full_name(self):
     def get_short_name(self):
         "Returns the short name for the user."
         return self.first_name
+
+    def get_groups(self):
+        return list(self.groups.all())
```
sergei-maertens, line 68: I cannot imagine this looks nice in the admin - this should just be a comma-separated list. Furthermore, this should be defined on the ModelAdmin as a helper method that does the proper formatting.

Additionally, you must modify the admin queryset to prefetch the groups, because with large amounts of users in the admin this will spawn an excessive amount of queries in the list view.

sergei-maertens (changes requested):

src/openforms/accounts/models.py
shea-maykinmedia: Admin wise it is rendered as a comma separated list.

Screenshot 2021-09-22 at 16 36 45

Do you still want this be a ModelAdmin helper method?

I will modify the admin queryset to prefetch the groups though.

sergei-maertens: > Do you still want this be a ModelAdmin helper method?

Yes - it is only relevant for the admin so let's not pollute the models with that. Additionally, that allows you to see the prefetch query & usage closely together instead of it being spread out over multiple files.

shea-maykinmedia:
Is there a reason we want a fixture vs a management command?  It seems like if we're going to call `loaddata` on a fixture we can just call a management command?

src/openforms/accounts/apps.py
```diff
@@ -20,8 +20,24 @@ def update_admin_index(sender, **kwargs):
     call_command("loaddata", "default_admin_index", verbosity=0)
 
 
+def update_groups(sender, **kwargs):
+    from django.contrib.auth.models import Group
+
+    Group.objects.all().delete()
```
sergei-maertens, line 26: This breaks all user-group relations! Do not do this

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #706, opened 2021-09-23
[#225] Skip not applicable steps

### LINKED ISSUE #225: Een gebruiker kan condities in formulieren opnemen. De gebruiker krijgt alleen die stappen of velden gepresenteerd ...
...die bij de ingestelde condities horen.

Eis: Elementen 15.
Zie ook NvI: 151

**Tasks**

- [x] Bug remains that you can go back a step, even if it is marked as n.v.t.

SilviaAmAm:
Fixes #225 

Front end PR: https://github.com/open-formulieren/open-forms-sdk/pull/73

Note: 
If the appointment fields are spread over multiple form steps and some of the steps are applicable and some not, when submitting the form there will be an error that the appointment data is not valid.

### DIFF
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -1,5 +1,6 @@
 import logging
 from dataclasses import dataclass
+from typing import Optional
 
 from django.conf import settings
 from django.core.mail import send_mail
@@ -334,7 +335,7 @@ class SubmissionStateLogicSerializer(serializers.Serializer):
 @dataclass
 class SubmissionStateLogic:
     submission: Submission
-    step: SubmissionStep = None
+    step: Optional[SubmissionStep] = None
 
 
 class SubmissionCompletionSerializer(serializers.Serializer):
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -16,7 +16,7 @@ from openforms.forms.models import FormStep
 
 from ...forms.validators import validate_not_maintainance_mode
 from ..constants import ProcessingResults, ProcessingStatuses
-from ..form_logic import evaluate_form_logic
+from ..form_logic import check_submission_logic, evaluate_form_logic
 from ..models import Submission, SubmissionStep, TemporaryFileUpload
 from .fields import NestedRelatedField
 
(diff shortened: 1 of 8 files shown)


sergei-maertens:
> Note:
> If the appointment fields are spread over multiple form steps and some of the steps are applicable and some not, when submitting the form there will be an error that the appointment data is not valid

This doesn't worksas intended/expected anyway because of how the `refreshOn` are configured in the frontend, so doing that is unsupported at the moment. Good to note it though!

src/openforms/submissions/api/serializers.py
```diff
@@ -330,7 +334,7 @@ class SubmissionStateLogicSerializer(serializers.Serializer):
 @dataclass
 class SubmissionStateLogic:
     submission: Submission
-    step: SubmissionStep
+    step: SubmissionStep = None
```
sergei-maertens, line 337:
```suggestion
    step: Optional[SubmissionStep] = None
```

src/openforms/submissions/form_logic.py
```diff
@@ -7,7 +7,6 @@
 
 from ..forms.models.form import FormLogic
 from ..prefill import JSONObject
-from .models import Submission, SubmissionStep
```
sergei-maertens, line 10: please replace this with a conditional import, otherwise linters will complain:

```py
from typing import TYPE_CHECKING

...

if TYPE_CHECKING:
    from .models import Submission, SubmissionStep
```

src/openforms/submissions/form_logic.py
```diff
@@ -81,3 +80,27 @@ def evaluate_form_logic(
     step._form_logic_evaluated = True
 
     return configuration
+
+
+def check_submission_logic(submission):
+    logic_rules = FormLogic.objects.filter(
+        form=submission.form,
+        actions__contains=[{"action": {"type": "step-not-applicable"}}],
```
sergei-maertens, line 88: can we have the action types defined in a constant somewhere and use those labels instead of magic strings?

src/openforms/submissions/form_logic.py
```diff
@@ -81,3 +80,27 @@ def evaluate_form_logic(
...
+            for action in rule.actions:
+                if action["action"]["type"] != "step-not-applicable":
+                    continue
+
+                step_to_modify_uuid = resolve(furl(action["form_step"]).pathstr).kwargs[
+                    "uuid"
+                ]
+                submission_step_to_modify = submission_state.get_submission_step(
+                    form_step_uuid=step_to_modify_uuid
+                )
```
sergei-maertens, line 105: This is now duplicated in a couple of places - I suggest adding a method `SubmissionState.resolve_step(resource_url: str) -> SubmissionStep` which hides the implementation details

sergei-maertens (changes requested):
Cool! Some code cleanup things, but looks neat in general.

sergei-maertens (approved):

PR #707, opened 2021-09-23
[#264] Integrated payments into confirmation-email flow 

Bartvaderkin:


src/openforms/emails/utils.py
```diff
@@ -27,7 +36,8 @@ def sanitize_content(content: str) -> str:
     config = GlobalConfiguration.get_solo()
 
     # strip out any hyperlinks that are not in the configured allowlist
-    replace_urls = partial(sanitize_urls, config.email_template_netloc_allowlist)
+    allowlist = get_system_netloc_allowlist() + config.email_template_netloc_allowlist
+    replace_urls = partial(sanitize_urls, allowlist)
     stripped = re.sub(URL_REGEX, replace_urls, content)
```
joeribekker, line 41: Can we add a logging.debug here that logs the URLs that were replaced?

src/openforms/payments/contrib/ogone/plugin.py
```diff
@@ -84,11 +88,11 @@ def handle_return(self, request, payment: SubmissionPayment):
 
     def handle_webhook(self, request):
         # unvalidated data
-        order_id = request.data.get("ORDERID")
+        order_id = request.data.get("orderID")
```
joeribekker, line 91: Are we sure about the casing? I remember it was called with uppercase?
If it can differ we should maybe sanitize all values to uppercase or something?

src/openforms/payments/contrib/ogone/plugin.py
```diff
@@ -99,10 +103,11 @@ def handle_webhook(self, request):
         except InvalidSignature as e:
             logger.warning(f"invalid SHASIGN for payment {payment}")
             logevent.payment_flow_failure(payment, self, e)
-            return HttpResponseBadRequest("bad shasign")
+            return HttpResponseBadRequest("bad shasign"), None
```
joeribekker, line 106: I see a mix of raise 404 (`get_object_or_404`) and return `HttpResponseBadRequest`.

Shouldn't we choose just to raise an exception instead of return `HttpResponseBadRequest`? This way, we also don't get the ackward tuple, and simply catch the exception in the view to return a `HttpResponseBadRequest`?

src/openforms/registrations/contrib/email/plugin.py
```diff
@@ -24,6 +25,12 @@ class EmailRegistration(BasePlugin):
     def register_submission(self, submission: Submission, options: dict) -> None:
         submitted_data = submission.get_merged_data()
 
+        # explicitly get a reference
+        set_submission_reference(submission)
+
+        subject = _("[Open Forms] {} - submission {}").format(
```
joeribekker, line 31: Can you please use named formatting? This makes translation easier :)

I saw a few cases but just mentioning this instance

src/openforms/submissions/admin.py
```diff
@@ -102,6 +102,12 @@ def has_add_permission(self, request, obj=None):
     def has_delete_permission(self, request, obj=None):
         return False
 
+    def order_id_str(self, obj):
+        return str(obj.order_id)
```
joeribekker, line 106: Should order ID not be a string anyway?

joeribekker (changes requested):
Looks good, minor comments.

PR #686, opened 2021-09-20
[#247] Added timeline logger and implemented logging events

Bartvaderkin:
Fixes #247

PR #715, opened 2021-09-24
Add additional logging

### LINKED ISSUE #260: De gebruiker werkt veilig met formulieren. De toepassing voldoet aan de Algemene verordening gegevensbescherming ...
...(AVG). Dit betekent dat:

a. De leverancier heeft bij de softwareontwikkeling privacy by design ingericht. Daarnaast heeft de leverancier privacy by default ingericht binnen de processen en systemen.
b. Er wordt gelogd wie waarom en wanneer welke persoonsgegevens verwerkt. Deze logging wordt eenvoudig ontsloten voor zowel de privacy-functionarissen van de gemeenten, als ook de inwoners wiens gegevens het betreft.
c. Wie heeft wanneer, en waarom welke gegevens 'verwerkt' 
 1. Voor welke handelingen en pogingen daartoe vindt logging plaats
 2. Voor welke elementen worden dergelijke handelingen gelogd.
 3. Wat er precies wordt vastgelegd.
 4. De wijze waarop gegevensuitwisseling via koppelvlakken wordt gelogd en geaudit.

d. Hoe lang de logs bewaard blijven, incl. de flexibiliteit mbt bewaartermijnen en de verschillende logs
e. In een logregel worden geen geval gevoelige gegevens opgenomen. Dit betreft niet alleen
persoonsgegevens maar ook gegevens waarmee de beveiliging doorbroken kan worden (zoals
wachtwoorden, inbelnummers, enz.)
f. Logbestanden worden zodanig beschermd dat deze niet aangepast of gemanipuleerd kunnen
worden.

Eis: Beveiliging En Privacy 6.
Zie ook NvI: 58, 113

**Tasks**

See also #247 to introduce timeline logger.
Specifically add logging statements for the Prefill plugins StUF-BG and HaalCentraal to...

* [x] Log the retrieval of personal data and which fields specifically and which attribute is used from the API/SOAP mapping. (see example below)
* [x] Log the viewing of admin submission details
* [x] Log the viewing of api submission details
* [x] Log the admin (and api?) export of submissions
* [x] These logs need to be tagged or labelled to identify AVG-related logs
* [x] These logs need a special view in the admin to see them (filtered on above label)

**Example: Prefill logging**

```
[2021-09-06 10:00:18] (Submission ae6f32): Prefill plugin "MyBG-service" (StUF-BG) retrieved personal data to prefill fields: Firstname (persoon.voornamen), Lastname (persoon.achternaam), Birthdate (persoon.geboortedatum).
```

**Example: User logging**

```
[2021-09-06 10:00:18] User "John Doe" viewed submission ae6f32 for form "Request passport".
```

shea-maykinmedia:
Fixes #260


sergei-maertens:
#686 is merged - please rebase on master :)

shea-maykinmedia:
Will do!

src/openforms/logging/models.py
```diff
@@ -93,3 +125,19 @@ def message(self) -> str:
...
+
+
+class AvgTimelineLogProxy(TimelineLogProxy):
+
+    objects = TimelineLogProxyManager()
+
+    class Meta:
+        proxy = True
+        verbose_name = _("avg timeline log entry")
+        verbose_name_plural = _("avg timeline log entries")
```
shea-maykinmedia, line 143: This is for the point `These logs need a special view in the admin to see them (filtered on above label)`.

To me it seemed that the best way to accomplish this was to create another proxy model which can then be used in the admin.  Maybe there is a better way?

shea-maykinmedia:
One thing I was thinking about with this logging is that we never seem to log the user.  There is the `fmt_user` function but it seems we never add the user to the `TimelineLogProxy` instance when we create it.  I'm not sure if this was intentional or if I was supposed to add that as part of this pull request?  Just wanted to mention it and if I'm supposed to add it let me know and I will do that. :)

Bartvaderkin:
@shea-maykinmedia We don't set a user for the form submission logging because it is anonymous and user objects are not created. The code still supports it in `fmt_user` to stay compatible with timelinelogger functionality (like if we would log admin interaction).

src/openforms/logging/logevent.py
```diff
@@ -348,3 +353,27 @@ def appointment_cancel_failure(appointment: "AppointmentInfo", plugin, error):
         plugin=plugin,
         error=error,
     )
+
+
+def view_submission_details_admin(submission: "Submission"):
```
Bartvaderkin, line 358: I would re-order the naming scheme so we get a bit of a namespace/grouping (from large to small terms), so if you put all of the events in a list related things group together

`view_submission_details_admin` -> `submission_details_view_admin`

`export_submissions` -> `submission_export_list`

src/openforms/logging/models.py
```diff
@@ -93,3 +125,19 @@ def message(self) -> str:
         return self.get_message()
 
     message.short_description = _("message")
+
+
+class TimelineLogProxyManager(models.Manager):
```
Bartvaderkin, line 130: I would name this `AVGTimelineLogProxyManager`

Bartvaderkin (changes requested):
I added a few minor comments.

src/openforms/logging/logevent.py
```diff
@@ -40,10 +44,18 @@ def _create_log(
     if error:
         extra_data["error"] = str(error)
 
+    extra_data["avg"] = tag_avg
+
+    if user and not isinstance(user, User):
+        # If user is not of correct type (eg. AnonymousUser) we can not
+        #   save it on the TimelineLogProxy model
+        user = None
```
shea-maykinmedia, line 52: I'm not sure if there is a better way to do this.  The `user` we save on the TimelineLog model needs to be what we use for `settings.AUTH_USER_MODEL`.  The current approach has two issues:

1. If we can the model we use for `settings.AUTH_USER_MODEL` this would break, though I think it's unlikely we'll change this setting.
2. It means we have to import `User` twice.  Once for the type checking and once in the function.  If we try importing it outside of the `if TYPE_CHECKING` then we get the `"AppRegistryNotReady: Apps aren't loaded yet."` exception and if we don't import it within the function then if type checking is turned off this code would break since the model would not be imported at all.

I'm happy to hear possible suggestions about how to improve this. :)

joeribekker: You can use `User = django.contrib.auth.get_user_model()` to get the model that was configured for the User model.

However, there's no need for this I think. Simply do: `if user.is_authenticated`. See: https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#attributes

src/openforms/logging/logevent.py
```diff
@@ -106,11 +118,13 @@ def pdf_generation_skip(submission: "Submission", submission_report):
...
 
 
-def prefill_retrieve_success(submission: "Submission", plugin):
+def prefill_retrieve_success(submission: "Submission", plugin, prefill_fields):
     _create_log(
         submission,
         "prefill_retrieve_success",
+        extra_data={"prefill_fields": prefill_fields},
         plugin=plugin,
+        tag_avg=True,
```
joeribekker, line 127: Can we make this more generic: Instead of having a parameter `tag_avg=True` we could do something like `tags=[TimelineLogTags.AVG]`?

I'm not sure if this will work in querysets though, like [here](https://github.com/open-formulieren/open-forms/pull/715/files#diff-2b3c27113b65828e17d17771483ffba20f8327d8449195d7fe02ed743cd0332cR133)

joeribekker (changes requested):
Looks good, one change needed. The change for tags instead of tag_avg is optional.

Issue #728, opened 2021-09-27
Issues noticed during demo sprint 7


sergei-maertens:
* [ ] Overzicht stap is altijd grijs -> some styling depending on if all (required) steps are completed or not?
* [ ] English "Log in or start the form anonymoysly" -> not all literals marked as `FormattedMessage` yet
* [ ] Paginering toevoegen aan logging view
* ~~Invullen formulier lijkt te haperen? Lukt niet meer om datum in te vullen? Achternaam borkt?~~ https://github.com/open-formulieren/open-forms-sdk/issues/79
* [ ] Afspraak annuleren geeft nog een javascript popup als je verkeerd emailadres invult -> generic `alert` on 403 should be replaced/caught with `ErrorBoundary` or something like that. Depends on if we expect it or not.
* [ ] Hijack werkt niet met 2FA

Bartvaderkin:
I looked into the 2FA+hijack issue but I couldn''t find a trivial way to do this: the 2FA does a few complicated things with the session store and OTP devices I couldn't easily fake/emulate in django hijack.

PR #738, opened 2021-09-29
Paginate TimelineLogProxyAdmin list view

shea-maykinmedia:
Fixes third point of #728 

**Screenshot**

Screenshot 2021-10-01 at 11 36 59

### DIFF
--- a/src/openforms/logging/templates/logging/admin_list.html
+++ b/src/openforms/logging/templates/logging/admin_list.html
@@ -1,5 +1,10 @@
 {% extends "admin/base_site.html" %}
-{% load i18n timeline %}
+{% load i18n timeline static %}
+
+{% block extrastyle %}
+    {{ block.super }}
+    <link rel="stylesheet" type="text/css" href="{% static "admin/css/changelists.css" %}">
+{% endblock %}
 
 
 {% block content %}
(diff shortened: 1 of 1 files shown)


sergei-maertens (changes requested):
The admin pages themselves are paginated, the view that is not paginated is http://localhost:8000/admin/logging/timelinelogproxy/auditlog/ - or rather, the view itself is paginated, but there are no pagination controls in the template

src/openforms/logging/templates/logging/admin_list.html
```diff
@@ -10,4 +15,19 @@
...
+    <div id="changelist">
+        <p class="paginator">
+            {% for i in paginator.page_range %}
+                {% if page_obj.number == i %}
+                    <span class="this-page">{{ i }}</span>
+                {% else %}
+                    <a href="?page={{ i }}" class="end">{{ i }}</a>
+                {% endif %}
+            {% endfor %}
+            {{ paginator.count }} {% if paginator.count == 1 %}{% trans "timeline log entry" %}{% else %}{% trans "timeline log entries" %}{% endif %}
```
shea-maykinmedia, line 29: I couldn't find a way to access the model's verbose name's in the template, since it's stored in `_meta`.  All the suggestions I saw online mentioned adding a template tag.  Not sure if we want that or if this is okay?

sergei-maertens: I can live with it :P the django admin typically puts `Model._meta` in the context as `opts`.

However, the correct django approach is to use the `{% plural %}` tag in `{% blocktrans %}`, see https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#blocktranslate-template-tag

sergei-maertens (approved):
Approving to get something working, but this will quickly show a very large list of page numbers and I think the django admin out of the box only shows a limited number of page numbers with an ellipsis in between. Would have to check the `ChangeList` class for that, I think.

PR #743, opened 2021-09-30
Fix z-index for adding a link

### LINKED ISSUE #731: Cannot use "link button" in Content Component
When you click the link button nothing seems to happen. 

Background windows becomes active where you can then type. Does not see normal behavior. 
issue31

shea-maykinmedia:
Fixes #731 

The cause of the bug is that we're setting the z-index of the FormIO popup to be higher than what the ckeditor was using for it's popup and so the 'add link' popup would be behind the FormIO popup.  By setting them to the same value the bug no longer occurs.

I also looked to see if any other options would be affected but I don't see any popups in this component or other ones.

**Screenshot**

Screenshot 2021-09-30 at 11 33 07

### DIFF
--- a/src/openforms/scss/components/builder/_builder.scss
+++ b/src/openforms/scss/components/builder/_builder.scss
@@ -38,7 +38,9 @@ li.nav-item {
   }
 }
 
-.formio-dialog {
+.formio-dialog,
+.ck-body .ck.ck-balloon-panel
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
sigh... again bootstrap and the formio stuff competing with django admin style.

PR #746, opened 2021-09-30
Update generate schema command to fail on warnings

### LINKED ISSUE #726: Set up CI job to check for schema generation warnings
drf-spectacular emits warnings for schema generation when things are ambiguous that it can't solve itself. This usually indicates a mistake in the schema documentation or some conflict with serializers in other apps, for example.

The warnings show in `runserver` when checking the API docs and also when generating the schema via the management command. CI should be set up to turn these warnings into errors to prevent errors from appearing in the API documentation.

This could be part of an existing job: https://github.com/open-formulieren/open-forms/blob/master/.github/workflows/ci.yml#L222

shea-maykinmedia:
Fixes #726 

Following the last `Note` from https://drf-spectacular.readthedocs.io/en/latest/client_generation.html#client-generation I've added the optional arguments so that if there are warnings then this job will fail and the build itself will fail as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Generate OAS
         run: |
-          python src/manage.py spectacular --lang=en > openapi.yaml
+          python src/manage.py spectacular --validate --fail-on-warn --lang=en --file openapi.yaml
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
so elegant :heart_eyes:

Issue #720, opened 2021-09-27
Errors in the Admin form editor are unclear


SilviaAmAm:
When an error happens, at the moment I look at the console errors to figure out what I need to change in the form definition. This should be displayed above the fields with mistakes.

sergei-maertens:
I had written this down too :) We need to set up a generic system to display validation errors in the admin. I've done a small bit of that already in the SDK, see https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/appointments/CancelAppointment.js#L72

sergei-maertens:
Screencast of current state in dev: https://watch.screencastify.com/v/YDdKIVUYQQ0UB47HKpfS

PR #752, opened 2021-10-01
Display validation errors

sergei-maertens:
Partially solves #720


joeribekker (approved):
Merged early to prevent code rot. Updated ticket with this change.

Issue #644, opened 2021-09-09
Translate upload text to Dutch


joeribekker:
![image](https://user-images.githubusercontent.com/96970/132691942-7bfc6f55-ef3b-4c6e-a262-a45a4e810ae9.png)

sergei-maertens:
Broader issue - _all_ FormIO literals need to be hooked up the the translation machinery. We need investigation on _how_ to do this though. :)

PR #755, opened 2021-10-01
[#644] Translate Formio in admin

SilviaAmAm:
Initial attempt to get translation of the Formio labels in the admin

#644


joeribekker (commented):
@SilviaAmAm where is the line from the ticket "Drop files to attach, or browse"?

SilviaAmAm:
I ran a different parser on the FormIO source, it picked up different strings :thinking: Maybe has to do with the configuration of the parser

PR #767, opened 2021-10-04
Issue/688 plugin tester

assignon:
- test_config(..) added to openforms.registrations.base.BasePlugin as abstract method and
Implemented for the ...zgw_apis.plugins.ZGWRegistration, Stuf_zds, Object api and Email plugins to test if the configuration is valid

- New view added to the layout of the admin to show the status of this plugin with the columns: label, plugin name, test_config result shown as "true/false/n.a." icon 
- Unit test added to tests the test_config() for the plugins


Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
joeribekker, line 1: This file should not be here. We don't work with Pipfiles.

requirements/dev.txt
```diff
@@ -56,7 +56,7 @@ cached-property==1.5.2
     #   -c requirements/ci.txt
```
joeribekker, line 56: The requirement files should not be changed unless the `*.in` files are changed. The `.in`-files control the libraries that are required.

src/openforms/prefill/contrib/kvk/constants.py
```diff
@@ -9,7 +9,7 @@ class Attributes(DjangoChoices):
 
     spec:    https://developers.kvk.nl/cms/api/uploads/api_zoeken_80071c75c9.yaml
     schema:  ResultaatItem
-    command: manage.py generate_prefill_from_spec --schema ResultaatItem --url https://developers.kvk.nl/cms/api/uploads/api_zoeken_80071c75c9.yaml
+    command:  generate_prefill_from_spec --schema ResultaatItem --url https://developers.kvk.nl/cms/api/uploads/api_zoeken_80071c75c9.yaml
```
joeribekker, line 12: ?

src/openforms/registrations/admin.py
```diff
@@ -1,4 +1,11 @@
 from .fields import RegistrationBackendChoiceField
+from django.contrib import admin
+from .contrib.zgw_apis.models import TestPlugin
+from .contrib.zgw_apis.plugin import ZGWRegistration as zwgr
+from .contrib.stuf_zds.plugin import StufZDSRegistration as szr
+from .contrib.objects_api.plugin import ObjectsAPIRegistration as oar
+from .contrib.email.plugin import EmailRegistration as er 
+from .contrib.demo.plugin import DemoFailRegistration as dfr
```
joeribekker, line 8: Imports from another package should use the full path. Since this is in the `registrations` package, all other paths should be absolute.

src/openforms/registrations/base.py
```diff
@@ -60,3 +60,10 @@ def update_payment_status(self, submission: "Submission"):
 
     def get_label(self):
         return self.verbose_name
+
+    @abstractmethod
+    def test_config(self):
```
joeribekker, line 65: The return type should be set on the base method. I used bool as example but, read the plugin comments first.

```suggestion
    def test_config(self) -> bool:
```

src/openforms/registrations/admin.py
```diff
@@ -12,3 +19,22 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
             return field
 
         return super().formfield_for_dbfield(db_field, request, **kwargs)
+
+
+class TestPluginAdmin(admin.ModelAdmin):
```
joeribekker, line 24: This admin model is more something for the config app. Move it there. Eventually, I want to test ALL plugins and not just the registration pluguins.

Also, there is no need to use the `admin.ModelAdmin` at all, since we don't have any (real) models to use it with. We can just add a TemplateView like done here (done with FormView but we don't need a form): https://github.com/open-formulieren/open-forms/blob/master/src/openforms/urls.py#L32-L36

src/openforms/registrations/admin.py
```diff
@@ -12,3 +19,22 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
...
+    change_list_template = 'admin/plugin_tester.html'
+
+    def changelist_view(self, request, extra_context=None):
+        context = extra_context or {}
+        context = {'plugins': [
+            {'name': 'ZGW API', 'test': zwgr.test_config()},
+            {'name': 'StUF-ZDS', 'test': szr.test_config()},
+            {'name': 'Objects API', 'test': oar.test_config()},
+            # {'name': 'Local', 'test': dfr.test_config()},
+        ], 'email': {'name': 'Email', 'test': er.test_config()},}
```
joeribekker, line 36: This works, but it's nicer to just iterate over all plugins in the `openforms.registratrations.registry.register`. We also don't need to import each plugin in that case.

src/openforms/registrations/contrib/demo/plugin.py
```diff
@@ -26,6 +26,9 @@ def get_reference_from_result(self, result: None) -> NoReturn:
     def update_payment_status(self, submission: "Submission"):
         print(submission)
 
+    def test_config(self) -> None:
+        print("test plugin")
```
joeribekker, line 30: Well... to make te config page do something, it should return `True` (since this one always works)

src/openforms/registrations/contrib/demo/plugin.py
```diff
@@ -41,3 +44,10 @@ def get_reference_from_result(self, result: None) -> NoReturn:
 
     def update_payment_status(self, submission: "Submission"):
         pass
+
+    def test_config():
+        test = []
+
+        test.append({'completed': False, 'error': 'geen fout melding', 'msg': 'Iets ging fout', 'name': 'name'})
```
joeribekker, line 51: So, about the return value for `test_config`.

It's probably nicer to return `True` if all works as expected and raise (a self defined) `ConfigException` using the error message:

```
if plugin is not configured:
  return False
if plugin test fails:
   raise ConfigException("failure message")
return True
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -85,3 +85,21 @@ def register_submission(
...
+
+    def test_config():
+        test = []
+        config = ObjectsAPIConfig.get_solo()
+        drc_client = config.drc_service.build_client()
+        orc_client = config.objects_service.build_client()
+
+        clients = [{'name': 'ORC (Overige)', 'type': 'object', 'client': orc_client},
+                   {'name': 'DRC (Informatieobjecten)', 'type': 'document', 'client': drc_client},
+                ]
```
joeribekker, line 97: You're duplicating the name that is also in the service model choices. No need to duplicate strings here.

src/openforms/registrations/contrib/objects_api/tests/test_plugin.py
```diff
@@ -0,0 +1,32 @@
...
+
+class OrcPluginTesterTest(TestCase):
+
+    def setUp(self):
+        # self.plugin = ZGWRegistration("zgw")
+        self.oac = ObjectsAPIConfig.get_solo()
+        drc_client = self.oac.drc_service.build_client()
+        orc_client = self.oac.objects_service.build_client()
+
+        self.lients = [{'type': 'object', 'client': orc_client},
```
joeribekker, line 16:
```suggestion
        self.clients = [{'type': 'object', 'client': orc_client},
```

Did you run these tests locally?

src/openforms/registrations/contrib/objects_api/tests/test_plugin.py
```diff
@@ -0,0 +1,32 @@
...
+
+        for i, client in enumerate(self.clients):
+            try:
+                client['client'].retrieve(client['type'], client['client'].base_url)
+                try:
+                    self.assertEqual(oar.test_config(), [])
+                except Exception as e:
+                    self.assertEqual(oar.test_config()[i]['error'], str(e))
+            except Exception as e:
+                pass
```
joeribekker, line 32: So, what are we testing here? When will this test fail?

src/openforms/templates/admin/plugin_tester.html
```diff
@@ -0,0 +1,94 @@
+{% extends 'admin/base_site.html' %}
```
joeribekker, line 1: This template could probably do without (or very little) custom css if you extend from the detail template.

Also, we don't import any external CSS if possible. The Django admin has a true/false icon built in.

joeribekker (changes requested):

src/openforms/registrations/admin.py
assignon: full path import implemented

src/openforms/registrations/admin.py
assignon: view added as TemplateView

sergei-maertens:
Closing this PR  to get peace of mind - it has merge conflicts anyway.

Issue #678, opened 2021-09-16
Validate backend options with the selected backend plugin serializer


Bartvaderkin:
Backends like the registration can define a serializer to decode their options from the Form object, but currently the validation on this runs only when pulling the configration data for use (like in the registration task) but it can fail the task because the data was not checked when saving the form.

@sergei-maertens  noted we probably could use https://github.com/maykinmedia/drf-polymorphic/ probably, so the API documentation is also updated with the expected schema depending on which plugin is selected

- [x] Backend /API validates options and returns DRF error response
- [x] Form Builder UI displays error: https://github.com/open-formulieren/open-forms/issues/678#issuecomment-944201087

joeribekker:
Labeled as bug because it's very hidden what goes wrong for the user.

joeribekker:
Registration backend nodig voor reproduceren

Bartvaderkin:
This happened on test with the console demo backend: https://sentry.maykinmedia.nl/maykin-media/open-forms/issues/291958/

joeribekker:
Keuze gemaakt voor eenvoudigere variant, zonder DRF maar met validators op basis van het "discrimator" field.

PR #817, opened 2021-10-15
[#678] Implemented validation of plugin options for registration and payment backends

Bartvaderkin:


Bartvaderkin:
This still needs the UI to pickup the validation error, see comment https://github.com/open-formulieren/open-forms/issues/678#issuecomment-944201087

PR #821, opened 2021-10-15
Improvements to confirmation email

### LINKED ISSUE #734: Email template add more forms
This ticket was rewritten into 2 tasks to address the actual problem.

**Tasks**

* [x] Move email template model/admin to the react Form designers as a tab
* [x] Add general email template to the general configuration (similar to confirmation text)

**Original report**

email templates --> edit --> click "+" sign behind "formulier" (NOg een formulier openen) opens a pop up to create a new form. 
One would expect the ability to add another existing form.

add_form_mail

shea-maykinmedia:
Fixes #734 

**Screenshots**

Screenshot 2021-11-12 at 11 51 16

Screenshot 2021-11-15 at 09 19 00

Screenshot 2021-11-17 at 13 50 51


src/openforms/config/models.py
```diff
@@ -37,6 +37,27 @@ class GlobalConfiguration(SingletonModel):
...
+    confirmation_email_content = models.TextField(
+        _("confirmation email content"),
+        help_text=_(
+            "Content of the confirmation email message. Can be overridden on the form level"
+        ),
+        default="""
+        Geachte heer/mevrouw,
+
+        Wij hebben uw inzending, met referentienummer {{ public_reference }}, in goede orde ontvangen.
+        """,
```
shea-maykinmedia, line 58: I'm not sure what we want for a default?

Also I couldn't seem to get the newlines working.  Not sure what the issue is but I'll look more into this once I know what the default will be and make sure it renders correctly.

src/openforms/forms/api/serializers.py
```diff
@@ -177,6 +185,58 @@ class Meta:
...
+                ConfirmationEmailTemplate.objects.create(
+                    form=instance, **confirmation_email_template
+                )
+        else:
+            try:
+                # If a complete email template is not given then delete the potential confirmation email template
+                #   This handles the case where a template was created but later cleared
+                instance.confirmation_email_template.delete()
+            except (AttributeError, ConfirmationEmailTemplate.DoesNotExist):
+                pass
```
shea-maykinmedia, line 236: I don't like this code, especially the `else`, but I'm not sure if there is a better way to do this given the one to one field and what is sent through the API.

The reason for this is that this seems to be what you need to do in the serializer when for creating one to one models.

The `else` is specifically so we can delete the confirmation email template through the api in the case that the user had previous entered a subject and content but then cleared them at a later point.

I do wonder if it would be better to get rid of the `ConfirmationEmailTemplate` and put the `subject` and `content` on the `Form`?  It's less nice from the model perspective but is nicer from an API perspective.

Either way I wanted to get some feedback on the best way to do this and thought highlighting this in a PR would be the best way to do it.  I am happy to do a Slack call to discuss this in depth. :)

src/openforms/submissions/tasks/emails.py
```diff
@@ -64,13 +63,6 @@ def send_confirmation_email(submission_id: int) -> None:
     if submission.confirmation_email_sent:
         return
 
-    try:
-        # access the reverse of the one2one
-        submission.form.confirmation_email_template
-    except (AttributeError, ObjectDoesNotExist):
-        logevent.confirmation_email_skip(submission)
-        return
```
shea-maykinmedia, line 72: This logic is now handled in `_send_confirmation_email`.  Just mentioning it since it might look a little weird that this code is just deleted.

sergei-maertens: Let's clean this up properly - `_send_confirmation_email` can raise a custom `SkipConfirmationEmail` exception, which the task here catches and then logs the appropriate message. I'd like to see the logging in a centralized place rather than smeared out over multiple functions, making it harder to follow what's logged where.

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -9,10 +9,12 @@
 
 from privates.test import temp_private_root
 
+from openforms.config.models import GlobalConfiguration
 from openforms.emails.models import ConfirmationEmailTemplate
 from openforms.emails.tests.factories import ConfirmationEmailTemplateFactory
+from openforms.utils.tests.html_assert import HTMLAssertMixin
 
-from ...utils.tests.html_assert import HTMLAssertMixin
+from ...forms.constants import ConfirmationEmailOptions
```
sergei-maertens, line 17:
```suggestion
from openforms.forms.constants import ConfirmationEmailOptions
```

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -277,11 +279,148 @@ def test_completed_submission_after_timeout_with_send_email(
...
+
+        with override_settings(CELERY_TASK_ALWAYS_EAGER=True):
+            # "execute" the celery task
+            send_confirmation_email(submission.id)
+
+        # Verify that email was sent
+        self.assertEqual(len(mail.outbox), 1)
+        message = mail.outbox[0]
+        self.assertEqual(message.subject, config.confirmation_email_subject)
+        self.assertEqual(message.body.strip(), config.confirmation_email_content)
```
sergei-maertens, line 318: Please compare to literal values, that makes tests easier to read and debug.

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -277,11 +279,148 @@ def test_completed_submission_after_timeout_with_send_email(
...
```
sergei-maertens, line 356: See above

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -277,11 +279,148 @@ def test_completed_submission_after_timeout_with_send_email(
...
+
+        with override_settings(CELERY_TASK_ALWAYS_EAGER=True):
+            # "execute" the celery task
+            send_confirmation_email(submission.id)
 
         # Verify that email was sent
         self.assertEqual(len(mail.outbox), 1)
+        message = mail.outbox[0]
+        self.assertEqual(message.subject, confirmation_email.subject)
+        self.assertEqual(message.body.strip(), confirmation_email.content)
```
sergei-maertens, line 390: See earlier remarks

src/openforms/submissions/utils.py
```diff
@@ -84,15 +86,50 @@ def send_confirmation_email(submission: Submission):
...
+    if (
+        submission.form.confirmation_email_option
+        == ConfirmationEmailOptions.form_specific_email
+    ):
+        try:
+            email_template = submission.form.confirmation_email_template
+            subject = email_template.subject
+            html_content = email_template.render(submission)
+            text_content = email_template.render(submission, {"rendering_text": True})
+        except (AttributeError, ObjectDoesNotExist):
```
sergei-maertens, line 98: All this should not be needed/possible if the validation is set up properly.

Please validate on the form serializer that `ConfirmationEmailOptions.form_specific_email` can only be picked if both subject/content of form-specific templates are present.

src/openforms/submissions/utils.py
```diff
@@ -84,15 +86,50 @@ def send_confirmation_email(submission: Submission):
...
+        == ConfirmationEmailOptions.global_email
+    ):
+        config = GlobalConfiguration.get_solo()
+        subject = Template(config.confirmation_email_subject).render(
+            Context({"form_name": submission.form.name})
+        )
+        html_content = config.render_confirmation_email_content(submission)
+        text_content = config.render_confirmation_email_content(
+            submission, {"rendering_text": True}
+        )
```
sergei-maertens, line 117: As mentioned earlier - please extract this into a new function to obtain the subject/body template to use.

sergei-maertens (changes requested):

src/openforms/emails/api/serializers.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework import serializers
+
+
+class ConfirmationEmailTemplateSerializer(serializers.Serializer):
```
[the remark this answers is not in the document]
shea-maykinmedia: The reason was so we can delete the `ConfirmationEmailTemplate` if a dict without a subject or body filled in doesn't throw a validation error.

For example if a `ConfirmationEmailTemplate` exists and the user wants to delete it they'll need to remove the subject and content which then leaves a dict like
```
{
   'subject': '',
   'content': ''
}
```
but this would thrown an error if this were a model serializer.

I guess a comment would be good to put there explaining this ;)

src/openforms/forms/api/serializers.py
```diff
@@ -227,6 +286,7 @@ def get_fields(self):
         # for export we want to use the list of plugin-id's instead of detailed info objects
         del fields["login_options"]
         del fields["payment_options"]
+        del fields["confirmation_email_template"]
```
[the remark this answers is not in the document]
shea-maykinmedia: Yes.  Since the ConfirmationEmailTemplate model existed but we weren't exporting it I figured we still wanted to not export it.

src/openforms/forms/api/serializers.py
```diff
@@ -181,6 +188,58 @@ class Meta:
...
+        confirmation_email_template = validated_data.pop(
+            "confirmation_email_template", None
+        )
+        instance = super().create(validated_data)
+
+        if (
+            confirmation_email_template
+            and confirmation_email_template.get("subject")
+            and confirmation_email_template.get("content")
+        ):
```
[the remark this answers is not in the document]
shea-maykinmedia: The reason for checking both is that the frontend (react admin) would send 
```
{
   'subject': '',
   'content': ''
}
```
if the user clears the fields related to the confirmation.  I will add the validation to the serializer to ensure either both are filled out or none but I wonder if we should still keep this conditional checking so it's clear, even though we could only check one or the other is present.

src/openforms/config/templates/emails/confirmation_email_content.html
```diff
@@ -0,0 +1,13 @@
+Geachte heer/mevrouw,<br>
```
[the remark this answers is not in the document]
shea-maykinmedia: It did seem to work with `{% blocktrans %}` or would this not be correct?

Screenshot 2021-11-17 at 11 04 06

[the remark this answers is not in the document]
sergei-maertens: we do not want this `{% blocktrans %}` to show in the admin, that's an internal template thing since we manage translations in the code, not in the content/database

src/openforms/emails/models.py
```diff
@@ -41,44 +57,11 @@ class ConfirmationEmailTemplate(models.Model):
...
     )
 
+    objects = ConfirmationEmailTemplateManager()
+
     class Meta:
         verbose_name = _("Confirmation email template")
         verbose_name_plural = _("Confirmation email templates")
 
     def __str__(self):
         return f"Confirmation email template - {self.form}"
```
sergei-maertens, line 67: Should be translatable string

src/openforms/emails/utils.py
```diff
@@ -173,3 +180,46 @@ def unwrap_anchors(html_str: str) -> str:
...
+    if submission.completed_on:
+        context["submission_date"] = submission.completed_on.date().strftime("%d-%m-%Y")
+
+    if submission.payment_required:
+        context["payment_required"] = True
+        context["payment_user_has_paid"] = submission.payment_user_has_paid
+        context["payment_url"] = build_absolute_uri(
+            reverse("payments:link", kwargs={"uuid": submission.uuid})
+        )
+        context["payment_price"] = str(submission.form.product.price)
```
sergei-maertens, line 206: See also #918 which should fix #917 and will result in merge conflict(s)

src/openforms/forms/api/serializers.py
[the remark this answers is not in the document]
sergei-maertens: imo this should also just call `ConfirmationEmailTemplate.objects.set_for_form`, it's supposed to be idempotent :)

src/openforms/forms/api/serializers.py
```diff
@@ -220,13 +257,60 @@ def validate_backend_options(self, attrs, backend_field, options_field, registry
...
+                            ErrorDetail(
+                                string=_("This field is required."), code="required"
+                            )
+                        ]
+                    }
+                )
+            for tag in [
+                "{% appointment_information %}",
+                "{% payment_information %}",
+            ]:
```
sergei-maertens, line 286: We have `openforms.emails.validators.DjangoTemplateValidator` already for this functionality.

src/openforms/submissions/utils.py
```diff
@@ -71,8 +79,46 @@ def remove_submission_uploads_from_session(
         remove_upload_from_session(attachment.temporary_file, session)
 
 
+def get_confirmation_email_components(submission: Submission):
+    if (
+        submission.form.confirmation_email_option
+        == ConfirmationEmailOptions.form_specific_email
+    ):
+        email_template = submission.form.confirmation_email_template
+        subject = email_template.subject
```
sergei-maertens, line 88: isn't this supposed to be a template too? Global config is a template, but form override then isn't? That violates the end-user principle of least surprise.

This entire function should just do the template resolving, and pass subject, html and text templates to render in the other function.

src/openforms/submissions/utils.py
```diff
@@ -71,8 +79,46 @@ def remove_submission_uploads_from_session(
...
+            submission, config.confirmation_email_content, {"rendering_text": True}
+        )
+        return subject, html_content, text_content
+
+
 def send_confirmation_email(submission: Submission):
-    email_template = submission.form.confirmation_email_template
+    logevent.confirmation_email_start(submission)
+
+    if submission.form.confirmation_email_option == ConfirmationEmailOptions.no_email:
```
sergei-maertens, line 113: it's weird that this check is smeared out across two functions - my initial feedback was to raise an exception if e-mail sending needs to be skipped for whatever reason, catch that exception and then abort the actual sending + log the skip.

That way you can keep things contained/scoped and easy to follow.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #854, opened 2021-10-25
[#678] Admin frontend for plugin validation

### LINKED ISSUE #678: Validate backend options with the selected backend plugin serializer

SilviaAmAm:
Fixes #678

### DIFF
--- a/src/openforms/js/components/admin/form_design/RegistrationFields.js
+++ b/src/openforms/js/components/admin/form_design/RegistrationFields.js
@@ -8,6 +8,38 @@ import Fieldset from '../forms/Fieldset';
 import Select from '../forms/Select';
 import Form from '@rjsf/core';
 
+const FormRjsfWrapper = ({ name, label, schema, formData, onChange, errors }) => {
+    let extraErrors = {};
+    for (let [key, msg] of errors) {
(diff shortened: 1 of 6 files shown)


sergei-maertens (approved):
Refactored so the code does not rely on magic numbers/indices and the components are more usable in a generic fashion.

I've also reverted the `error` class removal - this is something the django admin does and the React components should match the behaviour of the Django admin templates. Instead, I've modified the RJSF default styles to do the error-border highlight correctly.

Additionally, it's possible there are multiple validation errors for the same input, which show up as different array items from the backend and need to be merged back together in the `extraErrors`. This is now handled correctly in the wrapper.

PR #859, opened 2021-10-27
[#824] Update fieldset edit form

SilviaAmAm:
Fixes part #824 

The problems I saw while fixing #824 are:
- Field sets have both a 'legend' and a 'label' field. The label should be automatically populated to be equal to the legend, but this doesn't seem to work in Formio 4.13.x. Therefore, all fieldsets have the label 'Field Set'.
- The 'key' of a fieldset has to be updated manually in the API tab (instead of being automatically generated from the label like in other components). If one doesn't do this, the all the fieldsets have the same key 'fieldSet'

So the changes are:
- I changed the display tab so that it has a 'label' field and a 'key' field like for the other components. The legend field is automatically changed to equal the label (since the legend is the field that is actually used to make the 'title' of the component that is then shown in the form).
- I removed the API tab, since the 'key' attribute is now in the display tab.

### DIFF
--- a/src/openforms/js/components/form/edit/tabs.js
+++ b/src/openforms/js/components/form/edit/tabs.js
@@ -85,30 +85,6 @@ const TEXT_BASIC = {
 };
 
 
-const FIELDSET_BASIC = {
-    key: 'display',
-    label: 'Display',
-    components: [
-        {
-            type: 'textfield',
-            key: 'label',
-            label: 'Label'
-        },
-        {
-            type: 'textfield',
-            key: 'key',
-            label: 'Property Name'
-        },
-        {
-            type: 'checkbox',
-            key: 'hidden',
-            label: 'Hidden',
-            tooltip: 'A hidden field is still a part of the form, but is hidden from view.'
-        }
-    ]
-};
-
-
 const CHOICES_BASIC = {
     key: 'basic',
     label: 'Basic',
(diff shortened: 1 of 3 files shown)


src/openforms/js/components/form/edit/tabs.js
```diff
@@ -84,6 +84,31 @@ const TEXT_BASIC = {
     ]
 };
 
+
+const FIELDSET_BASIC = {
```
sergei-maertens, line 88: please move this to the fieldset component module, it's not used in any other components.

src/openforms/js/components/form/fieldset.js
```diff
@@ -0,0 +1,33 @@
...
+        super(component, options, data);
+    }
+
+    static editForm(...extend) {
+        const parentEditForm = FormioFieldSet.editForm();
+        parentEditForm.components[0].components = [
+            FIELDSET_BASIC,
+            // The 'API' tab is removed, since the only useful attribute it contained was the 'key', but we
+            // have this field in the FIELDSET_BASIC tab.
+            // TODO: can these tabs below be removed?
```
sergei-maertens, line 24: let's do it, seems too complex and we'll put it back if people request it

src/openforms/js/components/form/fieldset.js
```diff
@@ -0,0 +1,33 @@
...
+
+class FieldSet extends FormioFieldSet {
+    constructor(component, options, data) {
+        /* Field sets have both a 'legend' and a 'label' field. The label should be automatically populated to be equal
+        to the legend, but this doesn't seem to work in Formio 4.13.x. Since in open-forms we always use the label, in
+        this custom fieldset the legend is hidden and automatically filled with the value of the label.
+         */
+        if (component.label) {
+            component.legend = component.label;
+        }
```
sergei-maertens, line 14: Does this code run when you're changing the label? It seems odd to put this in the constructor

sergei-maertens (changes requested):

src/openforms/js/components/form/fieldset.js
SilviaAmAm: Yes, when the label property is changed in the edit form, the constructor is called and the legend gets updated. I agree though that it feels like a hack

sergei-maertens (approved):

PR #870, opened 2021-10-29
[#790] Added support to save map-component coordinates in registration backends

Bartvaderkin:


src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -95,6 +97,13 @@ def _safe_int(num):
             )
 
 
+def _point_coordinate(value):
+    if not value or not isinstance(value, list) or len(value) != 2:
+        return SKIP
+    # cast to str because floats would get localized in the template
```
sergei-maertens, line 103: shouldn't we just disable localization in the template then? https://docs.djangoproject.com/en/2.2/topics/i18n/formatting/#localize

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -110,6 +119,9 @@ class StufZDSRegistration(BasePlugin):
         # "initiator.aanschrijfwijze": FieldConf(RegistrationAttribute.initiator_aanschrijfwijze),
         "initiator.bsn": FieldConf(submission_field="bsn"),
         "initiator.kvk": FieldConf(submission_field="kvk"),
+        "lokatie": FieldConf(
```
sergei-maertens, line 122: is it called `lokatie`? Shouldn't that be `locatie`?

sergei-maertens (changes requested):

src/openforms/registrations/contrib/stuf_zds/plugin.py
Bartvaderkin: I forgot about the `unlocalize` filter.

src/openforms/registrations/contrib/stuf_zds/plugin.py
Bartvaderkin: It is `lokatie` in Stuf but I updated it anyway.

sergei-maertens (approved):

PR #873, opened 2021-11-02
Finalize showing backend validation errors from the backend

### LINKED ISSUE #720: Errors in the Admin form editor are unclear

sergei-maertens:
Fixes #720

* Added the missing tabs
* Refactored some components

### DIFF
--- a/src/openforms/js/components/admin/form_design/form-creation-form.js
+++ b/src/openforms/js/components/admin/form_design/form-creation-form.js
@@ -130,9 +130,9 @@ const FORM_FIELDS_TO_TAB_NAMES = {
     canSubmit: 'form',
     registrationBackend: 'registration',
     registrationBackendOptions: 'registration',
-    product: 'product',
-    paymentBackend: 'payment',
-    paymentBackendOptions: 'payment',
+    product: 'product-payment',
+    paymentBackend: 'product-payment',
+    paymentBackendOptions: 'product-payment',
     submissionsRemovalOptions: 'submission-removal-options',
     literals: 'literals',
 };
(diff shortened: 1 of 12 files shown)


src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -156,29 +158,32 @@ function reducer(draft, action) {
...
+
+            // remove any validation errors
+            draft.validationErrors = draft.validationErrors.filter(([key]) => key !== name);
+
+            // check which tabs still need the marker and which don't
+            const errorsPerTab = groupBy(draft.validationErrors, ([key]) => {
+                const [prefix, fieldPrefix] = key.split('.');
+                return FORM_FIELDS_TO_TAB_NAMES[fieldPrefix];
+            });
+            draft.tabsWithErrors = draft.tabsWithErrors.filter(tabId => tabId in errorsPerTab);
```
SilviaAmAm, line 186: This seems to work for the first tab, but not for the registratie tab: I enter an invalid email, submit, update the invalid field, but the error seems to keep showing.

SilviaAmAm (changes requested):

src/openforms/js/components/admin/form_design/form-creation-form.js
sergei-maertens: time to fix that :-) I tested with the form meta & data cleanup config tabs, so might have missed something :+1:

PR #882, opened 2021-11-04
[#217] Enable multiple values for input fields

SilviaAmAm:
Blocked by 
* https://github.com/open-formulieren/open-forms-sdk/issues/8
* https://github.com/open-formulieren/open-forms-sdk/issues/103

---

Part of #217

### DIFF
--- a/src/openforms/js/components/form/edit/tabs.js
+++ b/src/openforms/js/components/form/edit/tabs.js
@@ -33,6 +33,12 @@ const BASIC = {
         },
         {
             type: 'checkbox',
+            key: 'multiple',
+            label: 'Multiple values',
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):
I'd like to see the styling sorted out in the SDK before I merge this, otherwise we could end up in a situation that people start using this and then file bugs that it looks horrendous.

SilviaAmAm:
https://github.com/open-formulieren/open-forms-sdk/pull/102 Style PR

While I was working on the style, I noticed that the the multiple-value field doesn't seem to work properly in the SDK. I will make another issue about this, but we should probably wait to merge this until that is fixed: https://github.com/open-formulieren/open-forms-sdk/issues/103

PR #884, opened 2021-11-05
Fixes #863 -- fix date-picker z-index

### LINKED ISSUE #863: Form editor - date default value z-index is too low
Similar issue to #731 

When picking a default value for a date compoonent, the calendar popup is hidden behind the component edit modal.

image

sergei-maertens:
This was hidden because of higher z-indices from django apps.

### DIFF
--- a/src/openforms/scss/components/builder/_builder.scss
+++ b/src/openforms/scss/components/builder/_builder.scss
@@ -40,7 +40,8 @@ li.nav-item {
 }
 
 .formio-dialog,
-.ck-body .ck.ck-balloon-panel
+.ck-body .ck.ck-balloon-panel,
+div.flatpickr-calendar.open
 {
   z-index: 10000000;
 }


src/openforms/scss/components/builder/_builder.scss
```diff
@@ -40,7 +40,8 @@ li.nav-item {
 }
 
 .formio-dialog,
-.ck-body .ck.ck-balloon-panel
+.ck-body .ck.ck-balloon-panel,
```
SilviaAmAm, line 43: Seems to work nicely on chrome, but on firefox I don't see the date picker :S

SilviaAmAm (changes requested):

src/openforms/scss/components/builder/_builder.scss
sergei-maertens: mmm might be because I used the same value of another element, might have to increase that a bit. ugh, browsers...

PR #889, opened 2021-11-08
Various small fixes

### LINKED ISSUE #867: Issues noticed during sprint demo 8
- [x] Form designer (admin) -> "Search field(s)" is not translated yet
- [x] Unintended confusing characters end up in generated internal reference (1, l or I) -> remove that
- [x] Race condition with payments - browser return + webhook arrive at the same time and schedule sending the confirmation e-mail.
- #695
- [x] Missing (audit) log of update payment status with registration backend.

sergei-maertens:
Fixes #867 

* Override template to add missing translation
* Removed 1 and I from character set for reference generation
* Fixed race condition, now multiple confirmation e-mails should not be sent anymore
* Added extra audit log entries on payment update with the registration backend

### DIFF
--- a/src/openforms/submissions/tasks/emails.py
+++ b/src/openforms/submissions/tasks/emails.py
@@ -1,5 +1,8 @@
+import logging
+
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import DatabaseError, transaction
 
 from celery_once import QueueOnce
 
@@ -10,6 +13,8 @@ from ..models import Submission
 from ..utils import send_confirmation_email as _send_confirmation_email
 
 __all__ = ["maybe_send_confirmation_email"]
+
+logger = logging.getLogger(__name__)
 
 
 @app.task(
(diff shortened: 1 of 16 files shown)


src/openforms/submissions/tasks/emails.py
```diff
@@ -60,7 +66,18 @@ def send_confirmation_email(submission_id: int) -> None:
...
     the confirmation e-mail would be sent out twice, which is an acceptable risk.
     """
-    submission = Submission.objects.get(id=submission_id)
+    try:
+        submission = Submission.objects.select_for_update(nowait=True).get(
+            id=submission_id
+        )
+    except DatabaseError:
+        logger.debug(
+            "Submission %d confirmation e-mail task failed to acquire a lock. This is "
```
SilviaAmAm, line 75: Since this prevents the race condition, should the comment at the beginning of the function be changed?

src/openforms/submissions/tasks/registration.py
```diff
@@ -94,13 +94,13 @@ def get_reference_for_submission(submission: Submission) -> str:
 
 
 def get_random_reference() -> str:
-    # 36 characters with length 6 -> 36^6 possible combinations.
-    # that's roughly 1.5 billion combinations before we run out of options.
+    # 36 characters with length 6 -> 32^6 possible combinations.
```
SilviaAmAm, line 97:
```suggestion
    # 32 characters with length 6 -> 32^6 possible combinations.
```

SilviaAmAm (changes requested):

PR #899, opened 2021-11-11
[#896] Replace StringUUIDField with models.UUIDField

### LINKED ISSUE #896: Remove StringUUIDField
It's causing warnings and does not actually do anything useful that the default `models.UUIDField` lacks.

Warning, reported from Bart:

```
RemovedInDjango30Warning: Remove the context parameter from StringUUIDField.from_db_value(). 
```

Bartvaderkin:
Fixes #896


src/openforms/utils/fields.py
```diff
@@ -4,6 +4,12 @@
 
 
 class StringUUIDField(models.UUIDField):
+    """
+    DEPRECATED
+
+    remains here for references in historical migrations
+    """
+
```
sergei-maertens, line 12: https://docs.djangoproject.com/en/2.2/topics/migrations/#considerations-when-removing-model-fields this is a better option to actually remove the field :)

sergei-maertens (changes requested):

src/openforms/utils/fields.py
Bartvaderkin: Indeed, I'll replace the comment.

sergei-maertens (approved):

PR #907, opened 2021-11-12
Remove visual clutter in test runs

### LINKED ISSUE #654: Cleanup (test suite) console output
There is a lot of unnecessary log output that makes it difficult to track down actual problems.

Main use-case is the test suite (on Github actions) but we should also evaluate the actual logs.

- [x] DRF validation/permission errors
- [x] SOAP errors
- [x] ??

sergei-maertens:
Fixes #654

- [x] Disable `loaddata` output from post-migrate handler
- [x] Fix sudden reported deprecation warnings because of O365 lib (see https://github.com/O365/python-o365/issues/713)
- [x] Disable output from `createinitialsuperuser` management command in tests
- [x] Disable StUF output in tests
- [x] Disable exception handler logging in CI env


src/openforms/conf/ci.py
```diff
@@ -18,7 +18,17 @@
     "oidc": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
 }
 
-LOGGING = None  # shut up logging
+# shut up logging
+LOGGING["loggers"].update(
```
Bartvaderkin, line 22: I hope don't regret this when debugging build failures but we'll find out.

Bartvaderkin (approved):

src/openforms/conf/ci.py
sergei-maertens: it shouldn't and we can enable `DEBUG=1` envvar to have the exception re-raised :)

PR #930, opened 2021-11-15
[#926] Block form listing in HTML and API if user not staff

### LINKED ISSUE #926: Show forbidden message/empty page on the form list page when redirect is not configured and the user is not logged in
Currently, you see a list of forms when redirect is not configured or you are logged in to the admin.

This needs to change to ONLY show the list of forms when logged in as admin. Otherwise, an error/empty page or whatever is easiest to NOT show a list of forms. The API needs to block this too ofcourse.

Bartvaderkin:
Fixes #926

### DIFF
--- a/src/openforms/forms/api/viewsets.py
+++ b/src/openforms/forms/api/viewsets.py
@@ -12,6 +12,7 @@ from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema
 from rest_framework import parsers, permissions, response, status, views, viewsets
 from rest_framework.authentication import TokenAuthentication
 from rest_framework.decorators import action
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.mixins import ListModelMixin
 from rest_framework.request import Request
 from rest_framework.response import Response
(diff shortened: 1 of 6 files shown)


src/openforms/forms/api/viewsets.py
```diff
@@ -211,7 +212,7 @@ class FormViewSet(viewsets.ModelViewSet):
     lookup_url_kwarg = "uuid_or_slug"
     # lookup_value_regex = "[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}"
     serializer_class = FormSerializer
-    permission_classes = [IsStaffOrReadOnly]
+    permission_classes = [IsStaffOrReadOnlyNoList]
```
sergei-maertens, line 215: Be advised that this may break potential integration from CMS solutions which relied on this endpoint being public.

We will have to check/set-up token auth for this and document it.

sergei-maertens (approved):

PR #954, opened 2021-11-18
Add change appointment functionality

### LINKED ISSUE #680: De gebruiker kan een afspraak wijzigen via een url in een bevestigings e-mail.
Eis: Afspraken 3 (deel 2: wijzigen, deel 1 in #263 ).
Zie ook NvI: 60, 92

**Tasks**

* [x] ~Add API endpoint to change the appointment.~
* [x] To change an appointment, we can use the link (open forms, form url) that was used to create the appointment in the first place. So, point the user to the "appointment form"
* [x] The backend should make a copy of the original submission (data) into a new submission, keep a FK reference to the original, and load the new submission in the form
* [x] Improvement over the previous bullet: Allow the pre-selection of a product using a query param: `/form/my-form?product=XYZ` and use this in the change appointment link (see below).
* [x] Keep track that this submission replaces a previous submission (maybe add `Submission.previous_submission`)
* [x] The appoinment plugin behaviour should be adjusted so that when a previous submission is present, the previous submission appointment is cancelled AFTER the new appointment is created.
* [x] For now, we leave out the cancellation/deletion of the previous submission in a registration backend.
* [x] The confirmation email should indicate this concerns a change. Add " (wijziging)" to the end of the subject.
* [x] Add change link to the confirmation email: 
  **Afspraak wijzigen of annuleren**
  o [Ik wil mijn afspraak wijzigen]()
  o [Ik wil mijn afspraak annuleren]()

**Context**

The idea is that we will not implement any real "change" behaviour and call some appointment provider change API endpoint. Instead, we will simply combine "make new appointment + cancel old appoinment" and make the user feel he/she is changing the appointment. 

This choice is made because too much can be changed with numerous side effects that are hard to predict. What if you change the product and a different form fields should be present? Even if the cancel appointment API call fails for some reason, we can still make the new appointment. If a change appointment would fail, it's unclear what the status is exactly.

shea-maykinmedia:
Fixes #680 

This was a little hard to test since we don't have an external working client for creating appointments so I just mocked the create and delete appointment calls as well as the call to get the appointment details.

Also a change to the SDK isn't needed for this since the form save allows starting the form at a specific step and the SDK already preloads any data in the submission.

**Screenshots**
Screenshot 2021-11-19 at 09 47 15
Screenshot 2021-11-19 at 09 47 05
Screenshot 2021-11-19 at 09 45 15
Screenshot 2021-11-19 at 09 44 58


src/openforms/utils/email.py
```diff
@@ -25,7 +25,11 @@ def send_mail_plus(
         fail_silently=fail_silently,
     )
     mail = EmailMultiAlternatives(
-        subject, message, from_email, recipient_list, connection=connection
+        subject.replace("\n", ""),
```
shea-maykinmedia, line 28: There is an issue for this https://github.com/open-formulieren/open-forms/issues/964 but it's just something I did to get the confirmation emails working.  I think we can probably remove this.

src/openforms/appointments/utils.py
```diff
@@ -146,6 +146,20 @@ def book_appointment_for_submission(submission: Submission) -> None:
...
+                    submission.previous_submission.appointment_info.appointment_id
+                )
+                submission.previous_submission.appointment_info.status = (
+                    AppointmentDetailsStatus.cancelled
+                )
+                submission.previous_submission.appointment_info.save(
+                    update_fields=["status"]
+                )
+            except (AppointmentInfo.DoesNotExist, AppointmentDeleteFailed):
+                pass
```
sergei-maertens, line 162: it seems weird that we silently fail if appointment delete fails? This should at least be logged

src/openforms/appointments/views.py
```diff
@@ -41,3 +47,30 @@ def get_redirect_url(self, submission_uuid: uuid, token: str, *args, **kwargs):
             }
         )
         return f.url
+
+
+class VerifyChangeAppointmentLinkView(RedirectView):
+    def get_redirect_url(self, submission_uuid: uuid, token: str, *args, **kwargs):
+        submission = validate_url_and_get_submission(submission_uuid, token)
+
+        new_submission = submission.copy()
```
sergei-maertens, line 56: this sould probably be a deep copy, since the dicts in JSONFields are still the same instances and if we were to mutate these and (accidentally) save the original submission, this would now incorrectly be updated too.

src/openforms/appointments/views.py
```diff
@@ -41,3 +47,30 @@ def get_redirect_url(self, submission_uuid: uuid, token: str, *args, **kwargs):
...
+
+        add_submmission_to_session(new_submission, self.request.session)
+
+        step_url = new_submission.get_appointment_step_url()
```
sergei-maertens, line 60: badly named variable, should really be `step_slug` and not URL

src/openforms/submissions/models.py
```diff
@@ -236,6 +236,10 @@ class Submission(models.Model):
         ),
     )
 
+    previous_submission = models.ForeignKey(
+        "submissions.Submission", on_delete=models.DO_NOTHING, null=True, blank=True
```
sergei-maertens, line 240: you really should never use `models.DO_NOTHING` because then your database will start tripping up. Since the field is nullable, `models.SET_NULL` is more appropriate.

Effectively this means that when a submission is deleted (which happens after x number of days!), any change-appointment submissions now have lost their reference. I think this is acceptable.

src/openforms/submissions/models.py
```diff
@@ -528,6 +532,29 @@ def payment_registered(self) -> bool:
         # TODO support partial payments
         return self.payments.filter(status=PaymentStatus.registered).exists()
 
+    def get_appointment_step_url(self) -> str:
+        for form_step in self.form.formstep_set.all():
+            for component in form_step.iter_components(recursive=True):
+                if component.get("appointments", {}).get("showProducts"):
+                    return form_step.form_definition.slug
+
+        return ""
```
sergei-maertens, line 541: This is appointment specific logic/functionality that belongs in the `appointments` app. Should be a utility/service function from that app that's called where you need it. Do not overload different apps with functionality from other apps and keep them as decoupled as possible.

src/openforms/submissions/models.py
```diff
@@ -528,6 +532,29 @@ def payment_registered(self) -> bool:
...
+            form_url=self.form_url,
+            previous_submission=self,
+        )
+        for submission_step in self.submissionstep_set.all():
+            SubmissionStep.objects.create(
+                submission=copy,
+                form_step=submission_step.form_step,
+                data=submission_step.data,
+            )
+        return copy
```
sergei-maertens, line 556: The canonical Django way is to grab the object, set `object.pk = None` and save it again. Then you're sure all the required fields are copied along.

In this case for example, we lose the authentication information and possible other meta-information.

Not sure how to best approach this :thinking: 

This would probably also be better suited as a model mananger function: `Submission.objects.copy(old_submission)`

sergei-maertens (changes requested):

PR #958, opened 2021-11-18
Update the OAS and check if OAS is up to date

joeribekker:
This CI change causes PRs to fail if the OAS is not updated as part of the PR.

Before, we updated the OAS via the CI but this failed (see 2nd commit in this PR...). It's also more clear what changed in the API if the updated OAS is part of a PR.

I still kept the OAS generation inside the test step, and perform the comparison after the tests to prevent creating a whole new app installation just for that.


sergei-maertens (approved):
LGTM

PR #960, opened 2021-11-18
[#935] Improve FormStep.__str__()

### LINKED ISSUE #935: Add proper name to form steps
If you delete a form definition, you currently see a very cryptic name shown to the user.

If we can make this name more descriptive it would help alot: 

```
<form name> stap <index+1>: <form definiton name>
```
 
Example what we have now:
Melding formdef verwijderen

Bartvaderkin:
Fixes #935

### DIFF
--- a/src/openforms/forms/models/form_step.py
+++ b/src/openforms/forms/models/form_step.py
@@ -70,7 +70,14 @@ class FormStep(OrderedModel):
         verbose_name_plural = _("form steps")
 
     def __str__(self):
-        return _("Form step {order}").format(order=self.order)
+        if self.form_id and self.form_definition_id:
+            return _("{form_name} step {order}: {definition_name}").format(
+                form_name=self.form.admin_name,
+                order=self.order,
+                definition_name=self.form_definition.admin_name,
+            )
+        else:
+            return super().__str__()
 
     def iter_components(self, recursive=True):
         yield from self.form_definition.iter_components(recursive=recursive)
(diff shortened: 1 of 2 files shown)


src/openforms/forms/models/form_step.py
```diff
@@ -70,7 +70,7 @@ class Meta:
         verbose_name_plural = _("form steps")
 
     def __str__(self):
-        return _("Form step {order}").format(order=self.order)
+        return f"{self.form.admin_name} {_('step')} {self.order}: {self.form_definition.admin_name}"
```
sergei-maertens, line 73: - Do not use f-strings for `__str__` implementations, the entire template should be marked translatable
- When interpolating related objects, always consider the case that the relation is not set/saved yet. This crashes now for `str(FormStep())`

src/openforms/forms/tests/test_models.py
```diff
@@ -262,3 +262,13 @@ def test_form_definition_sensitive_information_returns_correct_fields(self):
             self.form_definition_with_sensitive_information.sensitive_fields,
             ["textFieldSensitive"],
         )
+
+
+class FormStepTestCase(TestCase):
```
sergei-maertens, line 267: Please add tests for `FormStep` instances without `form` or `form_definition` set.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #962, opened 2021-11-18
[#892] Email validation on blur

### LINKED ISSUE #892: Invalid email warning too strict
Raised by municipality Duiven - Taiga issue 79

When filling in the E-mail field you instantly get the warning "invalid_email", should give me the chance to enter a valid email instead of showing the warning instantly after typing the first letter of my email address. 
The whole field also moves when entering the field, this is annoying.

Also translate "invalid email"

givemeasec

See also #876

SilviaAmAm:
Fixes #892

### DIFF
--- a/src/openforms/js/components/form/email.js
+++ b/src/openforms/js/components/form/email.js
@@ -1,8 +1,26 @@
 import {Formio} from "formiojs";
 import DEFAULT_TABS, {ADVANCED, SENSITIVE_BASIC, REGISTRATION, VALIDATION} from "./edit/tabs";
 
+const FormioEmail = Formio.Components.components.email;
 
-class EmailField extends Formio.Components.components.email {
+class EmailField extends FormioEmail {
+    static schema(...extend) {
+        return FormioEmail.schema({
+            validateOn: 'blur'
+        }, ...extend);
+    }
+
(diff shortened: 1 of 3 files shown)


src/openforms/js/components/form/email.js
```diff
@@ -1,8 +1,26 @@
...
+    static get builderInfo() {
+        return {
+          title: 'Email',
+          group: 'advanced',
+          icon: 'at',
+          documentation: '/userguide/#email',
+          weight: 10,
+          schema: EmailField.schema()
+        };
+    }
```
sergei-maertens, line 22: Do we need this override? And if we do, is the title still translatable?

sergei-maertens (changes requested):

src/openforms/js/components/form/email.js
SilviaAmAm: It seems that without it, the `validateOn` remains in its default value of `change`. 
It should remain translatable, because in the template the string gets marked as translatable: https://github.com/formio/formio.js/blob/master/src/templates/bootstrap/builderSidebarGroup/form.ejs#L38

sergei-maertens (approved):

PR #973, opened 2021-11-19
[#891] Fix logic selectboxes

### LINKED ISSUE #891: Form logic not working for selectBoxes
Raised by municipality LV - Taiga issue 132

This does not work:

logic_fail_again

When using native logic in the "Bestandupload field", it does work. But our customer wants to use it cross form definitions (hence form logic), which is not possible with the native logic. 

Please fix. 

logic_notfail

SilviaAmAm:
Fixes #891 

Todo:

- [x] Think about cases with other operators
- [x] Write tests

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/Trigger.js
+++ b/src/openforms/js/components/admin/form_design/logic/Trigger.js
@@ -90,9 +90,16 @@ const parseJsonLogic = (logic) => {
     }
 
     // check if we're using a literal value, or a component reference
-    const compareValue = values[1];
+    let compareValue = values[1];
     let operandType = '';
     let operand = '';
+
+    // Selectboxes case: the component name contains the reference to which value, e.g. "selectComponentField.option1"
+    const splitComponent = component.split('.');
+    if (splitComponent.length > 1) {
+        component = splitComponent[0];
+        compareValue = splitComponent[1];
+    }
 
     if (jsonLogic.is_logic(compareValue)) {
         const op = jsonLogic.get_operator(compareValue);
@@ -261,8 +268,20 @@ const Trigger = ({ name, logic, onChange }) => {
         }
     }
 
-    const firstOperand = componentType === 'date' ?
-        {date: {var: triggerComponent}} : {var: triggerComponent};
+    let firstOperand;
+    switch (componentType) {
+        case 'date': {
+            firstOperand = {date: {var: triggerComponent}};
+            break;
+        }
+        case 'selectboxes': {
+            firstOperand = {var: `${triggerComponent}.${compareValue}`}
+            compareValue = true;
+            break;
+        }
+        default:
+            firstOperand = {var: triggerComponent};
+    }
 
     const jsonLogicFromState = {
         [operator]: [
--- a/src/openforms/js/components/admin/form_design/logic/Trigger.js
+++ b/src/openforms/js/components/admin/form_design/logic/Trigger.js
@@ -69,7 +69,7 @@ const TRIGGER_FIELD_ORDER = [
 ];
 
 
-const parseJsonLogic = (logic) => {
+const parseJsonLogic = (logic, allComponents) => {
     // Algorithm mostly taken from https://github.com/jwadhams/json-logic-js/blob/master/logic.js, combined
     // with our own organization.
     if (!logic || !Object.keys(logic).length) return {};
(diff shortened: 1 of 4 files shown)


src/openforms/js/components/admin/form_design/logic/Trigger.js
```diff
@@ -90,10 +90,17 @@ const parseJsonLogic = (logic) => {
...
+    let compareValue = values[1];
     let operandType = '';
     let operand = '';
 
+    // Selectboxes case: the component name contains the reference to which value, e.g. "selectComponentField.option1"
+    const splitComponent = component.split('.');
+    if (splitComponent.length > 1) {
+        component = splitComponent[0];
+        compareValue = splitComponent[1];
+    }
```
sergei-maertens, line 102: probably a bit too naive, since you can set up the `key` to be something like `foo.bar` which results in the following submission data:

```json
{"foo": {"bar": "<value>"}}
```

(that's how Formio tranforms it, using `lodash/set` under the hood)

A more robust approach will probably be:

```js
[component, ...compareValueBits] = component.split('.')
```

& then when building the `firstOperand`, you could just `compareValueBits.join('.')` again

src/openforms/submissions/tests/form_logic/test_modify_components.py
```diff
@@ -958,6 +958,54 @@ def test_dirty_data_only_diff_data_returned(self):
...
+                ]
+            },
+        )
+        FormLogicFactory.create(
+            form=form,
+            json_logic_trigger={
+                "==": [
+                    {"var": "currentPets.cat"},
+                    True,
+                ]
```
sergei-maertens, line 985: I don't think we can actually specify boolean values at the moment in the logic builder, they get interpreted as a string. Not sure if json-logic package handles this conversion?

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/logic/Trigger.js
SilviaAmAm: While updating this, I noticed that my change in the backed also breaks in case of components with dots in the key names. So I made changes there too. I changed a bit approach in the front end

src/openforms/submissions/tests/form_logic/test_modify_components.py
SilviaAmAm: Because of this: https://github.com/open-formulieren/open-forms/blob/fix/891-logic-selectboxes/src/openforms/js/components/admin/form_design/logic/Trigger.js#L276 the value is automatically set to a bool in the admin. It's a bit hackish maybe :sweat_smile:

sergei-maertens: Hrm, so that makes it not possible to attach logic if the value is `false` instead of `true`?

src/openforms/js/components/admin/form_design/logic/Trigger.js
```diff
@@ -261,8 +267,20 @@ const Trigger = ({ name, logic, onChange }) => {
...
+    switch (componentType) {
+        case 'date': {
+            firstOperand = {date: {var: triggerComponent}};
+            break;
+        }
+        case 'selectboxes': {
+            firstOperand = {var: `${triggerComponent}.${compareValue}`}
+            compareValue = true;
+            break;
+        }
```
sergei-maertens, line 280: Either we should:

- Parse the strings `"false"` and `"true"` from text-input as booleans `false` and `true` or
- Offer a dropdown for type selectboxes component where they can select `True` or `False` which maps to the booleans under the hood (preferred approach)

With the select boxes there's only two states - checked or not checked and you can attach logic to both. By always setting it to checked, you're taking away the option to do something if it's _not_ checked.

sergei-maertens (changes requested):

src/openforms/submissions/tests/form_logic/test_modify_components.py
SilviaAmAm: I thought that in that case you can use the `!=` operator

sergei-maertens (approved):

src/openforms/js/components/admin/form_design/logic/Trigger.js
SilviaAmAm: I thought that since the logic boxes only have 2 states, you can achieve the logic for when the checkbox is unselected by using the `!=` operator

Issue #974, opened 2021-11-19
Demo sprint 9 checklist


sergei-maertens:
**High prio**

- [x] Fix e-mail layout
    - [x] logo is scaled too big
    - [x] header bg color should be white (tweak design tokens)
- [x] Check failing submission: https://open-forms.test.maykin.opengem.nl/admin/submissions/submission/346/change/ (@sergei-maertens)
- [x] Investigate unexpected HTTP 403's while filling out the form - happens even when going from step to next step (@shea-maykinmedia)
- [x] SharePoint credentials @joeribekker 
- [x] Verify PDF output (confirmation) @SilviaAmAm 
- [x] update SDK translations (@sergei-maertens)
- [x] Submit/Next button hover-state is grey (should be dark blue, like the cookie hover buttons) (@sergei-maertens)
- [x] Opening payment URL from e-mail is broken (see Sentry) @sergei-maertens 

**Low prio**

- [x] Check that the default mail templates in global config are properly NL-version instead of EN (for new instances) @sergei-maertens 
- [ ] On FormIO form invalid -> logic checking spinner button keeps spinning
- [x] Component multiple values (array value) overview is not properly formatted as list (comma separated?) @SilviaAmAm 
- [ ] Logic: checkbox true conditions not working yet (#810) @SilviaAmAm 
- [ ] fieldsets not in summary: not just the fieldset, but the nested components aren't either! (#928)
- [x] Missing admin frontend translation `Receives confirmation email` @SilviaAmAm 
- [ ] The PDF report generation breaks with a multivalue field if that field is left empty (tested with email field). 

~~TinyMCE hover -> change cursor in "cursor" rather than default cursor (same as hovering a text input)~~ scrapped, iframe mess

sergei-maertens:
Not done

sergei-maertens:
mmm no, not closed.

PR #978, opened 2021-11-22
Hotfix/demo fixes

sergei-maertens:
Relates to #974


Bartvaderkin (approved):
We should add a ticket to cleanup `get_printable_data()` and related methods after the demo.

sergei-maertens:
Created #979 for the cleanup

PR #982, opened 2021-11-22
Printable data for radio/select components in submission report 

SilviaAmAm:
Part of #974

### DIFF
--- a/src/openforms/submissions/models.py
+++ b/src/openforms/submissions/models.py
@@ -375,7 +375,9 @@ class Submission(models.Model):
                     "value": merged_data[key],
                     "label": component.get("label", key),
                     "multiple": component.get("multiple", False),
-                    "values": component.get("values"),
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/models.py
```diff
@@ -436,6 +438,12 @@ def get_merged_data(self) -> dict:
 
         return merged_data
 
+    def _get_value_label(self, possible_values: list, value: str) -> str:
```
sergei-maertens, line 441: `self` is unused, so this can be a `@staticmethod`

I think in the longer term we should look into a system to define data formatters/getters, something like:

```py
@formatter("select")
def get_select_repr(component: dict, value: Any) -> str:
    values = component["data"]["values"]
    ...
    return foo
```

& then have fuctions for the `formio` component types.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1001, opened 2021-11-26
Technical (developer) documentation for Open Forms

sergei-maertens:
First pass at restructuring the "public" API documentation of Open Forms.


sergei-maertens:
Merging this, we can refine later and I want to prevent some code rot.

Issue #957, opened 2021-11-18
Users need to re-login when continuing a form or changing their appointment


joeribekker:
Currently, it's unclear how this works and/or what is the desired behaviour.

* [x] Investigate the current behaviour of a form, where authentication required, is continued with a resume-link (from "tussentijds opslaan").
* [x] Investigate the current behaviour of an appointnent form, where authentication required, where the user clicked on the link to change the appointment.

**Verduidelijking nodig**

* [x] Is inloggen opnieuw nodig na hervatten van formulier na tussentijds opslaan (mits het originele formulier inloggen verplicht had) -> JAAA
* [x] Is inloggen opnieuw nodig om afspraak te wijzingen (mits het originele formulier inloggen verplicht had) -> JAAA

SilviaAmAm:
This is what happens now when a user resumes a form by clicking on a link `/submission/<submission_id>/<token>/resume`:

1. The token and the submission ID are checked
2. If valid, the submission is re-added to the session
3. The user is redirected to the last edited step of the form in the SDK: `/stap/<step_slug>?submission_uuid=<submission_uuid>`
4. The SDK uses the submission id in the query param to retrieve the submission data and loads it in the state.

To make the user login again, the `ResumeSubmissionView` can check if the form requires a login and:

- if yes: redirect to a new `/resume?next='/step/step_slug?submission_uuid=<submission_uuid>'` where the user does the login and then gets redirected to the 'next' url
- if no: same as now

For cancelling/changing an appointment, the flow is similar as when resuming a form.

sergei-maertens:
@SilviaAmAm analysis/design on how to tackle this. This takes out the step where for re-login the user has to select the authentication method (as we re-use what was initially used), so I think there should be no or minimal SDK changes needed.

# Resume form with login flow

I've worked out the flows for both initial and resume authentication flows for a form.

Note that this approach explicitly takes into consideration the following issues/
user stories:

* #XXX: displaying authentication status in submission audit log
* #XXX: ability/mechanism for co-signing (additional authentication on top of main auth)

Below are two sections: a description of the flow users go through and a section
describing (some of) the technical changes needed in the codebase to facilitate this.

## Authentication flows

For completeness sake, both the initial and resume flows are described to see where
the overlap is, and how they work together. Changes to both flows are needed.

**Initial form start authentication**

1. User starts form and login options are presented (DigiD, eHerkenning, EIDAS)
2. User selects plugin, e.g. DigiD
[...]

PR #1004, opened 2021-11-29
[#957] Login when resuming form

SilviaAmAm:
Part of #957 

Frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/120

### DIFF
--- a/src/openforms/submissions/tests/test_endpoints.py
+++ b/src/openforms/submissions/tests/test_endpoints.py
@@ -8,6 +8,7 @@ from freezegun import freeze_time
 from furl import furl
 
 from openforms.config.models import GlobalConfiguration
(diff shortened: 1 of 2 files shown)


SilviaAmAm:
There are a few things that don't convince me about this approach:
- Even if you have a login screen on the `/resume` page, you can add `_start=1` to the url or click on the form step links on the sidebar and the form will start anyway because the submission is in the session (although adding `_start=1` to the URL has the same effect on the form start page).
- Currently, if the login fails the submission id remains in the session. If we were to clean the session after a failed login, the user will never be able to resume the form (since the resume token is valid only once and it is in the resume view that the submission is readded to the session).

SilviaAmAm:
Closed in favour of https://github.com/open-formulieren/open-forms/pull/1021

Issue #235, opened 2021-04-26
Een gebruiker kan via condities in een formulier meerdere of diverse betaalmogelijkheden beschikbaar hebben. Voorbeeld: ...


joeribekker:
...Bij resultaat A betaalt een bezoeker 50 euro, bij resultaat B 100 euro (deze eis valt buiten het MVP).

Eis: Sturing Formulier 9.
Zie ook NvI: 76, 158

**Tasks**

* [x] Add form logic action to set (and override) the price (if choice A then price=X; if choice B then price=Y)
* [x] Save the resulting price on the form
* [x] Make sure the overriden price is used when submitting the form (in summary and in payment)

sergei-maertens:
Discussed yesterday with Joeri.

The simple implementation would be:

1. Use static pricing (from connected product) or dynamic pricing (see 2.)
2. Define conditions just like in form logic "when component X has value Y" -> then set price A

Multiple rules are possible that change the price and the connected product price is no longer used.

joeribekker:
We will go for option 2, which will also imply to remove Products entirely, and set arbitrary prices on the form itself (instead of products). Ticket updated with specs.

joeribekker:
Issue #187 conflicts with my earlier comment...

I think both issues conflict a bit anyway: Manage prices centrally (#187) VS Allow per form price adjustments based on logic (this issue).

PR #1006, opened 2021-11-29
Dynamic price rules

sergei-maertens:
Plumbing for #235

**Changes**

- Added price logic rules data model
- Added price logic rules to the API


src/openforms/forms/api/validators.py
```diff
@@ -1,14 +1,118 @@
...
+        self.trigger_field = trigger_field
+        self.form_field = form_field
+
+    @staticmethod
+    def _get_field_from_data_or_instance(
+        field: str, data: dict, serializer: serializers.Serializer
+    ) -> Any:
+        if not (value := data.get(field)):
+            serializer_field = serializer.fields[field]
+            value = serializer_field.get_attribute(serializer.instance)
```
SilviaAmAm, line 76: Does this also work for selectBoxes, which would have a 'var' name such as `componentName.varName`?

src/openforms/forms/api/viewsets.py
```diff
@@ -83,12 +92,36 @@ class FormLogicViewSet(
 ):
     serializer_class = FormLogicSerializer
     queryset = FormLogic.objects.all()
-    permission_classes = [permissions.IsAdminUser]
+    permission_classes = [
+        permissions.IsAdminUser
+    ]  # TODO: connect with can_change_form admin permission!
```
SilviaAmAm, line 97: Todo for later?

src/openforms/forms/tests/test_api_form_price_logic.py
```diff
@@ -0,0 +1,164 @@
...
+        response = self.client.post(self.list_url, data=price_logic_data)
+
+        self.assertEqual(status.HTTP_201_CREATED, response.status_code)
+        price_logics_qs = FormPriceLogic.objects.all()
+        self.assertEqual(price_logics_qs.count(), 1)
+        price_logic = price_logics_qs.get()
+        self.assertEqual(price_logic.form, self.form)
+        self.assertEqual(price_logic.price, Decimal("15.00"))
+
+    def test_create_logic_with_dates(self):
```
SilviaAmAm, line 91: Might also be good to have a test with selectBoxes?

SilviaAmAm (changes requested):
Looks good! I added just some questions

src/openforms/forms/api/validators.py
sergei-maertens: this doesn't dive into Formio configuration at all - it's a generic variant of 
```py
form = data.get("form") or serializer.instance.form
```

src/openforms/forms/api/viewsets.py
sergei-maertens: yes, Joeri and I are aware of this and it's tracked _elsewhere_ :wink:

src/openforms/forms/tests/test_api_form_price_logic.py
sergei-maertens: opted to not do that since the test exists for the regular logic rules which follows the same validators. This test could also be dropped, but now there's a least something that validates that the validators aren't accidentally removed.

PR #1020, opened 2021-12-02
Implement co-signing flow (backend)

### LINKED ISSUE #296: De gebruiker kan tegelijk met een andere gebruiker via DigiD, eHerkenning en eIDAS inloggen en een formulier verzenden ...
...(co-signeren).

Eis: Authenticatie En Basisgegevens 4a.
Zie ook NvI: 78

## Specification

**Description of the flow**

[view/edit sequence diagram](https://sequencediagram.org/index.html#initialData=C4S2BsFMAIGEHsC0BlEBzAdiDboEMBXYAC2gDNx4B3aACgGIAmATgDYBKAKE7wGNh4AJzhJUmSIJ78h0AKIYAJogIBnCdwAOeQaF4gtGYNGQARANKctOkHoNGAQnwDWkRZe279eQ9BPoQJtyc8kqqEogAfKZmAFz4RMSuunjAMABk5CDg4NDwRORCALac0ZHRcX4qGuB4AJ4iKOgY0LzwhRrwGEncIcpqgmXmcbzgNk7QAER+aAET0ABGRALN2A1iza3tnd1SoABuKTDRJeaRjrwuinHIeHswKqka0AopeHQACgCqACpc55dKQaxaAACW+33e0EYAAZodAAPIWE5mM7OVwKOKCSAKEBY-jQATQAD0hBIRIAPCpwAQ0BEiTiZgoiQ8PESAPytdbIAjzQogFQqECdAC8lJ5fIFQowAEkTBE0l0AB7AUVkIrIR6fQTgCLcPj7Q7Qf7ozgKSD6kAHVLGczcY2KVEXdFxbAjAhmlrwdbQKx4QrQVbTEAKaAAJUgNVqGsOdrRDoiQZMcXDOLxRkJib1-EthszZotVpg9oU3AQjXEAwT-iT8RISRsMcTkUTcQOoxeqU4TcdAMx2NxkH43Hz2cLvmruxz1uLsad8eLcQAanh24cia0MCoCIUjgBBACyABl8AKJKBOipOMWe87npBUvjOU0fdo-QHmuHI9HO9eIgvoG2waGluvL8oKnTQLK+BoHg2APNAaiSp0V5xkof6oXEDxCDArSIIKmDPK8uTNCBErgRgs4AkCfapoORjzM4BLwNA5JqoIhQapAGhajqprmqOhozpOY7HMiN5XOQ968KQuH4SRwCGiSGggESewAIzMuKYFSioFKkdpnTSgodKyU0KFzoCURDFCsIImY0AADoYAA3hMT7iAoExxMAggEJAAA0kxkCAEYKCoXnQK5wWCA8AByfqQBFExqvAEwAL5pcO-EGtaom9GElblJ6hjYH5J7QKoBArkAA)
co-signing

## Design Principles

- No sensitive information can be modified by the client (BSN for example)
- State is saved on the backend only
- No sensitive data goes over the wire (from backend to SDK or vice versa)
- Re-use existing authentication plugins
- Generic implementation that can be used by multiple plugins, initial focus on DigiD

## Tasks

**Backend**

- [x] Add `Submission.co_sign_data` `JSONField` model field
    - Data structure is

      ```json
      {
        "plugin_id": "digid",
        "identifier": "123456782",
        "fields": {
            "first_name": "Foo",
            "last_name": "Bar"
        }
      }
      ```
    - `fields` in the above datastructure depends on the `plugin_id`
- [x] Extend authentication module/plugins (initially only DigiD/base plugin)
    - Add new parameter to `/auth/{slug}/{plugin_id}/start` endpoint: `coSignSubmission`
        - value is a submission ID filled out by the SDK
        - validate the submission ID against the session (look at 
          `openforms.submissions.api.permissions.ActiveSubmissionPermission`)
        - presence (and validity) of this parameter indicate a co-sign authentication 
          (as opposed to a form start authentication)
    - When redirecting to DigiD (or other plugins), the `coSignSubmission` or equivalent
      parameter should be added to the `RelayState` (or plugin-specific equivalent). The 
      goal is that the return path after auth can still detect co-sign auth flows
    - add `handle_co_sign` method on `BasePlugin`: the goal is to store the appropriate 
      data on the `Submission.co_sign_data` field. The submission is obtained via the
      `coSignSubmission` parameter.
    - in the return path of the auth flow, detect co-sign, validate submission ID again
      and invoke the new plugin method to store the data on the submission
- [x] Implement `/api/v1/submissions/<uuid>/co-sign` endpoint
    - it should return a response in the following shape:

      ```json
      {
        "coSigned": true,
        "fields": {
            "firstName": "Foo",
            "lastName": "Bar"
        }
      }
      ```
    - do NOT return a BSN, KVK ID or similar, only the state if it's co-signed or not
- [x] Define a custom FormIO component of type `co-sign`, which may be in its own ection in the form designer
    - ensure label can be specified
    - ensure help text can be specified
    - initially, hard-code it to use DigiD but in the future it should be possible to 
      select a possible authentication backend (similar to log in but only a single one)
- [ ] Add Co-Sign information ("co-signed by Foo Bar") to generated PDF
- [ ] Pass co-sign status/information to registration backend
    -> TODO: how exactly?
- [x] Implement filling out the fields after co-sign auth
    - [x] Implement a generic error message "something went wrong, please contact us" if pre-fill plugin/client fails (because of mis-configuration)
    - [x] Implement fetching the representational fields using the prefill machinery
- [ ] Update cleaning of sensitive data to wipe co-sign data

**SDK**

- [x] Implement client-side custom FormIO component of type `co-sign`
  - Probably should just render a react component, see the custom map component for this
  - Component should fetch coSign state from backend by hitting the 
    `/api/v1/submissions/<uuid>/co-sign` endpoint
  - if not cosigned, render the login button. Either hardcode DigiD at first, or expose
    `plugin_id=digid` in the component configuration initially.
  - re-use the authentication components (React) _if possible_
  - Below the button, the plugin-specific fields should be displayed if `coSign` state 
    is `true`: e.g. First name & Last name (either just as text, or read-only text fields)
- [x] implement On co-sign "authenticate" button click:
  - save the step data to not lose partially filled out form data (challenging - we need 
    to break out of the formio component to get access to the react FormStep component. 
    Possibly we can pass a callback down to the Formio options to invoke)
  - initiate the redirect to `/auth/{slug}/{plugin_id:digid}/start` endpoint, passing 
    the two query string parameters:
    - `coSignSubmission`: submission ID
    - `next`: the current step url (`window.location` should suffice)
- [x] After the backend redirects, the form step should re-load and the co-sign plugin
  reloads itself too, now getting a `coSign: true` state from the backend and it should
  render the appriopriate fields for display.

sergei-maertens:
Fixes #296

**Changes**

- Set up base co-sign flow: storing state on Submission, defining abstract base plugin method
- Added co-sign implementation for a subset of plugins:
    - DigiD-mock
    - DigiD
    - Demo
    - rest will follow later in another PR
- Implemented endpoint to read submission co-sign state

Note: I have mixed feelings about the way it's implemented at the moment, but not enough time to fix/re-design this :grimacing:


src/openforms/authentication/contrib/demo/plugin.py
```diff
@@ -14,13 +14,21 @@
 from openforms.forms.models import Form
 from openforms.utils.validators import BSNValidator
 
-from ...constants import FORM_AUTH_SESSION_KEY, AuthAttribute
+from ...base import BasePlugin
```
SilviaAmAm, line 17: On line 13 there is still `from openforms.authentication.base import BasePlugin` which should be removed

SilviaAmAm (approved):

src/openforms/authentication/contrib/demo/plugin.py
sergei-maertens: yeah, I spotted it in `feature/296-co-sign-formio-component` and fixed it there! Thanks for pointing it out though.

PR #1023, opened 2021-12-02
[#941] Added configuration check for Objects API registration 

Bartvaderkin:


src/openforms/registrations/contrib/objects_api/checks.py
```diff
@@ -0,0 +1,79 @@
...
+        "objecttype",
+        "informatieobjecttype_submission_report",
+        "informatieobjecttype_submission_csv",
+        "informatieobjecttype_attachment",
+    ]
+    for service_field in urls:
+        url = getattr(config, service_field)
+        if not url:
+            # they are all defaults
+            return
```
sergei-maertens, line 57: as soon as one is false-ish, then the check should abort completely? should this not be a `continue` statement rather than a `return` statement?

src/openforms/registrations/contrib/objects_api/checks.py
```diff
@@ -0,0 +1,79 @@
...
+
+        try:
+            res = requests.get(url)
+            res.raise_for_status()
+        except Exception as e:
```
sergei-maertens, line 62: as far as I understand, this makes an un-authenticated request, which will raise HTTP 403's and thus cause problems in the check, while the config would be correct, no?

The respective `client` object instead should have the correct configuration to make authenticated calls, or am I misreading this code?

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/checks.py
Bartvaderkin: yes it probably should, ill change it

src/openforms/registrations/contrib/objects_api/checks.py
Bartvaderkin: i assumed these were public but you're right and we should use the client

sergei-maertens (approved):

PR #1024, opened 2021-12-02
[#941] Added configuration check for Stuf-ZDS registration 

Bartvaderkin:


src/openforms/registrations/exceptions.py
```diff
@@ -1,5 +1,6 @@
 class RegistrationFailed(Exception):
-    pass
+    def __init__(self, *args, **kwargs):
+        self.response = kwargs.pop("response", None)
```
sergei-maertens, line 3: the `super().__init__(*args, **kwargs)` call is missing

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #1007, opened 2021-11-29
Configurable informatieobjecttype per upload field for ZGW/Objects API registration backend


stevenbal:
Currently, there can only be one informatieobjecttype that is used for file uploads in the Objects API registration backend. This should be configurable per file upload field

- [x] UploadFields should have an extra tab `Registratie` where an Informatieobjecttype can be supplied (see also: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/form/edit/tabs.js#L231)
- [ ] The Informatieobjecttype should be selected from a dropdown/radio menu (grouped by catalogus)

PR #1044, opened 2021-12-07
[DRAFT] Objects API registration IOtype configurable per upload field

stevenbal:
Fixes: #1007

* Allows using different InformatieObjecttypes for different upload fields, configurable via the admin interface


stevenbal:
@sergei-maertens since this PR makes it configurable per upload field which informatieobjecttype should be used when creating `EnkelvoudigInformatieObject`en for them, I would like the upload field name to be selectable with a dropdown.

Do you know if there is a way currently to dynamically build the `choices` for the field_names? It seems that the serializers do not get any context as to what form is being edited.

This is what it currently looks like in the registration configuration form:
image

stevenbal:
Closed in favor of https://github.com/open-formulieren/open-forms/pull/1164

PR #1048, opened 2021-12-08
[#941] Added configuration check for appointments module 

Bartvaderkin:


Bartvaderkin:
This is not the prettiest because it doesn't use the registry-pattern, but the closest I could get without refactoring too much.

sergei-maertens (approved):
It's contained - that's sufficient for now :)

PR #1049, opened 2021-12-08
[#941] Added configuration check for BAG address lookup 

Bartvaderkin:


Bartvaderkin:
This also bypassed the registry-pattern but the BAGClient is almost a 'plugin' so I ducktyped it as such.

sergei-maertens (approved):

Issue #990, opened 2021-11-24
The PDF report generation breaks with a multivalue field if that field is left empty (tested with email field).


sergei-maertens:
Tasks:

- [x] regression test
- [x] bugfix

PR #1058, opened 2021-12-10
[#990] Added regression test for multivalue field breaking PDF report

Bartvaderkin:


Bartvaderkin:
This is a big mess and really needs #1059

I did the quickfix that keeps the current situation and output: for example these values are not localized.

src/openforms/submissions/tests/factories.py
```diff
@@ -68,6 +68,7 @@ def from_components(
         cls,
         components_list: List[dict],
         submitted_data: dict = None,
+        with_report=True,
```
sergei-maertens, line 71: would be better suited as a trait:

```py
with_report = factory.Trait(
    report=factory.SubFactory(SubmissionReportFactory, submission=factory.SelfAttribute(".."))
)
```

not sure if this exact invocation of `SelfAttribute` works, but it should be something like that. See docs: https://factoryboy.readthedocs.io/en/stable/recipes.html#copying-fields-to-a-subfactory

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
:adhesive_bandage:

PR #1063, opened 2021-12-10
[#924] Update `can_submit` field

### LINKED ISSUE #924: Allow hiding the overview page which also disabled submissions
**Tasks**

* [x] Refactor and change the "Inzenden mogelijk" field to an pulldown: Inzenden: 1) Ja  2) Nee, met overzichtspagina  3) Nee, zonder overzichtspagina 
* [x] Update admin page
* [x] Hide the step "Bevestiging" in the step overview (in the frontend) if the 2nd option is selected
* [x] Hide the step "Bevestiging" AND "Overzicht" in the step overview (in the frontend) if the 3nd option is selected
* [x] If the 3rd option is selected, there should no longer be a "Volgende" (next) button in the last form step.

SilviaAmAm:
Fixes #924 

Frontend: https://github.com/open-formulieren/open-forms-sdk/pull/128


sergei-maertens (changes requested):
Looks like this will crash during migrations to me - chancing from BooleanField to CharField?

Additionally, if it doesn't, we definitely need a data migration to map from boolean to choice option.

src/openforms/forms/migrations/0009_transfer_can_submit.py
```diff
@@ -0,0 +1,40 @@
...
+def forwards_func(apps, schema_editor):
+    Form = apps.get_model("forms", "Form")
+    forms = Form.objects.all()
+
+    for form in forms:
+        if form.can_submit:
+            form.submission_allowed = SubmissionAllowedChoices.yes
+        else:
+            form.submission_allowed = SubmissionAllowedChoices.no_with_overview
+    Form.objects.bulk_update(forms, ["submission_allowed"])
```
sergei-maertens, line 17: FYI - this could be simplified to this:

```py
Form.objects.filter(can_submit=True).update(submission_allowed=SubmissionAllowedChoices.yes)
Form.objects.filter(can_submit=False).update(submission_allowed=SubmissionAllowedChoices.no_with_overview)
```

And more complicated single query (hurts readability):

```py
Form.objects.update(
    submission_allowed=Case(
        When(can_submit=True, then=Value(SubmissionAllowedChoices.yes)),
        When(can_submit=False, then=Value(SubmissionAllowedChoices.no_with_overview)),
    )
)
```
for completeness sake.

You can leave this as it is, it looks correct.

SilviaAmAm: Ohh good to know! :slightly_smiling_face:

sergei-maertens (approved):

Issue #1064, opened 2021-12-10
Issues noticed during sprint 10 demo preparation


sergei-maertens:
- [x] Multi-value ChoicesJS select is not displayed properly in summary (seen screenshot) @SilviaAmAm https://github.com/open-formulieren/open-forms-sdk/pull/126
      
  ![image](https://user-images.githubusercontent.com/5518550/145613543-2216add4-3664-4efa-8d5c-cca9a079f6b6.png)

- [ ] #1090

- [x] HTTP 500 response on status poll (no registration backend set up) (see Sentry 298511) **Prio medium** (@sergei-maertens)
    - [x] Ensure that `Form.submission_confirmation_template` is template syntax validated

- [ ] #1089
- [x] Appointment information in PDF is not nicely formatted, see screenshot  @joeribekker 
  
  ![image](https://user-images.githubusercontent.com/5518550/145617809-1d24a1af-6e39-4048-9004-fa775bd57027.png)

- [x] Confirmation e-mail sending fails (Sentry 298516, might be fixed by #1058) @sergei-maertens

sergei-maertens:
I've created new issues for the remaining items

Issue #1068, opened 2021-12-13
Refactor formio integration


sergei-maertens:
From #1099, #1059 and #1054 it's clear that we need to re-structure our code to manage (custom) formio types:

* Hooking up special custom types with backend mutations
* Formatting the values of vanilla and custom FormIO components

This ticket describes the technical design to resolve both mentioned issues.

**Design principles**

* We provide first class support for FormIO
* FormIO (or generally: vendor) specific integration code is contained in its own package (i.e. -> do not intertwine it with vendor-agnostic OF code)
* Pluggable & testable

**Tasks**

- [x] Create a new top-level package `openforms.formio`
    - [x] Subpackages `custom_types` and `formatters`
- [x] Set up the custom field registry
     - [ ] define the register in `openforms.formio.custom_types.registry` (move from `openforms.forms.custom_field_types`)
     - [x] Ensure that the register inherits from our base register class to make things consistent
     - [ ] Define a custom base type inheriting from `openforms.plugins.plugin.AbstractBasePlugin`
     - [ ] A custom type can have two modes of operation:
         - Return value to append FormIO component schema(s)
         - Mutates given FormIO component schema
     - [x] The handler must process components recursively (i.e.: use `FormDefinition.iter_components(...)`, currently not the case
     - [x] The handler also takes a submission object for context, instead of the request. The information possibly on the request should be persisted on the submission instead (such as authentication state).
     - [x] Split up hooks into part with and without input data ('static' impact + data-dynamic impact)
- [x] Set up the formatters registry in `openforms.formio.formatters.registry`
    - [x] A formatter takes two arguments: `value` and `component`, with `component` being the FormIO json schema
        - This has a big impact - the extra context of `component` is now not passed along and will lead to restructing in other places!
    - [x] Define a module with vanilla Formio formatters `openforms.formio.formatters.default`
    - [x] Define a default fallback formatter (could simply be `str(value)`), used for (custom) types without explicitly registered formatter, to avoid crashes
- [x] Consider marking formatters (or types?) as presentation only (vs. data) -> WYSYWIG, fieldsets, columns... (part of #1451)

SilviaAmAm:
In PR https://github.com/open-formulieren/open-forms/pull/1647#discussion_r886544587 I added a data mapping between the Formio components and the python type: https://github.com/open-formulieren/open-forms/blob/feature/1325-submission-variable-backend/src/openforms/formio/constants.py#L1

PR #1069, opened 2021-12-13
Fixes selects rendering in DPF

joeribekker:
Refs #1064


sergei-maertens (approved):
#1068 is ever so relevant :)

Issue #1077, opened 2021-12-13
Form Builder help texts for steps are broken


stevenbal:
In the form builder, the help texts for the form step details aren't working properly

When hovering the help text icon some fields, the next field is pushed to the right, for example:
![Screenshot from 2021-12-13 15-40-31](https://user-images.githubusercontent.com/29249171/145846431-d0aa23be-ab7d-4cc3-a998-067fa3434f26.png)

For the bottom field, the help text doesn't show:
![Screenshot from 2021-12-13 12-16-41](https://user-images.githubusercontent.com/29249171/145846505-bd9976dd-8ea7-4bbb-8610-c0cdffbe9a9f.png)

joeribekker:
Made several changes:

Special widgets:
![image](https://user-images.githubusercontent.com/96970/146174590-7c36c81a-3ede-446a-83d8-716a469ab228.png)

Read only fields:
![image](https://user-images.githubusercontent.com/96970/146174659-8ba63ad5-914b-41c4-9c7c-2922bf5e2b2c.png)

React stuff:
![image](https://user-images.githubusercontent.com/96970/146177501-260ad79b-f497-4b97-819d-b47ca6fa6775.png)

PR #1078, opened 2021-12-13
:lipstick: [#1077] Fix admin help text hover

stevenbal:
fixes: #1077

### DIFF
--- a/src/openforms/scss/_vars.scss
+++ b/src/openforms/scss/_vars.scss
@@ -16,5 +16,9 @@ $color_lightest: #fff;
 
 $color-error: #ba2121; // default django admin color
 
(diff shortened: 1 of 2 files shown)


stevenbal:
Fixed the help text hover:
Screenshot from 2021-12-13 17-14-33

sergei-maertens:
We had made a start using a popover library and this patch probably breaks
the style in other places as this is what's included in default project I
believe

sergei-maertens:
@joeri can you validate this, I think you originally made these styling setups?

joeribekker:
Made some improvements with what we have.

joeribekker (approved):

PR #1087, opened 2021-12-16
[#1075] Fixed datetime and date localisation display in printable data

### LINKED ISSUE #1075: Datums zijn in US-notatie in de PDF
Gerelateerd aan #1068

Bartvaderkin:
Closes #1075

### DIFF
--- a/src/openforms/submissions/models.py
+++ b/src/openforms/submissions/models.py
@@ -10,9 +10,11 @@ from django.contrib.postgres.fields import JSONField
 from django.core.files.base import ContentFile, File
 from django.db import models, transaction
 from django.template import Context, Template
+from django.template.defaultfilters import date as fmt_date
 from django.template.loader import render_to_string
 from django.urls import resolve
 from django.utils import timezone
+from django.utils.dateparse import parse_date, parse_datetime
 from django.utils.translation import gettext, gettext_lazy as _
 
 from celery.result import AsyncResult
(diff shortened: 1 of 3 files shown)


sergei-maertens (changes requested):
Checking the form with all elements (https://open-forms.test.maykin.opengem.nl/admin/forms/form/35/change/) I don't see a `type: datetime`, but only the types `date` and `time`. This is also confirmed by the types (keys) here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/formio_module.js#L28

So I suspect that the `datetime` block will never be invoked.

I'm also unsure now what happens with appointments, where the date/time is part of the `select` type and has no inherent date/time information. We do have meta-information on the component that relates it to the appointment date/time field though.

src/openforms/submissions/models.py
```diff
@@ -511,8 +513,24 @@ def _get_value_label(possible_values: list, value: str) -> str:
...
+            return list(filter(filter_func, value))
+        elif not filter_func(value):
+            return []
+        else:
+            return [value]
+
+    @classmethod
+    def _join_mapped(cls, func, value, seperator=", "):
+        # filter and map a single or multiple value into a joined string
+        return seperator.join(map(func, cls._get_values(value)))
```
sergei-maertens, line 532: `filter` and `map` aren't very pythonic - there are some voices that even claim it was a mistake to add them to Python.

can you rewrite these as list comprehensions instead - that's usually more readable and pythonic

sergei-maertens: Additionally, these two static/class methods seem to be used only in a single place - can't we just make these module-level helper functions instead of static/class methods? I don't mind fat models, but this is getting very bloated.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1093, opened 2021-12-17
:bug: Set indicatieGebruiksrecht to False for documents

stevenbal:
otherwise issues will occur when trying to close related Zaken


sergei-maertens (approved):

joeribekker (commented):
I agree with this.

PR #1096, opened 2021-12-17
Fix self-certifi flow with runserver

sergei-maertens:
Detection of REQUESTS_CA_BUNDLE is broken on runserver since the module
global mutation is not picked up in the runserver thread. This is now fixed
by looking at os.environ which is a mutable dict.

### DIFF
--- a/src/openforms/setup.py
+++ b/src/openforms/setup.py
@@ -24,8 +24,6 @@ from dotenv import load_dotenv
 from requests import Session
 from self_certifi import load_self_signed_certs as _load_self_signed_certs
 
-_certs_initialized = False
-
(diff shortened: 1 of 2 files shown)


joeribekker (approved):
Feels like it should be "fixed" in self-certify but confirmed this solved the issue.

sergei-maertens:
> Feels like it should be "fixed" in self-certify but confirmed this solved the issue.

Self-certifi should not introduce any other side-effects than setting the requests envvar. And self-certifi cannot know if you deliberarly set the requests envvar or not, so it doesn't know if it's a mistake or intentional.

Making sure that the initialization is only done once is the responsibility of the caller

PR #1108, opened 2021-12-22
Camunda: start process instance with submission data process variables

sergei-maertens:
Relates to #326

**Changes**

* Kick off process instance: latest version (by key) or specific version (by ID)
* Temporary: hardcoded process variables for the Camunda demo app (invoice)
* Error handling
* Small management command to run registration for a submission (dev only)


joeribekker (approved):
Looks good and works as described.

(had to use the sdk because submission=yes didn't get picked up by open forms project)

PR #1110, opened 2021-12-22
Add support for location -> product flow

joeribekker:
Instead of product -> location flow.


joeribekker:
@annashamray I assigned you in refinement because it adds code that was made to solve one of your tasks (but the code decayed since 2021). Specifically the support for multiple products and selecting the location first.

PR #1115, opened 2021-12-23
Select process variables from defined components

sergei-maertens:
Relates to #326 

First phase - simple process variable mapping.

**Changes**

- Manage process variables from the registration options
- Every form component can be added as a proces variable
- The Camunda variable name can be overridden, by default the Formio component.key is used
- Handle formio schema changes _after_ the variable mapping is configured (deleted components are deleted from mapping, changed components are updated in the mapping)
- Mapped variables are now included on process instance create. Formio types are converted to python types, which are in turn serialized to Camunda types through django-camunda.


src/openforms/js/components/admin/tables/ChangelistTable.js
```diff
@@ -1,10 +1,60 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 import ChangelistColumn from './ChangelistColumn';
 import {IconYes, IconNo, IconUnknown} from '../BooleanIcons';
 
 
+const ChangelistTableWrapper = ({ headColumns, children: body, extraModifiers=[] }) => (
+    <div className={classNames('changelist', 'changelist--react', ...extraModifiers.map(mod => `changelist--${mod}`))}>
```
SilviaAmAm, line 10: May be nice to have a utils function `getBEMClassName` like in the sdk?

SilviaAmAm (changes requested):
Looks good! Just some small comments/questions

src/openforms/js/components/admin/form_design/registrations/camunda/FormFields.js
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        // call parent event-handler with fully updated form data object
+        onChange(updatedFormData);
+    };
+
+    const onAddProcessVariable = () => {
+        const nextFormData = produce(formData, draft => {
+            if (!draft.processVariables) draft.processVariables = [];
```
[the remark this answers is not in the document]
sergei-maertens: Correct! you can do deep mutations in immutable objects (which `formData` is as it comes from props)

src/openforms/js/components/admin/form_design/registrations/camunda/FormFields.js
```diff
@@ -0,0 +1,223 @@
...
+
+    const onFieldChange = (event) => {
+        const {name, value} = event.target;
+        const updatedFormData = {...formData, [name]: value};
+
+        switch (name) {
+            // if the definition changes, reset the version
+            case 'processDefinition': {
+                updatedFormData.processDefinitionVersion = null;
+                break;
```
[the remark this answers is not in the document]
sergei-maertens: good shout, I'll have to think about that a bit more. I think it'd make sense to have a button/action too to "add all form fields", then a reset wouldn't hurt too much and it makes sense as you usually wouldn't define fields that you don't use anyway.

src/openforms/js/components/admin/tables/ChangelistTable.js
sergei-maertens: yeah the thought has occurred... we actually have bem.js as dependency already so I'll see if I can make that do what I want it to do.

sergei-maertens:
@joeribekker backend is ready for review

src/openforms/registrations/contrib/camunda/apps.py
```diff
@@ -1,12 +1,22 @@
...
+from django_camunda.utils import TYPE_MAP as CAMUNDA_TYPE_MAP
+
 
 class CamundaApp(AppConfig):
     name = "openforms.registrations.contrib.camunda"
     label = "registrations_camunda"
     verbose_name = _("Camunda registration plugin")
 
     def ready(self):
+        patch_django_camunda()
```
joeribekker, line 15: Why not update django_camunda? This doesn't seem OF specific?

joeribekker (approved):

src/openforms/registrations/contrib/camunda/apps.py
sergei-maertens: as discussed during standup - mapping a Python `Decimal` with fixed precision to a Java Double (with floating point precision errors) is probably not guaranteed to be correct, but a best effort we can do.

At some point we'll probably include it in django-camunda with a disclaimer or a setting to enable/override it.

PR #1116, opened 2021-12-24
:recycle: [#1068] FormIO formatters

stevenbal:
Adresses part of #1068 

Changes:
- Added formatters for the default FormIO fields we use
- Added feature flag to enable these custom formatters in `Submission.get_ordered_data_with_component_type` and `Submission.get_printable_data`

The formatters aren't used in the confirmation page yet


src/openforms/formio/formatters/default.py
```diff
@@ -0,0 +1,122 @@
...
+@register("selectboxes")
+class SelectBoxesFormatter(AbstractBasePlugin):
+
+    def format(self, component: Dict, value: str) -> str:
+        selected_labels = [
+            entry["label"]
+            for entry in component["values"]
+            if value.get(entry["value"])
+        ]
+        return ", ".join(selected_labels)
```
stevenbal, line 94: @sergei-maertens just checking, is this sort of what you had in mind for the formatters part of https://github.com/open-formulieren/open-forms/issues/1068? My idea would be that you could use it in `Submission.get_printable_data` like this:

```python
formatter = register[info["type"]] if info["type"] in register else register["default"]
printable_data[label] = formatter.format(info, info["value"])
```

sergei-maertens: yeah, that looks good!

src/openforms/formio/formatters/default.py
```diff
@@ -0,0 +1,122 @@
+from typing import Dict
+
+from openforms.plugins.plugin import AbstractBasePlugin
+
+from .registry import register
+
+
+@register("default")
+class DefaultFormatter(AbstractBasePlugin):
```
sergei-maertens, line 9: I'd define a base class specific to formatters and inherit from that rather than directly from `AbstractBasePlugin`:

```py
class FormioFormatter(AbstractBasePlugin):
    def __call__(self, component: dict, value: Any) -> str:
        return self.format(component, value)

    def format(self, component: dict, value: Any) -> str:
        raise NotImplementedError("%r must implement the 'format' method" % type(self))
```

The `__call__` approach is neat because you can then clean up calling code something like this:

```py
formatter = register[info["type"]] if info["type"] in register else register["default"]
formatted = formatter(info, info["value"])
```

The whole `info`/`info["value"]` situation in `get_printable_data` is messed up too, since it only pulls a subset of the keys. In my opinion it should just pass the entire formio component definition like how you'd get it from `iter_components` so that the full context is passed to each formatter.

sergei-maertens (changes requested):

sergei-maertens (changes requested):
Couple doubts/things to investigate:

* I think sometimes the value can be `None` too, especially in the context of multiple values allowed
* I'm missing the handling of `multiple` values components, and i think this should be part of the `__call__` method in the base formatter class. If there's multiple=True, then the `value` will be a list, and the `.format` method should be called on every item in the list.

src/openforms/formio/formatters/tests/test_default_formatters.py
```diff
@@ -0,0 +1,51 @@
...
+def load_json(filename: str):
+    with open(os.path.join(FILES_DIR, filename), "r") as infile:
+        return json.load(infile)
+
+
+class DefaultFormatterTestCase(TestCase):
+    def test_formatters(self):
+        all_components = load_json("all_components.json")["components"]
+        data = load_json("all_components_data.json")
+        expected = {
```
stevenbal, line 23: Since we do not have formatters for `map` and `signature` yet, should I leave them out for now?

stevenbal:
> I'm missing the handling of multiple values components, and i think this should be part of the __call__ method in the base formatter class. If there's multiple=True, then the value will be a list, and the .format method should be called on every item in the list.

Will the formatted value then be a list of formatted values? Or should the list be joined into a string?

src/openforms/formio/formatters/tests/test_default_formatters.py
sergei-maertens: yeah you can leave them out for now

src/openforms/formio/formatters/default.py
```diff
@@ -0,0 +1,152 @@
...
+    assert isinstance(value, str), "Expected value to be a str"
+
+    return value
+
+
+class FormioFormatter(AbstractBasePlugin):
+    def __call__(
+        self, component: dict, value: Any, multiple: bool = False
+    ) -> Union[str, List]:
+        if not value:
```
stevenbal, line 28: Should this explicitly check for `None`? Could `value` possibly be an empty list as well?

src/openforms/formio/formatters/default.py
```diff
@@ -0,0 +1,152 @@
...
+            return _("empty")
+
+        if multiple:
+            return [self.format(component, v) for v in value]
```
stevenbal, line 32: Should this return a list of formatted values, or should the formatted values be joined?

src/openforms/formio/formatters/default.py
sergei-maertens: I'm not sure - this should become clear by having all components in all variants and leaving stuff blank/filled out. Let's hold off on that and see what we uncover with testing.

src/openforms/formio/formatters/default.py
sergei-maertens: Since it's for display, let's just join them by comma.

sergei-maertens (approved):
I think this is okay to try out. I'm sure we'll see weird things :P

PR #1120, opened 2021-12-24
[#991] Aded support for Form._is_deleted 'soft-delete' in admin and API

### LINKED ISSUE #991: Hide deleted forms in the admin and API.
**Tasks**

* [x] Bulk and individual delete on forms, should only mark a form as deleted (do not really delete the form).
* [x] Hide forms, in the form list admin and API that are marked as deleted.
* [x] In the __str__, add `_(" (deleted)")` to the end if a form is marked as deleted. This makes it clear in the "inzendingen" admin page.
* [x] Make sure new submissions cannot be made onto deleted forms.

Bartvaderkin:
Closes #991

### DIFF
--- a/src/openforms/forms/admin/form.py
+++ b/src/openforms/forms/admin/form.py
@@ -86,6 +86,7 @@ class FormAdmin(
         return (
             super()
             .get_queryset(request)
+            .filter(_is_deleted=False)
             .annotate(anno_name=FirstNotBlank("internal_name", "name"))
         )
 
@@ -274,3 +275,12 @@ class FormAdmin(
     remove_from_maintenance_mode.short_description = _(
         "Remove %(verbose_name_plural)s from maintenance mode"
     )
+
+    def delete_model(self, request, form):
+        # override for soft-delete
+        form._is_deleted = True
+        form.save(update_fields=["_is_deleted"])
+
+    def delete_queryset(self, request, queryset):
+        # override for soft-delete
+        queryset.update(_is_deleted=True)
--- a/src/openforms/forms/admin/form.py
+++ b/src/openforms/forms/admin/form.py
@@ -39,6 +39,57 @@ class FormStepInline(OrderedTabularInline):
     extra = 1
 
 
+class FormDeletedListFilter(admin.ListFilter):
+    title = _("is deleted")
+    parameter_name = "deleted"
+
+    def __init__(self, request, params, model, model_admin):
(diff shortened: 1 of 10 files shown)


src/openforms/forms/admin/form.py
```diff
@@ -86,6 +86,7 @@ def get_queryset(self, request):
         return (
             super()
             .get_queryset(request)
+            .filter(_is_deleted=False)
```
sergei-maertens, line 89: I know the ticket states that it should be hidden in the admin, but I'm wondering if we maybe should use a custom list_filter that defaults to deleted=hidden?

src/openforms/forms/models/form.py
```diff
@@ -244,7 +244,10 @@ class Meta:
         verbose_name_plural = _("forms")
 
     def __str__(self):
-        return self.admin_name
+        if self._is_deleted:
+            return "{}{}".format(self.admin_name, _(" (deleted)"))
```
sergei-maertens, line 248: Always put as much as possible in the translatable string so that arguments can be re-ordered in the translations.

```suggestion
            return _("{name} (deleted)").format(name=self.admin_name)
```

sergei-maertens (changes requested):

src/openforms/forms/admin/form.py
sergei-maertens: Discussed with Joeri - we should go the custom List Filter route. You can check `openforms.submissions.admin.SubmissionTypeListFilter` for some inspiration.

Essentially, the filter should offer two options: "deleted forms" and "available forms", where the former are the soft-deleted forms. The default should be "available forms".

One of the reasons for this is that slugs are still unique over both deleted and non-deleted forms, and you must be able to debug/correct slug validation errors.

src/openforms/forms/admin/form.py
```diff
@@ -39,6 +39,57 @@ class FormStepInline(OrderedTabularInline):
...
 
 
+class FormDeletedListFilter(admin.ListFilter):
+    title = _("is deleted")
+    parameter_name = "deleted"
+
+    def __init__(self, request, params, model, model_admin):
+        super().__init__(request, params, model, model_admin)
+
+        self.request = request
```
sergei-maertens, line 49: didn't notice that in the submissions filter, but `self.request` does not seem to be used anywhere - can this be removed?

sergei-maertens (approved):

PR #1134, opened 2021-12-29
Camunda complex derived process vars

sergei-maertens:
Related to #326

**Changes**

Frontend:

* [x] Introduces a JSON editor where values can be JSON primitives, complex types or form field references
* [x] Form field reference values are to be evaluated using JSON logic
* [x] Variables are edited as complex camunda variables
* [ ] String variable values can interpolate form field reference values (using JSON logic) (separate PR)

Backend:

- [x] Persist complex variable in Camunda options form
- [x] Evaluate complex variables with the submission data and send to process instance create

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -445,12 +445,6 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/Form'
-          application/x-www-form-urlencoded:
-            schema:
-              $ref: '#/components/schemas/Form'
-          multipart/form-data:
-            schema:
-              $ref: '#/components/schemas/Form'
         required: true
       security:
       - tokenAuth: []
@@ -827,12 +821,6 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/Form'
-          application/x-www-form-urlencoded:
-            schema:
-              $ref: '#/components/schemas/Form'
-          multipart/form-data:
-            schema:
-              $ref: '#/components/schemas/Form'
         required: true
       security:
       - tokenAuth: []
@@ -866,12 +854,6 @@ paths:
       requestBody:
         content:
           application/json:
-            schema:
-              $ref: '#/components/schemas/PatchedForm'
-          application/x-www-form-urlencoded:
-            schema:
-              $ref: '#/components/schemas/PatchedForm'
-          multipart/form-data:
             schema:
               $ref: '#/components/schemas/PatchedForm'
       security:
@@ -929,12 +911,6 @@ paths:
       requestBody:
         content:
           application/json:
-            schema:
-              $ref: '#/components/schemas/FormAdminMessage'
-          application/x-www-form-urlencoded:
-            schema:
-              $ref: '#/components/schemas/FormAdminMessage'
-          multipart/form-data:
             schema:
               $ref: '#/components/schemas/FormAdminMessage'
         required: true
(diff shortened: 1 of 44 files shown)


src/openforms/js/components/admin/debug/definition.json
```diff
@@ -0,0 +1,61 @@
+{
```
sergei-maertens, line 1: this is an included sample file of the schema of such a complex variable definition. It serves no other purpose than documentation.

src/openforms/registrations/contrib/camunda/complex_variables.py
```diff
@@ -0,0 +1,187 @@
+"""
```
sergei-maertens, line 1: I'm not entirely happy with this code, but it works (according to the tests) so I'll let it simmer for a bit and revisit at a later stage when we have to touch this again.

src/openforms/registrations/contrib/camunda/fields.py
```diff
@@ -0,0 +1,24 @@
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import fields
+
+
+class NullField(fields.Field):
```
sergei-maertens, line 6: Bit of a weird one since the value and type are the same, but we need to validate this.

src/openapi.yaml
```diff
@@ -445,12 +445,6 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/Form'
```
joeribekker, line 447: I could not directly see the code that caused this change.

joeribekker (changes requested):

src/openforms/registrations/contrib/camunda/plugin.py
```diff
@@ -62,13 +63,25 @@ def get_process_variables(
         alias = simple_mappings[key]
         variables[alias] = to_python(component, value)
 
+    complex_variables = get_complex_process_variables(
+        options.get("complex_process_variables", []),
+        merged_data,
+    )
+
+    if intersection := set(complex_variables).intersection(set(variables)):
+        logger.warning("Complex variables overlap, check the keys %r", intersection)
```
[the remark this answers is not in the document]
sergei-maertens: yeah could actually do this at the main serializer level, come to think of it

src/openapi.yaml
sergei-maertens: `FormViewSet.parser_classes` and `FormViewSet.renderer_classes` :) the DRF default (in Django settings) allows JSON, FormData and multipart

src/openforms/forms/api/drf_camel_case.py
```diff
@@ -0,0 +1,29 @@
...
+
+        .. warning:: this also puts _all_ the plugin-specific keys into the same parser,
+           while we'd rather be able to just validate this depending on the applicable
+           registration backend. Future improvement!
+        """
+        ignore_fields = []
+        for plugin in register:
+            if not plugin.camel_case_ignore_fields:
+                continue
+            ignore_fields += list(plugin.camel_case_ignore_fields)
```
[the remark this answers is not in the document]
sergei-maertens: yeah, I've booked some time in the afternoon of the 17th to work on a POC for more fine-grained control, but we might look into different solutions alltogether.

Issue #1000, opened 2021-11-26
React Form editor allows editing when user does not have write access


stevenbal:
If React is enabled for the Form editor page via the `General configuration`, the `Form` admin page (`/admin/forms/form/<form_id>`) is editable for any user that is staff (even if these users do not have write permission)

Tested locally with a user that is staff and only has read permissions
![image](https://user-images.githubusercontent.com/29249171/143605575-1f7dda89-b036-455c-b010-f393e68d7679.png)

**Tasks**

* [x] Add permission class(es) to API endpoints (form app related)
* [x] Write access (put/patch/post/delete) is ONLY given to staff users that have `forms.change_form` permission 
* [x] Read access (get) to everything give to staff users that have `form.change_form` permission
* [x] Read access (get) to `forms.form_LIST` given to non-staff users that have `forms.view_form` permission
* [x] Read access (get) to `forms.form_DETAIL` given to anonymous users
* [x] Leave other app permissions as they are

PR #1140, opened 2021-12-30
[#1000] Added a first attempt at test matrix for the FormAPIPermissions 

Bartvaderkin:
For discussion: I started this but I don't know if we want it. This idea was to get a full overview with better readability.

Also depends a lot on the review of the actual permission so I'll leave it for now.


joeribekker:
Not finished yet and code is decaying.

PR #1143, opened 2021-12-31
Set signed assertions for DigiD to True

joeribekker:
The DigiD-eHerkenning library now supports this and according to the DigiD deployment docs it needs to be enabled.


sergei-maertens (approved):

SilviaAmAm (commented):
Looks good to me

PR #1154, opened 2022-01-04
Camunda integration: allow for strings with interpolated form fields

sergei-maertens:
Related to #326 

This is another step in the direction of allowing generic JSON logic expressions.

The UI is workable and an MVP approach - in the future, we can make this prettier using an editor-builder like Draft.js, but that was too much to get into. This replaces the existing prototype functionality where django templates can be used in the form of:

```django
Inzending

Contactpersoon: {{ contactName }}
```

which is used as values for JSON keys, for example.

TODO

- [x] Backend -> allow interpolate + interpret correctly (generic JSON logic evaluation)

### DIFF
--- a/src/openforms/js/compiled-lang/en.json
+++ b/src/openforms/js/compiled-lang/en.json
@@ -639,6 +639,12 @@
       "value": "Submissions will be deleted"
     }
   ],
+  "InHf6+": [
+    {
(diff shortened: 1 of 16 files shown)


joeribekker (approved):
Fijn die begrijpelijke tests!

Issue #1155, opened 2022-01-05
Show session timeout message


joeribekker:
The session cookie (`openforms_sessionid`) expire time is nicely set to the configured amount of time. However, after the time expired, it's not clear what is going on:

![image](https://user-images.githubusercontent.com/96970/148203563-0f18ceef-dc9a-4c20-8c9a-06d99ce0fe08.png)

```json
{
  "code": "permission_denied",
  "detail": "Je hebt geen toestemming om deze actie uit te voeren.",
  "instance": "urn:uuid:51cd6030-ff6e-46e3-aff3-8d258c2bc4ee",
  "status": 403,
  "title": "Je hebt geen toestemming om deze actie uit te voeren.",
  "type": "[...]/fouten/PermissionDenied/"
}
```

**Tasks**

* [x] Add popup that shows when the session has expired. When clicked OK, it redirects the user back to the start of the form (and login should happen again).

**Idea:**
We can do a "ping" to the server every minute that checks how much time left in the session (without refreshing the session). If the response is a 403 (I guess) the message is shown. This opens up the possibility to show a warning when there's less than 2 minutes left (which is not in scope of this ticket).

At the simplest level, we show the popup when any XHR-request returns a 403

sergei-maertens:
> We can do a "ping" to the server every minute that checks how much time left in the session (without refreshing the session)

ugh, don't like this, since this ping through the middleware will extend the session, unless you do view-specific exclusions in the middleware.

The cookie has an expiration date, but we'll probably not be able to access/read that from Javascript since it is (and should be) `httpOnly` :thinking:

sergei-maertens:
Better approach: update the middleware to include a response header with the timeout value, which the SDK reads and updates in local state after every API call. We can then have a component/alert that monitors this value and shows the appropriate error.

joeribekker:
> The cookie has an expiration date, but we'll probably not be able to access/read that from Javascript since it is (and should be)

Correct. Even when you don't do httpOnly, you cannot check it.

> Better approach: update the middleware to include a response header with the timeout value, which the SDK reads and updates in local state after every API call. We can then have a component/alert that monitors this value and shows the appropriate error.

Sounds good. I had that in mind as well but for some reason l thought JS could not read arbitrary HTTP headers and dismissed the idea. Google says I'm wrong :)

sergei-maertens:
[...]

PR #1156, opened 2022-01-05
Expose session expiry information

sergei-maertens:
#1155

Expose the session expiry as a response header on every (API) call. The frontend can process this header and calculate the timestamp of expiry relative to its own time. This will allow proper interpretation and handling of HTTP 403 responses due to expired sessions.

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -51,18 +51,27 @@ paths:
       responses:
         '204':
           description: No response body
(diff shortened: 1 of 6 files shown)


joeribekker (approved):
I was ready to comment that the header is also set for /admin but I think that might even be nice for a future feature that the admin shows a timeout message :)

PR #1161, opened 2022-01-06
Fix checkbox type is not correctly interpreted as boolean

### LINKED ISSUE #994: Logic: checkbox true conditions not working yet @SilviaAmAm
See https://github.com/open-formulieren/open-forms-sdk/pull/117#discussion_r756726120

**Tasks**

- [x] Fix checkbox-specific bug
- [x] Fix logic evaluation-on-initial-init

sergei-maertens:
Fixes #994 (partially)

Now checkbox logic rules are correctly using the boolean type rather
than a stringified boolean, which leads to correct evaluation in
the backend.

Additionally, some choices should have been stringified as they are
passed down to Select components and were failing prop type checks.


src/openforms/forms/models/form_definition.py
```diff
@@ -120,7 +120,7 @@ def delete(self, using=None, keep_parents=False):
         return super().delete(using=using, keep_parents=keep_parents)
 
     def iter_components(self, configuration=None, recursive=True):
-        from ..utils import iter_components
+        from openforms.formio.utils import iter_components
```
sergei-maertens, line 123: TODO: this import can probably be moved to the top of the module

src/openforms/submissions/form_logic.py
```diff
@@ -36,13 +38,29 @@ def evaluate_form_logic(
     step: "SubmissionStep",
     data: Dict[str, Any],
     dirty=False,
+    **context,
```
joeribekker, line 41: Is this (`**context`) broader than just "request" for a reason? Seems like request is the only thing that gets passed.

If `**context` is usefull, please document what it should/can contain. in this function.

joeribekker (changes requested):

src/openforms/submissions/form_logic.py
sergei-maertens: not really at this stage, though I figured I would just pass down the serializer context and it's up to the downstream functions to check if the context contains what it needs.

joeribekker (approved):

PR #1169, opened 2022-01-11
:arrow_up: Upgrade mozilla-django-oidc-db to 0.7.2

stevenbal:
permission checking causes a lot of backend initializations, which in turn causes a lot of get_solo() calls for the OIDC config, leading to a lot of Redis lookups

See issue in library: https://github.com/maykinmedia/mozilla-django-oidc-db/issues/30

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -243,7 +243,7 @@ maykin-django-two-factor-auth[phonenumbers]==2.0.4
     # via -r requirements/base.in
 git+https://github.com/maykinmedia/json-logic-py.git@1ceb93ad165d69a5639edc623d5613d0229c05e2#egg=maykin-json-logic-py
     # via -r requirements/base.in
-mozilla-django-oidc-db==0.6.0
+mozilla-django-oidc-db==0.7.2
     # via -r requirements/base.in
 mozilla-django-oidc==1.2.4
     # via mozilla-django-oidc-db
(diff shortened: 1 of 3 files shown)


stevenbal:
@joeribekker the large amount of cache calls still stands, do you think this would solve the issue? https://github.com/maykinmedia/mozilla-django-oidc-db/pull/29/commits/f76205997a073abcd49eab201ada540175763325

sergei-maertens (changes requested):
This won't work how you expect it to work.

LocMemCache is limited to the python process it's running in. So it will be shared by all threads in that process, but other python processes have their own cache instance.

What this means is that if you have deployed multiple workers (uwsgi processes/workers), and you change settings in the admin, only that process will have the cache busted but the other processes will still see the old cache. It gets even worse if you have multiple replicas (multiple app containers) with their own processes and memory state.

To fix this, you would have to restart uwsgi/the container(s).

stevenbal:
@sergei-maertens and what about https://github.com/maykinmedia/mozilla-django-oidc-db/commit/4e87575b1922b928a512954b34d4441f20d474be? This would prevent `get_solo()` from being called on `__init__` (which is what is causing 99% of the cache calls right now)

sergei-maertens:
@stevenbal not the most pretty solution, but more correct and it definitely has my preference to solve this in the library!

Another dirty workaround I have done before is caching the value on the `request` instance, so it's bound to the duration of the entire request and then discarded.

sergei-maertens:
@stevenbal see also https://github.com/maykinmedia/mozilla-django-oidc-db/commit/4e87575b1922b928a512954b34d4441f20d474be#r63496485. This would be our preferred approach

sergei-maertens (approved):

PR #1170, opened 2022-01-11
Configure Content-Security-Policy

### LINKED ISSUE #807: Analyse and fix CSP issues based on CSP reports
CSP and CSP reports was implemented in #791 

However, it's disabled by default. It should be enabled to analyse all hits and all hits should be solved.

**Tasks**

- [x] Ensure Formio.js machinery still works with unsafe-eval blocked
    - [x] Set up https://github.com/formio/protected-eval
    - [x] Pre-compile custom templates (done via ejs-loader in Webpack)
    - [x] Test all components
        - [x] File uploads
        - [x] Formio WYSIWYG content with inline styles
- [x] Ensure that nonces are added for inline styles
    - [x] Custom logo on form detail page
    - [x] Design tokens in master template
- [x] Fix specific components (https://github.com/open-formulieren/open-forms/issues/807#issuecomment-1019884345)
    - [x] Signature
    - [x] Maps (check)
    - [x] Columns (check)
    - [x] Free text (WYSIWYG & inline styles)
- [x] Add nonces to inline scripts
    - [x] Override `digid_herkenning/post_binding.html` template (5d90e00ce7b1b89115afde28614d92c74e7c91ab)
    - [x] Analytics scripts (bfb8c21b0665994c8efb22fb0816e17db6a7b2ba)
    - [x] Payments page script auto submit(bfb8c21b0665994c8efb22fb0816e17db6a7b2ba)

shea-maykinmedia:
Fixes #807 

**Changes**

- Configured the CSP
- Documented the CSP
- Made the necessary changes to not violate the CSP
    - Inline scripts received nonce
    - Pass CSP nonce to SDK init


joeribekker:
Updater from @shea-maykinmedia

Most of the issues in the front end are solved but a couple of components aren’t working as they should be.

**Signature**
```
[Report Only] Refused to apply inline style because it violates the following Content Security Policy 
directive: "default-src 'self' http://localhost:3000/ 'nonce-WKITyos8BFsoF29PVlbOkA=='". Either the 
'unsafe-inline' keyword, a hash ('sha256-7jPXitAmW5luCnpMQQGulV40bQXmdGjm0MJCccHWS8U='), 
or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event 
handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present. 
Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.
```
Possibly because it relies on an image?  Uses canvas https://www.w3schools.com/html/html5_canvas.asp as an html element

**Maps**
Doesn’t render the map, most likely because it relies on our custom template and custom templates are currently turned off

**Columns**
Doesn’t format properly, fields are one on top of another most likely because it relies on our custom template and custom templates are currently turned off

**Free Text (Vrijtekst)**
Just doesn’t render but I could not get the Free text component to render on master either so it may not be CSP related.  However I suspect this will have issues as well since I think it relies on TinyMCE and I don’t think TinyMCE is CSP compliant however this would still need to be looked into

src/openforms/conf/base.py
```diff
@@ -477,7 +478,9 @@
 RELEASE = config("RELEASE", GIT_SHA)
 
 # Base URL of where the SDK is hosted.
-SDK_BASE_URL = config("SDK_BASE_URL", "https://open-forms.test.maykin.opengem.nl/sdk")
+SDK_BASE_URL = config("SDK_BASE_URL", "https://open-forms.test.maykin.opengem.nl/sdk/")
+if not SDK_BASE_URL.endswith("/"):
+    base_url = f"{SDK_BASE_URL}/"
```
sergei-maertens, line 483: should fix `SDK_BASE_URL`, now missing trailing slashes still causes problems

src/openforms/conf/base.py
```diff
@@ -1009,8 +1014,8 @@
 #
 # Django CSP-report settings
 #
-CSP_REPORTS_SAVE = config("CSP_REPORTS_SAVE", False)  # save as model
-CSP_REPORTS_LOG = False  # logging
+CSP_REPORTS_SAVE = True  # save as model
+CSP_REPORTS_LOG = True  # logging
```
sergei-maertens, line 1018: should be envvars

src/openforms/forms/templates/forms/sdk_snippet.html
```diff
@@ -1,7 +1,7 @@
-{% load openforms %}{% if form %}{% with 'openforms-container' as div_id %}
+{% load openforms static %}{% if form %}{% with 'openforms-container' as div_id %}
```
sergei-maertens, line 1: I don't see any `{% static %}` usage?

sergei-maertens (changes requested):

src/openforms/conf/base.py
```diff
@@ -982,9 +986,12 @@
 # NOTE: make sure values are a tuple or list, and to quote special values like 'self'
 
 CSP_DEFAULT_SRC = [
-    "'self'"
+    "'self'",
+    SDK_BASE_URL,
 ]  # ideally we'd use BASE_URI but it'd have to be lazy or cause issues
 
+CSP_IMG_SRC = CSP_DEFAULT_SRC + ["data:", "https://service.pdok.nl/"]
```
joeribekker, line 993: Add comment what this is for.

sergei-maertens:
@joeribekker functionality wise, this is ready for review

docs/developers/csp.rst
```diff
@@ -0,0 +1,207 @@
...
+
+* The ``img-src`` requires ``https://service.pdok.nl/``, otherwise map components break.
+
+* The ``img-src`` requires ``data:``, otherwise map and signature components break.
+
+**CSP and WYSIWYG**
+
+Some form design relies on WYSIWYG editors, which results in HTML blobs in API endpoint
+resources that potentially have inline style (``<span style="...">...</span>``). The SDK
+uses React's ``dangerouslySetInnerHTML`` for this content.
```
joeribekker, line 91: Maybe hyperlink 
```suggestion
uses React's `dangerouslySetInnerHTML <https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml>`__ for this content.
```

docs/developers/sdk/embedding.rst
```diff
@@ -68,6 +76,14 @@ Available options
     The label of the Sentry environment to use, for example ``'production'``. Used in
     combination with ``sentryDSN``. Defaults to an empty string.
 
+Content Security Policy
+-----------------------
```
joeribekker, line 80:
```suggestion
Content Security Policy (CSP)
-----------------------------
```

docs/installation/config.rst
```diff
@@ -90,6 +90,25 @@ See: https://github.com/adamchainz/django-cors-headers
...
+* ``CSP_REPORTS_SAVE``: whether to save violation reports to the database. Defaults to
+  ``False``.
+
+* ``CSP_REPORTS_LOG``: whether to log violation reports. In containerized environments
+  this will be logged to ``stdout``, otherwise to file. If monitoring (Sentry) is set
+  up, reports will also be visible there through logging. Defaults to ``True``.
+
+* ``CSP_REPORT_PERCENTAGE``: float between 0 and 1.0, expressing the percentage of
+  violations that will be reported. Defaults to ``1.0`` (100%).
+
```
joeribekker, line 108: Should we make the list of allowed sources to be configurable? Now we only add PDOK but I can imagine that if an external plugin is created, you may want to add something here.

CSP_DEFAULT_SRC + CSP_IMG_SRC?

src/openforms/conf/base.py
```diff
@@ -995,25 +1005,31 @@
 # CSP_SANDBOX # too much
 
 CSP_UPGRADE_INSECURE_REQUESTS = False  # TODO enable on production?
-# CSP_INCLUDE_NONCE_IN = ["script-src"]  # if we inline we should at least have this
+
+CSP_EXCLUDE_URL_PREFIXES = (
+    # ReDoc/Swagger pull in external sources, so don't enforce CSP on API endpoints/documentation.
+    "/api/",
+    # FIXME: Admin pulls in bootstrap from CDN & has inline styles/scripts probably
```
joeribekker, line 1012: Will this cause a gazillion Sentry issues?

joeribekker (approved):
Small things, can be done in this PR or in new PR

src/openforms/conf/base.py
sergei-maertens: no, because CSP is essentially disabled for the admin. No headers are emitted at all for the admin, so there's nothing to report.

docs/installation/config.rst
sergei-maertens: Adding them as `CSP_EXTRA_FOO` settings

Issue #1174, opened 2022-01-11
Test results latest build


sergei-maertens:
Tested against version 5a8a30a35254ffedd1c9934654d911782eb53b93.

* Test form: `demo-sprint-11-testing` on the Maykin test env
* Admin interface in regular browser window
* Filling out the form in incognito window, Brave (Chromium 96) 

---

**Form designer**

- [x] BUG: Configuring appointments: added a phone number field in a different step, re-used a generic appointment fields step -> cannot configure the phone number component (dropdown does not change state) @sergei-maertens (fixed in 1ed8b77e24a73dca40b0a5ef2fa383327f025c88)
- [ ] #1190
- [ ] #1191
    - manual formatting (newlines, indentation) is reset when valid JSON is detected which outputs it all on a single line
    - font not being code/monospace makes it hard to read/edit the JSON
    - cursor jumps around once valid JSON is entered
    - basically currently the best way is to define the JSON in a code editor and then copy paste it into the form field
- [x] BUG: the complex JSON logic uses validation for simple logic at the moment, effectively making `and` and combined logic rules impossible to use. Caused by refactor with pricing/form logic & added validators. (fixed in a78d6b9fb0148a283ade0ab2e69a7b15038e21c0)

**Filling out the form**

- [x] BUG: Logging in with DigiD (login optional) brings the SDK into an infinite loading loop. Unsure what causes it, might be pre-fill related? @sergei-maertens (fixed in open-formulieren/open-forms-sdk#142)
    - It works when starting the form as anonymous user (!)
- [x] BUG: Multiple inputs (textfield) are broken (again) because of immutable state. The `+ add another` button click triggers this (likely fixed by https://github.com/open-formulieren/open-forms-sdk/pull/136) fix: open-formulieren/open-forms-sdk#142
- [x] BUG: Fieldset show/hide based on checkbox with default value `true` behaved a bit weird, but may be because of invalid logic configuration. To double check.
- [x] BUG: the WYSIWYG editor for the confirmation page template correctly allows arbitrary hyperlinks (filtering should only be done for templates going to e-mails w/r to phishing protection), however when the SDK renders this the URL appears to have been filtered out. Possibly some sanitation happens that shouldn't, OR the validator should be slapped on that field instead so users know to add it to the netloc allow list. @Bartvaderkin (fixed by https://github.com/open-formulieren/open-forms/pull/1177/commits/190c09c065852ddd02a2862d0768dba302dba118)
[...]

PR #1175, opened 2022-01-11
Backend bug fixes for sprint 11 demo

sergei-maertens:
#1174

* Fixed advanced logic validation
* Added more tests for advanced/simple logic rule validation
* Fixed appointments configuration tab in form designer


sergei-maertens:
Some tests are still failing, will look into that tomorrow

Bartvaderkin (changes requested):
I missing some new test code that fails the modified validator.

PR #1177, opened 2022-01-12
[#1174] Added SubmissionStepInline form_step to raw_id_fields

Bartvaderkin:


src/openforms/submissions/tests/test_submission_confirmation.py
```diff
@@ -90,3 +90,16 @@ def test_template_tags(self):
             "<p>some product information<p/>",
             confirmation_page_content,
         )
+
+    def test_content_not_sanitized(self):
+        config = GlobalConfiguration.get_solo()
+        config.submission_confirmation_template = "See http://example.com/bla"
+        config.save()
```
sergei-maertens, line 97: we switched patterns to use `unittest.mock` for the `get_solo` call, returning an unsaved `GlobalConfiguration` instance - check some other tests for this. That reduces the potential for breaking test isolation, since django-solo cache is enabled normally.

sergei-maertens (approved):
I see more tests in this module have the same pattern, let's clean that up in a separate PR/commit

PR #1188, opened 2022-01-13
Update NLX sample integration for local development

sergei-maertens:
Updated to the latest try-me git submodule & adapted our overrides.


sergei-maertens:
Deployed and working on test environment, verified locally too.

PR #1209, opened 2022-01-19
Date prefill

### LINKED ISSUE #1213: Datefields cannot be properly prefilled with dates
* [x] In the form builder in the admin, the date field should also have a tab to configure a prefill
* [x] When prefill values are processed, there should be a check to see if the field is a date field. If it is, then the date is formatted to "YYYY-MM-DD" (expected by the json-logic-py package).

SilviaAmAm:
Fixes #1213

Changes:
- In the form builder in the admin, the date field also has a tab to configure a prefill
- When prefill values are processed, there is a check to see if the field is a date field. If it is, then the date is formatted to "YYYY-MM-DD" (expected by the json-logic-py package).


joeribekker (changes requested):
Can you rebase the commit messages to include ticket https://github.com/open-formulieren/open-forms/issues/1213, otherwise, looks good

joeribekker (approved):

PR #1211, opened 2022-01-19
Make eIDAS char field longer

### LINKED ISSUE #1214: eIDAS pseudo ID cannot be properly stored
The documentation of eIDAS says that this field is a '32 bytes hex' https://afsprakenstelsel.etoegang.nl/display/as/EntityConcernedID%3APseudo

The field currently has a length of 9, and should be increased.

SilviaAmAm:
Fixes #1214 

The problem was that on the `Submission`, the field that contains the `Pseudo` id returned by eIDAS had a max_length of 9.

The documentation of eIDAS says that this field is a '32 bytes hex' https://afsprakenstelsel.etoegang.nl/display/as/EntityConcernedID%3APseudo


joeribekker (changes requested):
Can ou rebase commit messages to refer to issue #1214 ?

Otherwise looks good.

joeribekker (approved):
Just to clarify: 1 byte is represented in hex as a 2 character string, hence 64 and not 32 chars are needed.

PR #1231, opened 2022-01-21
[#1207] Cache prefill plugin values

### LINKED ISSUE #1207: Do not perform prefill actions upon every logic check
Our logic check is performed quite often when filling in a form.

When prefilling is configured, this can cause alot of calls to the StUF-BG-service or KvK API. There's not reason for this so we should prevent this.

**Tasks**

* [x] Refactor prefill to only be executed once (note: this differs from autocomplete features like addresses based on postalcode/number)

Bartvaderkin:
Fixes #1207


joeribekker (commented):
I had to checkout the use of nonlocal here but all seems to work. I had expected a `mock.assertNumCalls` or something. Leaving this for @sergei-maertens to have an opinion about but I think the general concept is good.

src/openforms/prefill/__init__.py
```diff
@@ -76,6 +78,75 @@ def apply_prefill(configuration: JSONObject, submission: "Submission", register=
...
+        # set None for fields not returned by plugin to block endless re-requesting
+        for plugin_id, fields in fetch_fields.items():
+            for field in fields:
+                path = Path(plugin_id, field)
+                try:
+                    glom(results, path)
+                except GlomError:
+                    assign(results, path, None, missing=dict)
+                    assign(cached, path, None, missing=dict)
+                    cache_dirty = True
```
sergei-maertens, line 137: This is too naive I think, the type of the data should match the type of the component.

Looking at some Formio configurations, I can see `defaultValue: ""` for text-based components. The pre-fill machinery with a pre-fill match would change that into `defaultValue: null`, which is not correct.

I'm okay with storing `None` in the DB (JSONField) as that is semantically correct, but we need to process that when setting the values so that the `defaultValue` is untouched if the result is `None`.

Possibly the implementation is correct, and the tests are set up wrong :-) so please check what happens after you fix the tests first.

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -299,3 +299,264 @@ def test_prefill_invalid_date(self, m_get_prefill_value):
...
+            "display": "form",
+            "components": [
+                {
+                    "id": "one",
+                    "type": "text",
+                    "prefill": {
+                        "plugin": "have",
+                        "attribute": "hit",
+                    },
+                    "defaultValue": None,
```
sergei-maertens, line 352:
```suggestion
                    "defaultValue": "",
```

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -299,3 +299,264 @@ def test_prefill_invalid_date(self, m_get_prefill_value):
...
+                },
+                {
+                    "id": "two",
+                    "type": "text",
+                    "prefill": {
+                        "plugin": "have",
+                        "attribute": "miss",
+                    },
+                    "defaultValue": None,
```
sergei-maertens, line 361:
```suggestion
                    "defaultValue": "",
```

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -299,3 +299,264 @@ def test_prefill_invalid_date(self, m_get_prefill_value):
...
+            configuration=form_step.form_definition.configuration,
+            submission=submission,
+            register=register,
+        )
+        self.assertEqual(counter, 1)
+
+        field = new_configuration["components"][0]
+        self.assertEqual(field["defaultValue"], "foo")
+        field = new_configuration["components"][1]
+        self.assertEqual(field["defaultValue"], None)
```
sergei-maertens, line 391:
```suggestion
        self.assertEqual(field["defaultValue"], "")
```

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -299,3 +299,264 @@ def test_prefill_invalid_date(self, m_get_prefill_value):
...
+            submission=submission,
+            register=register,
+        )
+        # still just one
+        self.assertEqual(counter, 1)
+
+        field = new_configuration["components"][0]
+        self.assertEqual(field["defaultValue"], "foo")
+        field = new_configuration["components"][1]
+        self.assertEqual(field["defaultValue"], None)
```
sergei-maertens, line 405:
```suggestion
        self.assertEqual(field["defaultValue"], "")
```

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -299,3 +299,264 @@ def test_prefill_invalid_date(self, m_get_prefill_value):
...
+
+    def test_apply_prefill_caches_values_across_steps_and_plugins(self):
+        """
+        similar to basic test except we check with multiple plugins, step and recurring prefills
```
sergei-maertens, line 409: Please update the `defaultValue`s too here for a realistic test case

sergei-maertens (changes requested):
The `prefill_data` data field should also be wiped/cleared as part of `Submission.remove_sensitive_data`.

sergei-maertens:
Also yeah, let me explicitly mention that this is a neat solution and neat code in general :)

sergei-maertens (approved):

PR #1234, opened 2022-01-21
[#1208] Implemented dynamic file upload urls in form configurations

### LINKED ISSUE #1208: Export includes upload domain, this should not be in there and dynamically set
Exporting a form contains the upload URL. This is required by FormIO but this makes sharing forms a little bit leaky AND changing the domain name basically invalidates the FormIO configuration.

**Tasks**

* [ ] Remove FormIO upload URL from definitions (or replace with: `https://www.example.com`) upon export
* [ ] Dynamically use the configured domain instead of storing the domain hardcoded in JSON.
* [ ] Upon import, set the correct domain (not needed when the above works?)

Bartvaderkin:
Fixes #1208

### DIFF
--- a/src/openforms/js/components/form/file.js
+++ b/src/openforms/js/components/form/file.js
@@ -204,7 +204,7 @@ class FileField extends BaseFileField {
             label: 'File Upload',
             key: 'file',
             storage: 'url',
-            url: getFullyQualifiedUrl('/api/v1/submissions/files/upload'),
+            url: "", // backend sets this
             options: "{\"withCredentials\": true}",
             webcam: false,
             input: true,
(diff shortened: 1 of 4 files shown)


joeribekker (approved):

src/openforms/js/components/form/file.js
```diff
@@ -204,7 +204,7 @@ class FileField extends BaseFileField {
             label: 'File Upload',
             key: 'file',
             storage: 'url',
-            url: getFullyQualifiedUrl('/api/v1/submissions/files/upload'),
```
joeribekker, line 207: This import could be removed as well but cleanup issue.

Issue #105, opened 2020-10-12
As developer I want an integration test to match backend and frontend code/api


joeribekker:
At the minimum, there needs to a complete flow test with a simple form. So, having a form, start it, fill it in, submit it, and make sure the submission succeeded.

sergei-maertens:
See PR remarks - we're not going to invest any more resources into this. The current flow with a dedicated tester seems to work well enough.

PR #1238, opened 2022-01-24
:construction_worker: Cypress integration tests

stevenbal:
Fixes: #105 

We could use the dashboard that Cypress provides for more insight and debugging possibilities, but it is subscription based.

Can be run locally with `node_modules/.bin/cypress run` or `node_modules/.bin/cypress open` if you want to see the browser.


stevenbal:
@sergei-maertens the fixtures for forms right now are rather big, would it be good to define the fixtures in a similar way as in the regular Django tests and then somehow take this configuration through the factories?

cypress/support/commands.js
```diff
@@ -0,0 +1,46 @@
...
+})
+
+// Find a form field by its key
+Cypress.Commands.add("formField", (fieldKey, timeout=10000) => {
+    return cy.get(`.openforms-form-control--${fieldKey}`, {timeout: timeout});
+})
+
+// Find the continue button
+Cypress.Commands.add("continueButton", (timeout=10000) => {
+    return cy.get('button[name="next"][type="submit"]', {timeout: timeout});
```
stevenbal, line 40: I guess it would be better if these elements have data attributes to find them, that way we don't have to rely on these kind of CSS selectors

sergei-maertens:
> the fixtures for forms right now are rather big, would it be good to define the fixtures in a similar way as in the regular Django tests and then somehow take this configuration through the factories?

a management command using the factories that does something like `src/manage.py generate_cypres_fixtures` wouldn't be that crazy :)

cypress/support/commands.js
sergei-maertens: I think I saw the NLX team use test IDs but I'm not sure

stevenbal:
> a management command using the factories that does something like src/manage.py generate_cypres_fixtures wouldn't be that crazy :)

Is the best approach here to load some configuration like this https://github.com/open-formulieren/open-forms/blob/3ed68803834d5c499a69f8930e678c5a6e9a1793/cypress/data_fixtures/definitions/data.json into the database and then dumping it to create the full fixture? I tried loading it into the container directly with a command, but we don't have `factory` installed in docker obviously :p

stevenbal:
Not sure if this was caused by rebasing, but now I seem to be running into issues because of DOM rerendering: https://github.com/cypress-io/cypress/issues/7306

sergei-maertens:
latest SDK patches have seen some changes to the way we interact with Formio, and Formio does mutate the DOM quite agressively, so that might be it?

stevenbal:
> latest SDK patches have seen some changes to the way we interact with Formio, and Formio does mutate the DOM quite agressively, so that might be it?

@sergei-maertens it does seem so. I'm not entirely sure how to fix this, because it does seem to be an unsolved problem (judging from the Cypress issue I linked). For `.click()` we could use `{force: true}`, but this doesn't seem to work for `.type()`, which is where we're running into problems now.

I'll check if adding more checks before trying to fill in textfields helps:

`You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.`

sergei-maertens:
We've discussed this and determined Cypress is not a viable option for our testing needs, unfortunately. Closing this PR to reduce the noise.

PR #1239, opened 2022-01-24
[#1127] Configuration hidden fields

### LINKED ISSUE #1127: Hidden fields are not part of the form and are not submitted
When you create a TextField which is hidden, it seems that the rendered HTML (nor the React state) does not include this field (and its possibly prefilled value) in the submit step.

We tried several things, like adding another field, causing a change event for that other field, but neither lead to the inclusion of the hidden field.

SilviaAmAm:
Fixes #1127 

Add the field `clearOnHide` to the configurations in the builder, so that the value of hidden fields can be included in the submission.

### DIFF
--- a/src/openforms/js/components/form/fieldset.js
+++ b/src/openforms/js/components/form/fieldset.js
@@ -21,7 +21,13 @@ const FIELDSET_BASIC = {
             key: 'hidden',
             label: 'Hidden',
             tooltip: 'A hidden field is still a part of the form, but is hidden from view.'
-        }
+        },
+        {
+            type: 'checkbox',
+            key: 'clearOnHide',
+            label: 'Clear on hide',
+            tooltip: 'Remove the value of this field from the submission if it is hidden',
+        },
     ]
 };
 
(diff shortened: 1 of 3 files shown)


joeribekker (changes requested):
So, the default of this `clearOnHide`  is `True`, right? Which means by default it doesn;t solve the issue itself.

I think we need to make this also clear in the tooltip!

src/openforms/js/components/form/fieldset.js
```diff
@@ -21,7 +21,13 @@ const FIELDSET_BASIC = {
...
             label: 'Hidden',
             tooltip: 'A hidden field is still a part of the form, but is hidden from view.'
-        }
+        },
+        {
+            type: 'checkbox',
+            key: 'clearOnHide',
+            label: 'Clear on hide',
+            tooltip: 'Remove the value of this field from the submission if it is hidden',
+        },
```
sergei-maertens, line 30: a fieldset doesn't have a value right, but then I'd assume the entire nested object with values is cleared? Can you confirm that?

SilviaAmAm: If you have a hidden fieldset then all the values of the children don't appear in the submission

sergei-maertens (approved):

PR #1245, opened 2022-01-26
[#1243] make onComponentMutated optional

### LINKED ISSUE #1243: Form definition configuration can't be added
When trying to add a form definition through `/admin/forms/formdefinition/add/`

This error appears in the console:

```
Uncaught TypeError: onComponentMutated is undefined
```

SilviaAmAm:
Fixes #1243

### DIFF
--- a/src/openforms/js/components/formio_builder/builder.js
+++ b/src/openforms/js/components/formio_builder/builder.js
@@ -92,13 +92,23 @@ const FormIOBuilder = ({ configuration, onChange, onComponentMutated, forceUpdat
         }
     );
 
+    if (onComponentMutated) {
+        return (
+            <FormBuilder
+                form={formRef.current}
+                options={BUILDER_OPTIONS}
+                onChange={formSchema => onChange(cloneDeep(formSchema))}
+                onSaveComponent={onComponentMutated.bind(null, 'changed')}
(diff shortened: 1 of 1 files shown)


src/openforms/js/components/formio_builder/builder.js
```diff
@@ -92,13 +92,23 @@ const FormIOBuilder = ({ configuration, onChange, onComponentMutated, forceUpdat
         }
     );
 
+    if (onComponentMutated) {
```
joeribekker, line 95: But, why wouldn't it be passed to the constructor? This seems like a workaround on a deeper issue?

SilviaAmAm: Because this component is used both in the Form view and in the FormDefinition view. What `onComponentMutate seem` to do in the case of the Form view (see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/form-creation-form.js#L736 and https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/form-creation-form.js#L290) is to check whether changes are needed for the Registration component. 

But this check doesn't need to happen for FormDefinitions that are being edited independently of the form.

joeribekker (approved):

PR #1257, opened 2022-01-28
[#1255] Warnings in reusable steps for date fields

### LINKED ISSUE #1255: Date fields in reusable step cause warning of form editing
Taiga issue 369

Whenever there is a reusable `FormDefinition` in a form, if a user updates that `FormDefinition` then a warning appears above the step:

```
U bewerkt een bestaande set van formuliervelden! Deze wijziging heeft effect op 2 formulieren.
```

However, for `FormDefinition` that have a date field, this warning is present before a user has updated anything.

This seems to come from the property `time_24hr` in the `widget` of the configuration of the date field. 

The configuration is saved as `time_24hr`, but when fetched from the API it is converted to camelCase so becomes `time24hr`. When the `useDetectConfigurationChanged` hook in the frontend checks if the configuration has changed, the 'original' configuration has `time_24hr` while the 'new' has `time_24hr`.

This causes the warning.

SilviaAmAm:
Fixes #1255 

The `time_24hr` was not the only problem.

The following fields in the widget for the date:
```
disabledDates
disableWeekends
disableWeekdays
disableFunction
readOnly
submissionTimezone
timezone
```

were being initialised to `undefined` and for some reason their value is not saved in the backend. So then when updating a form, these fields are automatically re-added to the configuration and that causes the configuration to not match the saved configuration => which causes the warning.

Not sure what the best way to fix this is, I attempted something.

### DIFF
--- a/src/openforms/forms/api/parsers.py
+++ b/src/openforms/forms/api/parsers.py
@@ -1,4 +1,5 @@
 from djangorestframework_camel_case.parser import CamelCaseJSONParser
+from djangorestframework_camel_case.render import CamelCaseJSONRenderer
 
 from .drf_camel_case import FormCamelCaseMixin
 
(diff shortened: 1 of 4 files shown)


sergei-maertens (approved):
I can live with it ;)

PR #1260, opened 2022-01-28
[#1251] Use named directory for self_certifi/requests CA bundle instead of mkdtemp()

### LINKED ISSUE #1251: Temporary certificate storage is not cleaned
This can cause quite alot of files/folders in the Docker overlay!

See: https://github.com/open-formulieren/open-forms/blob/5d16511b719e2c85c03c9ba68860a5a312dc6864/src/openforms/setup.py#L45

Bartvaderkin:
Fixes #1251

### DIFF
--- a/src/openforms/setup.py
+++ b/src/openforms/setup.py
@@ -12,14 +12,17 @@ they are available for Django settings initialization.
 import logging
 import os
 import sys
-import tempfile
 import warnings
+from pathlib import Path
 
 from django.conf import settings
 from django.urls import reverse
 from django.utils.http import is_safe_url
 
 import defusedxml
+import portalocker
+import redis
+from django_redis import get_redis_connection
 from dotenv import load_dotenv
 from requests import Session
 from self_certifi import load_self_signed_certs as _load_self_signed_certs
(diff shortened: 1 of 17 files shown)


sergei-maertens (changes requested):
I'd approach this a bit differently:

Add a custom setting `SELF_CERTIFI_DIR`, with a default of `os.path.join(BASE_DIR, "certifi_ca_bundle")`, and use that as `target_dir`. That gives the option at deployment time to mount a volume and pass the path to the mounted volume, completely avoiding overlayfs shenanigans. If not provided, it writes to the container layer but that's acceptable with this patch.

I'd also use `os.makedirs` while allowing the dir to already exist. `shutil.rmtree` is not needed, as self-certifi will overwrite the file anyway if it already exists.

Then, to combat race conditions, `threading.Lock` is not viable.  I am not sure if that still works with `uwsgi` with multiple processes, but if multiple containers are operating on the same volume, they definitely will not see each other's locks. I'd look into using https://pypi.org/project/portalocker/ maybe - it also supports Redis which is relevant in our set up too.

src/openforms/setup.py
```diff
@@ -47,10 +48,14 @@ def load_self_signed_certs() -> None:
     if _certs_initialized:
         return
 
-    # create target directory for resulting combined certificate file
-    target_dir = tempfile.mkdtemp()
-    _load_self_signed_certs(target_dir)
-    _certs_initialized = True
+    lock = RedisLock("load_self_signed_certs")
```
sergei-maertens, line 51: This will definitely require some more arguments/options to connect to the correct redis host, especially in container environments. You can take a look at `settings.CACHES` in `base.py` to see how we fetch the redis host information.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1266, opened 2022-01-28
Update NUM_PROXIES default

SilviaAmAm:


sergei-maertens (changes requested):
The default of `1` is correct, however please update the dev settings with `os.environ.setdefault("NUM_PROXIES", "0")`

SilviaAmAm:
But now the default in base.py is `0` (in the docker.py it is `1`). Should I still add `os.environ.setdefault("NUM_PROXIES", "0")` to dev.py?

joeribekker:
In retrospect, I think the default should be 1 and only dev should be 0.

SilviaAmAm:
Okay, i will update both!

sergei-maertens (approved):

PR #1276, opened 2022-01-31
Ensure embedded SDK build is pinned to same release

sergei-maertens:
The SDK statics are embedded in the backend build for convenience, but the version should
match the version of the backend too.

COPY --from= does not allow build args inline, we can work around this by defining a separate
stage which does support build args.

Vaguele related to Taiga 380 which saw a deploy with pinned versions and incorrect paths.


sergei-maertens:
This is okay for a first pass, but we will have to flesh this out more as the backend version can diverge from the SDK version.

SDK should stick to the _API version_ instead.

PR #1279, opened 2022-01-31
[#1278] Make prefill consistent with haal centraal standard

### LINKED ISSUE #1278: Haal Centraal _embedded attribute in response
Den Haag Taiga 255

The response from Haal centraal contain the data but not under the `_embedded` attribute. 

The `_embedded` attribute is only used in the list endpoint `/ingeschrevenpersonen` but not in the detail endpoint `/ingeschrevenpersonen/{burgerservicenummer}`.

https://vng-realisatie.github.io/Haal-Centraal-BRP-bevragen/redoc#operation/GetIngeschrevenPersoon

SilviaAmAm:
Fixes #1278 

This will break using Open-Personen for the prefill until it is updated to also be consistent with the standard.


sergei-maertens (approved):
Open Personen does things wrong w/r to the operation ID anyway

Issue #1310, opened 2022-02-07
File upload limit configuration is not properly validated in the form designer.


sergei-maertens:
This is a Formio bug in essence.

https://github.com/formio/formio.js/blob/4.13.x/src/components/file/editForm/File.edit.file.js#L226 lets you configure the max file size as a string (probably because of the MB/GB suffixes etc.). This is validated client-side in `validateMaxSize` which calls https://github.com/formio/formio.js/blob/4.13.x/src/components/file/File.js#L546 - this naively calls `parseFloat` on the entered number.

So, if a form designer in NL specifies `10,5 MB` for example, this will break because the `,` is probably not properly parsed as a float. It'll return `10` which is sort of acceptable, but unexpected.

Came across this from #1193

Issue #1193, opened 2022-01-14
File uploads and file size issues


joeribekker:
**Tasks**

* [x] Set and document Nginx default upload limit for the specific upload endpoint (suggestion: 50 MB)
* [x] Check and/or fix Open Forms configured upload limit (if configured to 1 MB, check that you cannot upload a file of 2 MB)
* [x] Show Nginx configured upload limit in Open Forms for upload fields (maybe even validate that it doesn't exceed this value)
* [x] For registration e-mails, do **not** add files as attachment but instead provide URLs to download these from the Open Forms server.
* [x] E-mail attachments can only be downloaded if you provide a valid hash or whatever as query-param.
* [x] For the confirmation email, only the file name is shown and the files are **not** attached
* [x] Make sure attachments (the actual files!) are also deleted when submissions are deleted.
* [x] Properly style file upload errors:
   ![image](https://user-images.githubusercontent.com/5518550/152831153-f958f1fd-89cb-4401-b171-737a43a4565e.png)

PR #1311, opened 2022-02-07
File uploads and maximum file sizes, part 1

sergei-maertens:
Partially solves #1193

* Added custom setting to specify upper limit to upload size, pulled from the environment. It supports the notation used in nginx, SI system and MiB/KiB notation.
* The instance upload limit is documented in the API spec and form designer. No form designer validation is added, since it's pretty :poop: validation to begin with. See #1310 for that.
* Runtime: added tests to validate that tinkering with `CONTENT_LENGTH` header does not fly
* Runtime: reduced DOS attack vector surface by only parsing upload bodies that are less than or equal to the configured upper limit.
* Runtime: when processing temporary uploads, the file sizes are checked against the configured maximum upload sizes in Formio component and blocked if they exceed the file size (this means the client-side validation was bypassed).

Client side validation tested in Brave (Chromium based) and Firefox, confirmed to be working:

image


joeribekker (approved):
Just a document thing, not blocking.

Even though the client side validation is "poop", it's probably more user friendly for larger files. don't expect float values anyway but more like 10 MB or 1 MB, etc. Plus I heard someone say the fileuploader is broken with CSP-headers :p

sergei-maertens:
> Even though the client side validation is "poop", it's probably more user friendly for larger files. don't expect float values anyway but more like 10 MB or 1 MB

It already goes right "by accident" and by relying on Javascript quirks to parse `"10 "` and `"10"` both as the float `10`. But yeah, don't want to invest too much effort in that aspect right now.

Issue #1301, opened 2022-02-03
Test the new formio formatters


sergei-maertens:
As part of #1068, Steven implemented Formio formatters in a pluggable way (which is to be used in the PDFs and confirmation e-mail). This is configured behind a [feature flag](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/config/models.py#L350)

So far this functionality has not been properly tested, which this ticket is about:

* manual testing
* automated testing

The current tests contain json fixtures for Formio: https://github.com/open-formulieren/open-forms/tree/master/src/openforms/formio/formatters/tests/files. These are a bit duplicated from https://github.com/open-formulieren/open-forms/tree/master/src/openforms/registrations/contrib/camunda/tests/files. These fixtures have been created by following these steps:

1. In the form designer admin, create a form with a single form step
2. Add all possible components to the form step
    * Make sure to have a single value variant
    * Make sure to have a multiple values variant (if applicable for the component type)
    * (new) Include variants of components for every option that may alter the representation (for example number of decimals for number/currency fields)
3. Save the data & form
4. Get the JSON configuration from the form definition detail/configuration endpoint and save to JSON file for tests (`all_components.json`)
5. Fill out the from from the UI and complete it
6. Save the `SubmissionStep.data` JSON to `all_components_data.json`

Once the fixtures have been generated/updated, define/update formatter tests (see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/formatters/tests/test_default_formatters.py#L21) for the legacy behaviour which is supposed to be mostly correct. The Camunda contrib should use the same `formio` fixtures, so remove the duplicated files/fixtures there.

Once that is in place (and any possible bugs fixed, for example the signature representation is not right yet - see #1293 and #1179), update the test to enable the feature flag and ensure it passes too. Fix any formatter bugs that you discover.

Then, make sure to also manually verify that the confirmation e-mail and PDF are properly formatted.

**Tasks**

- [ ] Update/extend Formio fixtures for tests
- [ ] Ensure tests are expanded/updated and passing with the feature flag disabled
- [ ] De-duplicate the `openforms.registrations.camunda.contrib.tests.files` fixtures (use the `openforms.formio.formatters.tests.files` fixtures instead)
- [ ] Update the tests to enable the feature flag
    - [ ] Fix any Formio formatters bugs you encounter
- [ ] Manually test the formatting of all components
    - [ ] Confirmation PDF
    - [ ] Confirmation e-mail
[...]

PR #1312, opened 2022-02-08
Deal with file upload "attachments"

sergei-maertens:
Partially solves #1193

* End-user uploads are no longer added as attachment to the email
* Instead, HTML mails have a clickable link and plain text emails have a link that can be copy-pasted
* The link points to a "standard" django view to download a submission attachment, using sendfile to optimize file IO/serving
* Access control on the view consists of:
    * user must be logged in
    * user must be staff user (as the admin login page is the configured login page)
    * user must have the `submissions.view_submissionfileattachment` permission, included in the groups fixture with the "Behandelaars" group

As part of this PR, some more refactoring was done to the displaying of form values in e-mails, it now checks if the rendering is set to text mode or not (if no information is available, the current behaviour is kept and html is assumed). There's also a helper interface to format a value as text/html, which will probably be implemented on the `openforms.formio.formatters` in the future. This helper allows us to tackle our concrete problem without needing the entire refactor and having to wait on #1301

### DIFF
--- a/docs/configuration/registration/email.rst
+++ b/docs/configuration/registration/email.rst
@@ -14,3 +14,19 @@ building the form.
     
     Email should be configured by the Open Forms supplier and also requires
     proper setup of SPF and/or DMARC-records.
+
+File uploads/attachments
+------------------------
+
+Any files uploaded as part of the submission are not added as email attachments.
+Instead, the e-mail contains download links to the individual files. These links require
+users to be authenticated and have the correct permissions:
+
+* users must be staff users
+* users must have permission to read file attachments. The standard group "Behandelaars"
+  offers this permission.
+
+.. note::
+
+    Email attachments are typicall restricted in size, making it impossible to actually
+    deliver the emails for further processing.
--- a/src/openforms/submissions/views.py
+++ b/src/openforms/submissions/views.py
@@ -1,6 +1,7 @@
 import logging
 import uuid
 
+from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.exceptions import PermissionDenied
 from django.utils.crypto import constant_time_compare
 from django.utils.translation import gettext_lazy as _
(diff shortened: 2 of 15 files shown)


docs/configuration/registration/email.rst
```diff
@@ -14,3 +14,19 @@ building the form.
...
+
+File uploads/attachments
+------------------------
+
+Any files uploaded as part of the submission are not added as email attachments.
+Instead, the e-mail contains download links to the individual files. These links require
+users to be authenticated and have the correct permissions:
+
+* users must be staff users
+* users must have permission to read file attachments. The standard group "Behandelaars"
```
joeribekker, line 26: Can you then also add it to the actual fixture :)

src/openforms/submissions/views.py
```diff
@@ -116,3 +121,61 @@ def get_form_resume_url(self, submission: Submission) -> str:
...
+        if not hasattr(self, "_object"):
+            obj = super().get_object(queryset=queryset)
+
+            if not constant_time_compare(obj.content_hash, content_hash):
+                logger.warning(
+                    "Unauthorized file download attempted (invalid content hash). Path: %s, user: %r",
+                    self.request.path,
+                    self.request.user,
+                )
+                raise PermissionDenied(_("File access denied."))
```
joeribekker, line 168: Does this redirect the user to the login screen or give a 403?

joeribekker (changes requested):
Looks good!

docs/configuration/registration/email.rst
sergei-maertens: that is in there already, or am I missing something? https://github.com/open-formulieren/open-forms/blob/master/src/openforms/fixtures/default_groups.json#L1607

src/openforms/submissions/views.py
sergei-maertens: 403 - this is a codepath where people are tampering with hashes, so a harsher error is okay.

Additionally, the django `AccessMixin` only redirects if the user is not authenticated:

in `django.contrib.auth.mixins.AccessMixin.handle_no_permission`

```py
    def handle_no_permission(self):
        if self.raise_exception or self.request.user.is_authenticated:
            raise PermissionDenied(self.get_permission_denied_message())
```

sergei-maertens:
I've styled the 403 page as well:

image

Issue #1265, opened 2022-01-28
eHerkenning/DigiD service parameters should be provided via CLI not as setting


joeribekker:
We only need them to generate stuff, or not?

SilviaAmAm:
In progress: https://bitbucket.org/maykinmedia/django-digid-eherkenning/pull-requests/20/draft-improve-metadata-generation

PR #1314, opened 2022-02-09
Improve eHerkenning/DigiD settings

SilviaAmAm:
Part of https://github.com/open-formulieren/open-forms/issues/1265

Pull request for the CLI arguments is here: https://bitbucket.org/maykinmedia/django-digid-eherkenning/pull-requests/20

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -909,6 +909,11 @@ EHERKENNING = {
             "requested_attributes": [],
             "language": "nl",
         },
+    ],
+}
+
+if EIDAS_SERVICE_INDEX:
+    EHERKENNING["services"].append(
         {
             "attribute_consuming_service_index": EIDAS_SERVICE_INDEX,
             "service_loa": EHERKENNING_LOA,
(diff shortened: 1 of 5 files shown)


sergei-maertens (approved):
I know too little about this to make a proper judgement, but code-wise this looks okay.

joeribekker:
FYI: Before merging, I want to do a review :)

sergei-maertens:
I intended to leave this to you indeed :)

src/openforms/conf/base.py
```diff
@@ -929,38 +891,21 @@
...
-                "urn:etoegang:1.11:attribute-represented:KvKnr",
-            ],
-            "privacy_policy_url": {
-                "nl": PRIVACY_POLICY_URL,
-            },
-            "herkenningsmakelaars_id": EHERKENNING_MAKELAAR_ID,
+            "requested_attributes": [],
+            "language": "nl",
         },
         {
```
joeribekker, line 897: Can we add an IF-statement to add the eIDAS part below dynamically? For example:

```python
if EIDAS_SERVICE_INDEX:
  EHERKENNING["services"].append(....)
```

Or does this cause other issues?

joeribekker (changes requested):

src/openforms/conf/base.py
SilviaAmAm: I think it should be fine

joeribekker (approved):

PR #1315, opened 2022-02-09
[#1216] Return deleted forms to staff users

### LINKED ISSUE #1216: Deleted forms cannot be retrieved via the API, causing issues when rendering a deleted form in the admin
If authenticated, return deleted forms. If not, don't return them.

SilviaAmAm:
Fixes #1216

### DIFF
--- a/src/openforms/forms/api/viewsets.py
+++ b/src/openforms/forms/api/viewsets.py
@@ -237,7 +237,7 @@ class FormViewSet(viewsets.ModelViewSet):
 
     parser_classes = (FormCamelCaseJSONParser,)
     renderer_classes = (FormCamelCaseJSONRenderer,)
-    queryset = Form.objects.filter(_is_deleted=False).prefetch_related(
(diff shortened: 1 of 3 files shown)


src/openforms/forms/api/viewsets.py
```diff
@@ -254,7 +254,7 @@ def get_queryset(self):
         queryset = super().get_queryset()
         if not self.request.user.is_staff:
             # if not staff then only show active
-            queryset = queryset.filter(active=True)
+            queryset = queryset.filter(_is_deleted=False).filter(active=True)
```
sergei-maertens, line 257: I think we should only return deleted forms for the detail endpoint (`self.action == "retrieve"`) and not list endpoints (`self.action == "list"`).

Third parties (authenticated) can integrate displaying a dropdown of forms to embed in their CMS - you only want to see non-deleted forms there. And the admin uses the detail/retrieve endpoint and does not use the list endpoint.

sergei-maertens (changes requested):

PR #1317, opened 2022-02-09
Clean up submission attachment files on submission delete

sergei-maertens:
Partially solves #1193

* [x] Made it explicit that confirmation emails may not have attachments
* [x] Added test + ensured that the attachment files (on disk) are deleted if a submission is delete
* [x] Refactored file deletion code into generic, drop-in abstraction
* [x] Ensure that files that are marked as sensitive data are deleted if the submission is kept, but sensitive data is being cleared


src/openforms/utils/files.py
```diff
@@ -0,0 +1,127 @@
...
+
+LogLevel = Literal[
+    "debug",
+    "info",
+    "warning",
+    "error",
+    "exception",
+    "critical",
+    "fatal",
+]
```
joeribekker, line 23: Bit overenginered for just that very specific `log_failed_deleted` function?

If anything, I want the level depend on the LOGGER settings in Django.

src/openforms/utils/files.py
```diff
@@ -0,0 +1,127 @@
...
+    def _raw_delete(self, using):
+        # raw delete is called in fast_deletes
+        self._delete_filefields_storage()
+        return super()._raw_delete(using)
+
+    _raw_delete.alters_data = True
+
+    def delete(self):
+        self._delete_filefields_storage()
+        return super().delete()
```
joeribekker, line 124: So, yes, this should be in a `transaction.on_commit` but you commented why this might be problematic (altho non-transactional tests are possible).

Wouldn't a valid alternative be to just move the actual file deletion AFTER the record deletion? This way any record issue will prevent actuall file deletion. Something like:

```suggestion
        filefields = self._get_filefields()
        result = super().delete()
        self._delete_filefields(filefields)
        return result
```

joeribekker (changes requested):

src/openforms/utils/files.py
sergei-maertens: > on-transactional tests are possible
yes they are, but they're slow and a better alternative is the `capture_on_commit` context manager that executes them. however this PR is already large enough that for now I didn't want to bother with updating those broken tests. Maybe I should just do it though.

The order of operations was a bit of a catch 22, and I need to think about it now the context manager was added for robustness. Without that logger catching errors, you would always end up in an inconsistent state:

* records gone, but files still present (file delete errors)
* files gone, but records still present (query error)

src/openforms/utils/files.py
sergei-maertens: this is a typehint expressing the levels, it still complies to the `LOGGING` configuration (will use the `openforms` logger). But I will hardcode it to `warning` indeed.

src/openforms/utils/files.py
sergei-maertens: ok, managed to swap it around, as the capture-and-log-exceptions behaviour altered things quite a bit. I had to drop the `queryset.iterator()` because of this though, since the delete query used in the helper method comes up empty after the actual delete. I don't think it'll be a problem now in terms of memory usage, but I left a comment in case it ever becomes a problem (and then it might just be better to hook into the `pre_delete` signal).

Because of the swapping around it was easy to hook it into the `transaction.on_commit` behaviour as well, so I've made that improvement.

joeribekker (approved):

src/openforms/utils/files.py
sergei-maertens: correct! the context manager was added as one of the final steps, so it's behaviour was not taken into account for the order of operations yet.

PR #1326, opened 2022-02-10
Clean up frontend build stack

sergei-maertens:
Ripped out `gulp` and a bunch of unused dependencies.

One "big" thing left is to upgrade to Webpack 5 and then dependabot should leave us alone again for a bit.


SilviaAmAm (approved):
:fire:

PR #1329, opened 2022-02-14
Improve backend + SDK release process

sergei-maertens:
Relates to #1264

* Decoupled backend + SDK versions for Docker image build time
    * The SDK version to bundle is `latest` if the backend build is also `latest`
    * Otherwise on CI the value of the file `.sdk-release` is used
* Extended release process documentation w/r to stable versions and backports of bugfixes
* Documented backend image build dependencies and order of operations w/r to the sdk image build


joeribekker (commented):
Shouldn't we make it similar to a pip requirement?

`>=1.0.0-rc.1`

joeribekker (approved):

sergei-maertens:
> Shouldn't we make it similar to a pip requirement?

Unfortunately that's not how docker image tags work at all, only exact matches can be resolved. And I'm not looking to implement that resolution stuff myself :P

PR #1331, opened 2022-02-15
[#1330] Warn user about logic errors

### LINKED ISSUE #1296: Refactor admin edit warnings into messagelist component
`src/openforms/js/components/admin/form_design/AuthenticationWarning.js` is now the third component to define the messagelist markup again.

This should instead use a component `MessageList` with children in the form of `<Warning>...</Warning>` (or `<Error />`) to properly keep this composable and DRY.

### LINKED ISSUE #1330: Warn user if logic involves non existing component
There are multiple issues here:

1. If a user updates the key of a component and this key was used in a logic rule (both logic in the separate tab and 'simple' logic within the modal to edit a component) no warning is shown that now the logic rule is wrong.
2. The 'simple' logic doesn't block saving the form if a non existent component is specified
3. The logic rules in the separate tab of the admin do block the form from being saved if the logic trigger is not valid (but the error is not clearly shown).

SilviaAmAm:
Fixes #1330 
Fixes #1296

**Changes**:
- Added a hook that checks whether the components used in 'simple logic' rules exist and shows a warning if they don't.
- When the key of a component is changed, the logic rules in 'the other tab' are updated to use the new key. 
- Show error returned by the backend when a logic rule is invalid.

### DIFF
--- a/src/openforms/js/compiled-lang/en.json
+++ b/src/openforms/js/compiled-lang/en.json
@@ -265,6 +265,36 @@
       "value": "Type"
     }
   ],
+  "8hSjw4": [
+    {
+      "type": 0,
+      "value": "Component "
+    },
+    {
+      "type": 1,
+      "value": "label"
+    },
+    {
+      "type": 0,
+      "value": " ("
+    },
+    {
+      "type": 1,
+      "value": "key"
+    },
+    {
+      "type": 0,
+      "value": ") uses a non-existent component key "
+    },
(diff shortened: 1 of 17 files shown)


sergei-maertens (approved):
The JSON stringify/parsing for the logic is a bit nasty, but I also get why you're doing it. Could probably be cleaned up in the future using json-logic internals, but let's get this merged for now.

PR #1332, opened 2022-02-15
Upgrade to Django 3.2

sergei-maertens:
Let's see what the test suite thinks


sergei-maertens:
Updated the styles for django-admin-index which were pretty broken. I also noticed a lot of the CSS overrides can probably be replaced with design tokens, as this is now supported to some extent in Django.

PR #1359, opened 2022-02-21
Integrate SDK build into Django collectstatic

sergei-maertens:
Building on the decoupled `RELEASE` and `SDK_RELEASE` build args, we can now mix and match the SDK bundled into the Docker image in a more fail-safe way.

* The `RELEASE` build arg specifies the version of the Open Forms backend we're building and is an environment variable baked into the backend image.
* The `SDK_RELEASE` build arg specifies which image tag to copy the SDK build artifact from, and is baked into the image as an environment variable as well.

* The SDK build artifacts are included in the `src/openforms/static/sdk` path, where they will be picked up by Django's collectstatic
    * Collectstatic puts this in `/app/static/sdk/`
    * Collectstatic applies the content-hash for cache busting
    * Collectstatic also copies the file without hash, so third parties can still point to `${BASE_URL}/static/sdk/<version>/open-forms-sdk.js` if desired (though we'd recommend hosting the SDK image explicitly)
* Our own templates have been adapted to use the `{% static `sdk/<version>/open-forms-sdk.js' %}` tag instead of relying on `SDK_BASE_URL`. In production, we get the URLs with hashes.
* Docs updated for local dev environment on how to symlink to a static build of the SDK
* :boom: the `SDK_BASE_URL` envvar/setting has been removed, `SDK_RELEASE` has been introduced with the proper defaults.


sergei-maertens:
Tested this set-up with docker-compose locally and runserver.

docker-compose call:

```
TAG=1.0.0-rc.3 SDK_TAG=1.0.0-rc.2 docker-compose up --build
```

joeribekker (approved):

PR #1365, opened 2022-02-22
Make file-uploads-as-attachments a configurable option

### LINKED ISSUE #1363: Make file-uploads-as-attachment configurable on the registration backend.
Taiga 411

#1193 changed the behaviour for the e-mail registration backend to no longer include file-uploads as attachments (for good reasons!).

However, this poses problems for some users of Open Forms, so we should make this behaviour configurable. The default should be the current situation (no mail attachments), while the registration backend must get an opt-in to use attachments instead.

**Tasks**

- [x] Add e-mail registration backend option to attach file uploads (default False)
- [x] Ensure that the files are added as attachments in those situations
- [x] Set up global default + form-override (must be `NullBooleanField` to discern between False/unset)

sergei-maertens:
Fixes #1363 

* Updated registration options with a null boolean select (none / true / false)
* On None/missing value -> global config (default: false) is used
* On truthy config -> attach file uploads to email after all
* Updated documentation to mention the runtime configuration


src/openforms/js/components/admin/RJSFWrapper.js
```diff
@@ -71,14 +77,43 @@ const CustomCheckboxWidget = ({
...
+            />
+        );
+    }
+
     // The CustomCheckboxWidget is rendered as a children of the CustomFieldTemplate, which makes styling a bit trickier
     return (
         <div className="rjsf-field__checkbox">
             <label className="rjsf-field__label" htmlFor={id}>
-                {schema.title}
+                {label}
```
sergei-maertens, line 116: label is derived from `schema.title`, but may be overridden (it seems?)

src/openforms/scss/vendor/_rjsf-field.scss
```diff
@@ -34,7 +34,7 @@
     }
   }
 
-  &__label {
+  &__label:first-letter {
     text-transform: capitalize;
   }
```
sergei-maertens, line 39: Foo Bar Baz -> Foo bar baz

src/openforms/registrations/contrib/email/config.py
```diff
@@ -21,7 +21,19 @@ class EmailOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
...
     )
+    attach_files_to_email = serializers.BooleanField(
+        label=_("attach files to email"),
+        allow_null=True,
+        default=None,  # falls back to the global default
+        help_text=_(
+            "Enable to attach file uploads to the registration email. Note that this "
+            "is the global default which may be overridden per form. Form designers "
+            "should take special care to ensure that the total file upload sizes do "
+            "not exceed the email size limit."
```
SilviaAmAm, line 37: This is currently also the help_text that is shown within the form registration tab
Screenshot from 2022-02-22 15-30-47

src/openforms/js/components/admin/RJSFWrapper.js
```diff
@@ -20,15 +21,20 @@ const CustomFieldTemplate = ({
     hidden,
     required,
     displayLabel,
+    schema,
```
SilviaAmAm, line 24: Missing PropTypes update :)

src/openforms/js/components/admin/RJSFWrapper.js
```diff
@@ -20,15 +21,20 @@ const CustomFieldTemplate = ({
...
 }) => {
+    const isNullBoolean = isSelect(schema) && schema.type.includes('boolean');
```
SilviaAmAm, line 26: How does this differentiate between a 'normal' bool and a nullable bool? I see that for the `attachFilesToEmail` field the type is `['boolean', 'null']`

src/openforms/js/components/admin/RJSFWrapper.js
```diff
@@ -71,14 +77,43 @@ const CustomCheckboxWidget = ({
...
+                options={{...props.options, enumOptions}}
+                id={id}
+                schema={{default: 'null', ...schema}}
+                value={stringValue}
+                disabled={disabled}
+                readonly={readonly}
+                label={label}
+                autofocus={autofocus}
+                onChange={_onChange}
+            />
```
SilviaAmAm, line 108: I was wondering why you added this code specific to the SelectWidget inside of the Checkbox widget instead of customising the the rjsf select widget? https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/core/src/components/widgets/SelectWidget.js#L49

SilviaAmAm (changes requested):

src/openforms/js/components/admin/RJSFWrapper.js
sergei-maertens: the `isSelect(schema)` looks at the presence of `enum` to determine if it's a select or a "regular" field. And our json-schema generator only outputs the enum if `field.allow_null is True`

joeribekker (approved):
If I understand correctly, saying yes will ALSO add the files as attachment and always as download link

sergei-maertens:
> If I understand correctly, saying yes will ALSO add the files as attachment and always as download link

Correct - didn't feel like we needed to rip out that mechanism as that's quite a bit more work.

joeribekker:
Agree, this approved.

PR #1389, opened 2022-03-01
Permanently delete forms

### LINKED ISSUE #988: Implement permanently deleting forms
Currently, if you "delete" a form, it's only marked as deleted.

We should implement the beat task to periodically check for forms to be deleted, and maybe instead track the timestamp for delete marker so we can configure the leeway.

Follow up of #143.

Or, maybe we should just hide the deleted forms from the list view in the admin :thinking:

sergei-maertens:
Fixes #988

For both bulk (from list page) and detail (from edit page, currently not exposed in React):

* First time deleting a form just marks it as deleted and keeps the related steps/form definitions
* Second time deleting permanently removes the form from the database, cascading the form steps. Single-use form definitions are also cascade-deleted, form definitions that are marked as re-usable are kept.

### DIFF
--- a/src/openforms/forms/admin/form.py
+++ b/src/openforms/forms/admin/form.py
@@ -355,11 +355,10 @@ class FormAdmin(
         Split between soft and hard deletes here.
 
         The admin has mutually exclusive filters, but let's not rely on that assumption.
+        Hard deletes need to be performed first, otherwise non-deleted forms get
+        soft-deleted and in the next steps _all_ (including the just created) soft-
+        deletes get hard-deleted.
         """
-
-        # soft-deletes
-        queryset.filter(_is_deleted=False).update(_is_deleted=True)
-
         # hard deletes - ensure we cascade delete the single-use form definitions as well
         soft_deleted = queryset.filter(_is_deleted=True)
         fds = list(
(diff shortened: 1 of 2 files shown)


src/openforms/forms/admin/form.py
```diff
@@ -334,10 +334,38 @@ def remove_from_maintenance_mode(self, request, queryset):
...
+        Split between soft and hard deletes here.
+
+        The admin has mutually exclusive filters, but let's not rely on that assumption.
+        """
+
+        # soft-deletes
+        queryset.filter(_is_deleted=False).update(_is_deleted=True)
+
+        # hard deletes - ensure we cascade delete the single-use form definitions as well
+        soft_deleted = queryset.filter(_is_deleted=True)
```
joeribekker, line 364: Please switch soft and hard blocks around to prevent accidental soft->hard deletes in one go.

joeribekker (changes requested):

src/openforms/forms/admin/form.py
sergei-maertens: test added and patched

joeribekker (approved):

PR #1394, opened 2022-03-02
[#1393] Prefilled data check on submission

### LINKED ISSUE #1393: Prefilled data should not be modifiable
Taiga issue 423

- [x] Do check on submission complete that the data of prefilled fields matches the data coming from the prefill
- [x] Show prefilled fields as read only in the UI
- [x] Display errors in SDK in case of an invalid submission

SilviaAmAm:
Fixes #1393 

Frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/174

### DIFF
--- a/src/openforms/js/components/form/edit/tabs.js
+++ b/src/openforms/js/components/form/edit/tabs.js
@@ -31,6 +31,15 @@ const BASIC = {
         CLEAR_ON_HIDE,
         IS_SENSITIVE_DATA,
         DEFAULT_VALUE,
+        {
+            // This doesn't work as in native HTML forms. Marking a field as 'disabled' only makes it read-only in the
+            // UI, but the data is still sent to the backend.
+            type: 'checkbox',
+            label: 'Read only',
+            tooltip: 'Make this component read only',
+            key: 'disabled',
+            input: true
+        },
     ]
 };
 
--- a/src/openforms/js/components/form/edit/tabs.js
+++ b/src/openforms/js/components/form/edit/tabs.js
@@ -1,7 +1,6 @@
 import {Utils} from 'formiojs';
 
 import {getFullyQualifiedUrl} from '../../../utils/urls';
-
 import {
     LABEL,
     KEY,
@@ -12,6 +11,7 @@ import {
     CLEAR_ON_HIDE,
     IS_SENSITIVE_DATA,
     DEFAULT_VALUE,
+    READ_ONLY,
 } from './options';
 
 /**
@@ -31,15 +31,6 @@ const BASIC = {
         CLEAR_ON_HIDE,
         IS_SENSITIVE_DATA,
         DEFAULT_VALUE,
-        {
-            // This doesn't work as in native HTML forms. Marking a field as 'disabled' only makes it read-only in the
-            // UI, but the data is still sent to the backend.
-            type: 'checkbox',
-            label: 'Read only',
-            tooltip: 'Make this component read only',
-            key: 'disabled',
-            input: true
-        },
     ]
 };
 
@@ -60,6 +51,7 @@ const TEXT_BASIC = {
     label: 'Basic',
     components: [
         ...BASIC.components,
+        READ_ONLY,
         {
             weight: 100,
             type: 'textfield',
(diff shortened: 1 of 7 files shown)


src/openforms/js/components/form/edit/tabs.js
```diff
@@ -64,6 +64,15 @@ const BASIC = {
...
         },
+        {
+            // This doesn't work as in native HTML forms. Marking a field as 'disabled' only makes it read-only in the
+            // UI, but the data is still sent to the backend.
+            type: 'checkbox',
+            label: 'Read only',
+            tooltip: 'Make this component read only',
+            key: 'disabled',
+            input: true
+        },
```
sergei-maertens, line 75: As mentioned during standup - this is probably not applicable for all types of components and we should check them all to see if they make sense or not.

src/openforms/js/lang/formio/nl.json
```diff
@@ -290,5 +290,7 @@
   "Specify the size of each column. The sum of all the widths should be 100%.": "Geef de grootte van elke kolom op. De som van alle breedtes moet 100% zijn.",
   "Hidden component": "Verborgen component",
   "Hide fieldset header": "Verberg veldengroepheader",
-  "Hide the line and the label above the fieldset in the form": "Verberg de koptekst en de lijn boven de veldengroep in het formulier"
+  "Hide the line and the label above the fieldset in the form": "Verberg de koptekst en de lijn boven de veldengroep in het formulier",
+  "Read only": "Alleen lezen",
```
sergei-maertens, line 294:
```suggestion
  "Read only": "Alleen-lezen",
```

src/openforms/submissions/api/validation.py
```diff
@@ -78,6 +85,27 @@ def validate_submission_completion(
...
+    invalid_prefilled_fields = []
+
+    for component in submission_step.form_step.iter_components():
+        if "prefill" not in component or component["prefill"]["plugin"] == "":
+            continue
+
+        plugin_name = component["prefill"]["plugin"]
+        attribute_name = component["prefill"]["attribute"]
+        if (
+            submission.data[component["key"]]
```
sergei-maertens, line 106: technically this should be checking `submission_step.data`, but I presume you're doing this on purpose as you can alter data from another step currently? If so, please add a comment to explain this, at some point we'll properly add namespacing which will make this not required anymore.

src/openforms/submissions/api/validation.py
```diff
@@ -78,6 +85,27 @@ def validate_submission_completion(
...
+
+def get_invalid_prefilled_fields(
+    submission_step: "SubmissionStep", submission: "Submission"
+) -> List[str]:
+    invalid_prefilled_fields = []
+
+    for component in submission_step.form_step.iter_components():
+        if "prefill" not in component or component["prefill"]["plugin"] == "":
+            continue
+
```
sergei-maertens, line 102: Should this check also not only apply if the component is marked as disabled/readonly? If you expose the disabled/readonly controls to form-designers, this would lead to unexpected results otherwise where they may allow prefilled fields to be altered?

src/openforms/js/components/form/edit/tabs.js
SilviaAmAm: For the file upload, password, signature, co-signing and select family member field I think it doesn't make much sense to have the read only option. I shall remove it for these fields.

For the select box, the checkboxes, radio buttons it could be useful to have it, but the styling is not good yet. 

The map I am not sure if it makes sense to have it, but the styling is not working yet for it :stuck_out_tongue:

SilviaAmAm: In the end I think I will just add it to the `TEXT_BASIC` tab

src/openforms/submissions/api/validation.py
SilviaAmAm: Good point

src/openforms/submissions/api/validation.py
SilviaAmAm: Mhmm I hadn't thought about the fact that other steps could be modified :thinking:  I think it makes sense for now to just check `submission_step.data`

sergei-maertens (approved):

PR #1400, opened 2022-03-04
[#1398] Max file size as integer

### LINKED ISSUE #1398: Commas in file upload file size break validation
Taiga issue 430

If one gives the maximum file size as `10MB` the validation works. If one uses `10,0MB` the validation raises an error even for files that are smaller than 10MB.

In addition, the help text of the field suggests that one can use a comma (`50,0MB`):
Screenshot from 2022-03-04 10-09-10

(But specifying `10.0MB` doesn't break it)

SilviaAmAm:
Fixes #1398

### DIFF
--- a/src/openforms/utils/context_processors.py
+++ b/src/openforms/utils/context_processors.py
@@ -1,5 +1,7 @@
 from django.conf import settings as django_settings
-from django.template.defaultfilters import filesizeformat
+from django.utils import formats
+from django.utils.html import avoid_wrapping
+from django.utils.translation import gettext, ngettext
 
 
 def settings(request):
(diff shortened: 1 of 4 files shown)


src/openforms/utils/context_processors.py
```diff
@@ -22,8 +24,52 @@ def settings(request):
...
+
+
+def filesizeformat_integers(bytes_):
+    """
+    Format the value like a 'human-readable' file size but with integer numbers (i.e. 13 KB, 4 MB,
+    102 bytes, etc.).
+
+    This is adapted from django/template/defaultfilters.py:838 (filesizeformat)
+    """
+    try:
```
sergei-maertens, line 41: I'm not really a fan of all the code duplication for a single-character change. Maybe we can be a bit more pragmatic and do something like:

```py
django_formatted = filesizeformat(bytes_)
decimal_suffix = formats.number_format(0, 1)[1:]
return django_formatted.replace(decimal_suffix, "", 1)
```

also not particularly elegant, but more focused on the problem at hand.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1402, opened 2022-03-07
KVK prefill rework (again)

### LINKED ISSUE #1399: KvK prefill assumes that a certain type of company is returned, but not all companies return the same information
Taiga issue 384

We need to add more fields, like "naam", "statutaireNaam" that are in the API of the KvK. Currently, "eersteHandelsnaam" is a relic/custom logic to get the first entry from the array field "handelsnamen". While at it, we can add "Eigenaar" properties.

sergei-maertens:
Fixes #1399

The KVK exposes different types of 'companies' in its API, manifesting
in different blobs of information being returned (essentially the
resources are polymorphic without being explicitly modeled that way
in the API).

For example, a B.V. has a hoofdvestiging, but a VVE does not.

The endpoint consulted for company information is adapted to fetch
the base-profile instead of the hoofdvestiging (which may not exist
for certain types of companies). The response contains the hoofd-
vestiging (if it exists) in the _embedded container (per HAL spec).

It seems that either the hoofdvestiging or the 'eigenaar' embedded
resource contains the addresses, if any are present at all, so our
post-processing extracts those on a best-effort basis.

Tests have been adapted for the base-profile responses, and some
kvk.nl test-environment data added for the VVE-case.

The test to check access paths from the constants is removed, as this
was just a copy of the schema anyway and did not reflect a real-world
scenario where keys can be missing - none of the schema properties
are marked as required so we have no guarantees.


src/openforms/forms/migrations/0020_migrate_kvk_prefill.py
```diff
@@ -0,0 +1,66 @@
...
+                continue
+
+            # these are lifted from the API schema in a custom fashion and don't need the
+            # prefix, see the plugin code
+            if attribute.startswith("bezoekadres.") or attribute.startswith(
+                "correspondentieadres."
+            ):
+                continue
+
+            new_attribute = f"_embedded.hoofdVestiging.{attribute}"
```
joeribekker, line 31:
```suggestion
            new_attribute = f"_embedded.hoofdvestiging.{attribute}"
```
```suggestion
            new_attribute = f"_embedded.hoofdVestiging.{attribute}"
```

src/openforms/forms/migrations/0020_migrate_kvk_prefill.py
```diff
@@ -0,0 +1,66 @@
...
+        for component in iter_components(form_definition.configuration, recursive=True):
+            plugin_id = glom(component, "prefill.plugin", default=None)
+            attribute = glom(component, "prefill.attribute", default=None)
+            if plugin_id != "kvk-kvknumber":
+                continue
+
+            if not attribute:
+                continue
+
+            new_attribute = attribute.replace("_embedded.hoofdVestiging.", "", 1)
```
joeribekker, line 50:
```suggestion
            new_attribute = attribute.replace("_embedded.hoofdvestiging.", "", 1)
```
```suggestion
            new_attribute = attribute.replace("_embedded.hoofdVestiging.", "", 1)
```

joeribekker, line 50:
```suggestion
            new_attribute = attribute.replace("_embedded.hoofdvestiging.", "", 1)
```
```suggestion
            new_attribute = attribute.replace("_embedded.hoofdVestiging.", "", 1)
```

joeribekker (changes requested):
Works but the migration is wrong (took me a while)

I found one annoying issue: The search field in FormIO doesn't return enough hits. If you search for "embedded" it returned 1 entry with "hoofdvestiging" in it. But Searching for "hoofdvestiging" would return 4 entries with "embedded" in them.

I think this is unrelated however, so I'll create an issue for the latter (#1416 1416)

PR #1405, opened 2022-03-08
[#840] Updated django-hijack and fixed for OTP/2-factor

### LINKED ISSUE #840: Hijack werkt niet met 2FA
@Bartvaderkin has investiged this as part of https://github.com/open-formulieren/open-forms/issues/728#issuecomment-944300088 and it seems not trivial (which is good, but also unfortunate)

Bartvaderkin:
Fixes #840 

Note the 3.x dropped almost all settings.

Styled notification:

image

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -36,8 +36,7 @@ django-cookie-consent
 django-cors-headers
 django-decorator-include
 django-digid-eherkenning
-django-hijack ~= 2.3.0
-django-hijack-admin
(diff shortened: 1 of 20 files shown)


sergei-maertens (approved):

sergei-maertens (changes requested):
Can we add some tests for this?

sergei-maertens (approved):

PR #1409, opened 2022-03-08
[#1408] Hardened TimelineLogProxy against disappeared content object

### LINKED ISSUE #1408: AVG Timeline proxy 500 error
Taiga 437

Bartvaderkin:
Fixes #1408 

(there is no database constraint on generic foreign key)

### DIFF
--- a/src/openforms/logging/models.py
+++ b/src/openforms/logging/models.py
@@ -1,6 +1,5 @@
 from typing import List
 
-from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.template.defaultfilters import capfirst
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):
IDs by themselves (especially deleted) are not very useful information, can we instead record the `str(obj)` into the `TimelineLog` object when it's created? Then we at least have _some_ idea what the (audit) log entry was about.

See also how the django admin does this: https://github.com/django/django/blob/main/django/contrib/admin/models.py#L42 and https://github.com/django/django/blob/main/django/contrib/admin/models.py#L69

Discussed on Slack and decided to move this to a separate issue/PR: #1413

sergei-maertens (approved):

PR #1412, opened 2022-03-09
[#1404] Make the default required value configurable 

### LINKED ISSUE #1404: Configure whether fields should be required by default
Taiga issue 424

-    Runtime optie toevoegen aan configuratie om  default “verplicht” op false/true te zetten
-    Runtime optie toevoegen aan configuratie om asterisk of (optioneel) te tonen op form – Op form detail endpoint aangeven om asterisk of (optioneel) te gebruiken.

SilviaAmAm:
Fixes  #1404

Front end PR: https://github.com/open-formulieren/open-forms-sdk/pull/177


src/openforms/forms/admin/form.py
```diff
@@ -113,17 +113,28 @@ class FormAdmin(
     search_fields = ("name", "internal_name")
 
     change_list_template = "admin/forms/form/change_list.html"
+    _global_config = None
+
+    @property
+    def global_config(self):
+        if not self._global_config:
+            self._global_config = GlobalConfiguration.get_solo()
+        return self._global_config
```
sergei-maertens, line 122: ModelAdmin classes/instances are not thread safe, you should absolutely avoid sharing state on these instances.

The only places where it's acceptable if really needed is in methods that are called for every request, they typically have the `request` argument.

sergei-maertens (changes requested):

docs/configuration/general/styling.rst
```diff
@@ -97,6 +97,23 @@ The configuration in the admin requires this to be provided as JSON, for example
 
 .. _style dictionary: https://amzn.github.io/style-dictionary/
 
+**Style of required fields in forms**
```
sergei-maertens, line 100: I don't think this is the best place for this documentation, https://open-forms.readthedocs.io/en/latest/manual/forms/index.html is probably a better candidate

sergei-maertens (changes requested):
Implementation looks good :+1:

docs/configuration/general/styling.rst
SilviaAmAm: Since this other section is in Dutch, I translated it. I am not sure about the translation for the name of the new config model fields though... they are not super good names in English either :thinking:

sergei-maertens: thanks, we'll take a look at the translations before merging :)

PR #1417, opened 2022-03-10
Pinned header heights and fixed config overview template.

joeribekker:


sergei-maertens:
This should upgrade to django-admin-index 2.0.0 and build from there, see also the changes in default-project that are relevant.

sergei-maertens (approved):

PR #1419, opened 2022-03-10
[#1418] Implemented Submission.isAuthenticated for LogoutButton visibility

### LINKED ISSUE #1418: Implement Submission.is_authenticated and use for LogoutButton visibility in SDK
Taiga 426

Instead of using the loginRequired boolean on Form (which can be False if login is optional) to show LogoutButton in the SDK we add an is_authenticated getter on Submission and pass that through the API so the SDK can use that instead of loginRequired. The getter checks if Submission.auth_method is set.

Bartvaderkin:
Fixes #1418

SDK part: https://github.com/open-formulieren/open-forms-sdk/pull/178

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -4532,6 +4532,10 @@ components:
           readOnly: true
           description: Whether the user is allowed to submit this form and whether
             the user should see the overview page.
+        isAuthenticated:
+          type: boolean
+          readOnly: true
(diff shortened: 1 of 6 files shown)


sergei-maertens:
on second thought, this isn't actually breaking any behaviour so let's schedule it for 1.1.0 release

sergei-maertens (approved):

PR #1421, opened 2022-03-11
[#1076] Added regex to postcode component

### LINKED ISSUE #1076: Postal Code field does not apply validation
Issue raised by gemeente Den Haag (Taiga 178)

It seems that the postal code field does not apply validation. An incomplete value can be entered, and the formstep can still be submitted:

image

SilviaAmAm:
Fixes #1076 

Add regex to postcode component

Frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/161

### DIFF
--- a/src/openforms/js/components/form/postcode.js
+++ b/src/openforms/js/components/form/postcode.js
@@ -11,8 +11,11 @@ class PostcodeField extends TextField {
             inputMask: '9999 AA',
             validateOn: 'blur',
             validate: {
-              customMessage: 'Invalid Postcode',
-              pattern: '^[1-9][0-9]{3} ?(?!sa|sd|ss|SA|SD|SS)[a-zA-Z]{2}$',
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
I'm curious where this regex comes from - can you add a documentation link perhaps?

SilviaAmAm:
It comes from here: https://stackoverflow.com/a/17898538/7146757 :sweat_smile: Plus https://nl.wikipedia.org/wiki/Postcodes_in_Nederland to check that SS, SA and SD are really not used :sweat_smile: 
Should I add this anyway?

sergei-maertens:
Yeah, add them as comments. I had a hunch there was a reason for SA/SS/SD not being used because of the Nazi-association, but this might not be obvious to everyone.

SilviaAmAm:
@sergei-maertens Comment added :)

sergei-maertens (approved):

PR #1426, opened 2022-03-11
[#1422] Prevent update of custom keys on label changes

### LINKED ISSUE #1422: Changing values of radio buttons in Editor changes keys
Taiga issue 396

It looks like that when one first updates the key (eigenschapnaam) of a radio button and then updates the values, the key is reset to the default. This seems to happen in the following Formio function:

https://github.com/formio/formio.js/blob/4.13.x/src/WebformBuilder.js#L1368

SilviaAmAm:
Fixes #1422

### DIFF
--- a/src/openforms/js/components/formio_builder/WebformBuilder.js
+++ b/src/openforms/js/components/formio_builder/WebformBuilder.js
@@ -1,4 +1,5 @@
 import {Formio} from 'formiojs';
+import BuilderUtils from 'formiojs/utils/builder';
 import cloneDeep from 'lodash/cloneDeep';
 
 const WebformBuilderFormio = Formio.Builders.builders.webform;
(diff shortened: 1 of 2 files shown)


SilviaAmAm:
After discussion: its better to try and overwrite the whole callback and prevent it from updating all components in the editForm.

To import builders, try to import from the react-formio, otherwise directly from formiojs

SilviaAmAm:
Finally found a way to make this work: https://watch.screencastify.com/v/mNCjcVk3vnK2I1B9AWmk

src/openforms/js/components/formio_builder/WebformBuilder.js
```diff
@@ -33,6 +34,80 @@ class WebformBuilder extends WebformBuilderFormio {
...
+                        BuilderUtils.uniquify(formComponents, event.data);
+                    }
+                }
+            }
+
+            // Update the component.
+            this.updateComponent(event.data.componentJson || event.data, event.changed);
+        };
+
+        this.editForm.events._events['formio.change'][existingOnChangeHandlers.length-1].fn = modifiedOnChangeCallback;
```
sergei-maertens, line 108: The only thing I'm a bit worried about is that this is index-based to patch the library handler, but I don't think we have any better alternatives. It's something to keep in mind if/when we upgrade our Formio version.

sergei-maertens (approved):

PR #1429, opened 2022-03-14
[#1313] :sparkles: Add option to auto redirect to selected auth backend

### LINKED ISSUE #1313: Option to start form directly
Via Den Haag (Taiga 183) and others:

Most of the time, a CMS page is put before the actual form page. This requires the user to click twice to start the actual form (once on the start button in the CMS, and once the start button of Open Forms).

Since there can be different login options (!) we can not simply "start" the form automatically. We could however create the option to "redirect" to the proper (login)action automatically, based on a query parameter.

`/my-form?autostart=digid` (digid auth) or `/my-form?autostart=start` (no auth), etc.

The CMS button can then include this query parameter to skip the OF-start button(s).

stevenbal:
Fixes #1313 

Replacement for https://github.com/open-formulieren/open-forms/pull/1195 because merge conflicts were getting out of hand

related PR in SDK: https://github.com/open-formulieren/open-forms-sdk/pull/116

Changes:
* Adds a configuration option to forms, that enables automatic redirection to a selected authentication backend


src/openforms/authentication/fields.py
```diff
@@ -17,6 +17,35 @@ def __init__(self, *args, **kwargs):
         self.validators.append(UniqueValuesValidator())
 
 
+class AuthenticationBackendChoiceField(CharField):
```
sergei-maertens, line 20: Do we still need those form fields after removing the non-react admin change form? I think this entire class can be safely removed.

src/openforms/forms/api/serializers.py
```diff
@@ -153,6 +153,10 @@ class FormSerializer(serializers.ModelSerializer):
         default=list,
     )
     login_options = LoginOptionsReadOnlyField()
+    auto_login_authentication_backend = serializers.CharField(
```
sergei-maertens, line 156: I am missing validation, tests and documentation that this value must be one of the _selected_ options in the `authentication_backends` field.

src/openforms/js/components/admin/form_design/AuthPluginAutoLoginField.js
```diff
@@ -0,0 +1,66 @@
...
+            <FormattedMessage
+                description="Auth plugin provided attributes suffix"
+                defaultMessage="(provides {attrs})"
+                values={{attrs: plugin.providesAuth.join(', ')}}
+            />
+        );
+        const label = (
+            <>
+                {plugin.label}
+                {plugin.providesAuth.length ? (<>{' '}{providedAttributes}</>) : null}
```
sergei-maertens, line 21: Either we should refactor this duplication of code into a separate component, or not display it. I'm in favour of the last option, since the auth plugins with provided attributes are already displayed before.

src/openforms/js/components/admin/form_design/AuthPluginAutoLoginField.js
```diff
@@ -0,0 +1,66 @@
...
+
+AuthPluginAutoLoginField.propTypes = {
+    availableAuthPlugins: PropTypes.arrayOf(PropTypes.shape({
+        id: PropTypes.string,
+        label: PropTypes.string,
+        providesAuth: PropTypes.arrayOf(PropTypes.string)
+    })),
+    selectedAuthPlugin: PropTypes.string,
+    onChange: PropTypes.func,
+    required: PropTypes.bool,
```
sergei-maertens, line 59: this prop is not used?

src/openforms/authentication/fields.py
stevenbal: I think we can, though I'll have to check how to fix the existing migrations for DH

src/openforms/forms/api/serializers.py
stevenbal: I removed the `AuthenticationBackendChoiceField` and added the validation you mentioned to the serializer

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -618,6 +628,7 @@ const getFormData = async (formUuid, dispatch) => {
             type: 'FORM_LOADED',
             payload: {
                 selectedAuthPlugins: form.loginOptions.map((plugin, index) => plugin.identifier),
+                selectedAuthPluginAutoLogin: form.autoLoginAuthenticationBackend,
```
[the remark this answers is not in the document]
stevenbal: Hmm, I renamed `selectedAuthPluginAutoLogin` to `autoLoginAuthenticationBackend`, but for some reason when I remove this, changing the selection and saving the form doesn't seem to update the value anymore

src/openforms/authentication/fields.py
sergei-maertens: yeah there are no constraints added or anything, so you should be able to un-apply the migration. Faking should not be needed. You can check the docs in our internal wiki page, I linked it on Slack this week.

sergei-maertens:
This should be a radio button instead of a checkbox - checkbox implies that multiple values are possible, which is not the case
image

stevenbal:
> This should be a radio button instead of a checkbox - checkbox implies that multiple values are possible, which is not the case

I thought about this as well, but I was sort of conflicted, because radio buttons can usually not be unchecked right? While here you would like to have the option to keep it empty. Maybe an extra "emtpy" option should be added?

src/openforms/forms/api/serializers.py
```diff
@@ -368,6 +380,26 @@ def validate_backend_options(self, attrs, backend_field, options_field, registry
...
+                {
+                    field_name: ErrorDetail(
+                        _(
+                            "The `{auto_login_authentication_backend}` must be one of the selected backends "
+                            "from `authentication_backends`"
+                        ).format(auto_login_authentication_backend=field_name),
+                        code="invalid",
+                    )
+                }
+            )
```
sergei-maertens, line 401: This can still lead to inconsistent situations on PATCH calls. If you remove available backends but don't touch the `autologin` backend, this validation won't catch it. Instead, you should use `openforms.api.utils.get_from_serializer_data_or_instance` which will compare against the existing record for partial updates. Please also add a test case for this situation.

src/openforms/forms/api/serializers.py
```diff
@@ -153,6 +153,14 @@ class FormSerializer(serializers.ModelSerializer):
         default=list,
     )
     login_options = LoginOptionsReadOnlyField()
+    auto_login_authentication_backend = serializers.CharField(
+        required=False,
+        allow_blank=True,
+        help_text=_(
+            "The authentication backend to which the user will be automatically "
+            "redirected upon starting the form. The chosen backend must be present in `authentication_backends`"
+        ),
```
sergei-maertens, line 162:
```suggestion
        help_text=_(
            "The authentication backend to which the user will be automatically "
            "redirected upon starting the form. The chosen backend must be present in "
            "`authentication_backends`."
        ),
```
Please also keep the line-lengths in mind, black is configured for 88 columns.

src/openforms/forms/tests/test_api_forms.py
```diff
@@ -51,6 +51,56 @@ def test_get_fields_is_not_filtered_for_staff(self):
             with self.subTest(field=field):
                 self.assertIn(field, FormSerializer.Meta.fields)
 
+    def test_auto_login_authentication_backend_validation_fails(self):
+        request = RequestFactory().get("/")
```
sergei-maertens, line 55: please take the opportunity to fix this `RequestFactory` - it should be `rest_framwork.test.APIRequestFactory` since we're dealing with API calls and serializers.

I also vastly prefer having a proper end-to-end test instead of these `RequestFactory` specific tests, we care about the client `post`/`put`/`patch` behaviour, the implementation itself does not matter. So please convert to proper endpoint tests in `FormsAPITests`

src/openforms/js/components/admin/form_design/AuthPluginAutoLoginField.js
```diff
@@ -0,0 +1,51 @@
...
+import Field from '../forms/Field';
+import {Checkbox} from '../forms/Inputs';
+
+const AuthPluginAutoLoginField = ({availablePlugins, selectedPlugins, selectedPlugin, onChange, errors}) => {
+    let filteredPlugins = availablePlugins.filter(plugin => selectedPlugins.includes(plugin.id))
+    const authCheckboxes = filteredPlugins.map(plugin => {
+        return (
+            <li key={plugin.id}>
+                <Checkbox
+                    name={`autoLoginAuthenticationBackend.${plugin.id}`}
```
sergei-maertens, line 14: this form field `name` attribute must match with the respective field name in the API for change events to fire properly.

if `name="foo"`, then the event will be something like:

```js
event = {
    target: {
        name: 'foo',
        value: 'bar',
    },
}
```

That is what the entire state/reducer machinery relies on to update the local state and push changes to the backend.

src/openforms/js/components/admin/form_design/AuthPluginAutoLoginField.js
```diff
@@ -0,0 +1,51 @@
...
+
+import Field from '../forms/Field';
+import {Checkbox} from '../forms/Inputs';
+
+const AuthPluginAutoLoginField = ({availablePlugins, selectedPlugins, selectedPlugin, onChange, errors}) => {
+    let filteredPlugins = availablePlugins.filter(plugin => selectedPlugins.includes(plugin.id))
+    const authCheckboxes = filteredPlugins.map(plugin => {
+        return (
+            <li key={plugin.id}>
+                <Checkbox
```
sergei-maertens, line 13: should be a radio, see the screenshot I posted. And to be able to reset it to no value, there should be an empty radio option added with `value=""`.

src/openforms/js/components/admin/form_design/AuthPluginAutoLoginField.js
```diff
@@ -0,0 +1,51 @@
...
+                    onChange={onChange}
+                    checked={selectedPlugin === plugin.id}
+                />
+            </li>
+        );
+    });
+
+    return (
+        <Field
+            name="formAuthPluginAutoLogin"
```
sergei-maertens, line 26: the `Field` component forwards the `name` prop to its children, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/forms/Field.js#L29

However, this doesn't work properly here because of the nested `<ul>`. Instead, you should let `AuthPluginAutoLoginField` take the `name` prop and make that the direct child of the `Field` component, and that component should in turn render the `<ul>`.

The correct approach is:

```jsx
<Field name="foo" {...rest}>
    <AuthPluginAutoLoginFieldOptions {...relevantProps} />
</Field>
```

with 

```jsx
const AuthPluginAutoLoginFieldOptions = ({name, ...relevantProps}) => (
    <ul>
        {plugins.map(plugin => (
            <li>
                <Radio name={name} {...radioProps} />
            </li>
        ))}
    </ul>
);
```

or keep them together in a single component but then repeat/fix the `name` prop.

src/openforms/js/components/admin/form_design/FormMetaFields.js
```diff
@@ -134,6 +136,16 @@ const FormMetaFields = ({
...
 
+            <FormRow>
+                <AuthPluginAutoLoginField
+                    availablePlugins={availableAuthPlugins}
+                    selectedPlugins={selectedAuthPlugins}
+                    selectedPlugin={autoLoginAuthenticationBackend}
+                    onChange={onChange}
+                    name="form.autoLoginAuthenticationBackend"
+                />
+            </FormRow>
```
sergei-maertens, line 147: I'd actually rather see the `Field` component lifted from `AuthPluginAutoLoginField` to here, just like line 120-128. And then instead of `TextInput`, you use `AuthPluginAutoLoginField`. The `AuthPluginField` approach is old and not how we want to do/compose things, it's a bad example.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -182,6 +183,14 @@ function reducer(draft, action) {
...
                     break;
                 }
+                case 'autoLoginAuthenticationBackend': {
+                    if (draft.autoLoginAuthenticationBackend === value) {
+                        draft.autoLoginAuthenticationBackend = "";
+                    } else {
+                        draft.autoLoginAuthenticationBackend = value;
+                    }
+                    break;
+                }
```
sergei-maertens, line 193: as mentioned before, this should not be needed if you use the proper `name` attributes/props in your form elements.

sergei-maertens (changes requested):

src/openforms/scss/components/builder/_radio.scss
```diff
@@ -0,0 +1,3 @@
+.vRadioLabel {
+    padding: 0 0 0 5px !important;
```
stevenbal, line 2: Not sure if this is the right place, but I noticed the padding of labels for radio buttons was off

sergei-maertens: it's not, but I'll take a look why it's off and how the default django styles solve it. possibly there's some markup/classnames missing in the components :)

Issue #1414, opened 2022-03-10
Remove support for non-formio from form admin


Bartvaderkin:
Drop use of feature flag `enable_react_form` in the form definition admin and related tests.

PR #1440, opened 2022-03-16
Delete dead code on custom fields

sergei-maertens:
This was only relevant when building a Django form field for the
respective custom fields, but we don't render Django forms anymore
after removing the enable_react_form feature flag in #1414


Bartvaderkin (approved):
I forgot about this ugly monkeypatch situation.

PR #1446, opened 2022-03-17
[#1441] Submission specific logout endpoint 

### LINKED ISSUE #1441: Separate form from admin sessions
Testing with demo plugins is really annoying atm.

You need to be logged in as admin to use demo plugins. But, specifically with prefill, you need to logout to reset your session. This however, also logs you out as admin...

## Spec added after refinement

**Current mechanisms clarification and context**

* There is only one Django session, backed by the session cookie. This session used to track both admin user authentication (`request.user`) and end-users without Django user who are filling out forms.
* We have a middleware that sets the session expiry based on the type of user (admin user or anon): https://github.com/open-formulieren/open-forms/blob/master/src/openforms/middleware.py#L29
* "Form sessions" are tracked by storing the IDs of started submissions in the django session, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/utils.py#L42. This allows users to have multiple submissions for the same or different forms in their session, allowing them to use multiple tabs. The frontend currently restricts this to one "form session" per form ID, i.e. you can fill out multiple different forms at the same time in different browser tabs/windows using the same Django session cookie. This is because the form ID is mapped to the submission ID in the local storage so that hard refreshes don't break the user experience.
* On logout, all authentication plugin session data must be destroyed. Every plugin can have [their own isolated/dedicated session key](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/contrib/digid/plugin.py#L71) which is converted in the generic `FORM_AUTH_SESSION_KEY`, which in turn [is recorded on the `Submission` instance](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/signals.py#L41). This is now performed by deleting the entire session, which also destroys the "admin session", i.e. you are logged out from the admin as staff user.

**Changes to make**

* [x] Scope the SDK/UI "form session" logout 
    - Currently, this does a `DELETE /api/v1/authentication/session` which destroys the entire session
    - [x] Add and call the endpoint for "form session" management: ``DELETE /api/v1/authentication/:uuid/session` with `uuid` the UUID of the submission.
    - [x] On succesful call, [remove the submission ID from the session](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/utils.py#L49) and
    - [x] Make sure to implement the auth clean up
        - Delete the `FORM_AUTH_SESSION_KEY` from the session (it has been recorded in started submissions already anyway), see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/signals.py#L68
        - Call plugin-specific logout cleanup, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/api/views.py#L60
        - Call submission auth cleanup, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/api/views.py#L57
* [x] Check all plugins for plugin-specific cleanup and implement it in the `logout` method
    - See https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/base.py#L68
    - This should at the least remove any session keys used exclusively by the plugin (DigiD, eHerkenning)
* [x] Deprecate the `/api/v1/authentication/session` endpoint, this will be removed in API version 2.0.0. OpenAPI 3 / drf-spectacular has constructions for this.
* [x] Document the [version requirements](https://open-forms.readthedocs.io/en/latest/developers/versioning.html#open-forms-sdk) - SDK 1.1 will require API version 1.1+

This allows us to implement the required changes without introducing breaking API changes:

* We add a new API endpoint (submission-specific)
* We swap the SDK over to use the new API endpoint

In the end, this means the `request.session` will no longer be flushed on "form session" logout, yet the reference to the submission will be gone in the session causing the SDK to start a new form submission.

Bartvaderkin:
Fixes #1441

SDK: https://github.com/open-formulieren/open-forms-sdk/pull/180

### DIFF
--- a/src/openforms/authentication/api/views.py
+++ b/src/openforms/authentication/api/views.py
@@ -3,11 +3,17 @@ from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework import authentication, permissions, serializers, status
+from rest_framework.generics import GenericAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from openforms.submissions.utils import get_submissions_from_session
+from openforms.submissions.utils import (
+    get_submissions_from_session,
+    remove_submission_from_session,
+)
 
+from ...submissions.api.permissions import ActiveSubmissionPermission
+from ...submissions.models import Submission
 from ...utils.api.views import ListMixin
 from ..registry import register
 from .serializers import AuthPluginSerializer
@@ -61,4 +67,38 @@ class AuthenticationLogoutView(APIView):
             plugin.logout(request)
 
         request.session.flush()
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class SubmissionLogoutView(GenericAPIView):
+    authentication_classes = ()
+    permission_classes = (ActiveSubmissionPermission,)
+    serializer_class = (
+        serializers.Serializer
+    )  # just to shut up some warnings in drf-spectacular
+
+    queryset = Submission.objects.all()
+    lookup_field = "uuid"
+
+    @extend_schema(
+        summary=_("Delete session"),
+        description=_(
+            "Calling this endpoint will clear the current form and submission from the session."
+        ),
+    )
+    @transaction.atomic()
+    def delete(self, request, *args, **kwargs):
+        submission = self.get_object()
+
+        remove_submission_from_session(submission, request.session)
+
+        # submissions = get_submissions_from_session(request.session)
+        # for submission in submissions:
+        #     submission.hash_identifying_attributes()
+        #     submission.save()
+        #
+        # for plugin in register.iter_enabled_plugins():
+        #     plugin.logout(request)
+
+        # request.session.flush()
         return Response(status=status.HTTP_204_NO_CONTENT)
--- a/src/openforms/authentication/api/views.py
+++ b/src/openforms/authentication/api/views.py
@@ -53,6 +53,7 @@ class AuthenticationLogoutView(APIView):
         description=_(
             "Calling this endpoint will clear the current user session and delete the session cookie."
         ),
+        deprecated=True,  # replaced with the submission specific SubmissionLogoutView
     )
     @transaction.atomic()
     def delete(self, request, *args, **kwargs):
--- a/src/openforms/authentication/api/views.py
+++ b/src/openforms/authentication/api/views.py
@@ -103,5 +103,4 @@ class SubmissionLogoutView(GenericAPIView):
         for plugin in register.iter_enabled_plugins():
             plugin.logout(request)
 
-        # request.session.flush()
         return Response(status=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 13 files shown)


src/openforms/authentication/api/views.py
```diff
@@ -3,12 +3,19 @@
...
 from rest_framework.views import APIView
 
-from openforms.submissions.utils import get_submissions_from_session
+from openforms.submissions.utils import (
+    get_submissions_from_session,
+    remove_submission_from_session,
+)
 
+from ...submissions.api.permissions import ActiveSubmissionPermission
+from ...submissions.models import Submission
```
sergei-maertens, line 16: Please use the absolute imports here as we're crossing django app boundaries

src/openforms/authentication/api/views.py
```diff
@@ -46,6 +53,7 @@ class AuthenticationLogoutView(APIView):
         description=_(
             "Calling this endpoint will clear the current user session and delete the session cookie."
         ),
+        deprecated=True,  # replaced with the submission specific SubmissionLogoutView
```
sergei-maertens, line 56: Can you please also amend the description of the endpoint with a pointer to the new endpoint. Something like "this endpoint is deprecated, instead use the submission-specific endpoint."

src/openforms/authentication/api/views.py
```diff
@@ -62,3 +70,38 @@ def delete(self, request, *args, **kwargs):
...
+    serializer_class = (
+        serializers.Serializer
+    )  # just to shut up some warnings in drf-spectacular
+
+    queryset = Submission.objects.all()
+    lookup_field = "uuid"
+
+    @extend_schema(
+        operation_id="submission_session_destroy",
+        summary=_("Delete session"),
```
sergei-maertens, line 87:
```suggestion
        summary=_("Delete form session"),
```
technically it's a submission session, but in all the other language we use "form session".

src/openforms/authentication/api/views.py
```diff
@@ -62,3 +70,38 @@ def delete(self, request, *args, **kwargs):
...
+        remove_submission_from_session(submission, request.session)
+
+        if FORM_AUTH_SESSION_KEY in request.session:
+            del request.session[FORM_AUTH_SESSION_KEY]
+
+        if not submission.auth_attributes_hashed:
+            submission.hash_identifying_attributes(save=True)
+
+        for plugin in register.iter_enabled_plugins():
+            plugin.logout(request)
```
sergei-maertens, line 105:
```suggestion
        for plugin in register:
            plugin.logout(request)
```
isn't the default behaviour to only loop over enabled plugins? so that could clean up the code a bit without leaking implementation details maybe?

Bartvaderkin: I think things could go wrong if we iterate over all plugins and try to logut on one that is not configured (eg: missing urls/config etc)

Bartvaderkin: I checked after discussion and the default register iterator iterates over everything included disabled plugins so we definitely want .iter_enabled_plugins() here.

sergei-maertens (approved):

PR #1450, opened 2022-03-18
[#1449] Custom setting for max number of files

### LINKED ISSUE #1449: File component should have a max number of uploaded files setting
Taiga issue DH 8

SilviaAmAm:
Fixes #1449 

Front end PR: https://github.com/open-formulieren/open-forms-sdk/pull/181


sergei-maertens (approved):
Translations are missing, but I did add a step to our test report form to check for missing translations so we can update them all at once before releasing a new version.

PR #1455, opened 2022-03-18
[#1453] Fixed formio file upload handler to not display JSON error

### LINKED ISSUE #1453: File upload component displays error JSON from backend instead of message
Taiga DH 11

Formio dumps the whole response in the message: https://github.com/formio/formio.js/blob/280ba15cffd64f5936a7a2cccd45322ac81b92e8/src/components/file/File.js#L765

Bartvaderkin:
Fixes #1453


sergei-maertens (changes requested):
I'm not liking at all how much Formio leaks into our API endpoint definitions and behaviour. I'm wondering if we can instead tap into the fetch plugin API to process the response from Open Forms into a format that Formio can understand: https://help.form.io/developers/fetch-plugin-api#plugin-hooks

This would move the entire solution/workaround to the SDK instead of the Open Forms API.

Motivation for this:

1. Alternative form renderers do not exist yet, but we do want to support them in a vendor agnostic way. Returning the validation errors as JSON object is the most portable way and error message formatting is for the client to figure out.
2. This format of validation errors does not comply with the Dutch API strategy which implements/prescribes an RFC for this format.
3. The custom renderer class does not behave the way a DRF renderer class should behave. Renderer classes should to be selected based on the `Accept` request header, which DRF itself implements. Those `isinstance` checks are not desired.
4. It's really hard to document the API behaviour in the API spec this way - especially because the behaviour of the custom renderer is not tied to an HTTP status code. This sucks for the clients implementing the API endpoint. On top of that, the API docs at the moment are just plain wrong (and even undocumented at the moment :grimacing: )
5. This change requires an API version 2.0 because the behaviour changed. 
6. If we decide to move away from Formio (which is going to happen at some point because of the various challenges), we have to make breaking changes again or be stuck with a non-ideal situation that no-one asked for.

Bartvaderkin:
I updated to PR to move the endpoint to the FormIO specific `/api/v1/formio/file/upload`.

src/openforms/formio/api/views.py
```diff
@@ -0,0 +1,75 @@
...
+        "Access to this view requires an active form submission. "
+        "Unclaimed temporary files automatically expire after {expire_days} day(s). \n\n"
+        "The maximum upload size for this instance is `{max_upload_size}`. Note that "
+        "this includes the multipart metadata and boundaries, so the actual maximum "
+        "file upload size is slightly smaller."
+    ).format(
+        expire_days=settings.TEMPORARY_UPLOADS_REMOVED_AFTER_DAYS,
+        max_upload_size=filesizeformat(settings.MAX_FILE_UPLOAD_SIZE),
+    ),
+)
```
sergei-maertens, line 36: The validation error response format needs to be documented explicitly in the API docs - now it's still emitting the standard JSON output.

src/openforms/submissions/api/renderers.py
```diff
@@ -21,3 +21,13 @@ class FileRenderer(renderers.BaseRenderer):
 
     def render(self, data, media_type=None, renderer_context=None):
         return data
+
+
+class JSONOrPlainTextRenderer(renderers.JSONRenderer):
```
sergei-maertens, line 26: shouldn't this inherit from the `CamelCaseJSONRenderer`?

I'm also thinking if it wouldn't be cleaner to implement `def finalize_response(self, request, response, *args, **kwargs):` on the particular view and look at `response.status_code`, and if that's `HTTP 400`, then change the `request.accepted_renderer` and `request.accepted_media_type`. This then allows you to implement a Plain Text renderer and a regular JSON renderer without these sort of nasty `isinstance` checks, since tecnically the string `"foo"` can still be regular JSON:


```py
>>> import json
>>> json.dumps("foo")
'"foo"'
```

Bartvaderkin: I applied this suggestion in the PR.

src/openforms/formio/api/views.py
Bartvaderkin: I applied the `responses` dictionary to `extend_schema` in the PR.

sergei-maertens (approved):

sergei-maertens:
Couple more changes I made:

* reverted the changes to the endpoint in submissions app, except for the deprecation
* fixed the API documentation - the return content type is `text/plain` for HTTP 400 and the actual type of the content is `str`, not `bytes`/binary content. The wire format is binary, but that's not what you document in an API spec, you document what is decoded by the client and what the format is then (text, json, yaml...)

sergei-maertens:
Backported in e37f7c22c5d8fd8052bb55d0b03017937918c31d

PR #1474, opened 2022-03-29
[#1472] Display errors for the actions in the logic rules

### LINKED ISSUE #1472: Admin errors for logic tab don't show which field caused the error
DH Taiga issue 13

When there is an error in the logic tab, it shows the red badge on the tab, but it doesnt show an error message near the field that caused the error.

See: https://github.com/open-formulieren/open-forms/pull/1474#pullrequestreview-925753078

We need to refactor how the Action component works. Right now all the logic for the different types of actions is managed by the action component, while we want to change it to be similar to https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/FormStepSummary/ComponentValueDisplay.js. Each Action type would have a separate component that manages the logic. These can then deal with displaying errors.

We also need to decide on the design of this page.

Todo:

- [x] Refactor action component https://github.com/open-formulieren/open-forms/pull/1496
- [x] Validate that component can't be empty (both in the trigger and the actions) https://github.com/open-formulieren/open-forms/pull/1499
- [x] Display errors in the actions https://github.com/open-formulieren/open-forms/pull/1507
- [x] Style

SilviaAmAm:
Fixes #1472 

This fixes the problem that the errors reported from the backend for the actions in a logic rule were not displayed.

### DIFF
--- a/src/openforms/js/components/admin/form_design/FormLogic.js
+++ b/src/openforms/js/components/admin/form_design/FormLogic.js
@@ -1,6 +1,7 @@
 import React, {useContext} from 'react';
 import PropTypes from 'prop-types';
 import {useIntl, FormattedMessage} from 'react-intl';
+import _ from 'lodash';
 
 import DeleteIcon from '../DeleteIcon';
 import Trigger from './logic/Trigger';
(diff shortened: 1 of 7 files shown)


sergei-maertens (changes requested):
I'm not satisfied yet with the UI nor with the code, and I do realize that it's all very complicated with the polymorphism going around!

Using screenshot for reference:

image

1. The alignment of the errors and inputs doesn't look particularly great
2. I'm missing the validation (backend) that a component must be specified
3. The `action.type` and `action.value` errors are bundled together. I think only one error should even be displayed, since the last field only becomes available/visible after selecting which attribute should be changed.

I think we need to take a bigger step back and look at the "framework" for plugging in all those different features + configuration of how to extract the relevant errors and display them. Ideally this would be declarative. We're definitely hitting some limits with the structure of the validation errors in `invalidParams` and having to map that back to the internal data structure, I'm going to keep that in mind towards the NL API strategy people as it's a bit silly we convert the nested structure in the backend to a flat list only to convert it back to a nested structure in the frontend.

sergei-maertens:
Removed the backport label - I get that it's annoying it's not available on 1.0.x, but it's not a critical bug so to reduce complexity of managing versions let's make this a 1.1 feature :)

PR #1477, opened 2022-03-29
[#1476] Add readonly opt to BSN/date/postcode

### LINKED ISSUE #1476: BSN, postcode and date component dont have "read-only" option
Taiga issue DH 16

The text components have the read only option, but not the BSN, postcode and the date component

SilviaAmAm:
Fixes #1476


src/openforms/js/components/form/edit/tabs.js
```diff
@@ -46,6 +46,14 @@ const SENSITIVE_BASIC = {
     }),
 };
 
+const SENSITIVE_READ_ONLY = {
```
sergei-maertens, line 49: all these different configurations are getting out of hand, I think it's starting to make more sense to just compose them in the actual form component because the re-use potential diminishes with every configuration change.

We're defining groups of config options and then in components filtering out certain options, in other places we add options... it's completely unclear what options are now actually included from looking at the code.

No need to change it for this PR, but keep it in mind for the next change we need to make and grab the opportunity to clean this up.

sergei-maertens (approved):

PR #1481, opened 2022-03-30
DMN component - HACKATHON POC

### LINKED ISSUE #1469: Prepare DMN (Camunda) integration for hackathon
There will be a hackathon mid-april to defer "business rule" evaluation from Open Forms to a DMN engine, at first instance Camunda. We have to prepare this because there's limited time to implement all the functionality at the hackathon itself.

Note that the technical direction  chosen here is to achieve the hackaton goals and not the definitive way to approach this. Especially in the context of larger future epics, expect rework where other abstraction layers will be used (such as "variables").

**Tasks**

- [x] Set up a local Camunda 7.16 (docker-compose) with DMN enabled
- [x] Set up some simple DMN tables to test/evaluate against
- [ ] Set up a custom Formio component
    - [x] Add ability to select a DMN table from the Camunda backend, specifying which table should be evaluated
    - [ ] Possibly add ability to map variables (component keys) from form design to input values for the DMN table
    - [x] Allow specifying a Django template fragment to display the resulting output in a formatted way (plain text only initially, future may be WYSIWYG)
    - [x] Allow specifying a key so the resulting value can also be sent to registration backends/referred to in logic
- [x] Evaluate the DMN table as part of the dynamic form configuration resolution.
    - [x] Investigate if this can be done in parallel or not with other calls (prefill, logic...)

Vaguele related to but not the same as #124

sergei-maertens:
Fixes #1469


sergei-maertens:
Hackathon is done

PR #1497, opened 2022-04-04
[#1495] Mutual TLS soap

### LINKED ISSUE #1495: Mutual TLS rework for SOAP-services
* [x] Move the `StUF`-model authorization fields, into the `SOAP Service` model
* [x] Replace the cert/key file-fields with our new `Certificate` model-relations (server and client)
* [x] Move SOAP-version field from `StUF` to `SOAP` model as well
* [x] Refactor the SOAP/StUF-client(s) to use these certificates for both client cert and server cert validation

image

SilviaAmAm:
Fixes #1495


src/stuf/migrations/0010_auto_20220404_1053.py
```diff
@@ -0,0 +1,68 @@
...
+    StufService = apps.get_model("stuf", "StufService")
+
+    updated_soap_services = []
+    for stuf_service in StufService.objects.all():
+        soap_service = stuf_service.soap_service
+        soap_service.endpoint_security = stuf_service.endpoint_security
+        soap_service.user = stuf_service.user
+        soap_service.password = stuf_service.password
+        soap_service.client_certificate.public_certificate = stuf_service.certificate
+        soap_service.client_certificate.private_key = stuf_service.certificate_key
```
sergei-maertens, line 17: this will crash if `client_certificate is None`

src/stuf/migrations/0010_auto_20220404_1053.py
```diff
@@ -0,0 +1,68 @@
...
+    SoapService.objects.bulk_update(
+        updated_soap_services,
+        [
+            "endpoint_security",
+            "user",
+            "password",
+            "client_certificate",
+            "soap_version",
+        ],
+    )
```
sergei-maertens, line 30: Fairly certain that this does not save the `soap_service.client_certificate` changes.

src/stuf/migrations/0010_auto_20220404_1053.py
```diff
@@ -0,0 +1,68 @@
...
+    StufService = apps.get_model("stuf", "StufService")
+
+    updated_stuf_services = []
+    for stuf_service in StufService.objects.all():
+        soap_service = stuf_service.soap_service
+        stuf_service.endpoint_security = soap_service.endpoint_security
+        stuf_service.user = soap_service.user
+        stuf_service.password = soap_service.password
+        stuf_service.certificate = soap_service.client_certificate.public_certificate
+        stuf_service.certificate_key = soap_service.client_certificate.private_key
```
sergei-maertens, line 43: Similar issues as above

src/stuf/migrations/0010_auto_20220404_1053.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("stuf", "0009_auto_20220404_1050"),
+    ]
+
+    operations = [
+        migrations.RunPython(forward, backwards),
```
sergei-maertens, line 67: I'd like some migration tests for this operation :) there is a base test class available for this: https://github.com/open-formulieren/open-forms/blob/c8709ef68ec4b8f75eb16a721eb533d6f8bd674a/src/openforms/utils/tests/test_migrations.py#L6

src/stuf/stuf_zds/tests/test_client.py
```diff
@@ -0,0 +1,142 @@
...
+    "soap",
+)
+
+
+@requests_mock.Mocker()
+class StufZdsClientTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        certificate_f = open(os.path.join(TEST_CERTIFICATES, "test.certificate"), "r")
+        key_f = open(os.path.join(TEST_CERTIFICATES, "test.key"), "r")
```
sergei-maertens, line 32: please use a context manager instead: `with open(...) as ...`

src/stuf/stuf_zds/tests/test_client.py
```diff
@@ -0,0 +1,142 @@
...
+    def test_mutual_tls(self, m):
+        stuf_service = StufServiceFactory.create(
+            soap_service__client_certificate=self.client_certificate,
+            soap_service__server_certificate=self.server_certificate,
+        )
+
+        with open(os.path.join(TEST_XML, "creeerZaak.xml"), "r") as f:
+            m.post(
+                stuf_service.soap_service.url,
+                content=Template(f.read()).render(Context({})).encode("utf-8"),
```
sergei-maertens, line 74: can't we use `render_to_string` here instead of dealing with these files ourselves?

src/stuf/stuf_zds/tests/test_client.py
```diff
@@ -0,0 +1,142 @@
...
+
+        client.create_zaak(zaak_identificatie="ZAAK-01", zaak_data={}, extra_data={})
+
+        history = m.request_history
+        request_with_tls = history[-1]
+
+        self.assertEqual(
+            self.server_certificate.public_certificate.path, request_with_tls.verify
+        )
+        self.assertTupleEqual(
```
sergei-maertens, line 87: please use the bare `self.assertEqual` - this will call the type-specific method, see https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

sergei-maertens (changes requested):

src/stuf/stuf_zds/tests/test_client.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def test_mutual_tls(self, m):
+        stuf_service = StufServiceFactory.create(
+            soap_service__client_certificate=self.client_certificate,
+            soap_service__server_certificate=self.server_certificate,
+        )
+
+        m.post(
+            stuf_service.soap_service.url,
+            content=render_to_string(os.path.join(TEST_XML, "creeerZaak.xml")).encode(
```
sergei-maertens, line 72: `render_to_string` should not take a full/absolute file path, but just the template name. In this case, that would be `stuf_zds/soap/creeerZaak.xml`. We effectively want to decouple from the file system here and just leverage the django template engine.

src/stuf/stuf_zds/tests/test_client.py
```diff
@@ -0,0 +1,142 @@
...
+    def test_mutual_tls_no_private_key(self, m):
+        stuf_service = StufServiceFactory.create(
+            soap_service__client_certificate=self.client_certificate_only,
+            soap_service__server_certificate=self.server_certificate,
+        )
+
+        with open(os.path.join(TEST_XML, "creeerZaak.xml"), "r") as f:
+            m.post(
+                stuf_service.soap_service.url,
+                content=Template(f.read()).render(Context({})).encode("utf-8"),
```
sergei-maertens, line 104: more raw file/`Template` use here

src/stuf/tests/test_migrations.py
```diff
@@ -0,0 +1,172 @@
...
+from ..constants import SOAPVersion
+
+TEST_CERTIFICATES = os.path.join(
+    settings.BASE_DIR,
+    "src",
+    "stuf",
+    "stuf_zds",
+    "tests",
+    "data",
+)
```
sergei-maertens, line 17: if you use `pathlib.Path`, you get usually a bit more readable code and it's the modern way to deal with filesystem paths in Python.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1499, opened 2022-04-06
[#1472] Improve backend validation of logic rules

SilviaAmAm:
Part of #1472

### DIFF
--- a/src/openforms/forms/api/serializers.py
+++ b/src/openforms/forms/api/serializers.py
@@ -622,6 +622,14 @@ class LogicPropertyActionSerializer(serializers.Serializer):
         ),
     )
 
+    def validate_state(self, value):
+        if value == "":
+            raise serializers.ValidationError(
+                self.fields["state"].error_messages["null"],
+                code="blank",
+            )
+        return value
+
 
 class LogicValueActionSerializer(serializers.Serializer):
     value = serializers.JSONField(
--- a/src/openforms/forms/api/serializers.py
+++ b/src/openforms/forms/api/serializers.py
@@ -43,6 +43,7 @@ from ..models import (
 )
 from .validators import (
     FormIOComponentsValidator,
+    JsonLogicActionValueValidator,
     JsonLogicTriggerValidator,
     JsonLogicValidator,
 )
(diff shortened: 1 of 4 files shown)


src/openforms/forms/api/serializers.py
```diff
@@ -689,16 +690,18 @@ def validate(self, data: dict) -> dict:
             and action_type in LogicActionTypes.requires_component
             and not component
         ):
-            # raises validation error
-            self.fields["component"].fail("blank")
+            raise serializers.ValidationError(
+                {"component": self.fields["component"].error_messages["blank"]}
```
sergei-maertens, line 694: I think we lost the `code="blank"` information here

src/openforms/forms/api/validators.py
```diff
@@ -123,6 +104,71 @@ def _is_advanced_logic(
...
+            if not isinstance(operand, JsonLogicTest):
+                continue
+
+            form = get_from_serializer_data_or_instance(
+                self.form_field, data, serializer
+            )
+
+            # some data missing, can't perform check
+            if not form:
+                return
```
sergei-maertens, line 129: also loop invariant, you can use this guard clause outside of the loop

src/openforms/forms/api/serializers.py
```diff
@@ -689,16 +690,20 @@ def validate(self, data: dict) -> dict:
             and action_type in LogicActionTypes.requires_component
             and not component
         ):
-            # raises validation error
-            self.fields["component"].fail("blank")
+            raise serializers.ValidationError(
+                {"component": self.fields["component"].error_messages["blank"]},
+                code="blank",
```
sergei-maertens, line 695: did you verify that this does indeed contain the code in the error response? Afaik it gets discarded if you pass a dict to `ValidationError` instead of just an err message string.

You should be able to easily make that assertion in the tests.

sergei-maertens (changes requested):

src/openforms/forms/api/serializers.py
SilviaAmAm: It looks like it is kept! I added the check in the tests

sergei-maertens (approved):

PR #1504, opened 2022-04-12
Compatibility new OneWelcome and Signicat brokers

SilviaAmAm:
Part of https://github.com/maykinmedia/django-digid-eherkenning/issues/4


sergei-maertens:
self-reminder: release package to PyPI

requirements/base.txt
```diff
@@ -243,7 +243,7 @@ kombu==5.2.3
     # via celery
 lockfile==0.12.2
     # via django-yubin
-lxml==4.7.1
+lxml==4.6.5
```
sergei-maertens, line 246: this downgrade is caused by python3-saml upstream

sergei-maertens (approved):

PR #1509, opened 2022-04-15
[#1482] Bulk export forms

### LINKED ISSUE #1482: Bulk action to export forms
DH Taiga issue 19

There should be a way to export many forms at the same time. This can involve downloading many forms (400+).

- [x] Create a custom action in the admin to bulk export forms.
- [x] In the confirmation screen, collect the email of the user requesting the export.
- [x] Once confirmed, a celery task should export all the forms and collect them in a zip file. When it is done, it should send an email to the user with the link to where this file can be downloaded.
- [x] The user needs to be logged in to download the file.
- [x] When the file is downloaded, it is marked as downloaded. 
- [x] Files that have been downloaded more than a week in the past should be deleted.

Examples of how to do this:
https://github.com/modelbrouwers/modelbrouwers/blob/develop/src/brouwers/users/management/commands/process_data_downloads.py
https://github.com/modelbrouwers/modelbrouwers/blob/develop/src/brouwers/users/management/commands/clear_data_downloads.py

SilviaAmAm:
Fixes #1482


src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,76 @@
...
+        initial_data = super().get_initial()
+        form = FormsUUIDsForm(
+            data={
+                "forms_uuids": self.request.session.get("forms_uuids"),
+                "username": self.request.user.username,
+            }
+        )
+        if form.is_valid():
+            initial_data.update(form.cleaned_data)
+        return initial_data
```
sergei-maertens, line 51: This is a weird construction, but if you refactor the action/view function this will not be relevant anymore.

However, I'd expect that the initial data would just be `"forms_uuids": self.request.session.get("forms_uuids")` and done. Not sure why things are still wrapped in another form that is not involved in the view itself.

The `username` construct is also odd - I'd expect to just pass around `request.user` or for celery tasks, `request.user.id`

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def form_valid(self, form):
+        process_forms_export.delay(**form.cleaned_data)
+        return super().form_valid(form)
+
+
+class DownloadExportedFormsView(DetailView):
```
sergei-maertens, line 58: this is missing the `LoginRequiredMixin` and the `UserPassesTestMixin` to check that the user is an admin user. The `self.site.admin_view` in admin URL registration covers that to some extent, but I prefer having an additional layer of protection at the view level itself to prevent accidental exposure in the future.

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,76 @@
...
+            raise PermissionDenied("URL is not valid")
+
+        if request.user.username != forms_export.username:
+            raise PermissionDenied("Wrong user requesting download")
+
+        forms_export.downloaded = True
+        forms_export.date_downloaded = date.today()
+        forms_export.save()
+
+        return FileResponse(open(forms_export.export_content.path, "rb"))
```
sergei-maertens, line 76: You're not using any of the `DetailView` functionality, so this might as well just be a `View` subclass.

Also, you can just pass the model in `get_object_or_404`:

```py
forms_export = get_object_or_404(FormsExport, pk=kwargs["pk"])
```

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,76 @@
...
+    def get(self, request, *args, **kwargs):
+        forms_export = get_object_or_404(FormsExport.objects, pk=kwargs["pk"])
+
+        token = kwargs["token"]
+        valid = exported_forms_token_generator.check_token(forms_export, token)
+        if not valid:
+            raise PermissionDenied("URL is not valid")
+
+        if request.user.username != forms_export.username:
+            raise PermissionDenied("Wrong user requesting download")
```
sergei-maertens, line 70: This really should be a `ForeignKey` to the user on the model instead of storing the username - usernames _can_ change. I'd also include this into the `get_object_or_404` check instead: `(..., pk=kwargs["pk"], user=request.user)`

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,76 @@
...
+
+        forms_export.downloaded = True
+        forms_export.date_downloaded = date.today()
```
sergei-maertens, line 73: please make this a datetime with more precision, and use `django.utils.timezone.now()`

src/openforms/forms/models/form.py
```diff
@@ -444,3 +446,37 @@ class FormLogic(models.Model):
         ),
         default=False,
     )
+
+
+class FormsExport(DeleteFileFieldFilesMixin, models.Model):
+    export_content = PrivateMediaFileField(
+        verbose_name=_("export content"),
+        upload_to="exports/%Y/%m/%d",
+        help_text=_("Zip file containing all the exported forms"),
```
sergei-maertens, line 455:
```suggestion
        help_text=_("Zip file containing all the exported forms."),
```

src/openforms/forms/models/form.py
```diff
@@ -444,3 +446,37 @@ class FormLogic(models.Model):
...
+    )
+    downloaded = models.BooleanField(
+        verbose_name=_("downloaded"),
+        help_text=_("Whether the zip file has already been downloaded"),
+        default=False,
+    )
+    date_downloaded = models.DateField(
```
sergei-maertens, line 462: As mentioned above, please make this a `DateTimeField`

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,76 @@
...
+        forms_export.save()
+
```
sergei-maertens, line 75: Additionally, let's also log this in the audit trail using the `openforms.logging.logevent` constructs

src/openforms/forms/models/form.py
```diff
@@ -444,3 +446,37 @@ class FormLogic(models.Model):
...
     )
+
+
+class FormsExport(DeleteFileFieldFilesMixin, models.Model):
+    export_content = PrivateMediaFileField(
+        verbose_name=_("export content"),
+        upload_to="exports/%Y/%m/%d",
+        help_text=_("Zip file containing all the exported forms"),
+    )
+    downloaded = models.BooleanField(
```
sergei-maertens, line 457: this can be implicitly derived as `self.date_downloaded is not None`, no need for this field I believe.

src/openforms/forms/models/form.py
```diff
@@ -444,3 +446,37 @@ class FormLogic(models.Model):
...
+        verbose_name=_("user email"),
+        help_text=_(
+            "The email address that the user entered when requesting the export"
+        ),
+    )
+    username = models.CharField(
+        verbose_name=_("username"),
+        help_text=_("The unique username of the user that requested the download"),
+        max_length=150,
+    )
```
sergei-maertens, line 478: Should be replaced with a single `user = models.ForeignKey(settings.AUTH_USER_MODEL)` and you can derive the username and email from those objects.

src/openforms/forms/models/form.py
```diff
@@ -444,3 +446,37 @@ class FormLogic(models.Model):
...
+
+    class Meta:
+        verbose_name = _("forms export")
+        verbose_name_plural = _("forms exports")
```
sergei-maertens, line 482: Please also implement `def __str__(self)`

src/openforms/forms/templates/admin/forms/form/export.html
```diff
@@ -0,0 +1,53 @@
...
+{% block content %}
+    <h1>{% trans 'Export forms' %}</h1>
+    <div id="content-main">
+        {% blocktrans %}
+            <div>
+                Please enter your email address below. Once your request has been processed, you
+                will receive a link where you can download a zip file with all the exported forms.
+                <br/><br/>
+            </div>
+        {% endblocktrans %}
```
sergei-maertens, line 29: Keep HTML out of translatable strings, as this provides an HTML injection risk depending on who manages the translations.

I also think that asking the e-mail address is only needed if there is no `request.user.email`, and would even go a step further to have them configure that in their user account before being able to make exports. That simplifies the amount of edge cases we need to deal with. This could be an early check in the admin action.

src/openforms/forms/templates/admin/forms/formsexport/email_content.html
```diff
@@ -0,0 +1,16 @@
...
+{{ download_url }}<br>
+<br>
+
+Note: the URL is only valid once.
+
+<br>
+Best wishes,<br>
+<br>
+Open Forms<br>
+{% endblocktrans %}
```
sergei-maertens, line 16: See earlier remark about the `blocktrans` contents. I think in the next review/iteration we'll have to check the copy of the email a bit better too.

src/openforms/forms/management/commands/delete_export_files.py
```diff
@@ -0,0 +1,23 @@
...
+        before_date = (
+            timezone.now() - timedelta(days=settings.FORMS_EXPORT_REMOVED_AFTER_DAYS)
+        ).date()
+
+        forms_exports = FormsExport.objects.filter(
+            downloaded=True, date_downloaded__lt=before_date
+        )
+
+        for forms_export in forms_exports:
+            forms_export.delete()
```
sergei-maertens, line 23: it's just a queryset, why not `forms_exports.delete()`?

src/openforms/forms/admin/tasks.py
```diff
@@ -0,0 +1,76 @@
...
+from ..models import Form
+from ..models.form import FormsExport
+from ..utils import export_form
+from .tokens import exported_forms_token_generator
+
+logger = logging.getLogger(__name__)
+
+
+@app.task
+def process_forms_export(forms_uuids, email, username):
```
sergei-maertens, line 28: please add type hints, makes it easier to reason about the intent and the interface contract

src/openforms/forms/admin/tasks.py
```diff
@@ -0,0 +1,76 @@
...
+    forms = Form.objects.filter(uuid__in=forms_uuids)
+
+    with tempfile.TemporaryDirectory(dir=settings.PRIVATE_MEDIA_ROOT) as temp_dir:
```
sergei-maertens, line 31: does this clean up/delete the `temp_dir` after all the code here has been executed? If that is the case, please add a comment documenting this behaviour :)

src/openforms/forms/admin/tasks.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+
+        zip_filepath = Path(temp_dir, f"forms-export_{uuid4()}.zip")
+        with ZipFile(zip_filepath, "w") as zipfile:
+            for output_file in output_files:
+                zipfile.write(output_file, arcname=os.path.basename(output_file))
+
+        with open(zip_filepath, "rb") as zipfile:
+            forms_export = FormsExport.objects.create(
+                export_content=File(zipfile, name=os.path.basename(zip_filepath)),
```
sergei-maertens, line 48: I feel like there should be some pathlib/Path instance attributes that you can use instead of having to use `os.path.basename`

src/openforms/forms/admin/tasks.py
```diff
@@ -0,0 +1,76 @@
+import logging
+import os
+import tempfile
+from pathlib import Path
+from uuid import uuid4
+from zipfile import ZipFile
+
+from django.conf import settings
+from django.core.files import File
+from django.core.mail import send_mail
```
sergei-maertens, line 10: please use our own wrappers in `openforms.email.utils`

sergei-maertens (changes requested):

src/openforms/forms/admin/tokens.py
```diff
@@ -0,0 +1,30 @@
...
+from openforms.tokens import BaseTokenGenerator
+
+
+class ExportedFormsTokenGenerator(BaseTokenGenerator):
+    """
+    Strategy object used to generate and check tokens for downloading a zip file of exported forms.
+    """
+
+    key_salt = "openforms.forms.admin.tokens.ExportedFormsTokenGenerator"
+    token_timeout_days = settings.SUBMISSION_REPORT_URL_TOKEN_TIMEOUT_DAYS
```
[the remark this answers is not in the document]
SilviaAmAm: Oups, had added it, forgot to replace it :sweat_smile:

src/openforms/forms/models/form.py
SilviaAmAm: I left the email field separate in case the user wants to enter a different email address or the email address associated with the user is blank.

src/openforms/forms/management/commands/delete_export_files.py
SilviaAmAm: Because otherwise the `.delete()` on the model doesn't get called (which removes the files from the filesystem).

I could use the signals, but I thought that this would be simpler? :sweat_smile:

sergei-maertens: Hrm, the code in `openforms.utils.files.DeleteFilesQuerySetMixin` should take care of that though :thinking:

SilviaAmAm: Ah I shall check again, maybe I made a mistake

SilviaAmAm: Yes, I made a mistake

sergei-maertens (commented):
Quick thought: exports that are never downloaded are now never pruned, right? So that could lead to excessive disk space usage if people keep requesting exports without the intention to download them.

SilviaAmAm:
Yes I was also thinking about this. I can change this so that after the download link expires then the files can be pruned

sergei-maertens:
> Yes I was also thinking about this. I can change this so that after the download link expires then the files can be pruned

maybe let's get rid of the entire token approach for this download - I don't mind if they download the export multiple times, there's no replay attack here as we check the user who created the export anyway.

So:

1. Track creation timestamp for export
2. Download URL can just contain the UUID of the export
3. On download, check that the `export.user == request.user`
4. Prune exports if they are older dan `x` number of days, irrespective if they've been downloaded or not

src/openforms/forms/admin/form.py
```diff
@@ -351,3 +363,43 @@ def delete_queryset(self, request, queryset):
...
+                ),
+                level=messages.ERROR,
+            )
+            return
+
+        selected_forms_uuids = queryset.values_list("uuid", flat=True)
+        form = ExportFormsForm(
+            initial={
+                "forms_uuids": [str(form_uuid) for form_uuid in selected_forms_uuids],
+                "email": request.user.email,
```
sergei-maertens, line 383: this is not needed anymore, right?

src/openforms/forms/admin/form.py
```diff
@@ -351,3 +363,43 @@ def delete_queryset(self, request, queryset):
...
+        )
+        context = dict(self.admin_site.each_context(request), form=form)
+        return TemplateResponse(request, "admin/forms/form/export.html", context)
+
+
+@admin.register(FormsExport)
+class FormsExportAdmin(admin.ModelAdmin):
+    list_display = ("uuid", "user", "datetime_requested")
+    list_filter = ("user",)
+    search_fields = ("user",)
```
sergei-maertens, line 394: fairly certain the search bar is broken now, try searching on a user username

src/openforms/forms/admin/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+
+from ..models import Form
+from ..models.form import FormsExport
+from ..utils import export_form
+
+logger = logging.getLogger(__name__)
+
+
+@app.task
+def process_forms_export(forms_uuids: list, email: str, user_id: int) -> None:
```
sergei-maertens, line 27: `email` should not be needed since we have `user_id` and associated `user.email`

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,65 @@
...
+from openforms.logging import logevent
+
+from ..models.form import Form, FormsExport
+from .tasks import process_forms_export
+
+
+class ExportFormsForm(forms.Form):
+    forms_uuids = SimpleArrayField(forms.UUIDField(), widget=forms.HiddenInput)
+    email = forms.EmailField(widget=forms.HiddenInput)
```
sergei-maertens, line 21: should not be needed as we have `request.user` available

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,65 @@
...
+    success_url = reverse_lazy("admin:forms_form_changelist")
+    success_message = _("Success! You will receive an email when your export is ready.")
+
+    def test_func(self):
+        return self.request.user.is_superuser
+
+    def form_valid(self, form):
+        process_forms_export.delay(
+            forms_uuids=form.cleaned_data["forms_uuids"],
+            email=form.cleaned_data["email"],
```
sergei-maertens, line 43: not needed

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,65 @@
...
+            user_id=self.request.user.id,
+        )
+        return super().form_valid(form)
+
+
+class DownloadExportedFormsView(LoginRequiredMixin, UserPassesTestMixin, View):
+    model = FormsExport
```
sergei-maertens, line 50: unused attribute

src/openforms/forms/admin/views.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_func(self):
+        return self.request.user.is_superuser
+
+    def get(self, request, *args, **kwargs):
+        forms_export = get_object_or_404(
+            FormsExport, uuid=kwargs["uuid"], user=request.user
+        )
+
+        if request.user != forms_export.user:
+            raise PermissionDenied("Wrong user requesting download")
```
sergei-maertens, line 61: will never happen because of the `user=request.user` filter above, instead they'll get a 404 (which is fine)

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -0,0 +1,153 @@
...
+
+        self.assertEqual(302, response.status_code)
+
+    def test_staff_cant_access(self):
+        user = UserFactory(is_staff=True, is_superuser=False)
+        self.client.force_login(user)
+
+        response = self.client.get(reverse("admin:forms_export"))
+
+        self.assertEqual(403, response.status_code)
```
sergei-maertens, line 33: Looks like we also need to update our fixture with the default group permissions to give them access to the export view, no?

Please add a test for a staff user that is not a superuser with the relevant permissions set up. There are some permission-related things already in either the factories or somewhere else in the codebase, you can search for `Permission` model.

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #1498, opened 2022-04-06
Payment emails are sent/received twice


joeribekker:
### Product versie / Product version

1.0.0-rc.4

### Omschrijf het probleem / Describe the bug

The payment confirmation to the municipality is recieved twice. Unknown why.

(Taiga Vught 5)

Bartvaderkin:
I tested this locally with both demo and our Ogone dev account and can't reproduce double emails.

sergei-maertens:
Using Bart's management command to find the amount of duplicated e-mails, we found:

* test: 3 payments total, 100% duplicated (6 emails sent)
* prod: 60 payments total, 93% duplicated

So we do have a structural problem, in the form of a race condition. Suspected problems is:

1. User initiates and completes payment
2. User is redirected back from Ogone to the form
3. The callback view is received and we sent out the payment confirmation
4. At the same time, the Ogone webhook calls back to the view and the payment confirmation is also sent out

3 & 4 are race conditions running at the same time. We might have to set up a test case with threading to reproduce this and then implement some sort of locking mechanism to mitigate this. Possibly we could also treat one confirmation path with a fixed minimum delay, so that the updated flag tracking payment/confirmation email sent is updated by the other execution path.

sergei-maertens:
These are the suspected racing handlers:

* https://github.com/open-formulieren/open-forms/blob/master/src/openforms/payments/views.py#L239
* https://github.com/open-formulieren/open-forms/blob/master/src/openforms/payments/views.py#L317

PR #1511, opened 2022-04-19
[#1498] Created a command to check for duplicate payment-confirmation email-registration mails

Bartvaderkin:


sergei-maertens (approved):
Looks alright and useful and I can live with the absence of tests, but please add the file to the coverage config in `setup.cfg` so that it's ignored in coverage reporting.

PR #1522, opened 2022-04-20
[#1510] Import bulk forms

### LINKED ISSUE #1510: Bulk import forms
### Thema / Theme

Admin

Related to #1482 (bulk export forms)

There should be the functionality to import forms exported in bulk

_No response_

_No response_

SilviaAmAm:
Fixes #1510

### DIFF
--- a/src/openforms/forms/admin/tasks.py
+++ b/src/openforms/forms/admin/tasks.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import tempfile
 import zipfile
 from pathlib import Path
@@ -13,6 +12,7 @@ from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
+from privates.storages import private_media_storage
 from rest_framework.exceptions import ValidationError
 
 from openforms.accounts.models import User
@@ -80,8 +80,8 @@ def process_forms_import(import_file: str, user_id: int) -> None:
     user = User.objects.get(id=user_id)
     failed_files = []
     # This deletes the temp dir once the context manager is exited
-    with tempfile.TemporaryDirectory(dir=Path(import_file).parent) as temp_dir:
-        with zipfile.ZipFile(import_file, "r") as zip_file:
+    with tempfile.TemporaryDirectory() as temp_dir:
+        with zipfile.ZipFile(private_media_storage.open(import_file), "r") as zip_file:
             for zipped_form_file in zip_file.infolist():
                 try:
                     # This normalises the path before extracting the files (to avoid writing outside the temp_dir)
@@ -95,7 +95,7 @@ def process_forms_import(import_file: str, user_id: int) -> None:
                     continue
 
     logevent.bulk_forms_imported(user=user, failed_files=failed_files)
-    os.remove(import_file)
+    private_media_storage.delete(import_file)
 
 
 @app.task
--- a/src/openforms/forms/admin/tasks.py
+++ b/src/openforms/forms/admin/tasks.py
@@ -13,9 +13,12 @@ from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
+from rest_framework.exceptions import ValidationError
+
 from openforms.accounts.models import User
 from openforms.celery import app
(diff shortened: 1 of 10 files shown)


src/openforms/forms/admin/views.py
```diff
@@ -54,3 +65,64 @@ def get(self, request, *args, **kwargs):
...
+    def _bulk_import_forms(self, import_file):
+        imports_dir = Path(settings.PRIVATE_MEDIA_ROOT, "imports")
+        if not imports_dir.exists():
+            os.mkdir(imports_dir)
+        imports_file = Path(imports_dir, f"import_forms_{uuid4()}.zip")
+        with open(imports_file, "wb") as f_private:
+            for chunk in import_file.chunks():
+                f_private.write(chunk)
+
+        process_forms_import.delay(str(imports_file), self.request.user.id)
```
sergei-maertens, line 128: I'd completely delegate this to the file storage:

```py
from privates.storages import private_media_storage

name = f"imports/import_forms_{uuid4()}.zip"
name = private_media_storage.save(name, import_file)

process_forms_import.delay(name, self.request.user.id)
```

src/openforms/forms/admin/tasks.py
```diff
@@ -70,6 +75,29 @@ def process_forms_export(forms_uuids: list, user_id: int) -> None:
         )
 
 
+@app.task
+def process_forms_import(import_file: str, user_id: int) -> None:
+    user = User.objects.get(id=user_id)
+    failed_files = []
+    # This deletes the temp dir once the context manager is exited
+    with tempfile.TemporaryDirectory(dir=Path(import_file).parent) as temp_dir:
+        with zipfile.ZipFile(import_file, "r") as zip_file:
```
sergei-maertens, line 84: I'd leverage the storage class here too, following the changes in the view:

```py
from privates.storages import private_media_storage

def process_forms_import(filename: str, user_id: int) -> None:
    # This deletes the temp dir once the context manager is exited
    with tempfile.TemporaryDirectory(dir=Path(import_file).parent) as temp_dir:
        with zipfile.ZipFile(private_media_storage.open(filename), "r") as zip_file:
            ...
```

src/openforms/forms/admin/tasks.py
```diff
@@ -70,6 +75,29 @@ def process_forms_export(forms_uuids: list, user_id: int) -> None:
...
+                        zip_file.extract(member=zipped_form_file, path=temp_dir)
+                    )
+                except ValidationError as exc:
+                    filename = Path(zipped_form_file.filename).name
+                    failed_files.append((filename, exc.detail))
+                    logger.error("Could not import form %s", filename)
+                    continue
+
+    logevent.bulk_forms_imported(user=user, failed_files=failed_files)
+    os.remove(import_file)
```
sergei-maertens, line 98: should use the storage class too: `private_media_storage.delete(filename)`

src/openforms/forms/tests/admin/test_tasks.py
```diff
@@ -60,3 +66,65 @@ def test_zip_file_contains_data(self):
...
+
+        process_forms_export(
+            forms_uuids=[form1.uuid, form2.uuid],
+            user_id=user.id,
+        )
+
+        cls.form_export = FormsExport.objects.get()
+        cls.user = user
+
+    def _copy_file_to_imports_tempdir(self):
```
sergei-maertens, line 86: better to use the private_media_storage here too

sergei-maertens (changes requested):

src/openforms/forms/admin/tasks.py
sergei-maertens: you can pass a filename or any file-like object to `ZipFile`, which is what's happening in my suggestion. We open it explicitly using the storage, and the storage is responsible for finding it on file system (or possibly in the future in Amazon S3 or other object storages)

It's basically using the abstraction that uploaded files may not be on the same server in the future, and this code pattern can handle both situations.

SilviaAmAm: Ohh i see, i shall update my update

sergei-maertens (approved):

PR #1528, opened 2022-04-22
[#1527] Hidden fields empty data

### LINKED ISSUE #1527: Data of previously filled fields that become hidden is still present in the summary

1.0/1.1

Taiga DH 27

If there is a form with logic that changes fields from hidden/visible, if the user enters data in one of these fields and then changes another field so that the previous field becomes hidden, the data is still kept in the submission and shown in the summary.

Minimal form:
1. Make a form with a radio button with values `a` and `b`. Add 2 text fields with key `fieldA` and `fieldB`. Both text fields are hidden.
2. Add 2 logic rules: if value of radiobutton is `a`, make `fieldA` visible. If value of radiobutton is `b`, make `fieldB` visible.

User:
1. Select radio button option `a`. 
2. Fill `fieldA` with some data
3. Now change radio button to `b`.
4. Fill `fieldB` with some data
5. Go to summary

Hier both fieldA and fieldB are shown with the data entered

Only the data of visible fields should be shown.

### Screen resolution

_No response_

### Browser

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1527

Front end PR: https://github.com/open-formulieren/open-forms-sdk/pull/192

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -29,6 +29,12 @@ def set_property_value(
     return configuration
 
 
+def get_component(configuration: JSONObject, key: str) -> JSONObject:
+    for component in iter_components(configuration=configuration, recursive=True):
+        if component["key"] == key:
+            return component
+
+
 @elasticapm.capture_span(span_type="app.submissions.logic")
 def evaluate_form_logic(
     submission: "Submission",
@@ -107,7 +113,11 @@ def evaluate_form_logic(
         data_diff = {}
         for key, new_value in step.data.items():
             original_value = data.get(key)
-            if new_value == original_value:
+            # Reset the value of any field that may have become hidden again after evaluating the logic
+            if original_value and get_component(configuration, key).get("hidden"):
+                data_diff[key] = defaults.get(key, "")
+                continue
+            elif new_value == original_value:
                 continue
             data_diff[key] = new_value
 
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/form_logic.py
```diff
@@ -107,7 +113,11 @@ def evaluate_form_logic(
         data_diff = {}
         for key, new_value in step.data.items():
             original_value = data.get(key)
-            if new_value == original_value:
+            # Reset the value of any field that may have become hidden again after evaluating the logic
+            if original_value and get_component(configuration, key).get("hidden"):
+                data_diff[key] = defaults.get(key, "")
+                continue
```
sergei-maertens, line 119: shouldn't this only apply if `clearOnHide` is set?

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
SilviaAmAm: Good point

sergei-maertens (approved):

sergei-maertens:
Backported in c6024978387b678ead2e377ece0a583f7dcc9a12, d967887636be1955f7733ace64333a32e5808d7d and 0cf66cbcb6f0589c2ceef7728264d5a0c4bc66b8

PR #1553, opened 2022-04-29
[#1550] FormIO fix for currency/number component

### LINKED ISSUE #1550: Currency component breaks form builder

1.1, 1.0

In the form builder (in the admin) the column where all the components are present disappears and in the console the following error is present:

```
Error: The 'value' provided to Text Mask needs to be a string or a number. The value received was:

 [null]
```
The form definition can then no longer be edited.

1. Create a new form
2. Add a currency component and save it (this by default has `defaultValue=null`)
3. Change the currency component to have `multiple=true` and save the component (this changes the default to `defaultValue=[null]`)
4. Change the currency component back to `multiple=false` and save the component (this does not change the default back)
5. Observe the form builder break

The default value should be switched back to `null`? 

In any case the form builder shouldn't break

### Screen resolution

_No response_

### Browser

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1550

### DIFF
--- a/src/openforms/js/components/form/currency.js
+++ b/src/openforms/js/components/form/currency.js
@@ -1,7 +1,47 @@
-import {Formio} from "formiojs";
-import {defineCommonEditFormTabs} from "./abstract";
-import {DECIMAL_PLACES} from "../form/edit/components"
-import CurrencyEditData from 'formiojs/components/currency/editForm/Currency.edit.data';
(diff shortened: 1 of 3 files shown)


sergei-maertens:
Backported in 20bfa594f49ddb91cddd1975b06bdb54d21e21a7

Issue #1451, opened 2022-03-18
Labels of components hidden (direct or by logic) show up in summary


Bartvaderkin:
Taiga DH 10

The ticket specifically mentions the case where parts of the form are conditional but the labels of hidden components still show up and potentially confuse the end-user into assuming they missed something.

**Tasks**

Prototyping was done in #1549

- [x] Set up renderer with different modes
- [x] Emit (child) nodes that all know how to render themselves
- [x] Implement confirmation email (HTML, text) using the renderer
- [x] Implement a CLI renderer for debugging purposes
- [x] Implement a confirmation PDF renderer
- [x] Implement the registration email using the renderer
- [x] Implement exports using the renderer
- [x] Add the FormIO options in form builder for `showIn<RenderMode>`
- [x] Write management command/diagnostic to detect usage of deprecated template tags/functionalities
- [x] Developer documentation
- [x] Add render management command (default to `RenderModes.cli`, optionally specify a render mode)
- [x] Implement `showInSummary` in the SDK
- [x] Add total cost (if relevant) to confirmation PDF

joeribekker:
Discussed, see results below

![IMG_20220324_120510](https://user-images.githubusercontent.com/96970/159903594-e0513d8a-f6d7-405e-ba97-971b3e9b7834.jpg)

sergei-maertens:
Done :tada:

PR #1554, opened 2022-04-29
Set up plumbing for FormIO in submission rendering

sergei-maertens:
Related to #1451

This adds a specialized Formio component node subclass, capable of emitting its own child nodes (if present) and formatting label/value appropriately.

On top of that, it puts the registry in place to register specific subclasses for formio components, and sets up the Formio component properties to control output in different render modes. This includes setting a default value for existing forms that don't have these keys defined (yet).

Commented out code and TODOs are still intentional and will follow in later PRs.

### DIFF
--- a/src/openforms/formio/rendering/tests/test_component_node.py
+++ b/src/openforms/formio/rendering/tests/test_component_node.py
@@ -1,0 +1,318 @@
+from unittest.mock import patch
+
+from django.test import TestCase
+
+from openforms.forms.tests.factories import FormFactory
+from openforms.submissions.rendering import Renderer, RenderModes
+from openforms.submissions.tests.factories import (
+    SubmissionFactory,
+    SubmissionStepFactory,
+)
+
+from ..constants import RenderConfigurationOptions
+from ..nodes import ComponentNode
+from ..registry import Registry
+
+# from rest_framework.reverse import reverse
+
+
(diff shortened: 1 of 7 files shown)


sergei-maertens:
Codecov seems to be lost/have misplaced the base commit, but the coverage reported locally is 93%

src/openforms/formio/rendering/tests/test_component_node.py
```diff
@@ -0,0 +1,318 @@
...
+from openforms.submissions.tests.factories import (
+    SubmissionFactory,
+    SubmissionStepFactory,
+)
+
+from ..constants import RenderConfigurationOptions
+from ..nodes import ComponentNode
+from ..registry import Registry
+
+# from rest_framework.reverse import reverse
```
SilviaAmAm, line 16: Maybe forgotten commented out import? :upside_down_face:

SilviaAmAm (approved):

src/openforms/formio/rendering/tests/test_component_node.py
sergei-maertens: indeed! I've fixed that in the follow up PR :)

sergei-maertens:
@Bartvaderkin if you have any feedback still for this PR, I'll process it in a separate PR

PR #1556, opened 2022-05-02
Implement specific formio type nodes

sergei-maertens:
Relates to #1451

This PR builds on top of #1554 and implements component-type specifics for Formio components.

**Changes**

- Fieldset and columns type implemented: only render output if there are visible children.
- WYSIWYG: in non-HTML mode, output a plain-text variant of the HTML, and only output HTML in PDF render mode (for a complete summary of the entire form)
- Columns: never output a label

### DIFF
--- a/src/openforms/formio/rendering/default.py
+++ b/src/openforms/formio/rendering/default.py
@@ -1,0 +1,102 @@
+from typing import Iterator, Union
+
+from django.utils.safestring import SafeString, mark_safe
+
+from openforms.emails.utils import strip_tags_plus  # TODO: put somewhere else
+from openforms.submissions.rendering.constants import RenderModes
+
+from ..utils import iter_components
+from .conf import RENDER_CONFIGURATION
+from .nodes import ComponentNode
+from .registry import register
+
+
+class ContainerMixin:
+    is_layout = True
+
+    @property
+    def is_visible(self) -> bool:
+        # fieldset does not support the showInFoo properties, so we don't use the super
+        # class.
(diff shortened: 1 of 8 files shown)


src/openforms/formio/rendering/default.py
```diff
@@ -0,0 +1,104 @@
...
+        # class.
+        if self.mode == RenderModes.export:
+            return True
+
+        if self.component.get("hidden") is True:
+            return False
+
+        render_configuration = RENDER_CONFIGURATION[self.mode]
+        if render_configuration.key is not None:
+            assert render_configuration.key not in self.component, (
```
SilviaAmAm, line 29: I am a bit confused by this :thinking: 
The documentation says:
_a component is visible for a given render mode if the component-level configuration says so (while falling back to some defaults for older configurations)._

But in this case we check that there is no configuration on this component because its a container and so part of the exceptions?

sergei-maertens: yeah, we will not expose the checkboxes for "Show in (confirmation email|PDF)" for columns and fieldsets, as that logic is entirely determined by child elements being visible or not (either by logic or explicit configuration). This removes confusion for the form designers.

The superclass implementation implements `is_visible` based on this configuration defined in code. The containers (fieldset, columns) basically _ignore_ this configuration defined in code based on the above assumption. If someone were to add the visibility options to the fieldset|columns component in the future, then this assertion will fail and point out that they need to handle that logic in the overridden `is_visible` too.

PR #1561, opened 2022-05-02
Feature/1451 cli renderer

sergei-maertens:
**Changes**

Implemented a management command & CLI render mode for submissions.

### DIFF
--- a/src/openforms/submissions/management/commands/render_report.py
+++ b/src/openforms/submissions/management/commands/render_report.py
@@ -5,7 +5,7 @@ On top of that, it is also a realistic example of using the high-level Python AP
 the renderer to output reports.
 """
 import inspect
-from typing import Type, Union
+from typing import List, Type, Union
 
 from django.core.management import BaseCommand
 
@@ -88,6 +88,9 @@ class Command(BaseCommand):
         for node in renderer:
             lead = get_indent_level(node)
             if isinstance(node, ComponentNode):
+                # do not emit empty lines
+                if not node.label and not node.display_value:
+                    continue
                 # extract label + value for tabulate data
                 tabulate_data.append([node.label, node.display_value])
                 prev_node_type = ComponentNode
@@ -105,7 +108,7 @@ class Command(BaseCommand):
 
         self._print_tabulate_data(tabulate_data)
 
-    def _print_tabulate_data(self, tabulate_data) -> None:
+    def _print_tabulate_data(self, tabulate_data: List[List[str]]) -> None:
         if not tabulate_data:
             return
         table = tabulate(tabulate_data)
--- a/src/openforms/submissions/management/commands/render_report.py
+++ b/src/openforms/submissions/management/commands/render_report.py
@@ -1,0 +1,114 @@
+"""
+Management command to simulate the submission renderer outputting to a particular mode.
+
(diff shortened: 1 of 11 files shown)


src/openforms/submissions/management/commands/render_report.py
```diff
@@ -0,0 +1,114 @@
...
+            "--render-mode",
+            choices=[c[0] for c in RenderModes.choices],
+            default=RenderModes.cli,
+            help=f"Simulate a particular render mode. Defaults to {RenderModes.cli}.",
+        )
+
+    def handle(self, **options):
+        submissions = Submission.objects.filter(
+            id__in=options["submission_id"]
+        ).order_by("id")
```
SilviaAmAm, line 72: Might be good to mention in the docs that the submissions will not be printed out in the order in which the user specified them?

src/openforms/submissions/management/commands/render_report.py
```diff
@@ -0,0 +1,114 @@
...
+    def render_submission(self, submission: int, render_mode: str, as_html=False):
+        renderer = Renderer(submission=submission, mode=render_mode, as_html=as_html)
+
+        self.stdout.write("")
+        self.stdout.write(f"Submission {submission.id} - ", ending="")
+
+        prev_node_type, tabulate_data = None, []
+
+        for node in renderer:
+            lead = get_indent_level(node)
```
SilviaAmAm, line 89: So if I understand correctly, here we ignore the 'depth' of each node and indent only based on whether its a `SubmissionStepNode` or a `ComponentNode`?

src/openforms/submissions/management/commands/render_report.py
```diff
@@ -0,0 +1,114 @@
...
+            if isinstance(node, ComponentNode):
+                # extract label + value for tabulate data
+                tabulate_data.append([node.label, node.display_value])
```
SilviaAmAm, line 92: I find it a bit awkward that if both label and display value are `""` then there is an empty line in the printed submission (this happens for example for columns).

Maybe we could skip completely empty lines?

src/openforms/submissions/management/commands/render_report.py
```diff
@@ -0,0 +1,114 @@
...
+                    tabulate_data = []
+
+            if lead:
+                self.stdout.write(lead, ending="")
+            self.stdout.write(node.render())
+            prev_node_type = type(node)
+
+        self._print_tabulate_data(tabulate_data)
+
+    def _print_tabulate_data(self, tabulate_data) -> None:
```
SilviaAmAm, line 108: Missing type hint :upside_down_face:

SilviaAmAm (changes requested):

src/openforms/submissions/management/commands/render_report.py
sergei-maertens: I don't think that really matters - the ID of each submission is printed as part of the output so that should make it obvious enough?

src/openforms/submissions/management/commands/render_report.py
sergei-maertens: Yep! We basically discard that and display the whole component tree from Formio in a table with just two columns without any structure representation.

PR #1562, opened 2022-05-02
Feature/1451 pdf rendering

sergei-maertens:
**Changes**

Refactored the PDF generator to make use of the renderer.

This was already pretty close with the `Report` data class, so that has been stripped and replaced with Renderer implementation bits. Added a couple tests for the actual core issue reported (hidden field labels still being visible -> no longer the case now).

Preview:
image

### DIFF
--- a/src/openforms/submissions/templates/report/submission_report.html
+++ b/src/openforms/submissions/templates/report/submission_report.html
@@ -40,24 +40,27 @@
                 {% endcomment %}
             </section>
 
-            {% for step in report.steps %}
-                <div class="submission-step">
-                    <h2 class="subtitle">{{ step.title }}</h2>
+            {% for submission_step_node in report.renderer.get_children %}
+                {% if submission_step_node.has_children %}
+                    <div class="submission-step">
+                        <h2 class="subtitle">{{ submission_step_node.render }}</h2>
+                        {% for component_node in submission_step_node.get_children %}
+                            {% if component_node.label or component_node.display_value %}
+                                <div class="submission-step-row {% if component_node.layout_modifier %}submission-step-row--{{ component_node.layout_modifier }}{% endif %}">
 
-                    {% for item in step %}
-                        <div class="submission-step-row {% if item.layout_modifier %}submission-step-row--{{ item.layout_modifier }}{% endif %}">
+                                    <div class="submission-step-row__label">
+                                        {{ component_node.label }}
+                                    </div>
 
-                            <div class="submission-step-row__label">
-                                {{ item.label }}
-                            </div>
+                                    <div class="submission-step-row__value">
+                                        {{ component_node.display_value }}
+                                    </div>
 
-                            <div class="submission-step-row__value">
-                                {{ item.display_value }}
-                            </div>
-
-                        </div>
-                    {% endfor %}
-                </div>
+                                </div>
+                            {% endif %}
+                        {% endfor %}
+                    </div>
+                {% endif %}
             {% endfor %}
 
             {% if report.co_signer %}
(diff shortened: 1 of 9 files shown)


src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -40,24 +40,27 @@ <h1 class="title">{{ report.form.name }}</h1>
...
+                    <div class="submission-step">
+                        <h2 class="subtitle">{{ submission_step_node.render }}</h2>
+                        {% for component_node in submission_step_node.get_children %}
+                            {% if component_node.label or component_node.display_value %}
+                                <div class="submission-step-row {% if component_node.layout_modifier %}submission-step-row--{{ component_node.layout_modifier }}{% endif %}">
 
-                    {% for item in step %}
-                        <div class="submission-step-row {% if item.layout_modifier %}submission-step-row--{{ item.layout_modifier }}{% endif %}">
+                                    <div class="submission-step-row__label">
+                                        {{ component_node.label }}
```
SilviaAmAm, line 52: At the moment, fieldsets that have been with the option `hideHeader` checked will still show the header

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -67,6 +67,96 @@ def test_pdf_renders_without_errors(self):
...
+        submission = SubmissionFactory.from_components(
+            [
+                {
+                    "key": "input1",
+                    "label": "Input 1",
+                    "type": "textfield",
+                    "hidden": True,
+                }
+            ],
+            # a multiple email field with no value entered sends a [None]
```
SilviaAmAm, line 88: Is this comment meant to be here? :upside_down_face:

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -67,6 +67,96 @@ def test_pdf_renders_without_errors(self):
...
+                        }
+                    ],
+                },
+                {
+                    "type": "content",
+                    "html": "<p>Some markup</p>",
+                    "key": "content",
+                },
+            ],
+            # a multiple email field with no value entered sends a [None]
```
SilviaAmAm, line 143: Same remark as above

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -67,6 +67,96 @@ def test_pdf_renders_without_errors(self):
...
+        # hidden component itself may not be visible
+        self.assertNotIn("Input 1", html)
+        # step has no visible children -> also not visible
+        self.assertNotIn(step_title, html)
+
+    def test_visible_output_included(self):
+        """
+        Assert that hidden components are not included in the report.
+
+        #1451 requires that components hidden (statically or) dynamically by logic
```
SilviaAmAm, line 105: Should there maybe also be a test that fields hidden by logic also are not shown?

SilviaAmAm (changes requested):

src/openforms/submissions/tests/test_tasks_pdf.py
sergei-maertens: I don't think so - the low-level interaction of logic is tested in the Renderer/Node unit tests, and the effect is that they set the `hidden` property, which is simulated here with static definitions.

src/openforms/submissions/templates/report/submission_report.html
sergei-maertens: oops! good one to catch in the unit test for `FieldSetNode`

sergei-maertens: fixed in 02bdd3c92ead0771e91bdbe097b13ec23ab326bf

Issue #1560, opened 2022-05-02
Prefill fields in a column field not working


rbakels:
### Product versie / Product version

1.0.5

### Omschrijf het probleem / Describe the bug

When a field is contained in a column (markup) field, then the prefill logic is not triggered. 

I'm working on a custom prefill plugin and the get_prefill_values() method is not triggered when the field is contained in a column field. When the field is within a fieldset or just added to the canvas at a position, it is working properly. 

### Stappen om te reproduceren / Steps to reproduce

1. Add a column (50-50) to a form 
2. Add a field to the column and add a prefill option
3. Do the same for the right column
4. Open the form and notice the data isn't rendered.

### Verwacht gedrag / Expected behavior

The prefill logic is executed for fields contained in a column.

### Screen resolution

bigger

### Browser

Desktop

### OS

MacOS

### Browser

Safari

rbakels:
I try to work on a patch myself (I'm not a Python coder though). I'll let you know if I have some update.

sergei-maertens:
Thanks for the report, from the description I can already guess where it's going wrong. Nice catch :)

PR #1563, opened 2022-05-02
[#1560] Fix to execute prefill on fields within a column format field

rbakels:
Fields within a column field were not rendering the selected prefill options.

Fixes: #1560

Changes:
 - Execute prefill configuration within column format fields.

### DIFF
--- a/src/openforms/prefill/__init__.py
+++ b/src/openforms/prefill/__init__.py
@@ -77,7 +77,6 @@ def apply_prefill(configuration: JSONObject, submission: "Submission", register=
 
     fields = _extract_prefill_fields(configuration)
     grouped_fields = _group_prefills_by_plugin(fields)
-
     results = _fetch_prefill_values_cached(grouped_fields, submission, register)
 
     # finally, ensure the ``defaultValue`` is set based on prefill results
@@ -174,15 +173,26 @@ def _fetch_prefill_values(
 
 def _extract_prefill_fields(configuration: JSONObject) -> List[Dict[str, str]]:
     prefills = []
-    components = configuration.get("components", [])
 
-    for component in components:
-        if prefill := component.get("prefill"):
-            if prefill.get("plugin"):
-                prefills.append(prefill)
-        else:
-            nested = _extract_prefill_fields(component)
-            prefills += nested
+    if ("type" in configuration):
+        fieldType = configuration.get("type")
+
+        if (fieldType == "columns"):
+            columns = configuration.get("columns", [])
+
+            for column in columns:
+                nested = _extract_prefill_fields(column)
+                prefills += nested
+    else:
+        components = configuration.get("components", [])
+
+        for component in components:
+            if prefill := component.get("prefill"):
+                if prefill.get("plugin"):
+                    prefills.append(prefill)
+            else:
+                nested = _extract_prefill_fields(component)
+                prefills += nested
 
     return prefills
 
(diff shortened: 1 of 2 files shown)


src/openforms/prefill/__init__.py
```diff
@@ -174,15 +173,26 @@ def invoke_plugin(item: Tuple[str, List[str]]) -> Tuple[str, Dict[str, Any]]:
...
+    else:
+        components = configuration.get("components", [])
+
+        for component in components:
+            if prefill := component.get("prefill"):
+                if prefill.get("plugin"):
+                    prefills.append(prefill)
+            else:
+                nested = _extract_prefill_fields(component)
+                prefills += nested
```
sergei-maertens, line 196: * Every FormIO component should always have the `type` property - if it's not present and causes tests to fail, then the test data setup is wrong.
* There's some duplicated handling of the FormIO (nested datastructures) here, for which we have  a utility function: `openforms.formio.utils.iter_components`. You can make that recursively loop over the entire configuration/component tree. It already has logic for `columns` type components. This code should have been using that in the first place
* Typically we also add a regression test for bugfixes to show the problem and prove that it was fixed by the patch
* Code style (nitpicking): Python uses `snake_case` rather than `camelCase`

So, collecting all those remarks, I think a better approach is to implement this function as:

```py
from openforms.formio.utils import iter_components

prefills = []
for component in iter_components(configuration, recursive=True):
    if not (prefill := component.get("prefill")):
        continue
    if prefill.get("plugin"):
        prefills.append(prefill)
return prefills
```

And then add a test case in https://github.com/open-formulieren/open-forms/blob/master/src/openforms/prefill/tests/test_prefill_hook.py#L109 for the `columns` type. You can see some example configurations here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/rendering/tests/test_vanilla_formio_components.py#L110

src/openforms/prefill/__init__.py
```diff
@@ -230,6 +240,15 @@ def _set_default_values(
...
     """
+    if "type" in configuration:
+        fieldType = configuration.get("type")
+
+        if fieldType == "columns":
+            columns = configuration.get("columns", [])
+
+            for column in columns:
+                _set_default_values(column, prefilled_values)
+
```
sergei-maertens, line 251: I think this can also be refactored to instead use `iter_components`

src/openforms/utils/tests/html_assert.py
```diff
@@ -15,14 +15,14 @@ def assertTagWithTextIn(self, tag, text, document_str):
         """
         check for html tags and their content while ignoring tag attributes
         """
-        if not re.search(fr"<{tag}(?: [^>]*)?>\s*{text}", document_str):
+        if not re.search(rf"<{tag}(?: [^>]*)?>\s*{text}", document_str):
```
sergei-maertens, line 18: this looks like an unexpected change - are you using a black version that's newer than the one pinned in the dev dependencies?

src/openforms/utils/tests/html_assert.py
```diff
@@ -15,14 +15,14 @@ def assertTagWithTextIn(self, tag, text, document_str):
...
             self.fail(f"cannot find <{tag}..>{text} in: {document_str}")
 
     def assertNotTagWithTextIn(self, tag, text, document_str):
         """
         check for html tags and their content while ignoring tag attributes
         """
-        if re.search(fr"<{tag}(?: [^>]*)?>\s*{text}", document_str):
+        if re.search(rf"<{tag}(?: [^>]*)?>\s*{text}", document_str):
```
sergei-maertens, line 25: this looks like an unexpected change - are you using a black version that's newer than the one pinned in the dev dependencies?

sergei-maertens (changes requested):

src/openforms/prefill/__init__.py
rbakels: Thanks for the comments. I will update the PR. 

:)

sergei-maertens:
@rbakels - we haven't seen any activity in a bit, so I'll take over this patch from you to get it in the next releases :-)

sergei-maertens:
Followed up in #1619

PR #1569, opened 2022-05-04
[#1568] Components in submission mismatch definition

### LINKED ISSUE #1568: Logic check error: 'NoneType' object has no attribute 'get'

1.0

```
AttributeError: 'NoneType' object has no attribute 'get'
  File "rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "openforms/submissions/api/viewsets.py", line 410, in logic_check
    new_configuration = evaluate_form_logic(
  File "elasticapm/traces.py", line 871, in decorated
    return func(*args, **kwds)
  File "openforms/submissions/form_logic.py", line 119, in evaluate_form_logic
    if component.get("hidden") and component.get("clearOnHide"):
```

I managed to reproduce this as follows:

Create a form with a step with 2 components:

- text field A
- text field B

Save the form.

As a user fill in the form: 
- Fill in a value for both fields.
- Then click on 'next'

As an admin, change the form definition: remove text field A

As a user, click in the form on 'back to the previous page'.

The logic check fails because there is no longer the component in the definition but there is still the component in the submission data.

_No response_

### Screen resolution

_No response_

### Browser

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1568

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -126,7 +126,11 @@ def evaluate_form_logic(
             # Reset the value of any field that may have become hidden again after evaluating the logic
             if original_value:
                 component = get_component(configuration, key)
-                if component.get("hidden") and component.get("clearOnHide"):
+                if (
+                    component
+                    and component.get("hidden")
+                    and component.get("clearOnHide")
+                ):
                     data_diff[key] = defaults.get(key, "")
                     continue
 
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):

sergei-maertens:
Backported in 987677973effcc7a361c7780c74617576cfc61ac and 5ca0f55c775441bc3a6018c852d00505e723f330

PR #1572, opened 2022-05-04
Remove submission get printable data

sergei-maertens:
Removed the `Submission.get_printable_data` method, which was used for data display in confirmation/registration e-mails (amongst others). All this has been replaced by `openforms.submissions.rendering.renderer.Renderer`.


src/openforms/registrations/contrib/email/templates/emails/email_registration.txt
```diff
@@ -9,8 +9,14 @@ Our reference: {{ public_reference }}
...
 {% endblocktrans %}
 {% endif %}
-{% for key, value in submitted_data %}
-- {{ key }}: {% display_value value %}{% endfor %}{% endautoescape %}
+
+{% for submission_step_node in renderer.get_children %}
+{{ submission_step_node.render }}
+
+{% for component_node in submission_step_node.get_children %}
+{% whitespace component_node.depth base=' ' %}- {{ component_node.label }}{% if not component_node.is_layout %}:{% endif %} {{ component_node.display_value }}{% endfor %}
```
SilviaAmAm, line 20: Also here check for empty label/value like in html template?

SilviaAmAm (changes requested):

src/openforms/registrations/contrib/email/templates/emails/email_registration.txt
sergei-maertens: Updated the PR #1571 for that!

PR #1581, opened 2022-05-09
Refactor submission export to use renderer

sergei-maertens:
Related to #1451

Submission exports are now done based on the submission renderer. All defined form components are always included, even for hidden fields.


src/openforms/submissions/exports.py
```diff
@@ -25,23 +31,44 @@ class ExportFileTypes:
...
+
+    .. note:: the queryset of submissions must all be of the same form!
+    """
+    # queryset *could* be empty
+    if not queryset:
+        return tablib.Dataset()
+
+    first_submission = queryset[0]
+    headers = ["Formuliernaam", "Inzendingdatum"]
+    for component_node in iter_submission_data_nodes(first_submission):
```
SilviaAmAm, line 55: I think there could be problems in the case where the definition of a form changes, and an admin tries to export multiple submissions of the same form (but that have different definitions).

SilviaAmAm (changes requested):

src/openforms/submissions/exports.py
sergei-maertens: hrm, good point, let's see...

1. Form fields have been removed that are present in submissions -> that's okay, apparently they weren't relevant otherwise they wouldn't be removed
2. Form fields were added that are missing in old submissions -> should not be a problem, export contains more headers, it's as if those fields were not filled out
3. Form fields have changed component keys - in that case, the downstream systems processing it will have issues too. Not sure if we can/should support this. And maybe variables may be the solution for this?

@joeribekker thoughts?

src/openforms/submissions/tests/test_exports.py
```diff
@@ -0,0 +1,118 @@
...
+from ..exports import create_submission_export
+from ..models import Submission
+from .factories import SubmissionFactory
+
+
+class ExportTests(TestCase):
+    @freeze_time("2022-05-09T13:00:00Z")
+    def test_complex_formio_configuration(self):
+        """
+        Assert that complex formio configurations with logic are exported correctly.
```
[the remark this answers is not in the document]
sergei-maertens: updated!

src/openforms/submissions/exports.py
```diff
@@ -7,7 +8,12 @@
 from lxml import etree
 from tablib.formats._json import serialize_objects_handler
 
+from openforms.submissions.models.submission import Submission
```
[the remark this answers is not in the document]
sergei-maertens: this is why I hate auto-import in editors

src/openforms/submissions/exports.py
sergei-maertens: Discussed with Joeri - we'll see if/when someone complains about this and #1586 is created to track this.

PR #1583, opened 2022-05-09
Add management command to check for template tag usage

sergei-maertens:
Should also be easily copy-pasteable in running production instances without deploying 1.1.0, to get a clear picture.

Related to #1451


sergei-maertens:
Checked prod instances, it's not used anywhere

PR #1589, opened 2022-05-10
add translation for validation error key

### LINKED ISSUE #1344: SDK file and time components with required and multiple show error code when invalid
- [x] If I have a required fileupload component with multiple=True but don't upload a file I get an error code string '`array_nonempty`' as part of the error message:

    `Dit veld is verplicht.`
    `array_nonempty`

- [x] The same message shows up with a time component, also when required and multiple and nothing entered.

sergei-maertens:
Fixes #1344

Formio runs both the multiple and required validators, leading to
multiple error messages being shown, which is not ideal.

image


SilviaAmAm (approved):

sergei-maertens:
Backported in 7351fb048e677d2df34a82b59ccddf3182b3e18d

PR #1597, opened 2022-05-12
Manage session expiry admin

### LINKED ISSUE #1308: Admin Form builder shows ambiguous error if user has been logged out
Whenever I am editing a form in the admin and I take too long, when I try to save I get the error:

Screenshot from 2022-02-07 15-04-19

While in the console it shows that the PUT to `/api/v1/forms/<form-uuid>` returns a 401 not authenticated response.

**Tasks**

- [x] Show a proper/clear error if an HTTP 401 is returned instead of the generic error
- [x] Display a timer indicating how long/when the session expires. We can use the `X-Session-Expires-In` response header like in the SDK
- [x] Show a modal if the session is close to expire with the option to extend the session
    - [x] Implement an `/api/v1/ping` endpoint for this purpose (require authenticated user or active submission session users), response is HTTP 204

sergei-maertens:
Fixes #1308

**Session expired (timer)**

image

**Endpoint returned HTTP 401**

image

### DIFF
--- a/src/openforms/js/components/admin/SessionStatus.js
+++ b/src/openforms/js/components/admin/SessionStatus.js
@@ -13,6 +13,7 @@ import {
 } from 'react-intl';
 
 import { sessionExpiresAt } from '../../utils/session-expiry';
+import { apiCall } from '../../utils/fetch';
 import FormModal from './FormModal';
 import ActionButton, {SubmitAction} from './forms/ActionButton';
 import SubmitRow from './forms/SubmitRow';
@@ -236,10 +237,9 @@ const SessionStatus = () => {
                         description: 'Extend session button',
                         defaultMessage: 'Stay logged in',
                     })}
-                    onClick={event => {
+                    onClick={async (event) => {
                         event.preventDefault();
-                        console.log('Ok');
-                        // TODO
+                        await apiCall('/api/v1/ping');
                     }}
                 />
             </>
--- a/src/openforms/js/components/admin/SessionStatus.js
+++ b/src/openforms/js/components/admin/SessionStatus.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+const SessionStatus = () => {
+    return (
+        'SessionStatus'
+    );
+};
+
+SessionStatus.propTypes = {
+};
+
+
+export default SessionStatus;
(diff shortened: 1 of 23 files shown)


src/openforms/js/utils/session-expiry.js
```diff
@@ -0,0 +1,38 @@
...
+
+const SessionExpiresInHeader = 'X-Session-Expires-In';
+
+let sessionExpiresAt = createGlobalstate({
+    date: null,
+    numSeconds: null,
+    authFailure: false,
+});
+
+const updateSesionExpiry = (seconds, authFailed=false) => {
```
SilviaAmAm, line 12:
```suggestion
const updateSessionExpiry = (seconds, authFailed=false) => {
```

src/openforms/js/components/admin/SessionStatus.js
```diff
@@ -0,0 +1,283 @@
...
+        showWarning,
+        () => {
+            setShowWarning(false);
+            reset();
+        },
+    ];
+};
+
+
+const SessionStatus = () => {
```
SilviaAmAm, line 127: I find this part a bit tricky to follow. If I understand correctly:
- `SessionStatus` looks at the global state (that is set by the `onResponseHook` in the `fetch` function) and sees in how long the session will expire.
- It sets a timeout (but I am not too clear what will run at the end of this timeout? Is it the `useTimeoutFn` in `useTriggerWarning`?)
- `useTriggerWarning` checks if warnings should be shown based on the time to expiry
- If the expiry date changes then the timeout is reset
- Based on `authFailure`, `isExpired`, `showWarning`, the right message is shown.

sergei-maertens: > I find this part a bit tricky to follow. If I understand correctly:
> 
> * `SessionStatus` looks at the global state (that is set by the `onResponseHook` in the `fetch` function) and sees in how long the session will expire.

Correct, and it also subscribes to global state updates. If the global state changes, the component re-renders.

> * It sets a timeout (but I am not too clear what will run at the end of this timeout? Is it the `useTimeoutFn` in `useTriggerWarning`?)

The `useTimeout` hook from `react-use` triggers a component re-render after the timeout. This is needed to re-calculate the derived booleans `isExpired` etc., because a fresher value for `now` is then used while the original `date` is still the same timestamp.

> * `useTriggerWarning` checks if warnings should be shown based on the time to expiry

Correct, this essentially does the same except it runs some time before the `useTimeout` that notifies you that the session _has_ expired. The warning notifies you that it _will_ expire.

> * If the expiry date changes then the timeout is reset

Correct

> * Based on `authFailure`, `isExpired`, `showWarning`, the right message is shown.

Correct

SilviaAmAm: I see! This is clear now, thank you :slightly_smiling_face:

PR #1598, opened 2022-05-12
update nginx location rules for fileuploads

### LINKED ISSUE #1593: Nginx upload path does not match API path
https://github.com/open-formulieren/open-forms/blob/a273ad7d0b29854f93640afba70cb16bc0719e48/deployment/templates/file-uploads.conf.tpl#L1

Hence, the 10 MB limit (in case of the screenshot) is used instead of the exception of 50 MB (again, in this screenshot) for that path.

image

sergei-maertens:
Fixes #1593

The #1453 fix/refactor moved the endpoint to be Formio specific,
which was forgotten in the nginx config.

### DIFF
--- a/deployment/templates/file-uploads.conf.tpl
+++ b/deployment/templates/file-uploads.conf.tpl
@@ -1,3 +1,23 @@
+location = /api/v1/formio/fileupload {
+    client_max_body_size {{ of_max_upload_size|mandatory }};
+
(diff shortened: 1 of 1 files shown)


sergei-maertens:
Backported in f85c301ff260addb4821ee85ff840ea2843d1fdc

Issue #1325, opened 2022-02-10
Adding variables to forms


SilviaAmAm:
It should be possible to define variables in the Admin form builder associated with a form:

- It should be possible to give these variables initial values when building the form. 
    - The values can have different types: numbers, strings, dates.
- The values in the variables can be used in the logic rules
- The values of the variables can be updated by the logic rules
- It should be possible to use the value of a field in the label of another field, like for example a new field could have the label 'email address of {{ surname }}'.

When the form is filled in by the user, these variables will be included in the data that is sent to the chosen registration backend.
It should also be possible to include them in a submission report (PDF) and in the confirmation email. 

**Problems with using hidden field**
- You have to choose a particular component with a specific type and then mark it as 'hidden'. 
- Each component has too many options that are not needed in the case of a variable.
- You then end up with many hidden component intertwined with visible components in the field without a clear differentiation.
- If you have many variables in a step, then you end up having a list of fields that is too long.

joeribekker:
Ok, I understand what they want. They basically want to sent arbitrary data to the registration backend. I think it's possible.

sergei-maertens:
Been analyzing this for a bit and came up with this initial technical specification

---

# Open Forms variables feature

Variables are an abstraction layer between the form definition and filled out values.
If implemented correctly, variables can be used in a number of places:

- Form component values
- Form component metadata (label, help text, content...)
- Calculated values from input data and/or other variables
- Logic (both reading from and writing to variables)

There is potential to leverage variables to manage client-side translations (e.g. by
attaching a language metadata element to a variable combined with the client requested
language).

## Requirements

* Define variable for every (input) component (automatically?)
* Define static variables with (initial) value
* Reference variable values as input for logic expressions
* Reference variables as target for logic actions
* Specify value output of a component to a (particular) variable
* Ability to handle collections (lists with indices) in a variable and/or collections
  of variables
* Reference and interpolate variables as part of component definitions (label, help
  text, HTML/WYSIWYG content...)
* Attach metadata to variables, such as:
    - user input vs. calculated result
    - variable origin (form component vs. explicitly defined)
    - type information (infer from component or explicitly define)
    - set/modified timestamps?
* Manage variables at the form-level in a dedicated tab
* Include form variables in imports/exports & version history
* Extract (nested) data from evaluation sources, e.g. DMN outputs a result -> point to
  a sub key

**Nice to have**

* Track variable usage/dependency tree, combined with order of components/logic rules +
[...]

PR #1604, opened 2022-05-13
[#1325 ] FormVariable backend

SilviaAmAm:
Part of #1325


src/openforms/forms/models/form_variable.py
```diff
@@ -0,0 +1,107 @@
...
+        on_delete=models.CASCADE,
+        null=True,
+    )
+    name = models.CharField(
+        verbose_name=_("name"),
+        help_text=_("Name of the variable"),
+        max_length=100,
+    )
+    # TODO Ideally should be namespaced with the form slug. Should add validator for format <form_slug>:<key>?
+    # and the key part should be of slug format
```
SilviaAmAm, line 43: I think I changed my mind about this while looking at the front end. I will add a unique_together constraint with for key/form and add the validation to the list serializer (https://github.com/encode/django-rest-framework/issues/6395)

src/openapi.yaml
```diff
@@ -869,6 +869,52 @@ paths:
           headers:
             X-Session-Expires-In:
               $ref: '#/components/headers/X-Session-Expires-In'
+  /api/v1/forms/{form_uuid_or_slug}/variables/bulk_update:
```
sergei-maertens, line 872:
```suggestion
  /api/v1/forms/{form_uuid_or_slug}/variables:
```

would probably be a nicer endpoint capturing the resource semantics?

src/openforms/forms/api/viewsets.py
```diff
@@ -517,3 +521,47 @@ def post(self, request, *args, **kwargs):
...
+        form_variables = self.get_queryset()
+        # We expect that all the variables that should be associated with a form come in the request.
+        # So we can delete any existing variables because they will be replaced.
+        form_variables.delete()
+
+        serializer = self.get_serializer(data=request.data, many=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
sergei-maertens, line 567: will have to be a HTTP_200, the 201 semantics really only apply to a single resource rather than a collection of resources (for example, DRF emits the `Location` header of the newly created object, which doesn't make any sense here)

src/openforms/forms/constants.py
```diff
@@ -31,3 +31,23 @@ class SubmissionAllowedChoices(DjangoChoices):
     no_without_overview = ChoiceItem(
         "no_without_overview", _("No (without overview page)")
     )
+
+
+class FormVariablesSources(DjangoChoices):
```
sergei-maertens, line 36:
```suggestion
class FormVariableSources(DjangoChoices):
```

src/openforms/forms/constants.py
```diff
@@ -31,3 +31,23 @@ class SubmissionAllowedChoices(DjangoChoices):
...
+    component = ChoiceItem("component", _("Component"))
+    user_defined = ChoiceItem("user_defined", _("User defined"))
+    static = ChoiceItem("static", _("Static"))
+
+
+class FormVariablesDataTypes(DjangoChoices):
```
sergei-maertens, line 42:
```suggestion
class FormVariableDataTypes(DjangoChoices):
```

src/openforms/forms/constants.py
```diff
@@ -31,3 +31,23 @@ class SubmissionAllowedChoices(DjangoChoices):
...
+    string = ChoiceItem("string", _("String"))
+    null = ChoiceItem("null", _("Null"))
+    boolean = ChoiceItem("boolean", _("Boolean"))
+    object = ChoiceItem("object", _("Object"))
+    array = ChoiceItem("array", _("Array"))
+    number = ChoiceItem("number", _("Number"))
+    datetime = ChoiceItem("datetime", _("Datetime"))
+
+
+class FormVariablesInitialValues(DjangoChoices):
```
sergei-maertens, line 52:
```suggestion
class FormVariableStaticInitialValues(DjangoChoices):
```

src/openforms/forms/models/form_variable.py
```diff
@@ -0,0 +1,105 @@
...
+    FormVariablesDataTypes,
+    FormVariablesInitialValues,
+    FormVariablesSources,
+)
+from .form import Form
+from .form_definition import FormDefinition
+
+
+def get_today():
+    return timezone.now().isoformat()
```
sergei-maertens, line 15: confusing - `today` implies a `date`, but `timezone.now()` returns a `datetime`. Also, please add type hints :-)

src/openforms/forms/models/form_variable.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    key = models.SlugField(
+        verbose_name=_("key"),
+        help_text=_("Key of the variable, should be unique with the form."),
+        max_length=100,
+    )
+    source = models.CharField(
+        verbose_name=_("source"),
+        help_text=_(
+            "Where will the data that will be associate with this variable come from"
```
sergei-maertens, line 50:
```suggestion
            "Where will the data that will be associated with this variable come from"
```

src/openforms/forms/models/form_variable.py
```diff
@@ -0,0 +1,105 @@
...
+        help_text=_(
+            "The format of the value that will be associated with this variable"
+        ),
+        blank=True,
+        max_length=250,
+    )
+    is_sensitive_data = models.BooleanField(
+        verbose_name=_("is sensitive data"),
+        help_text=_("Will this variable be associated with sensitive data?"),
+        default=True,
```
sergei-maertens, line 86:
```suggestion
        default=False,
```
and then consider setting it automatically from the component config/defaults in the UI?

src/openforms/forms/tests/variables/test_initial_values.py
```diff
@@ -0,0 +1,34 @@
...
+class FormVariablesTests(TestCase):
+    def test_get_initial_value(self):
+        static_variable = FormVariableFactory.create(
+            source=FormVariablesSources.static,
+            initial_value=FormVariablesInitialValues.today,
+        )
+        with self.subTest(part="static variable today"):
+            with freeze_time("2021-07-16T21:15:00Z"):
+                self.assertEqual(
+                    timezone.now().isoformat(), static_variable.get_initial_value()
```
sergei-maertens, line 19: use the actual literal expected value in tests, instead of duplicating the implementation. `timezone.now().isoformat()` is too much thinking to figure out what the expected result is. `"2021-07-16T21:15:00Z"` is immediately obvious.

src/openforms/forms/models/form_variable.py
```diff
@@ -0,0 +1,105 @@
...
+        help_text=_(
+            "Which attribute from the prefill response should be used to fill this variable"
+        ),
+        blank=True,
+        max_length=50,
+    )
+    data_type = models.CharField(
+        verbose_name=_("data type"),
+        help_text=_("The type of the value that will be associated with this variable"),
+        choices=FormVariablesDataTypes.choices,
```
SilviaAmAm, line 72: I am kind of undecided if this should have the type of the formio component and use something like https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/camunda/type_mapping.py#L44 to convert values :thinking:

src/openforms/forms/tests/variables/test_viewset.py
```diff
@@ -0,0 +1,164 @@
...
+
+        self.assertEqual(3, variables.count())
+        self.assertTrue(variables.filter(key=another_form_variable.key).exists())
+
+        form_variables = variables.filter(form=form)
+
+        self.assertEqual(2, form_variables.count())
+        self.assertTrue(form_variables.filter(key="today").exists())
+        self.assertTrue(form_variables.filter(key=form_variable1.key).exists())
+        self.assertFalse(form_variables.filter(key=form_variable2.key).exists())
```
sergei-maertens, line 110: here to - please use literal values instead of DRY-ing your code. Tests should basically throw out all the DRY-principles to make them obvious.

src/openforms/forms/models/form_variable.py
sergei-maertens: the camunda type mapping approach was a bandaid - ideally, that belongs in `openforms.formio` as part of the #1068.

So, if that sounds favourable, let's make work of it and sort out the Camunda registration backend as well to make use of it.

sergei-maertens: OTOH - for user-defined variables, we cannot infer the data type and will need that information from the form designers. This is also tackled to some degree in Camunda registration backend: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/camunda/complex_variables.py

src/openapi.yaml
SilviaAmAm: I am not entirely sure how to go about this (because it's a nested endpoint) :thinking:

src/openapi.yaml
```diff
@@ -869,6 +869,52 @@ paths:
...
+          multipart/form-data:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/FormVariable'
+        required: true
+      security:
+      - tokenAuth: []
+      - cookieAuth: []
+      responses:
```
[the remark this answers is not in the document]
SilviaAmAm: The only other endpoint that I could find that implements this is https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/api/viewsets.py#L141

I am unsure as to what is needed: for 400 it's a Error400Serializer with the usual fields `code`, `detail`, `invalidParams`...?
And similar for the other errors but with serializers containing the fields that are usually returned with those errors?

Or have I completely misunderstood?

src/openforms/forms/models/form_variable.py
SilviaAmAm: Yes :sweat_smile: I made it a datetime to avoid the mess of timezones and dates... maybe instead of 'today' i should call it 'now' or something like this?

src/openforms/forms/models/form_variable.py
SilviaAmAm: Mhmm yes indeed, and then if we replace Formio it would be better if the variables don't depend on the component types used in Formio :thinking: So maybe the frontend should figure out the type of the component and set the `data_type` accordingly

src/openforms/forms/api/viewsets.py
```diff
@@ -413,6 +433,53 @@ def admin_message(self, request, *args, **kwargs):
...
+            data=request.data, many=True, context={"request": request, "form": form}
+        )
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    @extend_schema(
+        summary=_("List form variables"),
+        description=_("List all FormVariables related to a form."),
```
sergei-maertens, line 460:
```suggestion
        description=_("List all variables defined for a form."),
```

src/openforms/forms/api/viewsets.py
```diff
@@ -233,6 +236,23 @@ def configuration(self, request: Request, *args, **kwargs):
         ),
         parameters=[UUID_OR_SLUG_PARAMETER],
     ),
+    variables_bulk_update=extend_schema(
+        summary=_("Bulk update/create/delete form variables"),
```
sergei-maertens, line 240:
```suggestion
        summary=_("Bulk configure form variables"),
```

src/openforms/forms/models/form_variable.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    is_sensitive_data = models.BooleanField(
+        verbose_name=_("is sensitive data"),
+        help_text=_("Will this variable be associated with sensitive data?"),
+        default=False,
+    )
+    initial_value = models.JSONField(
+        verbose_name=_("initial value"),
+        help_text=_("The initial value for this field"),
+    )
```
sergei-maertens, line 102:
```suggestion
    initial_value = models.JSONField(
        verbose_name=_("initial value"),
        help_text=_("The initial value for this field"),
    )
```
I think you need a `blank=True` here for the admin interface

sergei-maertens (changes requested):
Couple small tweaks and then we're good to go

src/openforms/forms/models/form_variable.py
SilviaAmAm: I tried it out and indeed I needed a `blank=True` but also a `null=True`.
And I also missed a `blank=True` for the form_definition FK :sweat_smile:

SilviaAmAm:
In the end I made almost everything read only in the admin and removed the delete button for static/component variables

sergei-maertens (approved):

PR #1612, opened 2022-05-16
Deprecated Submission.get_next_step as it's not actually used

sergei-maertens:
To be removed when we go for `2.0` together with all other deprecated code.

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -4651,6 +4651,7 @@ components:
           format: uri
           readOnly: true
           nullable: true
+          deprecated: true
         submissionAllowed:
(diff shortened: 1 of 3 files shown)


SilviaAmAm (approved):
We will also need to remember to remove it from https://github.com/open-formulieren/open-forms-sdk/blob/main/src/types/Submission.js#L11

sergei-maertens:
> We will also need to remember to remove it from https://github.com/open-formulieren/open-forms-sdk/blob/main/src/types/Submission.js#L11

Correct, came across it when checking if it was actually used or not.

PR #1619, opened 2022-05-17
Fix prefill fields in columns not working

### LINKED ISSUE #1560: Prefill fields in a column field not working

sergei-maertens:
Fixes #1560

Refactored prefill extract/set-values to use `openforms.formio.utils.iter_components` so that we get all components in the tree (and thus all layout variants).

### DIFF
--- a/src/openforms/formio/typing.py
+++ b/src/openforms/formio/typing.py
@@ -5,7 +5,10 @@ Formio components are JSON blobs adhering to a formio-specific schema. We define
 (subsets) of these schema's to make it easier to reason about the code operating on
 (parts of) the schema.
 """
+# TODO: on python 3.11+ we can use typing.NotRequired to mark keys that may be absent
 from typing import TypedDict
+
+from openforms.typing import JSONValue
 
 
 class OptionDict(TypedDict):
@@ -18,6 +21,11 @@ class OptionDict(TypedDict):
 
     value: str
     label: str
+
+
+class PrefillConfiguration(TypedDict):
+    plugin: str
+    attribute: str
 
 
 class Component(TypedDict):
@@ -40,3 +48,5 @@ class Component(TypedDict):
     label: str
     multiple: bool
     hidden: bool
+    defaultValue: JSONValue
+    prefill: PrefillConfiguration
--- a/src/openforms/prefill/__init__.py
+++ b/src/openforms/prefill/__init__.py
@@ -37,6 +37,7 @@ import elasticapm
 from glom import GlomError, Path, assign, glom
 from zgw_consumers.concurrent import parallel
 
+from openforms.formio.utils import iter_components
 from openforms.logging import logevent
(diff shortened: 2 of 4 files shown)


src/openforms/formio/typing.py
```diff
@@ -5,7 +5,9 @@
 (subsets) of these schema's to make it easier to reason about the code operating on
 (parts of) the schema.
 """
-from typing import TypedDict
+from typing import Any, Optional, TypedDict
```
SilviaAmAm, line 8: Any and Optional are unused I think

src/openforms/formio/typing.py
```diff
@@ -40,3 +42,4 @@ class Component(TypedDict):
     label: str
     multiple: bool
     hidden: bool
+    defaultValue: JSONValue
```
SilviaAmAm, line 45: Should we maybe also add the `prefill` field, since it is used  in `_set_default_values` later?

src/openforms/prefill/__init__.py
```diff
@@ -230,27 +226,27 @@ def _set_default_values(
...
-        ).get(configuration["prefill"]["attribute"])
+    for component in iter_components(configuration, recursive=True):
+        if not (prefill := component.get("prefill")):
+            continue
+        if not (plugin := prefill.get("plugin")):
+            continue
+        if not (attribute := prefill.get("attribute")):
+            continue
 
+        default_value = configuration.get("defaultValue")
```
SilviaAmAm, line 237: I think this should be
```suggestion
        default_value = component.get("defaultValue")
```

SilviaAmAm (changes requested):

src/openforms/formio/typing.py
sergei-maertens: my new linter doesn't make it that obvious anymore :<

src/openforms/formio/typing.py
sergei-maertens: I'm adding them, but the type checks are bit inconsistent - as these keys may be absent. You can signal that with `typing.NotRequired`, but that's only included in Python 3.11+

sergei-maertens:
Backported in 2370cf057d782a466486f6a5890f652415e7d9c5 and 313d6986428b45700c1b0d271e18ac7db9aa1be2

PR #1623, opened 2022-05-17
[#1621] Added Ctrl+S keyboard shortcut to Save & Continue to form changeform

### LINKED ISSUE #1621: Add usable save action to form designer for usability on big forms
### Thema / Theme

Form designer

Form designer user noted a usability issue when editing complex/tall forms where they have to scroll a long distance to reach the current 'Save and Continue' button.

It was mentioned to consider a floating Save button or a keyboard shortcut, ~~but Joeri proposed the shortcut, possibly because a floating element is bound to block some other UI elements.~~ *Shortcuts are problematic, we do a sticky element*

Via Taiga DH 40

- [x] ~~Ctrl+S shortcut to 'Save & Continue'~~
- [x] ~~Add to manual (add shortcuts section)~~
- [x] Use CSS solution to make the submit-row sticky in view

Improved usability in tall forms.

_No response_

Bartvaderkin:
Fixes #1621


sergei-maertens (changes requested):
Inline scripts are to be avoid, especially if/when we enable CSP in the admin.

Please move this to a new JS module somewhere in https://github.com/open-formulieren/open-forms/tree/master/src/openforms/js/components/admin and ensure it is imported/loaded in https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/index.js

Bartvaderkin:
I replaced the template based solution with the suggestion.

src/openforms/js/components/admin/Keyboard.js
```diff
@@ -0,0 +1,25 @@
...
+        event.preventDefault();
+        button.click();
+    }
+}
+
+const enableKeyboardShortcuts = () => {
+    window.addEventListener('DOMContentLoaded', (e) => {
+        document.addEventListener('keydown', event => {
+            // add a Ctrl+S keyboard listener as trigger for Django's standard "Save & Continue" button
+            if (event.ctrlKey && event.key === 's') {
```
sergei-maertens, line 18: I think this doesn't capture command key on Mac, see https://stackoverflow.com/a/3922353 - I think you also want to check `event.metaKey`

sergei-maertens (changes requested):
As discussed this morning, a "shortcuts" entry in the documentation is still missing.

src/openforms/js/components/admin/Keyboard.js
```diff
@@ -0,0 +1,25 @@
...
+        console.warn("Cannot find 'Save & Continue' button");
+    } else {
+        event.preventDefault();
+        button.click();
+    }
+}
+
+const enableKeyboardShortcuts = () => {
+    window.addEventListener('DOMContentLoaded', (e) => {
+        document.addEventListener('keydown', event => {
```
sergei-maertens, line 16: I'm also wondering if we shouldn't tap into `keyup` rather than `keydown`? That triggers the save when the keys are released.

sergei-maertens, line 16: I'm also wondering if we shouldn't tap into `keyup` rather than `keydown`? That triggers the save when the keys are released.

docs/manual/forms/shortcuts.rst
```diff
@@ -0,0 +1,13 @@
+.. shortcuts:
```
sergei-maertens, line 1:
```suggestion
.. _manual_forms_shortcuts:
```

sergei-maertens (changes requested):

sergei-maertens (approved):
Shiny

Issue #1630, opened 2022-05-19
Evaluate solution to block concurrent editing of Forms


Bartvaderkin:
### Thema / Theme

Form designer

### Omschrijving / Description

Users requested a solution to block multiple form designers from editing the same form and overwriting eachothers changes.

I proposed to try something simple like https://pypi.org/project/django-admin-edit-lock/ and @joeribekker asked to try this and see if it works for us.

Via Taiga DH 46

### Added value / Toegevoegde waarde

Improves collaborative experience.

Bartvaderkin:
I tried the django-admin-edit-lock package and while it works as described it isn't sufficient or ideal.

- First problem it only locks a single model admin and not related objects, so in our case we'd block the Form but other users would still be able to access the FormDefinitions though their own admins.
- Locking the Form and all is related FormDefinitions is not ideal as in practice this means a workflow where Forms often re-use shared FormDefintions can get annoying.

@joeir suggested it'd be better to setup a system that would put warning messages instead of locks, so user are informed others are accessing same data and can coordinate among themselves about the shared resources.

sergei-maertens:
Rough lines for initial design

* [ ] Set up a store to keep track of who is editing what (redis-based store)
    * [ ] End-user browser pings backend "now" with form/step that's being accessed
    * [ ] Backend sets expiry slightly larger than ping frequency (-> configure single source of truth for ping interval?)
    * [ ] Response contains messages about other users operating in the same space
    * [ ] Redis messages should use optimistic locking: https://redis.io/docs/manual/transactions/#cas depending on how we structure the keys
* [ ] Maybe reconsider websockets for push events?

PR #1635, opened 2022-05-20
[#1630] Added django-admin-edit-lock to FormAdmin

Bartvaderkin:
For evaluation purposes of #1630


sergei-maertens:
We are not continuing in this direction, so I'm closing this PR.

PR #1638, opened 2022-05-20
[#1621] Sticky CSS for admin buttons .submit-row

### LINKED ISSUE #1621: Add usable save action to form designer for usability on big forms

Bartvaderkin:
Fixes #1621

### DIFF
--- a/src/openforms/scss/admin/_app_overrides.scss
+++ b/src/openforms/scss/admin/_app_overrides.scss
@@ -93,7 +93,8 @@ $djai-border-width: 8px;
   z-index: 2;
 
   &.submit-row-extended {
-    bottom: -50px;
+    position: relative;
+    z-index: 3;
     margin-top: -24px;
     border-top: 0;
     padding-top: 0;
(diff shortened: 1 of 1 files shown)


Bartvaderkin:
I made it so only the top row (with the Save buttons) stick into view, and not the secondary/extended row (with the copy/export buttons) because they'd take up too much space for non-daily actions.

sergei-maertens (approved):

PR #1644, opened 2022-05-23
[#1643] Set default value for text validate.maxLength in Form Designer

### LINKED ISSUE #1643: Textarea component doesn't save maxLength if it's the default value
### Thema / Theme

Form designer

Set a default limit of 10kb on the Formio component

Via Taiga DH 53

Note there is a linked PR that sets the default, but there is a problem where the default value isn't saved in the created configuration.

_No response_

_No response_

Bartvaderkin:
Fixes #1643

### DIFF
--- a/src/openforms/js/components/form/edit/tabs.js
+++ b/src/openforms/js/components/form/edit/tabs.js
@@ -285,6 +285,7 @@ const TEXT_VALIDATION = {
       type: 'number',
       tooltip: 'The maximum length requirement this field must meet.',
       input: true,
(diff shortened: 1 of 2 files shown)


Bartvaderkin:
Note this is currenty broken due in Form.io: after adding a textarea and hitting save the configuration JSON in the database doesn't contain this `defaultValue`, even though it was visible on the tab. 

Only when I change the value and save does it pickup up we want to use the max-length validator.

sergei-maertens (approved):

PR #1647, opened 2022-05-24
[#1325] Submission value variables backend

SilviaAmAm:
Part of #1325


src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -0,0 +1,154 @@
...
+    class Meta:
+        verbose_name = _("Submission value variable")
+        verbose_name_plural = _("Submission values variables")
+
+    def __str__(self):
+        if self.form_variable:
+            return _("Submission value variable %(name)s") % {
+                "name": self.form_variable.name
+            }
+        return _("Submission value variable %(key)s") % {"key": self.key}
```
sergei-maertens, line 119: I think we prefer the more modern `_("{foo} bar").format(foo="baz")` format

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @classmethod
+    def bulk_create_or_update(cls, submission_step, data):
```
sergei-maertens, line 122: This should follow Django principles and be a manager method, not a model class method. It's similar to `Model.objects.bulk_create` and friends.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+        )
+
+        variables_to_create = []
+        variables_to_update = []
+        for variable in submission_step_variables:
+            key = variable.key
+            if key in data:
+                if data[key] != variable.value:
+                    variable.source = SubmissionValueVariableSources.user_input
```
sergei-maertens, line 140: A comment exlaining the logic behind this would be useful. I'm not sure I understand why this is the case.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -0,0 +1,154 @@
...
+                variable.value = data[key]
+            else:
+                variable.value = ""
+
+            if not variable.pk:
+                variables_to_create.append(variable)
+            else:
+                variable.modified_at = timezone.now()
```
sergei-maertens, line 148: shouldn't this happen automatically if the model field has `auto_now=True`?

src/openforms/submissions/tests/factories.py
```diff
@@ -110,10 +117,40 @@ def from_components(
...
+            form_variables.append(
+                FormVariableFactory.create(
+                    key=component["key"],
+                    is_sensitive_data=component.get("isSensitiveData", False),
+                    form=form,
+                    form_definition=form_definition,
+                    source=FormVariableSources.component,
+                )
+            )
+
```
sergei-maertens, line 134: this pattern is repeated quite often in different places, can't we define a manager method? something like:

```py
FormVariable.objects.create_for(form)
```

That could be used in factories, tests and in actual application code if needed? Seems useful for the data-migration for existing forms, for example.

src/openforms/submissions/tests/factories.py
```diff
@@ -110,10 +117,40 @@ def from_components(
...
+        data = submitted_data or {}
         SubmissionStepFactory.create(
-            submission=submission, form_step=form_step, data=submitted_data
+            submission=submission, form_step=form_step, data=data
         )
+        for form_variable in form_variables:
+            value = form_variable.get_initial_value()
+            if form_variable.key in data:
+                value = data.get(form_variable.key)
```
sergei-maertens, line 142:
```py
value = data.get(form_variable.key, form_variable.get_initial_value())
```
to use it as a fallback value?

src/openforms/submissions/tests/factories.py
```diff
@@ -140,6 +177,33 @@ class SubmissionStepFactory(factory.django.DjangoModelFactory):
...
+                continue
+
+            SubmissionValueVariableFactory.create(
+                submission=submission_step.submission,
+                key=variable.key,
+                value=value,
+                form_variable=variable,
+            )
+
+        return submission_step
```
sergei-maertens, line 205: Shouldn't we just do always do this and thus override the super `create`? or am I missing something?

src/openforms/submissions/tests/factories.py
```diff
@@ -170,3 +234,35 @@ class SubmissionFileAttachmentFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = SubmissionFileAttachment
+
+    @classmethod
+    def create_with_variable(
```
sergei-maertens, line 239: See above remark/question

src/openforms/submissions/tests/renderer/test_formio_integration.py
```diff
@@ -59,29 +60,28 @@ class SubmissionRendererIntegrationTests(TestCase):
...
                 "input2": "second input",
                 "amount": 1234.56,
                 "input4": "fourth input",
             },
         )
 
+        form = submission.form
+        form.name = "public name"
+        form.internal_name = "internal name"
+        form.save()
```
sergei-maertens, line 76: you should be able to pass that to the `SubmissionFactory` as kwargs with double underscore, that's the main added value of factory-boy :-)

sergei-maertens (changes requested):
General direction seems to be okay, see the detailed remarks for changes I'd make.

There are a couple general things that stand out to me:

1. The ORM is nice but dangerous - it's easy to shoot yourself in the foot by doing too many database queries. A couple of tests using `self.assertNumQueries` would help prevent performance regressions and give insight in the expected queries. I'd set those up for the hot code paths, like GET/PUT on submission step, POST submission and POST logic check.
2. You probably should leverage reverse relation accessors more, especially combined with prefetch queries. They also reduce the import noise and automatically scope your relevant data to the submission/step you're working with, because that's your starting point.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -0,0 +1,192 @@
...
+    objects = SubmissionValueVariableManager()
+
+    class Meta:
+        verbose_name = _("Submission value variable")
+        verbose_name_plural = _("Submission values variables")
+        unique_together = [["submission", "key"], ["submission", "form_variable"]]
+
+    def __str__(self):
+        if self.form_variable:
+            return _("Submission value variable %(name)s").format(
```
sergei-maertens, line 189:
```suggestion
            return _("Submission value variable {name}").format(
```

sergei-maertens (changes requested):

src/openforms/formio/rendering/nodes.py
```diff
@@ -118,6 +119,16 @@ def value(self) -> Any:
...
         """
+        conf = GlobalConfiguration.get_solo()
+        if conf.enable_form_variables:
+            value_variable = SubmissionValueVariable.objects.filter(
+                submission=self.step.submission, key=self.component["key"]
+            )
+            if value_variable.exists():
+                value_variable = value_variable.get()
+                return value_variable.value
+            return None
```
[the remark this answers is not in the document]
SilviaAmAm: After the changes to how the data is handled, there is no longer a query everytime a component obtains the value. I also added a test for this

sergei-maertens (approved):

PR #1649, opened 2022-05-25
[#1325] Link form variables to form definition

SilviaAmAm:
Part of #1325

### DIFF
--- a/src/openforms/js/components/admin/form_design/data.js
+++ b/src/openforms/js/components/admin/form_design/data.js
@@ -133,6 +133,8 @@ const updateOrCreateSingleFormStep = async (
     onFormDefinitionCreate(definitionResponse.data);
     onCreateFormStep(index, stepResponse.data.url, stepResponse.data.formDefinition);
   }
+
+  return stepResponse.data;
 };
 
 /**
@@ -150,7 +152,7 @@ const updateOrCreateFormSteps = async (
 ) => {
   const stepPromises = formSteps.map(async (step, index) => {
     try {
-      await updateOrCreateSingleFormStep(
+      return await updateOrCreateSingleFormStep(
         csrftoken,
         index,
         formUrl,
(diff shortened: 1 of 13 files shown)


src/openforms/js/components/admin/form_design/data.js
```diff
@@ -196,10 +211,32 @@ const savePriceRules = async (formUrl, csrftoken, priceRules, priceRulesToDelete
...
+      new URL(variable.formDefinition);
+    } catch (e) {
+      // Retrieve the URL of the definition from the formSteps which have already been saved
+      stateformSteps.map((step, index) => {
+        if (step._generatedId === variable.formDefinition) {
+          formDefinitionUrl = updatedFormSteps[index].formDefinition;
+        }
+      });
+    }
+    return {...variable, form: formUrl, formDefinition: formDefinitionUrl};
```
sergei-maertens, line 239: Is this something that can be done in the reducer rather than in the data layer possibly? E.g. when the form definition is created, dispatch an action to update the internal react state and do this merging/joining at that time rather than deferring it? I'm not sure if that would lead to simpler code though.

sergei-maertens (approved):

src/openforms/js/components/admin/form_design/data.js
SilviaAmAm: I am not sure, because this function `createOrUpdateFormVariables` is called from the `onSubmit` function which also makes the calls to save the `formDefinition`s. When the `createOrUpdateFormVariables` is called, the react state has not been updated yet with the results of the previous calls (so `stateFormSteps` doesn't contain the new `formDefinition` URLs yet) :thinking:

sergei-maertens: yeah I think we'd have to revisit that with immer's `produce` separately maybe :thinking:

PR #1650, opened 2022-05-27
[#1616] Dynamic CSP settings

### LINKED ISSUE #1616: Allow adding items to the CSP allow list
### Thema / Theme

Other

This can be added to the analytics settings to add some CSP-related domains, or on a general level.

In either case, it needs to be added for analytics at all times because they load external scripts. The same can currently be achieved with deployment settings but this is not really admin friendly when you can enable analytics only half-baked from the admin. With SiteImprove the CSP-domains are:

  CSP_EXTRA_DEFAULT_SRC: "siteimproveanalytics.com"
  CSP_EXTRA_IMG_SRC: "*.siteimproveanalytics.io"

Better admin experience.

_No response_

Bartvaderkin:
Fixes #1616


Bartvaderkin:
It is still using the middleware as the proposed `__getattr__` on the settings doesn't work (django settings aren't used like a regular module)

sergei-maertens (approved):

src/openforms/conf/base.py
```diff
@@ -232,7 +232,7 @@
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "axes.middleware.AxesMiddleware",
     "csp.contrib.rate_limiting.RateLimitedCSPMiddleware",
-    "csp.middleware.CSPMiddleware",
+    "openforms.utils.middleware.UpdateCSPMiddleware",
```
sergei-maertens, line 235: wouldn't these have to be swapped? since the responses are applied from bottom to top, our custom updates would have to be in addition to the policy set from the settings by the `RateLimitedCSPMiddleware`?

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1662, opened 2022-06-01
Migrate to design tokens

sergei-maertens:
This centralizes our source of truth for design-tokens (and their default values) in the NPM package.

Next step is building a more user-friendly editor for the design tokens.

To check still:

- [ ] Result in e-mail previews
- [ ] Result in confirmation PDF


src/openforms/config/migrations/0029_rename_design_tokens.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # clean up and remove nonsense token specs
+        new_value = remove_empty_tokens(new_value)
+        config.design_token_values = new_value
+        config.save(update_fields=["design_token_values"])
+
+    return operation
+
+
+class Migration(migrations.Migration):
```
sergei-maertens, line 80: I need to add a test for this still

PR #1663, opened 2022-06-02
Add support for Camunda-specific plugins

sergei-maertens:


src/openforms/registrations/contrib/camunda/registry.py
```diff
@@ -0,0 +1,21 @@
...
+or poll for the availability of a certain variable.
+"""
+from openforms.plugins.registry import BaseRegistry
+
+
+class CamundaHandlerRegistry(BaseRegistry):
+    module = "registrations:camunda"
+
+
+# Sentinel to provide the default registry. You an easily instantiate another
```
SilviaAmAm, line 19:
```suggestion
# Sentinel to provide the default registry. You an easily instantiate another
```
```suggestion
# Sentinel to provide the default registry. You can easily instantiate another
```

SilviaAmAm, line 19:
```suggestion
# Sentinel to provide the default registry. You an easily instantiate another
```
```suggestion
# Sentinel to provide the default registry. You can easily instantiate another
```

SilviaAmAm (commented):
The POC looks good!

If I understand correctly, now in addition to the 'Camunda registration backend' (backend registration plugin), there will be custom 'Camunda reference plugins' which can be called from the 'Camunda registration plugin' to retrieve a reference from Camunda and perform other operations.

If the above is correct, then the only thing I would say is that maybe we need to make it clearer in the names that these are 'different types' of plugins (i.e. the 'main registration' plugin and the 'obtain reference plugins') to avoid confusion? (Like for example in the `CamundaOptionsSerializer`, the `help_text` for the plugin field says "(Custom) Camunda registration plugin")

src/openforms/registrations/contrib/camunda/plugin.py
```diff
@@ -149,14 +151,30 @@ def register_submission(
...
+            plugin = camunda_plugin_register[plugin_id]
+            client = get_client()
+            reference = plugin.obtain_reference(client, submission)
+            if reference:
+                logger.info("Plugin %r returned reference: %s", plugin, reference)
+                return reference
+            else:
+                logger.info("Plugin %r returned empty reference", plugin)
+
         raise NoSubmissionReference("Deferred to Open Forms itself")
```
sergei-maertens, line 178: I think I'll move this up a level, as this does not entirely do semantically what I'd like it to. There should be:

1. registration result and references must be saved/persisted/committed after the result was received
2. then a post-registration hook should be invoked which can block the rest of the process before user-feedback is returned

Instead of applying side-effects in the `get_reference_from_result` - this really should just be concerned with data that has been persisted in the `result` dict already.

PR #1668, opened 2022-06-08
[#1325] Form variables in form versions

SilviaAmAm:
Part of #1325


src/openforms/forms/tests/test_restore_version.py
```diff
@@ -294,3 +298,204 @@ def test_restore_form_with_reusable_form_definition(self):
...
+
+class RestoreVersionsWithVariablesTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.export_blob = {
+            "forms": json.dumps(FORM),
+            "formDefinitions": json.dumps(FORM_DEFINITION),
+            "formSteps": json.dumps(FORM_STEP),
+            "formLogic": json.dumps([]),
+        }
```
sergei-maertens, line 362: make sure to call the `super` method. It's really annoying if we have to add mixins to existing test cases and not having them working because of the missing super call.

src/openforms/forms/utils.py
```diff
@@ -63,6 +68,12 @@ def form_to_json(form_id: int) -> dict:
 
     form_logic = FormLogic.objects.filter(form=form)
 
+    # Export only user defined variables and static variables
+    # The component variables should be regenerated from the form definition configuration
```
sergei-maertens, line 72: This means that old versions without variables in them are forwards-compatible, right? Except for the static variables, so maybe we'll need to do a check during restoring if all static variables are present?

sergei-maertens (changes requested):

src/openforms/forms/utils.py
SilviaAmAm: Yes indeed. I was unsure if the static variables should be recreated for every restore, but maybe they should be like the component variables. We can have a 'list' of all static variables that forms should have and it can be used also in the data migration :thinking:

sergei-maertens: that sounds like a solid approach!

sergei-maertens (changes requested):

src/openforms/variables/views.py
```diff
@@ -0,0 +1,31 @@
...
+            200: FormVariableSerializer(many=True),
+            403: ExceptionSerializer,
+        },
+    ),
+)
+class StaticFormVariablesView(ListMixin, APIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+    serializer_class = FormVariableSerializer
+    renderer_classes = [JSONRenderer]
```
sergei-maertens, line 28: This does not output what the API schema documents - is there a reason why we can't use the regular camelcase renderer?

SilviaAmAm: No, its a mistake. I removed it now so it uses the default `CamelCaseJSONRenderer`

sergei-maertens (approved):

PR #1669, opened 2022-06-09
Check for duplicate component keys in forms

SilviaAmAm:


src/openforms/forms/management/commands/check_duplicate_component_keys.py
```diff
@@ -0,0 +1,56 @@
...
+
+from tabulate import tabulate
+
+from openforms.formio.utils import iter_components
+
+from ...models.form import Form
+
+
+class Command(BaseCommand):
+    help = "Clear the form export files to free up disk space"
```
sergei-maertens, line 13: this doesn't seem correct :)

sergei-maertens (approved):

src/openforms/forms/management/commands/check_duplicate_component_keys.py
```diff
@@ -0,0 +1,56 @@
+from typing import List
+
+from django.core.management import BaseCommand
+
+from tabulate import tabulate
```
sergei-maertens, line 5: I don't think this dependency exists in the `stable/1.0.x` branch, so we'll have to remember to install that before running the script.

src/openforms/forms/management/commands/check_duplicate_component_keys.py
SilviaAmAm: Aiai copy paste :S

Issue #1680, opened 2022-06-14
Implement DMN engine(s) integration


sergei-maertens:
Related to #1469.

There is a wish to leverage DMN (decision modelling notation) engine(s) in Open Forms - at the least it should be a way to define form logic. Initially we support the Camunda DMN engine, but we leave the implementation open to alternative engines.

This issue specifies the core DMN evaluation layer which will serve as foundation for other features.

**Tasks**

- [x] Define DMN evaluation abstraction
    - [x] Given an _id_ and input variables, evaluate a decision table/decision definition and return the output variables
    - ~~Should have: introspection API to view which input is needed for a given decision definition~~ postponed
    - [x] Pluggable abstraction so that the engine can be specified/configured & engine-specific options can be passed along
- [x] Implement Camunda DMN through django-camunda
- [x] Document public DMN API and plugin development documentation
- [x] Implement API endpoints to list available plugins/plugin decision definitions & versions
- [x] Implement management command to evaluate a decision definition

Other related tickets: #124, #1325

PR #1685, opened 2022-06-15
Implement DMN API endpoints

sergei-maertens:
Related to #1680

**Changes**

- Added API endpoint to obtain a list of enabled/available plugins/engines
- Added API endpoint to get a list of available definitions for a given engine
- Added API endpoint to get a list of available versions for a given engine/definition
- Added API endpoint to retrieve the DMN XML for a given engine/definition

### DIFF
--- a/src/openforms/dmn/api/views.py
+++ b/src/openforms/dmn/api/views.py
@@ -58,6 +58,10 @@ VERSION_PARAMETER = OpenApiParameter(
 @extend_schema_view(
     get=extend_schema(
         summary=_("List available decision plugins"),
+        responses={
+            200: DecisionPluginSerializer,
+            403: ExceptionSerializer,
+        },
     ),
 )
 class PluginListView(ListMixin, APIView):
@@ -80,6 +84,7 @@ class PluginListView(ListMixin, APIView):
         responses={
             200: DecisionDefinitionSerializer(many=True),
             400: ValidationErrorSerializer,
+            403: ExceptionSerializer,
         },
     ),
 )
@@ -109,6 +114,7 @@ class DecisionDefinitionListView(
         responses={
             200: DecisionDefinitionVersionSerializer(many=True),
             400: ValidationErrorSerializer,
+            403: ExceptionSerializer,
         },
     ),
 )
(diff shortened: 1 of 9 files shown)


src/openforms/dmn/api/views.py
```diff
@@ -0,0 +1,172 @@
...
+    enum=lazy_enum(get_available_engines),
+)
+
+DEFINITION_PARAMETER = OpenApiParameter(
+    name="definition",
+    type=str,
+    location=OpenApiParameter.QUERY,
+    required=True,
+    description=_("Identifier of the definition to retrieve available versions for."),
+)
```
joeribekker, line 48: I noticed these 2 query params are required.

For the `/xml` endpoint you include the definiton as path-variable. Why do it as query param on other endpoints? Why is the engine not a path-variable? This feels inconsistent.

* `/api/v1/dmn/decision-definitions/versions` has required `engine` and `definition` **query** params.
* `/api/v1/dmn/decision-definitions/{definition}/xml` has required `engine` **query** param and (obviously) required `definition` **path** param
* `/api/v1/dmn/decision-definitions` has required `engine` **query** params - should it be required?

src/openforms/dmn/contrib/camunda/plugin.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+            for result in results
+        ]
+
+    @staticmethod
+    def get_definition_xml(definition_id: str, version: str = "") -> str:
+        with get_client() as client:
+            camunda_id = _get_decision_definition_id(client, definition_id, version)
+            xml_response = client.get(f"decision-definition/{camunda_id}/xml")
+        return xml_response["dmn_xml"]
```
joeribekker, line 121: Should have asked in previous PR but is `check_config` implemented? I thought I saw it but would have expected it here.

sergei-maertens: ah no, I forgot about it when porting the hackathon code!

sergei-maertens: Added that in #1683

PR #1688, opened 2022-06-17
Added script to fix type=selectboxes,multiple=true components

sergei-maertens:
Standalone script so we can easily take it and run it on older versions.

Note: this has been executed on our own test environments.

### DIFF
--- a/bin/fix_selectboxes_multiple.py
+++ b/bin/fix_selectboxes_multiple.py
@@ -1,0 +1,47 @@
+# reported via Taiga Venray, issue #17
+#
+# Execute by running:
+#
+#   python bin/fix_selectboxes_multiple.py
+#
+import sys
+from pathlib import Path
+
+import django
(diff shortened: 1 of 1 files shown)


joeribekker (approved):
Ok, but I'm missing a changelog entry that this should be run and why. And why is this not a data migration? I don't see much harm if it's run twice?

sergei-maertens:
A data migration causes problems when backporting the fix as the order of migrations is different in different version lines, and eventually causes InconsistenMigrationHistory issues when upgrading instances.

This script can be run on all supported versions of Open Forms in a standalone form, meaning we can also just copy it into older versions and execute it as fire and forget.

Imo it should only be run on environments that have problems, possibly just as a diagnostic and wouldn't merge this PR.

sergei-maertens:
Closing for aforementioned reasons.

Issue #388, opened 2021-05-31
Refactor StUF clients to use common base class


sergei-maertens:
A number of constructs are common for all StUF-based services, such as the "stuurgegevens" details. We can refactor the existing clients to make use of this.

- [x] Define base class with base context, containing things like `zender_organisatie` etc.
- [x] Define common/base templates for the XML (e.g. `stuurgegevens.xml`)

sergei-maertens:
The model refactor was done a while ago, we now have `stuf.SoapService` and `stuf.StufService`.

PR #1689, opened 2022-06-17
Use a unique reference number every time for StUF-ZDS requests

### LINKED ISSUE #1681: The same referentienummer is used more than once in StUF-ZDS communication

1.1.1

In some StUF-ZDS calls, we use the `submission.uuid` as a reference number ("referentienummer" in the StUF-ZDS "stuurgegevens"). This is wrong, because it may happen that a request with this referencenumber was already performed. Specially is the `submission.uuid` is used as referencenumber for both the `creeerZaak` and the `maakZaakdocument` calls, it is simply wrong.

The StUF-ZDS standard describes that the referencenumber (in combination with the sending organisation) is unique.
See: https://www.gemmaonline.nl/images/gemmaonline/f/fa/Stuf0301.pdf (section 4.3.1)

_No response_

Use a `str(unique.uuid4)` for all references.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

sergei-maertens:
Fixes #1681

I'm not sure how to write tests for this, and I think I'd rather do #388 first, because there is now a lot of code and template duplication between stuf-bg and stuf-zds.

### DIFF
--- a/src/openforms/registrations/contrib/stuf_zds/management/commands/stuf_zds_test_stp.py
+++ b/src/openforms/registrations/contrib/stuf_zds/management/commands/stuf_zds_test_stp.py
@@ -42,7 +42,6 @@ class Command(BaseCommand):
             "zds_zaaktype_status_code": "zt-st-code",
             "zds_zaaktype_status_omschrijving": "zt-st-omschrijving",
             "zds_documenttype_omschrijving_inzending": "dt-omschrijving",
-            "referentienummer": str(uuid.uuid4()),
         }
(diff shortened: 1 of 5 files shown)


joeribekker (approved):
Thanks for fixing globally.

sergei-maertens:
Backported in cbe959b9b2615213f6270e5484dd180fee56b499 and 20fe3122c7eabc7caec5eae042bf62db8513fec2

PR #1690, opened 2022-06-20
Emit step submission validation errors

sergei-maertens:
Fixes #1687

**Changes**

* Added endpoint for explicit validation of submission step data
* Modified step serializer to run upload validation
* Refactored function to attach files to submission step to use common generator
* Fixed camelizing `data.foo_bar` in validation errors
* Fixed error messages being improperly serialized

PR #1691, opened 2022-06-21
[#1677] Scoped formio key

### LINKED ISSUE #1677: Can't use scoped formio keys with variables
I'm trying to set a `foo.bar` key for a component in Formio to test nesting for another issue, but this is blocked by the variable creation validation.

This is valid formio configuration, so `.` should be allowed in the key of a variable. (which probably complicates things).

Additionally, the backend endpoint errored on that nested/list validation error, so that'll need some attention too.

Validation error from console:

```
rest_framework.exceptions.ValidationError: [{}, {}, {}, {}, {}, {}, {}, {}, {'name': [ErrorDetail(string='Dit veld mag niet leeg zijn.', code='blank')], 'key': [ErrorDetail(string='Voer een geldige "slug" in, bestaande uit letters, cijfers, lage streepjes of streepjes.', code='invalid')]}, {'name': [ErrorDetail(string='Dit veld mag niet leeg zijn.', code='blank')], 'key': [ErrorDetail(string='Voer een geldige "slug" in, bestaande uit letters, cijfers, lage streepjes of streepjes.', code='invalid')]}]
```

SilviaAmAm:
Fixes #1677


src/openforms/api/exception_handling.py
```diff
@@ -26,53 +26,60 @@ def _translate_exceptions(exc):
...
-                        err[
-                            "name"
-                        ] = f"{underscore_to_camel(field_name)}.{i}.{err['name']}"
-                        yield err
-
-        # nested validation - recursively call the function
-        if isinstance(error_list, dict):
-            for err in get_validation_errors(error_list):
-                err["name"] = f"{underscore_to_camel(field_name)}.{err['name']}"
+def get_validation_errors(validation_errors: Union[dict, list]):
```
sergei-maertens, line 29: these changes are going to cause merge conflicts with https://github.com/open-formulieren/open-forms/pull/1690 which needs backports, so I'll only merge this after the other PR was merged.

src/openforms/forms/tests/variables/test_viewset.py
```diff
@@ -193,3 +193,94 @@ def test_list_form_variables(self):
...
+
+            self.assertEqual(error["code"], "invalid")
+            self.assertEqual(
+                error["invalidParams"][0]["name"],
+                "0.key",
+            )
+            self.assertEqual(
+                error["invalidParams"][0]["reason"],
+                "Invalid variable key. It must only contain alphanumeric characters, underscores, "
+                "dots and dashes and should not be ended by dash or dot.",
```
sergei-maertens, line 277: this should be marked translatable, tests will fail when we update the translations otherwise.

sergei-maertens (changes requested):
dots in keys should also exhibit the same behaviour as in formio, e.g. `foo.bar` will become `{"foo": {"bar": "baz"}}`

I'd like to see some tests that reflect this behaviour through `submission.data` (or `step.data`), especially a case with variables `foo.bar` and `foo.baz` - they should construct an object with key `foo` and sub-keys `bar` and `baz`.

src/openforms/forms/tests/variables/test_viewset.py
sergei-maertens: oh I missed that decorator!

src/openforms/api/exception_handling.py
sergei-maertens: you should be able to rebase on master now for this

sergei-maertens (approved):

PR #1694, opened 2022-06-21
[#1325] Modify variables with logic

SilviaAmAm:
Part of #1325 

**Changes**
- In the admin form editor, for the actions there is a new type available, which is "variable". It gives the possibility to use JsonLogic to calculate the new value of a component. I think this will replace the action "value" once variables become the default.
- During logic evaluation, the value of the variables is updated.

Thing that currently happens if one uses a variable action: 
- User changes value of a field that triggers the update of a variable
- POST to `_check_logic` updates the step data with the new variable of the user defined variable
- I think the update of the step data triggers another POST to `_check_logic`
- The last POST to `_check_logic` updates the configuration with the data of the variables calculated in the first POST to `_check_logic`.

(So 2 calls to `_check_logic` instead of 1 if a field updates a variable)

**To do**
- Decide how much validation there should be? If the result of a calculation gives a value of a different type of the variable, what should we do?

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/actions/Actions.js
+++ b/src/openforms/js/components/admin/form_design/logic/actions/Actions.js
@@ -10,6 +10,8 @@ import {ComponentsContext} from '../../../forms/Context';
 import StepSelection from '../StepSelection';
 import {Action as ActionType, ActionError} from './types';
 import DSLEditorNode from '../DSLEditorNode';
+import {FeatureFlagsContext, FormVariablesContext} from '../../Context';
+import JsonWidget from '../../../forms/JsonWidget';
 
 const ActionProperty = ({action, errors, onChange}) => {
   const modifiablePropertyChoices = Object.entries(MODIFIABLE_PROPERTIES).map(([key, info]) => [
@@ -133,6 +135,32 @@ const ActionValue = ({action, errors, onChange}) => {
   );
 };
 
+const ActionVariableValue = ({action, errors, onChange}) => {
+  const allVariables = useContext(FormVariablesContext);
+
+  const getVariableChoices = variables => {
+    return variables.map(variable => [variable.key, variable.name]);
+  };
+
+  return (
+    <>
+      <DSLEditorNode errors={errors.variable}>
+        {/*TODO: This should be a searchable select for when there are a billion variables?*/}
+        <Select
+          name="variable"
+          choices={getVariableChoices(allVariables)}
+          allowBlank
+          onChange={onChange}
+          value={action.variable}
+        />
+      </DSLEditorNode>
+      <DSLEditorNode errors={errors.action?.value}>
+        <JsonWidget name="action.value" logic={action.action.value} onChange={onChange} />
+      </DSLEditorNode>
+    </>
+  );
+};
+
 const ActionStepNotApplicable = ({action, errors, onChange}) => {
   return (
     <DSLEditorNode errors={errors.formStep}>
@@ -150,6 +178,10 @@ const ActionComponent = ({action, errors, onChange}) => {
     }
     case 'value': {
       Component = ActionValue;
+      break;
+    }
+    case 'variable': {
+      Component = ActionVariableValue;
       break;
     }
     case '':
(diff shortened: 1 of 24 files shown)


src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -133,6 +135,32 @@ const ActionValue = ({action, errors, onChange}) => {
...
+  const allVariables = useContext(FormVariablesContext);
+
+  const getVariableChoices = variables => {
+    return variables.map(variable => [variable.key, variable.name]);
+  };
+
+  return (
+    <>
+      <DSLEditorNode errors={errors.variable}>
+        {/*TODO: This should be a searchable select for when there are a billion variables?*/}
```
sergei-maertens, line 148: yep, we should be able to use `react-select` for that!

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -133,6 +135,32 @@ const ActionValue = ({action, errors, onChange}) => {
...
+          name="variable"
+          choices={getVariableChoices(allVariables)}
+          allowBlank
+          onChange={onChange}
+          value={action.variable}
+        />
+      </DSLEditorNode>
+      <DSLEditorNode errors={errors.action?.value}>
+        <JsonWidget name="action.value" logic={action.action.value} onChange={onChange} />
+      </DSLEditorNode>
```
sergei-maertens, line 159: Is the intention here to use JSONLogic to calculate/set the value?

edit: yep, confirmed by https://github.com/open-formulieren/open-forms/pull/1694/files#diff-087878c531b0db8b37141d5a552a674d16470c37801f1e4bdb443ee110e2aab4R133

src/openforms/submissions/form_logic.py
```diff
@@ -124,7 +129,35 @@ def evaluate_form_logic(
...
+                    ):
+                        # Cases where either:
+                        # 1. The variable is not related to a particular step
+                        # 2. The variable being changed is part of a different step than the step currently being edited
+                        # In these cases, the changes are persisted to the database
+                        submission_variable.value = new_value
+                        submission_variable.source = (
+                            SubmissionValueVariableSources.logic
+                        )
+                        submission_variables_to_update.append(submission_variable)
```
sergei-maertens, line 155: we should not be able to do that when doing a logic check with dirty data though, right?

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
```diff
@@ -86,22 +88,28 @@ def evaluate_form_logic(
     # on that.
     rules = getattr(submission.form, "_cached_logic_rules", None)
     if rules is None:
-        rules = FormLogic.objects.filter(form=submission.form)
+        # TODO the ordering on PK will be replaced by ordering on an order field once it's implemented
+        rules = FormLogic.objects.filter(form=submission.form).order_by("pk")
```
sergei-maertens, line 92: It's coming in #1701 :-) hopefully I can finish it this week

SilviaAmAm:
The untested code was a left over from before the PR feedback, so I removed it!

src/openforms/submissions/form_logic.py
```diff
@@ -86,22 +88,28 @@ def evaluate_form_logic(
...
                 action_details = action["action"]
+                # TODO this action will be replaced by changing the value of variables
                 if action_details["type"] == LogicActionTypes.value:
-                    new_value = jsonLogic(action_details["value"], data)
+                    new_value = jsonLogic(
+                        action_details["value"], updated_submission_data
+                    )
                     configuration = set_property_value(
                         configuration, action["component"], "value", new_value
                     )
```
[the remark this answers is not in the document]
sergei-maertens: Formio itself keeps the configuration and values separated, so `value` inside of the component configuration is ignored by formio. The only reason it would be there, is that we use it for some reason.

I was curious because I saw some code duplication between `value` and `variable` action types, but I agree that we should probably just leave it and then remove it all once variables is done.

PR #1699, opened 2022-06-22
More developer documentation

sergei-maertens:


INSTALL.rst
```diff
@@ -211,6 +211,31 @@ All settings for the project can be found in
 ``src/openforms/conf``.
 The file ``local.py`` overwrites settings from the base configuration.
 
+Celery
```
joeribekker, line 214: I would not use the tool but the purpose:

```suggestion
Running background and periodic tasks
```

INSTALL.rst
```diff
@@ -211,6 +211,31 @@ All settings for the project can be found in
...
+======
+
+You can run celery beat and worker(s) in a shell too to have the asynchronous task
```
joeribekker, line 217: And mention we use Celery for this before diving into celery commands.

INSTALL.rst
```diff
@@ -221,8 +246,139 @@ Commands can be executed using:
...
+
+List the available decision definitions for a given engine.
+
+``check_duplicate_component_keys``
+----------------------------------
+
+Check all forms and report duplicated component keys.
+
+``export``
+----------
```
joeribekker, line 278: Just curious if we can't generate this list using sphinx stuff, and move it to docs? I doubt we are going to keep this up to date...

joeribekker (changes requested):

INSTALL.rst
sergei-maertens: I think Django has a custom sphinx extension for that, so we could look into that!

PR #1718, opened 2022-06-27
Feature/1682 logs for logic evaluation

### LINKED ISSUE #1682: As admin user, I want to see the logs for logic evaluation
### Thema / Theme

Admin

Logic can be created for each form. However, the logic evaluation is performed when someone fills in a form and this evaluation is not logged. To make debugging easier, we want to log these logic evaluations.

This allows an admin user to debug issues with the form.

The ticket is about logging the evaluation of logic rules such as these:

image

**Tasks**

* [x] Add button in Django admin "Submission" page to go to the "Logboek" (logging) - See: Mock 1 (below)
* [x] Create a new page in Django admin that is shown when the button above is clicked. This page needs to have the basic Django admin layout, and needs to be protected so it can only be used by admin users.
* [x] Hook into the logic evaluation to log all rules that are evaluated. You can store the result of each evaluation-call in a timelinelogger entry. 
* [x] For now, let's only log items that trigger logic rules. The evaluation logic can be found here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/form_logic.py#L41
* [x] Move the timeline logger inline from the submission page, to the newly created page
* [x] The format of a timeline logger entry that contains logic evaluation, should be roughly as Mock 2 (see below). You can store the data as given in that function linked above and don't resolve them to pretty names.

**Mock 1***
image

**Mock 2**
image

Ewen-Lgy:
Fixes #1682

### DIFF
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -438,6 +438,29 @@ def submission_export_list(form: "Form", user: "User"):
     )
 
 
+def submission_logic_evaluated(submission: "Submission", evaluated_rules):
+
+    log_evaluated_rules = []
+
+    for evaluated_rule in evaluated_rules:
+
+        rule = evaluated_rule["rule"]
+
+        evaluated_rule_data = {
+            "source_components": rule.json_logic_trigger,
+            "targeted_components": rule.actions,
+            "trigger": evaluated_rule["trigger"],
+        }
+
+        log_evaluated_rules.append(evaluated_rule_data)
+
+    _create_log(
+        submission,
+        "submission_logic_evaluated",
+        extra_data={"log_evaluated_rules": log_evaluated_rules},
+    )
+
+
 # - - -
 
 
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -438,7 +438,9 @@ def submission_export_list(form: "Form", user: "User"):
     )
 
 
-def submission_logic_evaluated(submission: "Submission", evaluated_rules):
+def submission_logic_evaluated(
+    submission: "Submission", evaluated_rules, updated_submission_data
+):
 
     log_evaluated_rules = []
 
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -442,6 +442,8 @@ def submission_logic_evaluated(
     submission: "Submission", evaluated_rules, updated_submission_data
 ):
 
+    if not evaluated_rules:
+        return
     log_evaluated_rules = []
 
     for evaluated_rule in evaluated_rules:
--- a/src/openforms/logging/templates/logging/events/submission_logic_evaluated.txt
+++ b/src/openforms/logging/templates/logging/events/submission_logic_evaluated.txt
@@ -1,0 +1,4 @@
+{% load i18n %}
+{% blocktrans trimmed with lead=log.fmt_lead submission=log.fmt_sub form=log.fmt_form %}
+   {{ lead }}: logged rule of {{ submission }} for form {{ form }}
+{% endblocktrans %}
--- a/src/openforms/logging/tests/test_events.py
+++ b/src/openforms/logging/tests/test_events.py
@@ -66,16 +66,14 @@ class EventTests(VariablesTestMixin, TestCase):
         logevent.submission_logic_evaluated(
             submission,
             [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
-            submission.get_merged_data(),
+            submission.data,
         )
 
-        logs = submission.logs.all()
         log = submission.logs.get()
         logged_rules = log.extra_data["log_evaluated_rules"]
 
-        self.assertEquals(1, logs.count())
-        self.assertEquals(2, len(logged_rules))
-        self.assertEquals(
+        self.assertEqual(2, len(logged_rules))
+        self.assertEqual(
             {
                 "trigger": True,
                 "source_components": json_logic_trigger,
@@ -83,7 +81,7 @@ class EventTests(VariablesTestMixin, TestCase):
             },
             logged_rules[0],
         )
-        self.assertEquals(
+        self.assertEqual(
             {
                 "trigger": False,
                 "source_components": json_logic_trigger_2,
--- a/src/openforms/logging/tests/test_events.py
+++ b/src/openforms/logging/tests/test_events.py
@@ -1,7 +1,7 @@
 from django.test import TestCase
 from django.utils.translation import gettext as _
 
-from openforms.forms.models.form import FormLogic
+from openforms.forms.models import FormLogic
 from openforms.logging import logevent
 from openforms.logging.models import TimelineLogProxy
 from openforms.submissions.tests.factories import SubmissionFactory
--- a/src/openforms/logging/tests/test_events.py
+++ b/src/openforms/logging/tests/test_events.py
@@ -1,12 +1,23 @@
 from django.test import TestCase
 from django.utils.translation import gettext as _
 
+from openforms.forms.models.form import FormLogic
 from openforms.logging import logevent
 from openforms.logging.models import TimelineLogProxy
 from openforms.submissions.tests.factories import SubmissionFactory
+from openforms.submissions.tests.mixins import VariablesTestMixin
 
 
-class EventTests(TestCase):
+class EventTests(VariablesTestMixin, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.submission = SubmissionFactory.from_data(
+            {
+                "firstname": "foo",
+                "birthdate": "2022-06-20",
+            },
+        )
+
     # test for specific events
     def test_submission_start(self):
         with self.subTest("anon"):
@@ -33,3 +44,61 @@ class EventTests(TestCase):
 
             log = TimelineLogProxy.objects.last()
             self.assertEqual(_("Anonymous user"), log.fmt_user)
+
+    def test_submission_logic_evaluated(self):
+
+        json_logic_trigger = {
+            ">": [{"date": {"var": "birthdate"}}, {"date": "2022-06-20"}]
+        }
+
+        json_logic_trigger_2 = {"==": [{"var": "firstname"}, "bar"]}
+
+        actions = [
+            {
+                "component": "firstname",
+                "formStep": "",
+                "action": {
+                    "type": "property",
+                    "property": {"value": "disabled", "type": "bool"},
+                    "state": True,
+                },
+            }
+        ]
+
+        actions_2 = [
+            {
+                "component": "",
+                "formStep": "",
+                "action": {
+                    "type": "disable-next",
+                    "property": {"type": "", "value": ""},
+                    "value": "",
+                    "state": "",
+                },
+            }
+        ]
+
+        actual = self.submission.get_merged_data()
+        expected = {"firstname": "foo", "birthdate": "2022-06-20"}
+        rule = FormLogic(
+            form=self.submission.form,
+            json_logic_trigger=json_logic_trigger,
+            actions=actions,
+        )
+        rule_2 = FormLogic(
+            form=self.submission.form,
+            json_logic_trigger=json_logic_trigger_2,
+            actions=actions_2,
+        )
+
+        logevent.submission_logic_evaluated(
+            self.submission,
+            [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
+        )
+        logs = self.submission.logs.all()
+        log = logs.last()
+        logged_rules = log.extra_data["log_evaluated_rules"]
+
+        self.assertEquals(1, logs.count())
+        self.assertEquals(2, len(logged_rules))
+        self.assertEquals(actual, expected)
--- a/src/openforms/logging/tests/test_events.py
+++ b/src/openforms/logging/tests/test_events.py
@@ -1,7 +1,7 @@
 from django.test import TestCase
 from django.utils.translation import gettext as _
 
-from openforms.forms.models import FormLogic
+from openforms.forms.tests.factories import FormLogicFactory
 from openforms.logging import logevent
 from openforms.logging.models import TimelineLogProxy
 from openforms.submissions.tests.factories import SubmissionFactory
@@ -9,14 +9,6 @@ from openforms.submissions.tests.mixins import VariablesTestMixin
 
 
 class EventTests(VariablesTestMixin, TestCase):
-    @classmethod
-    def setUpTestData(cls):
-        cls.submission = SubmissionFactory.from_data(
-            {
-                "firstname": "foo",
-                "birthdate": "2022-06-20",
-            },
-        )
 
     # test for specific events
     def test_submission_start(self):
@@ -47,58 +39,55 @@ class EventTests(VariablesTestMixin, TestCase):
 
     def test_submission_logic_evaluated(self):
 
+        submission = SubmissionFactory.from_data(
+            {
+                "firstname": "foo",
+                "birthdate": "2022-06-20",
+            },
+        )
+
         json_logic_trigger = {
             ">": [{"date": {"var": "birthdate"}}, {"date": "2022-06-20"}]
         }
 
         json_logic_trigger_2 = {"==": [{"var": "firstname"}, "bar"]}
 
-        actions = [
-            {
-                "component": "firstname",
-                "formStep": "",
-                "action": {
-                    "type": "property",
-                    "property": {"value": "disabled", "type": "bool"},
-                    "state": True,
-                },
-            }
-        ]
-
-        actions_2 = [
-            {
-                "component": "",
-                "formStep": "",
-                "action": {
-                    "type": "disable-next",
-                    "property": {"type": "", "value": ""},
-                    "value": "",
-                    "state": "",
-                },
-            }
-        ]
-
-        actual = self.submission.get_merged_data()
-        expected = {"firstname": "foo", "birthdate": "2022-06-20"}
-        rule = FormLogic(
-            form=self.submission.form,
+        rule = FormLogicFactory(
+            form=submission.form,
             json_logic_trigger=json_logic_trigger,
-            actions=actions,
+            actions=[],
         )
-        rule_2 = FormLogic(
-            form=self.submission.form,
+        rule_2 = FormLogicFactory(
+            form=submission.form,
             json_logic_trigger=json_logic_trigger_2,
-            actions=actions_2,
+            actions=[],
         )
 
         logevent.submission_logic_evaluated(
-            self.submission,
+            submission,
             [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
+            submission.get_merged_data(),
         )
-        logs = self.submission.logs.all()
-        log = logs.last()
+
+        logs = submission.logs.all()
+        log = submission.logs.get()
         logged_rules = log.extra_data["log_evaluated_rules"]
 
         self.assertEquals(1, logs.count())
         self.assertEquals(2, len(logged_rules))
-        self.assertEquals(actual, expected)
+        self.assertEquals(
+            {
+                "trigger": True,
+                "source_components": json_logic_trigger,
+                "targeted_components": rule.actions,
+            },
+            logged_rules[0],
+        )
+        self.assertEquals(
+            {
+                "trigger": False,
+                "source_components": json_logic_trigger_2,
+                "targeted_components": rule_2.actions,
+            },
+            logged_rules[1],
+        )
--- a/src/openforms/scss/components/admin/_logic-evaluated-table.scss
+++ b/src/openforms/scss/components/admin/_logic-evaluated-table.scss
@@ -1,0 +1,19 @@
+.logic-rules-tables {
+    width: 100%;
+}
+
+.property-name-width {
+    width: 70%;
+}
+
+.rule-width {
+    width: 60%;
+}
+
+.trigger-width {
+    width: 10%;
+}
+
+.rule-list {
+    list-style: none;
+}
--- a/src/openforms/scss/components/admin/_logic-evaluated-table.scss
+++ b/src/openforms/scss/components/admin/_logic-evaluated-table.scss
@@ -1,19 +1,19 @@
 .logic-rules-tables {
-    width: 100%;
+  width: 100%;
 }
 
 .property-name-width {
-    width: 70%;
+  width: 70%;
 }
 
 .rule-width {
-    width: 60%;
+  width: 60%;
 }
 
 .trigger-width {
-    width: 10%;
+  width: 10%;
 }
 
 .rule-list {
-    list-style: none;
+  list-style: none;
 }
--- a/src/openforms/submissions/admin.py
+++ b/src/openforms/submissions/admin.py
@@ -18,7 +18,7 @@ from openforms.logging.logevent import (
 )
 from openforms.logging.models import TimelineLogProxy
 from openforms.payments.models import SubmissionPayment
-from openforms.submissions.views import LogsEvaluatedLogicView
+from openforms.submissions.admin_views import LogsEvaluatedLogicView
 
 from ..utils.admin import ReadOnlyAdminMixin
 from .constants import IMAGE_COMPONENTS, RegistrationStatuses
@@ -326,7 +326,7 @@ class SubmissionAdmin(admin.ModelAdmin):
         urls = super().get_urls()
         my_urls = [
             path(
-                "<int:submission_id>/evaluated-logic-logs",
+                "<int:submission_id>/logs/logic/",
                 LogsEvaluatedLogicView.as_view(),
                 name="logs-evaluated-logic",
             )
--- a/src/openforms/submissions/admin.py
+++ b/src/openforms/submissions/admin.py
@@ -4,6 +4,7 @@ from django import forms
 from django.contrib import admin, messages
 from django.contrib.contenttypes.admin import GenericTabularInline
 from django.template.defaultfilters import filesizeformat
+from django.urls import path
 from django.utils.translation import gettext_lazy as _, ngettext
 
 from privates.admin import PrivateMediaMixin
@@ -17,6 +18,7 @@ from openforms.logging.logevent import (
 )
 from openforms.logging.models import TimelineLogProxy
 from openforms.payments.models import SubmissionPayment
+from openforms.submissions.views import LogsEvaluatedLogicView
 
 from ..utils.admin import ReadOnlyAdminMixin
 from .constants import IMAGE_COMPONENTS, RegistrationStatuses
@@ -319,6 +321,17 @@ class SubmissionAdmin(admin.ModelAdmin):
     retry_processing_submissions.short_description = _(
         "Retry processing %(verbose_name_plural)s."
     )
+
+    def get_urls(self):
+        urls = super().get_urls()
+        my_urls = [
+            path(
+                "<int:submission_id>/evaluated-logic-logs",
+                LogsEvaluatedLogicView.as_view(),
+                name="logs-evaluated-logic",
+            )
+        ]
+        return my_urls + urls
 
 
 @admin.register(SubmissionReport)
--- a/src/openforms/submissions/admin_views.py
+++ b/src/openforms/submissions/admin_views.py
@@ -1,0 +1,25 @@
+from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
+from django.views.generic import ListView
+
+from .models import Submission
+
+
+class LogsEvaluatedLogicView(PermissionRequiredMixin, LoginRequiredMixin, ListView):
+
+    template_name = "submission_logs/submission_logs.html"
+    context_object_name = "logs_activity"
+    permission_required = "submissions.view_submission"
+
+    def get_queryset(self):
+        self.id = self.kwargs["submission_id"]
+        self.submission = Submission.objects.get(id=self.id)
+        self.queryset = self.submission.logs.filter(
+            template="logging/events/submission_logic_evaluated.txt"
+        ).order_by("timestamp")
+        return super().get_queryset()
+
+    def get_context_data(self):
+        context = super().get_context_data()
+        context["opts"] = Submission._meta
+        context["original"] = self.submission
+        return context
--- a/src/openforms/submissions/admin_views.py
+++ b/src/openforms/submissions/admin_views.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.shortcuts import get_object_or_404
+from django.utils.translation import gettext as _
 from django.views.generic import ListView
 
 from openforms.utils.mixins import UserIsStaffMixin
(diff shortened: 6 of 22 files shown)


src/openforms/submissions/api/viewsets.py
```diff
@@ -198,6 +200,15 @@ def _complete(self, request, *args, **kwargs):
 
         logevent.form_submit_success(submission)
 
+        rules = FormLogic.objects.filter(form=submission.form)
```
Ewen-Lgy, line 203: I decided to log the not triggered rule just before to submit the submission. If I did in the evaluation logic, I would have had potentially  the same logic rule triggered and not triggered in different logs

src/openforms/logging/logevent.py
```diff
@@ -437,6 +440,58 @@ def submission_export_list(form: "Form", user: "User"):
...
+
+    # Assuming key is unique in all form steps 
+    for targeted_component in targeted_components :
+        # Adding the label of the key and step name associated if key exists 
+        if 'key' in targeted_component.keys() :
+            for formstep in formsteps :
+                if targeted_component['key'] in formstep.form_definition.get_all_keys() :
+                    targeted_component['label'] = formstep.form_definition.get_key_label(targeted_component['key'])
+                    targeted_component['step_name'] = formstep.form_definition.name
+                    break
```
sergei-maertens, line 484: This can benefit from some guard clauses to remove indentation from your code:

```suggestion
    for targeted_component in targeted_components :
        # Adding the label of the key and step name associated if key exists 
        if not (key := target_component.get("key")):
            continue
        for formstep in formsteps:
            if key not in formstep.form_definition.get_all_keys():
                continue
            targeted_component['label'] = formstep.form_definition.get_key_label(key)
            targeted_component['step_name'] = formstep.form_definition.name
            break
```

This is an example of flattening the loops and using guard clauses - additionally I think it's better to build a mapping _before_ entering the actions loop where you map formio component key to the form step it belongs, because this information is a loop invariant and doesn't need to be re-calculated every time you introspect/enhance an action.

src/openforms/logging/templates/logging/events/submission_logic_evaluated.txt
```diff
@@ -0,0 +1,4 @@
+{% load i18n %}
+{% blocktrans trimmed with lead=log.fmt_lead user=log.fmt_user submission=log.fmt_sub form=log.fmt_form %}
```
sergei-maertens, line 2: `user` is unused

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +38,42 @@ def test_submission_start(self):
 
             log = TimelineLogProxy.objects.last()
             self.assertEqual(_("Anonymous user"), log.fmt_user)
+
+    def test_submission_logic_evaluated(self):
+        submission = SubmissionFactory.create()
+        form = FormFactory.create()
+        form_step = FormStepFactory.create(form = form)
+        submission_step = SubmissionStepFactory.create(submission = submission, form_step = form_step)
```
sergei-maertens, line 46: You can cut down a lot of boilerplate with `SubmissionFactory.from_data` or `SubmissionFactory.from_components`

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +38,42 @@ def test_submission_start(self):
...
+            }
+        }]
+        rule = FormLogic(form=form, json_logic_trigger=json_logic_trigger, actions=actions)
+        logevent.submission_logic_evaluated(submission, rule)
+        log = TimelineLogProxy.objects.filter(
+            template="logging/events/submission_logic_evaluated.txt", 
+            object_id = submission.id)[0]
+        
+        self.assertIsNotNone(log)
+
```
sergei-maertens, line 79: Please structure your tests nicely with whitespace:

```
def test_something(self):
    ...  # test data setup

    result = call_thing_your_are_testing(...)

    ... # make assertions
```

having a blank line before and after the call of the thing you're testing makes the structure of `setup / call / assert` more obvious and easier to understand/debug

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +38,42 @@ def test_submission_start(self):
...
+                "state": True
+            }
+        }]
+        rule = FormLogic(form=form, json_logic_trigger=json_logic_trigger, actions=actions)
+        logevent.submission_logic_evaluated(submission, rule)
+        log = TimelineLogProxy.objects.filter(
+            template="logging/events/submission_logic_evaluated.txt", 
+            object_id = submission.id)[0]
+        
+        self.assertIsNotNone(log)
```
sergei-maertens, line 78: 1. Accessing by index `[0]` does not guarantee there haven't been multiple records created
2. `queryset[0]` will never be None, as that would raise an `IndexError` instead.

-> assert the expected number of objects with `qs.count()` and/or use `queryset.get()` to guarantee only one record exists.

Then, make assertions on the content of the object, here you'd want to validate `expected_data` and validate that the introspection from formio was done correctly (i.e. correct step name, correct component label... are recorded).

src/openforms/scss/components/admin/_logic-evaluated-table.scss
```diff
@@ -0,0 +1,19 @@
...
+    width: 60%;
+}
+
+.trigger-width {
+    width: 10%;
+}
+
+.rule-list {
+    list-style: none;
+}
```
sergei-maertens, line 19: We tend to stick with http://getbem.com/ for component structure, so after the backend feedback has processed, we should walk through that in more detail :-)

src/openforms/submissions/api/viewsets.py
```diff
@@ -198,6 +200,15 @@ def _complete(self, request, *args, **kwargs):
...
+        logs = TimelineLogProxy.objects.filter(
+            template="logging/events/submission_logic_evaluated.txt", 
+            object_id = submission.id)
+        for rule in rules :
+            # rule has never been triggered
+            if rule.id not in [log.extra_data['rule_id'] for log in logs]:
+                logevent.submission_logic_evaluated(submission, rule)
+
```
sergei-maertens, line 211: @joeribekker submission completion does not trigger another logic rule evaluation (and I don't see why it should), did you request for these rules to be logged (again)? I'm not convinced this should be here.

src/openforms/submissions/form_logic.py
```diff
@@ -90,10 +95,11 @@ def evaluate_form_logic(
         submission.form._cached_logic_rules = rules
 
     submission_state = submission.load_execution_state()
-
+    
     updated_data = deepcopy(step.data)
     for rule in rules:
         if jsonLogic(rule.json_logic_trigger, data):
+            logevent.submission_logic_evaluated(submission, rule, trigger=True)
```
sergei-maertens, line 102: hrm, we only log the rule if it evaluates to true, but not if it's not triggered. That doesn't seem right to me - I'd expect all the evaluated rules to be logged.

src/openforms/submissions/templates/admin/submissions/submission/change_form.html
```diff
@@ -1,6 +1,12 @@
 {% extends 'admin/change_form.html' %}
 {% load i18n solo_tags %}
 
+{% block object-tools-items %}{{ block.super }}
```
sergei-maertens, line 4:
```suggestion
{% block object-tools-items %}
{{ block.super }}
```

src/openforms/submissions/templates/submission_logs/rules.html
```diff
@@ -0,0 +1,18 @@
+{% if node.nodeType == 'operation' %}
+    <li class="rule-list"><b> {{node.operator}}: </b> </li>
```
sergei-maertens, line 2: `<b>` is not recommended anymore these days. Either use the more semantic `<strong>` or apply the styling with CSS.

src/openforms/submissions/templates/submission_logs/rules.html
```diff
@@ -0,0 +1,18 @@
...
+    <ul>
+{% else %}
+
+    <li class="rule-list">
+        <b>{{node.stepName}}: {{node.key}} &#40;{{node.label}}&#41;</b> <i>{{node.operator}}</i> <b> {{node.value}}</b>
```
sergei-maertens, line 7: Please put spaces in the curly braces

```suggestion
        <b>{{ node.stepName }}: {{ node.key }} &#40;{{ node.label }}&#41;</b> <i>{{ node.operator }}</i> <b> {{ node.value }}</b> 
```

src/openforms/submissions/templates/submission_logs/submission_logs.html
```diff
@@ -0,0 +1,69 @@
...
+&rsaquo;{% trans 'Evaluated Logic Logs'|capfirst%}
+</div>
+{% endblock %}
+
+{% block content %}
+    <h1>{% trans  "Logging" %}</h1>
+    {% block field_sets %}
+        {% for log in logs_activity  %}
+            <fieldset class="module aligned collapse collapsed">
+                <h2>{{ log.name|date:'Y-m-d H:i:s' }} - {% trans 'Logic' %} - </h2>
```
sergei-maertens, line 24: is `.name` a date/datetime?

src/openforms/submissions/templates/submission_logs/submission_logs.html
```diff
@@ -0,0 +1,69 @@
...
+                    </thead>
+                    <tbody>
+                        {% for action in log.targeted_components %}
+                            <tr>
+                                <td>
+                                    {% for node in log.source_components  %}
+                                    {% include "submission_logs/rules.html" %}
+                                    {% endfor %}
+                                </td>
+                                <td>{% if log.trigger %}{% trans "Yes" %} {% else %} {% trans "No" %} {% endif %} </td>
```
sergei-maertens, line 54:
```suggestion
                                <td>{% if log.trigger %}{% trans "Yes" %} {% else %} {% trans "No" %} {% endif %} </td>
```
Look up the `yesno` template filter :smile:

src/openforms/submissions/urls.py
```diff
@@ -15,4 +15,9 @@
         SubmissionAttachmentDownloadView.as_view(),
         name="attachment-download",
     ),
+    path(
+        "<int:submission_id>/evaluated-logic-logs",
+        LogsEvaluatedLogic.as_view(),
+        name="logs-evaluated-logic",
+    )
```
sergei-maertens, line 22: This url belongs in the `SubmissionAdmin.get_urls` instead of in the "public" url patterns.

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,57 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+    return ""
+
+
+def find_key_value(value: Union[JsonLogicTest, str]) -> str :
+    """Find the value associated to the key in the json_trigger_rule"""
+    if isinstance(value, JsonLogicTest) :
+        return find_key_value(value.values[0])
+
+    return value
```
sergei-maertens, line 71: I'm not sure what this does? What happens if the order of `values` is swapped?

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,57 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+def formating_json_logic(logics :Union[dict, list, JsonLogicTest], candidates, submission):
```
sergei-maertens, line 73: this function name should be something more accurate like `convert_json_logic_to_logging_tree` or something like that.

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,57 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+                'operator' :logics.operator,
+                'children' : formating_json_logic(logics.values, candidates, submission)
+                }
+        return ""
+    
+    else :
+        logic_list = []
+        for logic in logics :
+            logic_list.append(formating_json_logic(logic, candidates, submission))
+        return logic_list
```
sergei-maertens, line 105: that doesn't seem right if `logics` is a dict

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,57 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+        for value in logics.values:
+            if value.operator in ['var', 'date'] :
+                key = find_component_key(logics, candidates)
+                form_step_name, label = '', ''
+                formsteps = submission.form.formstep_set.all()
+                for formstep in formsteps:
+                    if key in formstep.form_definition.get_all_keys() :
+                        label = formstep.form_definition.get_key_label(key)
+                        form_step_name = formstep.form_definition.name
+                        break
```
sergei-maertens, line 84: This seems like a blob of code that's repeated in a couple of places

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,57 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+                return {
+                    'nodeType' : 'leaf',
+                    'operator': logics.operator,
+                    'key': key,
+                    'value': find_key_value(logics.values[1]),
```
sergei-maertens, line 90: this doesn't account for unary operators I think, for example: `{"!": [True]}`

sergei-maertens (changes requested):
The tests are a bit lacking in assertions made and possible complex situations that can arise, please expand on those.

Then, when CI will run you'll also see that the `black` and `isort` code formatting fails, so please run those and/or set up a pre-commit hook. You can take a look at https://github.com/maykinmedia/captain-hook for this

src/openforms/submissions/templates/submission_logs/submission_logs.html
Ewen-Lgy: It is a date, small confusion because it is used as the title of the log entry, I change that name

src/openforms/utils/json_logic.py
Ewen-Lgy: @sergei-maertens It retrieves the value compared in the condition no matter we compare either the field to a value, or another field/date :
Capture d’écran du 2022-06-30 15-10-52

src/openforms/utils/json_logic.py
Ewen-Lgy: I do think it is impossible that logics is a dict. But maybe you said so because in the signature I was alowing dict (and that was a mistake)

src/openforms/logging/tests/test_events.py
```diff
@@ -1,12 +1,23 @@
...
-class EventTests(TestCase):
+class EventTests(VariablesTestMixin, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.submission = SubmissionFactory.from_data(
+            {
+                "firstname": "foo",
+                "birthdate": "2022-06-20",
+            },
+        )
```
sergei-maertens, line 19: Try to keep your test setup/data _in_ the test itself rather than splitting it over multiple test case methods. I'd really just put this inside of `def test_submission_logic_evaluated`

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +44,61 @@ def test_submission_start(self):
...
+                "component": "",
+                "formStep": "",
+                "action": {
+                    "type": "disable-next",
+                    "property": {"type": "", "value": ""},
+                    "value": "",
+                    "state": "",
+                },
+            }
+        ]
```
sergei-maertens, line 79: I don't think we really do anything with the actions here in the test, so maybe don't specify them at all? When creating the rule, you can use `actions=[]` and it'll be allowed by the database :)

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +44,61 @@ def test_submission_start(self):
...
+
+        actual = self.submission.get_merged_data()
+        expected = {"firstname": "foo", "birthdate": "2022-06-20"}
```
sergei-maertens, line 82: This + the assertion at the end doesn't really fit the scope of this test - you already set up the data in the submission factory so this is pretty much a given. I'd just remove this completely.

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +44,61 @@ def test_submission_start(self):
...
+        rule = FormLogic(
+            form=self.submission.form,
+            json_logic_trigger=json_logic_trigger,
+            actions=actions,
+        )
+        rule_2 = FormLogic(
+            form=self.submission.form,
+            json_logic_trigger=json_logic_trigger_2,
+            actions=actions_2,
+        )
```
sergei-maertens, line 92: I'd rather use the `FormLogicFactory` here instead of the model itself, for consistency reasons.

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +44,61 @@ def test_submission_start(self):
...
+
+        logevent.submission_logic_evaluated(
+            self.submission,
+            [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
+        )
+        logs = self.submission.logs.all()
```
sergei-maertens, line 98:
```suggestion
        logevent.submission_logic_evaluated(
            self.submission,
            [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
        )
        
        logs = self.submission.logs.all()
```
a visual separation between "call" and "assertions" makes it easier to understand what is being tested

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +44,61 @@ def test_submission_start(self):
...
+        log = logs.last()
+        logged_rules = log.extra_data["log_evaluated_rules"]
+
+        self.assertEquals(1, logs.count())
+        self.assertEquals(2, len(logged_rules))
```
sergei-maertens, line 103: maybe `log = submission.logs.get()` - this will crash if there's more or less than 1 record (and thus fail the test).

I'd also like to see some assertions about the exact structure of `logged_rules` rather than just testing how many objects are there

src/openforms/submissions/admin.py
```diff
@@ -17,6 +18,7 @@
 )
 from openforms.logging.models import TimelineLogProxy
 from openforms.payments.models import SubmissionPayment
+from openforms.submissions.views import LogsEvaluatedLogicView
```
sergei-maertens, line 21: I'd move this view implementation to `openforms.submissions.admin_views`. Later, we should restructure the admin code so that it's a folder `admin/` with python modules inside, similar to `openforms.forms.admin`.

Splitting between `views.py` and `admin_views.py` makes it more obvious for developers in which context these views are used (actual end-users vs. staff users)

src/openforms/submissions/admin.py
```diff
@@ -320,6 +322,17 @@ def retry_processing_submissions(self, request, queryset):
         "Retry processing %(verbose_name_plural)s."
     )
 
+    def get_urls(self):
+        urls = super().get_urls()
+        my_urls = [
+            path(
+                "<int:submission_id>/evaluated-logic-logs",
```
sergei-maertens, line 329:
```suggestion
                "<int:submission_id>/logs/logic/",
```
bit cleaner URL

src/openforms/submissions/form_logic.py
```diff
@@ -93,8 +97,10 @@ def evaluate_form_logic(
 
     updated_step_data = deepcopy(step.data)
     updated_submission_data = deepcopy(data)
+    evaluated_rules = list()
```
sergei-maertens, line 100:
```suggestion
    evaluated_rules = []
```
for consistency with the rest of the code

src/openforms/submissions/form_logic.py
```diff
@@ -93,8 +97,10 @@ def evaluate_form_logic(
...
     for rule in rules:
         if jsonLogic(rule.json_logic_trigger, updated_submission_data):
+            evaluated_rules.append({"rule": rule, "trigger": True})
```
sergei-maertens, line 103: I'd restructure this quite a bit - now the `if: else` are pretty far apart and ideally we keep the logging code in just a single place:

```suggestion
        trigger_rule = jsonLogic(rule.json_logic_trigger, updated_submission_data)
        evaluated_rules.append({"rule": rule, "trigger": trigger_rule})
        if not trigger_rule:
            continue
        ...  # for action in rules.actions...
```
& this way you can remove the `else` completely a couple lines down.

The `if not trigger_rule: continue` construct also allows us to simplify the code by removing a nesting level

src/openforms/submissions/form_logic.py
```diff
@@ -136,8 +142,13 @@ def evaluate_form_logic(
...
                     updated_step_data[variable_key] = new_value
                     updated_submission_data = {**data, **updated_step_data}
-
+        else:
+            evaluated_rules.append({"rule": rule, "trigger": False})
     step.data = DirtyData(updated_step_data)
+    if evaluated_rules:
+        logevent.submission_logic_evaluated(
+            submission, evaluated_rules, updated_submission_data
+        )
```
sergei-maertens, line 151: I'd consider always calling this and doing the check `if evaluated_rules` inside of `logevent.submission_logic_evaluated`

src/openforms/submissions/form_logic.py
```diff
@@ -8,10 +9,13 @@
 from openforms.formio.utils import get_default_values, iter_components
 from openforms.forms.constants import LogicActionTypes
 from openforms.forms.models import FormLogic
+from openforms.logging import logevent
 from openforms.prefill import JSONObject
 
 from .models.submission_step import DirtyData
 
+logger = logging.getLogger(__name__)
```
sergei-maertens, line 17: this does not seem used, does it?

src/openforms/submissions/templates/admin/submissions/submission/change_form.html
```diff
@@ -1,6 +1,13 @@
 {% extends 'admin/change_form.html' %}
 {% load i18n solo_tags %}
 
+{% block object-tools-items %}
+{{ block.super }}
+  <li>
+    <a href="{% url 'admin:logs-evaluated-logic' submission_id=original.pk %}">{% trans "Logging" %}</a>
```
sergei-maertens, line 7:
```suggestion
    <a href="{% url 'admin:logs-evaluated-logic' submission_id=original.pk %}">{% trans "Logic logs" %}</a>
```
the rest of the logs are still on this same page

src/openforms/submissions/tests/test_admin.py
```diff
@@ -143,3 +146,72 @@ def test_change_view_displays_logs_if_not_avg(self):
...
+        )
+
+    def setUp(self):
+        super().setUp()
+        self.admin_user = UserFactory.create(
+            is_superuser=True, is_staff=True, app=self.app
+        )
+        self.anonymous_user = UserFactory.create(
+            is_superuser=False, is_staff=False, app=self.app
+        )
```
sergei-maertens, line 168: we have a decorator to disable 2fa in `openforms.tests.utils` - better to just decorate the entire test case and then you can just set up your users in `setUpTestData` instead of here.

src/openforms/submissions/tests/test_admin.py
```diff
@@ -143,3 +146,72 @@ def test_change_view_displays_logs_if_not_avg(self):
...
+                "firstname": "foo",
+                "birthdate": "2022-06-20",
+            },
+        )
+
+    def setUp(self):
+        super().setUp()
+        self.admin_user = UserFactory.create(
+            is_superuser=True, is_staff=True, app=self.app
+        )
```
sergei-maertens, line 165: we have `StaffUserFactory` and `SuperUserFactory` too :-)

sergei-maertens (changes requested):

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +36,58 @@ def test_submission_start(self):
...
+            submission,
+            [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
+            submission.get_merged_data(),
+        )
+
+        logs = submission.logs.all()
+        log = submission.logs.get()
+        logged_rules = log.extra_data["log_evaluated_rules"]
+
+        self.assertEquals(1, logs.count())
```
sergei-maertens, line 76:
```suggestion
        self.assertEqual(1, logs.count())
```
-> `assertEquals` is deprecated in favor of `assertEqual`

src/openforms/logging/tests/test_events.py
```diff
@@ -33,3 +36,58 @@ def test_submission_start(self):
...
+        rule_2 = FormLogicFactory(
+            form=submission.form,
+            json_logic_trigger=json_logic_trigger_2,
+            actions=[],
+        )
+
+        logevent.submission_logic_evaluated(
+            submission,
+            [{"rule": rule, "trigger": True}, {"rule": rule_2, "trigger": False}],
+            submission.get_merged_data(),
```
sergei-maertens, line 69:
```suggestion
            submission.data,
```
we prefer the property accessor instead of calling the method :)

src/openforms/logging/tests/test_events.py
sergei-maertens, line 76:
```suggestion
        log = submission.logs.get()  # implicitly fails the test if there are not exactly one log record
        logged_rules = log.extra_data["log_evaluated_rules"]
```

src/openforms/submissions/admin_views.py
```diff
@@ -0,0 +1,25 @@
...
+        self.queryset = self.submission.logs.filter(
+            template="logging/events/submission_logic_evaluated.txt"
+        ).order_by("timestamp")
+        return super().get_queryset()
+
+    def get_context_data(self):
+        context = super().get_context_data()
+        context["opts"] = Submission._meta
+        context["original"] = self.submission
+        return context
```
sergei-maertens, line 25: Couple improvements to be made here, I'll start with the anti-patterns first:

1. you typically shouldn't assign anything to `self` outside of the `__init__` method. (the `self.id = ...` doesn't really make much sense, especially because the data is already available as `self.kwargs["submission_id"]`. In its current form, you're sort of "polluting the namespace". A local variable would've been sufficient
2. assigning `self.queryset = ...` suffers from the problem above, but it's also not really how you're supposed to implement this. Django has a default implementation for `get_queryset` that:
    - either checks if `queryset` is set, if it is -> `return self.queryset.all()`
    - if no queryset is set, check if `model` is set and get the queryset from that: `return self.model.objects.all()`

    You can override this by just returning a queryset from the `get_queryset` method and django will be fine with that - you definitely don't need to overwrite the instance attribute.

Then, some further improvements:

1. You can make the `self.submission` thing a bit more self-contained by implementing it as a property with an internal cache. That makes it so you do the database lookup only once, without doing too much `self.foo = bar` stuff - it's a bit of a trade-off. In my version, it is at least made "obvious" that this is some internal caching API by using the `_` prefix.
2. Since this view is used in the admin, you typically want to enforce that it's an admin user viewing it, on top of having the required permission. We have an existing mixin for that in `openforms.utils.mixins.UserIsStaffMixin`. Your tests should also reflect this in the form of setting up a non-staff user with the permission - they should not have access to this view!

All in all, I'd rewrite this as:

```py
from openforms.utils.mixins import UserIsStaffMixin
from django.shortcuts import get_object_or_404


class LogsEvaluatedLogicView(UserIsStaffMixin, PermissionRequiredMixin, ListView):

    template_name = "submission_logs/submission_logs.html"
    context_object_name = "logs_activity"
    permission_required = "submissions.view_submission"

    @property
    def submission(self):
        if not hasattr(self, "_submission"):
            self._submission = get_object_or_404(Submission, id=self.kwargs["submission_id"])
        return self._submission

    def get_queryset(self):
        logic_log_template = "logging/events/submission_logic_evaluated.txt"
        queryset = self.submission.logs.filter(template=logic_log_template)
        return queryset.order_by("timestamp")

    def get_context_data(self):
        context = super().get_context_data()
        context.update({
            "opts": Submission._meta,
            "original": self.submission,
        })
        return context
```

src/openforms/submissions/templates/submission_logs/submission_logs.html
```diff
@@ -0,0 +1,55 @@
...
+{% block content %}
+    <h1>{% trans  "Logic logs" %}</h1>
+    {% block field_sets %}
+        {% if not logs_activity %}
+            {% trans "No logic rules for that submission" %}
+        
+        {% else %}
+            {% for log in logs_activity  %}
+                <fieldset class="module aligned collapse collapsed">
+                    <h2>{{ log.timestamp|date:'Y-m-d H:i:s' }} - {% trans 'Logic' %} - </h2>
```
sergei-maertens, line 28: I think this would benefit the readability by using the `timesince` template filter: https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#timesince

Then, you typically want to combine that with semantic markup so you can still see the exact timestamp on hover:

```suggestion
                    <h2>
                        <time datetime="{{ log.timestamp.isoformat }}" title="{{ log.timestamp|date:'Y-m-d H:i:s' }}">{{ log.timestamp|timesince }}</time>
                        - {% trans 'Logic' %}
                    </h2>
```

src/openforms/submissions/templates/submission_logs/submission_logs.html
```diff
@@ -0,0 +1,55 @@
...
+                    <h4> {% trans "Logic rules" %} </h4>
+                    <table class="logic-rules-tables">
+                        <thead>
+                            <th class="rule-width"> {% trans "Rule" %} </th>
+                            <th class="trigger-width"> {% trans "Trigger" %} </th>
```
sergei-maertens, line 33:
```suggestion
                            <th class="trigger-width"> {% trans "Triggered" %} </th>
```

src/openforms/submissions/templates/submission_logs/submission_logs.html
```diff
@@ -0,0 +1,55 @@
...
+                            <th> {% trans "Action" %} </th>
+                        </thead>
+                        <tbody>
+                            {% for rule in log.extra_data.log_evaluated_rules %}
+                                    <tr>
+                                        <td>
+                                            {{ rule.source_components }}
+                                        </td>
+                                        <td>{{rule.trigger|yesno|capfirst}} </td>
```
sergei-maertens, line 42:
```suggestion
                                        <td>{{ rule.trigger|yesno|capfirst }} </td>
```

src/openforms/submissions/tests/test_admin.py
```diff
@@ -143,3 +148,71 @@ def test_change_view_displays_logs_if_not_avg(self):
...
+            user=self.anonymous_user,
+            status=status.HTTP_403_FORBIDDEN,
+        )
+
+    def test_valid_permission_view(self):
+        self.app.get(
+            reverse("admin:logs-evaluated-logic", args=(self.submission.pk,)),
+            user=self.admin_user,
+            status=status.HTTP_200_OK,
+        )
```
sergei-maertens, line 180: this needs a bit more nuance - you also want to test the following user types:

* authenticated user, not staff, no permissions
* authenticated user, staff, no permissions
* authenticated user, no staff, with permissions
* authenticated user, staff, with permissions (but not superuser)

sergei-maertens (changes requested):

src/openforms/submissions/templates/submission_logs/submission_logs.html
Ewen-Lgy: I agree, but the problem is that if you add a tag inside the H2 one, it will be disabled if the collapse element is hidden. The only way to display is to add text inside the H2 tag directly I think

PR #1733, opened 2022-06-29
Make zaaktype fields optional — [#1723]

### LINKED ISSUE #1723: Make setting a zaakstatus optional
### Thema / Theme

Admin

Currently, our pattern to create a Zaak is to also set the Zaakstatus. This is configured with mandatory fields in the form registration tab for StUF ZDS.

We need to adjust the react+model to make this optional. If not filled in, don't include the status in the StUF ZDS template (while at it, make zaaktype description optional as well - so only code is required)

We need the same for ZGW APIs but there it's a separate API call to status that needs to be done or not.

To accommodate Roxit

_No response_

dariomory:
Fixes #1723

### DIFF
--- a/src/openforms/registrations/contrib/stuf_zds/plugin.py
+++ b/src/openforms/registrations/contrib/stuf_zds/plugin.py
@@ -37,16 +37,16 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
         help_text=_("Zaaktype code for newly created Zaken in StUF-ZDS"),
     )
     zds_zaaktype_omschrijving = serializers.CharField(
-        required=True,
+        required=False,
         help_text=_("Zaaktype description for newly created Zaken in StUF-ZDS"),
     )
 
     zds_zaaktype_status_code = serializers.CharField(
-        required=True,
+        required=False,
         help_text=_("Zaaktype status code for newly created zaken in StUF-ZDS"),
     )
     zds_zaaktype_status_omschrijving = serializers.CharField(
-        required=True,
+        required=False,
         help_text=_("Zaaktype status omschrijving for newly created zaken in StUF-ZDS"),
     )
 
--- a/src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
+++ b/src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
@@ -317,7 +317,10 @@ class StufZDSPluginTests(VariablesTestMixin, StufTestBase):
         }
 
         plugin = StufZDSRegistration("stuf")
-        result = plugin.register_submission(submission, form_options)
+        serializer = plugin.configuration_options(data=form_options)
+        self.assertTrue(serializer.is_valid())
+
+        result = plugin.register_submission(submission, serializer.validated_data)
         self.assertEqual(
             result,
             {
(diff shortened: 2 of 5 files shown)


sergei-maertens (changes requested):
The options form has changed, but I'm not seeing any tests or changes to the StUF-ZDS plugin code to actually deal with the situation where no values are provided for these fields.

The ticket describes what should or shouldn't happen in the StUF-ZDS registration.

sergei-maertens (changes requested):

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -37,16 +37,19 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
...
     zds_zaaktype_status_code = serializers.CharField(
-        required=True,
+        required=False,
+        allow_null=True,
         help_text=_("Zaaktype status code for newly created zaken in StUF-ZDS"),
     )
     zds_zaaktype_status_omschrijving = serializers.CharField(
-        required=True,
+        required=False,
+        allow_null=True,
```
sergei-maertens, line 52: there is no reason to set `allow_null=True` - string-based inputs should only have one type of empty value: the empty string.

You need to update the test to use empty strings instead to reflect actual user input.

sergei-maertens (changes requested):

src/openforms/registrations/contrib/stuf_zds/plugin.py
dariomory: @sergei-maertens The API sends null values in the payload:

image

sergei-maertens: it should be empty strings instead of null

src/stuf/stuf_zds/templates/stuf_zds/soap/creeerZaak.xml
```diff
@@ -86,8 +90,12 @@
...
         <ZKN:heeft StUF:entiteittype="ZAKSTT" StUF:verwerkingssoort="T">
             <ZKN:gerelateerde StUF:entiteittype="STT" StUF:verwerkingssoort="I">
-                <ZKN:code>{{ zds_zaaktype_status_code }}</ZKN:code>
-                <ZKN:omschrijving>{{ zds_zaaktype_status_omschrijving }}</ZKN:omschrijving>
+                {% if zds_zaaktype_status_code %}
+                    <ZKN:code>{{ zds_zaaktype_status_code }}</ZKN:code>
+                {% endif %}
+                {% if zds_zaaktype_status_omschrijving %}
+                    <ZKN:omschrijving>{{ zds_zaaktype_status_omschrijving }}</ZKN:omschrijving>
+                {% endif %}
```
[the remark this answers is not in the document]
sergei-maertens: clarified - it is!

src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -645,6 +660,416 @@ def test_retried_registration_with_internal_reference(self, m, mock_task):
...
+        self.assertEqual(
+            TimelineLogProxy.objects.filter(
+                template="logging/events/stuf_zds_success_response.txt"
+            ).count(),
+            6,
+        )
+
+    @patch("celery.app.task.Task.request")
+    def test_plugin_optional_fields_missing_status_description(self, m, mock_task):
+        submission = SubmissionFactory.from_components(
```
sergei-maertens, line 869: I'm missing the variant where `zds_zaaktype_status_omschrijving` is set but `zds_zaaktype_status_code` isn't

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Backported in b45116d09908784cf7504766aeb58e7003cb820e, 5a736729a56f5caf97ac753bd4edf45e5f1f54d4 for stable/1.0.x and ea4d386225ffece8c81a2c74dc3103a9501a4249, e7dfbb3cbd938048cf83de0467d0da6bee47792a for stable/1.1.x

PR #1735, opened 2022-06-30
Convert JSONSchema to FormIO

### LINKED ISSUE #1734: Import a Product from the PDC API to generate the basic form
### Thema / Theme

Admin

With the upcomming PDC API featuring Products with a ProductAanvraag containing a JSON schema, we can link a form to a Product. However, the PDC API is not done and there's more feature that are usefull when linking product.

So, for now, we just want to import a JSON schema, which will generate a basic Open Forms form!

**Tasks**

* [ ] Continue on code in ...
* [ ] Add a button "Import Product" to the Product page in the Open Forms admin
* [ ] Add view for the Import Product page to select a JSON-schema file.
* [ ] Once the JSON-schema is imported, the result should be a Open Forms form instance with 1 step, containing all fields that are in the JSON-schema. The form name can be the title from the JSON-schema and the step name can be "Product velden".

_No response_

_No response_

joeribekker:
Fixes #1734 

Code by @frekkenSnork but was hanging in a fork of a fork.

### DIFF
--- a/src/openforms/products/utils/entry.py
+++ b/src/openforms/products/utils/entry.py
@@ -1,0 +1,26 @@
+import json
+
+from .factory import FieldFactory
+
+
+def convert_json_schema_to_py(json_schema: dict) -> dict:
+    """
+    reads json schema in json format and return dict of fields objects
+    """
+    final_result = {"components": []}
+    properties = json_schema.get("properties", {})
+    for key, content in properties.items():
+        type_flag = content.get("type")
+        if type_flag != "object":
+            # check if attr "required" exists
+            if json_schema.get("required") is not None:
+                required = key in json_schema.get("required")
+            else:
+                required = None
+            single_field = FieldFactory.create(key, required=required, content=content)
+        else:
+            single_field = FieldFactory.create(key, content=content)
+        elem = json.dumps(single_field.dict_repr)
+        final_result["components"].append(json.loads(elem))
+
+    return final_result
--- a/src/openforms/products/utils/factory.py
+++ b/src/openforms/products/utils/factory.py
@@ -1,0 +1,61 @@
+from .formio_classes import (
+    DateTimeField,
+    DayField,
+    EmailField,
+    FieldSet,
+    NumberField,
+    RadioField,
+    SelectBoxesField,
+    TextAreaField,
+    TextField,
+    TimeField,
+)
+
+
+class StringTypeFactory:
+    """create instances matching form io classes based on JSON schema type==string"""
+
+    CLASS_TYPES = {
+        "date": DayField,
+        "time": TimeField,
+        "date-time": DateTimeField,
+        "email": EmailField,
+    }
+
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """make distinction for string format fields"""
+        type_format = kwargs["content"].get("format")
+        _class = cls.CLASS_TYPES.get(type_format, None)
+        if type_format is not None:
+            return _class(*args, **kwargs)
+        else:
+            if kwargs["content"].get("maxLength") is not None:
+                return TextField(*args, **kwargs)
+            else:
+                return TextAreaField(*args, **kwargs)
+
+
+class FieldFactory:
+    """create instances matching form io classes based on different JSON schema types"""
+
+    CLASS_TYPES = {
+        "string": StringTypeFactory,
+        "number": NumberField,
+        "integer": NumberField,
+        "array": SelectBoxesField,
+        "boolean": RadioField,
+        "object": FieldSet,
+    }
+
+    @classmethod
+    def create(cls, *args, **kwargs):
+        _type = kwargs.get("content").get("type")
+        _class = cls.CLASS_TYPES.get(_type, None)
+        if _type == "string":
+            obj = StringTypeFactory.create(*args, **kwargs)
+        elif _type == "object":
+            obj = FieldSet(*args, **kwargs)
+        else:
+            obj = _class(*args, **kwargs)
+        return obj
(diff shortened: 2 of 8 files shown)


src/openforms/products/utils/formio_classes.py
```diff
@@ -0,0 +1,143 @@
...
+        self.defaultValue = content_dict.get("default")
+        self.input = True
+
+
+class FieldSetBase(BaseField):
+    """parent class with basic attributes for a fieldset or panel (form io)"""
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.input = False
```
sergei-maertens, line 33:
```suggestion
class FieldSetBase(BaseField):
    """parent class with basic attributes for a fieldset or panel (form io)"""
    input = False
```

src/openforms/products/utils/formio_classes.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class FieldSet(FieldSetBase):
+    def __init__(self, *args, **kwargs):
+        from .entry import convert_json_schema_to_py
+
+        super().__init__(*args)
+        self.label = "Field Set"  # docs
+        self.type = "fieldset"
+        self.components = convert_json_schema_to_py(kwargs["content"])["components"]
```
sergei-maertens, line 43:
```suggestion
class FieldSet(FieldSetBase):
    label = "Field Set"  # docs
    type = "fieldset"
    
    def __init__(self, *args, **kwargs):
        from .entry import convert_json_schema_to_py

        super().__init__(*args)
        self.components = convert_json_schema_to_py(kwargs["content"])["components"]
```

src/openforms/products/utils/tests/test_nested_object.py
```diff
@@ -0,0 +1,60 @@
+from django.test import TestCase
+
+from ..entry import convert_json_schema_to_py
+
+
+class NumberFieldTestCase(TestCase):
+    def setUp(self):
+        self.json_schema_nested_object = {
```
sergei-maertens, line 8: always call `super().setUp()` in tests

src/openforms/products/utils/tests/test_string.py
```diff
@@ -0,0 +1,59 @@
...
+        self.assertEqual(
+            result["components"][0]["validate"].get("pattern"), "^[A-Z0-9]?$"
+        )
+
+    def test_string_textarea(self):
+        result = convert_json_schema_to_py(self.json_schema_textarea)
+        self.assertEqual(result["components"][0]["type"], "textarea")
+        self.assertEqual(result["components"][0]["key"], "review")
+        self.assertEqual(result["components"][0]["description"], "Product review")
+        self.assertEqual(result["components"][0]["validate"].get("required"), False)
```
sergei-maertens, line 59: Please keep your test data setup, test call and assertions together in the test method. Now the `setUp` method sets up the test data for ALL the tests, but there's only ever one thing relevant.

The most readable test cases don't have a `setUp` method and follow the pattern:

```
def test_something(...):
    # set up test data, one or multiple lines

    # call the thing you're testing

    # make assertions, one or more lines
```

Note the deliberate blank lines to separate the set up / test / assertions parts of a test. Ideally, a test is understood by looking at just the method and nothing more.

sergei-maertens (commented):
I'd also re-organize the code to not be dumped in some generic `utils` package, but instead something like `json_schema_importer` or something like that.

src/openforms/products/utils/entry.py
```diff
@@ -0,0 +1,26 @@
...
+        type_flag = content.get("type")
+        if type_flag != "object":
+            # check if attr "required" exists
+            if json_schema.get("required") is not None:
+                required = key in json_schema.get("required")
+            else:
+                required = None
+            single_field = FieldFactory.create(key, required=required, content=content)
+        else:
+            single_field = FieldFactory.create(key, content=content)
```
joeribekker, line 22: I don't see why object is special (in this function). I even think an object can be required, which now cannot be the case. The code can be easier (please doublecheck!)

```suggestion
        required = key in json_schema.get("required", [])
        single_field = FieldFactory.create(key, required=required, content=content)
```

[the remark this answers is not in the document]
joeribekker: I think its covered by FieldFactory
```
        "array": SelectBoxesField,
```

src/openforms/products/utils/factory.py
```diff
@@ -0,0 +1,61 @@
...
+        "number": NumberField,
+        "integer": NumberField,
+        "array": SelectBoxesField,
+        "boolean": RadioField,
+        "object": FieldSet,
+    }
+
+    @classmethod
+    def create(cls, *args, **kwargs):
+        _type = kwargs.get("content").get("type")
```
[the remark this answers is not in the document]
frekkenSnork: changed return value to empty dict

src/openforms/products/utils/entry.py
frekkenSnork: Thought about it. Looks  having a fieldset as "required" not logic; some single fields in the fieldset could. But I'll explore it further.

joeribekker: well, whether you do something with it in the FieldSet is upto the FieldSet class - not this function.

src/openforms/products/utils/formio_classes.py
```diff
@@ -0,0 +1,143 @@
...
+                "minWords": None,
+                "maxWords": None,
+            }
+        )
+
+
+class DayField(Field):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.type = "day"
```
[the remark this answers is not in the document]
joeribekker: I pointed Tanja at Form.IO, and only now we're making the move to Open Forms, but yes you have a point ofcoirse.

frekkenSnork:
Need to add/change logic in case of "enum" presence in the field description

frekkenSnork:
Need a feedback how to split flow for unique user choice: SelectField vs RadioField  depending on JSON schema.Thx

PR #1740, opened 2022-07-01
Update allowed headers for nonce CSP header — [#1730]

### LINKED ISSUE #1730: Csp nonce creates problem concerning CORS policy

latest

Passing the CSPNonce to the SDK init call causes error concerning CORS policy.

"Access to fetch at ... from origin ... has been blocked by CORS policy: Request header field x-csp-nonce is not allowed by Access-Control-Allow-Headers in preflight response."

Embed the form in a cross domain context.

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

dariomory:
Fixes #1730


src/csp_post_processor/__init__.py
```diff
@@ -1,7 +1,9 @@
+from .constants import NONCE_HTTP_HEADER
 from .processor import post_process_html
 
 __version__ = "0.1.0"
 
-__all__ = ["post_process_html"]
+__all__ = ["post_process_html", "default_headers"]
 
 default_app_config = "csp_post_processor.apps.CSPPostProcessConfig"
+default_headers = (NONCE_HTTP_HEADER,)
```
sergei-maertens, line 9: there's no need for this indirection, make it explicit and just import the `NONCE_HTTP_HEADER` in the settings and add it to the `CORS_ALLOW_HEADERS` setting

sergei-maertens (changes requested):

src/csp_post_processor/__init__.py
dariomory: Thanks for the suggestion. I prefer this because it is more consistent with current settings and there's little indirection in exposing a variable from a package.

sergei-maertens: we want settings to be as obvious as possible, especially when all the relevant code is under our control. When looking at the `base.py` module, I want to be able to see exactly what is in a setting without having to track down function calls or aliases or whatever. Please apply the change.

sergei-maertens: > there's no need for this indirection, make it explicit and just import the `NONCE_HTTP_HEADER` in the settings and add it to the `CORS_ALLOW_HEADERS` setting

in `src/openforms/conf/base.py`:

```py
# imports at the top
from csp_post_processor.constants import NONCE_HTTP_HEADER

...

# CORS settings section
CORS_ALLOW_HEADERS = (
    list(default_cors_headers)
    + [NONCE_HTTP_HEADER]
    + config("CORS_EXTRA_ALLOW_HEADERS", split=True, default=[])
)
```

src/csp_post_processor/__init__.py
```diff
@@ -1,7 +1,9 @@
+from .constants import NONCE_HTTP_HEADER
 from .processor import post_process_html
 
 __version__ = "0.1.0"
 
-__all__ = ["post_process_html"]
+__all__ = ["post_process_html", "default_csp_headers"]
 
 default_app_config = "csp_post_processor.apps.CSPPostProcessConfig"
+default_csp_headers = (NONCE_HTTP_HEADER,)
```
sergei-maertens, line 9: I am reiterating my original feedback:

> there's no need for this indirection, make it explicit and just import the `NONCE_HTTP_HEADER` in the settings and add it to the `CORS_ALLOW_HEADERS` setting

in `src/openforms/conf/base.py`:

```py
# imports at the top
from csp_post_processor.constants import NONCE_HTTP_HEADER

...

# CORS settings section
CORS_ALLOW_HEADERS = (
    list(default_cors_headers)
    + [NONCE_HTTP_HEADER]
    + config("CORS_EXTRA_ALLOW_HEADERS", split=True, default=[])
)
```

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Backported in 27adbd5abe8e060325fc5d2254ba35901b4bb233 and 7211830a52776788172bf43aaac158514d10418c

PR #1753, opened 2022-07-05
[#1747] Bugs default values with variables

### LINKED ISSUE #1747: Default values for the variables

latest

Issues:
- Here there are initial values for the variables of different types: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/models/form_variable.py#L36. But, `boolean` should have `False` as in Formio, `float` should have `None`. 
- The file component is of type object, but having `{}` as default value breaks it.
- When checking what data we should return here https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/models/submission_value_variable.py#L36 we check that the value is not an empty string. We should check against the default value/empty value.
- If a variable has a default value, even if the form field is hidden it ends up in the final form data. => Note: if the logic makes it visible and then hidden again, also without variables this value ends up in the data. So this is a separate bug (https://github.com/open-formulieren/open-forms/issues/1755).

SilviaAmAm:
Fixes #1747 

**Changes**

- Before, all variables related to a step were included in the step data with their default value at the beginning of the step.
Now, if a variable is not in the configuration (i.e. its hidden) then it is not added to the data.
- The component without an explicit default value have a corresponding component variable with `initial_value=None`.
- The function `bulk_create_or_update_from_data` deletes any saved variables which are no longer present in the data.
- User defined variables are initialised (saved) at the beginning of filling out the form (after prefills).

### DIFF
--- a/src/openforms/forms/models/form_variable.py
+++ b/src/openforms/forms/models/form_variable.py
@@ -35,13 +35,24 @@ STATIC_INITIAL_VALUES = {FormVariableStaticInitialValues.now: get_now}
 
 INITIAL_VALUES = {
     FormVariableDataTypes.string: "",
-    FormVariableDataTypes.boolean: "",
+    FormVariableDataTypes.boolean: False,
     FormVariableDataTypes.object: {},
     FormVariableDataTypes.array: [],
-    FormVariableDataTypes.int: "",
-    FormVariableDataTypes.float: "",
+    FormVariableDataTypes.float: None,
     FormVariableDataTypes.datetime: "",
     FormVariableDataTypes.time: "",
+}
+
+COMPONENT_DATATYPES = {
+    "date": "datetime",
+    "time": "time",
+    "file": "array",
+    "currency": "float",
+    "number": "float",
+    "checkbox": "boolean",
+    "selectboxes": "object",
+    "npFamilyMembers": "object",
+    "map": "array",
 }
 
 
@@ -77,28 +88,33 @@ class FormVariableManager(models.Manager):
             ):
                 continue
 
-            form_variables.append(
-                self.model(
-                    form=form_step.form,
-                    form_definition=form_step.form_definition,
-                    prefill_plugin=glom(
-                        component,
-                        Path("prefill", "plugin"),
-                        default="",
-                        skip_exc=KeyError,
-                    ),
-                    prefill_attribute=glom(
-                        component,
-                        Path("prefill", "attribute"),
-                        default="",
-                        skip_exc=KeyError,
-                    ),
-                    key=component["key"],
-                    is_sensitive_data=component.get("isSensitiveData", False),
-                    source=FormVariableSources.component,
-                    data_type=get_component_datatype(component),
-                )
+            new_variable = self.model(
+                form=form_step.form,
+                form_definition=form_step.form_definition,
+                prefill_plugin=glom(
+                    component,
+                    Path("prefill", "plugin"),
+                    default="",
+                    skip_exc=KeyError,
+                ),
+                prefill_attribute=glom(
+                    component,
+                    Path("prefill", "attribute"),
+                    default="",
+                    skip_exc=KeyError,
+                ),
+                key=component["key"],
+                is_sensitive_data=component.get("isSensitiveData", False),
+                source=FormVariableSources.component,
+                data_type=get_component_datatype(component),
             )
+            default_value = component.get("defaultValue")
+            if default_value is not None:
+                new_variable.initial_value = default_value
+            else:
+                new_variable.initial_value = new_variable.get_initial_value()
+
+            form_variables.append(new_variable)
 
         return self.bulk_create(form_variables)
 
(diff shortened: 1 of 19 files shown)


src/openforms/forms/models/form_variable.py
```diff
@@ -212,7 +202,7 @@ def get_initial_value(self):
         if self.source == FormVariableSources.static:
             return STATIC_INITIAL_VALUES[self.initial_value]()
         else:
-            return self.initial_value or INITIAL_VALUES[self.data_type]
+            return self.initial_value or None
```
sergei-maertens, line 206: How does this behave with an empty string is default value? I'm not sure how Formio saves that in the form builder.

src/openforms/forms/models/form_variable.py
```diff
@@ -97,6 +86,7 @@ def create_for_formstep(self, form_step: "FormStep") -> List["FormVariable"]:
                     is_sensitive_data=component.get("isSensitiveData", False),
                     source=FormVariableSources.component,
                     data_type=get_component_datatype(component),
+                    initial_value=component.get("defaultValue"),
```
SilviaAmAm, line 89: Make a `get_default_value` in utils for formio documenting that Formio has the empty value and default value concepts + links to github

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -360,7 +360,7 @@ function reducer(draft, action) {
 
       // Issue #1729 - Workaround for bug in FormIO
       if (mutationType === 'changed' && !schema.multiple && Array.isArray(schema.defaultValue)) {
-        schema.defaultValue = getEmptyValue(schema);
+        schema.defaultValue = null;
```
SilviaAmAm, line 363: comment that formio will fallback to the emptyvalue of the component.

src/openforms/submissions/api/viewsets.py
```diff
@@ -119,6 +120,7 @@ def perform_create(self, serializer):
         conf = GlobalConfiguration.get_solo()
         if conf.enable_form_variables:
             prefill_variables(serializer.instance)
+            initialise_variables_unrelated_to_a_step(serializer.instance)
```
SilviaAmAm, line 123: When resuming a form we will have to refresh the static variables?
Maybe we shouldn't persist static variables to databases?

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -18,41 +18,82 @@
...
+
+    @property
+    def saved_variables(self):
+        return {
+            variable_key: variable
+            for variable_key, variable in self.variables.items()
+            if variable.pk
+        }
 
     def get_variable(self, key: str) -> Optional["SubmissionValueVariable"]:
```
SilviaAmAm, line 41: Cannot return none?

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -18,41 +18,82 @@
...
 class SubmissionValueVariablesState:
-    variables: Dict[str, "SubmissionValueVariable"]
+    submission: "Submission"
+    _variables: Dict[str, "SubmissionValueVariable"] = None
+
+    def __init__(self, submission: "Submission"):
+        self.submission = submission
+
+    @property
+    def variables(self):
```
SilviaAmAm, line 28: Add type hints

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -115,12 +151,17 @@ def bulk_create_or_update_from_data(
...
         variables_to_update = []
+        variables_ids_to_delete = []
         for key, variable in submission_variables.items():
             try:
                 variable.value = glom(data, key)
             except PathAccessError:
                 if update_missing_variables:
-                    variable.value = variable.form_variable.get_initial_value()
+                    if variable.pk:
+                        variables_ids_to_delete.append(variable.id)
```
SilviaAmAm, line 161: Maybe just remove the deleted variables from memory? Save some queries

src/openforms/js/components/admin/form_design/variables/utils.js
```diff
@@ -64,7 +64,7 @@ const updateFormVariables = (
           prefillPlugin: newComponent.prefill?.plugin || '',
           prefillAttribute: newComponent.prefill?.attribute || '',
           isSensitiveData: newComponent.isSensitiveData,
-          initialValue: newComponent.defaultValue || '',
+          initialValue: getDefaultValue(newComponent),
```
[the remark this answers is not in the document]
SilviaAmAm: Created an issues for this: https://github.com/open-formulieren/open-forms/issues/1760

src/openforms/submissions/api/viewsets.py
SilviaAmAm: Created issue: https://github.com/open-formulieren/open-forms/issues/1761

sergei-maertens (approved):

PR #1757, opened 2022-07-06
:truck: Move FormLogicFactory from submissions to forms app

sergei-maertens:
The model is defined in the forms app, so that's where the factory
model belongs too.


src/openforms/forms/tests/test_api_form_logic.py
```diff
@@ -14,8 +14,11 @@
 
 from openforms.accounts.tests.factories import SuperUserFactory
 from openforms.forms.models import FormLogic
-from openforms.forms.tests.factories import FormFactory, FormStepFactory
-from openforms.submissions.tests.form_logic.factories import FormLogicFactory
+from openforms.forms.tests.factories import (
```
SilviaAmAm, line 17: These could be relative imports

SilviaAmAm (approved):

PR #1758, opened 2022-07-06
Support postponing logic rule evaluation until a certain step is reached

### LINKED ISSUE #1546: Logic rules in other steps can affect whether the current step can be submitted

From SDK commit https://github.com/open-formulieren/open-forms-sdk/commit/1bdf03a9d20887113e0cdb9d57b1c255b3d8436f

It looks like that when there is a rule that applies to a different step evaluates to true when the form is not filled can cause the first step to not be submittable. 

Create a form with two steps:
1. Step 1: only a content component
2. Step 2: two textfields, one called `aField` and the other `bField`

Add an advanced logic rule with trigger:
`{"==":[{"var":"aField"},{"var":"bField"}]}`
and with action 'blokkeer doorgaan naar de volgende stap'.

Start the form, see that Step 1 cannot be submitted ('next' button is blocked)

Step 1 should be submittable

### Screen resolution

_No response_

### Browser

_No response_

### OS

_No response_

### Browser

_No response_

sergei-maertens:
Fixes #1546

Screencast: https://watch.screencastify.com/v/MNRbjnogrR9GR42XRgaA

**Changes**

- [x] Add trigger point/step to logic rule model
- [x] Expose setting the trigger via the API
    - Validate that the step belongs to the related form
    - Validate that the step exists
- [x] Implement skipping rules during evaluation that belong to future steps
- [x] Implement the frontend code/UI to configure the trigger step for a logic rule (advanced configuration?)


src/openforms/js/components/admin/form_design/StepSelection.js
```diff
@@ -0,0 +1,39 @@
...
+  // look up the step from the array of steps in the context
+  const step = formSteps.find(element => element.url == formStepUrl);
+  return {
+    step,
+    stepName: getStepDisplayName(step),
+  };
+};
+
+export default StepSelection;
+export {useFormStep, StepSelection};
```
SilviaAmAm, line 39: Why exporting `StepSelection` both as default and not default?

src/openforms/scss/components/admin/_logic-trigger.scss
```diff
@@ -1,12 +1,29 @@
...
+@import '~microscope-sass/lib/bem';
 @import '~microscope-sass/lib/color';
+@import '~microscope-sass/lib/responsive';
 
 $logic-line-breakpoint: 1800px !default;
 
+$logic-trigger-border: solid 1px #eee;
+
+.logic-trigger-container {
+  padding: 1em;
```
SilviaAmAm, line 10: Shouldn't the variables in the `src/openforms/ui/static/ui/scss/_settings.scss` file be used instead of hardcoded values?

SilviaAmAm (commented):
Looks good :)

Maybe it would be good to mention this new option in the documentation?

src/openforms/js/components/admin/form_design/StepSelection.js
sergei-maertens: didn't want to break existing `default` imports and also just want to expose using only named imports. it doesn't make that much difference to be honest, so I can remove one if you'd like?

src/openforms/scss/components/admin/_logic-trigger.scss
sergei-maertens: yes but no? :P those variables aren't very portable/exact science - spacing ratio's heavily depend on their parent context. Besides that, the `UI` scss stuff is still leftovers from the prototype that should be revisited at some point.

And I'd rather use CSS variables than sass variables for this kind of stuff so you can tackle it with theming more easily (e.g. a theme with larger spacing for people that desire it vs. a compact theme for people who want to see as much as possible on a single page)

src/openforms/forms/models/form_variable.py
```diff
@@ -86,6 +86,7 @@ def create_for_formstep(self, form_step: "FormStep") -> List["FormVariable"]:
                         skip_exc=KeyError,
                     ),
                     key=component["key"],
+                    name=component["key"],
```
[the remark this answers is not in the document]
sergei-maertens: I think I ran into it mostly while debugging tests and the 'name' being an empty string, so it didn't carry much information to see what was going on. The key is the only thing (almost) guaranteed to be set in tests, so I think I prefer it for that reason? Maybe it should be a `component.get("label") or component["key"]` instead, that way it'll still work nicely in non-test code and also work in tests?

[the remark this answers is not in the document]
SilviaAmAm: Sounds good!

src/openforms/js/components/admin/form_design/StepSelection.js
SilviaAmAm: Ok! I think its fine, just wanted to double check that it was intentional :)

PR #1764, opened 2022-07-08
[#1710] Added editGrid to form builder

### LINKED ISSUE #1710: Repeating components in Forms
### Thema / Theme

Other

DH Taiga 42

Body of the Taiga issue: 

In een formulier er is een veldengroep waar de gebruiker informatie over hun auto moet toevoegen: 

--Veldengroep auto-----  
Kenteken:  
Kleur: …  
Electrisch: …

Heb je nog een auto? 
ja /nee

---------------------------------

Als de gebruiker nog een auto heeft dan verschijnt hetzelfde veldengroep nog een keer.

Dit momenteel wordt gebouwd in openforms door de veldengroep te herhaalen voor het maximale aantal keren dat de gebruikers gegevens zou kunnen invullen (en daarna met logica die zichtbaar/verborgen te maken).

Dit is moeilijk te doen als de veldengroep veel keer moet worden herhaald.

=> This should be implemented using the variables:
There should be a variable keeping track of how many times a group should be repeated and then with a loop the field can be repeated the desired number of times in the configuration.

_No response_

_No response_

SilviaAmAm:
Fixes #1710
 
**Changes**

- Added repeating groups (edit grid in Formio) to the form builder
- Added rendering node 

Related frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/215


src/openforms/formio/rendering/default.py
```diff
@@ -154,3 +157,67 @@ def display_value(self) -> str:
...
+                    ...
+                ]
+            }
+
+        So we need to repeat the child nodes of the configuration and associate them with the data
+        provided by the user.
+        """
+        repeats = 0
+        if self.value:
+            repeats = len(self.value)
```
sergei-maertens, line 201:
```suggestion
        repeats = len(self.value) if self.value else 0
```
a bit more concise?

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -362,13 +362,19 @@ function reducer(draft, action) {
         schema.defaultValue = null;
       }
 
+      // TODO: This could break if a reusable definition is used multiple times in a form
+      const step = draft.formSteps.filter(
+        step => step.formDefinition === formDefinition || step._generatedId === formDefinition
+      )[0];
```
sergei-maertens, line 368: * you can/should use `Array.find` for this rather than `.filter(...)[0]`
* looks a bit like a candidate for https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/StepSelection.js#L35 or rather split out the matching/filter functionality into a utility function that's used here and in the hook? I'm looking to "contain" the implementation details of the private `_generatedId` vs. the URL thing

sergei-maertens (changes requested):
Cool stuff! Some nitpicking here and there.

I *think* I'd like to see each repeating group in an edit group as a `<li>` element of an `<ol>` rather than making that implicit, but I think we'll have to check that with Joeri :)

sergei-maertens:
@SilviaAmAm Joeri and I discussed how we want to present the repeating groups and came to this conclusion:

For plain text e-mail, we can do an extra indentation level:

```
Your reference is: OF-E3PU7W

Samenvatting

Car details

- Personal Details
 - Name: John
 - Surname: Doe
 - Email: test@email.nl
- Cars
  1. - Brand: Citroen
     - Radio: Blue
     - How old is the car (years): 4
  2. - Brand: Fiat
     - Radio: White
     - How old is the car (years): 78

Kind regards,

Open Forms
```

For the HTML based things (the other screenshots), we can add a label (in bold, is okay) above each repeated group, e.g. in this case **Car 1**, **Car 2** etc. You could add an extra option to the component design `itemLabel` at the same level as the `label` and then you interpolate the `index + 1` at the end of it.

SilviaAmAm:
Decided to go for 
```
Car details

- Personal Details 
 - Name: Silvia
 - Surname: Amam
- Cars 
 - Car 1 
  - Brand: Fiat
  - Radio: Red
  - How old is the car (years): 6
 - Car 2 
  - Brand: Maserati
  - Radio: Blue
  - How old is the car (years): 5
```
For the plain text email confirmation, because it makes it easier to format

SilviaAmAm:
Email:
Screenshot from 2022-07-15 12-21-26
 PDF:
 
Screenshot from 2022-07-15 12-22-32

src/openforms/emails/templates/emails/templatetags/form_steps_summary.html
```diff
@@ -20,7 +20,8 @@ <h3 style="margin: 0 0 1em; font-size: 1.2em;">{{ submission_step_node.render }}
             {% with add_bottom_margin=forloop.last %}
                 {% if component_node.label or component_node.display_value %}
                     {% ifchanged component_node.layout_modifier %}
-                        {% if component_node.layout_modifier == 'fieldset' %}
+                        {% has_bold_label component_node.layout_modifier as make_bold%}
+                        {% if make_bold %}
```
sergei-maertens, line 24: This could/should be done as providing the modifiers in the rendering context rather than inventing a custom template tag for this. I feel the custom tag  is too specific and spreads out the formatting logic across different tags/template files.

sergei-maertens (approved):

PR #1766, opened 2022-07-08
Avoid invalidating token based on accessed date — [#1719]

### LINKED ISSUE #1719: Allow submission reports to be downloaded for a limited time
### Thema / Theme

API

Currently, you can only download a submission report once! After that you get a 403 (see screenshot). The link should be valid for 1 hour instead so you can download it multiple times.

UPDATE: I heard from sergei that our current token generator can work on day-level. So if it's valid for 1 day thats also fine.

image

Less calls to the support desk

_No response_

dariomory:
Fixes #1719


src/openforms/submissions/api/views.py
```diff
@@ -30,8 +30,7 @@
     summary=_("Download the PDF report"),
     description=_(
         "Download the PDF report containing the submission data. The endpoint requires "
-        "a token which is tied to the submission from the session. Once the PDF is "
-        "downloaded, this token is invalidated. The token also automatically expires "
+        "a token which is tied to the submission from the session. The token also automatically expires "
```
sergei-maertens, line 33:
```suggestion
        "a token which is tied to the submission from the session. The token automatically expires "
```

sergei-maertens (changes requested):

sergei-maertens (changes requested):
@dariomory I've rebased this on master so that we don't have such a large number of changes in the `django.po` file (ran `makemessages` on master).

Please fix the failing OAS check so that we can merge this.

sergei-maertens (approved):

Issue #1611, opened 2022-05-16
Form logic editor action doesn't handle unsaved steps


Bartvaderkin:
### Product versie / Product version

1.0.x development (e9e90f16)

### Omschrijf het probleem / Describe the bug

The select-widget for FormSteps in the logic editor doesn't handle unsaved steps and uses a blank string as the option value instead of a URL.

### Stappen om te reproduceren / Steps to reproduce

1. Edit a Form and add a new Step but don't save.
2. In the Logica tab create a rule and as action choose "mark step as not applicable" and try to select the unsaved step.
3. Note the widget won't select the step but jumps back to the empty option.
4. In you inspect the html element you notice the value attribute of the new step's option element is empty

### Verwacht gedrag / Expected behavior

I'd expect to be able to select the items presented in the widget.

### Screen resolution

1920 x 1080

### Device

Desktop

### OS

Linux

### Browser

Firefox

PR #1767, opened 2022-07-11
Refactor form-saving machinery in admin

sergei-maertens:
This refactor was overdue and is necessary to sort out #1611 without (too) ugly hacks.

**Changes**

* All of the actual data loading when opening/starting a (new) form is contained in the `data/` package, with a module per "topic"
* All of the API calls that need to be made for saving a form are also contained in this package.
* Validation (and other 4xx) errors are now thrown by the unsafe API calls (post, put, delete...) and caught by the callers where they are collected for display

Since we optimize a bunch of things by making API calls in parallel (for both loading and persisting data), we need to be able to distinguish between succeeded and failed calls. For succeeded calls, the state needs to be update (so that we don't create the same object _again_ after correcting validation errors) and for failed calls, we need to display the validation errors so the end user can correct these.

This means that we specifically intercept validation error exceptions and process them, while any other errors propagate upwards and result in a "generic error message". When doing calls in parallel, we no longer discard the entire batch as soon as one of them invalidates, but we process both the succeeded calls and update the state while also returning validation errors.

This allowed for some refactors where we can rely on the incrementally updated state for subsequent calls, and that enables the solution for #1611 by using temporary references rather than broken/no references at all to steps that only exist in memory.

### DIFF
--- a/src/openforms/js/components/admin/form_design/data/complete-form.js
+++ b/src/openforms/js/components/admin/form_design/data/complete-form.js
@@ -1,0 +1,275 @@
+import produce from 'immer';
+
+import {ExtendableError, ValidationErrors} from '../../../../utils/exception';
+import {post, put, apiDelete} from '../../../../utils/fetch';
+import {FORM_ENDPOINT} from '../constants';
+import {saveLogicRules, savePriceRules} from './logic';
+import {updateOrCreateFormSteps} from './steps';
+import {createOrUpdateFormVariables} from './variables';
+import {createFormVersion} from './versions';
+
+/**
+ * Save the form itself without any related objects.
+ */
+const saveForm = async (state, csrftoken) => {
+  const {
+    newForm: isNewForm,
+    form: {uuid},
+  } = state;
+  const formDetails = produce(state, draft => {
+    return {
+      ...draft.form,
+      literals: draft.literals,
+      authenticationBackends: draft.selectedAuthPlugins,
+    };
+  });
+
+  const createOrUpdate = isNewForm ? post : put;
+  const endpoint = isNewForm ? FORM_ENDPOINT : `${FORM_ENDPOINT}/${uuid}`;
+
+  // throws on HTTP 400, HTTP 401 or any non-OK status.
+  let response;
+  try {
+    response = await createOrUpdate(endpoint, csrftoken, formDetails, true);
+  } catch (e) {
+    // wrap validation errors so the component knows where to display the errors
+    if (e instanceof ValidationErrors) {
+      e.context = 'form';
+      throw e;
+    }
+    // unknown, re-throw
+    throw e;
+  }
+
+  // update with the backend generated data, like UUID and URL. Note that this is a noop
+  // for form updates.
+  const newState = produce(state, draft => {
+    const {uuid, url, literals} = response.data;
+    draft.form.uuid = uuid;
+    draft.form.url = url;
+    draft.literals = literals;
+    draft.newForm = false; // it's either created now, or updated -> both are not 'new form anymore'
+  });
+  return newState;
+};
+
+/**
+ * Save the form steps and their related form definitions + report back any
+ * validation errors.
+ */
+const saveSteps = async (state, csrftoken) => {
+  const createdFormDefinitions = [];
+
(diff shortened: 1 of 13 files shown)


src/openforms/js/components/admin/form_design/data/steps.js
```diff
@@ -0,0 +1,123 @@
...
+  try {
+    definitionResponse = await definitionCreateOrUpdate(
+      definitionEndpoint,
+      csrftoken,
+      definitionData,
+      true
+    );
+    // handle any unexpected API errors
+    if (!definitionResponse.ok) {
+      throw new FormException(
```
sergei-maertens, line 38: this is another thing to get rid off at some point

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -0,0 +1,276 @@
...
+      variable.form = formUrl;
+      if (variable.formDefinition == null) continue;
+      // if the variable.formDefinition is not a URL, we have to resolve it against the
+      // temporary client-side ID. This also allows us to update the state with the
+      // actual resolved resource URLs.
+      try {
+        new URL(variable.formDefinition);
+      } catch {
+        const formStep = stepsByGeneratedId[variable.formDefinition];
+        variable.formDefinition = formStep.formDefinition;
```
SilviaAmAm, line 182: In the case where I add a user defined variable with nothing filled in (invalid) and then I save the form, I see an error in the console => formStep is undefined

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -685,49 +643,9 @@ function reducer(draft, action) {
...
-        if (fieldPrefix === 'form' && fieldName === 'literals') {
-          key = err.name;
-        } else if (fieldPrefix === 'logic-rules') {
-          key = err.name;
-        } else {
-          key = `${fieldPrefix}.${err.name}`;
-        }
-        return [key, err.reason];
-      });
-      draft.validationErrors = [...draft.validationErrors, ...prefixedErrors];
```
SilviaAmAm, line 723: I noticed that now if we submit a form and the errors are in the form or form step tab, once the user resolves the errors, the warning icon in the tab disappears.
In the logic tab, this doesn't happen (the icon remains also when all errors are resolved)

SilviaAmAm (changes requested):

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -0,0 +1,276 @@
...
+    newState = await saveForm(state, csrftoken);
+  } catch (e) {
+    if (e instanceof ValidationErrors) {
+      return [state, [e]];
+    }
+    throw e;
+  }
+
+  // then, ensure the form definitions and steps are persisted
+  [newState, stepsValidationErrors] = await saveSteps(newState, csrftoken);
```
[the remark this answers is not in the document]
sergei-maertens: Found the cause - user defined variables don't have `formDefinition = null`, but `formDefinition = ''` and I was checking for null. Will be fixed in my rebase :)

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1012,194 +920,54 @@ const FormCreationForm = ({csrftoken, formUuid, formUrl, formHistoryUrl}) => {
...
       return;
     }
 
+    dispatch({
+      type: 'SUBMIT_DONE',
+      payload: {
+        updatedState: newState,
+        validationErrors,
+      },
+    });
```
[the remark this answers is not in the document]
sergei-maertens: yeah, the `dispatch` state change was not propagated to the state being passed to the `saveCompleteForm` function, causing it to briefly reset. Fixed it now!

src/openforms/js/components/admin/form_design/data/complete-form.js
sergei-maertens: Fixed, was the cause of https://github.com/open-formulieren/open-forms/pull/1767#discussion_r919870341 as well

PR #1768, opened 2022-07-11
 Reusable form definition can be marked as non-reusable [#1702]

### LINKED ISSUE #1702: Reusable form definition can be marked as non-reusable even if it's used in another form

1.0.9, 1.1.2, latest

If a reusable form definition is used in multiple forms, there is no mechanism to prevent someone to uncheck the box "reusable".

In the admin:
1. Create a new form with label "Test form 1"
2. Add a new step
3. Create a new form definition with label "Test reusable definition"
4. Check the box "reusable?"
5. Save the form.
6. Create a new form with label "Test form 2"
7. Add a new step
8. Select a reusable definition. Pick the one created above ("Test reusable definition")
9. Uncheck the box "reusable?"
10. Save the form

Now if we go in the form definition tab of the admin, we see that the "Test reusable definition" is used in 2 forms, but is marked as non reusable.
Screenshot from 2022-06-23 16-20-23

If a reusable definition is used in multiple forms, one shouldn't be able to mark it as not-reusable.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

dariomory:
Fixes #1702

### DIFF
--- a/src/openforms/forms/tests/admin/test_form_definition.py
+++ b/src/openforms/forms/tests/admin/test_form_definition.py
@@ -133,20 +133,19 @@ class TestFormDefinitionAdmin(WebTest):
         step = FormStepFactory.create(form_definition__is_reusable=True)
         FormStepFactory.create(form_definition=step.form_definition)
 
-        self.client.force_login(user=self.user)
-
         response = self.app.get(
             reverse(
                 "admin:forms_formdefinition_change",
                 kwargs={"object_id": step.form_definition.pk},
-            )
+            ),
         )
         response.form["is_reusable"] = False
         response = response.form.submit()
 
         self.assertInHTML(
             _(
-                "Cannot change is_reusable from true to false if the form definition is used in more than one formstep."
+                "This form definition cannot be marked as 'not-reusable' as it is used "
+                "in multiple existing forms."
             ),
             str(response.content),
         )
--- a/src/openforms/forms/tests/admin/test_form_definition.py
+++ b/src/openforms/forms/tests/admin/test_form_definition.py
@@ -126,3 +126,27 @@ class TestFormDefinitionAdmin(WebTest):
             deleted_form_url,
             str(response.content),
         )
+
+    def test_cannot_mark_form_definition_as_non_reusable_if_used_in_multiple_forms(
+        self,
+    ):
+        step = FormStepFactory.create(form_definition__is_reusable=True)
+        FormStepFactory.create(form_definition=step.form_definition)
+
+        self.client.force_login(user=self.user)
+
+        response = self.app.get(
+            reverse(
+                "admin:forms_formdefinition_change",
+                kwargs={"object_id": step.form_definition.pk},
+            )
+        )
+        response.form["is_reusable"] = False
+        response = response.form.submit()
+
+        self.assertInHTML(
+            _(
+                "Cannot change is_reusable from true to false if the form definition is used in more than one formstep."
+            ),
+            str(response.content),
+        )
--- a/src/openforms/forms/tests/test_api_formdefinition.py
+++ b/src/openforms/forms/tests/test_api_formdefinition.py
@@ -178,7 +178,8 @@ class FormDefinitionsAPITests(APITestCase):
         self.assertEqual(
             response.json()["invalidParams"][0]["reason"],
             _(
-                "Cannot change is_reusable from true to false if the form definition is used in more than one formstep."
+                "This form definition cannot be marked as 'not-reusable' as it is used "
+                "in multiple existing forms."
             ),
         )
 
--- a/src/openforms/forms/tests/test_api_formdefinition.py
+++ b/src/openforms/forms/tests/test_api_formdefinition.py
@@ -1,6 +1,7 @@
 from unittest.mock import patch
 
 from django.urls import reverse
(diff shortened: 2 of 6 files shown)


src/openforms/forms/tests/test_api_formdefinition.py
```diff
@@ -156,6 +156,35 @@ def test_update(self):
         self.assertEqual(True, definition.login_required)
         self.assertIn({"label": "New field"}, definition.configuration["components"])
 
+    def test_update_is_reusable_unsuccessful_with_multiple_forms(self):
+        user = StaffUserFactory.create(user_permissions=["change_form"])
```
sergei-maertens, line 160: permissions are not relevant for this test case, so you can use `SuperUserFactory` instead.

src/openforms/forms/tests/test_api_formdefinition.py
```diff
@@ -156,6 +156,35 @@ def test_update(self):
...
+        self.client.force_authenticate(user=user)
+
+        step = FormStepFactory.create(
+            form__internal_name="InternalForm",
+            form_definition__internal_name="InternalDef",
+            order=1,
+        )
```
sergei-maertens, line 167:
```suggestion
        step = FormStepFactory.create()
```
* form name is not relevant for the test
* form definition name is not relevant for the test
* order is not relevant
* order starts at `0`
* order is set implicitly by `django-ordered-model` anyway

src/openforms/forms/tests/test_api_formdefinition.py
```diff
@@ -156,6 +156,35 @@ def test_update(self):
...
+        FormStepFactory.create(
+            form_definition=step.form_definition,
+        )
```
sergei-maertens, line 170: this is an impossible situation - the form definition is used in two forms but not marked as `reusable`.

**Mistakes**

Your test setup is wrong to begin with.

src/openforms/forms/tests/test_api_formdefinition.py
```diff
@@ -156,6 +156,35 @@ def test_update(self):
...
+            url,
+            data={
+                "isReusable": False,
+            },
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(
+            response.json()["invalidParams"][0]["reason"],
+            "Cannot change is_reusable from true to false if the form definition is used in more than one formstep.",
```
sergei-maertens, line 185: **Mistake**

Assertion message is not marked translatable. This test will crash when we update the Dutch translations.

sergei-maertens (changes requested):
**Missing tests**

* Test that you cannot mark a re-usable form definition as not-re-usable in the `FormDefinition` admin if it's used in multiple forms.

src/openforms/forms/tests/admin/test_form_definition.py
```diff
@@ -126,3 +126,27 @@ def test_deleted_forms_not_in_used_in(self):
...
+
+    def test_cannot_mark_form_definition_as_non_reusable_if_used_in_multiple_forms(
+        self,
+    ):
+        step = FormStepFactory.create(form_definition__is_reusable=True)
+        FormStepFactory.create(form_definition=step.form_definition)
+
+        self.client.force_login(user=self.user)
+
+        response = self.app.get(
```
sergei-maertens, line 138: this `self.client.force_login` does nothing, because `self.client` and `self.app` have nothing to do with each other.

src/openforms/forms/validators.py
```diff
@@ -44,3 +46,23 @@ def validate_not_deleted(form):
             _("Form is deleted."),
             code="invalid",
         )
+
+
+def validate_form_definition_is_reusable(instance: FormDefinition, **kwargs) -> None:
+    """
+    - Ensure that ``is_reusable`` can only be switched from true to false
+      if the form definition is used in less than 2 formsteps.
+    """
```
sergei-maertens, line 55:
```suggestion
    """
    Validate the integrity of the ``is_reusable`` flag.
    
    ``is_reusable`` can only be switched from true to false if the form definition is used in less than
    two distinct forms.
    """
```
`used_in` returns forms, not formsteps. In practice, this should be the same, but let's use correct wording.

src/openforms/forms/validators.py
```diff
@@ -44,3 +46,23 @@ def validate_not_deleted(form):
...
+        return
+
+    is_reusable = kwargs.get("is_reusable", instance.is_reusable)
+
+    if is_reusable is False and instance.used_in.count() > 1:
+        raise ValidationError(
+            {
+                "is_reusable": _(
+                    "Cannot change is_reusable from true to false if the form definition is used in more than one formstep."
+                ),
```
sergei-maertens, line 66:
```suggestion
                "is_reusable": _(
                    "This form definition cannot be marked as 'not-reusable' as it is used in multiple existing forms."
                ),
```
better user feedback.

sergei-maertens (changes requested):
Last details

sergei-maertens (approved):

sergei-maertens (changes requested):
Turns out after rebasing on master the build is completely broken.

sergei-maertens (approved):
I fixed the build after rebasing and did some more cleanups.

PR #1778, opened 2022-07-12
Plugin configuration behavior [#1320] 

### LINKED ISSUE #1320: Make sure all plugins from the configuration overview behave properly
The plugin overview uses very theoritical responses to check whether a connection can be made or not. Various fixes need to be done to make this work at all. This ticket is about the "Configuration overview" page.

**Tasks**

* [x] Only show and call `check_config` for plugins if they are enabled in the General settings (see screenshot 1). This should clean up the list and make the page faster to load.
* [x] StUF-ZDS cannot work with a fake BSN (causes upstream issues). This check should be refactored to only do a connection/ssl test or retrieve the WSDL if possible.
* [x] Haal Centraal, same reason, refactor to do a GET-requests that is invalid to simple check the connection/ssl
* [x] KvK registration plugin uses the wrong GET-request. Use the proper one (Basisprofiel API)
* [x] Refactor the the view to show only the error message: "Failed to validate the configuration." in case of an error. In the alt/title attribute, you can put the whole error. This should clean up the view.

Screenshot 1:
image

Screenshot 2:
image

dariomory:
Fixes #1320

### DIFF
--- a/src/openforms/config/data.py
+++ b/src/openforms/config/data.py
@@ -8,8 +8,8 @@ from django.contrib.admin.templatetags.admin_list import _boolean_icon
 class Entry:
     name: str
     status: Optional[bool] = None
-    status_message: str = ""
     actions: Optional[List[Tuple[str, str]]] = None
+    error: str = ""
 
     @property
     def status_icon(self):
--- a/src/openforms/config/templates/admin/config/overview.html
+++ b/src/openforms/config/templates/admin/config/overview.html
@@ -29,7 +29,14 @@
                         {% for entry in section.entries %}
                             <tr>
                                 <td><strong>{{ entry.name }}</strong></td>
-                                <td>{{ entry.status_icon }}{% if entry.status_message %} {{ entry.status_message|linebreaksbr }}{% endif %}</td>
+                                <td {% if entry.error %}title="{{ entry.error }}"{% endif %}>
+                                    {{ entry.status_icon }}
+                                    {% if entry.status is False %}
+                                        {% trans "Failed to validate the configuration." %}
+                                    {% elif entry.status is None %}
+                                        {% trans "Skipped" %}
+                                    {% endif %}
+                                </td>
                                 <td>
                                     {% for label, url in entry.actions %}
                                         {% if url %}
--- a/src/openforms/config/tests/test_config_check.py
+++ b/src/openforms/config/tests/test_config_check.py
@@ -1,31 +1,97 @@
 from django.test import TestCase
 from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
+
+import requests_mock
 
 from openforms.accounts.tests.factories import StaffUserFactory, UserFactory
+from openforms.config.models import GlobalConfiguration
+from stuf.stuf_zds.models import StufZDSConfig
+from stuf.stuf_zds.tests.test_backend import load_mock
+from stuf.tests.factories import StufServiceFactory
 
 
 class ConfigCheckTests(TestCase):
-    def setUp(self):
-        super().setUp()
+    url = reverse("config:overview")
 
     def test_access_permission(self):
-        url = reverse("config:overview")
         with self.subTest("anon"):
-            response = self.client.get(url)
+            response = self.client.get(self.url)
             self.assertEqual(response.status_code, 302)
 
         with self.subTest("user"):
             self.client.force_login(UserFactory())
-            response = self.client.get(url)
+            response = self.client.get(self.url)
             self.assertEqual(response.status_code, 302)
 
         with self.subTest("staff"):
             self.client.force_login(StaffUserFactory())
-            response = self.client.get(url)
+            response = self.client.get(self.url)
             self.assertEqual(response.status_code, 403)
 
         with self.subTest("staff with permission"):
             user = StaffUserFactory(user_permissions=["configuration_overview"])
             self.client.force_login(user)
-            response = self.client.get(url)
+            response = self.client.get(self.url)
             self.assertEqual(response.status_code, 200)
+
+    def test_disabled_plugins_are_skipped(self):
+        config = GlobalConfiguration.get_solo()
+        config.plugin_configuration = {
+            "registrations": {
+                "stuf-zds-create-zaak": {"enabled": False},
+                "email": {"enabled": True},
+            },
+            "prefill": {
+                "kvk-kvknumber": {"enabled": False},
+            },
+        }
+        config.save()
+
+        user = StaffUserFactory(user_permissions=["configuration_overview"])
+        self.client.force_login(user)
+
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(
+            response,
+            _("StUF-ZDS"),
+        )
+        self.assertNotContains(
+            response,
+            _("KvK Company by KvK number"),
+        )
+        self.assertContains(
+            response,
+            _("Email registration"),
+        )
+
+    @requests_mock.Mocker()
+    def test_check_config_service_failure_correctly_reported(self, m):
+        self.service = StufServiceFactory.create()
+        config = StufZDSConfig.get_solo()
+        config.service = self.service
+        config.gemeentecode = "1234"
+        config.save()
+        m.get(
+            "http://zaken/soap/?wsdl",
+            status_code=500,
+            content=load_mock("soap-error.xml"),
+            headers={"content-type": "text/xml"},
+        )
+
+        user = StaffUserFactory(user_permissions=["configuration_overview"])
+        self.client.force_login(user)
+
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(
+            response,
+            f"""<td title="{_('Could not connect: {exception}').format(
+                exception='Error while making backend request: HTTP 500'
+            )}""",
+        )
+        self.assertContains(
+            response,
+            f'{_("Failed to validate the configuration")}',
+        )
--- a/src/openforms/config/tests/test_config_check.py
+++ b/src/openforms/config/tests/test_config_check.py
@@ -1,5 +1,9 @@
+from unittest.mock import patch
+
+from django.contrib.admin.templatetags.admin_list import _boolean_icon
 from django.test import TestCase
 from django.urls import reverse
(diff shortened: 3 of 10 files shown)


src/openforms/config/views.py
```diff
@@ -24,7 +23,7 @@ def _subset_match(requested: Optional[str], checking: str) -> bool:
     return requested == checking
 
 
-class ConfigurationView(UserIsStaffMixin, PermissionRequiredMixin, TemplateView):
+class ConfigurationView(PermissionRequiredMixin, UserIsStaffMixin, TemplateView):
```
sergei-maertens, line 26: why was the order of these mixins changed?

src/openforms/config/data.py
```diff
@@ -10,6 +10,7 @@ class Entry:
     status: Optional[bool] = None
     status_message: str = ""
     actions: Optional[List[Tuple[str, str]]] = None
+    error: Optional[str] = None
```
sergei-maertens, line 13:
```suggestion
    error: str = ""
```
`bool(error)` in both situation is `False` for the default value.

Improvement: keep types to check for "empty" value simple.

src/openforms/config/views.py
```diff
@@ -95,6 +94,7 @@ def get_register_entries(self, register):
     def get_plugin_entry(self, plugin: Any) -> Entry:
         # undocumented query string support - helps for developers ;)
         requested_plugin = self.request.GET.get("plugin")
+        status_message, status, error = None, True, None
```
sergei-maertens, line 97:
```suggestion
        status_message, status, error = None, True, ""
```
-> matches type declaration for `Entry`

src/openforms/config/views.py
```diff
@@ -107,19 +107,12 @@ def get_plugin_entry(self, plugin: Any) -> Entry:
...
-            status_message = _("Internal error: {exception}").format(exception=e)
-            status = False
-        else:
-            status_message = None
-            status = True
-
+            status_message, status, error = (
+                _("Failed to validate the configuration."),
+                False,
+                e,
```
sergei-maertens, line 114: **Mistakes**

* `e` is of `Exception` type, while the type hint expects `str` type

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -113,39 +113,20 @@ def get_co_sign_values(cls, identifier: str) -> Tuple[Dict[str, Any], str]:
...
                 raise InvalidPluginConfiguration(_("Service not selected"))
 
             client = config.service.build_client()
-            client.retrieve(
+            client.list(
                 "ingeschrevenpersonen",
-                burgerservicenummer=check_bsn,
                 request_kwargs=self.request_kwargs,
+                params={"burgerservicenummer": "123"},
             )
```
sergei-maertens, line 126: Requirement:

> refactor to do a GET-requests that is invalid to simple check the connection/ssl

**Mistake**

You are not doing an invalid GET request.

**Rationale**

We can not make "real" calls _as described in the ticket_. At best, we can do connection tests.

src/stuf/stuf_zds/client.py
```diff
@@ -18,11 +18,7 @@
 from requests import RequestException, Response
 
 from openforms.config.models import GlobalConfiguration
-from openforms.logging.logevent import (
-    stuf_zds_failure_response,
-    stuf_zds_request,
-    stuf_zds_success_response,
-)
+from openforms.logging import logevent
```
sergei-maertens, line 21: Next time, do these refactors in a separate PR - they clutter the diff for PR review.

src/stuf/stuf_zds/client.py
```diff
@@ -395,44 +391,25 @@ def create_zaak_attachment(
...
+            logevent.stuf_zds_request(self.service, url)
+            response = requests.get(
+                url,
+                auth=self.service.get_auth(),
+                cert=self.service.get_cert(),
+                verify=self.service.get_verify(),
             )
-        except RegistrationFailed as e:
-            if not e.response:
+            if response.status_code < 200 or response.status_code >= 400:
```
sergei-maertens, line 404:
```suggestion
            if not response.ok:
```

src/stuf/stuf_zds/client.py
```diff
@@ -395,44 +391,25 @@ def create_zaak_attachment(
...
-        context = self._get_request_base_context()
+        url = f"{self.service.get_endpoint(EndpointType.beantwoord_vraag)}?wsdl"
 
         try:
-            self._make_request(
-                template,
-                context,
-                endpoint_type=EndpointType.beantwoord_vraag,
-                soap_action="geefZaakdetails_ZakLv01",
+            logevent.stuf_zds_request(self.service, url)
```
sergei-maertens, line 397: this is NOT a StUF-ZDS request, nor should it be logged

src/openforms/config/templates/admin/config/overview.html
```diff
@@ -29,7 +29,7 @@ <h1>{% trans 'Configuration' %}</h1>
                         {% for entry in section.entries %}
                             <tr>
                                 <td><strong>{{ entry.name }}</strong></td>
-                                <td>{{ entry.status_icon }}{% if entry.status_message %} {{ entry.status_message|linebreaksbr }}{% endif %}</td>
+                                <td {% if entry.error %}title="{{ entry.error }}"{% endif %}>{{ entry.status_icon }}{% if entry.status_message %} {{ entry.status_message|linebreaksbr }}{% endif %}</td>
```
sergei-maertens, line 32: Requirement:

> Refactor the the view to show only the error message: "Failed to validate the configuration." in case of an error. In the alt/title attribute, you can put the whole error. This should clean up the view.

The literal text "Failed to validate the configuration." may be missing if there was no exception raised.

```suggestion
                                <td {% if entry.error %}title="{{ entry.error }}"{% endif %}>{{ entry.status_icon }}{% if entry.status == False %}{% trans "Failed to validate the configuration." %}{% endif %}</td>
```

We don't need `status_message` at all anymore I think.

sergei-maertens (changes requested):
**Requirements errors**

> Only show and call check_config for plugins if they are enabled in the General settings (see screenshot 1). This should clean up the list and make the page faster to load.

-> You don't check if the plugin is enabled before running the config tests.

src/openforms/config/views.py
dariomory: Just a minor change to avoid IDE warnings:

image

Apparently it violates DJG301. I'm okay placing the builtin mixins first or last if there is some consistency.

dariomory:
@sergei-maertens 

> -> You don't check if the plugin is enabled before running the config tests.

Looks like this is already in place (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/config/views.py#L89)

P.S. Thank you for making the extra effort with simplifying the points.

sergei-maertens:
But why does your IDE give that warning? The order is deliberate so that
the staff status is checked first.

src/openforms/config/views.py
sergei-maertens: So apparently this is a PyCharm check (https://pycharm-security.readthedocs.io/en/latest/checks/DJG301.html) that isn't smart enough to see that the mixins to the left of it are also `AccessMixin`. Your IDE is giving wrong information in this case.

Don't blindly follow IDE warnings without understanding what you are changing, and don't fix what isn't broken. Let's revert this change.

sergei-maertens (commented):
I'm missing tests:

* Assert that disabled plugins are skipped
* Assert that service failures lead to correct reporting in the UI

The coverage of `config.views` is only 80%.

src/openforms/config/tests/test_config_check.py
```diff
@@ -1,31 +1,98 @@
...
+            },
+        }
+        config.save()
+
+        user = StaffUserFactory(user_permissions=["configuration_overview"])
+        self.client.force_login(user)
+
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertInHTML(
```
sergei-maertens, line 56: django has a built in assertion for this: `assertContains`

src/openforms/config/tests/test_config_check.py
```diff
@@ -1,31 +1,98 @@
...
+            _("StUF-ZDS"),
+            response.rendered_content,
+            count=0,
+        )
```
sergei-maertens, line 60: I would expect this to NOT be present since the plugin is disabled

src/openforms/config/tests/test_config_check.py
```diff
@@ -1,31 +1,98 @@
...
+        self.assertInHTML(
+            _("KvK Company by KvK number"),
+            response.rendered_content,
+            count=0,
+        )
```
sergei-maertens, line 65: I would expect this to NOT be present since the plugin is disabled

src/openforms/config/tests/test_config_check.py
```diff
@@ -1,31 +1,98 @@
...
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            f"""<td title="{_('Could not connect: {exception}').format(exception='Error while making backend request: HTTP 500')}""",
+            response.rendered_content,
+        )
+        self.assertIn(
+            f'{_("Failed to validate the configuration")}',
+            response.rendered_content,
+        )
```
sergei-maertens, line 98:
```suggestion
        self.assertContains(response, _("Failed to validate the configuration"))
```

src/openforms/config/tests/test_config_check.py
```diff
@@ -1,31 +1,98 @@
...
+            headers={"content-type": "text/xml"},
+        )
+
+        user = StaffUserFactory(user_permissions=["configuration_overview"])
+        self.client.force_login(user)
+
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            f"""<td title="{_('Could not connect: {exception}').format(exception='Error while making backend request: HTTP 500')}""",
```
sergei-maertens, line 92: this is not readable, please break it up in separate lines.

sergei-maertens (changes requested):

dariomory:
Thanks for the minor feedback @sergei-maertens. Please don't forget to include **why** the changes should be made.

PR #1779, opened 2022-07-13
Feature/1751 display logic rules ncely

Ewen-Lgy:
This feature introduces a nice display for logic rules instead of the untreated json_logic_trigger and actions related. This feature takes into account advanced rules, unary operators.

Also, the logic display is now realized in python directly instead of templates, which allows more flexibilty (That is why some templates files have been deleted)

I just adapted the previous tests from the #1682 feature but maybe adding some tests to test the new utils functions might be relevant.

### DIFF
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -3,8 +3,10 @@ from typing import TYPE_CHECKING, List, Optional, Tuple
 
 from django.db.models import Model
 
+from openforms.forms.constants import LogicActionTypes
 from openforms.logging.constants import TimelineLogTags
 from openforms.payments.constants import PaymentStatus
+from openforms.utils.json_logic import ComponentMeta, introspect_json_logic
 
 if TYPE_CHECKING:  # pragma: nocover
     from openforms.accounts.models import User
@@ -439,29 +441,105 @@ def submission_export_list(form: "Form", user: "User"):
 
 
 def submission_logic_evaluated(
-    submission: "Submission", evaluated_rules, updated_submission_data
+    submission: "Submission", evaluated_rules, resulting_data
 ):
-
     if not evaluated_rules:
         return
-    log_evaluated_rules = []
+    evaluated_rules_list = []
 
+    submission_state = submission.load_execution_state()
+    form_steps = submission_state.form_steps
+
+    # Keep a mapping for each component's key for each component in the entire form.
+    # The key of the mapping is the component key, the value is a dataclasse composed of
+    # form_step and component defintion
+    components_map = {}
+    for form_step in form_steps:
+        for component in form_step.form_definition.iter_components():
+            components_map[component["key"]] = ComponentMeta(form_step, component)
+
+    input_data = []
+    all_variables = submission.form.formvariable_set.distinct("key").in_bulk(
+        field_name="key"
+    )
+    component_key_lookup = {
+        LogicActionTypes.value: "component",
+        LogicActionTypes.property: "component",
+        LogicActionTypes.variable: "variable",
+    }
     for evaluated_rule in evaluated_rules:
 
         rule = evaluated_rule["rule"]
+        rule_introspection = introspect_json_logic(
+            rule.json_logic_trigger, components_map, resulting_data
+        )
+
+        # Gathering all the input component of each evaluated rule
+        input_data.extend(rule_introspection.get_input_components())
+
+        targeted_components = []
+        for action in rule.actions:
+            action_details = action["action"]
+            component_key = action.get(component_key_lookup.get(action_details["type"]))
+            component_meta = components_map.get(component_key)
+            component = component_meta.component if component_meta else None
+
+            # figure out the best possible label
+            # 1. fall back to component label if there is a label, else empty string
+            # 2. if there is a variable, use the name if it's set, else fall back to
+            # component label
+            label = component.get("label", "") if component else ""
+            if component_key in all_variables:
+                label = all_variables[component_key].name or label
+
+            action_log_data = {
+                "key": component_key,
+                "type_display": LogicActionTypes.get_choice(
+                    action_details["type"]
+                ).label,
+                "label": label,
+                "step_name": component_meta.form_step.form_definition.name
+                if component_meta
+                else "",
+                "value": "",
+            }
+
+            # process the value
+            if action_details["type"] == LogicActionTypes.value:
+                action_log_data["value"] = action_details["value"]
+            elif action_details["type"] == LogicActionTypes.property:
+                action_log_data.update(
+                    {
+                        "value": action_details["property"]["value"],
+                        "state": action_details["state"],
+                    }
+                )
+            elif action_details["type"] == LogicActionTypes.variable:
+                action_logic_introspection = introspect_json_logic(
+                    action_details["value"], components_map, resulting_data
+                )
+                action_log_data["value"] = action_logic_introspection.as_string()
+            targeted_components.append(action_log_data)
 
         evaluated_rule_data = {
-            "source_components": rule.json_logic_trigger,
-            "targeted_components": rule.actions,
+            "raw_logic_expression": rule_introspection.expression,
+            "readable_rule": rule_introspection.as_string(),
+            "targeted_components": targeted_components,
             "trigger": evaluated_rule["trigger"],
         }
 
-        log_evaluated_rules.append(evaluated_rule_data)
+        evaluated_rules_list.append(evaluated_rule_data)
+
+    # de-duplication of input data
+    deduplicated_input_data = {node["key"]: node for node in input_data}
 
     _create_log(
         submission,
         "submission_logic_evaluated",
-        extra_data={"log_evaluated_rules": log_evaluated_rules},
+        extra_data={
+            "evaluated_rules": evaluated_rules_list,
+            "input_data": deduplicated_input_data,
+        },
     )
 
 
--- a/src/openforms/setup.py
+++ b/src/openforms/setup.py
@@ -43,6 +43,7 @@ def setup_env():
     load_self_signed_certs()
     monkeypatch_requests()
     monkeypatch_mozilla_django_oidc_get_from_settings()
+    monkeypatch_json_logic()
 
 
 def load_self_signed_certs() -> None:
@@ -140,3 +141,13 @@ def mute_deprecation_warnings():
     """
     if not sys.warnoptions:
         warnings.simplefilter("ignore", DeprecationWarning, append=False)
+
+
+# TODO Apply this fix to the fork of json-logic.py
+def monkeypatch_json_logic():
+
+    from json_logic import operations
+
+    operations["in"] = (
+        lambda a, b: a in b if "__contains__" in dir(b) and a is not None else False
+    )
--- a/src/openforms/submissions/templates/admin/submissions/submission/logs.html
+++ b/src/openforms/submissions/templates/admin/submissions/submission/logs.html
@@ -17,14 +17,12 @@
 {% endblock %}
 
 {% block content %}
-    <h1>{% trans  "Logic logs" %}</h1>
     {% block field_sets %}
         {% if not logs_activity %}
             {% trans "No logic rules for that submission" %}
         
         {% else %}
             {% for log in logs_activity  %}
-                <span class="timestamp-log"></span>
                 <fieldset class="module aligned collapse collapsed fieldset-log">
                     <h2>
                         <time datetime="{{ log.timestamp.isoformat }}" title="{{ log.timestamp|date:'Y-m-d H:i:s' }}">
@@ -32,22 +30,48 @@
                         </time>
                         - {% trans 'Logic' %} -
                     </h2>
-                    <h4> {% trans "Logic rules" %} </h4>
-                    <table class="logic-rules-tables">
+                    <h4> {% trans "Variables" %} </h4>
+                    <table class="logic-logs logic-logs--variables">
                         <thead>
-                            <th class="rule-width"> {% trans "Rule" %} </th>
-                            <th class="trigger-width"> {% trans "Triggered" %} </th>
+                            <th class="logic-logs__col-header logic-logs__col-header--name"> {% trans "Property name" %} </th>
+                            <th class="logic-logs__col-header logic-logs__col-header--value"> {% trans "Value" %} </th>
+                        </thead>
+                        <tbody>
+                            {% for key, value in log.extra_data.input_data.items  %}
+                            <tr>
+                                <td>{{ value.step_name }}: <code>{{ key }}</code> &#40;{{ value.label }}&#41;</td>
+                                <td>{{ value.actual_value }}</td>
+                            </tr>
+                            {% endfor %}
+                        </tbody>
+                    </table>
+                    <h4> {% trans "Logic rules" %} </h4>
+                    <table class="logic-logs logic-logs--rules">
+                        <thead>
+                            <th class="logic-logs__col-header logic-logs__col-header--rule"> {% trans "Rule" %} </th>
+                            <th class="logic-logs__col-header logic-logs__col-header--trigger"> {% trans "Triggered" %} </th>
                             <th> {% trans "Action" %} </th>
                         </thead>
                         <tbody>
-                            {% for rule in log.extra_data.log_evaluated_rules %}
+                            {% for rule in log.extra_data.evaluated_rules %}
                                     <tr>
                                         <td>
-                                            {{ rule.source_components }}
+                                            {{ rule.readable_rule }}
                                         </td>
                                         <td>{{ rule.trigger|yesno|capfirst }} </td>
                                         <td>
-                                            {{ rule.targeted_components }}
+                                        {% for action in rule.targeted_components %}
+                                            {% if action.key %}
+                                                <b>{{ action.type_display }}</b>: {{ action.step_name }}: <code>{{ action.key }}</code> ({{ action.label }}):
+                                                    {% if not 'state' in action %}
+                                                        <i>{{ action.value }}</i>
+                                                    {% else %}
+                                                        <i>{{ action.value }} = {{ action.state|yesno|capfirst }}</i>
+                                                    {% endif %}
+                                            {% else %}
+                                                <b>{{ action.type_display }}</b>
+                                            {% endif %}
+                                        {% endfor %}
                                         </td>
                                     </tr>
                             {% endfor %}
--- a/src/openforms/submissions/tests/test_variables/test_num_queries.py
+++ b/src/openforms/submissions/tests/test_variables/test_num_queries.py
@@ -127,8 +127,9 @@ class SubmissionVariablesPerformanceTests(VariablesTestMixin, APITestCase):
         # 8. Delete submission attachment files
         # 9. RELEASE SAVEPOINT
         # 10. Delete submission values
-        # 11. Create log for rules
-        with self.assertNumQueries(11):
+        # 11. Retrieve all form_variables
+        # 12. Creation of timelinelog
+        with self.assertNumQueries(12):
             evaluate_form_logic(submission, submission_step2, data)
 
     def test_update_step_data(self):
--- a/src/openforms/utils/json_logic.py
+++ b/src/openforms/utils/json_logic.py
@@ -1,14 +1,20 @@
 """
 Utilities to parse/process jsonLogic expressions.
 """
-from dataclasses import dataclass, field
-from typing import List, Union
+from dataclasses import asdict, dataclass, field
+from typing import TYPE_CHECKING, Any, Dict, List, Union
 
 from json_logic import jsonLogic
+
+if TYPE_CHECKING:
+    from openforms.formio.typing import Component
+    from openforms.forms.models import FormStep
 
 __all__ = ["JsonLogicTest"]
 
 JSONLogicValue = Union[str, int, "JsonLogicTest"]
+JsonLogicExpression = Dict[str, Any]
+JsonLogicNode = Union["JsonLogicLeaf", "JsonLogicOperation"]
 
 
 @dataclass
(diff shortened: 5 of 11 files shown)


src/openforms/logging/logevent.py
```diff
@@ -441,18 +449,94 @@ def submission_export_list(form: "Form", user: "User"):
...
+    # Mapping each form's key to :
+    # - the form_step it belongs
+    # - The component related to it
+    keys_mapped_to_form = {}
+    keys_mapped_to_component = {}
+    for formstep in formsteps:
+        keys_mapped_to_form = {
+            **{component["key"]: formstep for component in formstep.iter_components()},
+            **keys_mapped_to_form,
+        }
```
sergei-maertens, line 465: I think it's a bit more concise to "update" the dict for every formstep:

```py
keys_mapped_to_form.update({
    component["key"]: formstep
    for component in formstep.iter_components()
})
```

src/openforms/logging/logevent.py
```diff
@@ -441,18 +449,94 @@ def submission_export_list(form: "Form", user: "User"):
...
+        keys_mapped_to_component = {
+            **flatten(formstep.form_definition.configuration),
+            **keys_mapped_to_component,
+        }
```
sergei-maertens, line 469: I would simplify this too:

```py
keys_mapped_to_component.update(flatten(formstep.form_definition.configuration))
```

src/openforms/logging/logevent.py
```diff
@@ -441,18 +449,94 @@ def submission_export_list(form: "Form", user: "User"):
 def submission_logic_evaluated(
     submission: "Submission", evaluated_rules, updated_submission_data
 ):
```
sergei-maertens, line 451: I'd rename the variables here to be a bit more generic - now the mechanisms of the actual logic evaluation are leaking into this log function, but that's not really relevant.

```suggestion
def submission_logic_evaluated(
    submission: "Submission", evaluated_rules, resulting_data
):
```

I am wondering though - the startpoint data is essentially `submission.data`, but as part of the logic evaluation, this data can change values. This masks the actual values that are used for evaluation of a particular rule. There is a delta possible between `submission.data` and `resulting_data` - even for the same key!

src/openforms/logging/logevent.py
```diff
@@ -441,18 +449,94 @@ def submission_export_list(form: "Form", user: "User"):
...
+            keys_mapped_to_form,
+            updated_submission_data,
+        )
+
+        stringified_json_tree = stringify_json_tree(source_components)
+
+        # will remove duplicate components that are involved in several rules
+        remove_duplicate_components_from_json_tree(
+            source_components, involved_components
+        )
```
sergei-maertens, line 488: I don't understand what this does - can you explain?

src/openforms/logging/logevent.py
```diff
@@ -441,18 +449,94 @@ def submission_export_list(form: "Form", user: "User"):
...
+        for targeted_component in targeted_components:
+            key = targeted_component.get("key")
+            # Adding the label of the key and step name associated if key exists
+            if not key:
+                continue
+            # Might happens that a form variable is not related to a component
+            if targeted_component.get(
+                "isVariable"
+            ) and not keys_mapped_to_component.get(key):
+                targeted_component["label"] = key
```
sergei-maertens, line 527: we should look up the form variables in that case and grab the name of the variable

src/openforms/setup.py
```diff
@@ -140,3 +141,12 @@ def mute_deprecation_warnings():
     """
     if not sys.warnoptions:
         warnings.simplefilter("ignore", DeprecationWarning, append=False)
+
+
+def monkeypatch_json_logic():
```
sergei-maertens, line 146: can you please add a TODO that we need to apply this fix to our fork of json-logic-py?

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -17,21 +17,34 @@
...
+                    <h4> {% trans "Variables" %} </h4>
+                    <table class="logic-rules-tables">
+                        <thead>
+                            <th class="property-name-width"> {% trans "Property name" %} </th>
+                            <th> {% trans "Value" %} </th>
+                        </thead>
+                        <tbody>
+                            {% for key, value in log.extra_data.involved_components.items  %}
+                            <tr>
+                                <td>{{ value.stepName }}: {{ key }} &#40;{{ value.label }}&#41;</td>
```
sergei-maertens, line 42:
```suggestion
                                <td>{{ value.stepName }}: <code>{{ key }}</code> &#40;{{ value.label }}&#41;</td>
```
+ some nice styling will make this stand out visually

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -43,11 +56,24 @@ <h4> {% trans "Logic rules" %} </h4>
...
                                         <td>
-                                            {{ rule.source_components }}
+                                            {{ rule.rule_as_str }}
                                         </td>
                                         <td>{{ rule.trigger|yesno|capfirst }} </td>
                                         <td>
-                                            {{ rule.targeted_components }}
+                                        {% for action in rule.targeted_components %}
+                                            {% if action.key %}
+                                                <b>{{ action.step_name }}:{{ action.key }} &#40;{{ action.label }}&#41;</b><i> {{ action.type }}</i>
```
sergei-maertens, line 65:
```suggestion
                                                <b>{{ action.step_name }}:{{ action.key }} ({{ action.label }})</b><i> {{ action.type }}</i> 
```

you don't need to use HTML escape codes for these things :)

src/openforms/submissions/tests/test_variables/test_num_queries.py
```diff
@@ -127,8 +127,11 @@ def test_evaluate_form_logic_with_rules(self):
         # 8. Delete submission attachment files
         # 9. RELEASE SAVEPOINT
         # 10. Delete submission values
-        # 11. Create log for rules
-        with self.assertNumQueries(11):
+        # 11. Retrieve formsteps
+        # 12. Retrieve form_defintion related to formstep 1
+        # 13. Retrieve form_defintion related to formstep 2
+        # 14. Creation of timelinelog
+        with self.assertNumQueries(14):
```
sergei-maertens, line 134: we'll want to do something about this, because the number of queries is quickly growing and that hurts performance

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,123 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
         return JsonLogicTest.from_expression(value)
 
     raise NotImplementedError(f"Unknown value type: {type(value)}")
+
+
+def find_json_logic_test_value(value: JsonLogicTest) -> str:
+
+    """returns the value inside a JsonLogicTest"""
```
sergei-maertens, line 56:
```suggestion
def find_json_logic_test_value(value: JsonLogicTest) -> str:
    """returns the value inside a JsonLogicTest"""
```
docstrings should not be separated by a blank line

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,123 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+    if isinstance(value, JsonLogicTest) and value.operator == "date":
+        return find_json_logic_test_value(value.values[0])
+    if isinstance(value, JsonLogicTest) and value.operator == "var":
+        return value.values[0]
```
sergei-maertens, line 60: Technically I could define some `{"var": {"var": "foo"}}` logic I tink? that would probably break this parsing :grimacing:

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,123 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+    else:
+        return value
+
+
+def convert_json_logic_in_json_tree(
+    logics, keys_mapped_to_component, keys_mapped_to_form, updated_submission_data
+):
+
+    """returns a dict (potentially nested) with complemetentary data regarding the field"""
```
sergei-maertens, line 70:
```suggestion
    """returns a dict (potentially nested) with complemetentary data regarding the field"""
```
docstring hygiene :)

I would also very much like to see some examples of the shape of data returned, because this docstring doesn't really tell me anything useful except that it's a dict with some structure returned. It's also unclear what "the field" is in this context.

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,123 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+    if isinstance(logics, JsonLogicTest):
+        logic_dict = {}
+        logic_dict["nodeType"] = "operation"
+        logic_dict["operator"] = logics.operator
+        logic_dict["children"] = []
```
sergei-maertens, line 76:
```suggestion
        logic_dict = {
            "nodeType": "operation",
            "operator": logics.operator,
            "children": [],
        }
```
this avoids repetition and makes the code more readable + makes the datastructure more obvious

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +49,123 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+        for logic in logics:
+            logic_list.append(
+                convert_json_logic_in_json_tree(
+                    logic,
+                    keys_mapped_to_component,
+                    keys_mapped_to_form,
+                    updated_submission_data,
+                )
+            )
+        return logic_list
```
sergei-maertens, line 132: this can probably be a list comprehension

```suggestion
        return [
            convert_json_logic_in_json_tree(
                logic,
                keys_mapped_to_component,
                keys_mapped_to_form,
                updated_submission_data,
            ) for logic in logics
        ]
```

sergei-maertens (changes requested):

src/openforms/utils/json_logic.py
Ewen-Lgy: yes indeed :grimacing:

src/openforms/logging/logevent.py
sergei-maertens: ah okay, it's de-duplication for the "variables" section then if I understand it correctly?

Ewen-Lgy: Yes, exactly

src/openforms/logging/logevent.py
```diff
@@ -439,20 +447,97 @@ def submission_export_list(form: "Form", user: "User"):
...
+            elif action_details["type"] == LogicActionTypes.variable:
+                log_action["key"] = action["variable"]
+                logic = JsonLogicTest.from_expression(action_details["value"])
+                src_components = convert_json_logic_in_json_tree(
+                    logic,
+                    keys_mapped_to_component,
+                    keys_mapped_to_form,
+                    resulting_data,
+                )
+                log_action["value"] = stringify_json_tree(src_components)
```
sergei-maertens, line 513: If all the above has been implement, you should be able to simplify this as:

```py
action_logic_introspection = introspect_json_logic(action_details["value"], components_map, resulting_data)
log_action["value"] = action_logic_introspection.as_string()
```

src/openforms/logging/logevent.py
```diff
@@ -439,20 +447,97 @@ def submission_export_list(form: "Form", user: "User"):
...
+            source_components, involved_components
+        )
+        targeted_components = []
+        for action in rule.actions:
+            log_action = {}
+
+            # Action verifications
+            action_details = action["action"]
+            log_action["value"] = ""
+            log_action["type"] = action_details["type"]
```
sergei-maertens, line 496:
```suggestion
            # normalize actions for log record display
            action_details = action["action"]
            log_action = {
                "value": "",
                "type": action_details["type"],
            }
```

src/openforms/logging/logevent.py
```diff
@@ -439,20 +447,97 @@ def submission_export_list(form: "Form", user: "User"):
...
+
+        stringified_json_tree = stringify_json_tree(source_components)
+        print(stringified_json_tree)
+
+        # will remove duplicate components that are involved in several rules
+        remove_duplicate_components_from_json_tree(
+            source_components, involved_components
+        )
+        targeted_components = []
+        for action in rule.actions:
```
sergei-maertens, line 490: In this loop, you are doing a lot of "extra" work that is probably not needed. During the loop, you basically set up a bunch of data and add it to `targeted_components`, and then you post-process `targeted_components` before passing it to the evaluated rule data.

It'd probably make more sense to do all the processing during the for loop already and convert it in exactly the data structure you want, e.g.:

```py
        all_variables = submission.form.formvariable_set.in_bulk(field_name="key")  # move this OUTSIDE of the parent loop - it only needs to be calculated once.
        targeted_components = []
        component_key_lookup = {
            LogicActionTypes.value: "component",
            LogicActionTypes.property: "component",
            LogicActionTypes.variable: "variable",
        }
        for action in rule.actions:
            action_details = action["action"]
            component_key = component_key_lookup.get(action_details["type"])
            component_meta = components_map.get(component_key)
            component = component_meta[1] if component_meta else None

            # figure out the best possible label
            # 1. fall back to component label if there is a label, else empty string
            # 2. if there is a variable, use the name if it's set, else fall back to 
            #    component label
            label = component.get("label", "") if component else "" 
            if component_key in all_variables:
                label = all_variables[component_key].name or label
            
            action_log_data = {
                "type": action_details["type"],
                "type_display": LogicActionTypes.labels[action_details["type"]],
                "label": label or "",
                "step_name": component_meta[0].form_definition.name if component_meta else "",
                "value": "",
            }

            # process the value
            if action_details["type"] == LogicActionTypes.value:
                action_log_data["value"] = action_details["value"]
            elif action_details["type"] == LogicActionTypes.property:
                action_log_data.update({
                    "value": action_details["property"]["value"],
                    "state": action_details["state"],
                })
            elif action_details["type"] == LogicActionTypes.variable:
                action_logic_introspection = introspect_json_logic(action_details["value"], components_map, resulting_data)
                action_log_data["value"] = action_logic_introspection.as_string()
            target_components.append(action_log_data)
```

In this form, it should be good to directly include in the log record extra data.

src/openforms/logging/logevent.py
```diff
@@ -439,20 +447,97 @@ def submission_export_list(form: "Form", user: "User"):
...
+                )
+                targeted_component["step_name"] = keys_mapped_to_form[
+                    key
+                ].form_definition.name
+            break
 
         evaluated_rule_data = {
-            "source_components": rule.json_logic_trigger,
-            "targeted_components": rule.actions,
+            "rule_as_str": stringified_json_tree,
```
sergei-maertens, line 539:
```suggestion
            "readable_rule": stringified_json_tree,
```
perhaps?

I think you'd still want to include the raw underlying expression as well, just for logging purposes. It's possible that the logic rule is edited after it's logged, and then all you have left is the string-representation. If you have to debug things based on logs, it would definitely be useful to have the raw logic expression stored. This also applies in the event of bugs in the string-representation calculation - we would be able to generate a new string representation and fix that.

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -17,21 +17,34 @@
...
+                        </thead>
+                        <tbody>
+                            {% for key, value in log.extra_data.involved_components.items  %}
+                            <tr>
+                                <td>{{ value.stepName }}: <code>{{ key }}</code> &#40;{{ value.label }}&#41;</td>
+                                <td>{{ value.actualValue }}</td>
+                            </tr>
+                            {% endfor %}
+                        </tbody>
+                    </table>
```
sergei-maertens, line 47: Now is the time to talk about BEM :)

The rough strokes of this article are good to familiarize yourself: https://css-tricks.com/bem-101/

Summarized - the idea is that you split your styling based on the idea of **blocks** and **elements** (the B and E from BEM). Elements only exist in the context of their parent block. With BEM, this means you use classnames in the form of `.block` and `.block__element`. One block can have many different elements, and markup-wise, they may even be nested. You cannot have elements of elements (meaning `.block__element__other-element` is illegal.

Modifiers apply to both blocks and elements, and are essentially used to implement variations of the same thing.

Here, your tables act as blocks, and I would use a class name that appropriately reflects that: `.logic-logs`. I see two variants: a table for input variables, and a table for evaluated rules, so in the markup I'd expect something like:

```html
<table class="logic-logs logic-logs--variables">
    ...
</table>

<table class="logic-logs logic-logs--rules">
    ...
</table>
```
We've applied the modifiers to the block so that we can target the styling appropriately.


Next, you have your header rows/columns - those only make sense in the context of their parent table, so they should be elements. I'd suggest something like `<th class="logic-logs__col-header">...</th>`. You can see the element syntax (double underscore).

Finally, since the different types of tables require different widths, we can target those using modifiers, e.g.:

```html
<table class="logic-logs logic-logs--variables">
    <thead>
        <th class="logic-logs__col-header logic-logs__col-header--name">{% trans "Property name" %}</th>
        <th class="logic-logs__col-header logic-logs__col-header--value">{% trans "Value" %}</th>
    </thead>
</table>

<table class="logic-logs logic-logs--rules">
    <thead>
        <th class="logic-logs__col-header logic-logs__col-header--rule"> {% trans "Rule" %} </th>
        <th class="logic-logs__col-header logic-logs__col-header--trigger"> {% trans "Triggered" %} </th>
        <th> {% trans "Action" %} </th>
        <th></th>
    </thead>
</table>
```

You would then write some SCSS in the form of:

```scss
.logic-logs {
  width: 100%;

  &--variables {
    .logic-logs__col-header {
      &--name {
        width: 70%;
      }
    }
  }

  &--rules {
    .logic-logs__col-header {
      &--rule { 
        width: 60%;
      }
      &--trigger {
        width: 10%;
      }
    }
  }
}
```

the net result is that your styles are fairly well isolated and scoped within the parent `.logic-logs` block.

sergei-maertens (changes requested):
The table layout is not correct yet:
image

The header row only has 3 columns, while the action row has 4 columns and that causes visual issues. You might consider using the `colspan` attribute to ensure you always have the correct number of table columns.

Additionally, the `disable-next` action should probably display the correctly localized label instead of the raw value, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/constants.py#L6 which has the mapping of value-label.

src/openforms/logging/logevent.py
Ewen-Lgy: It seems that we can't use in_bulk() on not unique fields, and key isn't one

src/openforms/logging/logevent.py
```diff
@@ -439,29 +441,110 @@ def submission_export_list(form: "Form", user: "User"):
...
-        log_evaluated_rules.append(evaluated_rule_data)
+        evaluated_rules_list.append(evaluated_rule_data)
+
+    # de-duplication of input data
+    de_duplicated_input_data = [
+        dict(data_tuple) for data_tuple in {tuple(data.items()) for data in input_data}
+    ]
+
+    # Mapping input data with its key
+    input_data = {data["key"]: data for data in de_duplicated_input_data}
```
sergei-maertens, line 539: took me a while to get what is going on here :P

I think a more readable version is:

```py
    deduplicated_input_data = {node["key"]: node for node in input_data}
```

You rely on the key-uniqueness for de-duplication. If a node occurs multiple times, it will simply overwrite an earlier assignment and that's fine, because they are identical anyway.

Then you can pass `deduplicated_input_data` to the `_create_log` call directly as well.

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -17,37 +17,61 @@
...
+                    <h4> {% trans "Variables" %} </h4>
+                    <table class="logic-logs logic-logs--variables">
+                        <thead>
+                            <th class="logic-logs__col-header logic-logs__col-header--name"> {% trans "Property name" %} </th>
+                            <th class="logic-logs__col-header logic-logs__col-header--value"> {% trans "Value" %} </th>
+                        </thead>
+                        <tbody>
+                            {% for key, value in log.extra_data.input_data.items  %}
+                            <tr>
+                                <td colspan="1">{{ value.stepName }}: <code>{{ key }}</code> &#40;{{ value.label }}&#41;</td>
```
sergei-maertens, line 42: `colspan="1"` is the default - you should only specify it if you span multiple columns.

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+class ComponentMeta:
+    form_step: "FormStep"
+    component: Dict[str, Any]
+
+
+@dataclass
+class IntrospectionResult:
+
+    expression: dict
+    rule_tree: Dict[Any, str]
```
sergei-maertens, line 67:
```suggestion
    rule_tree: Dict[str, Any]
```
the dict keys are strings, the values are anything.

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+    def as_string(self) -> str:
+        return stringify_json_tree(self.rule_tree)
+
+    def get_input_components(self):
+        return get_leaves(self.rule_tree)
+
+
+def introspect_json_logic(
+    expression: Dict[Any, str],
```
sergei-maertens, line 77: your types at not very consistent - I think I'd define at the module level a type:

```py
JsonLogicExpression = Dict[str, Any]
```

you can then reuse `JsonLogicExpression` for the type hints everywhere (instead of `dict` or `Dict[...]`)

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+    components_map: Dict[Any, ComponentMeta],
```
sergei-maertens, line 78:
```suggestion
    components_map: Dict[str, ComponentMeta],
```
this is a mapping of component key to meta, so should be `str` rather than `Any` (since the keys are strings).

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+    input_data: Dict[str, Any],
+) -> IntrospectionResult:
+    logic_expression = JsonLogicTest.from_expression(expression)
+    rule_tree = convert_json_logic_in_json_tree(
+        logic_expression, components_map, input_data
+    )
+    print(rule_tree)
```
sergei-maertens, line 85: don't leave `print` statements in code please :-)

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+@dataclass
+class IntrospectionResult:
+
+    expression: dict
+    rule_tree: Dict[Any, str]
+
+    def as_string(self) -> str:
+        return stringify_json_tree(self.rule_tree)
+
+    def get_input_components(self):
```
sergei-maertens, line 72: I'd like a return type hint here

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+def find_json_logic_test_value(value: JsonLogicTest) -> str:
+    """returns the value inside a JsonLogicTest"""
+    if isinstance(value, JsonLogicTest) and value.operator in ["date", "var"]:
+        return find_json_logic_test_value(value.values[0])
+    else:
+        return value
+
+
+def convert_json_logic_in_json_tree(logics, component_map, resulting_data):
```
sergei-maertens, line 97: I'd like a return type hint here

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+                "actualValue":"2022-06-21",
+                "stepName":"definition-birthdate",
+                "label":"Birthdate"
+            }
+        ],
+        "value":"2022-06-20"
+    }
+
+    :rtype: dict
+    :return: rule with additional informations related to the involved component
```
sergei-maertens, line 126: this *would* also benefit from dataclasses rather than dicts (helps a lot with the type hinting & readability)

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+                logic_dict["value"] = value
+
+        return logic_dict
+
+    # array of rules
+    else:
+        return [
+            convert_json_logic_in_json_tree(logic, component_map, resulting_data)
+            for logic in logics
+        ]
```
sergei-maertens, line 185: I'd swap the if statement to make use of a guard clause - treat the case where it isn't a `JsonLogicTest` first and get it out of the way, and then you can remove an indentation level (flat is better than nested!). So the structure becomes:

```py
if not isinstance(logics, JsonLogicTest):
    return [...]

logic_dict = {
    "nodeType": "operation",
    "operator": logics.operator,
    "children": [],
}
for index, value in enumerate(logics.values):
    ...
return logic_dict
```

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+
+def stringify_json_tree(node: Dict[Any, str]):
+    """transform the json_tree in a string of rules"""
+
+    # Unary operation
+    if not node["children"]:
+        return f"{node['operator']} {node['value']}"
+
+    tmp_string = ""
```
sergei-maertens, line 195:
```suggestion
    output = ""
```
better name?

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+        if child["nodeType"] == "leaf":
+            tmp_string += child["key"]
+        else:
+            tmp_string += f"({stringify_json_tree(child)})"
+
+        if node.get("value"):
+            tmp_string += f" {node['operator']} {node['value']}"
+        else:
+            print()
```
sergei-maertens, line 207:
```suggestion
```

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+    """transform the json_tree in a string of rules"""
+
+    # Unary operation
+    if not node["children"]:
+        return f"{node['operator']} {node['value']}"
+
+    tmp_string = ""
+    children_number = len(node["children"]) - 1
+    for index, child in enumerate(node["children"]):
+
```
sergei-maertens, line 198:
```suggestion
    children_number = len(node["children"])
    for index, child in enumerate(node["children"]):
        last = index == children_number - 1

```

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+        else:
+            tmp_string += f"({stringify_json_tree(child)})"
+
+        if node.get("value"):
+            tmp_string += f" {node['operator']} {node['value']}"
+        else:
+            print()
+            tmp_string += (
+                f"""{f" {node['operator']} " if index != children_number else ''}"""
+            )
```
sergei-maertens, line 210:
```suggestion
            output += f" {node['operator']} " if not last else ""
```
I think? I don't see the point of the outer f-string (triple quoted)

src/openforms/utils/json_logic.py
```diff
@@ -49,3 +52,174 @@ def normalize_value(value: Union[dict, list, tuple, int, str]) -> JSONLogicValue
...
+
+    if node["nodeType"] == "leaf":
+        return [node]
+    else:
+        if not node["children"]:
+            return []
+        input_components = []
+        for child in node["children"]:
+            input_components.extend(get_leaves(child))
+        return input_components
```
sergei-maertens, line 225: guard clauses can flatten your code considerably

```suggestion
def get_leaves(node: Dict[Any, str]):
    if node["nodeType"] == "leaf":
        return [node]
    return sum(
        get_leaves(child) for child in node["children"], 
        []
    )
```

sergei-maertens (changes requested):
The nitpicking step! :-)

src/openforms/utils/json_logic.py
Ewen-Lgy: Definitely

src/openforms/utils/json_logic.py
Ewen-Lgy: Something like JsonLogicOperation and JsonLogicValue ?

sergei-maertens (approved):

sergei-maertens:
Now all that's left is cleaning up the commit history and then we're good to merge!

PR #1797, opened 2022-07-21
webpack "absolute" imports

sergei-maertens:
All the goodness of `resolve.modules` from CRA in our backend project :)


SilviaAmAm (approved):
:tada:

PR #1804, opened 2022-07-22
[#1799] Copy variables when copying form

### LINKED ISSUE #1799: Copying form doesn't create new FormVariables

latest

When a form is copied, the related `FormVariables` are not created (in the same way that they are when the form is imported for example).

We should have a checklist in the documentation for things that need to be checked when adding a new feature to a form. This list should contain at least the following points:
- Importing / exporting  functions
- Copying functions
- Bulk export  / import functions

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1799

### DIFF
--- a/docs/developers/checklists.rst
+++ b/docs/developers/checklists.rst
@@ -1,0 +1,21 @@
+.. _checklists_for_developers:
+
+
+=========================
+Checklists for developers
+=========================
+
+
+Adding features to forms
+========================
+
+When you add a new feature to the ``forms`` module, make sure that the following features still work:
+
+#. Importing / exporting forms in the admin
+#. Bulk importing / exporting forms in the admin
+#. Copying a form (both with the admin action from the form list page and with the button from the form detail page)
+
+Also good to check that:
+
+#. FormVariables behave as expected
+#. Forms with reusable form definitions work as expected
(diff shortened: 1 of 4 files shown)


docs/developers/checklists.rst
```diff
@@ -0,0 +1,21 @@
+.. _checklists_for_developers:
```
sergei-maertens, line 1: we usually just mimick the filesystem path for these labels

```suggestion
.. _developers_checklists:
```

src/openforms/forms/tests/test_copy.py
```diff
@@ -184,3 +192,63 @@ def test_form_copy_staff_required(self):
         )
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+
+class CopyFormWithVarsTest(VariablesTestMixin, APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        user = SuperUserFactory.create()
```
sergei-maertens, line 200: missing `super().setUpTestData()` call

sergei-maertens (changes requested):

src/openforms/forms/tests/test_copy.py
SilviaAmAm: everytime :weary:

sergei-maertens (approved):

Issue #1711, opened 2022-06-24
Additional static variables


SilviaAmAm:
### Thema / Theme

Other

### Omschrijving / Description

DH Taiga issue 107

The list of desired new variables is:
- Date of today (with various formats like 6/23/2022 or vrijdag 10 maart 2006)
- Current month
- Current day
- Current Year
- OTAP: the name of the environment. Should be Test, Acceptation or Production. This can leverage the setting `settings.ENVIRONMENT`.
- Current month name (tbd if in english or dutch)
- Current day name
- Current time (with varying formats, like 12:33, 12:33:26)
- Current week number
- Form Name
- Form ID

SilviaAmAm:
Need to clarify why all these different formats are needed. Is it for displaying or for calculating things?

sergei-maertens:
With the implementation #1869 of #1708 we've settled on using (a sandboxed version of) the django template engine. This exposes the built in template tags and filters, so they can use the date/time formatters for all the display variations of "current timestamp". That tackles most of the extra requested variables, turning this ticket mostly into a documentation issue.

- [ ] Document support for built in template tags/filters for form designers
    - Link to Django (3.2!) built in template filters/tags documentation
    - Provide a number of examples of date formats for the requested "now" variations in a table (column template bit -> column resulting value) so there's an easy lookup table. Perhaps include a column with the description as in this ticket as well.
- [ ] OTAP: add `settings.ENVIRONMENT` as a static variable once #1825 is merged
- [ ] Form name: add as static variable once #1825 is merged (use the submissions app, not the forms app!)
- [ ] Form ID (UUID?): add as static variable once #1825 is merged (use the submissions app, not the forms app!)

sergei-maertens:
we'll hear feedback if it's not good what we did/do, but it seems like we're not getting any refinement soon

Issue #1708, opened 2022-06-24
Interface to handle inserting variables in Formio configuration


SilviaAmAm:
### Thema / Theme

Other

### Omschrijving / Description

Comment referring to this feature: https://github.com/open-formulieren/open-forms/pull/1647#discussion_r886544587

There needs to come a way to inject variables into the:
- label of certain components
- the html of content components
- the default value of components: (see Taiga DH issue 54, user story 114)

**Tasks**

- [x] Profile how the templating solution performs
- [x] depending on profiling results, investigate optimization paths

**Implementation**

- [x] Implement templating component properties
- [x] Implement form designer validation for template syntax errors

**Extra**

Probably split this in different ticket

- [ ] #1900
- [ ] #1902
- [ ] Variables in min / max configuration of date field: #1878

sergei-maertens:
Some technical analysis of what the options are. Feel free to add suggestions/remarks.

## Constraints

**Simple variables**

- `now` -> datetime (python datatype) or ISO-8601 string (for display)
- `someKey`: `123`
- `someKey`: `123.45`
- `someKey`: `"foo"`
- `someKey`: `None`
- `someKey`: `False`

-> value is always a JSON primitive: `null`, `string`, `number`, `boolean`.

**Complex variabeles**

- `someKey`: `{"foo": {"bar": "baz"}}`
- `someKey`: `["foo, "bar"]`

## What exists?

Check for existing tools to avoid re-inventing the wheel

**JQ**

URL: https://stedolan.github.io/jq/manual/

* also used in Zaken API 1.2, fits ecosystem.
* well-known for people dealing with JSON on the CLI
* https://pypi.org/project/jq/ - Python bindings exist
* originally CLI only, so really meant for parsing strings into powerful expressions
* To figure out: extensible or not?
* Supports some built in operators/functions

**glom**

URL: https://glom.readthedocs.io/en/latest/

* already used throughout the codebase
* you can do a lot with simple string specs, but might need python-level access for
  complex operations -> limited for use in formio?
* To figure out: extensible or not?

**json logic**

URL: https://jsonlogic.com/operations.html

* used for logic & actions already, single "expression language"
* very verbose
* not sure how to get an editor for json logic into formio form builder
* not sure how to display this in "human readable" format
* definitely extensible!

**jinja2**

URL: https://palletsprojects.com/p/jinja/

* Similar to Django template language
* Exposes more direct python acess though...
* extensible
[...]

PR #1808, opened 2022-07-22
[#1761] Refresh static vars

### LINKED ISSUE #1761: Refresh static variables
### Thema / Theme

API

Currently, variables that are not related to a form step (user defined variables and static variables) are persisted to the database when the form is started.

The problem with this is that if the form is saved and then resumed a week later, the value of the static variable "now" will be a week old.

Possibilities:
- Refresh static variables
- Not persisting static variables to the backend at all

_No response_

_No response_

SilviaAmAm:
Fixes #1761


src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -133,6 +137,19 @@ def remove_variables(self, keys: list) -> None:
             if key in self._variables:
                 del self._variables[key]
 
+    # TODO: Argument request present for when we will use the data in the session to put
+    #  data returned from DigiD/eHerkenning into static vars
+    def static_data(self, request: "Request") -> dict:
+        if self._static_data is None:
+            static_form_variables = self.submission.form.formvariable_set.filter(
+                source=FormVariableSources.static
+            )
```
sergei-maertens, line 146: I'm wondering - if we are not saving the values in the DB, is there even any point in storing these with the form in the DB?

* we are populating the API endpoint from a hardcoded python list of static variables
* we only do this so the form designer can load that information to include them in the form designer UI
* the form designer includes them in the API call
* we do some "tricky" things during import/copy etc. to ensure that the static variables are present

I think we can ditch static variables being persisted in the DB (even at form-design time) completely and only keep the definition in python code. Advantages:

1. We don't have to query them, ever -> free performance
2. We don't have to deal with new/updated variables in data migrations when we make changes to them in newer versions of OF -> easier upgrade path
3. Related to above - existing forms get the static variables "for free" when upgrading to 1.2.x
4. We can still present them in the UI from the API endpoint we already have, just don't save them in the backend

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -133,6 +137,19 @@ def remove_variables(self, keys: list) -> None:
...
+            static_variables_data = {}
+            for variable in static_form_variables:
+                static_variables_data[variable.key] = variable.get_initial_value()
+            self._static_data = static_variables_data
```
sergei-maertens, line 150:
```suggestion
            self._static_data = {
                variable.key: variable.get_initial_value() for variable in static_form_variables
            }
```

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
```diff
@@ -69,9 +94,12 @@ def evaluate_form_logic(
     # check what the default data values are
     defaults = get_default_values(configuration)
 
+    submission_variables_state = submission.load_submission_value_variables_state()
+    static_data = submission_variables_state.static_data(context.get("request"))
+
     # merge the default values and supplied data - supplied data overwrites defaults
     # if keys are present in both dicts
-    data = {**defaults, **data}
+    data = DataForLogic(initial_data={**defaults, **data}, static_data=static_data)
```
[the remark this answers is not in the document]
SilviaAmAm: Yes, I couldn't think of a good name and the solution was so simple :sweat_smile:

src/openforms/forms/models/form_variable.py
```diff
@@ -205,19 +205,15 @@ def __str__(self):
...
-            data_type=FormVariableDataTypes.datetime,
-            initial_value="now",
-        )
+    def get_static_data() -> List[dict]:
+        now = {
+            "name": "Now",
+            "key": "now",
+            "data_type": FormVariableDataTypes.datetime,
+            "initial_value": STATIC_INITIAL_VALUES["now"](),
+        }
```
sergei-maertens, line 217: just a comment - but I think I'd still wrap this as `FormVariable` instances with all the convenience methods, especially in light of #1708 and #1711 where we probably just want to define a python list of `FormVariable` instances.

src/openforms/submissions/form_logic.py
```diff
@@ -147,21 +177,23 @@ def evaluate_form_logic(
...
                 variable_key = action["variable"]
                 updated_step_data[variable_key] = new_value
-                updated_submission_data = {**data, **updated_step_data}
+                data_container.update_data(updated_step_data)
     step.data = DirtyData(updated_step_data)
 
     # Logging the rules
-    logevent.submission_logic_evaluated(submission, evaluated_rules, data)
+    logevent.submission_logic_evaluated(
+        submission, evaluated_rules, data_container.data
```
sergei-maertens, line 188: this should be `{**data_container.static_data, **data_container.initial_data}` - now you're still passing the mutated data to the log event instead of the original input data.

src/openforms/submissions/tests/test_variables/test_num_queries.py
```diff
@@ -59,7 +59,7 @@ def test_evaluate_form_logic_without_rules(self):
         #    variables with prefill data, so the defaultValue on the components can be set.
         # 2. Retrieve all logic rules related to a form
         # 3. Load submission state: Retrieve formsteps,
-        # 4. Load submission state: Retrieve submission steps
+        # 4.  Load submission state: Retrieve submission steps
```
sergei-maertens, line 62:
```suggestion
        # 4. Load submission state: Retrieve submission steps
```

sergei-maertens (changes requested):
Form variable validation will still have to check for key collisions with the static variables though.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -824,9 +825,7 @@ const FormCreationForm = ({csrftoken, formUuid, formUrl, formHistoryUrl}) => {
       endpoint: `${PRICE_RULES_ENDPOINT}?form=${formUuid}`,
       stateVar: 'priceRules',
     });
-  } else {
-    // only fetch this data if we're creating a new form
-    pluginsToLoad.push({endpoint: STATIC_VARIABLES_ENDPOINT, stateVar: 'formVariables'});
+    pluginsToLoad.push({endpoint: STATIC_VARIABLES_ENDPOINT, stateVar: 'staticVariables'});
```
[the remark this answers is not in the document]
SilviaAmAm: Aah good point, I only read the comment `// only fetch this data if we're creating a new form` and not the one above :see_no_evil:

sergei-maertens (approved):

PR #1809, opened 2022-07-25
Feature/1752 allow filtering in the logic logging

### LINKED ISSUE #1752: Allow filtering in the logic logging
Add a JS/CSS filter to the logic logging page using Django admin layout.

image

The default filter should be "Ja". Selecting a filter option works on ALL entries in the logging and it can work completely with JS/CSS, it just has to look like the Django admin one (which does NOT work with JS/CSS but on view/template level).

Ewen-Lgy:
Fixes #1752

### DIFF
--- a/src/openforms/js/components/admin/submissions/filter.js
+++ b/src/openforms/js/components/admin/submissions/filter.js
@@ -3,23 +3,23 @@ window.onload = function () {
   triggeredLink = document.querySelector('.trigger-filter-list > .selected > a');
 
   // Retrieve the "All" corresponding string, depending the language
-  all = document.querySelector('#trigger-filter-list__link--all').dataset.trigger;
+  all = document.querySelector('.logic-rules-filter__link--all').dataset.trigger;
 
   // Only triggered logs are displayed first
   document
-    .querySelectorAll('.logic-table-body__trigger-' + triggeredLink.dataset.trigger)
+    .querySelectorAll('.logic-logs__col-body--trigger-' + triggeredLink.dataset.trigger)
     .forEach(trigger => {
-      trigger.parentNode.classList.remove('logic-table-body__row--hidden');
+      trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
     });
 };
 
-document.querySelectorAll('.trigger-filter-list__link').forEach(link => {
+document.querySelectorAll('.logic-rules-filter__link').forEach(link => {
   link.addEventListener('click', event => {
     event.preventDefault();
 
     // Hiding all the rows first
-    document.querySelectorAll('.logic-table-body__row').forEach(row => {
-      row.classList.add('logic-table-body__row--hidden');
+    document.querySelectorAll('.logic-logs__row-body').forEach(row => {
+      row.classList.add('logic-logs__row-body--hidden');
     });
 
     let triggered = event.target.dataset.trigger;
@@ -30,12 +30,12 @@ document.querySelectorAll('.trigger-filter-list__link').forEach(link => {
 
     // Showing only rules corresponding to the selected trigger link
     if (triggered == all) {
-      document.querySelectorAll('.logic-table-body__row').forEach(row => {
-        row.classList.remove('logic-table-body__row--hidden');
+      document.querySelectorAll('.logic-logs__row-body').forEach(row => {
+        row.classList.remove('logic-logs__row-body--hidden');
       });
     } else {
-      document.querySelectorAll('.logic-table-body__trigger-' + triggered).forEach(trigger => {
-        trigger.parentNode.classList.remove('logic-table-body__row--hidden');
+      document.querySelectorAll('.logic-logs__col-body--trigger-' + triggered).forEach(trigger => {
+        trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
       });
     }
   });
--- a/src/openforms/js/components/admin/submissions/filter.js
+++ b/src/openforms/js/components/admin/submissions/filter.js
@@ -1,44 +1,48 @@
-const init = async () => {
+function init() {
   // Retrieve the filter link that displays only triggered rules
   const triggeredLink = document.querySelector('.logic-rules-filter > .selected > a');
 
-  // Only triggered rules are displayed
-  document
-    .querySelectorAll('.logic-logs__col-body--trigger-' + triggeredLink.dataset.trigger)
-    .forEach(trigger => {
-      trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
-    });
-};
-
-document.querySelectorAll('.logic-rules-filter__link').forEach(link => {
-  link.addEventListener('click', event => {
-    event.preventDefault();
-
-    // Hiding all the rows first
-    document.querySelectorAll('.logic-logs__row-body').forEach(row => {
-      row.classList.add('logic-logs__row-body--hidden');
-    });
-
-    const filterValue = event.target.dataset.trigger;
-
-    // Change the focus of the selected filter
-    document.querySelector('.selected').classList.remove('selected');
-    event.target.parentNode.classList.add('selected');
-
-    // Showing only rules corresponding to the selected trigger link
-    if (filterValue == 'all') {
+  document.querySelectorAll('.logic-rules-filter__link').forEach(link => {
+    link.addEventListener('click', event => {
+      // Hiding all the rows first
       document.querySelectorAll('.logic-logs__row-body').forEach(row => {
-        row.classList.remove('logic-logs__row-body--hidden');
+        row.classList.add('logic-logs__row-body--hidden');
       });
-    } else {
-      document
-        .querySelectorAll('.logic-logs__col-body--trigger-' + filterValue)
-        .forEach(trigger => {
-          trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
-        });
-    }
+
+      const filterValue = event.target.dataset.trigger;
+
+      // Change the focus of the selected filter
+      document.querySelector('.logic-rules-filter .selected').classList.remove('selected');
+      event.target.parentNode.classList.add('selected');
+
+      let rows;
+      switch (filterValue) {
+        case 'all': {
+          rows = document.querySelectorAll('.logic-logs__row-body');
+          break;
+        }
+        case 'yes': {
+          rows = [...document.querySelectorAll('.logic-logs__col-body--trigger-yes')].map(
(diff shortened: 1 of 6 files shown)


src/openforms/js/components/admin/submissions/filter.js
```diff
@@ -0,0 +1,44 @@
+window.onload = function () {
+  // Retrieve the filter link that displays only triggered rules
+  triggeredLink = document.querySelector('.logic-rules-filter > .selected > a');
```
sergei-maertens, line 3: we generally use strict javascript, which implies declaring all variables using `const` (preferably) or `let` (if the value needs to be re-assigned).

Global variables are a Bad Practice so we avoid using those.

Also. rather than `window.onload`, we tap into the `DOMContentLoaded` event, but even that may not be needed, e.g. in `js/components/admin/form-category.js`:

```js
if (document.readyState !== 'loading') {
  init();
} else {
  document.addEventListener('DOMContentLoaded', init);
}
```

src/openforms/scss/components/admin/_logic-evaluated-table.scss
```diff
@@ -18,5 +18,15 @@
...
+        display: none;
+      }
+    }
   }
 }
+
+#changelist-filter {
+  width: 10%;
+  margin: 0 !important;
+}
```
sergei-maertens, line 32: it seems to me that this would also affect the styling on _different_ admin pages.

If you replicate the markup from the django changelist (including the filter classnames etc.), there should not be any need to override these styles.

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -22,6 +22,25 @@
...
+                        <ul class="logic-rules-filter">
+                            <li>
+                                <a class="logic-rules-filter__link logic-rules-filter__link--all" href="" data-trigger="{% trans 'All' %}" title="{% trans "All" %}">{% trans "All" %}</a>
+                            </li>
+                            <li class="selected">
+                                <a class="logic-rules-filter__link" href="" data-trigger="{% trans 'Yes' %}" title="{% trans "Yes" %}">{% trans "Yes" %}</a>
+                            </li>
+                            <li>
+                                <a class="logic-rules-filter__link" href="" data-trigger="{% trans 'No' %}" title="{% trans "No" %}">{% trans "No" %}</a>
+                            </li>
```
sergei-maertens, line 39: * you typically don't want to include translatable strings into `data-` attributes - they are normally only used for programmatic access.
* I don't think you need the `logic-rules-filter__link--all`, since you can do a `.querySelector('.logic-rules-filter__link[data-trigger="all"]')` in Javascript to filter down on data-attribute as well.
* The `title` attributes don't add any value compared to the link text, so I'd leave them out.
* I'd use `href="#"` so that if the Javascript click capture fails, you don't get a full page reload

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -54,11 +73,11 @@ <h4> {% trans "Logic rules" %} </h4>
                         </thead>
                         <tbody>
                             {% for rule in log.extra_data.evaluated_rules %}
-                                    <tr>
+                                    <tr class="logic-logs__row-body logic-logs__row-body--hidden">
                                         <td>
                                             {{ rule.readable_rule }}
                                         </td>
-                                        <td>{{ rule.trigger|yesno|capfirst }} </td>
+                                        <td class="logic-logs__col-body logic-logs__col-body--trigger-{{ rule.trigger|yesno|capfirst }}">{{ rule.trigger|yesno|capfirst }} </td>
```
sergei-maertens, line 80: I'd make this an `{% if %} ...` statement instead of relying on `yesno` which puts things through i18n, that will also make the code more predictable/readable.

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -81,5 +100,6 @@ <h4> {% trans "Logic rules" %} </h4>
             {% endfor %}
         {% endif %}
     {% endblock %}
+    <script src="{% static "bundles/core-js.js" %}"></script>
     <script src="/static/admin/js/collapse.js"></script>
```
sergei-maertens, line 104: I'd swap the order of the script inclusions - our bundle is custom and should be last (it's also much larger, so the collapse can _maybe_ load a bit earlier).

I've also only now noticed you don't use the `{% static %}` tag for the collapse, so please modify that as well.

sergei-maertens (changes requested):
The filter layout should be similar to a changelist page in the admin - now you have the filter above the log records instead of on the right-hand side:

image

src/openforms/submissions/templates/admin/submissions/submission/logs.html
```diff
@@ -1,8 +1,8 @@
 {% extends "admin/change_form.html" %}
-{% load i18n admin_urls static %}
+{% load i18n static admin_urls admin_list forms_admin_list %}
```
sergei-maertens, line 2: `forms_admin_list` is unused, nor is `admin_list` or `admin_urls` I tink

```suggestion
{% load i18n static %}
```

src/openforms/js/components/admin/submissions/filter.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Only triggered rules are displayed
+  document
+    .querySelectorAll('.logic-logs__col-body--trigger-' + triggeredLink.dataset.trigger)
+    .forEach(trigger => {
+      trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
+    });
+};
+
+document.querySelectorAll('.logic-rules-filter__link').forEach(link => {
```
sergei-maertens, line 13: you typically only want to start performing DOM queries when the DOM has finished loading, so you'd move this entire block inside the `init` function

src/openforms/js/components/admin/submissions/filter.js
```diff
@@ -0,0 +1,47 @@
...
+  // Retrieve the filter link that displays only triggered rules
+  const triggeredLink = document.querySelector('.logic-rules-filter > .selected > a');
+
+  // Only triggered rules are displayed
+  document
+    .querySelectorAll('.logic-logs__col-body--trigger-' + triggeredLink.dataset.trigger)
+    .forEach(trigger => {
+      trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
+    });
+};
```
sergei-maertens, line 11: easier and more robust (since this code is duplicated here and in the click event handler) is to _use_ the click handler by simulating a click on the trigger link:

```js
triggeredLink.click()
```

Note that you must prevent crashes on other pages, so only do that if `triggeredLink` actually exists.

src/openforms/js/components/admin/submissions/filter.js
```diff
@@ -0,0 +1,47 @@
...
+
+    // Hiding all the rows first
+    document.querySelectorAll('.logic-logs__row-body').forEach(row => {
+      row.classList.add('logic-logs__row-body--hidden');
+    });
+
+    const filterValue = event.target.dataset.trigger;
+
+    // Change the focus of the selected filter
+    document.querySelector('.selected').classList.remove('selected');
```
sergei-maertens, line 25: I'd make this queryselector a bit more specific to prevent unintended side-effects elsewhere:

```suggestion
    document.querySelector('.logic-rules-filter .selected').classList.remove('selected');
```

src/openforms/js/components/admin/submissions/filter.js
```diff
@@ -0,0 +1,47 @@
...
+      document.querySelectorAll('.logic-logs__row-body').forEach(row => {
+        row.classList.remove('logic-logs__row-body--hidden');
+      });
+    } else {
+      document
+        .querySelectorAll('.logic-logs__col-body--trigger-' + filterValue)
+        .forEach(trigger => {
+          trigger.parentNode.classList.remove('logic-logs__row-body--hidden');
+        });
+    }
```
sergei-maertens, line 39: you can clean up this code to avoid the repetition:

```suggestion
    // Showing only rules corresponding to the selected trigger link
    let rows;
    switch (filterValue) {
      case 'all': {
        rows = document.querySelectorAll('.logic-logs__row-body');
        break;
      }
      case 'yes': {
        rows = document.querySelectorAll('.logic-logs__row-body.logic-logs__col-body--trigger-yes');
        break;
      }
      case 'no': {
        rows = document.querySelectorAll('.logic-logs__row-body.logic-logs__col-body--trigger-no');
        break;
      }
    }
    for (const row of rows) {
      row.classList.remove('logic-logs__row-body--hidden');
    }
```

sergei-maertens (changes requested):
I'm not sure what's happening now, but it looks like the updated markup is causing some alignment issues:

image

src/openforms/submissions/templates/admin/submissions/submission/logs.html
Ewen-Lgy: admin_urls is used with the admin_urlname filter

src/openforms/js/components/admin/submissions/filter.js
```diff
@@ -0,0 +1,51 @@
...
+        }
+      }
+      for (const row of rows) {
+        row.classList.remove('logic-logs__row-body--hidden');
+      }
+    });
+  });
+
+  // Simulates the click on the default selected link
+  triggeredLink.click();
```
sergei-maertens, line 44: try opening http://localhost:8000/admin/config/globalconfiguration/ in your browser and open the dev-tools, I suspect there will be a javascript error because of this line :-)

src/openforms/scss/admin/_admin_theme.scss
```diff
@@ -166,6 +166,10 @@ div.breadcrumbs {
   }
 }
 
+#changelist table thead th {
+  padding: 8px;
+}
```
sergei-maertens, line 171: this should not be a global override, but instead scoped to your own table classes where the problem is happening. Other list pages don't have this padding problem.

I'm also wondering if it doesn't make more sense to mimic the Django admin list markup for table headers instead of overriding the CSS. I believe they put some `div`s inside of the `th` normally.

sergei-maertens (changes requested):

src/openforms/scss/admin/_admin_theme.scss
Ewen-Lgy: Yes it could be a better approach

sergei-maertens (approved):

PR #1811, opened 2022-07-25
[#1697] Deal with duplicate keys in variables and form components

### LINKED ISSUE #1697: Unique keys

latest

1. If I create a component with key 'var1' in step1 and a component with key 'var1' in step 2, the second component is not linked to a variable (because of the keys conflict)
2. If I add a user defined variable and then a component with the same key, I have two variables with duplicate keys (this will raise an error in the backend but it is not yet shown in the frontend).

_No response_

1. Unique variables should be created for both components
2. There should be a warning when creating the component that this key is already in use?

**Note**
We should first discuss namespacing before tackling this issue!

SilviaAmAm:
Fixes #1697 

The cases that I tested are:
- Create 2 form components with same keys
- Create a user defined var and then a component with the same key
- Create a form component with key 'now' (clash with static variable)
- Create a user defined var with key 'now' (clash with static variable)

### DIFF
--- a/src/openforms/forms/api/serializers/form_variable.py
+++ b/src/openforms/forms/api/serializers/form_variable.py
@@ -21,23 +21,24 @@ class FormVariableListSerializer(serializers.ListSerializer):
         static_data_keys = [item.key for item in FormVariable.get_static_data()]
 
         existing_form_key_combinations = []
-        for item in attrs:
+        errors = {}
+        for index, item in enumerate(attrs):
             key_form_combination = (item["key"], item["form"].slug)
             if key_form_combination in existing_form_key_combinations:
-                raise ValidationError(
-                    _("The form and key attributes must be unique together"),
-                    code="unique",
-                )
+                errors[f"{index}.key"] = _("Key not unique within form")
+                continue
 
             if item["key"] in static_data_keys:
-                raise ValidationError(
-                    _(
-                        "The variable key cannot be equal to any of the following values: {static_data}."
-                    ).format(static_data=", ".join(static_data_keys)),
-                    code="unique",
-                )
+                errors[f"{index}.key"] = _(
+                    "The variable key cannot be equal to any of the following values: {static_data}."
+                ).format(static_data=", ".join(static_data_keys))
+                continue
 
             existing_form_key_combinations.append(key_form_combination)
+
+        if errors:
+            raise ValidationError(errors, code="unique")
+
         return attrs
 
 
--- a/src/openforms/forms/api/serializers/form_variable.py
+++ b/src/openforms/forms/api/serializers/form_variable.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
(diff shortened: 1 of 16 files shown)


src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -21,23 +21,24 @@ def validate(self, attrs):
...
-        for item in attrs:
+        errors = {}
+        for index, item in enumerate(attrs):
             key_form_combination = (item["key"], item["form"].slug)
             if key_form_combination in existing_form_key_combinations:
-                raise ValidationError(
-                    _("The form and key attributes must be unique together"),
-                    code="unique",
-                )
+                errors[f"{index}.key"] = _("Key not unique within form")
```
sergei-maertens, line 28: Maybe use `serializers.ErrorDetail(_("The variable key must be unique within a form"), code="unique")` instead for more metadata?

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -21,23 +21,24 @@ def validate(self, attrs):
         static_data_keys = [item.key for item in FormVariable.get_static_data()]
 
         existing_form_key_combinations = []
-        for item in attrs:
+        errors = {}
```
sergei-maertens, line 24: probably a `defaultdict(list)` would be more accurate, since you can have multiple errors for the same item/field

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -21,23 +21,24 @@ def validate(self, attrs):
...
-                )
+                errors[f"{index}.key"] = _(
+                    "The variable key cannot be equal to any of the following values: {static_data}."
+                ).format(static_data=", ".join(static_data_keys))
+                continue
 
             existing_form_key_combinations.append(key_form_combination)
+
+        if errors:
+            raise ValidationError(errors, code="unique")
```
sergei-maertens, line 40: since it's a collection of errors (dict), the `code` is discarded by DRF *I think*. Using `ErrorDetail` in the raw error is better suited to preserve that.

src/openforms/js/components/admin/form_design/variables/StaticData.js
```diff
@@ -41,7 +41,7 @@ const StaticData = () => {
       <ChangelistTableWrapper headColumns={headColumns} extraModifiers={['fixed']}>
         {staticData.map((item, index) => {
           return (
-            <tr className={`row${(index % 2) + 1}`}>
+            <tr className={`row${(index % 2) + 1}`} key={index}>
```
sergei-maertens, line 44: I'd use `item.key` as key instead, which should be a stable identity.

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
```diff
@@ -32,6 +33,30 @@ const SensitiveData = ({isSensitive}) => {
   );
 };
 
+const Td = ({variable, fieldName}) => {
```
sergei-maertens, line 36: can you define `Td.propTypes` please?

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
```diff
@@ -44,11 +69,16 @@ const VariableRow = ({index, variable}) => {
     return '';
   };
 
+  const hasErrors = !!Object.entries(variable.errors || {}).length;
```
sergei-maertens, line 72: this `Object.entries(objectLike || {}).length` is repeated quite often - can we check if `lodash` has something for this, and otherwise put it in a utility?

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
```diff
@@ -138,7 +168,7 @@ const EditableVariableRow = ({index, variable, onDelete, onChange}) => {
           <Select
             name="formDefinition"
             choices={formStepsChoices}
-            value={variable.formDefinition}
+            value={variable.formDefinition || ''}
```
sergei-maertens, line 171: perhaps we should fix that in the `Select` component itself - if the prop `value` is `null`, replace it with empty string?

src/openforms/js/components/admin/form_design/variables/utils.js
```diff
@@ -141,20 +142,42 @@ const updateFormVariables = (
...
+const checkForDuplicateKeys = (formVariables, staticVariables) => {
+  let validationErrors = [];
+  let existingKeys = [];
+
+  staticVariables.concat(formVariables).map((variable, index) => {
+    if (existingKeys.includes(variable.key)) {
+      validationErrors.push([`variables.${index}.key`, 'unique']);
+
+      if (!variable.errors) variable.errors = {};
+      variable.errors['key'] = 'unique';
```
sergei-maertens, line 166: maybe we can skip the whole constant for error messages, and use `defineMessage` directly here? I don't think we need this layer of indirection, and it makes it possible to cater the exact error message depending on the situation (component vs. user defined variable).

sergei-maertens (changes requested):
Found a couple annoying edge cases :grimacing: 

**Component keys in step clashing**

image

1. Define step one with component key `textField`
2. Define step two with a component key `textField`
3. Error is correctly displayed for the Component variables
4. Edit step two component key `textField` into `testVariable`

-> the variable from step one has been updated as well now, while that should not have been touched.

Perhaps you can find/restrict the right variable to update based on `variable.formDefinition` which is either a URL or a temporary ID?

**User defined unique message**

image

I'd expect the same "Key not unique within form" message or maybe even "This key is already used in a form component" instead of just the error code (?).

SilviaAmAm:
Now the cases of: 
- Add component `textFieldA`, `textFieldB`, change `textFieldB` to `textFieldA` and then back to `textFieldB` 
- Add component `testField`, add user defined variable `testField` (which becomes automatically `testField1`) and then changing the component to `testField1`

seem to work as expected.

sergei-maertens (approved):

PR #1815, opened 2022-07-27
Fix problems with searching form.io pulldowns — [#1783]

### LINKED ISSUE #1783: Searching in FormIO pulldowns (like the prefill KvK field selector) doesn't find all fields

1.1.3

Most likely a bug or feature in the select2/underlying library that FormIO uses but searching for "kvknummer" finds 1 instance, while if you scroll you can find 2. Finding "vestigingsnummer" doesn't yield any results, but its there.

Most likely caused by the fact that these matches are further up in the strings and the match/index doesn't go that far?

Searching gives no result:
image

Scrolling does:
image

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

dariomory:
Fixes #1783

Peek 2022-07-27 11-40

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "copy-to-clipboard": "^3.3.1",
         "core-js": "^3.11.1",
         "flatpickr": "^4.6.9",
-        "formiojs": "^4.13.13",
+        "formiojs": "~4.13.0",
         "immer": "^9.0.6",
         "json-logic-js": "^2.0.1",
         "leaflet": "^1.7.1",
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):
The version range caused formio to be upgraded to 4.14 instead of the latest patch release in 4.13.

Upgrading to newer minor versions is something that needs proper planning, we might do that for the 1.3.0 version and should also upgrade the SDK version then, which is currently stuck on 4.12.x with some backported patches.

PR #1820, opened 2022-07-28
[#1760] Variables creation validation

### LINKED ISSUE #1760: Variables creation validation
### Thema / Theme

Other

At the moment, when saving building a form, the default value of the variables is inferred from the component type in  the frontend:
https://github.com/open-formulieren/open-forms/blob/fix/1747-default-values-variables/src/openforms/js/components/admin/form_design/variables/utils.js#L47

- [x] It would be good to simplify the frontend and do this in the backend, since the form definition has already been saved once the form variables are saved
- [x] The backend should also do some validation that the key referred by the variables really exists in the form definition.

_No response_

_No response_

SilviaAmAm:
Fixes #1760

**Changes**
Backend validation to check that all component variables have a corresponding field in the configuration.

Note about `initial_value` of the variables in the tests:
The variables created in the backend during tests by factories should have realistic data because the variables are created by the FormStepFactory [here](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/tests/factories.py#L98) using `create_from_step` which sets the `initial_value` to `None` if no `defaultValue` has been specified in the configuration (while it uses the `defaultValue` if it has been set). This is the same behaviour of the frontend.

=> In the end the backend sets the initial_value and the data_type in the save() method


sergei-maertens (approved):
Couple cleanups, shouldn't be too bad.

I am re-thinking the first task though and the remark of having to look up things - in the serializer we need to look up the component anyway to validate it. Couldn't we override the `save` method of `FormVariable` and derive the default value from there? Something like:

```py
    def save(self, *args, **kwargs):
        self._derive_initial_value()
        super().save(*args, **kwargs)

    def _derive_initial_value(self):
        if self.source != FormVariableSources.compnent:
            return
        if not (fd := self.form_definition):
            return

        if self.initial_value:
            return

        component = get_component(fd.configuration, self.key)
        self.initial_value = component.get("defaultValue", None)
```

performance wise this shouldn't be a big problem, because the serializer sets `variable.form_definition`, so there's no lookup there. And we remain within the context of our `form_definition`, so there's no traversal of relations, right?

SilviaAmAm:
I updated it so that now the backend sets both the `initial_value` and the `data_type` for component variables.
I am not entirely sure what to do in the front end: if we remove the current code, you would only see the `initial_value` and `data_type` populated after saving the form :thinking:

src/openforms/formio/utils.py
```diff
@@ -2,6 +2,7 @@
 from datetime import date, datetime
 from typing import Any, Dict, Iterator, List
 
+from ..forms.constants import FormVariableDataTypes
```
sergei-maertens, line 5:
```suggestion
from openforms.forms.constants import FormVariableDataTypes
```

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -3,7 +3,10 @@
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 
-from openforms.forms.models import FormVariable
+from openforms.submissions.form_logic import get_component
```
sergei-maertens, line 6: probably better to move `get_component` to `openforms.formio.utils`

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1821, opened 2022-07-29
[#1762] Variables in simple logic editor

### LINKED ISSUE #1762: Add variables to the simple logic menus
### Thema / Theme

Form designer

Currently, one can only specify variables in the trigger of advanced rules.

It would be good to be able to specify them in the simple rules too.

SilviaAmAm:
Fixes #1762 

**Note**
I left the selection of the action "Change the value of a field/component" to use the components instead of the variables, since we will remove this action when the variable feature flag is removed. => https://github.com/open-formulieren/open-forms/issues/1822

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -4398,7 +4398,7 @@ components:
           title: JSON logic
           description: The trigger expression to determine if the actions should execute
             or not. Note that this must be a valid JsonLogic expression, and the first
-            operand must be a reference to a component in the form.
+            operand must be a reference to a variable in the form.
         order:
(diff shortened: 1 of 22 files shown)


sergei-maertens (approved):
Some non-code remarks:

1. Vertical alignment of date input is a bit janky: 
image

2. Checkbox (type boolean) setting the value presents a textarea rather than a checkbox/dropdown with yes/no options: 
image

SilviaAmAm:
Added an issue for the comments: https://github.com/open-formulieren/open-forms/issues/1845

PR #1825, opened 2022-08-01
[#1661] Set auth static data

### LINKED ISSUE #1661: Make sure the login attributes are set as variables
Make sure the DigiD / eHerkenning + machtigingen is stored in the variables in such a way that we can add logic (psuedo):

```
Als eherkenning_machtiginggever_kvknr is leeg, dan blokkeer formulier
```

SilviaAmAm:
Fixes #1661 

**Changes**: 
We now have an `auth_identifier` static variable, which contains all information:
```
{ "plugin": "digid", "attribute": "bsn", "value": "111222333", "machtigen": {}}
```
And then there are 3 more static vars (`auth_bsn`, `auth_kvk` and `auth_pseudo`) which only contain the value of the  `auth_identifier` and are `""` if no value is present for that particular attribute.

Every app can define their static variables through the register.

### DIFF
--- a/src/openforms/authentication/static_variables/static_variables.py
+++ b/src/openforms/authentication/static_variables/static_variables.py
@@ -1,0 +1,73 @@
+from typing import TYPE_CHECKING, Optional
+
+from django.utils.translation import gettext_lazy as _
+
+from openforms.authentication.constants import AuthAttribute
+from openforms.forms.constants import FormVariableDataTypes
+from openforms.variables.base import BaseStaticVariable
+from openforms.variables.registry import static_variables_register
+
+if TYPE_CHECKING:  # pragma: nocover
+    from openforms.authentication.utils import FormAuth
+    from openforms.submissions.models import Submission
+
+
+@static_variables_register("auth_identifier")
+class AuthIdentifier(BaseStaticVariable):
+    name = _("Authentication identifier")
+    data_type = FormVariableDataTypes.object
+
+    def get_initial_value(
+        self, submission: Optional["Submission"]
+    ) -> Optional["FormAuth"]:
+        if not submission or not hasattr(submission, "auth_info"):
+            return None
+
+        from openforms.authentication.utils import FormAuth
+
+        auth_data = FormAuth(
+            plugin=submission.auth_info.plugin,
+            attribute=submission.auth_info.attribute,
+            value=submission.auth_info.value,
+            # TODO what is the structure of the data in the machtigen field?
+            machtigen=submission.auth_info.machtigen,
+        )
+
+        return auth_data
+
+
+def get_auth_value(submission: Optional["Submission"], attribute: str) -> str:
+    if not submission or not hasattr(submission, "auth_info"):
+        return ""
+
+    if submission.auth_info.attribute == attribute:
+        return submission.auth_info.value
+
+    return ""
+
+
+@static_variables_register("auth_bsn")
+class AuthBSN(BaseStaticVariable):
+    name = _("Authentication BSN")
+    data_type = FormVariableDataTypes.string
+
+    def get_initial_value(self, submission: Optional["Submission"]) -> str:
+        return get_auth_value(submission, AuthAttribute.bsn)
+
+
+@static_variables_register("auth_kvk")
+class AuthKvK(BaseStaticVariable):
+    name = _("Authentication KvK")
+    data_type = FormVariableDataTypes.string
+
+    def get_initial_value(self, submission: Optional["Submission"]) -> str:
+        return get_auth_value(submission, AuthAttribute.kvk)
+
+
+@static_variables_register("auth_pseudo")
+class AuthPseudo(BaseStaticVariable):
+    name = _("Authentication pseudo")
+    data_type = FormVariableDataTypes.string
+
+    def get_initial_value(self, submission: Optional["Submission"]) -> str:
+        return get_auth_value(submission, AuthAttribute.pseudo)
--- a/src/openforms/authentication/static_variables/static_variables.py
+++ b/src/openforms/authentication/static_variables/static_variables.py
@@ -2,17 +2,19 @@ from typing import TYPE_CHECKING, Optional
 
 from django.utils.translation import gettext_lazy as _
 
-from openforms.authentication.constants import AuthAttribute
-from openforms.forms.constants import FormVariableDataTypes
 from openforms.variables.base import BaseStaticVariable
-from openforms.variables.registry import static_variables_register
+from openforms.variables.constants import FormVariableDataTypes
+from openforms.variables.registry import register_static_variable
+
+from ..constants import AuthAttribute
 
 if TYPE_CHECKING:  # pragma: nocover
-    from openforms.authentication.utils import FormAuth
     from openforms.submissions.models import Submission
 
+    from ..utils import FormAuth
 
-@static_variables_register("auth_identifier")
+
+@register_static_variable("auth_identifier")
 class AuthIdentifier(BaseStaticVariable):
     name = _("Authentication identifier")
     data_type = FormVariableDataTypes.object
@@ -20,10 +22,10 @@ class AuthIdentifier(BaseStaticVariable):
     def get_initial_value(
         self, submission: Optional["Submission"]
     ) -> Optional["FormAuth"]:
-        if not submission or not hasattr(submission, "auth_info"):
+        if not submission or not submission.is_authenticated:
             return None
 
-        from openforms.authentication.utils import FormAuth
+        from ..utils import FormAuth
 
         auth_data = FormAuth(
             plugin=submission.auth_info.plugin,
@@ -37,7 +39,7 @@ class AuthIdentifier(BaseStaticVariable):
 
 
 def get_auth_value(submission: Optional["Submission"], attribute: str) -> str:
-    if not submission or not hasattr(submission, "auth_info"):
+    if not submission or not submission.is_authenticated:
         return ""
 
     if submission.auth_info.attribute == attribute:
@@ -46,28 +48,31 @@ def get_auth_value(submission: Optional["Submission"], attribute: str) -> str:
     return ""
 
 
-@static_variables_register("auth_bsn")
+@register_static_variable("auth_bsn")
 class AuthBSN(BaseStaticVariable):
     name = _("Authentication BSN")
     data_type = FormVariableDataTypes.string
 
-    def get_initial_value(self, submission: Optional["Submission"]) -> str:
+    @staticmethod
+    def get_initial_value(submission: Optional["Submission"]) -> str:
         return get_auth_value(submission, AuthAttribute.bsn)
 
 
-@static_variables_register("auth_kvk")
+@register_static_variable("auth_kvk")
 class AuthKvK(BaseStaticVariable):
     name = _("Authentication KvK")
     data_type = FormVariableDataTypes.string
 
-    def get_initial_value(self, submission: Optional["Submission"]) -> str:
+    @staticmethod
+    def get_initial_value(submission: Optional["Submission"]) -> str:
         return get_auth_value(submission, AuthAttribute.kvk)
 
 
-@static_variables_register("auth_pseudo")
+@register_static_variable("auth_pseudo")
 class AuthPseudo(BaseStaticVariable):
     name = _("Authentication pseudo")
     data_type = FormVariableDataTypes.string
 
-    def get_initial_value(self, submission: Optional["Submission"]) -> str:
+    @staticmethod
+    def get_initial_value(submission: Optional["Submission"]) -> str:
         return get_auth_value(submission, AuthAttribute.pseudo)
--- a/src/openforms/authentication/tests/test_static_variables.py
+++ b/src/openforms/authentication/tests/test_static_variables.py
@@ -17,6 +17,13 @@ class TestStaticVariables(TestCase):
         static_data = FormVariable.get_static_data(auth_info.submission)
 
         self.assertEqual(5, len(static_data))
+        (
+            form_auth_var,
+            auth_identifier_var,
+            auth_bsn_var,
+            auth_kvk_var,
+            auth_pseudo_var,
+        ) = static_data
         self.assertEqual(
             {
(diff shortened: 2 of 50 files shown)


src/openforms/forms/models/form_variable.py
```diff
@@ -27,14 +28,27 @@
...
     return timezone.now().isoformat()
 
 
-STATIC_INITIAL_VALUES = {FormVariableStaticInitialValues.now: get_now}
+def get_auth_identifier(request: Optional["Request"]) -> Optional[str]:
+    if not request or not hasattr(request, "session"):
+        return None
+
+    # TODO what is the structure of the data in the machtigen field?
+    return request.session.get(FORM_AUTH_SESSION_KEY)
```
sergei-maertens, line 45: I think we should aim to keep django apps decoupled and avoid using the request/session as a persistent data structure:

1. forms app ideally shouldn't have to be aware of `authentication` app internals
2. the request session data is not guaranteed to always be available - in fact, after authenticating, the `authentication` apps is responsible for persisting the auth details asap into the `Submission` model (see `signals.py` there)
3. session data is just a mechanism to persist some information during redirect chains and ideally would be wiped once it's persisted
4. session data is a bit tricky perhaps with the form suspend/resume - it does get set if you re-auth, but it all feels a bit fragile

A submission is essentially a "form session", with the added benefit that it acts as a namespace (you can have different details for different submissions all at the same time). I think it's better to pass around the `submission` instance instead of the request/session and extract the data from there.

On that note, `submissions/authentication` depends on `forms` app, and we did a lot of work to remove `authentication` app imports from submissions/forms - I think it's a bit of a shame that we then re-introduce this tight coupling.

Maybe there should be a method/utility in submissions app `Submission.get_auth_identifier_variable`? Maybe we should have a per-app discovery of variables or a registry for them so that every app can declare their own variables and implementations?

sergei-maertens (changes requested):

SilviaAmAm:
After discussion: 
We will do the refactor of the submission auth attributes first and then rework this PR.

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -0,0 +1,73 @@
+from typing import TYPE_CHECKING, Optional
+
+from django.utils.translation import gettext_lazy as _
+
+from openforms.authentication.constants import AuthAttribute
```
sergei-maertens, line 5: this can (should?) be a relative import

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -0,0 +1,73 @@
...
+from openforms.forms.constants import FormVariableDataTypes
+from openforms.variables.base import BaseStaticVariable
+from openforms.variables.registry import static_variables_register
+
+if TYPE_CHECKING:  # pragma: nocover
+    from openforms.authentication.utils import FormAuth
```
sergei-maertens, line 11: this can (should?) be a relative import

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -0,0 +1,73 @@
...
+
+@static_variables_register("auth_identifier")
+class AuthIdentifier(BaseStaticVariable):
+    name = _("Authentication identifier")
+    data_type = FormVariableDataTypes.object
+
+    def get_initial_value(
+        self, submission: Optional["Submission"]
+    ) -> Optional["FormAuth"]:
+        if not submission or not hasattr(submission, "auth_info"):
```
sergei-maertens, line 23: can we use `submission.is_authenticated` instead of the `hasattr` call?

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -0,0 +1,73 @@
...
+            return None
+
+        from openforms.authentication.utils import FormAuth
```
sergei-maertens, line 26: this can (should?) be a relative import

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -0,0 +1,73 @@
...
+            value=submission.auth_info.value,
+            # TODO what is the structure of the data in the machtigen field?
+            machtigen=submission.auth_info.machtigen,
+        )
+
+        return auth_data
+
+
+def get_auth_value(submission: Optional["Submission"], attribute: str) -> str:
+    if not submission or not hasattr(submission, "auth_info"):
```
sergei-maertens, line 40: `submission.is_authenticated` here as well?

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -0,0 +1,73 @@
...
+
+    return ""
+
+
+@static_variables_register("auth_bsn")
+class AuthBSN(BaseStaticVariable):
+    name = _("Authentication BSN")
+    data_type = FormVariableDataTypes.string
+
+    def get_initial_value(self, submission: Optional["Submission"]) -> str:
```
sergei-maertens, line 54: can be a `@staticmethod` as the `self` is not used

src/openforms/authentication/tests/test_static_variables.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self.assertEqual(5, len(static_data))
+        self.assertEqual(
+            {
+                "plugin": "test-plugin",
+                "attribute": AuthAttribute.bsn,
+                "value": "111222333",
+                "machtigen": {AuthAttribute.bsn: "123456789"},
+            },
+            static_data[1].initial_value,
```
sergei-maertens, line 27: I'd get rid of the magic numbers here by doing some destructuring:

```py
form_auth_var, auth_bsn_var, auth_kvk_var, auth_pseudo_var = static_data
```
and then do the assertions on those variables.

src/openforms/formio/service.py
```diff
@@ -26,21 +26,17 @@ def get_dynamic_configuration(
     configuration = handle_custom_types(
         configuration, request=request, submission=submission
     )
-    configuration = insert_variables(
-        configuration, submission=submission, request=request
-    )
+    configuration = insert_variables(configuration, submission=submission)
```
sergei-maertens, line 29: self-reminder: this conflicts with the changes in #1869, rebase that feature branch after this PR is merged

src/openforms/forms/constants.py
```diff
@@ -55,3 +55,7 @@ class FormVariableDataTypes(DjangoChoices):
 
 class FormVariableStaticInitialValues(DjangoChoices):
     now = ChoiceItem("now", _("Now"))
+    auth_identifier = ChoiceItem("auth_identifier", "Authentication identifier")
+    auth_bsn = ChoiceItem("auth_bsn", "Authentication BSN")
+    auth_kvk = ChoiceItem("auth_kvk", "Authentication KvK")
+    auth_pseudo = ChoiceItem("auth_pseudo", "Authentication pseudo")
```
sergei-maertens, line 61: do we need this? is this going to be annoying to extend if other apps define their own static variables? It feels like this is superseded by the variables registry

src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
```diff
@@ -119,6 +120,49 @@ def test_static_variable_in_trigger(self, m_get_static_data):
...
+                key="auth_identifier",
+                data_type=FormVariableDataTypes.object,
+                initial_value={
+                    "plugin": "digid",
+                    "attribute": AuthAttribute.bsn,
+                    "value": "123456782",
+                },
+            )
+        ],
+    )
```
sergei-maertens, line 137: can we replace this mocking structure with replacing the variable registry instance? that decouples it from the actual implementation and makes our tests a bit more robust

src/openforms/submissions/form_logic.py
```diff
@@ -95,9 +95,7 @@ def evaluate_form_logic(
 
     step_index = submission_state.form_steps.index(step.form_step)
 
-    data_container = DataContainer(
-        state=submission_variables_state, request=context.get("request")
-    )
+    data_container = DataContainer(state=submission_variables_state)
```
sergei-maertens, line 98: self-reminder - probably also conflicts with #1869

src/openforms/submissions/tests/test_variables/test_num_queries.py
```diff
@@ -118,17 +119,18 @@ def test_evaluate_form_logic_with_rules(self):
...
-        # 5.  Retrieve the submission variables to be deleted
-        # 6.  Retrieve the submission attachment files to be deleted
-        # 7.  SAVEPOINT
-        # 8.  Delete submission attachment files
-        # 9.  RELEASE SAVEPOINT
-        # 10. Delete submission values
-        # 11. Retrieve all form_variables
-        # 12. Creation of timelinelog
-        with self.assertNumQueries(12):
+        # 3.  Retrieve auth info for the submission
```
sergei-maertens, line 122: I *think* we should be able to `select_related` this and avoid this extra query

src/openforms/variables/base.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass
+
+from openforms.forms.models import FormVariable
+
+
+@dataclass
+class BaseStaticVariable:
+    identifier: str
+
+    def get_initial_value(self, *args, **kwargs):
```
sergei-maertens, line 10: you could use `@abc.ABC` and `@abc.abstractmethod` here, and you can add the `# pragma: nocover` comment since measuring coverage for this isn't relevant.

src/openforms/variables/base.py
```diff
@@ -0,0 +1,19 @@
...
+        raise NotImplementedError()
+
+    def get_static_variable(self, *args, **kwargs):
+        return FormVariable(
+            name=self.name,
```
sergei-maertens, line 15: shouldn't we define `name` and `data_type` as a class property too? You can add them as `dataclasses.field(init=False)` so they are type hinted but not required for the `__init__` method.

src/openforms/variables/registry.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    A registry for static variables.
+    """
+
+    module = "variables"
+
+
+# Sentinel to provide the default registry. You can easily instantiate another
+# :class:`Registry` object to use as dependency injection in tests.
+static_variables_register = Registry()
```
sergei-maertens, line 14: I'd rename this to `register_static_variable` so it reads a bit more natural when used as a decorator.

In the `FormVariable.get_static_data` module you can alias the import to `static_variables_register` so that also still reads like proper English :)

src/openforms/variables/static_variables/static_variables.py
```diff
@@ -0,0 +1,16 @@
...
+from ..registry import static_variables_register
+
+
+@static_variables_register("now")
+class Now(BaseStaticVariable):
+    name = _("Now")
+    data_type = FormVariableDataTypes.datetime
+
+    def get_initial_value(self, *args, **kwargs):
+        return timezone.now().isoformat()
```
sergei-maertens, line 16: while implementing #1708, I noticed that I want access to the raw underlying datetime object and not the stringified version (so I can use the `date` and `time` template filters), so I'd drop the `.isoformat()`.

```suggestion
        return timezone.now()
```

src/openforms/variables/static_variables/static_variables.py
```diff
@@ -0,0 +1,16 @@
+from django.utils import timezone
+from django.utils.translation import gettext_lazy as _
+
+from openforms.forms.constants import FormVariableDataTypes
```
sergei-maertens, line 4: I think it makes more sense to move these variable-related constants to `openforms.variables`.

At this point, I think our app organization & dependencies is more looking like:

1. `openforms.forms` & `openforms.variables` on the same level
2. `openforms.submissions` makes use of forms & variables

All three can be considered "core" Open Forms. This is probably relevant for the documentation PR as well.

sergei-maertens (changes requested):

src/openforms/forms/constants.py
SilviaAmAm: Yes, it doesn't look to be used anywhere anymore

src/openforms/submissions/tests/renderer/test_renderer.py
```diff
@@ -123,6 +123,7 @@ def test_performance_num_queries(self):
         # 4. Get the step-specific data from submission variable values (TODO: this can probably be optimized away?)
         # 5. Load submission state: get form steps
         # 6. Load submission state: get submission steps
-        # 7. Query the form logic rules for the submission form (and this is cached)
-        with self.assertNumQueries(7):
+        # 7. Retrieve the submission auth info
```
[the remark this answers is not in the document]
SilviaAmAm: To avoid this query, we need the auth_info to already be on the submission object that is passed to the renderer.

src/openforms/submissions/tests/test_variables/test_num_queries.py
```diff
@@ -57,9 +57,10 @@ def test_evaluate_form_logic_without_rules(self):
         # 1. get_dynamic_configuration: injects variables in configuration and *currently* does 1 query to retrieve
         #    variables with prefill data, so the defaultValue on the components can be set.
         # 2. Retrieve all logic rules related to a form
-        # 3. Load submission state: Retrieve formsteps,
-        # 4. Load submission state: Retrieve submission steps
-        with self.assertNumQueries(4):
+        # 3. Retrieve auth info for the submission
```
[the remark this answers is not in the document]
SilviaAmAm: For this, we need the submission object passed to `evaluate_form_logic` to already have the auth_info. I added the `auth_info` prefetch on the `get_object` of the endpoint (src/openforms/submissions/api/viewsets.py#L361).

But since this only tests `evaluate_form_logic` Im not sure if I should remove the query from the assertions?

src/openforms/submissions/tests/renderer/test_renderer.py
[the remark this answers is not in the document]
SilviaAmAm: Discussed: should find the places where the renderer is called and make sure the auth is on the submission

sergei-maertens (approved):

PR #1826, opened 2022-08-01
Hidden fields present (bug) 

### LINKED ISSUE #1755: Hidden fields with default value present in data

1.1, latest, probably also 1.0

Hidden fields with a default value that are made visible by logic and then hidden again are not removed from the data.

**Note**
This is also a problem with the variables

In the admin
1. Create a new form
2. Add a checkbox 
3. Add a text field with hidden checked and 'Test data' as default value
4. Add an advanced logic rule with trigger `{"==":[{"var": "checkbox"}, true]}` and action `change a property `of the `textField`, property `hidden` value `false`.
5. Save the form

In the SDK:
1. Check the box
2. Uncheck the box
3. Submit the form

In the Admin, check the submission. The data contains the field `textField: 'test data'`

The hidden field should be removed from the data

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

Relates to #1527  and the patch for that (#1528)

dariomory:
Fixes #1755


sergei-maertens:
Where is the regression test?

src/openforms/submissions/tests/form_logic/test_modify_components.py
```diff
@@ -172,6 +172,88 @@ def test_hiding_component_empties_its_data(self):
         self.assertEqual(configuration, expected)
         self.assertEqual("", submission_step.data["component2"])
 
+    def test_hiding_component_with_default_value_empties_its_data(self):
```
sergei-maertens, line 175: This is not a good regression test. We need a regression test that replicates what happens in the "how to reproduce" part of the ticket.

Additionally, regression tests should always have a reference back to the ticket, the preferred way is to use the test tag: `@tag("gh-1755")` so we can easily run and isolate the test and see why this test exists.

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
```diff
@@ -182,27 +184,26 @@ def evaluate_form_logic(
         # debounced client-side data changes
         data_diff = {}
 
-        for key, new_value in step.data.items():
-            original_value = initial_data.get(key)
+        for key in map(lambda x: x["key"], configuration["components"]):
```
sergei-maertens, line 187: This does not handle nested components.

src/openforms/submissions/form_logic.py
```diff
@@ -182,27 +184,26 @@ def evaluate_form_logic(
...
                     continue
 
-            if new_value == original_value:
+            if new_value is empty or new_value == original_value:
                 continue
             data_diff[key] = new_value
 
         # only return the 'overrides'
         step.data = DirtyData(data_diff)
-
```
sergei-maertens, line 205: this line is here for a visual separation with nasty instance-state mutations, please put it back.

sergei-maertens (changes requested):
Solid regression test :+1:

src/openforms/submissions/form_logic.py
```diff
@@ -182,21 +184,22 @@ def evaluate_form_logic(
         # debounced client-side data changes
         data_diff = {}
 
-        for key, new_value in step.data.items():
-            original_value = initial_data.get(key)
+        components = iter_components(configuration)
+        for key in map(lambda x: x["key"], components):
```
sergei-maertens, line 188:
```suggestion
        for (key := component["key"]) in iter_components(configuration):
```

I don't understand why we can't loop over `step.data.items()` though. As far as I can see, we get the merged object of (saved data + dirty data), where dirty data overwrites the saved data if keys collide.

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
dariomory: From my observations `step.data` often has missing keys, and here we have to iterate over them.

dariomory: Anyhow I'd like to discuss this function in more detail. It is doing too much at once and will be a primary source for bugs. The logic of mutation operations needs a revision.

sergei-maertens: > From my observations step.data often has missing keys, and here we have to iterate over them.

This should be made clear by adding comments to the code.

> Anyhow I'd like to discuss this function in more detail. It is doing too much at once and will be a primary source for bugs. The logic of mutation operations needs a revision.

This function has been refactored only a week ago with more separation of concerns. "it will be a primary source of bugs"  -> you know we write tests to prevent those, right? Those are unfounded claims.

There are business requirements and technical requirements from the libraries we use. This function does a lot, but it is well documented why things are the way they are and it is well-tested.

dariomory: Good to know it is already being refactored :)

I agree writing tests is an excellent to prevent bugs, note that it is unrelated to my point about following simple principles of clean architecture.

src/openforms/submissions/tests/form_logic/test_submission_logic.py
```diff
@@ -150,6 +152,92 @@ def test_check_logic_with_full_datetime(self):
...
+
+        logic_check_endpoint = reverse(
+            "api:submission-steps-logic-check",
+            kwargs={"submission_uuid": submission.uuid, "step_uuid": form_step.uuid},
+        )
+
+        response_json = self.client.post(
+            logic_check_endpoint, data={"data": {"checkbox": True}}
+        ).json()
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
```
sergei-maertens, line 219: this refers to the original response of the GET detail endpoint. The code however makes it look like you meant to get the status code of the logic check call?

src/openforms/submissions/tests/form_logic/test_submission_logic.py
```diff
@@ -150,6 +152,92 @@ def test_check_logic_with_full_datetime(self):
...
+            response_json["step"]["formStep"]["configuration"]["components"][0][
+                "hidden"
+            ],
+        )
+
+        response_json = self.client.post(
+            logic_check_endpoint,
+            data={"data": {"checkbox": False, "textfield": "Test data"}},
+        ).json()
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
```
sergei-maertens, line 230: this refers to the original response of the GET detail endpoint. The code however makes it look like you meant to get the status code of the logic check call?

src/openforms/submissions/form_logic.py
```diff
@@ -186,21 +188,22 @@ def evaluate_form_logic(
         # debounced client-side data changes
         data_diff = {}
 
-        for key, new_value in step.data.items():
-            original_value = initial_data.get(key)
+        for component in iter_components(configuration):
```
sergei-maertens, line 191: I requested an inline comment to be added why we iterate over all the components rather than over `step.data`.

src/openforms/submissions/form_logic.py
```diff
@@ -186,21 +188,22 @@ def evaluate_form_logic(
...
+            key = component["key"]
+            new_value = updated_step_data.get(key, empty)
+            original_value = initial_data.get(key, empty)
             # Reset the value of any field that may have become hidden again after evaluating the logic
-            if original_value:
+            if original_value is not empty:
                 component = get_component(configuration, key)
```
sergei-maertens, line 197: this line is not required anymore, we already have the `component` from the for loop

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
joeribekker: @dariomory 

```suggestion
        # Iterate over all components instead of `step.data`, to take hidden fields into account (See: #1755)
        for component in iter_components(configuration):
```

joeribekker (changes requested):

sergei-maertens (approved):

PR #1827, opened 2022-08-02
[#1798] Delete submissions forms from admin

### LINKED ISSUE #1798: Forms with submissions can't be deleted

latest

DH Taiga 150

It is not possible to delete a form (which has related submissions) from the admin. The submissions are related to logs, which can't be deleted (even by a superuser).

- Create a new form in the admin, with one step and one field.
- Save the form.
- Fill in the form in the frontend and submit it.
- Now back in the admin try to remove the form. It will say that the form can't be deleted because there is a related submission
- Go to the submission and try to delete it. It will say that it is not possible to delete it because there is a log associated with it.

It should be possible to delete the submissions.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1798 

I ended up adding the possibility to remove logs from the admin for the users with the right permissions. 

The removal of the objects through the admin is logged in the `LogEntry` model ([here](https://github.com/django/django/blob/main/django/contrib/admin/options.py#L948) ) by default.

### DIFF
--- a/src/openforms/logging/tests/test_admin.py
+++ b/src/openforms/logging/tests/test_admin.py
@@ -1,14 +1,17 @@
 from django.contrib.auth.models import Permission
-from django.test import TestCase
 from django.urls import reverse
+from django.utils import timezone
 
-from openforms.accounts.tests.factories import StaffUserFactory
+from django_webtest import WebTest
+
+from openforms.accounts.tests.factories import StaffUserFactory, SuperUserFactory
 from openforms.logging import logevent
+from openforms.logging.models import TimelineLogProxy
 from openforms.logging.tests.factories import TimelineLogProxyFactory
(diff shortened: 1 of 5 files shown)


sergei-maertens (commented):
I'd like to discuss this with Joeri next week - I think we intentionally made it so that even superusers cannot remove log records, because all their actions also have to be audit-logged. Now superusers can view personal data and remove all traces of them viewing that data.

sergei-maertens (changes requested):
Discussed - audit logs may not be deleted, not even by superusers.

Can you please:

* Add a test to ensure even superusers cannot delete audit logs
* remove the `GenericRelation` field and instead where it's used replace that with direct `TimelineLogProxy` queries?

sergei-maertens (approved):

PR #1830, opened 2022-08-03
Feature/1771 deleting not reusable form definition

### LINKED ISSUE #1771: When a form step is deleted that is not re-usable, the form definition still exists
I think we should cascade delete non-reusable form definitions when a form step is being deleted - they can't be selected anymore anyway and only pollute the unique slug namespace.

Ewen-Lgy:
Fixes #1771

### DIFF
--- a/src/openforms/forms/models/form_step.py
+++ b/src/openforms/forms/models/form_step.py
@@ -79,5 +79,11 @@ class FormStep(OrderedModel):
         else:
             return super().__str__()
 
+    def delete(self, using=None, keep_parents=False):
+        super().delete(using=using, keep_parents=keep_parents)
+
+        if not self.form_definition.is_reusable:
+            self.form_definition.delete()
+
     def iter_components(self, recursive=True, **kwargs):
         yield from self.form_definition.iter_components(recursive=recursive, **kwargs)
--- a/src/openforms/forms/models/form_step.py
+++ b/src/openforms/forms/models/form_step.py
@@ -79,8 +79,8 @@ class FormStep(OrderedModel):
         else:
             return super().__str__()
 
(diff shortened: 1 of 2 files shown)


src/openforms/forms/models/form_step.py
```diff
@@ -79,5 +79,11 @@ def __str__(self):
         else:
             return super().__str__()
 
+    def delete(self, using=None, keep_parents=False):
+        super().delete(using=using, keep_parents=keep_parents)
+
+        if self.form_definition.pk is not None and not self.form_definition.is_reusable:
```
SilviaAmAm, line 85: Why the check on `self.form_definition.pk`?

SilviaAmAm (changes requested):
Just a question/comment :slightly_smiling_face: 

I noticed that removing the non-reusable form definitions was already the behaviour for the admin actions (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/admin/form.py#L368) so no need to do anything about those.

src/openforms/forms/models/form_step.py
Ewen-Lgy: I noticed that in a test(https://github.com/open-formulieren/open-forms/blob/0d58b3a333cfb217f4da0e824aab2257c272f1f1/src/openforms/forms/tests/test_api_import_export.py#L159), we were deleting a formdefinition before the formstep itself. Without this verification, it would raises an error since the form_definition doesn't exist anymore. Nonetheless, I suspect that it wouldn't be possible to replicate this behavior in a real scenario

sergei-maertens (approved):

SilviaAmAm (approved):

PR #1843, opened 2022-08-05
Feature/1833 form with empty prefill data cannot be submitted

### LINKED ISSUE #1833: Form with empty prefill data cannot be submitted

1.0.12

Gemeente Venray Taiga 26

The form contains fields that with prefill (Stuf-BG). When logging in with a fake BSN no data is prefilled in the fields. In the admin logs it says that the data retrieved is `{"filed": null, ...}`. 
The prefilled fields are read only, so they can't be edited by the user. 

When submitting the form, the backend returns a 400, with the following error:

```
{"incompleteSteps":[],"submissionAllowed":"yes","invalidPrefilledFields":["Voornaam","Tussenvoegsel","Achternaam","Straatnaam","Huisnummer","Huisletter","Huisnummertoevoeging","Postcode","Woonplaats"]}
```

Form bezwaar-gemeentelijke-belastingen on the venray test environment. 

The form should be submitted.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

Ewen-Lgy:
Fixes #1833

### DIFF
--- a/src/openforms/submissions/api/validation.py
+++ b/src/openforms/submissions/api/validation.py
@@ -107,6 +107,8 @@ def get_invalid_prefilled_fields(
         attribute_name = component["prefill"]["attribute"]
 
         if submission.prefill_data[plugin_name][attribute_name] is None:
+            # the value will be `None` if there is no actual prefill data available, so there is nothing to compare to. This
+            # especially applies to test-environments without real prefill-connections.
             continue
 
         if (
--- a/src/openforms/submissions/api/validation.py
+++ b/src/openforms/submissions/api/validation.py
@@ -107,11 +107,12 @@ def get_invalid_prefilled_fields(
         attribute_name = component["prefill"]["attribute"]
 
         if submission.prefill_data[plugin_name][attribute_name] is None:
-            initial_value = component["defaultValue"]
-        else:
-            initial_value = submission.prefill_data[plugin_name][attribute_name]
+            continue
 
-        if submission_step.data[component["key"]] != initial_value:
+        if (
+            submission_step.data[component["key"]]
+            != submission.prefill_data[plugin_name][attribute_name]
+        ):
             invalid_prefilled_fields.append(component["label"])
 
     return invalid_prefilled_fields
--- a/src/openforms/submissions/api/validation.py
+++ b/src/openforms/submissions/api/validation.py
@@ -105,10 +105,13 @@ def get_invalid_prefilled_fields(
 
         plugin_name = component["prefill"]["plugin"]
         attribute_name = component["prefill"]["attribute"]
-        if (
-            submission_step.data[component["key"]]
-            != submission.prefill_data[plugin_name][attribute_name]
-        ):
+
+        if submission.prefill_data[plugin_name][attribute_name] is None:
+            initial_value = component["defaultValue"]
+        else:
+            initial_value = submission.prefill_data[plugin_name][attribute_name]
+
+        if submission_step.data[component["key"]] != initial_value:
             invalid_prefilled_fields.append(component["label"])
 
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/api/validation.py
```diff
@@ -105,10 +105,13 @@ def get_invalid_prefilled_fields(
...
-            submission_step.data[component["key"]]
-            != submission.prefill_data[plugin_name][attribute_name]
-        ):
+
+        if submission.prefill_data[plugin_name][attribute_name] is None:
+            initial_value = component["defaultValue"]
+        else:
+            initial_value = submission.prefill_data[plugin_name][attribute_name]
+
+        if submission_step.data[component["key"]] != initial_value:
```
sergei-maertens, line 114:
```suggestion

        if submission.prefill_data[plugin_name][attribute_name] is None:
            # the value will be `None` if there is no actual prefill data available, so there is nothing to compare to. This
            # especially applies to test-environments without real prefill-connections.
            continue
```
perhaps? though I do realize this leaves open the option to spoof data

sergei-maertens (changes requested):
I'm not sure if this is the right approach. We now cannot tell the difference from `null` as an actual value vs. `null` because the prefill data is missing.

The `defaultValue` construct is also a bit wonky, since usually prefill results in `defaultValue` being set.

This check is supposed to explicitly check for people spoofing data that is only supposed to be entered via pre-fill through "read-only" fields. Perhaps we should just check if the prefill data is `None` and then `continue`, and leave that edge case for testing environments etc.?

src/openforms/submissions/api/validation.py
```diff
@@ -105,10 +105,13 @@ def get_invalid_prefilled_fields(
 
         plugin_name = component["prefill"]["plugin"]
         attribute_name = component["prefill"]["attribute"]
-        if (
-            submission_step.data[component["key"]]
-            != submission.prefill_data[plugin_name][attribute_name]
-        ):
+
+        if submission.prefill_data[plugin_name][attribute_name] is None:
+            initial_value = component["defaultValue"]
```
SilviaAmAm, line 110: I think here we need to keep the old code path if `submission.prefill_data[plugin_name][attribute_name]` is not `None`, but if it is `None` we could check that `submission_step.data[component["key"]] == component["defaultValue"]` (and if they are different add the component label to the `invalid_prefilled_fields`).

On the other hand, I was thinking that there may be issues with the default values (Formio does some funny things with default values), so I'm still not 100% sure we should do this extra check :thinking:

SilviaAmAm (changes requested):

src/openforms/submissions/api/validation.py
SilviaAmAm: Oups, didn't see Sergei had already reviewed! :sweat_smile:

Ewen-Lgy: Yes ahah :smile: If defaultValue cannot be trusted, I think we should go for the Sergei's solution and keeping things simple for now then

SilviaAmAm (approved):

sergei-maertens (approved):

sergei-maertens:
Backports:

* `stable/1.0.x`: a501af8c51aca0986eaaaaaca3ec03ddd01e0622, 77efd1939c13dfb82c80ca92a76d47957414fc9b and 845b9b5f1b0a240889a8b10d5d178b69bce5cf52
* `stable/1.1.x`: 72f98f75e9349d6d8050fb7712a902324f917e52, 8cbd3d8f807df3b6896279d5f91f92c618c5074c and 63184b2f7553a5fe784f2d02c6b71386d4668eb1

PR #1848, opened 2022-08-09
Feature/1294 analytics tools should play nice with csp headers

### LINKED ISSUE #1294: Analytics tools should manage CSP and Cookie settings automatically

**Tasks**

Ewen-Lgy:
Fixes #1294 

**Summary:**
- New app created: `Analytics_tools`
- Moved analytics tools and csp management into that new app
- Added boolean field to dynamically enable or disable analytics tool
- Related analytics tool Cookies and csp are created and deleted automatically
- Cookies and CSP entries are stored in json file
- Logging of enabling and disabling these analytics tools


src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,312 @@
...
+            for cookie in cookies:
+                Cookie.objects.get_or_create(
+                    name=cookie["name"],
+                    path=cookie["path"],
+                    domain=settings.ALLOWED_HOSTS[0],
+                    cookiegroup=cookie_group,
+                )
+        else:
+            for cookie in cookies:
+                Cookie.objects.filter(name=cookie["name"]).delete()
```
sergei-maertens, line 181:
```suggestion
            Cookie.objects.filter(name__in=[cookie["name"] for cookie in cookies]).delete()
```

src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def update_csp(self, csps: dict, create: bool):
+        if create:
+            for csp in csps:
+                CSPSetting.objects.get_or_create(
+                    directive=csp["directive"], value=csp["value"]
+                )
+        else:
+            for csp in csps:
+                CSPSetting.objects.filter(value=csp["value"]).delete()
```
sergei-maertens, line 191: this can also be a single `.filter(...).delete()` call instead of multiple queries

src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,312 @@
...
+            return super().save(*args, **kwargs)
+        original_object = self.__class__.objects.get(pk=self.pk)
+        # For each analytics provider, we check if :
+        # - the value of the provider's boolean field is changing
+        # - one of the fields required is not filled out while the providers's boolean field one is --> we need to reset the boolean field to False
+        if original_object.enable_google_analytics != self.enable_google_analytics or (
+            not self.google_analytics_can_be_enabled() and self.enable_google_analytics
+        ):
+            if not self.google_analytics_can_be_enabled():
+                self.enable_google_analytics = False
```
sergei-maertens, line 205: this makes it possible for the end user to check a checkbox with incomplete configuration, and then the checkbox will be silently unchecked, right?

I think it would be better to implement the validation in the `def clean(...)` method and raise a `django.core.exceptions.ValidationError`, something along the lines "If you enable $provider, you must fill out all the required fields.". That provides explicit user feedback, and then in this `save` method, you only have to check the value of the `is_<provider>_enabled` and/or the boolean flag.

src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,312 @@
...
+        if create:
+            for csp in csps:
+                CSPSetting.objects.get_or_create(
+                    directive=csp["directive"], value=csp["value"]
+                )
+        else:
+            for csp in csps:
+                CSPSetting.objects.filter(value=csp["value"]).delete()
+
+    def save(self, *args, **kwargs):
```
sergei-maertens, line 193: django has a preferred order for methods within a model: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#model-style

src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,312 @@
...
+                AnalyticsTools.siteimprove, self.siteimprove_can_be_enabled()
+            )
+        super().save(*args, **kwargs)
+
+    def update_analytics_tool(
+        self,
+        analytics_tool: str,
+        is_activated: bool,
+        site_id: Optional[str] = None,
+        site_url: Optional[str] = None,
```
sergei-maertens, line 271: these are quite tool-specific arguments in a generic method. Maybe it would be better suited to pass in a list of string replacements? e.g. `[("SITE_ID", site_id), ("SITE_URL", site_url)]`

src/openforms/analytics_tools/tests/test_csp_update.py
```diff
@@ -15,10 +16,33 @@ def parse_csp_policy(header_value):
...
-@override_settings(
-    CSP_REPORT_ONLY=False,
-)
+@override_settings(CSP_REPORT_ONLY=False, SOLO_CACHE=None)
 class CSPUpdateTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        config = AnalyticsToolsConfiguration.get_solo()
```
sergei-maertens, line 25: if possible, we prefer mocking django-solo models so that you don't have to deal with cache and cross-contamination in tests

src/openforms/analytics_tools/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+    def setUp(self):
+        super().setUp()
+        self.config = AnalyticsToolsConfiguration.get_solo()
+        self.config.analytics_cookie_consent_group = CookieGroup.objects.get(
+            varname="analytical"
+        )
+        self.config.gtm_code = "GTM-XXXX"
+        self.config.ga_code = "UA-XXXXX-Y"
+        self.config.save()
+        self.addCleanup(lambda: AnalyticsToolsConfiguration.get_solo().delete())
```
sergei-maertens, line 29: see earlier remark about mocking

src/openforms/analytics_tools/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_analytics_tool_properly_enabled(self):
+
+        self.config.enable_google_analytics = True
+        self.config.save()
+
+        for cookie in self.json_cookies:
+            self.assertIsNotNone(Cookie.objects.get(name=cookie["name"]))
```
sergei-maertens, line 37: 1. use `with self.subTest` when doing loops in tests
2. this will never return None, but instead raise `Cookie.DoesNotExist` if the cookie wasn't created.

src/openforms/analytics_tools/tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+        self.config.enable_google_analytics = True
+        self.config.save()
+
+        # Deletion of cookies
+        self.config.enable_google_analytics = False
+        self.config.save()
+
+        for cookie in self.json_cookies:
+            with self.assertRaises(Cookie.DoesNotExist):
+                Cookie.objects.get(name=cookie["name"])
```
sergei-maertens, line 53: maybe cleaner is to use `self.assertFalse(Cookie.objects.filter(name=cookie["name"]).exists())`

src/openforms/analytics_tools/utils.py
```diff
@@ -0,0 +1,42 @@
...
+        cookies = json.load(f)
+        if site_id is None:
+            return cookies
+        for cookie in cookies:
+            cookie["name"] = cookie["name"].replace(SITE_ID, str(site_id))
+            cookie["name"] = cookie["name"].replace(
+                DOMAIN_HASH,
+                get_domain_hash(settings.ALLOWED_HOSTS[0], cookie["path"]),
+            )
+        return cookies
```
sergei-maertens, line 25: working with a list of `token -> replacements` (or a dict, doesn't really matter) would be a bit more generic and clean, and you can still reuse piwik/piwik pro/matomo through a utility function.

src/openforms/config/migrations/0031_auto_20220718_1544.py
```diff
@@ -0,0 +1,61 @@
...
+
+    dependencies = [
+        ("config", "0030_piwik_pro_support"),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name="CSPSetting",
+        ),
+        migrations.AddField(
```
sergei-maertens, line 16: I think these migration operations can be cleaned up and/or re-generated while moving the model to the new app

sergei-maertens (changes requested):
Can you also please add a data migration for the moved analytics config from `config` app to the new one? It should copy over the values of the fields and if filled out, also enable the respective tool.

src/openforms/analytics_tools/models.py
Ewen-Lgy: Yes, I agree. It would also make the save method more readable

src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,308 @@
...
+        super().save(*args, **kwargs)
+
+    def clean(self):
+        if self.enable_google_analytics and not self.is_google_analytics_enabled:
+            raise ValidationError(
+                _(
+                    "If you enable {analytics_tool}, you must fill out all the required fields".format(
+                        analytics_tool="Google Analytics"
+                    )
+                )
```
sergei-maertens, line 253:
```suggestion
                _("If you enable {analytics_tool}, you must fill out all the required fields").format(
                    analytics_tool="Google Analytics"
                )
```
this is a common mistake - you want to apply the `.format` on the string marked for translation, not mark the resulting interpolated string translatable.

src/openforms/analytics_tools/models.py
```diff
@@ -0,0 +1,308 @@
...
+        if (
+            original_object.enable_matomo_site_analytics
+            != self.enable_matomo_site_analytics
+        ):
+            string_replacements_list = [
+                ("SITE_ID", self.matomo_site_id),
+                ("SITE_URL", self.matomo_url),
+                (
+                    "DOMAIN_HASH",
+                    lambda x: get_domain_hash(settings.ALLOWED_HOSTS[0], cookie_path=x),
```
sergei-maertens, line 196: I'd pass the entire `cookie` object instead of just the `path` to make it bit more generic:

```suggestion
                    lambda cookie: get_domain_hash(settings.ALLOWED_HOSTS[0], cookie_path=cookie["path"]),
```

src/openforms/config/migrations/0031_move_analytics_tools_data.py
```diff
@@ -0,0 +1,154 @@
...
+    get_cookies,
+    get_csp,
+    get_domain_hash,
+    update_analytical_cookies,
+    update_csp,
+)
+
+
+class Migration(migrations.Migration):
+    def move_analytics_config(apps, _):
```
sergei-maertens, line 17: please define this outside of the `Migration` class :)

src/openforms/config/migrations/0031_move_analytics_tools_data.py
```diff
@@ -0,0 +1,154 @@
...
+        AnalyticsToolsConfiguration = apps.get_model(
+            "analytics_tools", "AnalyticsToolsConfiguration"
+        )
+        GlobalConfiguration = apps.get_model("config", "GlobalConfiguration")
+        CookieGroup = apps.get_model("cookie_consent", "CookieGroup")
+
+        analytics_config = AnalyticsToolsConfiguration.objects.first()
+        config = GlobalConfiguration.objects.first()
+
+        if analytics_config and config:
```
sergei-maertens, line 28: use a guard clause here to de-indent the code and make it flatter:

```py
if not analytics_config or not config:
    return
```

src/openforms/config/migrations/0031_move_analytics_tools_data.py
```diff
@@ -0,0 +1,154 @@
...
+
+            if config.analytics_cookie_consent_group:
+                analytics_config.analytics_cookie_consent_group = (
+                    CookieGroup.objects.get(pk=config.analytics_cookie_consent_group.pk)
+                )
```
sergei-maertens, line 33:
```suggestion
            analytics_config.analytics_cookie_consent_group = config.analytics_cookie_consent_group
```

* if it's not set, on the `config`, then it'll just be `None` being assigned, which is fine
* if it exists, you can just directly set the instance - no need to re-query them, you already have the model instance.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1853, opened 2022-08-11
Profiling management command for variable injection

sergei-maertens:
Relates to #1708

This uses a formio configuration from a test environment with a total of 21 components, and runs some template evaluation for a number of attributes. Then, the number of components is multiplied by a factor to determine how it scales with form size.

Results on my local env:

```
Configuration components scale factor: 1
Input configuration has 21 components.
100 loop(s), average duration: 7.151ms

Configuration components scale factor: 10
Input configuration has 210 components.
100 loop(s), average duration: 70.849ms

Configuration components scale factor: 100
Input configuration has 2100 components.
100 loop(s), average duration: 712.561ms
```

With a single iteration (one-pass, what the endpoints would do) it's worse:

```
Configuration components scale factor: 1
Input configuration has 21 components.
1 loop(s), average duration: 23.657ms

Configuration components scale factor: 10
Input configuration has 210 components.
1 loop(s), average duration: 77.419ms

Configuration components scale factor: 100
Input configuration has 2100 components.
1 loop(s), average duration: 718.980ms
```

However, I can't pinpoint what exactly it is that makes it slower, since the scaled up configurations do perform similarly as above.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,6 +45,7 @@ omit =
     src/openforms/plugins/management/commands/disable_demo_plugins.py
     src/openforms/payments/management/commands/checkpaymentemaildupes.py
     src/openforms/registrations/contrib/email/views.py
+    src/openforms/formio/management/commands/profile_variable_injection.py
 
 [coverage:report]
 skip_covered = True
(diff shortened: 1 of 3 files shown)


sergei-maertens:
Swapping to Jinja2 template engine without any bells or whistles gave comparable results:

```
Configuration components scale factor: 1
Input configuration has 21 components.
10 loop(s), average duration: 17.908ms

Configuration components scale factor: 10
Input configuration has 210 components.
10 loop(s), average duration: 171.119ms

Configuration components scale factor: 100
Input configuration has 2100 components.
10 loop(s), average duration: 1703.485ms
```

This is still in the order of ~1ms per component.

sergei-maertens:
We've added more tracing using APM and get proper insights through Elastic APM

PR #1857, opened 2022-08-11
Refactor submission auth

SilviaAmAm:
Part of #1661


src/openforms/authentication/migrations/0002_migrate_submission_data.py
```diff
@@ -0,0 +1,76 @@
...
+    Submission = apps.get_model("submissions", "Submission")
+
+    submissions = Submission.objects.filter(~Q(auth_plugin=""))
+
+    auth_info_to_create = []
+    for submission in submissions:
+        if hasattr(submission, "auth_info"):
+            continue
+
+        value = submission.bsn or submission.kvk or submission.pseudo
```
sergei-maertens, line 20: that'll always pick `submission.bsn` if it's hashed already. I think you need to select the appropriate attribute based on the plugin used to grab the correct field.

Probably best to hardcode in the migration which plugin ID maps to which `AuthAttribute`

src/openforms/authentication/migrations/0002_migrate_submission_data.py
```diff
@@ -0,0 +1,76 @@
...
+                attribute_hashed=submission.auth_attributes_hashed,
+                submission=submission,
+            )
+        )
+
+    AuthInfo.objects.bulk_create(auth_info_to_create)
+
+
+def backwards(apps, schema_editor):
+    AuthInfo = apps.get_model("of_authentication", "AuthInfo")
```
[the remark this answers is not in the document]
sergei-maertens: okay, before merging I'll try them out locally forwards/backwards!

src/openforms/authentication/models.py
```diff
@@ -0,0 +1,77 @@
...
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from openforms.contrib.kvk.validators import validate_kvk
+from openforms.utils.validators import validate_bsn
+
+from .constants import AuthAttribute
+
+
+# TODO: what about co-sign data?
```
sergei-maertens, line 11: Let's leave it as is for now, but it should get a similar construct indeed. There's some other co-sign rework in the pipeline, we can bring this in when we're dealing with that.

src/openforms/authentication/models.py
```diff
@@ -0,0 +1,77 @@
...
+        to="submissions.Submission",
+        verbose_name=_("Submission"),
+        on_delete=models.CASCADE,
+        help_text=_("Submission related to this authentication information"),
+        related_name="auth_info",
+    )
+
+    class Meta:
+        verbose_name = _("Authentication Information")
+        verbose_name_plural = _("Authentication Information")
```
sergei-maertens, line 52:
```suggestion
        verbose_name = _("Authentication details")
        verbose_name_plural = _("Authentication details")
```

src/openforms/authentication/tests/test_migrations.py
```diff
@@ -0,0 +1,34 @@
...
+        Form = apps.get_model("forms", "Form")
+
+        form = Form.objects.create(name="Form", slug="form")
+        submission = Submission.objects.create(
+            form=form,
+            kvk="123456789",
+            bsn="",
+            pseudo="",
+            auth_plugin="test-kvk-plugin",
+            auth_attributes_hashed=True,
```
sergei-maertens, line 22: that doesn't make sense with the unhashed values - can you add some explicit checks for submissions with/without hashed details? Keep in mind that hashing a submission also hashes the _empty_ fields!

src/openforms/authentication/utils.py
```diff
@@ -24,8 +25,19 @@ def store_auth_details(submission: Submission, form_auth: FormAuth) -> None:
...
+        auth_info.save()
+        return
+
+    auth_data = AuthInfo(
+        plugin=plugin,
+        attribute=attribute,
+        value=form_auth["value"],
+        submission=submission,
+    )
+    auth_data.save()
```
sergei-maertens, line 43: could be `AuthInfo.objects.update_or_create(submission=submission, defaults=...)` instead?

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -72,9 +72,14 @@ def _get_values_for_bsn(cls, bsn: str, attributes: Iterable[str]) -> Dict[str, A
     def get_prefill_values(
         cls, submission: Submission, attributes: List[str]
     ) -> Dict[str, Any]:
-        if not submission.bsn:
+        if (
+            not hasattr(submission, "auth_info")
+            or submission.auth_info.attribute != AuthAttribute.bsn
+        ):
+            #  If there is no bsn we can't prefill any values so just return
+            logger.warning("No BSN associated with submission, cannot prefill.")
```
sergei-maertens, line 80:
```suggestion
            logger.info("No BSN associated with submission, cannot prefill.")
```
This situation probably happens on test environments, I think we can use a lower logging level.

src/openforms/prefill/contrib/stufbg/plugin.py
```diff
@@ -124,11 +124,15 @@ def _get_values_for_bsn(
     def get_prefill_values(
         self, submission: Submission, attributes: List[str]
     ) -> Dict[str, Any]:
-        if not submission.bsn:
+        if (
+            not hasattr(submission, "auth_info")
+            or submission.auth_info.attribute != AuthAttribute.bsn
+        ):
             #  If there is no bsn we can't prefill any values so just return
+            logger.warning("No BSN associated with submission, cannot prefill.")
```
sergei-maertens, line 132:
```suggestion
            logger.info("No BSN associated with submission, cannot prefill.")
```

src/openforms/submissions/models/submission.py
```diff
@@ -364,9 +324,8 @@ def is_completed(self):
     def remove_sensitive_data(self):
         from .submission_files import SubmissionFileAttachment
 
-        self.bsn = ""
-        self.kvk = ""
-        self.pseudo = ""
+        if hasattr(self, "auth_info"):
+            self.auth_info.clear_sensitive_data()
```
sergei-maertens, line 328: perhaps instead of all these `hasattr` checks, we can use `if self.is_authenticated`? That makes the code quite a bit more readable, though it's a bit of a leaky abstraction.

sergei-maertens (changes requested):
I think you can also either remove or update the roadmap entry for the auth info rework stuff.

src/openforms/authentication/utils.py
```diff
@@ -17,15 +18,10 @@ class FormAuth(TypedDict):
...
     assert (
         attribute in AuthAttribute.values
     ), f"Unexpected auth attribute {attribute} specified"
 
-    # DO NOT log this in plain text, it is considered sensitive information (BSN for example)
-    identifier = form_auth["value"]
-    submission.auth_plugin = plugin
-    setattr(submission, attribute, identifier)
-    submission.save(update_fields=["auth_plugin", attribute])
+    AuthInfo.objects.update_or_create(submission=submission, **form_auth)
```
sergei-maertens, line 27:
```suggestion
    AuthInfo.objects.update_or_create(submission=submission, defaults=form_auth)
```
so that it only matches on `submission` and not the combination of all kwargs :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1860, opened 2022-08-12
Seperate digid/eherkenning certificates

sjoerdie:
Separate digid/eherkenning certificates: need to be able to use old / new certs when certificate renewal is taking place and separate brokers load metadata at separate times cause one or the other not to work.


src/openforms/conf/base.py
```diff
@@ -859,6 +859,8 @@
 DIGID_METADATA = config("DIGID_METADATA", "")
 SSL_CERTIFICATE_PATH = config("SSL_CERTIFICATE_PATH", "")
 SSL_KEY_PATH = config("SSL_KEY_PATH", "")
```
SilviaAmAm, line 861: These are no longer used anywhere I think, so it might be good to remove them

SilviaAmAm (changes requested):

src/openforms/conf/base.py
sjoerdie: check, is done.

SilviaAmAm (approved):

PR #1863, opened 2022-08-12
Refactor logic evaluation implementation

sergei-maertens:
Related to #1708

Refactored the logic evaluation algorithm so that now we first work towards resolving the variable state (logic actions can change variable values), and then we apply the dynamic formio configuration using this resolved state/values.

The logic actions have been split up in a new interface so that we can clean up the code and side-effects applied from logic actions. This also all makes it a bit more logging friendly to get insight on what is happening when.

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -1,4 +1,3 @@
-import copy
 from copy import deepcopy
 from typing import TYPE_CHECKING, Any, Dict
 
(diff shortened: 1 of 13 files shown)


src/openforms/submissions/form_logic.py
```diff
@@ -134,72 +114,97 @@ def evaluate_form_logic(
...
         # debounced client-side data changes
         data_diff = {}
+
         for key, new_value in step.data.items():
-            original_value = data_container.initial_data.get(key)
+            original_value = initial_data.get(key)
             # Reset the value of any field that may have become hidden again after evaluating the logic
             if original_value:
                 component = get_component(configuration, key)
+                default = get_component_default_value(component)
```
SilviaAmAm, line 192: Does this handle this case? https://github.com/open-formulieren/open-forms/issues/1755

src/openforms/submissions/logic/datastructures.py
```diff
@@ -0,0 +1,69 @@
...
+        self.state.set_values(updates)
+
+    def get_updated_step_data(self, step: SubmissionStep) -> DataMapping:
+        relevant_variables = self.state.get_variables_in_submission_step(
+            step, include_unsaved=True
+        )
+        return {
+            key: variable.value
+            for key, variable in relevant_variables.items()
+            if variable.value and variable.value != variable.form_variable.initial_value
```
SilviaAmAm, line 68: I think this might be problematic for variables with bool values where `variable.value=False`

SilviaAmAm (changes requested):

src/openforms/submissions/form_logic.py
sergei-maertens: I don't know - it might, it might not, but there's a separate PR for that in progress already so we'll see how they merge and if it's still required or not.

src/openforms/submissions/logic/datastructures.py
sergei-maertens: yeah I don't like this check at all, let's see if the tests still pass without the `if variable.value` part.

sergei-maertens: tests pass locally with this change, so let's see what CI thinks

PR #1864, opened 2022-08-15
Ensure that logic variable references are updated

### LINKED ISSUE #1806: Fieldname change in business logic once fieldname is changed
### Thema / Theme

Other

When a fieldname is used in a logic rule and that fieldname is changed, the rule is not working anymore. (pointing to an empty field)
Request: once an existing fieldname is changed then check if this fieldname is used in logic rule and if so, change it accordingly.

Improvement of stability and integrity of existing forms once they are changed.

_No response_

sergei-maertens:
Fixes #1806

Component key references were correctly updated already when a Formio
component key changes, however, the variable references in logic
actions weren't yet.

This patch ensures that the variable references are also updated, and applies the same logic when (user-defined) variable keys change.

### DIFF
--- a/src/openforms/js/components/admin/form_design/form-creation-form.js
+++ b/src/openforms/js/components/admin/form_design/form-creation-form.js
@@ -47,7 +47,7 @@ import {
   getFormComponents,
   findComponent,
   checkKeyChange,
-  replaceComponentKeyInLogic,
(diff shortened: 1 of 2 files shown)


SilviaAmAm (approved):
Might be good to also rename the function  `replaceComponentKeyInLogic` since we also replace variables keys?

PR #1865, opened 2022-08-15
Feature/1294 analytics tools should play nice with csp headers

### LINKED ISSUE #1294: Analytics tools should manage CSP and Cookie settings automatically

Ewen-Lgy:
Fixes #1294 

Added test for each analytics tools in order to increase the test coverage

### DIFF
--- a/src/openforms/analytics_tools/tests/test_models.py
+++ b/src/openforms/analytics_tools/tests/test_models.py
@@ -1,3 +1,7 @@
+import uuid
+
+from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.test import TestCase, override_settings
 
 from cookie_consent.models import Cookie, CookieGroup
@@ -6,13 +10,13 @@ from openforms.config.models import CSPSetting
 
 from ..constants import AnalyticsTools
 from ..models import AnalyticsToolsConfiguration
-from ..utils import get_cookies, get_csp
+from ..utils import get_cookies, get_csp, get_domain_hash
 
 
 @override_settings(
     SOLO_CACHE=None,
 )
-class AdminTests(TestCase):
+class GoogleAnalyticsTests(TestCase):
     @classmethod
     def setUpTestData(cls):
         super().setUpTestData()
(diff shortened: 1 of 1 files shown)


src/openforms/analytics_tools/tests/test_models.py
```diff
@@ -68,11 +77,91 @@ def test_analytics_tool_properly_disabled(self):
...
+        ]
+        cls.json_cookies = get_cookies(
+            analytics_tool=AnalyticsTools.matomo,
+            string_replacement_list=string_replacements_list,
+        )
+
+        cls.json_csp = get_csp(
+            analytics_tool=AnalyticsTools.matomo,
+            string_replacement_list=string_replacements_list,
+        )
```
sergei-maertens, line 121: I think I'd avoid re-usin the implementation code here, and instead just use a hardcoded list of cookies/csp directives in the test for what you'd expect.

That means that if it turns out something is missing, we can update the test and if someone messes with the JSON files without knowing what they're doing, we have tests that check for the correct lists of cookies/CSP directives.

So in short - please hardcode this in the test rather than using the utility functions.

src/openforms/analytics_tools/tests/test_models.py
```diff
@@ -81,4 +170,279 @@ def test_analytics_tool_enabled_but_related_fields_are_not(self):
...
+
+    def setUp(self):
+        super().setUp()
+        self.config = AnalyticsToolsConfiguration.get_solo()
+        self.config.analytics_cookie_consent_group = CookieGroup.objects.get(
+            varname="analytical"
+        )
+        self.config.piwik_url = self.piwik_url
+        self.config.piwik_site_id = self.piwik_site_id
+        self.addCleanup(lambda: AnalyticsToolsConfiguration.get_solo().delete())
```
sergei-maertens, line 224: this is the "not-interesting" part that you could put in a mixin or base test case class, used by all the tool-specific test cases. That would cut down a bit on setup code.

I think I'd also split up the test file into a test file per analytics provider.

sergei-maertens (changes requested):

PR #1866, opened 2022-08-15
 feature/1294-adding-test-for-analytics-tools 

### LINKED ISSUE #1294: Analytics tools should manage CSP and Cookie settings automatically

Ewen-Lgy:
Fixes #1294


src/openforms/analytics_tools/tests/mixin.py
```diff
@@ -0,0 +1,16 @@
+from django.test import TestCase, override_settings
+
+from cookie_consent.models import CookieGroup
+
+from openforms.analytics_tools.models import AnalyticsToolsConfiguration
+
+
+@override_settings(SOLO_CACHE=None, ALLOWED_HOSTS=["*"])
+class AnalyticsToolsMixinTestCase(TestCase):
```
sergei-maertens, line 9:
```suggestion
class AnalyticsToolsMixin:
```
mixins typically don't have a base class (but they do have to call the `super` methods)

You can provide a convenience base test case using the mixin like this then:

```py
@override_settings(SOLO_CACHE=None, ALLOWED_HOSTS=["*"])
class AnalyticsTestCase(AnalyticsToolsMixin, TestCase):
    pass
```

(I'm not sure how the `@override_settings` works with base classes though)

sergei-maertens (changes requested):
Much cleaner!

sergei-maertens (approved):

PR #1872, opened 2022-08-17
[#1845] Display improvements

### LINKED ISSUE #1845: Logic tab display problems

latest

Based on https://github.com/open-formulieren/open-forms/pull/1821#pullrequestreview-1064828740

Some non-code remarks:

1. Vertical alignment of date input is a bit janky: 
image

2. Checkbox (type boolean) setting the value presents a textarea rather than a checkbox/dropdown with yes/no options: 
image

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1845 

Screenshot from 2022-08-17 14-44-24

I am not really convinced that having a dropdown for the checkboxes in the action is a good idea, because:
- the backend expects JSONLogic for the `value` field (so I used the `{"!!": [true]}` and `{"!": [true]}` for the dropdown options)
- One might want to use a more complex JSONLogic expression to assign the variable bool value (with and, or... operators) 
- The other fields with types that are not string (datetimes, times, ...) could also end up having a result of wrong type, but for those we don't have a mechanism to avoid this error.
Screenshot from 2022-08-17 14-43-42

### DIFF
--- a/src/openforms/js/compiled-lang/en.json
+++ b/src/openforms/js/compiled-lang/en.json
@@ -1695,12 +1695,6 @@
       "value": "Form definition"
     }
   ],
-  "jAraWj": [
-    {
-      "type": 0,
-      "value": "change the value of a component"
-    }
-  ],
   "jU/t8O": [
     {
       "type": 0,
@@ -1783,6 +1777,12 @@
     {
       "type": 0,
       "value": "Add item"
+    }
+  ],
+  "li/63B": [
+    {
+      "type": 0,
+      "value": "No"
     }
   ],
   "mH3X/G": [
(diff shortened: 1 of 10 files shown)


sergei-maertens:
I agree with the concerns about the boolean dropdown. What needs to happen to that JSON logic expression field is (out of scope for this ticket):

* allow primitives for values, rather than requiring an object with an operator inside (this allows `"foo"`, `null`, `true`, `123.45` etc as values - they are valid JsonLogic expressions too).
* common recipes so you can easily select a boolean etc., depending on the type of the selected variable. This probably requires  #1897 first

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #1875, opened 2022-08-18
[#1579] Variables in registration backend

### LINKED ISSUE #1579: Allow the use of variables in the registration backends
Addition to #1325 

With the introduction of variables, it would be nice to let the registration backend have access to these variables.

One of the nice things here is that we don't have to put everything in a form. For example, an internal BAG-id for a location can currently only be given to a registration backend if we add a hidden field to the form that is prefilled with a BAG-id. With variables, we simply let the variable be prefilled and never communicate this to the form - but instead attach the variable to a ZGW-API or StUF-ZDS field.

How this works exactly, it up for more discussion but as a start:

**StUF-ZDS**
- [x] Allow a variable to be added as ExtraElement or set it on any StUF-ZDS field #3238 

**ZGW-API**
- [x] Allow a variable to be added as Eigenschap or set it on any ZGW API field

**Objects API**
- [x] Allow a variable to be used a field-value of the matching Objecttype

**Email** - #2040
- [x] Allow a variable to be included in the email content or subject.
- [x] Allow variables to be included in the attachment.

**Camunda**
- [ ] Refactor the camunda process variable approach to make use of the form variables.

etc.

SilviaAmAm:
Related to #1579

### DIFF
--- a/src/openforms/submissions/tests/factories.py
+++ b/src/openforms/submissions/tests/factories.py
@@ -6,7 +6,6 @@ from django.utils import timezone
 
 import factory
 import magic
-from faker import Faker
 from glom import PathAccessError, glom
 
 from openforms.forms.tests.factories import (
@@ -149,9 +148,9 @@ class SubmissionFactory(factory.django.DjangoModelFactory):
 
         configuration = {"components": components}
 
-        faker = Faker()
+        form_definition_index = cls._setup_next_sequence()
         form_definition = FormDefinitionFactory.create(
-            name=f"definition-{faker.word()}", configuration=configuration
+            name=f"definition-{form_definition_index}", configuration=configuration
         )
         form_step = FormStepFactory.create(form=form, form_definition=form_definition)
 
--- a/src/openforms/submissions/tests/factories.py
+++ b/src/openforms/submissions/tests/factories.py
@@ -6,6 +6,7 @@ from django.utils import timezone
 
 import factory
 import magic
+from faker import Faker
 from glom import PathAccessError, glom
 
 from openforms.forms.tests.factories import (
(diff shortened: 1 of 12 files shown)


src/openforms/variables/rendering/nodes.py
```diff
@@ -0,0 +1,90 @@
...
+            return self.variable.key
+        return self.variable.form_variable.name
+
+    @property
+    def display_value(self) -> Union[str, Any]:
+        # in export mode, expose the raw datatype
+        if self.mode == RenderModes.export:
+            return self.variable.value
+
+        # TODO add a registry of formatters like for Formio ComponentNodes
```
SilviaAmAm, line 86: I wasn't sure if I should do this, since we are only rendering for registration, cli and export (not for submission report pdf, confirmation email or summary page) :thinking:

sergei-maertens: let's not do it, but add a comment that this would be needed if we'll render in different modes.

sergei-maertens (changes requested):
Couple small remarks

src/openforms/submissions/tests/factories.py
```diff
@@ -148,8 +149,9 @@ def from_components(
 
         configuration = {"components": components}
 
+        faker = Faker()
         form_definition = FormDefinitionFactory.create(
-            name=f"definition-{key}", configuration=configuration
+            name=f"definition-{faker.word()}", configuration=configuration
```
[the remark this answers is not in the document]
SilviaAmAm: I ended up using https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Factory._setup_next_sequence to get the index of the sequence

sergei-maertens (approved):

PR #1876, opened 2022-08-18
[#1856] Prevent JsonLogicValidator from causing a 500 error

### LINKED ISSUE #1856: 500 errors ignored when saving logic rules in admin ignored

latest

While creating a logic rule in the admin, if a 500 happens, no error is displayed, the rule disappears and nothing is said to the user about the error.

- In the admin, create a new form
- Add a field with key  `textFieldA`

- Save and continue editing the form
- Add a logic rule with the following trigger
```
{
  "==": [
    {
      "var": "textFieldA"
    },
    "test",
    "invalid"
  ]
}
```
- Save the form => No errors shown
- Look at the logic rules: the rule created is not present

An error should be shown that something went wrong

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1856

### DIFF
--- a/src/openforms/forms/api/validators.py
+++ b/src/openforms/forms/api/validators.py
@@ -21,7 +21,7 @@ class JsonLogicValidator:
     def __call__(self, value: dict):
         try:
             jsonLogic(value)
-        except ValueError:
+        except Exception:
             raise serializers.ValidationError(_("Invalid JSON logic."), code="invalid")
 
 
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):

sergei-maertens:
Backported in d754e2c834efe593f28c265690b2c19f2c19ff73 and 9b6ab530a3863a4332e58f9d646d43f8078d1644, the test has not been backported as the bulk endpoint is new in 1.2.

PR #1894, opened 2022-08-19
Azure AD OIDC

### LINKED ISSUE #1861: Allow employees to login to a form using OpenID-connect
DH Taiga 95

**Tasks**

* [x] Add an Authentication plugin for OpenID-connect (this is VERY similar to: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/contrib/digid_eherkenning_oidc/plugin.py)
* [x] Make sure the plugin re-uses the general OpenID-connect configuration (the one under the Account menu settings in the admin).
* [x] Make sure we can configure a claim that holds the employee-number (much like the kvk-number in above linked code).
* [x] Make sure the employee number is saved on the form submission (much like how the bsn, kvk, pseudo id are saved at the moment). It should be hashed (or whatever is used for bsn, etc.)
* [x] Test the system as a whole against Azure AD and as an employee-number claim we can use the email (configured as mapping in the admin), since I'm sure you will get that.

dariomory:
Fixes #1861


sergei-maertens:
@dariomory we need to discuss this first thing tomorrow, it is not going in the right direction at all.

sergei-maertens:
Discussed via call - a new branch and PR will be created.

PR #1911, opened 2022-08-23
Document the variable injection in formio configuration

sergei-maertens:
This covers part of the documentation needs in #1711 as well.


docs/manual/forms/form_fields.rst
```diff
@@ -17,17 +17,58 @@ Basic
...
+   https://github.com/open-formulieren/open-forms/pull/1893)
+
+U kunt variabelen gebruiken in het ontwerp van formulieren, bijvoorbeeld om een
+(voor-)ingevulde naam in het label van een ander formulierveld weer te geven. Hiervoor
+is de :ref:`sjabloonfunctionaliteit <manual_templates>` beschikbaar.
+
+.. note::
+
+    Alle velden in een formulier zijn als variabele beschikbaar. Daarnaast zijn er ook
+    altijd een aantal standaardvariabelen beschikbaar én u kunt zelf ook variabelen
```
SilviaAmAm, line 44: In the variable docs I think we referred to these as 'vaste variabelen'? Not sure if its worth mentioning here instead of standaardvariabelen

SilviaAmAm (approved):

docs/manual/forms/form_fields.rst
sergei-maertens: yeah, I'll update this PR :-) consistency is better

PR #1926, opened 2022-08-25
[#1871] Check if component is visible in the frontend

### LINKED ISSUE #1871: Fields made visible by front end logic can't be filled

latest

DH Taiga user story 120 + issue 159, user story 167 + issue 162

If a hidden field is made visible by the frontend logic (the logic specified in the component), then it cannot be filled.

Create a form in the admin:
- Add a radio component with choices yes/no
- Add a text area which is hidden. In the logic tab of the component, configure it so that it should be visible if the value of the radio component is `yes`.
- Save the form

Fill in the form in the front end.
- Select `yes` for the radio button
- The text area appears. Try to fill in data. After the logic check the data disappears.

The data should not disappear.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1871


src/openforms/formio/utils.py
```diff
@@ -221,3 +221,30 @@ def conform_to_mask(value: str, mask: str) -> str:
             )
 
     return "".join(result)
+
+
+def is_visible_in_frontend(component: dict, data: dict) -> bool:
```
sergei-maertens, line 226:
```suggestion
def is_visible_in_frontend(component: JSONObject, data: DataMapping) -> bool:
```
these type hints are more precise (both available in `openforms.typing`)

src/openforms/formio/utils.py
```diff
@@ -221,3 +221,30 @@ def conform_to_mask(value: str, mask: str) -> str:
...
+    conditional = component.get("conditional", {})
+
+    if conditional.get("show") is None:
+        return not hidden
+
+    trigger_component_key = conditional.get("when")
+    if data.get(trigger_component_key) == conditional.get("eq"):
+        return conditional.get("show")
+
+    return not conditional.get("show")
```
sergei-maertens, line 250: I think this can be made a bit more readable and flatter

```suggestion
    hidden = component.get("hidden")
    conditional = component.get("conditional")
    if not conditional or (conditional_show := conditional.get("show")) is None:
        return not hidden

    trigger_component_key = conditional.get("when")
    trigger_component_value = data.get(trigger_component_key)
    compare_value = conditional.get("eq")
    return conditional_show if trigger_component_value == compare_value else not conditional_show
```

sergei-maertens (changes requested):
Not 100% sure about the refactor of the code, but I think we can/should avoid the repeated calls to `dict.get(key)` with the same key.

sergei-maertens (approved):

PR #1932, opened 2022-08-26
[#1855] Variables off sync

### LINKED ISSUE #1855: Refreshing the admin form editor after a failed save leaves variables in off sync state

latest

In the following flow, the variables are left in a state that is not in sync with the form definition:
- Edit form in editor
- Save form => Receives an error
- Refresh page
This causes a lot of errors for the variables, because the form definition has been saved, but not the variables, so that then they are no longer in sync

- In the admin, create a new form
- Add a form step
- Add a text field 
- Add another text field
- Save and continue editing the form
- Now remove the last text field
- Add a new text field with an invalid key like `test invalid` (with the space to make it invalid)
- Save form => This should give an error
- Refresh page
- Now the variables no longer match the components in the form definition

The variables should remain in sync with the form definition

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #1855

### DIFF
--- a/src/openforms/forms/api/viewsets.py
+++ b/src/openforms/forms/api/viewsets.py
@@ -20,6 +20,7 @@ from rest_framework.response import Response
 from openforms.api.pagination import PageNumberPagination
 from openforms.api.serializers import ExceptionSerializer, ValidationErrorSerializer
 from openforms.utils.patches.rest_framework_nested.viewsets import NestedViewSetMixin
+from openforms.variables.constants import FormVariableSources
 
 from ..messages import add_success_message
 from ..models import (
@@ -32,7 +33,7 @@ from ..models import (
     FormVersion,
 )
 from ..utils import export_form, import_form
-from .filters import FormLogicFilter, FormPriceLogicFilter
+from .filters import FormLogicFilter, FormPriceLogicFilter, FormVariableFilter
 from .parsers import (
     FormCamelCaseJSONParser,
     IgnoreConfigurationFieldCamelCaseJSONParser,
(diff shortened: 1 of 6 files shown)


src/openforms/forms/api/viewsets.py
```diff
@@ -517,6 +518,17 @@ def variables_bulk_update(self, request, *args, **kwargs):
...
+            filter_serializer = FormVariableSourceSerializer(
+                data=self.request.query_params
+            )
+            is_valid = filter_serializer.is_valid()
+            if not is_valid:
+                return Response([], status=status.HTTP_200_OK)
+
+            form_variables = form_variables.filter(
+                source=filter_serializer.validated_data["source"]
+            )
```
sergei-maertens, line 531: wouldn't it make more sense to use https://pypi.org/project/django-filter/? Afaik it also has API schema introspection capabilities in combination with drf-spectacular.

src/openforms/forms/tests/variables/test_viewset.py
```diff
@@ -205,6 +207,83 @@ def test_list_form_variables(self):
...
+
+        url = furl(
+            reverse(
+                "api:form-variables",
+                kwargs={"uuid_or_slug": form.uuid},
+            )
+        )
+        url.args["source"] = FormVariableSources.user_defined
+
+        response_user_defined = self.client.get(url.url)
```
sergei-maertens, line 241: usually I try to only use `furl` in situations where complicated URL manipulation is needed, but with the django test cases this isn't needed since you can pass a dict for query params:

```suggestion
        response_user_defined = self.client.get(
            reverse("api:form-variables", kwargs={"uuid_or_slug": form.uuid}),
            {"source": FormVariableSources.user_defined},
        )
```

sergei-maertens (changes requested):

src/openforms/forms/api/viewsets.py
SilviaAmAm: Yes I wanted to use it, but I couldn't quite figure out how to make it work for a viewset custom action (since the variables endpoint is an action in the form viewset).

sergei-maertens: I think you can leverage the `filter_backends` keyword argument, see https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

sergei-maertens (approved):

PR #1936, opened 2022-08-26
Add data migration for form explanation message — [#1933]

### LINKED ISSUE #1933: Remove login message on the form start page
By default, the form start page says something like: "Log in of start het formulier anoniem." depending on the authentication options.

This is currently not adjustable. My suggestion is to remove this text and let the users fill this in with the start page template if they want to.

**Tasks**

* [x] Remove message from the API/SDK (wherever it is).
* [x] Make a data migration that appends this message (depending on the auth options) to the `form.explanation_template`

dariomory:
Fixes #1933

- Depends on SDK changes: https://github.com/open-formulieren/open-forms-sdk/pull/223


dariomory:
@sergei-maertens Please double check the internationalization aspect.

sergei-maertens (changes requested):

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 3.2.14 on 2022-08-26 15:29
+from django.db import migrations
+
+
+def append_explanation_template_message(apps, schema_editor):
+    from openforms.authentication.registry import register as auth_register
+    from django.utils.translation import gettext_lazy as _
```
sergei-maertens, line 7: why is this a local import?

You can also use the non-lazy variant, we aren't defining any translations strings at the module level.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Form = apps.get_model("forms", "Form")
+    for form in Form.objects.all():
+        plugins = form.authentication_backends
+        login_options_count = len(
+            [plugin_id for plugin_id in plugins if plugin_id in auth_register._registry]
```
sergei-maertens, line 13: there's no need to use private API.

You can use either:

1. `auth_register.items()`
2. `plugin_id in auth_register`
3. `[plugin.identifier for plugin in auth_register]`

To get the currently available plugins.

Additionally, you can build this list of available plugins/plugin identifiers _outside_ of the for loop, as it doesn't change because of the loop.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+        if form_login_required:
+            message = _("Please authenticate to start the form.")
+        elif can_login:
+            message = _("Please authenticate or start the form anonymously.")
+        else:
+            message = _("Please click the button below to start the form.")
+
+        form.explanation_template += f'<p id="start-form">{message}</p>'
```
sergei-maertens, line 28: if there is HTML in the translated messages, you now have broken HTML/possible script injection.

Use `django.utils.html.format_html` when building HTML in python.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
```diff
@@ -0,0 +1,56 @@
...
+    for form in Form.objects.all():
+        plugins = form.authentication_backends
+        login_options_count = len(
+            [plugin_id for plugin_id in plugins if plugin_id in auth_register._registry]
+        )
+        can_login = login_options_count > 0
+        form_login_required = any(
+            form_step.form_definition.login_required
+            for form_step in form.formstep_set.all()
+        )
```
sergei-maertens, line 19: you can prefetch the form steps to prevent a cascade of queries. But even better would be to filter only the forms that require login directly in the queryset on line 10.

src/openforms/js/lang/en.json
```diff
@@ -339,11 +339,6 @@
     "description": "Form step internal name field help text",
     "originalDefault": "Internal name of the form definition used in this form step"
   },
-  "C6xxed": {
-    "defaultMessage": "Example initial",
-    "description": "Variable table initial value title",
-    "originalDefault": "Example initial"
-  },
```
sergei-maertens, line 346: why are these files changed?

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
```diff
@@ -0,0 +1,56 @@
...
+        form.save()
+
+
+def remove_explanation_template_message(apps, schema_editor):
+    from bs4 import BeautifulSoup
+
+    Form = apps.get_model("forms", "Form")
+
+    for form in Form.objects.all():
+        soup = BeautifulSoup(form.explanation_template, "html.parser")
```
sergei-maertens, line 38: we can use lxml, it's a dependency of the project anyway.

sergei-maertens (changes requested):
Where are the tests for the data migration?

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: I want to separate concerns between migration and reverse migration.

Can you explain the advantages of using the non-lazy variant in this particular case?

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: > Additionally, you can build this list of available plugins/plugin identifiers outside of the for loop, as it doesn't change because of the loop.

The`plugins` variable is coming from `form.authentication_backends`.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: We can, but we don't need the extra functionality offered by lxml.

src/openforms/js/lang/en.json
dariomory: I am just running `./bin/makemessages`

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: Fortunately there is no html in the translated messages. Good tip :)

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: Good catch. Is optimizing querysets in a migration file worth the time investment?

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: > I want to separate concerns between migration and reverse migration.

Imports in python always go to the top of the module, unless you are dealing with circular imports.

For the same reason, the `register` import should probably also just go to the top of the module.

> Can you explain the advantages of using the non-lazy variant in this particular case?

It doesn't output lazy strings, but actual translated strings. The lazy variant in Django only exists because often translations need to be defined at module level (so loaded/evaluated at import-time) and be translated at runtime.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: I am talking about `auth_register`, and the registered plugins inside.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: Yeah until sometimes slips by and another contributor adds HTML in the translation messages.

Security should always be in your mind whenever you're dealing with HTML outside of the template engine. It's not a tip, it's pointing out a security vulnerability.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: Always consider the scale of the data you're dealing with. If an organization has a 1000 forms, that's 1000 additional queries that could've been avoided.

So yes, it is worth the time investment, especially if you don't know how many objects could be affected. Assume the worst.

src/openforms/js/lang/en.json
sergei-maertens: I've run makemessages on `master` branch, if you rebase, these changes should no longer be in the diff.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: Except that lxml uses C-bindings and is much more performant in parsing the HTML, that alone is enough reason to use it if you have it.

dariomory: lxml will add html and body tags to the string. This is a migration file and we are not concerned about a small gain in performance.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: I don't see a strong reason to make this change.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
dariomory: > It doesn't output lazy strings, but actual translated strings. The lazy variant in Django only exists because often translations need to be defined at module level (so loaded/evaluated at import-time) and be translated at runtime.

What advantage are we gaining from outputting translated strings over lazy strings in this migration?

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: `_registry` starts with an underscore prefix, indicating it is private API. That is plenty of reason to clean this up properly.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: This is a Django convention per the docs: https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#standard-translation

You should use `gettext` unless there's concrete reason to use the lazy variant. The lazy variant messes a lot with type hints where you declare a `str` typehint, so it is preferential. But the main reason is the Django convention to use the non-lazy call.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: > Can you provide me a source?

You really should be able to see what happens if `message` contains HTML tags (possibly `<script>` tags). They will not be escaped.

For a source, I'm fairly certain that XSS is in the OWASP top 10.

> What would happen if a contributor added HTML in the translation message?

See above, it will not be escaped.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
sergei-maertens: > lxml will add html and body tags to the string

now _that's_ an actual valid reason to consider not using it, because it's too much hassle to strip those out.

> This is a migration file and we are not concerned about a small gain in performance.

Yes we are, for those environments were downtime is not acceptable.

The application runs migrations during startup. While the app is migrating, it is NOT available as a backend to serve traffic. Migrations should complete as fast as possible.

joeribekker (changes requested):
Sorry, too soon.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
joeribekker: @dariomory: Using a private API is bad practice, so please:

```suggestion
            [plugin_id for plugin_id in plugins if plugin_id in auth_register]
```

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
joeribekker: @dariomory Typing out this whole discussion took more time than adding a prefetch, add the prefetch please.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
joeribekker: @dariomory Please use  `django.utils.html.format_html` and don't use the `id="start-form"` in it because it will be lost once someone saves the form again in the admin.

src/openforms/forms/migrations/0042_explanation_template_data_migration.py
joeribekker: I'm going to be practical here, and we can remove the backward migration entirely. This because the `id` will also not be used (see above).

NOTE: This is a PO-decision not a technical decision :)

joeribekker (changes requested):
Please rebase before you make the suggested changes.

sergei-maertens (approved):

PR #1937, opened 2022-08-26
Validate template expressions in form definition admin and endpoint

### LINKED ISSUE #1708: Interface to handle inserting variables in Formio configuration

sergei-maertens:
Fixes #1708

- [x] Implement admin-level validation
- [x] Implement API level validation
- [x] Display validation errors in form designer
  This is a bit of a shitty solution - the errors are now displayed _above_ the form builder. I can't figure out how to pass the custom backend validation errors down to the form designer so that it displays the component with an error state AND displays the error in the form builder, all while highlighting the correct configuration tab.


src/openforms/js/components/admin/form_design/FormStepDefinition.js
```diff
@@ -253,15 +281,11 @@ FormStepDefinition.propTypes = {
   errors: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.string)),
 };
 
-const ConfigurationErrors = ({errors = []}) => {
-  const configurationErrors = errors.filter(([name, err]) => name === 'configuration');
-  const [hasConfigurationErrors, formattedConfigurationErrors] =
-    normalizeErrors(configurationErrors);
-
-  if (!hasConfigurationErrors) return null;
+const ConfigurationErrorList = ({errors = []}) => {
```
SilviaAmAm, line 284: Why not use https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/forms/ErrorList.js#L4 ?

SilviaAmAm (commented):
Looks good!

I notice that atm if there is an error shown in the builder and the user updates a field, the error is not cleared (this happens for validation errors on other fields in the editor). It's not too big a deal, but I thought I'd mention it for consistency

src/openforms/js/components/admin/form_design/FormStepDefinition.js
sergei-maertens: I didn't realize that's (almost) the same! I saw it because in this file there was some manual error list building and figured we didn't have anything for it yet.

I'll update to use the re-usable component.

sergei-maertens:
> I notice that atm if there is an error shown in the builder and the user updates a field, the error is not cleared (this happens for validation errors on other fields in the editor). It's not too big a deal, but I thought I'd mention it for consistency

Yes, I noticed it too. Clearing that would involve hooking into the formio configuration change, figuring out which component changed, what the path in the tree of it is and then removing the errors for that component :thinking: 

Not impossible, but also not easy. I'll see if I can get that information without too much pain.

src/openforms/js/components/admin/form_design/FormStepDefinition.js
sergei-maertens: In the end we needed `MessageList` component rather than `ErrorList` because of the classnames involved, and I saw that we've been doing some funky stuff with that anyway so I refactored it. Now we are only using existing components.

sergei-maertens:
@SilviaAmAm the error now clears!

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -122,6 +122,39 @@ const parseValidationErrors = (errors, prefix) => {
...
+// it similar to the backend.
+function* _componentWithPathGenerator(components, prefix = 'components') {
+  for (let index = 0; index < components.length; index++) {
+    const component = components[index];
+    const fullPath = `${prefix}.${index}`;
+    yield [fullPath, component];
+
+    // taken from FormioUtils.eachComponent
+    const hasColumns = component.columns && Array.isArray(component.columns);
+    // const hasRows = component.rows && Array.isArray(component.rows);
```
SilviaAmAm, line 135: This is commented out instead of removed in case we at some point enable the table component?

SilviaAmAm (approved):

src/openforms/js/components/admin/form_design/utils.js
sergei-maertens: correct!

PR #1893, opened 2022-08-19
[#1705] Variables documentation

SilviaAmAm:
Fixes #1705 

TODO:
- User defined Variables with prefill or in calculations?

PR #1939, opened 2022-08-29
Add additional static variables

### LINKED ISSUE #1711: Additional static variables

sergei-maertens:
Fixes #1711

I'll create a separate PR with some minor cleanup/refactor - we can get rid of the `submission` usage in the `forms` app completely.

### DIFF
--- a/src/openforms/variables/base.py
+++ b/src/openforms/variables/base.py
@@ -1,7 +1,9 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from typing import Optional
 
 from openforms.forms.models import FormVariable
+from openforms.submissions.models import Submission
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


src/openforms/variables/base.py
```diff
@@ -1,7 +1,9 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from typing import Optional
 
 from openforms.forms.models import FormVariable
+from openforms.submissions.models import Submission
```
SilviaAmAm, line 6:
```suggestion
if TYPE_CHECKING:
    from openforms.submissions.models import Submission
```
Since it's only used in the type hints?
(=> then quotes around "Submission" in the type hints)

SilviaAmAm (approved):
Should I add documentation about these new static vars in https://github.com/open-formulieren/open-forms/pull/1893?

src/openforms/variables/base.py
sergei-maertens: the `TYPE_CHECKING` workaround is mostly used to solve circular imports, if it works without, that leads to cleaner code.

sergei-maertens:
> Should I add documentation about these new static vars in #1893?

No - that will quickly go out of sync if we add/modify/remove static variables. As far as I can see, at the moment we don't have any static variables documented, not in your referenced PR either. They will show up in the form designer, so let's stick with that for now.

Ideally, we can write a Sphinx extension at some point to document the variables through introspection, using the registry we set up, so then it will automatically be in sync.

PR #1940, opened 2022-08-29
Refactored the public API for (static) variables 

sergei-maertens:
Instead of `FormVariable.get_static_data` (and leaking `Submission` objects into the `forms` app), we can now use `openforms.variables.service.get_static_variables` for a better separation of concerns.


SilviaAmAm (approved):

sergei-maertens:
Failing test is a Heisenbug... weird.

PR #1945, opened 2022-08-31
[#1896] Primitives and arrays are valid jsonLogic

### LINKED ISSUE #1896: Relax JsonLogic validation in form designer logic tab
### Thema / Theme

Form designer

The logic tab in the form designer has the action to set the value of a component/variable, which uses JsonLogic under the hood.

If you want to simply set a simple primitive value (boolean, string, number...) for a variable, then you cannot literally enter this as it fails JsonLogic validation, which expects an object with an operator key inside.

Currently, you need to workaround this by using a relevant operator, e.g. `{"cat": ["the string I want"]}` instead of the more simple expression `"the string I want"`.

More intuitive logic rules editing.

Both the JS and python libraries accept primitives as valid JsonLogic expressions, so this should not pose any problems.

SilviaAmAm:
Fixes #1896 

The json-logic-js function `is_logic` only accepts objects: https://github.com/jwadhams/json-logic-js/blob/master/logic.js#L176 (while if we enter primitive values/arrays in the [play-with-it page](https://jsonlogic.com/play.html) these are accepted).

### DIFF
--- a/src/openforms/js/components/admin/forms/JsonWidget.js
+++ b/src/openforms/js/components/admin/forms/JsonWidget.js
@@ -12,6 +12,28 @@ const jsonFormat = value => {
   return JSON.stringify(value, null, 2);
 };
 
+const isJsonLogic = jsonExpression => {
+  // jsonLogic accepts primitives
+  if (
+    jsonExpression == null || // typeof null -> 'object'
+    typeof jsonExpression === 'string' ||
+    typeof jsonExpression === 'boolean' ||
+    typeof jsonExpression === 'number'
+  ) {
+    return true;
+  }
+
+  if (Array.isArray(jsonExpression)) {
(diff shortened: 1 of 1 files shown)


sergei-maertens:
> Fixes #1896
> 
> The json-logic-js function `is_logic` only accepts objects: https://github.com/jwadhams/json-logic-js/blob/master/logic.js#L176 (while if we enter primitive values/arrays in the [play-with-it page](https://jsonlogic.com/play.html) these are accepted).

Looking at `rule_like`, that seems to cover for primitives: https://github.com/jwadhams/json-logic-js/blob/master/logic.js#L415

So I think we're good in allowing primitives, which is in line with the playground.

sergei-maertens (approved):

PR #1949, opened 2022-08-31
Global default for allowed filetypes

### LINKED ISSUE #1921: Global default for allowed file types
### Thema / Theme

Form designer

DH Taiga user story 166 + issue 165

Currently, at DH they always configure the same allowed file types in the upload component. They would like to be able to configure a global default and then this would be used automatically in the upload components.

Less error-prone than having to configure the same thing multiple time

_No response_

dariomory:
Fixes #1921


src/openforms/config/models.py
```diff
@@ -191,6 +192,15 @@ class GlobalConfiguration(SingletonModel):
             "'(optional)' and required fields are unmarked."
         ),
     )
+    form_upload_default_file_types = ArrayField(
+        models.CharField(max_length=12, choices=mimetypes.types_map.items()),
+        verbose_name=_("Default allowed file upload types"),
+        help_text=_(
+            "Provide a list of default allowed file upload types. If empty, all extensions are allowed."
```
joeribekker, line 199: > If empty, all extensions are allowed.

Is this true? I thought you need to pass `*` (All) as an option in the extensions list.

joeribekker (changes requested):
I'm not sure if this is how I want it.

From what I see, this will change the default value in the form designer when a) I open the file-component, and b) the extensionlist is empty.

So, this is not really a global default. It's a global initial value for newly placed components. My first thought is that this is confusing. And when MS introduces "docxx" we still need to change all file fields.

Am I correct that it works this way? If so, we need to discuss alternative implementations.

src/openforms/config/models.py
dariomory: Yes, this is correct. `*` and an empty list are equivalent.

dariomory:
@joeribekker Exactly, this works as a global initial value.

I understand we need something more versatile that will apply to all already existing forms. I suppose we can apply it when any form has an empty list of allowed filetypes?

joeribekker (changes requested):
I gave it some thought and want to go in a different direction: Instead of global default that is only set upon the creation of the file component, I want to maintain a global value that - when updated - is reflected in all file components without the need to change every file-component in every form.

To that end:

* [x] Refactor these choices to a DjangoChoices version: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/form/file.js#L33
* [ ] Make sure the DjangoChoices are then used to generate this JS. This makes sure the Python and JS-list are kept in sync.
* [ ] Refactor the model configuration field `form_upload_default_file_types` to use the DjangoChoices.
* [ ] Add a new checkbox to the file component (in the form designer), below the filetype dropdown: "Use globally configured filetypes"
* [ ] Whenever the above checkbox is checked, the `fileTypes` in the file component JSON should include the currently configured `form_upload_default_file_types`

We don't need any new API endpoints for this, so much of the current code can be gone.

dariomory:
@joeribekker Assuming this is all correct my plans for the SDK are:

- If `useConfigFiletypes` is true: Update the formio file validation pattern with global configuration **or** simply allow the backend to perform file pattern validation (already done here).

joeribekker:
I don't think any ask change is needed

dariomory:
> I don't think any sdk change is needed

It may be needed to avoid using the regular validation pattern when the checkbox is true and use the global configuration instead.

sergei-maertens:
> @joeribekker Assuming this is all correct my plans for the SDK are:
> 
> * If `useConfigFiletypes` is true: Update the formio file validation pattern with global configuration **or** simply allow the backend to perform file pattern validation (already done here).

This should be done dynamically in the backend, see `openforms.formio.service.update_configuration_for_request`.

No changes to the SDK are required.

joeribekker (approved):
Looks good to me!

src/openforms/config/constants.py
```diff
@@ -53,3 +53,28 @@ class CSPDirective(DjangoChoices):
     # CSP_UPGRADE_INSECURE_REQUESTS  # Include upgrade-insecure-requests directive. A boolean. False See: upgrade-insecure-requests
     # CSP_BLOCK_ALL_MIXED_CONTENT  # Include block-all-mixed-content directive. A boolean. False See: block-all-mixed-content
     # CSP_INCLUDE_NONCE_IN  # Include dynamically generated nonce in all listed directives, e.g. CSP_INCLUDE_NONCE_IN=['script-src'] will add 'nonce-<b64-value>' to the script-src directive.
+
+
+class UploadFileType(DjangoChoices):
+    ALL = ChoiceItem("*", label="All")
```
joeribekker, line 59: Please convert all vars to lowercase (ALL -> all) as a convention.

For the `label="All"` this can be translated as well: `_("All")`

src/openforms/submissions/tests/test_submission_attachment.py
```diff
@@ -14,6 +14,7 @@
 from openforms.forms.tests.factories import FormStepFactory
 from openforms.tests.utils import disable_2fa
 
+from ...config.models import GlobalConfiguration
```
joeribekker, line 17: Should be absolute import since its in a different app.

src/openforms/formio/service.py
```diff
@@ -53,12 +56,23 @@ def update_configuration_for_request(configuration: dict, request: Request) -> N
 
     The configuration is modified in the context of the provided :arg:`submission`.
     """
-    update_urls_in_place(configuration, request)
+    pipeline = (
+        update_urls_in_place,
+        update_default_file_types,
+    )
```
joeribekker, line 62: A pattern emerges here! But we already have a pattern: Can you refactor this into a register and register these "formatters" like here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/formatters/service.py  and https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/formatters/default.py

We probably will have more to come!

joeribekker (changes requested):

sergei-maertens (approved):

Issue #1944, opened 2022-08-30
Navigate between submission steps only if the previous ones are completed


SilviaAmAm:
### Thema / Theme

Other

### Omschrijving / Description

DH Taiga user story 121 + issue 28

If a user is not logged in as admin, they shouldn't be able to skip steps by clicking on the progress indicator or using the urls of following steps.

Todo:

Part 1: 
- [x] Add a middleware that sets the X-CSRF-Token header https://github.com/open-formulieren/open-forms/pull/1952
- [x] Update the SDK so that it extracts the CSRF token and saves it in memory
- [x] Remove workarounds from the SDK that currently set the CSRF Token https://github.com/open-formulieren/open-forms-sdk/pull/224

Part 2:
- [x] Add session authentication to the Submission/SubmissionStep endpoints. 
- [x] Document header in API schema
- [x] Add the backend logic to check if the user is admin and if they can skip steps. https://github.com/open-formulieren/open-forms/pull/1958
- [x] Add header `X-Is-Staff` so that the SDK knows whether the user is allowed to navigate between steps https://github.com/open-formulieren/open-forms/pull/2006
- [ ] Update the SDK so that navigating between steps that are not completed is not possible https://github.com/open-formulieren/open-forms-sdk/pull/231

PR #1952, opened 2022-09-01
[#1944] Set csrf token header

SilviaAmAm:
Fixes Part 1 of #1944


sergei-maertens (changes requested):
I'm missing the updated API schema documentation, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/api/drf_spectacular/hooks.py#L38 for the drf-spectacular hook

sergei-maertens (approved):

PR #1962, opened 2022-09-02
Manually resend failed registrations — [#1957]

### LINKED ISSUE #1957: I cannot manually resend failed registrations after they reached the sent limit
When you check a submission in the list view, and click the bulkaction retry submission, it doesn't do anything.

This manual action should always work!

image

dariomory:
Fixes #1957

### DIFF
--- a/src/openforms/submissions/admin.py
+++ b/src/openforms/submissions/admin.py
@@ -168,7 +168,12 @@ class SubmissionAdmin(admin.ModelAdmin):
         "created_on",
         "completed_on",
     )
-    list_filter = (SubmissionTypeListFilter, "form")
+    list_filter = (
+        SubmissionTypeListFilter,
+        "registration_status",
+        "needs_on_completion_retry",
+        "form",
+    )
     search_fields = (
         "form__name",
         "uuid",
@@ -301,7 +306,7 @@ class SubmissionAdmin(admin.ModelAdmin):
     )
 
     def retry_processing_submissions(self, request, queryset):
-        submissions = queryset.filter(needs_on_completion_retry=True)
+        submissions = queryset.filter(registration_status=RegistrationStatuses.failed)
         messages.success(
             request,
             ngettext(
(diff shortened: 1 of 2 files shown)


joeribekker (changes requested):
This is correct but, now that I see this, I remember why this was done the way it is.

It allows an admin to simply select everything and retry what needs to be retried. If you add an admin filter on `RegistrationStatuses` and `ProcessingResults` (above the form-filter please) then I'll approve it in full.

src/openforms/submissions/admin.py
```diff
@@ -301,7 +306,7 @@ def export_xml(self, request, queryset):
     )
 
     def retry_processing_submissions(self, request, queryset):
-        submissions = queryset.filter(needs_on_completion_retry=True)
+        submissions = queryset.all()
```
joeribekker, line 309:
```suggestion
        submissions = queryset.filter(registration_status=RegistrationStatuses.failed)
```

This will allow us to resend failed registrations (if they are no longer tried automatically or even if they still are).

joeribekker (changes requested):
In retrospect, one more change!

joeribekker (approved):

joeribekker (changes requested):
Can you change this test: https://github.com/open-formulieren/open-forms/pull/1962/commits/9971cf381d66195284ff55495847381043c11e26#diff-e2a486848e14c62d1ecd9b783307ee3d1be3822bb99cf8b090a6411f42853319R112

To reflect the new behaviour. Currently it is by accident that it works.

Please add one more test that has `needs_...retry=False` and `registration_status=failed` to show that that one is also processed.

sergei-maertens (approved):

sergei-maertens:
Backported in fdfd3db61176f85a95cab905a2f3027879ab92f1 and 7f030e9dc2a3a9f650ce31c38768d018ffb25a90

PR #1966, opened 2022-09-02
Making different logic checks/evaluations consistent

### LINKED ISSUE #1913: Align check_submission_logic with evaluate_form_logic
https://github.com/open-formulieren/open-forms/blob/082e41707dee4516ba5dfa14f6058a684eb41442/src/openforms/submissions/form_logic.py#L211

* does not honor `trigger_from_step` -> should be based on the latest completed step in the submission
* duplicates some code in evaluate_form_logic -> put that in private utility functions
* doesn't use the data container with full view of variable values, which may be used in the logic triggers

sergei-maertens:
Fixes #1913

We have two variants of logic checks:

1. Running explicitly in the context of a form step
2. Running on the submission as a whole

The second one did not take into account recent features such as:

* variables (static, user defined)
* logic actions changing values of variables that are used in other rules' trigger
* disabling of step-submission

**TODO**

Refactor the code to avoid the repetition and properly abstract this away:

- handle logging the evaluation
- handle skipping certain logic actions (formio configuration is not relevant for flavour 2)

### DIFF
--- a/src/openforms/submissions/logic/rules.py
+++ b/src/openforms/submissions/logic/rules.py
@@ -1,0 +1,100 @@
+from typing import Iterable, List, Optional
+
+from openforms.forms.models import FormLogic, FormStep
+
+from ..models import Submission, SubmissionStep
+
+
+def _include_rule(form_steps: List[FormStep], rule: FormLogic, step_index: int) -> bool:
+    # rules that always apply
+    if not rule.trigger_from_step:
+        return True
+
+    trigger_from_index = form_steps.index(rule.trigger_from_step)
+    # if the current step is before the trigger-from step, do not include the rule
+    if step_index < trigger_from_index:
+        return False
+
+    return True
+
+
+def get_rules_to_evaluate(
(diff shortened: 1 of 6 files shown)


src/openforms/submissions/logic/rules.py
```diff
@@ -0,0 +1,163 @@
...
+    :arg on_rule_check: Optional callable taking a :class:`EvaluatedRule` instance as
+      sole argument. Useful to gather metadata about rule evaluation.
+    :returns: An iterator yielding :class:`ActionOperation` instances.
+    """
+    for rule in rules:
+        with elasticapm.capture_span(
+            "evaluate_rule",
+            span_type="app.submissions.logic",
+            labels={"ruleId": rule.pk},
+        ):
```
sergei-maertens, line 143: This should result in:

* getting timings of the evaluation of a single rule
* get some aggregated insights of how many rules are evaluated within a transaction (APM should say "x number of calls" or something like that in the UI)

SilviaAmAm (approved):

Issue #1964, opened 2022-09-02
Problems discovered while preparing demo


SilviaAmAm:
### Product versie / Product version

latest

### Omschrijf het probleem / Describe the bug

Needed for demo:

- [x] Cant upload files with multiple values selected https://github.com/open-formulieren/open-forms/pull/1965
- [x] Custom CSS classes on content field looks weird https://github.com/open-formulieren/open-forms-sdk/pull/225
![Screenshot from 2022-09-02 15-09-00](https://user-images.githubusercontent.com/19154114/188152882-141b0d21-7391-4c7b-9423-b2e88e8013d3.png)

Others:

- [x] Introspect_json_logic (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/logging/logevent.py#L500) crashes if a trigger has the value `true`. https://github.com/open-formulieren/open-forms/pull/1968
- [x] User defined variables of type float can't be given an initial value of zero https://github.com/open-formulieren/open-forms/pull/1974
- [ ] ~Something is going wrong with repeating groups in some specific cases (See below for info).~ => Weird, can't reproduce anymore :thinking: 
- [x] Logging gives wrong information for Haal centraal prefill: in tijdlijn logboekvermelding, we see "Prefillplugin "Haal Centraal" (haalcentraal) meldt: Informatie met succes opgehaald om velden te prefillen: ..." but in sentry: 
`404 Client Error: Not Found for url: https://haal-centraal-test.nl/api/ingeschrevenpersonen/123123123` https://github.com/open-formulieren/open-forms/pull/1976

### Stappen om te reproduceren / Steps to reproduce

How to reproduce problem repeating groups:
Example form: 
[problem-repeating-groups.zip](https://github.com/open-formulieren/open-forms/files/9478227/problem-repeating-groups.zip)

Components in form:
- Radio A, choices a, b
- Repeating group A with a number field. Hidden unless radio A = a
- Radio B choices b, c
- Repeating group B with a number field.  Hidden unless radio B = b

Fill in the form, select `a` in first radio=> Filling in the repeating group gives the error:
```
Uncaught TypeError: can't define array index property past the end of an array with non-writable length
    saveRow EditGrid.js:842
```

PR #1968, opened 2022-09-05
[#1964] Introspect json logic primitives

SilviaAmAm:
Part of #1964

### DIFF
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -470,7 +470,7 @@ def submission_logic_evaluated(
     initial_data: dict,
     resolved_data: dict,
 ):
-    from openforms.submissions.logic.rules import get_targeted_components
+    from openforms.submissions.logic.log_utils import get_targeted_components
 
     if not evaluated_rules:
         return
@@ -495,7 +495,7 @@ def submission_logic_evaluated(
         rule = evaluated_rule.rule
 
         trigger = rule.json_logic_trigger
-        if not isinstance(trigger, dict) and not isinstance(trigger, list):
+        if not isinstance(trigger, (dict, list)):
             # The trigger was a primitive, no need to introspect
             evaluated_rule_data = {
                 "raw_logic_expression": trigger,
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -471,6 +471,8 @@ def submission_logic_evaluated(
     initial_data: dict,
     resolved_data: dict,
 ):
+    from openforms.submissions.logic.rules import get_targeted_components
+
     if not evaluated_rules:
         return
     evaluated_rules_list = []
@@ -479,8 +481,8 @@ def submission_logic_evaluated(
     form_steps = submission_state.form_steps
 
     # Keep a mapping for each component's key for each component in the entire form.
-    # The key of the mapping is the component key, the value is a dataclasse composed of
-    # form_step and component defintion
+    # The key of the mapping is the component key, the value is a dataclass composed of
+    # form_step and component definition
     components_map = {}
     for form_step in form_steps:
         for component in form_step.form_definition.iter_components():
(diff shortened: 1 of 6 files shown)


src/openforms/submissions/logic/rules.py
```diff
@@ -161,3 +162,60 @@ def iter_evaluate_rules(
...
                     yield operation
+
+
+def get_targeted_components(
+    rule: FormLogic,
+    component_key_lookup: dict,
+    components_map: dict,
+    all_variables: Dict[str, FormVariable],
+    initial_data: dict,
+) -> List[dict]:
```
sergei-maertens, line 173: I think I'd prefer this method to also invoke `compile_action_operation`, and have every (relevant class) implement the `get_targeted_components`. I want to avoid if/else mayhem based on the logic action type, but instead delegate that to a specialized class.

sergei-maertens (changes requested):

src/openforms/submissions/logic/rules.py
```diff
@@ -161,3 +161,22 @@ def iter_evaluate_rules(
...
+
+def get_targeted_components(
+    rule: FormLogic,
+    components_map: dict,
+    all_variables: Dict[str, FormVariable],
+    initial_data: dict,
+) -> List[dict]:
+    targeted_components = []
+    for action in rule.actions:
+        action = ActionOperation.from_action(action)
```
sergei-maertens, line 174: `compile_action_operation` is used in other places as the public API

```suggestion
        action = compile_action_operation(action)
```

I'd also use a different variable name, since this changes the type of `action` from `dict` to `ActionOperation`, which I expect `mypy`/editors to complain about.

src/openforms/submissions/logic/actions.py
```diff
@@ -40,6 +42,15 @@ def apply(self, step: SubmissionStep, configuration: JSONObject) -> None:
...
         pass
 
+    def get_action_log_data(
+        self,
+        component_map: Dict[str, ComponentMeta],
+        all_variables: Dict[str, FormVariable],
+        initial_data: dict,
+    ) -> dict:
+        """Get action information to log"""
+        pass
```
sergei-maertens, line 52: base implementation does not match type hint - I think I'd opt for `return None` as default implementation and change the type hint to be `-> Optional[JSONObject]`, as the return variable must be json-serializable.

src/openforms/submissions/logic/actions.py
```diff
@@ -97,9 +148,56 @@ def apply(self, step: SubmissionStep, configuration: JSONObject) -> None:
...
+        return cls(variable=action["variable"], value=action["action"]["value"])
+
+    def get_action_log_data(
+        self,
+        component_map: Dict[str, ComponentMeta],
+        all_variables: Dict[str, FormVariable],
+        initial_data: dict,
+    ) -> dict:
+        # Check if it's a primitive value, which doesn't require introspection
+        if not isinstance(self.value, dict):
```
sergei-maertens, line 181: in the previous patch I also saw some `isinstance` checks on `list` - is that still covered by `introspect_json_logic` then?

src/openforms/logging/logevent.py
```diff
@@ -471,6 +470,8 @@ def submission_logic_evaluated(
     initial_data: dict,
     resolved_data: dict,
 ):
+    from openforms.submissions.logic.rules import get_targeted_components
```
sergei-maertens, line 473: could we maybe move that into a dedicated `openforms.submissions.logic.log_utils` module? I'd like to keep the other modules focused on effective execution.

src/openforms/logging/logevent.py
```diff
@@ -490,70 +491,37 @@ def submission_logic_evaluated(
...
-        LogicActionTypes.value: "component",
-        LogicActionTypes.property: "component",
-        LogicActionTypes.variable: "variable",
-    }
     for evaluated_rule in evaluated_rules:
-
         rule = evaluated_rule.rule
+
+        trigger = rule.json_logic_trigger
+        if not isinstance(trigger, dict) and not isinstance(trigger, list):
```
sergei-maertens, line 498:
```suggestion
        if not isinstance(trigger, (dict, list)):
```

sergei-maertens (changes requested):
I think it's a good first step to cleaning up that `logevent` function a bit :+1: There's more to gain, but that will come in due time.

src/openforms/submissions/logic/actions.py
SilviaAmAm: Mhm I think actually it's a mistake here. Because the list may have json logic inside, so it should be introspected.

sergei-maertens (approved):

Issue #1987, opened 2022-09-06
Error on any URL happens in middleware


joeribekker:
### Product versie / Product version

latest (6-9-2022)

### Omschrijf het probleem / Describe the bug

This happens but I cannot pinpoint the exact cause:
```
AttributeError: 'NoneType' object has no attribute 'namespace'
  File "django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "openforms/middleware.py", line 48, in __call__
    if request.resolver_match.namespace != app_name:
```

Happens often on `favicon.ico` but also on just `/admin`

Sentry event ID: c15cb47a05774820b54fe14779b15117

SilviaAmAm:
Discussed with sergei: we are going to change the check to `request.path.startswith(/api)`

PR #1990, opened 2022-09-06
[#1987] Replace check for api endpoints

SilviaAmAm:
I didn't write a test, because I am not really sure what the problem here exactly is / how to reproduce it :disappointed:


sergei-maertens:
> I didn't write a test, because I am not really sure what the problem here exactly is / how to reproduce it disappointed

I think there's some funky stuff going on with viewsets/routers in DRF as well. The existing test is fine for this.

sergei-maertens (approved):

PR #2006, opened 2022-09-08
[#1944] Special header for SDK to navigate between steps

SilviaAmAm:
Part of #1944

### DIFF
--- a/src/openforms/middleware.py
+++ b/src/openforms/middleware.py
@@ -6,6 +6,7 @@ from openforms.config.models import GlobalConfiguration
 
 SESSION_EXPIRES_IN_HEADER = "X-Session-Expires-In"
 CSRF_TOKEN_HEADER_NAME = "X-CSRFToken"
+CAN_NAVIGATE_BETWEEN_STEPS_HEADER_NAME = "X-Is-Staff"
 
 
 class SessionTimeoutMiddleware:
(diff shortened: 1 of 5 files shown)


src/openforms/middleware.py
```diff
@@ -6,6 +6,7 @@
 
 SESSION_EXPIRES_IN_HEADER = "X-Session-Expires-In"
 CSRF_TOKEN_HEADER_NAME = "X-CSRFToken"
+CAN_NAVIGATE_BETWEEN_STEPS_HEADER_NAME = "X-Is-Staff"
```
sergei-maertens, line 9: I'd rename this to `IS_STAFF_HEADER_NAME` and keep it generic.

src/openforms/middleware.py
```diff
@@ -49,3 +50,28 @@ def __call__(self, request):
...
+
+        # Only add the header if it's an api endpoint
+        if not request.path.startswith("/api"):
+            return response
+
+        response[CAN_NAVIGATE_BETWEEN_STEPS_HEADER_NAME] = "false"
+
+        user = request.user
+        if user.is_staff and user.has_perm("forms.change_form"):
+            response[CAN_NAVIGATE_BETWEEN_STEPS_HEADER_NAME] = "true"
```
sergei-maertens, line 75: I forgot about the permission check - maybe we should rename it to `X-Is-Form-Designer` rather than `X-Is-Staff` after all :thinking:

sergei-maertens (changes requested):

src/openforms/middleware.py
SilviaAmAm: Yes I think `X-Is-Form-Designer` fits better

PR #2026, opened 2022-09-12
Enable more complex date range validation

### LINKED ISSUE #1878: Improve date validation configuration
### Thema / Theme

Other

DH taiga user story 117, issue 50
DH Taiga user story 168, issue 161

What they basically want is, in the Min/Max fields: {{ field1 }} + 14

Currently, the date field has 2 fields "max date" and "min date":
datefields

It is not possible here to specify that a date should not be in the future or that it should be between today and in 2 weeks time. It is possible to do this in the logic tab, but this makes it difficult to maintain.

The possibilities of validation that DH would like are:
- Specific max/min date range (already present)
- Max/min date range + or - X number of days (?)
- In the future (including and excluding today)
- In the past (including and excluding today)
- Larger than the date in another field +/- X days
- Smaller than the date in another field +/- X days

_No response_

_No response_

sergei-maertens:
Fixes #1878

Screencast with form designer UX: https://watch.screencastify.com/v/BXiFyRpkZ0CMk4D6wF3f

**Tasks**

- [x] Set up form designer UX
- [x] Implement processing config into `datePicker.minDate` and `datePicker.maxDate`
- [x] Translations


src/openforms/js/components/form/date.js
```diff
@@ -10,6 +10,230 @@ import {
 
 const DateTimeField = Formio.Components.components.datetime;
 
+const MESSAGES = {
```
sergei-maertens, line 13: TODO: add translation to formio translations dict

src/openforms/scss/components/builder/_columns.scss
```diff
@@ -47,3 +47,14 @@ $bootstrap-ratio-desktop: math.div(12, $grid-columns-desktop);
...
+[ref='editForm'] {
+  .formio-component-columns.row {
+    margin-left: -7.5px;
+    margin-right: -7.5px;
+  }
+  .columns > * {
+    margin-left: 7.5px;
+    margin-right: 7.5px;
+  }
+}
```
sergei-maertens, line 60: Using columns otherwise looked ugly with broken padding/margins, as they assumed bootstrap styles.

src/openforms/formio/dynamic_config/apps.py
```diff
@@ -0,0 +1,9 @@
+from django.apps import AppConfig
+
+
+class FormioDynamicConfigApp(AppConfig):
```
joeribekker, line 4: Nu snap ik wat je gisteren bedoelde met laat dat register patroon in global filetypes maar zitten :)

src/openforms/formio/dynamic_config/date.py
```diff
@@ -0,0 +1,143 @@
...
+        if base_value is None:
+            return None
+
+        assert (
+            base_value.tzinfo is not None
+        ), "Expected the input variable to be timezone aware!"
+        base_date = timezone.localtime(value=base_value).date()
+
+        delta = relativedelta(
+            years=cast(int, glom(config, "delta.years", default=None) or 0),
```
joeribekker, line 132: Why not `default=0` without the `or`?

src/openforms/formio/dynamic_config/date.py
```diff
@@ -0,0 +1,143 @@
...
+                glom(component, f"openForms.{key}", default=None),
+            )
+            if config is None:
+                continue
+
+            if (mode := config.get("mode")) is None:
+                continue
+
+            # formio has set the datePicker.{key} value
+            if mode == "fixedValue":
```
joeribekker, line 76: I see a bunch of.magic strings: fixed value, future, past, add, subtract, relativeToVariable, includeToday. Maybe define constants (like the now-variable).

src/openforms/formio/dynamic_config/date.py
```diff
@@ -0,0 +1,143 @@
...
+        base_value = cast(
+            Optional[datetime],
+            glom(data, config["variable"], default=None),
+        )
+        # can't do calculations on values that don't exist
+        if base_value is None:
+            return None
+
+        assert (
+            base_value.tzinfo is not None
```
joeribekker, line 127: But, date and time components (and the values people put in) are not time aware. Do we/formio convert them somewhere else?

src/openforms/formio/dynamic_config/registry.py
```diff
@@ -0,0 +1,40 @@
...
+
+from ..typing import Component
+
+
+class BasePlugin(AbstractBasePlugin):
+    is_enabled: bool = True
+
+    @staticmethod
+    def mutate(component: Component, data: DataMapping) -> None:  # pragma: nocover
+        ...
```
joeribekker, line 15: If this is valid I learned something today. Same as `pass` I guess.

src/openforms/formio/dynamic_config/tests/test_date_component.py
```diff
@@ -0,0 +1,265 @@
...
+        return new_configuration["components"][0]
+
+    def test_legacy_configuration_still_works(self):
+        # legacy configuration = without the openForms.minDate keys etc.
+        component = {
+            "type": "date",
+            "key": "aDate",
+            "datePicker": {
+                "minDate": None,
+                "maxDate": "2022-09-08T00:00:00+02:00",
```
joeribekker, line 43: I assume you checked that formio uses this (timezone aware) date and not regular dates? It's almost worth writing a test to check that with a JS test.

joeribekker (changes requested):

src/openforms/formio/dynamic_config/apps.py
sergei-maertens: ja, de boilerplate heeft wel even wat voeten in de aarde :grimacing:

src/openforms/formio/dynamic_config/date.py
sergei-maertens: Hrm, the `now` variable constant is done because it's external to this module and importing the actual variable from a different app felt too much like violating separation of concerns.

The "magic strings" are type-checked in editors (and with mypy once we get that set up) through the type declaration in the `TypedDict` at https://github.com/open-formulieren/open-forms/pull/2026/files#diff-19dff21508a2142d75a9ac65d2293e0f4cff5acc1506d0544ec04c261ef558dfR26 - this effectively specifies the allowed collection of values for `mode`.

I just tried out defining it as a constant and using that constant in the type declaration like `mode: Literal[MODE_FIXED_VALUE, "future", "past", "relativeToVariable"]`, but that's not valid python - they must all be literals.

src/openforms/formio/dynamic_config/date.py
sergei-maertens: Our `date` component inherits from Formio's `datetime` component (which out of the box has options to disable the date or time aspect), and that's why it internally stores and communicates everything using full datetime strings.

Dates also inherently lack timezone information - Sept 13 at 1am midnight is Sept 13 in NL, but Sept 12 in UTC, so you need that timezone information to be able to figure out the correct date in the local timezone of the user.

src/openforms/formio/dynamic_config/date.py
sergei-maertens: If the dict has `{"years": None}`, this results in `None` being passed to `relativedelta`, which then crashes. This normalizes missing and empty variables both alike to the int `0`.

src/openforms/formio/dynamic_config/registry.py
sergei-maertens: yeah it is, often used for pure type-hinting code where you overload a function like this: 

```py
from typing import overload

@overload
def foo(arg: str) -> str: ...

@overload
def foo(arg: int) -> int: ...

def foo(arg: int | str) -> int | str:
    # actual implementation
```

it indicates that string input results in string output, int input results in int output, while without the overloads you could theoretically get one type as input and return the other as output.

src/openforms/formio/dynamic_config/tests/test_date_component.py
sergei-maertens: yep, this is what ends up in the JSON when you pick an explicit min/max date - a full datetime string.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -289,3 +293,57 @@ def __str__(self):
...
+        # can't see any type information...
+        if not self.form_variable_id:
+            return self.value
+
+        # we expect JSON types to have been properly stored (and thus not as string!)
+        data_type = self.form_variable.data_type
+        if data_type in (
+            FormVariableDataTypes.string,
+            FormVariableDataTypes.boolean,
+            FormVariableDataTypes.object,
```
sergei-maertens, line 319: I'm not sure yet how we'll deal with recursion if there's a stringified datetime inside of an object, but let's tackle that when we actually run into such a situation.

src/openforms/formio/dynamic_config/date.py
joeribekker: check.

joeribekker (approved):

PR #2030, opened 2022-09-13
:zap: Optimizing API endpoints

sergei-maertens:
A combination of analysis using Elastic APM and django-silk to figure out the origin of excessive queries, followed by some workarounds to remove the queries that are not absolutely needed.

* Logic check endpoint with test case used for profiling: 18 -> 11 queries
* Renderer with variables node: 8 -> 3 queries

Let's get this merged and tested on our test environment, so we can check again with APM what the new situation is (and run the profiling benchmark again).

### DIFF
--- a/src/openforms/conf/dev.py
+++ b/src/openforms/conf/dev.py
@@ -92,7 +92,7 @@ CACHES.update(
 # Library settings
 #
 
-ELASTIC_APM["DEBUG"] = True
+ELASTIC_APM["DEBUG"] = config("DISABLE_APM_IN_DEV", default=True)
 
 # Django debug toolbar
 INSTALLED_APPS += ["debug_toolbar", "ddt_api_calls"]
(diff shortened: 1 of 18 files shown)


src/openforms/conf/dev.py
```diff
@@ -92,7 +92,7 @@
 # Library settings
 #
 
-ELASTIC_APM["DEBUG"] = True
+# ELASTIC_APM["DEBUG"] = True
```
sergei-maertens, line 95: self-reminder: re-enable DEBUG mode! (and maybe convert this to an envvar)

src/openforms/formio/dynamic_config/date.py
```diff
@@ -113,13 +113,16 @@ def calculate_delta(
...
         )
-        # can't do calculations on values that don't exist
-        if base_value is None:
+        # can't do calculations on values that don't exist or are empty
+        if not base_value:
             return None
 
+        # if it's not empty-ish, it's a datetime
+        base_value = cast(datetime, base_value)
+
```
sergei-maertens, line 125: Had a failing test because the variable default value was a string for type datetime (in case no value is set yet), which can happen, so this is now handled and caught too.

joeribekker (approved):
Lijkt me correcte vervangingen voor queries maar daadwerkelijke performance winst moet blijken uit meting before en after.

Issue #1922, opened 2022-08-25
Performance profiling of 2.0.0 release


sergei-maertens:
We're seeing some performance regressions on bigger forms in the logic evaluation. Before releasing 2.0.0, we need to do extensive testing and fixing.

* Ooievaarspas form on our own test environment is displaying these issues (~ 10 steps, ~ 100 logic rules)
* Look at APM to get insight in queries that can be optimized
* Add additional APM decorators in various places for more fine-grained profiling/execution times

**Tasks**

- [x] Optimize saving form variables -> 3 queries per variable it seems at the moment, see APM link on Slack (maykin test, form UUID: 61ffd593, Aanvraag stadspas)

PR #2036, opened 2022-09-15
:zap: [#1922] -- Optimize the bulk form variables endpoints

sergei-maertens:
Related to #1922

Performance improvement variables bulk update, for a case with 30 component variables and 20 user defined variables:

* before: 84 queries
* after: 5 queries

**Variables list**

* Removed unnecessary prefetch/select_related from variables-list endpoint
* Added select_related on variables-list endpoint for variable form and form-definition
* Variables-list endpoint now has a fixed number of queries: 2 (+ session/middleware queries etc.)

**Variables bulk update**

Every form variable was being validated individually, meaning that a FK lookup was performed for the form and the (optional) form definition, giving a complexity somewhere between `O(n)` and `O(2*n)` with n the number of form variables.

The form validation/lookup is not needed in the first place, since the endpoint is scoped within a form detail endpoint. The form definition validation is a bit more complex, but again, multiple variables can belong to the same form-definition, so doing this query multiple times is wasted performance.

The performance patch collects the form and it's related form definitions upfront and passes them via the serializer context down to a custom field which replaces the DB lookups with context lookups. This results in a fixed number of queries: `O(5)`.

### DIFF
--- a/src/openforms/api/fields.py
+++ b/src/openforms/api/fields.py
@@ -1,0 +1,31 @@
+from rest_framework import relations
+
+
+class RelatedFieldFromContext(relations.HyperlinkedRelatedField):
+    """
+    Look up the object in the serializer context.
+    """
+
+    def __init__(self, context_name="objects", *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # key to use to look up the object in the context, which is a dict mapping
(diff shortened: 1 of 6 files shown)


joeribekker (approved):
Looks great. I had some doubt on resetting the validator because having this done elsewhere is not enforced. The DB should have catch it

PR #2040, opened 2022-09-15
Variables in email registration

sergei-maertens:
#1579

* User-defined variables are already included via Silvia's earlier implementation
* Added support and help-text for variables in the e-mail subject.

Note that I chose the `vars` namespace for use in the template engine - `variables` is maybe better but a bit too long? Either way, let me know if you can think of a better namespace.


joeribekker:
Refinement: Closed with Sergei's permission.

PR #2047, opened 2022-09-16
Squash database migrations

sergei-maertens:
In 2.1.0 we can remove the original migration files and convert the squashed migrations into normal migration.


joeribekker:
I validated by checking out 1.1.5, migrate, switch to this branch and migrate further.
Also, did a dropdb, and migrated from scratch

both work.

PR #2065, opened 2022-09-22
Implement upgrade path checks

### LINKED ISSUE #1979: Track version number and block upgrades until minimum version is reached
From Slack discussion about requiring to upgrade to 1.1.x before going to 2.0

* Track the current version somewhere (database, maybe part of a solo-model that is updated through every new version with a data migration?)
* Implement a system check that checks the current code version against the current DB version and error/block if the conditions aren't met
* Define in code/config what the minimum previous version must be for an upgrade (following semver?) e.g.:
  
  ```py
  UPGRADE_PATHS = {
      "1.1": "1.0.3",
      "2.0": "1.1.5",
      "3.0": "2.0",
  }
  ```
  This could also simply be a file in the root of the project though.
* Latest/dev/master should not be affected by this

sergei-maertens:
Fixes #1979 

- [x] Record the "current" version in the database (automatically on every install/upgrade)
- [x] Implement checks based on from_version -> to_version
    - [x] Check supported version ranges
    - [x] Verify/run certain management commands
    - [x] Verify/run certain 'arbitrary' scripts
- [x] Check that latest/dev still work properly
- [x] Tests!

Thoughts:

- Consider using `openforms.__version__` instead of relying on `RELEASE` envvar?

-> not doing that, since custom builds/forks will probably forget to update this variable and `RELEASE` is a bit more flexible/reliable for that.


SilviaAmAm (approved):

sergei-maertens:
New model is backported in 922e1460374d2c36252c60635a1c8d2bb1a52410 and c5551e50c336adfc959603753ce5422507c25de5

PR #2088, opened 2022-09-28
Update 2.0.0 changelog

sergei-maertens:
* Added changelog entries from 1.0.14 and 1.1.6
* Fixed mistake in version range (1.0.16 does not exist, 1.0.14 does)
* Added 2.0.0-beta.0 as a supported version range to update from
* Moved 1.0.x links to "unsupported versions" section

Tomorrow (Sept. 29th) we will release the new stable versions, 2.0.0 will follow next week hopefully.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -265,6 +265,14 @@ double check with the list of breaking changes in mind.
 * Improved documentation on embedding the SDK
 * [#921] Added decision tree docs
 
+1.1.7 (2022-10-04)
+==================
+
+1.1.6 was broken due to a bad merge conflict resolution.
+
+* [#2095] Fixed accidentally removing the OF layer on top of Formio
+* [#1871] Ensure that fields hidden in frontend don't end up in registration e-mails
+
 1.1.6 (2022-09-29)
 ==================
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-2.0.0 "Règâh" (2022-??-??) - in testing
+2.0.0 "Règâh" (2022-10-??) - in testing
 =======================================
 
 *The symbol of The Hague is the stork, a majestic bird, which is somewhat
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -268,6 +268,57 @@ double check with the list of breaking changes in mind.
...
+* [#1832] Prevent and handle location service rate limit errors
+* [#1960] Ensure design tokens override default style
+* [#1957] Fixed not being able to manually retry errored submission registrations having
+  exceeded the retry limit
+* [#1867] Added more StUF-ZDS/ZGW registration fields.
+* Added missing translation for max files
+* [#2011] Worked around thread-safety issue when configuring Ogone merchants in the admin
+* [#2066] Re-added key validation in form builder
+
+**Upgrade preprations**
```
joeribekker, line 287: No need to point this out here or make a separate heading for it. It's relevant for 2.0.0 where this is mentioned

CHANGELOG.rst
```diff
@@ -268,6 +268,57 @@ double check with the list of breaking changes in mind.
...
+
+The following additions are required to prepare upgrading to 2.0.0. Please see the
+release notes of 2.0.0 on how to use them.
+
+* [#2055] Added management command to check for invalid keys
+* [#1979] Added model to track currently deployed version
+
+.. note:: This is the FINAL 1.0.x release - support for this version has now ended. We
+   recommend upgrading to 2.0.x.
```
joeribekker, line 320:
```suggestion
   recommend upgrading to the latest major version.
```

CHANGELOG.rst
```diff
@@ -268,6 +268,57 @@ double check with the list of breaking changes in mind.
...
+* [#1832] Prevent and handle location service rate limit errors
+* [#1960] Ensure design tokens override default style
+* [#1957] Fixed not being able to manually retry errored submission registrations having
+  exceeded the retry limit
+* [#1867] Added more StUF-ZDS/ZGW registration fields.
+* Added missing translation for max files
+* [#2011] Worked around thread-safety issue when configuring Ogone merchants in the admin
+* [#2066] Re-added key validation in form builder
+
+**Upgrade preprations**
```
joeribekker, line 311: Same

CHANGELOG.rst
```diff
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-2.0.0 "Règâh" (2022-??-??) - in testing
+2.0.0 "Règâh" (2022-10-??) - in testing
```
joeribekker, line 5: Testing can be gone right?

Also, sorry no possibility to comment on that line, I feel like there should be a summary of the upgrade procedure:

1. Update to 1.1.6
2. Run management command 1 and fix issues
3. Run management command 2 and fix issues
4. If all issues solved, upgrade to 2.0.0

joeribekker (changes requested):

CHANGELOG.rst
sergei-maertens: I'll remove it :+1:

CHANGELOG.rst
sergei-maertens: I'll remove it :+1:

CHANGELOG.rst
sergei-maertens: I thought we had covered that with the "manual interventions required", but probably we can reword that a bit.

W/r to the "in testing" - as long as 2.0.0 final is not out yet, it's still in testing. We still need to fix a couple other bugs and then re-test before we can release.

PR #2094, opened 2022-09-29
[#2093] Fix wrong action structure after conversion

### LINKED ISSUE #2093: Util function `advanced_formio_logic_to_backend_logic` creates invalid logic rules

2.0.0-beta.0

The JSON of the actions created during the conversion from advanced frontend logic to backend logic is:
```json
{
    "property": {
       "value": "validate",
       "type": "json"
     },
     "state": {"required": true}
}
```
While it should be:

```json
{
  "action": {
    "property": {
       "value": "validate",
       "type": "json"
     },
     "state": {"required": true}
  },
  "component": "bla"
}
```

SilviaAmAm:
Fixes #2093


src/openforms/forms/utils.py
```diff
@@ -320,7 +321,9 @@ def parse_trigger(trigger_info: JSONObject) -> Optional[JSONObject]:
         return trigger
 
 
-def parse_actions(actions: List[JSONObject]) -> List[JSONObject]:
+def parse_actions(
+    actions: List[JSONObject], component_key: str = None
```
SilviaAmAm, line 325: I left this with component_key=None so we don't need to update the detection script

sergei-maertens (approved):

Issue #2042, opened 2022-09-15
Performance optimizations


sergei-maertens:
* [x] `GlobalConfiguration.get_solo()` is called a lot and hits the redis cache often  (PR: #2044)
    * Use local-memory cache wrapper around redis cache?
    * Callers like: `openforms.plugins.plugin.AbstractBasePlugin.is_enabled` are expensive when looping over plugin registries
* [x] Form logic rules bulk endpoints -> apply same performance as with form variables
* [x] Throttling (DRF) checks hit cache often as well (don't see an immediate workaround)
* [x] PUT form variables does _a lot_ of `openforms.formio.utils.get_component` calls

PR #2109, opened 2022-10-05
Reduce the amount of (unnecessary) iter_components calls

sergei-maertens:
Related to #2042

Performance test run results on a simple form. A more complex form had hundreds of `iter_components` calls

| **Metric**              | Before | After |
|-------------------------|--------|-------|
| Avg latency (ms)        | 479    | 278   |
| 90 percentile (ms)      | 886    | 420   |
| # iter_components calls | 45     | 6     |

Note that I attribute the speedup mostly to the reduced APM overhead (implicit result by removing `iter_components` calls)

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -5,12 +5,12 @@ from django.utils.functional import empty
 
 import elasticapm
 
-from openforms.formio.service import get_dynamic_configuration, inject_variables
-from openforms.formio.utils import (
-    get_component_empty_value,
-    is_visible_in_frontend,
-    iter_components,
+from openforms.formio.service import (
+    FormioConfigurationWrapper,
+    get_dynamic_configuration,
+    inject_variables,
 )
+from openforms.formio.utils import get_component_empty_value, is_visible_in_frontend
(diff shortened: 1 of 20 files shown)


src/openforms/submissions/form_logic.py
```diff
@@ -119,10 +114,10 @@ def evaluate_form_logic(
 
     # we need to apply the context-specific configurations before we can apply
     # mutations based on logic, which is then in turn passed to the serializer(s)
-    # TODO: refactor this to rely on variables state, which will move this down to
-    configuration = deepcopy(configuration)
-    configuration = get_dynamic_configuration(
-        configuration,
+    # TODO: refactor this to rely on variables state
+    # config_wrapper = FormioConfigurationWrapper(deepcopy(configuration))
```
SilviaAmAm, line 118: Was this meant to remain as a comment?

SilviaAmAm (approved):

src/openforms/submissions/form_logic.py
sergei-maertens: nope, I'll get rid of it! :fire:

PR #2112, opened 2022-10-05
[#2004] Post processing hook for CSRF token

### LINKED ISSUE #2004: Automatically document X-CSRFToken header through DRF spectacular extension
### Thema / Theme

API

Extracted from this comment: https://github.com/open-formulieren/open-forms/pull/1958#discussion_r965732245

We should automatically document the required `X-CSRFToken` header in endpoints with unsafe methods instead of adding 

```
parameters=[
            OpenApiParameter(
                CSRF_TOKEN_HEADER_NAME,
                OpenApiTypes.STR,
                location=OpenApiParameter.HEADER,
                required=True,
            )
        ],
```
to every request.

We can/should automatically document this through a drf-spectacular extension (in a different PR).

Examples in the docs: https://drf-spectacular.readthedocs.io/en/latest/blueprints.html

_No response_

_No response_

SilviaAmAm:
Fixes #2004

### DIFF
--- a/src/openforms/api/drf_spectacular/hooks.py
+++ b/src/openforms/api/drf_spectacular/hooks.py
@@ -1,6 +1,8 @@
 from django.utils.translation import gettext_lazy as _
 
+from drf_spectacular.plumbing import build_parameter_type
 from drf_spectacular.settings import spectacular_settings
+from drf_spectacular.utils import OpenApiParameter
 
 from openforms.middleware import (
     CSRF_TOKEN_HEADER_NAME,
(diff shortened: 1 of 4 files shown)


src/openforms/api/drf_spectacular/hooks.py
```diff
@@ -65,3 +67,36 @@ def add_middleware_headers(result, generator, request, public):
     )
 
     return result
+
+
+CSRF_TOKEN_PARAMETER = build_parameter_type(
+    name=CSRF_TOKEN_HEADER_NAME,
+    schema={"type": "string"},
+    location=OpenApiParameter.HEADER,
+    required=True,
```
sergei-maertens, line 76: I think it should only be required if session auth is the only authentication method for the endpoint. If token auth is an option, then the parameter is not required :grimacing:

src/openforms/api/drf_spectacular/hooks.py
```diff
@@ -65,3 +67,36 @@ def add_middleware_headers(result, generator, request, public):
...
+                continue
+
+            for item in operation["security"]:
+                if "cookieAuth" in item:
+                    break
+            else:
+                continue
+
+            if "parameters" not in operation:
+                operation["parameters"] = []
```
sergei-maertens, line 99:
```suggestion
            operation.setdefault("parameters", [])
```

sergei-maertens (changes requested):

src/openforms/api/drf_spectacular/hooks.py
```diff
@@ -65,3 +67,36 @@ def add_middleware_headers(result, generator, request, public):
...
+        for operation_method, operation in path.items():
+            if operation_method not in ["post", "put", "patch", "delete"]:
+                continue
+
+            if "security" not in operation:
+                continue
+
+            for item in operation["security"]:
+                if "cookieAuth" in item:
+                    break
```
sergei-maertens, line 94: I think the detection can be a bit more generic and not necessarily rely on the name, but rather if `type == apiKey` and `in == cookie`

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2128, opened 2022-10-06
Columns options in form builder

### LINKED ISSUE #2113: Align columns
### Thema / Theme

Frontend

Den Haag Taiga user story 280, issue 274

The columns alignment doesn't look right:
kolom uitlijning

This happens also in the form editor:
kolom uitlijning in de editor

_No response_

_No response_

sergei-maertens:
Fixes #2113 

* The (nested) boxes for columns now stretch to look more like a grid
* The labels of components are now limited to a single line (only in the builder!) and truncated with an ellipsis

image

image

### DIFF
--- a/src/openforms/scss/components/builder/_columns.scss
+++ b/src/openforms/scss/components/builder/_columns.scss
@@ -38,7 +38,7 @@ $bootstrap-ratio-desktop: math.div(12, $grid-columns-desktop);
   }
 
   .column {
-    $col-sizes: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11;
+    $col-sizes: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
 
     @each $col-size in $col-sizes {
       &--span-#{$col-size} {
--- a/src/openforms/scss/components/builder/_columns.scss
+++ b/src/openforms/scss/components/builder/_columns.scss
@@ -24,6 +24,10 @@ $bootstrap-ratio-desktop: math.div(12, $grid-columns-desktop);
   display: flex;
   width: 100%;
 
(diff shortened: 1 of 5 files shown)


src/openforms/scss/components/builder/_columns.scss
```diff
@@ -38,8 +42,17 @@ $bootstrap-ratio-desktop: math.div(12, $grid-columns-desktop);
...
+    flex-direction: column;
+
+    padding-left: 3px;
+    padding-right: 3px;
 
+    & > * {
+      flex-grow: 1;
+    }
+
+    $col-sizes: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11;
```
SilviaAmAm, line 55: Since you added `{label: '12/12 (100%)', value: '12'},` to the column component configuration, do you need to also add the `12` to the column sizes?

SilviaAmAm (changes requested):
When I checkout the branch locally I see this:

This branch:
Screenshot from 2022-10-07 11-14-00

Master:
Screenshot from 2022-10-07 11-15-07

It seems that the width of the page is still a problem? (I tried both firefox and chromium, and updated the form to not have the old column configuration :thinking: )

src/openforms/scss/components/builder/_columns.scss
sergei-maertens: ah yeah, did it in the SDK but forgot to do so here :facepalm:

sergei-maertens:
@SilviaAmAm the `min-width: 0` and `flex-grow: 1` combination seems to fix it. Of course flexbox also has weird issues :facepalm:

SilviaAmAm (approved):

PR #2144, opened 2022-10-11
[#2117] Refactored create_submission_export() to support differences in exported fields

### LINKED ISSUE #2117: Exporting multiple forms causes 500 with difference in fields

latest

Data (headers) from a form can be different between entries of the same form - this could be caused by steps being flagged as invalid which excludes the fields from this step in the data.

The resulting data (and the headers) of the fields are not identical which causes a issues when exporting results and causes a 500 like sentry: 317224

1. Create a form where a step is marked non applicable
2. Submit the form twice: once with and once without the step
3. Export both results 

The form with the most headers should be used to create the csv, json etc. 
or
The steps marked as non applicable should always be included in the data of all results.

Bartvaderkin:
This also supports combining Submission data from multiple Forms into one export

Fixes #2117

### DIFF
--- a/src/openforms/submissions/exports.py
+++ b/src/openforms/submissions/exports.py
@@ -2,6 +2,7 @@ import dataclasses
 from typing import Iterator
 
 from django.http import HttpResponse
+from django.utils.datastructures import OrderedSet
 from django.utils.timezone import make_naive
 
 import tablib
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/exports.py
```diff
@@ -43,32 +44,54 @@ def create_submission_export(queryset: SubmissionQuerySet) -> tablib.Dataset:
...
         )
         submission_data = [
             submission.form.admin_name,
             inzending_datum,
-            *[data_node.value for data_node in iter_submission_data_nodes(submission)],
+            # use blank for missing values
+            *(row.get(key, "") for key in seen_keys),
         ]
         data.append(submission_data)
+
```
sergei-maertens, line 94: While typing this out, I'm also wondering if the problem is not "simpler" that there's a bug with hidden (layout) components not showing for `export` RenderMode in `iter_submission_data_nodes` - as the initial bug report was about different submissions for the same form, so they should all have the same keys and all keys in the form (visible or hidden) should be exported always, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/rendering/conf.py#L14 Perhaps the approach shouldn't be to grab the keys/headers from the first submission, but rather from the entire form of the (first) submission.

---

Thoughts before I realized the above paragraph

In general this approach looks good to me, I'm a bit concerned about memory usage for large datasets though - we're storing all the data of the entire queryset in memory in the `rows` variable before even initializing a tablib dataset, in addition to already having the entire result set loaded in the submission queryset cache. Tablib in itself already isn't the greatest for memory usage (all data is loaded into memory rather than being able to stream/chunk it), so paying this penalty twice/thrice seems like something to avoid - especially if we need to replace tablib in the future.

With that context in mind...

* Can we leverage `queryset.iterator()` so we get memory-efficient (and server-side cursors!) processing of the submissions?
* Perhaps we could set up the tablib dataset all the way at the beginning, and mutate `dataset.headers` on the fly when we detect a new key?
* Then every iteration of the loop over the `queryset.iterator` can add the tablib data row with the correct dimensions (after the first iteration, you should only have to add 'padding' on the right to match the length of `dataset.headers`, as this will never shrink)

src/openforms/submissions/tests/test_exports.py
```diff
@@ -167,3 +172,112 @@ def test_user_defined_variables_in_export(self):
                 "Some value",
             ),
         )
+
+    @freeze_time("2022-05-09T13:00:00Z")
+    def test_submission_export_with_mixed_fields_and_forms(self):
```
sergei-maertens, line 177: Can you add a regression test that reproduces the reported issue _without_ mixed forms? That's the case that primarily needs to be solved. I'm a bit on the fence of the practicality of exporting mixed forms - this requires a lot of upfront knowledge to the importing party and they can't tell if a key belongs to the form in the row or not based on whether the value is empty or not.

sergei-maertens (changes requested):

Bartvaderkin:
@sergei-maertens I made a new PR with just the regression test and a fix in the rendering: #2157

sergei-maertens:
> @sergei-maertens I made a new PR with just the regression test and a fix in the rendering: #2157

Thanks - it's nice to see that certain limitations have been lifted here with this approach, though I still doubt the usefulness of this. So for now I'll close this PR as #2157 fixed the actual issue organizations were running into, and we'll see if the request pops up in the future to change these exports.

PR #2151, opened 2022-10-11
[#2050] Removed SubmissionFileAttachment.form_key field and using variables instead

### LINKED ISSUE #2050: Remove SubmissionFileAttachment.form_key
https://github.com/open-formulieren/open-forms/blob/1061d18c75c5f57d44b6670b56f72f196b959eef/src/openforms/submissions/models/submission_files.py#L131

There is an open TODO pointed out in #2048, but it looks like it impacts the codebase too much for easy removal (and it conflicts with the squashed migrations in #2047), so we're making it a separate issue.

Bartvaderkin:
Fixes #2050


src/openforms/submissions/migrations/0061_backfill_submissionfileattachment_variable.py
```diff
@@ -0,0 +1,55 @@
...
+def backfill_attachment_submission_variables_from_form_key(apps, schema_editor):
+    SubmissionStep = apps.get_model("submissions", "SubmissionStep")
+    SubmissionValueVariable = apps.get_model("submissions", "SubmissionValueVariable")
+
+    qs = SubmissionStep.objects.annotate(Count("attachments")).filter(
+        attachments__count__gt=0
+    )
+    qs = qs.only("id").prefetch_related("attachments")
+
+    for submission_step in qs.iterator():
```
sergei-maertens, line 16: `qs.iterator()` discards the prefetch cache, so I'd either swap it around an loop over the attachments + do a `select_related` on the submission step, or drop the `prefetch_related` call.

src/openforms/submissions/tests/factories.py
```diff
@@ -256,29 +255,32 @@ class Meta:
     @classmethod
     def create(
         cls,
+        form_key: str = None,
```
sergei-maertens, line 258:
```suggestion
        form_key: Optional[str] = None,
```
or use the empty string as default

sergei-maertens (changes requested):
Tiny remarks, otherwise this looks neat and tidy!

sergei-maertens (approved):

PR #2157, opened 2022-10-12
[#2117] Fixed issue in rendering non-submitted/non-applicable steps

### LINKED ISSUE #2117: Exporting multiple forms causes 500 with difference in fields

Bartvaderkin:
Fixes #2117

Different approach based on feedback in #2144

### DIFF
--- a/src/openforms/submissions/rendering/renderer.py
+++ b/src/openforms/submissions/rendering/renderer.py
@@ -50,7 +50,7 @@ class Renderer:
         Return the submission steps in the correct order.
         """
         execution_state = self.submission.load_execution_state()
-        return [step for step in execution_state.submission_steps if step.pk]
+        return execution_state.submission_steps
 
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):

sergei-maertens:
Backported in a84dfbd5696d898a5ddf6aa01eb3d042d96440fc

PR #2158, opened 2022-10-12
[#1899] Error readonly formatted fields

### LINKED ISSUE #1899: Sending forms with BSN-prefill in read-only fields for birthday and postcode fails with 'Call did not validate on the backend'

1.1.5

When trying to send a form where the data has been pre-filled w/that of a BSN test account, sending the form fails with a notice 'Call did not validate on the backend'. Taking a look in the dev-tools, shows that it judges the fields for postcode and birthdate to be incorrect: **invalidPrefilledFields":["Geboortedatum","Postcode"]**

The fields don't show validation-errors in the form itself:
image

**Edit 14-09: The problem disappears when disabling read-only on these fields**_

1. Open the form https://haarlemmermeer.test.open-formulieren.nl/prefill-test-bsn
2. Use 900174341 or 900174316 for a BSN
3. Fill in e-mail and telephone number
4. Try to send the form

Form sends without warning and shows the confirmation page

### Screen resolution

_No response_

### Device

Desktop

### OS

Windows

### Browser

Chrome, Firefox

SilviaAmAm:
Fixes #1899

### DIFF
--- a/src/openforms/formio/normalization.py
+++ b/src/openforms/formio/normalization.py
@@ -1,5 +1,6 @@
 import logging
 from abc import abstractmethod
+from datetime import date, datetime
 from typing import Any
 
 from openforms.plugins.plugin import AbstractBasePlugin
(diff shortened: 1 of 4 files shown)


sergei-maertens:
As discussed on Slack - in `master` (with variables), this check is moved to the `SubmissionStepSerializer` validation since the prefill data is available from the variables in that situation. There will be a separate PR for the patch on `master` as the code diverges too much between stable/1.1.x and master.

sergei-maertens (approved):

PR #2200, opened 2022-10-17
[#2062] Moved the "Print this page" click handler to new public.js exernal javascript bundle to satisfy CSP

### LINKED ISSUE #2062: "Pagina afdrukken" blocked by CSP

latest

When trying to print a page CSP is triggered preventing the window.print() command to execute.

```json

{
  "csp-report": {
    "document-uri": "https://open-formulieren-test.domein.nl/form/",
    "referrer": "",
    "violated-directive": "script-src-elem",
    "effective-directive": "script-src-elem",
    "original-policy": "default-src 'self'; img-src 'self' data: https://service.domain.nl/; style-src 'self' 'nonce-3nRQ5zYUpD131Zux7qupVw=='; script-src 'self' 'nonce-3nRQ5zYUpD131Zux7qupVw=='; frame-ancestors 'self'; base-uri 'self'; frame-src 'self'; report-uri /csp/report/",
    "disposition": "enforce",
    "blocked-uri": "inline",
    "line-number": 1,
    "source-file": "https://open-formulieren-test.domein.nl/form/",
    "status-code": 200,
    "script-sample": ""
  }
}

```

Click "Pagina Afdrukken" from the a11y toolbar and watch console print the CSP error.

The browser should initiate a print screen to print the current page.

### Screen resolution

_No response_

### Device

Desktop

### OS

Linux

### Browser

Chrome, Firefox

Bartvaderkin:
Fixes #2062

### DIFF
--- a/src/openforms/js/public.js
+++ b/src/openforms/js/public.js
@@ -1,3 +1,5 @@
+import {onLoaded} from './utils/dom';
+
 const init = () => {
   // connect the "Print this page" link element to a handler
   document.querySelectorAll('.a11y-toolbar__window-print-action').forEach(elem => {
@@ -7,8 +9,4 @@ const init = () => {
   });
 };
 
-if (document.readyState !== 'loading') {
-  init();
-} else {
-  document.addEventListener('DOMContentLoaded', init);
-}
+onLoaded(init);
--- a/src/openforms/js/public.js
+++ b/src/openforms/js/public.js
@@ -1,12 +1,18 @@
-import {onLoaded} from './utils/dom';
+import {onLoaded} from 'utils/dom';
 
-const init = () => {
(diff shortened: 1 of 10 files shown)


src/openforms/js/public.js
```diff
@@ -0,0 +1,8 @@
+document.addEventListener('DOMContentLoaded', event => {
```
sergei-maertens, line 1: so turns out that this approach is to be preferred:

```
if (document.readyState !== 'loading') {
  init();
} else {
  document.addEventListener('DOMContentLoaded', init);
}
```

... probably we should write a utility `onLoaded` function for that so we can just do `onLoaded(init);` at the end of the module.

sergei-maertens (approved):

PR #2201, opened 2022-10-17
[#1924] Dark theme fixes

### LINKED ISSUE #1924: Dark scheme of the portal is unworkable

latest

The "dark scheme" of the Form designer is unworkable. Most of the text is unreadable and in some cases the action buttons as well.

Screenshot 2022-08-25 at 16 33 07
Screenshot 2022-08-25 at 16 33 31

Please disable it and make it so that the default is always applied. Now I am blocked because of this unfinished functionality.

### Screen resolution

_No response_

### Device

Desktop

### OS

iOS

### Browser

Chrome

SilviaAmAm:
Fixes #1924

### DIFF
--- a/src/openforms/js/components/admin/form_design/Editor.js
+++ b/src/openforms/js/components/admin/form_design/Editor.js
@@ -9,6 +9,9 @@ const TinyMCEEditor = ({content, onEditorChange}) => {
   const editorRef = useRef(null);
   const tinyMceUrl = useContext(TinyMceContext);
   const intl = useIntl();
+  // TODO Django 4.2: use explicit theme names rather than the media query approach:
+  // https://github.com/django/django/blob/main/django/contrib/admin/static/admin/css/dark_mode.css#L36
+  const useDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches;
 
   return (
     <>
@@ -16,7 +19,12 @@ const TinyMCEEditor = ({content, onEditorChange}) => {
         tinymceScriptSrc={tinyMceUrl}
         onInit={(evt, editor) => (editorRef.current = editor)}
         value={content}
-        init={{...tinyMceConfig, language: intl.locale}}
+        init={{
+          ...tinyMceConfig,
+          language: intl.locale,
+          skin: useDarkMode ? 'oxide-dark' : 'oxide',
+          content_css: useDarkMode ? 'dark' : 'default',
+        }}
         onEditorChange={onEditorChange}
       />
     </>
--- a/src/openforms/scss/components/builder/_builder.scss
+++ b/src/openforms/scss/components/builder/_builder.scss
@@ -77,8 +77,13 @@ div.flatpickr-calendar.open {
   }
 
   .choices__input {
-    background-color: var(--form-input-bg) !important;
-    border-color: var(--border-color) !important;
+    background-color: var(--form-input-bg);
+    border-color: var(--border-color);
+  }
+
+  .choices__input.choices__input--cloned {
+    background-color: var(--form-input-bg);
+    border-color: var(--border-color);
   }
 
   .choices__item--selectable.is-highlighted {
@@ -223,5 +228,9 @@ div.flatpickr-calendar.open {
         border-color: var(--hairline-color) var(--hairline-color) var(--body-bg);
       }
     }
+  }
+
+  .formio-button-add-another {
+    color: var(--body-fg) !important;
   }
 }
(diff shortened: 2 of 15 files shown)


src/openforms/js/components/admin/form_design/Editor.js
```diff
@@ -9,14 +9,20 @@ const TinyMCEEditor = ({content, onEditorChange}) => {
   const editorRef = useRef(null);
   const tinyMceUrl = useContext(TinyMceContext);
   const intl = useIntl();
+  const useDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches;
```
sergei-maertens, line 12: can you add a TODO for Django 4.2 - they now use explicit theme names rather than the media query approach: https://github.com/django/django/blob/main/django/contrib/admin/static/admin/css/dark_mode.css#L36

src/openforms/scss/components/builder/_builder.scss
```diff
@@ -69,35 +69,33 @@ div.flatpickr-calendar.open {
   padding-top: 10px;
 }
 
-[ref='editForm'] {
```
sergei-maertens, line 72: I think we had things scoped under this selector to prevent unexpected side effects with other unrelated components that happen to use the same classnames (it may have been django-admin-index). What's the reason for removing this?

src/openforms/scss/components/builder/_builder.scss
```diff
@@ -69,35 +69,33 @@ div.flatpickr-calendar.open {
...
+    border-color: var(--border-color);
   }
 
-  .choices__list + .choices__input {
-    border: none;
-    border-radius: unset;
-    padding: 4px 0 4px 2px;
+  .choices__input {
+    background-color: var(--form-input-bg) !important;
+    border-color: var(--border-color) !important;
```
sergei-maertens, line 81: why are the `!important` flags required? Preferably you override styles by using a more specific selector, which is reverted here in this PR.

src/openforms/scss/vendor/_rjsf-field.scss
```diff
@@ -25,6 +25,11 @@
     input[type='email'] {
       width: $input-field-size; // override from admin/rjsf default
     }
+
+    select {
+      background-color: var(--form-input-bg) !important;
+      border-color: var(--border-color) !important;
+    }
```
sergei-maertens, line 32: maybe try using `color-scheme: dark` here too and see how that reacts? If we can avoid the `!important`, we should.

src/openforms/scss/components/builder/_builder.scss
SilviaAmAm: I removed because the the select dropdown in the editor has these classes but I couldn't find a parent with the `ref=editForm`. :thinking:

src/openforms/scss/vendor/_rjsf-field.scss
SilviaAmAm: It gets a bit of a different colour:
Screenshot from 2022-10-19 12-36-53
Should I still go for it or leave the the colours?

SilviaAmAm:
>     * The product edit form still has light-theme WYSIWYG fields, those are rendered via the django package afaik. These should also be adapted to use the correct theme (and preferably in a generic way)

I see there is this issue for it: https://github.com/jazzband/django-tinymce/issues/375. I'll look into it and see if I can figure out how to make it detect the dark theme!

src/openforms/js/components/admin/form_design/Editor.js
```diff
@@ -9,14 +9,22 @@ const TinyMCEEditor = ({content, onEditorChange}) => {
...
       <Editor
         tinymceScriptSrc={tinyMceUrl}
         onInit={(evt, editor) => (editorRef.current = editor)}
         value={content}
-        init={{...tinyMceConfig, language: intl.locale}}
+        init={{
+          ...tinyMceConfig,
+          language: intl.locale,
+          skin: useDarkMode ? 'oxide-dark' : 'oxide',
+          content_css: useDarkMode ? 'dark' : 'default',
```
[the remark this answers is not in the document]
SilviaAmAm: made an issue for this: https://github.com/open-formulieren/open-forms/issues/2215

src/openforms/scss/vendor/_rjsf-field.scss
sergei-maertens: Hrm, probably because we explicitly set the background for the text inputs rather than using `color-scheme: dark` for those. Let's go with the explicit colors then - I'm curious which styles overrides this color though that require the `!important`.

PR #2207, opened 2022-10-18
[#1180] Update gtag and google tag manager

### LINKED ISSUE #1180: Google Analytics does not measure form steps correctly
Currently, Google Analytics measures views and pages but all forms are opened in the same page (called "Openforms").

We need to change the title of the window to match the form, and maybe even include the step, to get this working. This also makes the window title show nicely, so win-win.

SilviaAmAm:
Fixes #1180 

Front end PR: https://github.com/open-formulieren/open-forms-sdk/pull/269

### DIFF
--- a/docs/developers/csp.rst
+++ b/docs/developers/csp.rst
@@ -84,6 +84,11 @@ do not rely on external CDNs (such as the Formio CDN, Google Fonts...).
 
 * The ``img-src`` requires ``data:``, otherwise map and signature components break.
 
+* If Google Tag manager is enabled, the ``img-src`` requires ``www.googletagmanager.com`` to allow the execution of the
+  Tag Manager container code. More information can be found `here`_.
+
+.. _here: https://developers.google.com/tag-platform/tag-manager/web/csp#enable_the_container_tag_to_use_csp
+
 **CSP and WYSIWYG**
 
 Some form design relies on WYSIWYG editors, which results in HTML blobs in API endpoint
--- a/src/openforms/scss/components/admin/_google-tag-manager.scss
+++ b/src/openforms/scss/components/admin/_google-tag-manager.scss
@@ -1,0 +1,6 @@
+.google-tag-manager {
+  &__noscript {
+    display: none;
+    visibility: hidden;
+  }
+}
--- a/src/openforms/scss/components/admin/_google-tag-manager.scss
+++ b/src/openforms/scss/components/admin/_google-tag-manager.scss
@@ -1,6 +1,0 @@
-.google-tag-manager {
-  &__noscript {
-    display: none;
-    visibility: hidden;
-  }
-}
--- a/src/openforms/templates/master.html
+++ b/src/openforms/templates/master.html
@@ -29,6 +29,14 @@
 </head>
 
 <body class="view__body">
+    {% if enable_analytics and analytics_tools_config.gtm_code %}
+        {# Load the second part of the Google tag manager snippet #}
+        <!-- Google Tag Manager (noscript) -->
+        <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ analytics_tools_config.gtm_code }}"
+        height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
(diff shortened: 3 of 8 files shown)


src/openforms/templates/master.html
```diff
@@ -29,6 +29,14 @@
 </head>
 
 <body class="view__body">
+    {% if enable_analytics and analytics_tools_config.gtm_code %}
```
sergei-maertens, line 32: shouldn't this also only be emitted if Google Analyics is explicitly enabled - not only based on whether the code is filled out or not?

sergei-maertens (changes requested):

src/openforms/templates/master.html
SilviaAmAm: This snippet should be present when the other Google Tag Manager script is present (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/analytics_tools/templates/analytics_tools/google.html#L8). Unless i read it wrong, I think the other one relies on `enable_analytics` and `analytics_tools_config.gtm_code`?

sergei-maertens: I see that the model has the field `enable_google_analytics`, but this is not used anywhere now - this should be used similar to `{% if analytics_tools_config.piwik_enabled %}` in `src/openforms/analytics_tools/templates/analytics_tools/piwik.html`

The idea is that you can configure your integration without enabling it.

src/openforms/templates/master.html
```diff
@@ -29,6 +29,14 @@
 </head>
 
 <body class="view__body">
+    {% if enable_analytics and analytics_tools_config.enable_google_analytics and analytics_tools_config.gtm_code %}
+        {# Load the second part of the Google tag manager snippet #}
+        <!-- Google Tag Manager (noscript) -->
+        <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ analytics_tools_config.gtm_code }}"
+        height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
```
sergei-maertens, line 36: I think this inline style will be blocked by the CSP

sergei-maertens (changes requested):

docs/developers/csp.rst
```diff
@@ -84,6 +84,11 @@ do not rely on external CDNs (such as the Formio CDN, Google Fonts...).
 
 * The ``img-src`` requires ``data:``, otherwise map and signature components break.
 
+* If Google Tag manager is enabled, the ``img-src`` requires ``www.googletagmanager.com`` to allow the execution of the
+  Tag Manager container code. More information can be found `here`_.
+
+.. _here: https://developers.google.com/tag-platform/tag-manager/web/csp#enable_the_container_tag_to_use_csp
+
```
sergei-maertens, line 91: This should be added to the automatic CSP configuration when activating/deactivating an analytics tool: `src/openforms/analytics_tools/contrib/google_analytics/csp_headers.json`

src/openforms/templates/master.html
```diff
@@ -29,6 +29,14 @@
 </head>
 
 <body class="view__body">
+    {% if enable_analytics and analytics_tools_config.enable_google_analytics and analytics_tools_config.gtm_code %}
+        {# Load the second part of the Google tag manager snippet #}
+        <!-- Google Tag Manager (noscript) -->
+        <noscript><iframe class="google-tag-manager__noscript" src="https://www.googletagmanager.com/ns.html?id={{ analytics_tools_config.gtm_code }}"
```
sergei-maertens: this is not correct BEM naming - an element cannot exist without a parent block (there's no `.google-tag-manager` class anywhere)

src/openforms/scss/components/admin/_google-tag-manager.scss
```diff
@@ -0,0 +1,6 @@
+.google-tag-manager {
```
sergei-maertens: not the correct location - this is not an admin component, but a public component (it gets loaded on the form detail page)

sergei-maertens (approved):

Issue #1862, opened 2022-08-12
When employees are logged in to a form, registration backends should add them as being an employee


joeribekker:
Followup on #1861

When employees login on behalf of a resident, a submission should typically have 2 persons attached to it:

1. The employee (employee number)
2. The person (bsn) or company (kvk) it involves

However, this is NOT a requirement since an employee can also fill in an internal form that is only used by employees and not meant for residents. So, we could also only have the employee number.

We need to modify the registration backend plugins

# Technical approach

Two aspects are relevant in this flow:

* We require "authentication details" to be able to prefill/register the correct
  attributes of the subject of the form
* We want third parties to be able to initiate/complete a form on behalf of a subject
  AND keep track of this fact.

## Constraints

* `Submission.auth_info` (reverse one-to-one of `AuthInfo.submission`) is the foundation
  for prefill/registration and not easily refactored (again)

* `AuthInfo` _does_ contain all the fields we'd need to keep track of the third party
  (called "registrator" from now on)

## Data model changes

* [x] Create abstract base model `BaseAuthInfo` from current model `AuthInfo` (except for the `submission` `OneToOneField`)
* [x] Derive `AuthInfo` from `BaseAuthInfo` (with the `submission` field) - this keeps the existing machinery intact
* [x] Add model `RegistratorInfo` derived from `BaseAuthInfo`, with an additional `submission` `OneToOneField`. Use the `_registrator` related name.
* [x] Define `Submission.registrator` as `Submission._registrator or Submission.auth_info`, meaning that if no explicit Registrator` is provided, the usual auth details is used for this purpose.

## Employee login flow

Currently, the employee logs in using OIDC by:

1. Click button -> redirected to Open Forms backend
2. Open Forms backend redirects to identity provider (IDP)
3. Employee logs in with IDP (if no active session exists)
4. Employee is redirected back to Open Forms backend
5. Open Forms backend creates/updates Django user + logs this user in. Employee details
   are stored in session.
6. Open Forms backend redirects back to form
7. SDK creates submission
8. Backend fires `submission_start` signal
9. Backend creates `AuthInfo` with employee details from the session and relates it to
   the submission.

This flow needs to be modified so the employee can additionally specify which
person/company they are acting on behalf of:

* [x] Once the employee is redirected to the Open Forms backend (5.) after succesful
      authentication and login, display an intermediate screen (POST form)

      * Button to skip additional information (register this explicitly!)
      * Input field for BSN
      * Input field for KVK

[...]

PR #2216, opened 2022-10-19
[#1862] Implemented registrator-subject view and storage

Bartvaderkin:
Part of #1862


src/openforms/authentication/urls.py
```diff
@@ -1,10 +1,19 @@
...
+    AuthenticationReturnView,
+    AuthenticationStartView,
+    RegistratorSubjectInfoView,
+)
 
 app_name = "authentication"
 
 urlpatterns = [
+    path(
+        "<slug:slug>/registrator-subject",
```
Bartvaderkin, line 13: I added this URL with the slug but in hindsight we don't actually use the form (maybe be should).

src/openforms/authentication/contrib/org_oidc/plugin.py
```diff
@@ -13,6 +13,24 @@
 from ...registry import register
 
 
+def _reverse_plus(name, *, args=None, kwargs=None, request=None, query=None):
```
Bartvaderkin, line 16: Proposal to make this `_reverse_plus()` a shared util so we have less of this noisy `reverse`/`furl` everywhere (the blacked versions of those lines are terrible)

sergei-maertens: let's do it - you can put it in `openforms.utils.urls` and ensure it's picked up by the developer documentation.

I'd also consider adding a kwarg `make_absolute: bool = False` and if true, pass the result through `build_absolute_uri`

src/openforms/authentication/contrib/org_oidc/plugin.py
```diff
@@ -23,20 +41,25 @@ class OIDCAuthentication(BasePlugin):
...
+        redirect_url = _reverse_plus(
+            "org-oidc:init",
+            request=request,
+            query={
+                "next": auth_return_url,
+            },
+        )
+        if request.user.is_authenticated:
+            # logout user if logged in with other account
+            auth.logout(request)
```
sergei-maertens, line 61: can you expand this comment a bit that we do this because otherwise staff users get redirect to internal admin pages instead of the intended next page due to some internal oidc-library state in the session?

src/openforms/authentication/contrib/org_oidc/tests/test_auth_procedure.py
```diff
@@ -298,11 +326,22 @@ def test_callback_url_creates_logged_in_djang_ouser(
...
         mock_verify_token.return_value = user_claims
         mock_get_userinfo.return_value = user_claims
         mock_store_tokens.return_value = {"whatever": 1}
 
+        # setup user group
+        solo = OpenIDConnectConfig.get_solo()
+        group = GroupFactory(
+            name="registrators",
+            permissions=["of_authentication.can_register_client_submission"],
+        )
```
sergei-maertens, line 340: let's add this group with permission to the default groups fixture!

src/openforms/authentication/contrib/org_oidc/tests/test_auth_procedure.py
```diff
@@ -298,11 +326,22 @@ def test_callback_url_creates_logged_in_djang_ouser(
...
+        solo.save()
+        solo.default_groups.add(group)
```
sergei-maertens, line 342: I'd swap these calls - the save triggers cache invalidation.

src/openforms/authentication/contrib/org_oidc/tests/test_auth_procedure.py
```diff
@@ -332,20 +372,39 @@ def test_callback_url_creates_logged_in_djang_ouser(
...
+
+        # we got our user
         user = User.objects.get()
         self.assertTrue(user.is_staff)
         self.assertTrue(user.username, "some_username")
         self.assertTrue(user.email, "admin@example.com")
         self.assertTrue(user.first_name, "John")
         self.assertTrue(user.last_name, "Doe")
         self.assertTrue(user.employee_id, "my_id_value")
+        self.assertEqual(list(user.groups.all()), [group])
```
sergei-maertens, line 388: you can use `self.assertQuerySetEqual` too

src/openforms/authentication/forms.py
```diff
@@ -0,0 +1,40 @@
...
+        label=_("KvK number of client"),
+        required=False,
+        max_length=8,
+        validators=[validate_kvk],
+    )
+    skip_subject = forms.BooleanField(
+        label=_("Continue without additional information"),
+        required=False,
+        initial=False,
+        # TODO make a widget instead of hacking a <button> in the form template
```
sergei-maertens, line 25: solving this at the HTML level is okay, as long as it has a (web)test

src/openforms/authentication/models.py
```diff
@@ -102,3 +102,10 @@ class RegistratorInfo(BaseAuthInfo):
     class Meta:
         verbose_name = _("Registrator authentication details")
         verbose_name_plural = _("Registrator authentication details")
+
+        permissions = [
+            (
+                "can_register_client_submission",
```
sergei-maertens, line 108:
```suggestion
                "can_register_customer_submission",
```
perhaps that's a better term to avoid confusion with server/client, municipalities also always tend to speak about "klantreizen" so it lines up better.

src/openforms/authentication/signals.py
```diff
@@ -59,19 +63,43 @@ def set_auth_attribute_on_session(
         )
         raise PermissionDenied(_("Demo plugins require an active admin session."))
 
+    # TODO check this log is accurate
     logger.debug(
         "Persisting form auth to submission %s. Plugin: '%s' (setting attribute '%s')",
         instance.uuid,
         plugin,
         attribute,
     )
```
sergei-maertens, line 72: I'd include an additional info item whether a registrator initiated this or not (basically: `registrator_subject is not None` - or maybe better -> rely on `request.user` to provide this info.

src/openforms/authentication/signals.py
```diff
@@ -59,19 +63,43 @@ def set_auth_attribute_on_session(
...
+        if registrator_subject.get("skipped_subject_info"):
+            # If continued without extra details, store the employee details in AuthInfo
+            store_auth_details(instance, form_auth)
+        else:
+            # If KVK/BSN is given, store those details in the AuthInfo model and the employee details in the Registrator model.
+            auth_save = {
+                "value": registrator_subject["value"],
+                "attribute": registrator_subject["attribute"],
+                # TODO we don't have a plugin to define here? things break if this doesn't exist (like the logout view)
+                "plugin": registrator_subject.get("plugin", "registrator"),
```
sergei-maertens, line 85: good enough for now!

src/openforms/authentication/tests/test_admin.py
```diff
@@ -51,3 +51,49 @@ def test_validate_invalid_kvk(self):
         self.assertContains(
             form_response, "KvK-nummer moet 8 tekens lang zijn.", html=True
         )
+
+
+class RegistratorInfoAdminValidationTest(WebTest):
+    def setUp(self):
+        super().setUp()
+        self.user = UserFactory.create(is_superuser=True, is_staff=True, app=self.app)
```
sergei-maertens, line 59:
```py
@classmethod
def setUpTestData(cls):
    ...
```

is a bit more performant for this as it only runs once for the entire testcase. Additionally, we can drop the `app=self.app` requirement if we slap the `@disable_2fa` decorator on the test.

src/openforms/authentication/signals.py
```diff
@@ -59,19 +63,43 @@ def set_auth_attribute_on_session(
...
+            }
+            registrator_save = {
+                "value": form_auth["value"],
+                "attribute": form_auth["attribute"],
+                "plugin": form_auth["plugin"],
+            }
+            # TODO the "machtigingen" key dispeared
```
sergei-maertens, line 92: can you elaborate?

src/openforms/authentication/tests/test_signals.py
```diff
@@ -75,6 +83,77 @@ def test_attribute_set_for_demo_plugin_with_staff_user(self):
...
+        instance.refresh_from_db()
+
+        # check the property doesn't alias both
+        self.assertNotEqual(instance.auth_info, instance.registrator)
+
+        # check the clients info is stored as .registrator
+        self.assertEqual(instance.auth_info.attribute, AuthAttribute.bsn)
+        self.assertEqual(instance.auth_info.value, "123")
+        # TODO check what we want from this
+        # self.assertEqual(instance.auth_info.plugin, "registrator")
```
sergei-maertens, line 118: let's leave it unspecified - as long as existing tests don't break, we don't provide any guarantees about it.

src/openforms/authentication/tests/test_views.py
```diff
@@ -384,3 +391,179 @@ def test_co_sign_flow_invalid_submission_id(self):
...
+
+        f.args["next"] = "http://not-in-whitelist.net/foo"
+        self.subject_url_not_allowed_next = f.url
+
+        f.args["next"] = self.form_url
+        self.subject_url = f.url
+
+        self.user = UserFactory(
+            user_permissions=["of_authentication.can_register_client_submission"]
+        )
```
sergei-maertens, line 430: perhaps put this in `setUpTestData` instead to keep the tests as fast as possible :)

src/openforms/authentication/tests/test_views.py
```diff
@@ -384,3 +391,179 @@ def test_co_sign_flow_invalid_submission_id(self):
...
+            self.assertContains(response, _("Use either BSN or KvK"), html=True)
+
+        with self.subTest("valid bsn"):
+            response = self.app.get(self.subject_url, status=200, user=self.user)
+            form = response.forms["registrator-subject"]
+            form["bsn"] = "115736499"
+            response = form.submit(status=302)
+            self.assertRedirects(response, self.form_url, fetch_redirect_response=False)
+
+        with self.subTest("valid bsn"):
```
sergei-maertens, line 521:
```suggestion
        with self.subTest("valid kvk"):
```

src/openforms/authentication/urls.py
sergei-maertens: let's leave it like this for consistency with the other URLs/endpoints - maybe in the future we'll need it or it'll be relevant input for a plugin.

src/openforms/authentication/utils.py
```diff
@@ -26,3 +29,17 @@ def store_auth_details(submission: Submission, form_auth: FormAuth) -> None:
     ), f"Unexpected auth attribute {attribute} specified"
 
     AuthInfo.objects.update_or_create(submission=submission, defaults=form_auth)
+
+
+def store_registrator_details(
+    submission: Submission, registrator_auth: BaseAuth
```
sergei-maertens, line 35: I'd like a more specific type hint than `BaseAuth` which also contains the `skip_subject_info` key with boolean value

src/openforms/authentication/views.py
```diff
@@ -373,3 +381,89 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+        ),
+        302: None,
+        (400, "text/html"): OpenApiResponse(
+            response=str, description=_("Bad request. Invalid parameters were passed.")
+        ),
+        (403, "text/html"): OpenApiResponse(
+            response=str, description=_("Forbidden. Invalid authentication.")
+        ),
+    },
+)
```
sergei-maertens, line 428: we can delete all this - it's an internal page that doesn't need to be documented in the API spec (and I'm fairly certain it's not picked up either because it's not a `rest_framework.views.APIView`)

src/openforms/templates/authentication/registrator_subject_info.html
```diff
@@ -0,0 +1,41 @@
+{% extends 'ui/views/abstract/detail.html' %}
```
sergei-maertens, line 1: we'll tackle styling this in another PR - I'll also check with Joeri whether we want this to have public UI or admin styling.

sergei-maertens (changes requested):
Couple small remarks, but this is really nice work!

src/openforms/authentication/contrib/org_oidc/tests/test_auth_procedure.py
Bartvaderkin: Although this snippet isn't needed with the group in a fixture

src/openforms/authentication/signals.py
Bartvaderkin: If there is a `machtigingen` key/value in the `form_auth` it won't get copied here. I don't think it is relevant here though.

src/openforms/authentication/utils.py
Bartvaderkin: If there is a `skip_subject_info` in the dictionary in the session then it won't reach here, and a new dict conforming `BaseAuth` is created in `set_auth_attribute_on_session()`. I could make a type of what is in the session but it would be like a `Union[BaseAuth, TypedDictWithOnlySkipSubjectInfo]`

sergei-maertens:
```
class SkipSubjectInfo(TypedDict):
    skip_subject_info: bool

RegistratorInfo = Union[BaseAuth, SkipSubjectInfo]
```
is probably the compromise to go for then - you can then use the `RegistratorInfo` type hint and editors will pick it up correctly.

src/openforms/authentication/signals.py
sergei-maertens: yeah you're right - maybe detect if there's a non-empty value and emit a `logger.warning(...)` in that case that the data was discarded. It shouldn't happen, but if it _does_ happen, we at least make the problem visible.

src/openforms/authentication/utils.py
Bartvaderkin: But `store_registrator_details()` doesn't do anything with `skip_subject_info`, so that union would be invalid in that function signature.

sergei-maertens (approved):

PR #2217, opened 2022-10-19
Improve example config for future devs

CharString:
Eager celery tasks do database queries in these suites.


CharString:
I've reverted the test modifications and change the conf instead. Copying the example will not break tests anymore.

sergei-maertens (approved):

PR #2218, opened 2022-10-19
🐛 [#1929] Validate file type/extension on upload

### LINKED ISSUE #1929: Upload field accepts files with re-named (false) extensions

1.1.5

When I upload a file where the mimetype doesn't match the extension (renamed from `file.zip` to `file.jpg`, `file.jpg` to `file.png` etc.), this isn't caught by the upload component. It is caught only upon trying to send the form. However, this doen't lead to a (visible) error message, so the user is left in the dark about why the form won't send.

Component accepts the file ( file_all_types_iphone.png is actually a renamed .jpg)
image

The error as shown in the network-tab upon trying to send:
`{"type":"https://haarlemmermeer.test.open-formulieren.nl/fouten/ValidationError/","code":"invalid","title":"Invalid input.","status":400,"detail":"","instance":"urn:uuid:9a1e4850-6468-4c82-8130-4e9ec6998807","invalidParams":[{"name":"data.Voeg een foto of kopie toe van een geldig paspoort of een identiteitskaart","code":"invalid","reason":"The file 'file_all_types_iphone.png' is not a valid file type."}]}`

_No response_

For the upload component to catch that the mimetype doesn't match the extension, and display an error accordingly.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

CharString:
Fixes #1929

Some refactoring could be done such that we override `application/octet-stream` on the value with the correct mime-type. Then the validator on form submission can trust the content_type and just needs to check whether it's in the set of accepted ones.

That does follow [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle), and should be okay on POST requests, but wouldn't be what I'd expect of this serializer is used for a PUT request.


src/openforms/formio/api/serializers.py
```diff
@@ -44,3 +46,24 @@ def get_url(self, instance) -> str:
...
+                    % {"filename": value.name, "file-type": f".{ext}"}
+                )
+        elif magic.from_buffer(value.read(2048), mime=True) != value.content_type:
+            raise serializers.ValidationError(
+                _("The file %(filename)s is not a %(file-type)s.")
+                % {"filename": value.name, "file-type": value.content_type}
+            )
+
+        # valid
+        return value
```
sergei-maertens, line 69: Can we implement this as a re-usable DRF validator class/function, and then call that as part of the `openforms.submissions.attachments.validate_uploads` flow? I'd like to keep this validation machinery/logic in a single place to avoid implementations diverging in logic.

src/openforms/formio/api/serializers.py
```diff
@@ -44,3 +46,24 @@ def get_url(self, instance) -> str:
             kwargs={"uuid": instance.uuid},
             request=request,
         )
+
+    def validate_file(self, value: UploadedFile):
+        if value.content_type == "application/octet-stream":
+            ext = value.name.split(".")[-1]
```
sergei-maertens, line 52: we prefer the modern `pathlib.Path(value.name).suffix`

sergei-maertens (changes requested):

src/openforms/formio/api/serializers.py
CharString: That's what I had initially. I <3 pathlib, but though it is semantically incorrect, it's not really a path, just a name. i.e. checking on `.exists()` leads to drawing incorrect conclusions about the state of the file etc.

sergei-maertens: fair enough. I suppose you could argue that it's a relative path ( :wink: ) but I'm not married to it.

I also did a quick check for the content type lookup (e.g. with `.tar.gz` file where you'd have multiple suffixes), but in the end only the last bit is relevant (this would yield a `gzip` content type, tar does not have one for itself) which is also in line with expectations.

CharString: lol, relative to what? The `django.core.file.uploadedfile.UploadedFile` can be a proxy to memory, a cdn, anything really. I went with `PurePath.suffix` as a compromise :handshake:

magic allows for inspection into these types like zipfiles and tarballs. Should the check in `openforms.submissions.attachments.validate_uploads` check if all the contents of these archive files are allowed too?

src/openforms/formio/api/serializers.py
CharString: I've refactored for re-use. I didn't squash the last refactor commit, because the :beers: gitmoji could be equally valid. Co-author Hertog Jan :wink:. Have a nice weekend.

src/openforms/formio/utils.py
```diff
@@ -140,7 +140,7 @@ def get_component_default_value(component) -> Optional[Any]:
     return default_value
 
 
-def mimetype_allowed(mime_type: str, allowed_mime_types: List[str]) -> bool:
+def mimetype_allowed(mime_type: str, allowed_mime_types: Container[str]) -> bool:
```
sergei-maertens, line 143: it's not wrong to use `Container` - but it does expand the API a bit beyond intended usage - the allowed mimetypes typically comes from the formio configuration which uses a list (since it's a JSON datatype, where tuples/sets/dicts don't exist).

Duplicate entries aren't really a problem for the implementation, so I'd revert this back and use lists instead (zen of python - preferably one way to do something :wink: )

src/openforms/submissions/attachments.py
```diff
@@ -141,30 +141,22 @@ def validate_uploads(submission_step: SubmissionStep, data: Optional[dict]) -> N
         else:
             allowed_mime_types = glom(component, "file.type", default=[])
 
+        validate_file = mime_type_validator(allowed_mime_types)
```
sergei-maertens, line 144: see earlier remark about the class-based validator implementation while still re-using the actual validation code. That should make the code more straightforward to read and maintain.

src/openforms/submissions/attachments.py
```diff
@@ -141,30 +141,22 @@ def validate_uploads(submission_step: SubmissionStep, data: Optional[dict]) -> N
...
-            validation_errors[key].append(invalid_file_type_error)
-            continue
+            try:
+                validate_file(uploaded_file)
+            except ValidationError as e:
+                logger.warning(
+                    "Blocking submission upload %d because of invalid mime type check",
+                    upload.id,
+                )
+                validation_errors[key].append(e.detail)
```
sergei-maertens, line 159: do we have a check that the validation error code is  still `"invalid"` here - that information is not passed explicitly anymore in the validator (it may be the default code).

sergei-maertens (changes requested):
More nitpicking :sweat_smile:

src/openforms/api/validators.py
```diff
@@ -22,3 +31,59 @@ def __call__(self, data: dict, serializer: serializers.Serializer):
...
+
+    def validate_file(value: UploadedFile) -> UploadedFile:
+        head = value.read(2048)
+        mime_type = magic.from_buffer(head, mime=True)
+        if value.content_type == "application/octet-stream":
+            ext = PurePath(value.name).suffix
+            m = magic.Magic(extension=True)
+            extensions = m.from_buffer(head).split("/")
+            if extensions == ["???"]:
+                pass  # magic db doesn't know extensions; decide later on mime-type
```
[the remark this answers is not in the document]
CharString: Hmmm. on my machine magic doesn't know the extension for image/gif, maybe your fancy rolling Arch knows more.  I'll add an explicit test with a mocked magic to decouple it... `mock.MagicMock('magic')` I love it already <3

src/openforms/api/validators.py
```diff
@@ -22,3 +31,59 @@ def __call__(self, data: dict, serializer: serializers.Serializer):
...
+                    _("The file '{filename}' is not a {file_type}.").format(
+                        filename=value.name, file_type=ext
+                    )
+                )
+        elif mime_type != value.content_type:
+            raise serializers.ValidationError(
+                _("The file '{filename}' is not a {file_type}.").format(
+                    filename=value.name, file_type=value.content_type
+                )
+            )
```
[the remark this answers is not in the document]
CharString: :+1:  And fallback to the mime-type for the `???` case.

src/openforms/formio/utils.py
CharString: I'm in love with `PathLike` and programming to an interface/protocol instead of an implementation in general. If an algorithm just iterates, use an `Iterable` or `Sequence`, if it just checks with `in`, a `Container` is what the algorithm expects and actually uses. It's all over the Django code too. Almost everywhere you can interchangeably use lists or tuples everywhere. "Explicit is better than implicit", this explicitly communicates what part of the `List` api is actually expected. "... Although that way may not be obvious at first unless you're Dutch." :wink:

src/openforms/api/tests/test_validators.py
```diff
@@ -0,0 +1,76 @@
...
+    _original_language = get_language()
+
+    def setUp(self):
+        self.CORRECT_GIF.seek(0)
+
+    def setUpClass():
+        activate("en")
+
+    def tearDownClass():
+        activate(MimeTypeValidatorTests._original_language)
```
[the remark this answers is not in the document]
CharString: `override_settings` only works with subclasses of `django.test.SimpleTestCase` this is a plain vanilla `unittest.TestCase` because it doesn't hit any of the django magic, it unittests a function.
Changing to a `SimpleTestCase` is more than 2 orders slower. At the rate I run tests during development, that counts.
```
........                                                                 [100%]
8 passed in 0.08s

==================================== ERRORS ====================================
_________ ERROR collecting src/openforms/api/tests/test_validators.py __________
test_validators.py:16: in <module>
    class MimeTypeValidatorTests(TestCase):
../../../../env/lib/python3.8/site-packages/django/test/utils.py:392: in __call__
    return self.decorate_class(decorated)
../../../../env/lib/python3.8/site-packages/django/test/utils.py:469: in decorate_class
    raise ValueError(
E   ValueError: Only subclasses of Django SimpleTestCase can be decorated with override_settings
=========================== short test summary info ============================
ERROR test_validators.py - ValueError: Only subclasses of Django SimpleTestCa...
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
1 error in 0.27s

[1]+  Stopped                 vi test_validators.py
(env) chars@Macaca:~/Projects/open-forms/src/openforms/api/tests$ fg
vi test_validators.py
........                                                                 [100%]
8 passed in 10.15s
```

src/openforms/api/validators.py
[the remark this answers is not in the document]
sergei-maertens: or use something like `uuid.uuid4` for a generated extension, that will surely not find anything.

`mock` is (a necessary) evil and preferably avoided where possible :P

src/openforms/formio/utils.py
sergei-maertens: but it does limit your implementation for future refactoring - stretching it up to `Container` does remove your "freedom" to alter implementation details and call `List` specific interface (e.g. access the `allowed_mime_types.index(<item>)` to include this information in a validation error.

In general I try to be a specific and constrained as possible, because it's easier to relax constraints than tighten them in the future.

src/openforms/api/tests/test_validators.py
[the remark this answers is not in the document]
sergei-maertens: that's odd - I can't imagine where that time is spent since `SimpleTestCase` isn't even allowed to touch the database.

Are you running tests locally with `--keepdb`? You can also tweak your local postgres instance to not flush to disk (I have set up a second cluster just for running tests locally on port 5433) to keep things fast.

We don't really use bare `unittest.TestCase` anywhere exactly because we're always operating in the Django context and expect to be able to use the django testing tools.

Also running it locally to see the difference, you may be onto something with the magic DB being different:

```
======================================================================
ERROR: test_accepts_unknown_extensions (openforms.api.tests.test_validators.MimeTypeValidatorTests)
  File "/home/bbt/code/open-forms/src/openforms/api/tests/test_validators.py", line 54, in test_accepts_unknown_extensions
    self.assertIs(validators.mime_type_validator(file), file)
  File "/home/bbt/code/open-forms/src/openforms/api/validators.py", line 88, in mime_type_validator
    return validate_file(arg)
  File "/home/bbt/code/open-forms/src/openforms/api/validators.py", line 68, in validate_file
rest_framework.exceptions.ValidationError: [ErrorDetail(string="The file 'pixel.gif' is not a .gif.", code='invalid')]

Ran 8 tests in 0.019s
```

Timings:

* `unittest.TestCase`: 0.019s
* `django.test.SimpleTestCase` without `--keepdb`: 0.004s
* `django.test.SimpleTestCase` with `--keepdb`: 0.004s

So I'm not seeing that two orders of magnitude slowdown :thinking:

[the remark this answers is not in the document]
CharString: Good catch on coupling to the translations! This removes the need for the class setup and teardown:

```python
@patch('openforms.api.validators._', lambda s: s)  # decouple from translations by returning msgid
class MimeTypeValidatorTests(TestCase):
```

sergei-maertens (approved):
Thanks - I didn't convey the message right about moving the validator code, I meant to move the validator into the `openforms.formio` package rather than moving everything (including util) into the `openforms.api` package, so I amended that in this PR.

Let's get this merged!

sergei-maertens (approved):

PR #2220, opened 2022-10-20
[#2206] Prefill race condition

### LINKED ISSUE #2206: AttributeError: 'NoneType' object has no attribute 'prefill_plugin'

2.0.0-rc.0

Sentry 317885

Seen on Den Haag test environment

```python
    def get_prefill_variables(self) -> List["SubmissionValueVariable"]:
        prefill_vars = []
        for variable in self.variables.values():
            prefill_plugin = variable.form_variable.prefill_plugin # Error here!
            if prefill_plugin == "":
                continue
            prefill_vars.append(variable)
        return prefill_vars
```

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #2206


src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -296,6 +296,13 @@ class SubmissionValueVariable(models.Model):
         blank=True,
         auto_now=True,
     )
+    is_initially_prefilled = models.BooleanField(
```
SilviaAmAm, line 299: Actually `SubmissionValueVariable` already had a source field, but this could be overwritten by user_input/logic (when we eventually sort out the source field to be set by these changes). Then for the prefill data check we wouldn't retrieve these variables.

So I added a new field to track the prefill

sergei-maertens (approved):

sergei-maertens:
Applied to stable/2.0.x in 053bd4e6abafa7290f801521a2828af5bb3e1950 and 2be7b3971a285925aff857109299dcf9ba8cfd2e

Issue #2228, opened 2022-10-24
Configure Django settings to enable internationalization/localization


sergei-maertens:
Epic: #2170

**Tasks**

* [x] Enable Django's `LocaleMiddleware` (to detect the language preferences from the browser)
* [x] Configure the language cookie settings for optimal security:
    - `httpOnly` flag enabled
    - `Secure` flag enabled when in an HTTPS context (see `IS_HTTPS` setting)
    - `Samesite` must be "None" for cross-origin operation (like embedding the SDK in a third party site), otherwise "Lax" is ok

PR #2233, opened 2022-10-24
:wrench: [#2228] Prepare settings for i18n

stevenbal:
Fixes: #2228 

Changes:
* Add settings for i18n

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -46,6 +46,12 @@ LANGUAGES = [
     ("en", _("English")),
     ("nl", _("Dutch")),
 ]
+LANGUAGE_COOKIE_HTTPONLY = True
+LANGUAGE_COOKIE_NAME = "openforms_language"
+LANGUAGE_COOKIE_SAMESITE = config(
+    "LANGUAGE_COOKIE_SAMESITE", default="None" if IS_HTTPS else "Lax"
+)
+LANGUAGE_COOKIE_SECURE = IS_HTTPS
 
 TIME_ZONE = "Europe/Amsterdam"  # note: this *may* affect the output of DRF datetimes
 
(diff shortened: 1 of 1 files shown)


src/openforms/conf/base.py
```diff
@@ -47,6 +47,14 @@
     ("nl", _("Dutch")),
 ]
 
+LANGUAGE_COOKIE_HTTPONLY = True
+
+LANGUAGE_COOKIE_NAME = "openforms_language"
```
stevenbal, line 52: Not sure if we want this to be configurable

sergei-maertens: nope, fixed value is fine (session cookie name is also fixed)

src/openforms/conf/base.py
```diff
@@ -47,6 +47,14 @@
...
+
+LANGUAGE_COOKIE_SAMESITE = "Lax"
```
sergei-maertens, line 54: this needs to be `"None"` for cross-origin use, see the similar CSRF & Session cookie setings

stevenbal: Ah okay, so whenever forms are embedded, you would want this to be "None"

sergei-maertens: yep!

src/openforms/conf/base.py
```diff
@@ -47,6 +47,16 @@
...
 
+LANGUAGE_COOKIE_HTTPONLY = True
+
+LANGUAGE_COOKIE_NAME = "openforms_language"
+
+LANGUAGE_COOKIE_SAMESITE = config(
+    "LANGUAGE_COOKIE_SAMESITE", default="None" if IS_HTTPS else "Lax"
+)
+
+LANGUAGE_COOKIE_SECURE = IS_HTTPS
```
sergei-maertens, line 58: final remark - can you remove the blank lines between the language settings to be consistent with the rest of the settings file? :grimacing:

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #2230, opened 2022-10-24
Add django-modeltranslation to manage Django model fields translations


sergei-maertens:
Epic #2170 

**Tasks**

- [ ] Add django-modeltranslation dependency, with admin integration
- [ ] Make `forms.Form` model fields translatable:
    - `name`
    - `submission_confirmation_template`
    - `begin_text`
    - `previous_text`
    - `change_text`
    - `confirm_text`
    - `explanation_template`
- [ ] Make `forms.FormStep` fields translatable:
    - `previous_text`
    - `save_text`
    - `next_text`
- [ ] Make `forms.FormDefinition` fields translatable:
    - `name`
- [ ] Make `emails.ConfirmationEmailTemplate` fields translatable:
    - `subject`
    - `content`
- [ ] Make `config.GlobalConfiguration` fields translatable:
    - `submission_confirmation_template`
    - `confirmation_email_subject`
    - `confirmation_email_content`
    - `save_form_email_subject`
    - `save_form_email_content`
    - `form_previous_text`
    - `form_change_text`
    - `form_confirm_text`
    - `form_begin_text`
    - `form_step_previous_text`
    - `form_step_save_text`
    - `form_step_next_text`
    - `privacy_policy_label`

PR #2235, opened 2022-10-24
:globe_with_meridians: [#2230] Setup modeltranslations

stevenbal:
Fixes: #2230 

Changes:
* Setup modeltranslations for several models
* Add data migration to update translation fields data


src/openforms/forms/tests/test_migrations.py
```diff
@@ -228,3 +237,33 @@ def test_inline_edit_is_true(self):
...
+                ]
+            },
+        )
+
+    def test_update_translation_fields(self):
+        self.form_definition.refresh_from_db()
+
+        self.assertEqual(self.form_definition.name, "Vertaalbare naam")
+        self.assertEqual(self.form_definition.name_nl, "Vertaalbare naam")
+        self.assertEqual(self.form_definition.name_en, None)
```
stevenbal, line 269: It seems that by default translatable fields are `None`, is this an issue?

src/openforms/forms/migrations/0051_update_translation_fields.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.16 on 2022-10-26 07:09
+
+from django.core.management import call_command
+from django.db import migrations
+
+
+def update_translation_fields(apps, schema_editor):
```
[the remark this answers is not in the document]
sergei-maertens: no it's fine, those new fields/languages can be empty initially. I think at that point we'll also have to implement the feature to specify a subset of languages for a form.

src/openforms/forms/tests/test_migrations.py
sergei-maertens: I don't necessarily like it, but it's part of the 3rd party library implementation details so we'll just have to go along with it. I think we'll be able to solve this at the serializer level to turn it into an empty string.

sergei-maertens (approved):

PR #2236, opened 2022-10-24
:bug: emit openforms-theme as default theme

### LINKED ISSUE #2225: Apply openforms-theme class in backend...
... and switch to class name selector for design tokens rather than `:root`

sergei-maertens:
Fixes #2225

Unless another theme is explicitly specified, emit the openforms-theme class for the design tokens and include the theme-scoped design token CSS variables rather than overriding the :root selector.

Newer SDK builds specify design token values for components other than our own --of prefix, which breaks themes other than our own.


SilviaAmAm (approved):

sergei-maertens:
Backport: 71635b02d3941325809bc258e5cf26bac058afbf

Issue #2197, opened 2022-10-16
Token authorisation


joeribekker:
### Thema / Theme

Plugins

### Omschrijving / Description

Gemeente Den Haag heeft een voorstel gedaan om services (voor nu, enkel de ZGW APIs - Open Zaak) te kunnen bevragen indien een **token** wordt meegestuurd.

Een korte omschrijving van hoe de token exchange in zijn werk zal gaan (oorspronkelijk geschreven voor GZAC maar aangepast op Open Formulieren):

1. Een gebruiker logt in met DigiD/eHerkenning via OpenID-connect (Keycloak)
2. Open Formulieren verkrijgt zo een access-token van Keycloak.
3. Open Formulieren stuurt het access-token van de gebruiker naar de Keycloak token-exchange endpoint
4. Keycloak genereert een nieuw access-token en geeft deze terug.
5. Open Formulieren gebruikt dit nieuwe access-token om Haal Centraal gegevens op te halen of om Open Zaak te benaderen.
6. Haal Centraal of Open Zaak valideren het accesstoken op certificaat en inhoud, en geven indien valide, hun response terug.

### Added value / Toegevoegde waarde

ABAC

### Aanvullende opmerkingen / Additional context

<s>Mogelijke oplossingsrichting:

1. We voegen een soort hook toe aan de plugins, waarmee we kunnen inhaken op de headers (of breder) van de aan te roepen services.
2. We passen de (toch al voor Den Haag specifieke) DigiD/eHerkenning met OIDC-plugins aan. Deze plugin haalt direct een nieuw access-token op. Deze plugin registreert zichzelf op de hook van de andere plugins om het nieuwe access-token mee te sturen.</s>

--- 

**Technical analysis & implementation guidelines**

These guidelines are sourced from the documentation/POC design document (v1.32) and checking the implementation in Valtimo.

In addition, the mechanisms in `requests` and `zds_client` libraries are checked to see how it all fits into our chain.

## Principles that make this work and need to be understood to implement this

* End-user (citizen, company, staff...) authenticate on forms using OpenID Connect protocol, implemented by Keycloak
* On successful authentication, Open Forms receives an (opaque) OAUTH2 access token used to retrieve the user details (cfr. OpenID flow). We do *not* inspect or parse this access tokens, as access tokens are _opaque_ to the client (OAUTH2 principle).
[...]

PR #2237, opened 2022-10-24
[#2197] Token authorisation: pre_request hook

SilviaAmAm:
Part of #2197 

**Changes**
- Add the `pre_requests` registry
- Add a new client which calls the registered pre_requests hooks
- Change the ZGW_CONSUMERS_CLIENT_CLASS setting to point to the new client class

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -231,6 +231,7 @@ INSTALLED_APPS = [
     # Apps registering static variables
     "openforms.variables.static_variables.apps.StaticVariables",
     "openforms.authentication.static_variables.apps.AuthStaticVariables",
+    "openforms.pre_requests",
 ]
 
 MIDDLEWARE = [
@@ -856,7 +857,7 @@ SPECTACULAR_SETTINGS = {
 #
 # ZGW Consumers
 #
-ZGW_CONSUMERS_CLIENT_CLASS = "zgw_consumers.client.ZGWClient"
+ZGW_CONSUMERS_CLIENT_CLASS = "openforms.pre_requests.clients.PreRequestZGWClient"
 ZGW_CONSUMERS_TEST_SCHEMA_DIRS = [
     os.path.join(BASE_DIR, "src/openforms/registrations/contrib/zgw_apis/tests/files"),
     os.path.join(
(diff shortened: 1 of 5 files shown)


src/openforms/pre_requests/base.py
```diff
@@ -0,0 +1,12 @@
...
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class PreRequestHookBase(ABC):
+    identifier: str
+
+    @abstractmethod
+    def __call__(self, method: str, url: str, **kwargs) -> Any:
```
sergei-maertens, line 11: I'd make the return type `None` since we don't do anything with it anyway.

src/openforms/pre_requests/tests/test_clients.py
```diff
@@ -0,0 +1,47 @@
...
+
+class PreRequestHooksTest(TestCase):
+    def test_pre_request_hook(self):
+        register = Registry()
+
+        @register("test-hook")
+        class PreRequestHook(PreRequestHookBase):
+            def __call__(self, url, method, **kwargs):
+                headers = kwargs.get("headers", {})
+                headers.update({"test": "test"})
```
sergei-maertens, line 22: A more realistic example would ensure the headers are set:

```suggestion
            def __call__(self, url, method, **kwargs):
                kwargs.setdefault("headers", {})
                kwargs["headers"].update({"test": "test"})
```

src/openforms/conf/base.py
```diff
@@ -231,6 +231,7 @@
     # Apps registering static variables
     "openforms.variables.static_variables.apps.StaticVariables",
     "openforms.authentication.static_variables.apps.AuthStaticVariables",
+    "openforms.pre_requests",
```
sergei-maertens, line 234: we don't need to add it to `INSTALLED_APPS` since:

* there are no models
* there is no `AppConfig.ready()` that imports/loads code

sergei-maertens (changes requested):

sergei-maertens (approved):
Will you add developer documentation in another PR? :)

SilviaAmAm:
Yes :)

PR #2242, opened 2022-10-24
Feature [#2229] Implement api endpoints for i18n

### LINKED ISSUE #2229: Implement API endpoints for i18n
Epic #2170

- [ ]  Add `GET /api/v1/i18n/info` endpoint, return data:
  ```json
  {
    "languages": [
      {
        "code": "en",
        "name": "English"
      },
      {
        "code": "nl",
        "name": "Nederlands"
      }
    ],
    "current": "nl"
  }
  ```
  * exposes the available languages
  * exposes which language the user has selected (this is based on the `LocaleMiddleware`)
  * in-browser language detection via javascript is unreliable, best to have the backend expose this
  * Take an optional `?hint=` querystring param, which will contain the value from the `html.lang` attribute - use this is fallback (browser preference should always override this if possible).

- [ ] Add `PUT /api/v1/i18n/language` endpoint to set the language, request body `{"code": "nl"}`. This needs to result in the language cookie being set, so check Django's i18n tools for that

CharString:
Unsure about the spec of the 400 on the PUT.  https://docs.geostandaarden.nl/api/API-Strategie-ext/#error-handling-0 is marked not normative and under development.

Closes #2229

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -2405,9 +2405,7 @@ paths:
       operationId: i18n_info_retrieve
       summary: List available languages and the currently active one.
       tags:
-      - i18n
-      security:
-      - {}
+      - translations
       responses:
         '200':
           content:
@@ -2422,8 +2420,6 @@ paths:
                       name: English
                     - code: nl
                       name: Nederlands
-                    - code: fy
-                      name: frysk
                     current: nl
                   summary: Selected Dutch
           description: ''
@@ -2439,7 +2435,7 @@ paths:
       operationId: i18n_language_update
       summary: Set the desired language
       tags:
-      - i18n
+      - translations
       requestBody:
         content:
           application/json:
@@ -2452,8 +2448,6 @@ paths:
             schema:
               $ref: '#/components/schemas/LanguageCode'
         required: true
-      security:
-      - {}
       responses:
         '204':
           description: No response body
@@ -2469,6 +2463,19 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ValidationError'
+          description: ''
+          headers:
+            X-Session-Expires-In:
+              $ref: '#/components/headers/X-Session-Expires-In'
+            X-CSRFToken:
+              $ref: '#/components/headers/X-CSRFToken'
+            X-Is-Form-Designer:
+              $ref: '#/components/headers/X-Is-Form-Designer'
+        5XX:
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Exception'
           description: ''
           headers:
             X-Session-Expires-In:
@@ -4764,8 +4771,8 @@ components:
       type: string
     AvailableLanguagesEnum:
       enum:
-      - en
       - nl
+      - en
       type: string
     BlankEnum:
       enum:
@@ -5572,7 +5579,7 @@ components:
         code:
           allOf:
           - $ref: '#/components/schemas/AvailableLanguagesEnum'
-          description: RFC5646 language tag, e.g. 'en' or 'en-us'
+          description: RFC5646 language tag, e.g. `en` or `en-us`
         name:
           type: string
           description: Language name in its local representation. e.g. "fy" = "frysk",
@@ -5586,7 +5593,7 @@ components:
         code:
           allOf:
           - $ref: '#/components/schemas/AvailableLanguagesEnum'
-          description: RFC5646 language tag, e.g. 'en' or 'en-us'
+          description: RFC5646 language tag, e.g. `en` or `en-us`
       required:
       - code
     LanguageInfo:
@@ -5601,7 +5608,7 @@ components:
         current:
           allOf:
           - $ref: '#/components/schemas/AvailableLanguagesEnum'
-          description: RFC5646 language tag, e.g. 'en' or 'en-us'
+          description: RFC5646 language tag, e.g. `en` or `en-us`
       required:
       - current
       - languages
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -2400,6 +2400,83 @@ paths:
               $ref: '#/components/headers/X-CSRFToken'
             X-Is-Form-Designer:
               $ref: '#/components/headers/X-Is-Form-Designer'
+  /api/v2/i18n/info:
+    get:
+      operationId: i18n_info_retrieve
+      summary: List available languages and the currently active one.
+      tags:
+      - i18n
+      security:
+      - {}
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/LanguageInfo'
+              examples:
+                SelectedDutch:
+                  value:
+                    languages:
+                    - code: en
+                      name: English
+                    - code: nl
+                      name: Nederlands
+                    - code: fy
+                      name: frysk
+                    current: nl
+                  summary: Selected Dutch
+          description: ''
+          headers:
+            X-Session-Expires-In:
+              $ref: '#/components/headers/X-Session-Expires-In'
+            X-CSRFToken:
+              $ref: '#/components/headers/X-CSRFToken'
+            X-Is-Form-Designer:
+              $ref: '#/components/headers/X-Is-Form-Designer'
+  /api/v2/i18n/language:
+    put:
+      operationId: i18n_language_update
+      summary: Set the desired language
+      tags:
+      - i18n
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/LanguageCode'
+          application/x-www-form-urlencoded:
+            schema:
+              $ref: '#/components/schemas/LanguageCode'
+          multipart/form-data:
+            schema:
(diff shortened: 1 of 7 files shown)


src/openforms/conf/base.py
```diff
@@ -850,6 +850,7 @@
     },
     "ENUM_NAME_OVERRIDES": {
         "IncompleteSubmissionsRemovalMethodEnum": "openforms.data_removal.constants.RemovalMethods",
+        "AvailableLanguagesEnum": "django.conf.settings.LANGUAGES",
```
sergei-maertens, line 853: I saw the warning in the failing CI build, but I'm curious if this is still needed when a different serializer + custom field is used for the language code choices

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+from django.test import override_settings
+from django.urls import reverse
+from django.utils.translation import gettext as _
+
+from hypothesis import given, settings, strategies as st
+
+# from hypothesis.extra.django import TestCase
+from rest_framework.test import APISimpleTestCase
+
+Language = namedtuple("Language", "conf local")
```
sergei-maertens, line 12: now that dataclasses exist, let's use them rather than namedtuples

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class I18NAPITests(APISimpleTestCase):
+    databases = ["default"]
```
sergei-maertens, line 16: please don't override settings/configurations that are already the default. it's highly unlikely we'll ever have a multi-db setup

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+                    Language(("tr", _("Turkish")), "Türkçe"),
+                ]
+            ),
+            unique=True,
+        )
+    )
+    @settings(deadline=500)
+    def test_info_contains_available_langauges(self, languages):
+        with override_settings(
+            USE_I18N=True,
```
sergei-maertens, line 33: don't override settings that are already the current value (and cannot be altered by the env):

* changing the setting from `True` to `False` would break the functionality at runtime for real deploys
* having it overridden here would *not* point out this misconfiguration. if someone messes up with the (default) settings, CI needs to fail

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+                [
+                    Language(("en", _("English")), "English"),
+                    Language(("nl", _("Dutch")), "Nederlands"),
+                    Language(("tr", _("Turkish")), "Türkçe"),
+                ]
+            ),
+            unique=True,
+        )
+    )
+    @settings(deadline=500)
```
sergei-maertens, line 30: I've looked a bit more into hypothesis and while I do see the value in it, I'm afraid it's a bit too opaque. Nobody wants to be thinking about or debugging tests to figure out why they are failing, instead, you want to be debugging your actual useful code. We apply the principle that tests should be stupidly easy to follow:

1. data setup
2. call the thing being tested
3. make assertions

This implies:
* avoid putting data/state setup in `setUp` methods or private helper methods -> to understand a failing test, ideally you want to look at only the failing test code without jumping around
* DO repeat yourself - linear straightforward code is easy to follow
* use explicit parametrization if needed and use `self.subTest` with sufficient/relevant context

I find I have to really think here to understand what this test is actually doing, and that hurts maintainability. So please make this a test without hypothesis usage. Instead, we can structure the test more like:

```py
    @override_settings(
        LANGUAGES=(
            ("en", _("English")),
            ("nl-nl", _("Dutch")),
            ("nl-be", _("Flemish")),
        ),
        LANGUAGE_CODE="nl-nl",
    )
    def test_language_info_endpoint(self):
        url = reverse("api:i18n:info")

        response = self.client.get(url)

        expected = {
            "languages": [
                {
                    "code": "en",
                    "name": "English",
                },
                {
                    "code": "nl-nl",
                    "name": "Nederlands",
                },
                {
                    "code": "nl-be",
                    "name": "Vlaams",
                },
            ],
            "current": "nl-nl",
        }
        self.assertEqual(response.json(), expected)
```

This implies:

* fixed order from the settings
* isolated the test from runtime configuration by specifying a _different_ `LANGUAGES` setting and pinning the `LANGUAGE_CODE`

You could also add a smoke test that runs without `@override_settings`, but that's a bit overkill probably.

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+            USE_I18N=True,
+            LANGUAGES=[lang.conf for lang in languages],
+        ):
+            # expect language names in their local representation
+            expected = [
+                {"code": lang.conf[0], "name": lang.local} for lang in languages
+            ]
+            url = reverse("api:i18n:info")
+            response = self.client.get(url)
+            available = response.json()["langauges"]
```
sergei-maertens, line 42: typo is propagated into the test

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+            ("en", _("English")),
+            ("nl", _("Dutch")),
+        ],
+    )
+    def test_info_contains_active_language(self):
+        # default fallback
+        url = reverse("api:i18n:info")
+        response = self.client.get(url)
+        current = response.json()["current"]
+        self.assertEqual(current, "en")
```
sergei-maertens, line 58: we also try to visually structure the tests:

```
# set up test data/state 1
# set up test data/state 2

# call thing being tested

# make assertion 1
# make assertion 2
```
with blank lines to break up the three "blocks"

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+        LANGUAGE_CODE="en",
+        LANGUAGES=[
+            ("en", _("English")),
+            ("nl", _("Dutch")),
+        ],
+        MIDDLEWARE=[
+            "django.contrib.sessions.middleware.SessionMiddleware",
+            "django.middleware.locale.LocaleMiddleware",
+            "django.middleware.common.CommonMiddleware",
+        ],
```
sergei-maertens, line 71: you can leave this out to test our settings/configuration isn't missing a/the middleware (Steven's PR got merged that enabled the middleware, so you can rebase this)

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    def test_put_overrides_negotiated_current(self):
+        SELECT = "en"
+        url = reverse("api:i18n:language")
+        response = self.client.put(
+            url,
+            data={"code": SELECT},
+            HTTP_ACCEPT_LANGUAGE="nl-BE, nl;q=0.9, en;q=0.8",
+        )
```
sergei-maertens, line 102: use literal values in test to remove indirection and make the test easy to understand (so DO repeat yourself ;) )

src/openforms/translations/tests/test_api.py
```diff
@@ -0,0 +1,131 @@
...
+            "django.contrib.sessions.middleware.SessionMiddleware",
+            "django.middleware.locale.LocaleMiddleware",
+            "django.middleware.common.CommonMiddleware",
+        ],
+    )
+    def test_put_language_checks_for_availability(self):
+        url = reverse("api:i18n:language")
+        # ትግርኛ is not available
+        response = self.client.put(url, data={"code": "ti"})
+        self.assertEqual(response.status_code, 400)
```
sergei-maertens, line 131: you'll also want to make an assertion that the `name` and `code` of `invalidParams[0]` is as expected

sergei-maertens (changes requested):
Please also rebase on master to gets Steven's work w/r to the settings available for your tests.

src/openforms/conf/base.py
CharString: That's based on the value of the enum, not the name. It warns. But in this case it's valid, they are after all the same Enum i.e. the available languages.

src/openapi.yaml
```diff
@@ -2400,6 +2400,85 @@ paths:
...
+  /api/v1/i18n/info:
+    get:
+      operationId: i18n_info_retrieve
+      summary: Available languages and the currently selected one.
+      tags:
+      - i18n
+      security:
+      - tokenAuth: []
+      - cookieAuth: []
+      - {}
```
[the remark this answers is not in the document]
CharString: It's empty now.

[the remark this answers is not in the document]
sergei-maertens: I found this is coming from https://github.com/tfranzel/drf-spectacular/blob/edab4be1b1499cb032bab09e87ee9eeb3a1f61e1/drf_spectacular/openapi.py#L339 and this structure does not seem to be entirely in line with the OpenAPI 3 specification/intent. You can get rid of it by declaring `permission_classes = ()` as an empty tuple.

It's not crucial for us now and doesn't affect runtime behaviour, but at some point we'll have to look into contributing back to drf-spectacular to get this properly sorted.

sergei-maertens (approved):

PR #2249, opened 2022-10-26
[#2197] Developers documentation token exchange

SilviaAmAm:
Part of #2197

### DIFF
--- a/docs/developers/extensions.rst
+++ b/docs/developers/extensions.rst
@@ -8,9 +8,10 @@ Keycloak token exchange extension
 =================================
 
 The municipality of Den Haag started a proof of concept for the addition of Attribute Based Access Control (ABAC) to the
-ZGW APIs. This means that an authenticated user should only be able to retrieve data from the ZGW APIs which they are
-entitled to see. In the proof of concept, the Keycloak preview feature called "`token exchange`_" is leveraged to get a
-OAUTH2 access token that is then sent to the downstream APIs to determine what the user is allowed to retrieve.
+ZGW APIs. This means that an authenticated user should only be able to retrieve data from external services (such as
+Haal Centraal BRP bevragen or retrieving case data from the ZGW APIs) which they are entitled to see.
+In the proof of concept, the Keycloak preview feature called "`token exchange`_" is leveraged to get a
+(OAUTH2) access token that is then sent to the downstream APIs to determine what the user is allowed to retrieve.
 
 Open Forms makes requests to APIs like, for example, Haal Centraal when retrieving user data for the prefill at the
 start of a submission. To work with Den Haag's proof of concept, we need to obtain the access token from Keycloak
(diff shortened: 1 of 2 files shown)


docs/developers/extensions.rst
```diff
@@ -0,0 +1,49 @@
...
+==========
+Extensions
+==========
+
+Keycloak token exchange extension
+=================================
+
+The municipality of Den Haag started a proof of concept for the addition of Attribute Based Access Control (ABAC) to the
+ZGW APIs. This means that an authenticated user should only be able to retrieve data from the ZGW APIs which they are
+entitled to see. In the proof of concept, the Keycloak preview feature called "`token exchange`_" is leveraged to get a
```
sergei-maertens, line 12:
```suggestion
ZGW APIs. This means that an authenticated user should only be able to retrieve data from external services (such as Haal Centraal BRP bevragen or retrieving case data from the ZGW APIs) which they are
entitled to see. In the proof of concept, the Keycloak preview feature called "`token exchange`_" is leveraged to get a
```

docs/developers/extensions.rst
```diff
@@ -0,0 +1,49 @@
...
+OAUTH2 access token that is then sent to the downstream APIs to determine what the user is allowed to retrieve.
```
sergei-maertens, line 13:
```suggestion
(OAUTH2) access token that is then sent to the downstream APIs to determine what the user is allowed to retrieve.
```

it doesn't necessarily have to be an OAUTH access token, as long as it uses a Bearer token in the `Authorization` header, and even then there's some flexibility to use different headers or token formats (at least on our end).

If keycloak is configured to return a ZGW API's JWT, we can use that as is!

docs/developers/extensions.rst
```diff
@@ -0,0 +1,49 @@
...
+to send it to Haal Centraal. Because the token exchange is not a Keycloak core feature (disabled by default), this
+process was implemented in Open Forms as an extension: `open-forms-ext-token-exchange`_.
+
+The flow works as follows:
+
+#. A user logs in with DigiD/eHerkenning/eIDAS/DigiD Machtigen/eHerkenning Bewindvoering through Keycloak (OpenID
+   connect). Keycloak returns a token in the payload along with the BSN/KvK/Pseudo ID.
+
+   * The authentication plugins in Open Forms extract the user information from the payload and add it to the session
+     under the :class:`openforms.authentication.constants.FORM_AUTH_SESSION_KEY` key.
```
sergei-maertens, line 26:
```suggestion
     under the :class:`openforms.authentication.constants.FORM_AUTH_SESSION_KEY` key.
```
this would rather be `attribute` or `param` or something like that, instead of `class` (as it is just a module attribute)

docs/developers/extensions.rst
```diff
@@ -0,0 +1,49 @@
...
+   * The :class:`AuthenticationReturnView` of the authentication plugins fires a signal to notify that the authentication
+     was successful if the :class:`openforms.authentication.constants.FORM_AUTH_SESSION_KEY` key is present in the session.
+   * The ``open-forms-ext-token-exchange`` extension receives the signal and extracts the Keycloak token from the request.
+     This token is saved in thread local memory.
```
sergei-maertens, line 30:
```suggestion
   * The ``open-forms-ext-token-exchange`` extension receives the signal and extracts the Keycloak token from the request.
     This token is kept in application memory for later access (and never persisted to disk to avoid leaking credentials _if_ the application were to be compromised).
```

docs/developers/extensions.rst
```diff
@@ -0,0 +1,49 @@
...
+
+#. The submission is started and Open Forms requests prefill data from Haal Centraal:
+
+   * The client that makes the request to the ZGW API is :class:`openforms.pre_requests.clients.PreRequestZGWClient`.
+     This client is a subclass of :class:`zgw_consumers.client.ZGWClient` which overrides the ``pre_request`` method so
+     that any pre-request hooks registered in Open Forms are run before performing the request.
+   * The ``open-forms-ext-token-exchange`` extension registers a pre-request hook which adds a
+     `custom authentication class`_ to the request.
+   * The custom authentication class :class:`token_exchange.auth.TokenAccessAuth` checks if the URL to which the
+     request is being made has a service with an associated :class:`token_exchange.models.TokenExchangeConfiguration`.
```
sergei-maertens, line 40: `token_exchange` references will never resolve/link because they aren't part of our codebase or dependencies. You can simply use double backticks instead.

Maybe later when we're all grown up we'll set up intersphinx to link to external API documentation.

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #2232, opened 2022-10-24
Add toggle to enable/disable translations per form


sergei-maertens:
Epic #2170

---

Migration wise best to implement this after #2230 

---

**Tasks**

- [x] Add a new field to the `Form` model: `i18n_enabled` - default to `False` for backwards compatibility
- [x] Expose the field in the public API endpoint (the SDK needs to be aware of this configuration)
- [x] Ensure that when i18n is disabled, the language is _forced_ to `settings.LANGUAGE_CODE` (i.e. `HTTP_LANGUAGE` and/or language cookie are ignored)

PR #2261, opened 2022-10-31
:globe_with_meridians: [#2232] Add Form.i18n_enabled to model and API

stevenbal:
Fixes: #2232 

Requires https://github.com/open-formulieren/open-forms-sdk/pull/272 to be merged as well

Changes:
* Add Form.i18n_enabled to model and API
* Force the language to the default if translations are not enabled


src/openforms/forms/api/viewsets.py
```diff
@@ -388,6 +391,45 @@ def export(self, request, *args, **kwargs):
...
+                response=[status.HTTP_201_CREATED],
+            ),
+        ],
+    )
+    @transaction.atomic
+    @action(
+        detail=True,
+        methods=["post"],
+        authentication_classes=(),
+        permission_classes=(),
```
stevenbal, line 415: This endpoint probably shouldn't require the user to be authenticated/have permissions?

src/openforms/utils/tests/test_migrations.py
```diff
@@ -13,13 +13,14 @@ class TestMigrations(TestCase):
...
     migrate_to = None
+    extra_from_migrations = []
 
     def setUp(self):
         assert self.migrate_from and self.migrate_to and self.app, (
             "TestCase '%s' must define migrate_from, migrate_to and app properties"
             % type(self).__name__
         )
-        self.migrate_from = [(self.app, self.migrate_from)]
+        self.migrate_from = [(self.app, self.migrate_from)] + self.extra_from_migrations
```
stevenbal, line 23: I'm not sure about this, but it seems that when migrating to `migrate_from` the database state is properly migrated, but the models in the apps registry (retrieved in line 26) apparently not, causing issues: https://github.com/open-formulieren/open-forms/pull/2261/files#diff-c301558f7c67e1e1cd54331d556c2a6f4476a25ca524a5ac81d81a720b3e8245R26

sergei-maertens: I made some changes to this in effb84f2750a67a1be9d1c1d88cdc4f173597fb8, so I'd like to see if that resolves it.

src/openforms/forms/api/viewsets.py
sergei-maertens: indeed not, because they typically wouldn't have started a submission yet :+1:

src/openforms/forms/api/viewsets.py
```diff
@@ -388,6 +391,45 @@ def export(self, request, *args, **kwargs):
...
+        In case translation is not enabled for the Form, the default language is set to
+        settings.LANGUAGE_CODE.
+        """
+        response = Response(status=status.HTTP_204_NO_CONTENT)
+        instance = self.get_object()
+
+        if not instance.translation_enabled:
+            translation.activate(settings.LANGUAGE_CODE)
+            response = set_language_cookie(response, settings.LANGUAGE_CODE)
+
```
sergei-maertens, line 430: I think we want to reply to the frontend with what the currently active language is so the internal state can be updated and the appropriate frontend translations loaded, without having to make an additional API call to retrieve the current language.

sergei-maertens (changes requested):

src/openforms/utils/tests/test_migrations.py
stevenbal: Works!

src/openforms/forms/api/viewsets.py
stevenbal: Added in https://github.com/open-formulieren/open-forms/pull/2261/commits/58950cee3e76ac98cc7d4c0238d8df673f43cdcb

By the way, do you prefer separate commits for PR feedback, or rebasing and adding changes to the existing commits?

sergei-maertens: I don't really have a preference except for when a fix needs to be backported, then definitely rebasing.

sergei-maertens (changes requested):

src/openforms/forms/api/viewsets.py
```diff
@@ -388,6 +391,47 @@ def export(self, request, *args, **kwargs):
...
+        responses={status.HTTP_201_CREATED: FormSerializer},
+        parameters=[
+            UUID_OR_SLUG_PARAMETER,
+            OpenApiParameter(
+                name="Location",
+                type=OpenApiTypes.URI,
+                location=OpenApiParameter.HEADER,
+                description="URL of the created resource",
+                response=[status.HTTP_201_CREATED],
+            ),
```
sergei-maertens, line 407: we're not emiting a HTTP 201 and there's no resource to redirect to

```suggestion
```

src/openforms/forms/api/viewsets.py
```diff
@@ -388,6 +391,47 @@ def export(self, request, *args, **kwargs):
         response["Content-Length"] = len(response.content)
         return response
 
+    @extend_schema(
+        summary=_("Set default language for form"),
+        tags=["forms"],
+        request=None,
+        responses={status.HTTP_201_CREATED: FormSerializer},
```
sergei-maertens, line 398: you'll need to use/define a proper serializer for the response data here, and use HTTP 200 instead of 201

src/openforms/forms/api/viewsets.py
```diff
@@ -388,6 +391,47 @@ def export(self, request, *args, **kwargs):
...
+        settings.LANGUAGE_CODE.
+        """
+        response = Response(status=status.HTTP_200_OK)
+        instance = self.get_object()
+
+        if not instance.translation_enabled:
+            translation.activate(settings.LANGUAGE_CODE)
+            response = set_language_cookie(response, settings.LANGUAGE_CODE)
+
+        response.data = {"active_language": translation.get_language()}
```
sergei-maertens, line 431: I'd prefer to see a real serializer used for the response data to keep the implementation/documentation together.

Using https://github.com/open-formulieren/open-forms/blob/master/src/openforms/translations/api/serializers.py:

```py
class CurrentLanguageSerializer(serializers.Serializer):
    active_language = LanguageCodeField()


def activate_default_language(self, request, *args, **kwargs):
    form = self.get_object()
    response = Response(status=status.HTTP_200_OK)
    current_language = translation.get_language()
    if not form.translation_enabled:
        translation.activate(settings.LANGUAGE_CODE)
        current_language = translation.get_language()
        set_language_cookie(response, current_language)

    response.data = CurrentLanguageSerializer(instance={"active_language": current_language}).data
    return response
```

src/openforms/forms/tests/test_api_forms.py
```diff
@@ -1120,3 +1122,51 @@ def test_detail_shows_translated_values_based_on_cookie(self):
...
+        response = self.client.post(url, HTTP_ACCEPT_LANGUAGE="en")
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, {"active_language": "nl"})
+        self.assertIn(settings.LANGUAGE_COOKIE_NAME, response.cookies)
+
+        language_cookie = response.cookies[settings.LANGUAGE_COOKIE_NAME]
+
+        self.assertEqual(language_cookie.key, settings.LANGUAGE_COOKIE_NAME)
+        self.assertEqual(language_cookie.value, settings.LANGUAGE_CODE)
```
sergei-maertens, line 1145:
```suggestion
        self.assertEqual(language_cookie.value, "nl")
```
let's make this explicit, if we change the `LANGUAGE_CODE` setting, we'll have weird results and there's misaligned because the `HTTP_ACCEPT_LANGUAGE` header would now suddenly match.

There's a chance in the future we allow configuring the default language via an environment variable.

src/openforms/forms/tests/test_api_forms.py
```diff
@@ -1120,3 +1122,51 @@ def test_detail_shows_translated_values_based_on_cookie(self):
...
+        self.assertEqual(language_cookie["expires"], "")
+        self.assertEqual(language_cookie["path"], "/")
+        self.assertEqual(language_cookie["comment"], "")
+        self.assertEqual(language_cookie["domain"], "")
+        self.assertEqual(language_cookie["max-age"], "")
+        self.assertEqual(language_cookie["secure"], "")
```
sergei-maertens, line 1151: that's a bit odd since the value is coupled to the `IS_HTTPS` setting - I don't think we need to explicitly test this

src/openforms/forms/tests/test_api_forms.py
```diff
@@ -1120,3 +1122,51 @@ def test_detail_shows_translated_values_based_on_cookie(self):
...
+        """
+        Default language should be set, because translation is disabled (ignoring any language headers)
+        """
+        form = FormFactory.create(translation_enabled=False)
+
+        url = reverse(
+            "api:form-set-default-language", kwargs={"uuid_or_slug": form.uuid}
+        )
+        self.client.force_authenticate(user=None)
+        response = self.client.post(url, HTTP_ACCEPT_LANGUAGE="en")
```
sergei-maertens, line 1136:
```suggestion

        response = self.client.post(url, HTTP_ACCEPT_LANGUAGE="en")
```
stylistic nitpicking to break up visual blocks in tests:

```
# 1. test data setup
# blank line
# 2. call the thing you're testing
# blank line
# 3. make assertions
```

src/openforms/translations/api/views.py
```diff
@@ -78,14 +79,5 @@ def put(self, request: Request) -> Response:
...
-            key=settings.LANGUAGE_COOKIE_NAME,
-            value=serializer.data["code"],
-            max_age=settings.LANGUAGE_COOKIE_AGE,
-            domain=settings.LANGUAGE_COOKIE_DOMAIN,
-            httponly=settings.LANGUAGE_COOKIE_HTTPONLY,
-            path=settings.LANGUAGE_COOKIE_PATH,
-            samesite=settings.LANGUAGE_COOKIE_SAMESITE,
-            secure=settings.LANGUAGE_COOKIE_SECURE,
-        )
+        response = set_language_cookie(response, serializer.data["code"])
```
sergei-maertens, line 82:
```suggestion
        set_language_cookie(response, serializer.data["code"])
```

src/openforms/utils/translations.py
```diff
@@ -44,3 +46,17 @@ def __enter__(self):
...
+        key=settings.LANGUAGE_COOKIE_NAME,
+        value=language_code,
+        max_age=settings.LANGUAGE_COOKIE_AGE,
+        domain=settings.LANGUAGE_COOKIE_DOMAIN,
+        httponly=settings.LANGUAGE_COOKIE_HTTPONLY,
+        path=settings.LANGUAGE_COOKIE_PATH,
+        samesite=settings.LANGUAGE_COOKIE_SAMESITE,
+        secure=settings.LANGUAGE_COOKIE_SECURE,
+    )
+    return response
```
sergei-maertens, line 62:
```suggestion
def set_language_cookie(response: Response, language_code: str) -> None:
    response.set_cookie(
        key=settings.LANGUAGE_COOKIE_NAME,
        value=language_code,
        max_age=settings.LANGUAGE_COOKIE_AGE,
        domain=settings.LANGUAGE_COOKIE_DOMAIN,
        httponly=settings.LANGUAGE_COOKIE_HTTPONLY,
        path=settings.LANGUAGE_COOKIE_PATH,
        samesite=settings.LANGUAGE_COOKIE_SAMESITE,
        secure=settings.LANGUAGE_COOKIE_SECURE,
    )
```
I wouldn't return anything because that sort of gives the impression that this function does not have side-effects, but the `set_cookie` method only produces side effects.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2263, opened 2022-10-31
Upgrade zgw-consumers and gemma-zds-client

SilviaAmAm:


src/openforms/pre_requests/clients.py
```diff
@@ -1,15 +1,17 @@
...
 from .registry import register as registry
 
 
 class PreRequestZGWClient(ZGWClient):
-    def pre_request(self, method: str, url: str, **kwargs) -> Any:
-        result = super().pre_request(method, url, **kwargs)
+    def pre_request(
+        self, method: str, url: str, kwargs: Optional[dict] = None, **old_kwargs
+    ) -> Any:
+        result = super().pre_request(method, url, kwargs, **old_kwargs)
```
sergei-maertens, line 12: you can drop the `old_kwargs` I think, that's just a compatibility layer.

sergei-maertens (changes requested):

Issue #2278, opened 2022-11-01
Add Submission.language_code field which stores the language the submission is created in


stevenbal:
Epic https://github.com/open-formulieren/open-forms/issues/2170

PR #2281, opened 2022-11-01
:globe_with_meridians: [#2278] Add Submission.language_code

stevenbal:
Closes: #2278


src/openforms/submissions/models/submission.py
```diff
@@ -241,6 +242,15 @@ class Submission(models.Model):
         "submissions.Submission", on_delete=models.SET_NULL, null=True, blank=True
     )
 
+    language_code = models.CharField(
+        _("language code"),
+        max_length=2,
+        default=settings.LANGUAGE_CODE,
```
stevenbal, line 248: Not sure if we want a default or if we want it to be explicitly filled (with a data migration to account for existing `Submission`s)

sergei-maertens: Django might populate existing data with the `default`, but I'll have to check.

Old submissions are pruned anyway, so I think we can leave it empty. This would then only apply to the edge case where the language code is included to the registration backend and has an empty value, so as long as we have an explicit test case for that behaviour, it should be fine.

src/openforms/submissions/models/submission.py
```diff
@@ -241,6 +242,15 @@ class Submission(models.Model):
...
+        help_text=_(
+            "The code (ISO 639-1 format) of the language used to fill in the Form."
```
sergei-maertens, line 250: As remarks in Chris' PR, this is not an ISO-639-1 code. he found out it's RFC5646 :)

sergei-maertens (changes requested):

src/openforms/submissions/models/submission.py
stevenbal: So no default and set `blank=True`? What would we want to test in that test case, that it is not possible to complete a submission with an empty language code?

src/openforms/submissions/migrations/0067_alter_submission_language_code.py
```diff
@@ -0,0 +1,25 @@
...
+        ("submissions", "0066_remove_submissionvaluevariable_language"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="submission",
+            name="language_code",
+            field=models.CharField(
+                choices=[("nl", "Dutch"), ("en", "English")],
+                default=django.utils.translation.get_language,
```
stevenbal, line 19: I think this migration is needed in addition to `0065` to ensure that existing submissions have their language_code set to `""`

sergei-maertens: exactly my train of thoughts :+1:

PR #2290, opened 2022-11-02
[#1901] Run CSP postprocessor on content component's html to fix inline styles

### LINKED ISSUE #1901: Size of pictures inside free text components doesn't get respected

1.1.5

When setting the size of a picture through the GUI of the 'free text' component...
image

It gets respected in the form-editor view:
image

But not in the form itself:
image

1. Go to https://haarlemmermeer.test.open-formulieren.nl/admin/forms/form/95/change/
2. Evaluate how the size of the pictures is set
3. Compare with resulting form: https://haarlemmermeer.test.open-formulieren.nl/erker/stap/erker

The size of pictures in the forms should match how they were set in the form-editor.

### Screen resolution

_No response_

### Device

Desktop

### OS

Windows

### Browser

Chrome, Firefox

Bartvaderkin:
Fixes #1901 (the resize operation created an inline style attribute)

### DIFF
--- a/src/csp_post_processor/processor.py
+++ b/src/csp_post_processor/processor.py
@@ -1,12 +1,13 @@
 import hashlib
 import logging
 from collections import defaultdict
-from typing import Union
+from typing import List, Optional, Union
 
 from django.http import HttpRequest
 
 import html5lib
 import lxml.html
+import tinycss2
 from lxml import etree
 from rest_framework.request import Request
 
@@ -19,7 +20,12 @@ def get_html_id(node):
     return str(id(node))  # CPython: memory address, so should be unique enough
 
 
-def post_process_html(html: str, request: Union[HttpRequest, Request]) -> str:
+def post_process_html(
+    html: str,
+    request: Union[HttpRequest, Request],
+    *,
+    allowed_css_declarations: Optional[List[str]] = None,
+) -> str:
     """
     Replacing inline style attributes with an inline <style> element with nonce added.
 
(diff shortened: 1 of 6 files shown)


src/openforms/formio/service.py
```diff
@@ -86,3 +88,8 @@ def update_default_file_types(component: Component, **kwargs) -> None:
     if component.get("type") == "file" and component.get("useConfigFiletypes"):
         config = GlobalConfiguration.get_solo()
         component["filePattern"] = ",".join(config.form_upload_default_file_types)
+
+
+def update_content_inline_csp(component: Component, request: Request) -> None:
+    if component.get("type") == "content":
+        component["html"] = post_process_html(component["html"], request)
```
sergei-maertens, line 95: I'd probably extend this post processor to accept an allow-list of CSS attributes to extract - reason being is that content components are not purely "trusted" content from form-designers, but can contain malicious user input if the form designer users variables inside the HTML (and the form submission data is passed as template context to those HTML blobs).

This should all be escaped by default by Django, but it's possible form designers disable auto-escape.

Bartvaderkin: Cool, I added a PR with some tinycss2.

src/csp_post_processor/processor.py
```diff
@@ -1,12 +1,13 @@
...
 import logging
 from collections import defaultdict
-from typing import Union
+from typing import List, Optional, Union
 
 from django.http import HttpRequest
 
 import html5lib
 import lxml.html
+import tinycss2
```
sergei-maertens, line 10: we should make this dependency explicit - it's installed via WeasyPrint, but now it's a first part dependency.

sergei-maertens (approved):

PR #2334, opened 2022-11-10
[#2304] Convert form step URLs in UUIDs

### LINKED ISSUE #2304: Upgrade from 2.0.0-rc.0 to 2.0.0 breaks logic rules with links to form steps

2.0.0-rc.0 -> 2.0.0

In 2.0.0-rc.0, the api path had `v1`. 

Forms created with logic rules with the action "mark form step as not-applicable" had a link to the form step. This URL contains `v1` in the path.

In 2.0.0, the api path has `v2`.

When upgrading, the URL of the form steps in the logic rules is not updated to `v2`. This breaks the logic rule. It also blocks importing/exporting the form (it gives a `form_step: 'ongeldige hyperlink` error).

Checkout `2.0.0-rc.0`
- create a new form, with a step and a radio field with values `a` and `b`.
- Create a logic rule with trigger If **radio > is equal to > a** and action **mark step as not applicable > select step created above**
- Save form

Checkout `2.0.0`
- Look at the action of the logic rule. It can't find the form step anymore. If you look at the URL, it still contains `v1`.

The URL in the action should automatically be converted to `v2`

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #2304


src/openforms/forms/api/serializers/logic/fields.py
```diff
@@ -0,0 +1,30 @@
...
+class ActionFormStepUUIDField(UUIDField):
+    """
+    UUID Field that checks the validity of the UUID but accepts empty strings (since not all rules actions need to have
+    a UUID specified for the form_step field.
+
+    It also doesn't convert the UUID to a UUID object, since this is not JSON serializable and the action is a JSON
+    field.
+    """
+
+    def to_internal_value(self, data):
```
sergei-maertens, line 15: looks like a lot of code is taken from drf - can't we lift more on their implementation?

Something like:

```py
value = super().to_internal_value(data)
return str(value)
```

alternatively, you can also specify a different JSON encoder on the model field, and then you can use django's JSONEncoder which understands the UUID type

(I'd rather not do that now because of backporting migrations woes, but a separate PR would maybe be in order for that?)

src/openforms/forms/api/serializers/logic/validators.py
```diff
@@ -0,0 +1,25 @@
...
+    def __call__(self, step_uuid: UUID):
+        if step_uuid == "":
+            return
+
+        try:
+            FormStep.objects.get(uuid=step_uuid)
+        except ObjectDoesNotExist:
+            raise serializers.ValidationError(
+                _("Invalid form step specified in logic action."), code="invalid"
+            )
```
sergei-maertens, line 25: this will be a performance nightmare with the bulk endpoint - can you look into the custom field that fetches data from the serializer context and use a similar approach (separate commit please, so we can still backport the 'naive' fix).

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -103,19 +102,16 @@ class LogicComponentActionSerializer(serializers.Serializer):
             )
         ),
     )
-    form_step = URLRelatedField(
+    form_step = ActionFormStepUUIDField(
```
sergei-maertens, line 105: I guess this technically is a breaking change in the API - maybe we should keep the old field and deprecate it, and add a new field `form_step_uuid`?

sergei-maertens (changes requested):

src/openforms/forms/api/serializers/logic/fields.py
SilviaAmAm: There is still the problem of the empty UUIDs, so I will go for the first option!

src/openforms/forms/api/serializers/logic/fields.py
```diff
@@ -0,0 +1,31 @@
...
+        value = str(super().to_internal_value(data))
+
+        # Check that the form step UUID matches a UUID of a form step
+        # related to the form
+        related_form_steps = self.parent.context["form_steps"]
+        if value not in related_form_steps:
+            self.error_messages["invalid"] = _(
+                "Invalid form step specified in logic action."
+            )
+            self.fail("invalid", value=value)
```
sergei-maertens, line 29: I'd rather see `default_error_messages` extended on the class level (DRF merges them all together) with a different key and then fail with that key:

```suggestion
            self.fail("invalid-step", value=value)
```

src/openapi.yaml
```diff
@@ -5753,8 +5753,17 @@ components:
           type: string
           format: uri
           nullable: true
-          description: De formulierstap die wordt beïnvloed door de actie. Dit veld
-            is verplicht als het actietype `step-not-applicable` is, anders optioneel.
+          description: '**DEPRECATED** The form step that will be affected by the
+            action. This field is required if the action type is `step-not-applicable`,
+            otherwise optional.'
```
sergei-maertens, line 5758: I think we can add a drf-spectacular decorator to annotate this properly

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -151,10 +163,14 @@ def validate(self, data: dict) -> dict:
...
-            and not form_step
+            and (not form_step and not form_step_uuid)
         ):
             raise serializers.ValidationError(
-                {"form_step": self.fields["form_step"].error_messages["null"]},
+                {
+                    "form_step_uuid": self.fields["form_step_uuid"].error_messages[
+                        "null"
+                    ]
+                },
```
sergei-maertens, line 173: Probably if `form_step` is provided but `form_step_uuid` isn't, we need to convert the former into the latter in the data (for backwards compatibility reasons)

edit: I see that `src/openforms/submissions/logic/actions.py` handles both old and new situation

src/openforms/forms/migrations/0054_replace_urls_with_uuids.py
```diff
@@ -0,0 +1,43 @@
...
+                action["form_step_uuid"] = form_step_uuid
+                rule_modified = True
+
+        if rule_modified:
+            updated_rules.append(rule)
+
+    FormLogic.objects.bulk_update(updated_rules, fields=["actions"])
+
+
+class Migration(migrations.Migration):
```
sergei-maertens, line 33: self-reminder: backporting this migration is going to be fun again

sergei-maertens (changes requested):
I'll pick up the feedback & rebase so it's more easily backported :grimacing:

PR #2342, opened 2022-11-13
Upgrade to Python 3.10

sergei-maertens:
Let's see how CI fares...


sergei-maertens:
I suspect the tests are failing because of https://github.com/python/cpython/commit/9fc720e5e4f772598013ea48a3f0d22b2b6b04fa - where we should use `mimetypes.types_map` rather than `from mimetypes import types_map`

PR #2349, opened 2022-11-14
Styling intermediate screen after OIDC login

### LINKED ISSUE #2258: Style the intermediate screen for staff OIDC login
Related to #1862

#2216 adds the intermediate screen where an employee of the organization can enter a BSN or KVK number in case they're filling out a form on behalf of that other party. This is currently "unstyled".

Styling requirements:

- [x] Use the public UI styling (as opposed to the admin styling) by leveraging the SDK stylesheets and markup
- [x] Use either tabs or radio selection to toggle between "continue as staff", "continue as citizen" or "continue as company"
- [x] Only the relevant input field should be shown depending on the tab/radio option choosen

sergei-maertens:
Fixes #2258

image

**Changes**

* Show relevant fields depending on context
* Display auth mode in inline radio (stacked on mobile)


src/openforms/templates/includes/forms/field_wrapper.html
```diff
@@ -0,0 +1,15 @@
+{# Takes a 'field' context argument, which should be an instance of a BoundField #}
+<div class="openforms-form-control">
+    <div class="openforms-form-control openforms-form-control--{{ type|default:'textfield' }}">
+        <label for="{{ field.auto_id }}" class="openforms-label">
+            {{ field.label }}
+        </label>
+        <div>
+            {{ field }}
+        </div>
+        {% if field.help_text %}<div class="openforms-help-text">{{ field.help_text }}</div>{% endif %}
```
SilviaAmAm, line 10: I wonder if we should use something like `aria-describedby` so that a screen reader knows to which input the help-text corresponds?

src/openforms/authentication/forms.py
```diff
@@ -3,20 +3,39 @@
...
+    mode = forms.ChoiceField(
+        label=_("Continue as..."),
+        choices=ModeChoices.choices,
+        required=False,
+        widget=OpenFormsRadioSelect(inline=True),
+    )
     bsn = forms.CharField(
-        label=_("BSN number of client"),
+        label=_("BSN of client"),
+        help_text=_("Social security number of the customer"),
```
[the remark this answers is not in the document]
sergei-maertens: I'm normalizing it to "customer" - citizen/company are different entities but they're both customers. A citizen can represent a company, but ultimately it's the company itself that's the customer.

src/openforms/templates/includes/forms/field_wrapper.html
sergei-maertens: I looked into it but it's quite messy since we don't have pure template based form rendering yet. This would require us to set the `aria-describedby` attribute in python on the form field widget, but it also needs to know if the label/help text are going to be rendered or not since multiple values apply (and I think both label and help text describe the input).

PR #2317, opened 2022-11-07
:globe_with_meridians: [#2313] Add translations field to relevant (internal) API endpoints

stevenbal:
Closes #2313 

* Implements translations fields for (internal) API endpoints

Probably shouldn't be merged until the frontend for this is ready (https://github.com/open-formulieren/open-forms/pull/2353)

Issue #2315, opened 2022-11-07
Implement saving formio translations


sergei-maertens:
Epic #2170

Formio by default applies translations to a number of properties:

* `component.label`
* `component.description`
* `component.placeholder`
* `component.tooltip`
* `option.label` (for select, radio etc.)

Dynamic translations are a bit sketchy with Formio, so we need to build our own solution to
be able to inject these. Furthermore, we will ensure translations remain re-usable and
store them with the `FormDefinition` model.

- [ ] Add a JSONField `FormDefinition.translations`. The intended structure inside is:
  ```json
  {
    "nl": {
      "<componentPropertyValue>": "<Dutch translation>"
    }
  }
  ```
- [ ] Hook up loading and saving this translations data in the form designer (see #2314)

stevenbal:
@sergei-maertens while working on this, I noticed that the solution I was working on does not really seem to work too well in the case of reused translations. For instance, say I have a field `Field1` with the following translations: `{"en": {"foo": "bar"}}`. If I then create a second field `Field2` that uses the same literal `foo` somewhere and navigate to the `Translations` tab, there are no translations shown there yet (not until the configuration is saved and reloaded).

Is this a big issue?

sergei-maertens:
> @sergei-maertens while working on this, I noticed that the solution I was working on does not really seem to work too well in the case of reused translations. For instance, say I have a field `Field1` with the following translations: `{"en": {"foo": "bar"}}`. If I then create a second field `Field2` that uses the same literal `foo` somewhere and navigate to the `Translations` tab, there are no translations shown there yet (not until the configuration is saved and reloaded).
> 
> Is this a big issue?

It's far from ideal, but I think for the time being we can accept this is the way it is. Once there's a "ready enough" implementation to review, we can revisit this aspect and check how we can fit a solution in the implementation ("now" or later).

stevenbal:
#2437 is ready for review, I split off three issues to address the remaining stuff:
- https://github.com/open-formulieren/open-forms/issues/2475
- https://github.com/open-formulieren/open-forms/issues/2477
- https://github.com/open-formulieren/open-forms/issues/2478

PR #2353, opened 2022-11-15
:lipstick: [#2314] Translation tabs in Form designer

### LINKED ISSUE #2314: Implement translations UI in form designer
Depends on #2313 to save the translations, epic #2170

We need to implement/add a basic UI to manage translations in the form designer. In the future it's likely we'll get an export/import functinality as well for outsourced translations.

## Form tab

Wireframe
image

- [x] Reorganize tab content in sections `Details` and `Configuration` - we can use the blue or grey dividers used in other tabs
  
  * "Details" contains the translatable fields (name, explanation template)
  * Configuration contains all the other fields

- [x] Add tabbed translation interface for the translatable fields. `settings.LANGUAGE_CODE` should be the first, and the remaining languages are additional tabs. Activating a tab shows the same form but with the content from that particular language.

## Steps and fields tab

Wireframe
image

- [x] Wrap the "Form definition" section fields in translation tabs
- [x] Fields that are not translatable should become read-only except for when the primary language tab is active.

## Confirmation tab

- [x] Add translation tabs wrapper to every section (section header goes _above_ the tabs)

## Literals tab

- [x] Add translation tabs wrapper

## Formio component UI

Loading and persisting these translations is tracked via #2315

This concerns the edit component modal (left pane with config, right pane with preview)

- [x] Add a tab "translations" to every component
- [x] Add a nested tab interface with a tab for every supported language
- [x] Within a language tab, allow editing (datagrid?) the key-value map.
- [x] Only take translatable fields into account: label, description, select/radio option labels, content HTML...

Formio does not support context-aware translations, it's simply a key-value map. So the left side (key) should display the values that are filled out in the other form designer tabs, while the right side will have the translated version for that key.

## General considerations

These apply to all the tabs/sections

stevenbal:
Depends on https://github.com/open-formulieren/open-forms/pull/2317

Issue #2314 

TODO:
- [ ] dynamic initialState translation.languages
- [x] programmatically determine default language?


src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -86,10 +87,36 @@ const initialFormState = {
...
+        nl: {},
+        en: {},
+      },
+    },
     confirmationEmailOption: 'global_email',
     explanationTemplate: '',
     autoLoginAuthenticationBackend: '',
+    // TODO dynamic
+    translations: {
+      en: {
```
stevenbal, line 101: Currently the languages and keys are hardcoded, I'm not sure if this is an issue that could be addressed later of or if it needs to be tackled in this PR

stevenbal:
Some screenshots of the new tabs:

Formdetails:
Screenshot 2022-11-28 at 16-38-33 formulier wijzigen Open Formulieren

Formsteps:
Screenshot 2022-11-28 at 16-38-43 formulier wijzigen Open Formulieren

Confirmation:
Screenshot 2022-11-28 at 16-42-13 formulier wijzigen Open Formulieren
Screenshot 2022-11-28 at 16-42-19 formulier wijzigen Open Formulieren

Literals:
Screenshot 2022-11-28 at 16-39-22 formulier wijzigen Open Formulieren

Tabs should also indicate if there are errors for a specific language
Screenshot 2022-11-28 at 16-40-28 formulier wijzigen Open Formulieren

stevenbal:
This PR doesn't implement the formIO component part of https://github.com/open-formulieren/open-forms/issues/2314, because that depends on https://github.com/open-formulieren/open-forms/issues/2315

sergei-maertens (changes requested):
There are some broad changes/patterns I'd like to see addressed that apply to most of the code:

There is a lot of imperative programming that would really benefit from React's declarative nature. Instead of passing down the props all the way down the tree and then explicitly mapping over languages to build some JSX to then render, I'd rather see some code structure along the lines of:

```jsx
<LanguageTabs>
    <ErrorList />
    <FormRow>...</FormRow>
</LanguageTabs>
```
with the idea that the `LanguageTabs` component repeats its children for every supported language, making the tree structure a lot clearer.

This would also benefit from tracking some values/props in Context rather than having to pass them down via components that don't care about this information, e.g. the `LanguageTabs` component above should then be able to:

```jsx
const {languages, defaultLanguage} = useContext(LanguagesContext);
return languages.map(React.Children.each(children, ...));
```

etc.

I think this effort is too much to address all in this PR and postpone the backend branch further, so maybe it should be addressed in a separate PR so we can go ahead with this first iteration.

sergei-maertens:
Merging this as is, there's some follow up work to be done which is tracked in #2418

PR #2354, opened 2022-11-15
[#1899] Prefill data not formatted

### LINKED ISSUE #1899: Sending forms with BSN-prefill in read-only fields for birthday and postcode fails with 'Call did not validate on the backend'

SilviaAmAm:
Fixes #1899


src/openforms/submissions/models/submission.py
```diff
@@ -297,6 +299,20 @@ def save_registration_status(self, status, result):
...
+    def total_configuration_wrapper(self):
+        if not self._total_configuration_wrapper:
+            state = self.load_execution_state()
+            form_steps = state.form_steps
+            if len(form_steps) == 0:
+                return
+
+            wrapper = copy.deepcopy(form_steps[0].form_definition.configuration_wrapper)
+            for form_step in form_steps[1:]:
+                wrapper += form_step.form_definition.configuration
```
sergei-maertens, line 312: I don't like the magic method use wit _different_ types, so I'd say to either use each step's configuration_wrapper and sum it, or use a custom method `wrapper.add_to_configuration(step_configuration)`.

I also don't see why the `copy.deepcopy` is needed? The `FormConfigurationWrapper` doesn't implement deepcopy afaik so I'm expecting shallow copies anyway.

src/openforms/submissions/models/submission.py
```diff
@@ -297,6 +299,20 @@ def save_registration_status(self, status, result):
...
 
+    @property
+    def total_configuration_wrapper(self):
+        if not self._total_configuration_wrapper:
+            state = self.load_execution_state()
+            form_steps = state.form_steps
+            if len(form_steps) == 0:
+                return
+
+            wrapper = copy.deepcopy(form_steps[0].form_definition.configuration_wrapper)
```
sergei-maertens, line 310: it'd probably make more sense to instantiate `FormioConfigurationWrapper` and treat it as a separate object used in this particular scenario, rather than have it cached on the `form_definition`:

```
wrapper = FormioConfigurationWrapper(form_steps[0].form_definition.configuration_wrapper)
```

src/openforms/formio/datastructures.py
```diff
@@ -42,6 +42,20 @@ def __contains__(self, key: str) -> bool:
...
+        self._configuration["components"] += new_components
+
+        if self._cached_component_map is not None:
+            self._cached_component_map = {
+                **self._cached_component_map,
+                **{
+                    component["key"]: component
+                    for component in iter_components(configuration, recursive=True)
+                },
+            }
```
sergei-maertens, line 56: if you do `def __add__(self, other: "FormioConfigurationWrapper"):`, you could leverage the properties of that instance:

```py
self.component_map.update(other.component_map)
```

otherwise, a simple for loop would be more readable:

```py
for component in iter_components(configuration, recursive=True):
    self._cached_component_map[component["key"]] = component
```

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Similar fix on stable branch applied in df8096c5fb3ac6d73684445abdd68e274d551e49 and 929955e2325dbaa41103e5ddfd0487e0f788fb99

PR #2356, opened 2022-11-16
add logging for empty prefill plugin values

### LINKED ISSUE #2348: Log successful and empty prefill responses

2.0.0

No logevent is emitted if an empty set of values is returned here: https://github.com/open-formulieren/open-forms/blob/68e01d477642aad6aab2d86f21212cf5f70328e5/src/openforms/prefill/__init__.py#L72

_No response_

All code paths should result in a log line.

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

pi-sigma:
- Fixes #2348 by raising a ValueError and emitting the appropriate logevent if plugin returns an empty set of prefill values

### DIFF
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -170,6 +170,16 @@ def prefill_retrieve_success(submission: "Submission", plugin, prefill_fields):
     )
 
 
+def prefill_retrieve_empty(submission: "Submission", plugin, prefill_fields):
+    _create_log(
+        submission,
+        "prefill_retrieve_empty",
+        extra_data={"prefill_fields": prefill_fields},
+        plugin=plugin,
+        tags=[TimelineLogTags.AVG],
+    )
+
+
 def prefill_retrieve_failure(submission: "Submission", plugin, error: Exception):
     _create_log(
         submission,
--- a/src/openforms/prefill/__init__.py
+++ b/src/openforms/prefill/__init__.py
@@ -72,6 +72,8 @@ def _fetch_prefill_values(
         else:
             if values:
                 logevent.prefill_retrieve_success(submission, plugin, fields)
+            else:
+                logevent.prefill_retrieve_empty(submission, plugin, fields)
             return plugin_id, values
 
     with parallel() as executor:
--- a/src/openforms/prefill/tests/test_prefill_hook.py
+++ b/src/openforms/prefill/tests/test_prefill_hook.py
@@ -2,8 +2,10 @@ import logging
 from copy import deepcopy
 from unittest.mock import patch
 
+from django.template.defaultfilters import escape_filter
 from django.test import TransactionTestCase
 from django.utils.crypto import get_random_string
+from django.utils.translation import gettext as _
 
 from openforms.config.models import GlobalConfiguration
 from openforms.formio.datastructures import FormioConfigurationWrapper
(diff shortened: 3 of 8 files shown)


pi-sigma:
The test is almost a copy of the previous one (`test_prefill_no_result_and_default_value_set`), except that it specifically targets logging. I'm not sure if it's better to tuck on the logging test to the previous one (and save the code duplication), or if it's better to create a separate test (as it is now).

pi-sigma:
I can't reproduce the failing tests locally. Everything runs fine on my end.

src/openforms/prefill/__init__.py
```diff
@@ -64,13 +64,19 @@ def invoke_plugin(item: Tuple[str, List[str]]) -> Tuple[str, Dict[str, Any]]:
 
         try:
             values = plugin.get_prefill_values(submission, fields)
+            if not values:
+                try:
+                    raise ValueError(f"Empty values for prefill plugin '{plugin_id}'")
+                except ValueError as e:
+                    logger.exception(f"exception in prefill plugin '{plugin_id}'")
+                    logevent.prefill_retrieve_failure(submission, plugin, e)
+                    return plugin_id, values
```
sergei-maertens, line 73: this doesn't seem right - the fact that there's an empty values being returned doesn't mean/imply that there were errors/exceptions.

Instead below on line 72 I'd expect an `else` which logs a different message: `logevent.prefill_retrieve_empty` which explicitly mentions that "yes, the prefill API calls worked, but no data was returned" as that is the blind spot we currently have in logs

sergei-maertens (changes requested):

src/openforms/logging/logevent.py
```diff
@@ -170,6 +170,16 @@ def prefill_retrieve_success(submission: "Submission", plugin, prefill_fields):
     )
 
 
+def prefill_retrieve_empty(submission: "Submission", plugin, prefill_fields):
+    _create_log(
+        submission,
+        "prefill_retrieve_empty",
```
sergei-maertens, line 176: the template for this logevent is missing in the `src/openforms/logging/templates/logging/events/` directory

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -219,6 +220,32 @@ def get_prefill_values(submission, attributes):
...
+
+        apply_prefill(
+            configuration=config,
+            submission=submission,
+            register=register,
+        )
+
+        logs = TimelineLogProxy.objects.all()
+
+        self.assertNotEqual(logs.count(), 0)
```
sergei-maertens, line 244: either use `self.assertTrue(logs.exists())` or an explicit count: `self.assertEqual(logs.count(), 1)`

is there a reason to expect more than one log entry here?

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -219,6 +220,32 @@ def get_prefill_values(submission, attributes):
...
+        for log in logs:
+            with self.subTest(log):
+                self.assertEqual(log.event, "prefill_retrieve_empty")
```
sergei-maertens, line 247: the missing template would probably have been caught if you include an assertion on `log.get_message()`

src/openforms/submissions/tests/form_logic/test_submission_logic.py
```diff
@@ -773,6 +773,11 @@ def test_evaluate_logic_with_default_values(self):
         )
 
     def test_evaluate_logic_log_event_triggered(self):
+        # Logs need to be cleared, otherwise an AssertionError is raised in the
+        # logs.count test due to logevents created elsewhere for plugins with
+        # empty prefill values
+        TimelineLogProxy.objects.all().delete()
```
sergei-maertens, line 779: I suspect this is because of the threadpool that's used in the prefill functionality. Therefore, (prefill) tests should instead add a cleanup function in the `setUp`:

```py
def setUp(self):
    super().setUp()

    self.addCleanup(lambda: TimelineLogProxy.objects.all().delete())
```

sergei-maertens (changes requested):

src/openforms/submissions/tests/form_logic/test_submission_logic.py
pi-sigma: I've tried several ways to add the cleanup method to the testcases that deal with prefill plugins so as to treat the source of the failing tests rather than the symptoms, but to no avail. The only thing that works for me is clearing the logs in the functions that fail.

sergei-maertens (approved):

PR #2361, opened 2022-11-16
Add ServiceFetchConfiguration

### LINKED ISSUE #2332: Define data model to configure data-fetching actions (JSON services)
Epic #2169

This initially focuses on JSON-based services. SOAP/StUF will be handled in a separate model.

- [x] Define model `variables.ServiceFetchConfiguration` with the fields:
      * `service` FK to `zgw_consumers.Service` - this will be used for authentication credentials etc. 
      * `path` - URL relative to `Service.api_root` -> endpoint to hit
      * `method` - `GET|POST` (POST because Haal Centraal goes that way, this should not be used to mutate data!)
      * `headers` JSONField - merged with default service headers. If duplicates exist, this value overrides duplicates.
      * `query_params` - querystring parameters (not used with POST)
      * `body`: JSONField, used with POST only and only supports JSON (for now)
      * `data_mapping_type`: `none|JsonLogic|jq`
      * `mapping_expression` JSONField - actual object/JSON/primitive for mapping type jsonLogic, string to be validated with python-jq if mapping type is `jq`
- [x] Add python-jq and implement mapping expression validation
- [x] Expose the model `variables.ServiceFetchConfiguration` in the admin (for developers/superusers primarily)

CharString:
Closes #2332


src/openforms/variables/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class Variables(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "openforms.variables"
+    verbose_name = "Variables"
```
sergei-maertens, line 7: please mark the verbose_name translatable

src/openforms/variables/admin.py
```diff
@@ -0,0 +1,8 @@
+from django.contrib import admin
+
+from .models import ServiceFetchConfiguration
+
+
+@admin.register(ServiceFetchConfiguration)
+class ServiceFetchConfigurationAdmin(admin.ModelAdmin):
+    pass
```
sergei-maertens, line 8: always give the admin some love in terms of `list_display`, `raw_id_fields` etc.

src/openforms/variables/models.py
```diff
@@ -0,0 +1,86 @@
...
+        max_length=250,
+        blank=True,  # allow "" for the root, but not None
+        default="",
+        help_text=_("path relative to the Service API root"),
+    )
+    method = models.CharField(
+        _("HTTP method"),
+        max_length=4,
+        choices=ServiceFetchMethods.choices,
+        default="GET",
```
sergei-maertens, line 27:
```suggestion
        default=ServiceFetchMethods.get,
```

src/openforms/variables/models.py
```diff
@@ -0,0 +1,86 @@
...
+        help_text=_("POST is allowed, but should not be used to mutate data"),
+    )
+    headers = models.JSONField(
+        _("HTTP request headers"),
+        blank=True,
+        null=True,
+        help_text=_(
+            "Additions and overrides for the HTTP request headers as defined in the Service."
+        ),
+    )
```
sergei-maertens, line 37: probably good to add a validator here so that we don't get illegal header names/values. E.g. `{"foo bar": 30.0}` does not make any sense

src/openforms/variables/models.py
```diff
@@ -0,0 +1,86 @@
...
+        choices=DataMappingTypes.choices,
+    )
+    mapping_expression = models.JSONField(
+        _("Mapping expression"),
+        blank=True,
+        null=True,
+        help_text=_("For jq, pass a string containing the filter expression"),
+    )
+
+    def clean_fields(self, exclude=None):
```
sergei-maertens, line 65: wrong api to override - use `def clean(...)` for fields that depend on each other.

src/openforms/variables/models.py
```diff
@@ -0,0 +1,86 @@
...
+        elif self.data_mapping_type == DataMappingTypes.jq:
+            try:
+                jq.compile(self.mapping_expression)
+            except ValueError as e:
+                errors["mapping_expression"] = str(e)
+        elif self.data_mapping_type == DataMappingTypes.json_logic:
+            try:
+                jsonLogic(self.mapping_expression)
+            except ValueError as e:
+                errors["mapping_expression"] = str(e)
```
sergei-maertens, line 83: I think we already have a re-usable validator for this

src/openforms/variables/tests/test_models.py
```diff
@@ -0,0 +1,72 @@
+from django.core.exceptions import ValidationError
+from django.test import SimpleTestCase
+
+from ..models import DataMappingTypes, ServiceFetchConfiguration
+
+
+class ServiceFetchConfigurationTests(SimpleTestCase):
+    def test_body_on_get_request_should_be_empty(self):
+        null = ServiceFetchConfiguration(
+            method="GET",
```
sergei-maertens, line 10: please use the constants from django choices instead of using "magic strings"

sergei-maertens (changes requested):

CharString:
Tried to avoid adding a new dependency, but JSONField  userinput can be anything.

src/openforms/variables/models.py
```diff
@@ -0,0 +1,101 @@
...
+
+from .constants import DataMappingTypes, ServiceFetchMethods
+from .validators import HeaderValidator
+
+
+class ServiceFetchConfiguration(models.Model):
+    service = models.ForeignKey(
+        "zgw_consumers.Service",
+        on_delete=models.PROTECT,
+    )
```
sergei-maertens, line 18: please add `verbose_name=_("service")` so we can get that translated too in the admin

src/openforms/variables/admin.py
```diff
@@ -0,0 +1,17 @@
...
+    raw_id_fields = [
+        "service",
+    ]
+    list_display = [
+        "service",
+        "path",
+    ]
+    list_filter = [
+        "service",
+    ]
```
sergei-maertens, line 17: 1. let's add `method` to the `list_display` too
2. `search_fields` should also be defined, and it should search through the `path` and `service__api_root`

src/openforms/variables/models.py
```diff
@@ -0,0 +1,101 @@
...
+        null=True,
+        help_text=_(
+            'Request body for POST requests (only "application/json" is supported)'
+        ),
+    )
+    data_mapping_type = models.CharField(
+        _("mapping expression language"),
+        max_length=10,
+        blank=True,
+        null=True,
```
sergei-maertens, line 59: don't use `null=True` for text-based fields, as comparing/querying for empty values has some gotchas. `blank=True` is sufficient.

src/openforms/variables/tests/test_validators.py
```diff
@@ -0,0 +1,206 @@
...
+OBS_TEXT = "".join((chr(i) for i in range(0x80, 0x100)))
+SP = "\x20"
+HTAB = "\x09"
+CR = "\x0d"
+LF = "\x0a"
+NUL = "\x00"
+FIELD_VALUE_ALPHABET = "".join((VCHAR, OBS_TEXT, SP, HTAB))
+
+
+JSONValue: TypeAlias = (
```
sergei-maertens, line 21: JSONValue is defined in `openforms.typing` already (though I forgot `bool`, it seems)

sergei-maertens (changes requested):
I'm a bit surprised there doesn't seem to be an obvious HTTP header name/value parsing/validating utility in the stdlib or Django itself that makes all of this needed, but I couldn't find anything quickly myself.

I'm a bit concerned about the readability of these tests and I do think that the validation is a bit overkill, but it *looks* correct to me so let's go forward with it.

For the record, a more pragmatic approach would've probably been to try and chuck the headers into requests and see if it blows up or not:

```
>>> requests.get("https://example.com", headers={123: "foo"})
  File "<input>", line 1, in <module>
    requests.get("https://example.com", headers={123: "foo"})
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/requests/
api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/requests/
api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/bbt/code/open-forms/src/openforms/setup.py", line 132, in new_requ
est
    return self._original_request(*args, **kwargs)
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/requests/
sessions.py", line 528, in request
    prep = self.prepare_request(req)
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/requests/
sessions.py", line 462, in prepare_request
    headers=merge_setting(request.headers, self.headers, dict_class=CaseInsensit
iveDict),
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/requests/
sessions.py", line 70, in merge_setting
    merged_setting.update(to_key_val_list(request_setting))
  File "/usr/lib/python3.10/_collections_abc.py", line 1000, in update
    self[key] = value
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/requests/
structures.py", line 51, in __setitem__
    self._store[key.lower()] = (key, value)
AttributeError: 'int' object has no attribute 'lower'
```
The implementation in this PR is definitely _better_ though.

src/openforms/variables/admin.py
CharString: :slightly_smiling_face:  I had `method` in there, but removed it because in my happy place there should only be `GET`s in there. :see_no_evil:

sergei-maertens (approved):

Issue #2351, opened 2022-11-14
Changes to translations API for improved caches.


CharString:
### Thema / Theme

API

### Omschrijving / Description

`POST /api/v2/forms/{uuid_or_slug}/activate_default_language` and `PUT /api/v2/i18n/language` look idempotent and safe to me; doing them multiple times results in the same state on the client, and no server state is changed in either request. They can be made cacheable by turning them into GET requests. The `code` in the request body for the PUT, should be moved to either the path or the parameters, whichever fits best in the style of the rest of the API.

The `GET /api/v2/i18n/info` response contains the current language, and depends on the value of a the language cookie, so a `Vary` header is needed to make that clear.

### Added value / Toegevoegde waarde

Caching proxies can help scale. Using correct semantics in the API helps maximising cache potential.

sergei-maertens:
there are some DRF/OpenAPI peculiarities here and application context where the `@never_cache` is applied to all the API endpoints _by design_ so HTTP caching is not an argument to make this change.

Further discussion can follow after the current sprint

joeribekker:
Refinement: This might be useful for the tax office to tackle caching for the scale they need eventually. Not picking up now.

Issue #2255, opened 2022-10-26
(SDK) Ensure the correct (static) translations are loaded


sergei-maertens:
Epic #2170 - depends on #2253

The language info endpoint returns information about the current language (either inferred from browser preferences or explicitly set). Use this selected language through the SDK stack:

- [x] Ensure that the correct static literal translations are loaded and used (`react-intl` based)
- [x] Ensure that the correct formio language is used
    - Download relevant formio translations from the backend
    - Pass down the locale in the formio options so that validation errors etc. are shown in the correct language
- [x] Ensure that the correct `flatpickr` date-picker locale/translations are loaded and used

PR #2363, opened 2022-11-17
Convey language information via HTTP headers

sergei-maertens:
Related to #2255

After some discussion with Chris about the impact on the SDK and in light of #2351, relying on the HTTP headers will reduce the amount of API calls needed to make.

Changing the client language is no server state change, so that side-effect can be performed inside the GET form-detail call.

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -1488,8 +1488,19 @@ paths:
     get:
       operationId: forms_retrieve
       description: "Retrieve the details/configuration of a particular form. \n\n\
-        For forms that don't have translations enabled, the default language is activated,\
-        \ otherwise the browser preferences determine the active language."
+        A form is a collection of form steps, where each form step points to a formio.js\
+        \ form definition. Multiple definitions are combined in logical steps to build\
+        \ a multi-step/page form for end-users to fill out. Form definitions can be\
+        \ (and are) re-used among different forms.\n\n**Warning: the response data\
+        \ depends on user permissions**\n\nNon-staff users receive a subset of the\
+        \ documented fields which are usedfor internal form configuration. These fields\
+        \ are:\n\n- `internalName`\n- `registrationBackend`\n- `registrationBackendOptions`\n\
+        - `paymentBackend`\n- `paymentBackendOptions`\n- `product`\n- `category`\n\
+        - `isDeleted`\n- `submissionConfirmationTemplate`\n- `submissionsRemovalOptions`\n\
+        - `confirmationEmailTemplate`\n- `confirmationEmailOption`\n- `displayMainWebsiteLink`\n\
+        \nIf the form doesn't have translations enabled, its default language is forced\
+        \ by setting a language cookie and reflected in the Content-Language response\
+        \ header. Normal HTTP Content Negotiation rules apply."
       summary: Retrieve form details
       parameters:
       - in: header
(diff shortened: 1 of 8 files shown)


src/openforms/forms/tests/test_api_forms.py
```diff
@@ -1085,8 +1085,10 @@ def setUpTestData(cls):
         )
 
     def test_detail_shows_translated_values_based_on_request_header(self):
-
+        self.en_form.translation_enabled = True
+        self.en_form.save()
```
CharString, line 1089: Should this live in the `make_translated` decorator?

src/openforms/forms/api/viewsets.py
```diff
@@ -175,6 +174,11 @@ def configuration(self, request: Request, *args, **kwargs):
     retrieve=extend_schema(
         summary=_("Retrieve form details"),
         parameters=[UUID_OR_SLUG_PARAMETER],
+        description=_(
+            "Retrieve the details/configuration of a particular form. \n\nFor forms "
+            "that don't have translations enabled, the default language is activated, "
+            "otherwise the browser preferences determine the active language."
```
CharString, line 180: The API response is about a single Form. So the language doesn't have to be generalised. More like:

> Iff the form has no translations enabled, its default language is forced with a `Set-Cookie` header and reflected in the `Content-Language` header. Normal HTTP Content Negotiation rules are applied/followed/honoured.

I don't envy the person that has to translate "iff", but I like it when documentation tells me the precise mechanisms and what to expect. I expect a reader of an OpenAPI spec to know or willing to lookup details of the HTTP specs.

src/openapi.yaml
```diff
@@ -1396,35 +1482,14 @@ paths:
...
-        - `paymentBackend`
-        - `paymentBackendOptions`
-        - `product`
-        - `category`
-        - `isDeleted`
-        - `submissionConfirmationTemplate`
-        - `submissionsRemovalOptions`
-        - `confirmationEmailTemplate`
-        - `confirmationEmailOption`
-        - `displayMainWebsiteLink`
```
CharString, line 1427: Where did this go? I don't see this disappear in the code diff.

CharString (changes requested):
Love how this simplifies the flow and leans on established HTTP features.
One aside may be for mixed content endpoints, like the one for available languages, that contain content in multiple languages.

src/openforms/forms/tests/test_api_forms.py
sergei-maertens: hrm no, because `make_translated` also works on models/factories other than `Form` while this field is only scoped to the `Form` model.

I'll break out the data set up into this test method rather than rely on `setUpTestData` though, because this is not a good pattern.

src/openforms/forms/api/viewsets.py
sergei-maertens: Your expectation is very optimistic, from personal experience, but I agree ;)

src/openapi.yaml
sergei-maertens: oh I missed this, probably because of the explicit `retrieve` implementation and/or description that's now set rather than taken from the docstring. I'll sort it out!

sergei-maertens: it's because I added `description` to `@extend_schema` which overrules the default (viewset docstring), I've incorporated it now.

PR #2368, opened 2022-11-18
[#2367] Test migration going wrong

### LINKED ISSUE #2367: Automatic conversion of advanced front end logic goes wrong with selectboxes

1.1.x migration to 2.0.x

In this PR: https://github.com/open-formulieren/open-forms/pull/2049 we added a a method to convert "advanced frontend logic" to backend logic.

This conversion does not work properly when the trigger refers to the value of a `selectboxes` component.
For example, a form with configuration:
```json
{
    "components": [
        {
            "type": "selectboxes",
            "key": "test1",
            "values": [
                {
                    "label": "A",
                    "value": "a",
                },
                {
                    "label": "B",
                    "value": "b",
                },
            ],
        },
        {
            "type": "fieldset",
            "key": "test2",
            "components": [],
            "logic": [
                {
                    "name": "Rule 1",
                    "trigger": {
                        "type": "simple",
                        "simple": {
                            "show": True,
                            "when": "test1",
                            "eq": "a",
                        },
                    },
                    "actions": [
                        {
                            "name": "Action 1",
                            "type": "property",
                            "property": {
                                "label": "Hidden",
                                "value": "hidden",
                                "type": "boolean",
                            },
                            "state": False,
                        }
                    ],
                }
            ],
        },
    ]
}
```
Should end up having a trigger with json:
```json
{'==': [{'var': 'test1.a'}, True]}
```
While it has this:
```json
{'==': [{'var': 'test1'}, 'a']}
```

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #2367


sergei-maertens (changes requested):
Can you please make this PR to `stable/2.0.x` and add the migration conforming to that migration history? This will be some serious rework because you're using utilities that have been introduced after 2.0.0, so you'll have to replace those with what's available :(

PR #2371, opened 2022-11-21
[2219] update fix for CSS-unit issue with design tokens in email header logo

### LINKED ISSUE #2219: Setting logo size with design-token makes logo in e-mail display unscaled

1.1.7

When setting the size of the logo with design-tokens thus:
image

The logo in the forms and PDF resizes accordingly, however, it seems to throw off the resizing done in the e-mail: the image is displayed in its original size (a png of ~1000px in this case).

image

Without these design tokens in place the image is resized to a (OpenForms default) size:
image

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

pi-sigma:
Fixes #2219.

- the original fix no longer works for the latest version of MS Outlook's desktop client
- the update adds the logo height/width as HTML tags with the unit removed (or an empty string if the unit is anything other than "px")

### DIFF
--- a/src/openforms/emails/templates/emails/wrapper.html
+++ b/src/openforms/emails/templates/emails/wrapper.html
@@ -10,10 +10,8 @@
                     <td align="left" style="padding: 20px; color: {{ style.header.fg }}; background-color: {{ style.header.bg }};">
                         {% if logo_url %}
                             <a style="text-decoration: none;" href="{{ main_website_url }}">
-                                <img style="{% spaceless %}
-                                    {% if style.logo.width %}width: {{ style.logo.width }}; {% endif %}
-                                    {% if style.logo.height %}height: {{ style.logo.height }}; {% endif %}
-                                {% endspaceless %}" src="{{ logo_url }}" />
+                                <img {% if style.logo.width_attr %}width="{{ style.logo.width_attr }}" {% endif %} {% if style.logo.height_attr %}height="{{ style.logo.height_attr }}" {% endif %}
+				                style="{% spaceless %} {% if style.logo.width %}width: {{ style.logo.width }}; {% endif %} {% if style.logo.height %}height: {{ style.logo.height }}; {% endif %} {% endspaceless %}" src="{{ logo_url }}" />
                             </a>
                         {% endif %}
                     </td>
--- a/src/openforms/emails/tests/test_send_mail_html.py
+++ b/src/openforms/emails/tests/test_send_mail_html.py
@@ -1,9 +1,13 @@
 from django.conf import settings
 from django.core import mail
 from django.core.exceptions import SuspiciousOperation
+from django.template.loader import get_template
 from django.test import TestCase
 
+from pyquery import PyQuery
+
 from openforms.config.models import GlobalConfiguration
+from openforms.emails.context import _get_design_token_values
 from openforms.emails.utils import send_mail_html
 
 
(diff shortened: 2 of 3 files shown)


src/openforms/emails/templates/emails/wrapper.html
```diff
@@ -10,7 +10,9 @@
                     <td align="left" style="padding: 20px; color: {{ style.header.fg }}; background-color: {{ style.header.bg }};">
                         {% if logo_url %}
                             <a style="text-decoration: none;" href="{{ main_website_url }}">
-                                <img style="{% spaceless %}
+                                <img {% if style.logo.width %}width="{{ style.logo.width_attr }}" {% endif %}
+				     {% if style.logo.height %}height="{{ style.logo.height_attr }}" {% endif %}
+				    style="{% spaceless %}
```
sergei-maertens, line 15: if it's set to `auto`, this will still output the `width`/`height` attribute but with an empty value. Does that make sense? What happens in e-mail clients when that occurs?

pi-sigma: Only the value for the img tag attribute is empty; the CSS values are untouched. Hence it should display correctly (I've confirmed this for web clients; still waiting to hear from Joeri for MS Outlook's desktop client, which is the real culprit).

pi-sigma: My bad, I forgot to modify the `if`-test in the template to check the `attr` values of the `img` attributes. Now the `width` will simply not be present if it's set to `auto`

sergei-maertens: please add integration tests for the actual e-mail rendering in that case - the template filter unit tests pass but didn't catch this problem, this implies you're lacking some tests :)

src/openforms/emails/tests/test_send_mail_html.py
```diff
@@ -102,3 +104,72 @@ def test_oversize_content_raise_suspiciosu_operation(self):
...
+            "main_website_url": "https://logoresizingdoneright.com",
+            "style": {
+                "logo": {
+                    "height": "150px",
+                    "width": "auto",
+                    "height_attr": "150",
+                    "width_attr": "",
+                }
+            },
+            "logo_url": "https://haarlemmermeer.open-formulieren.nl/media/logo/Haarlemmermeer-logo-2018-RGB-transparant_vsdibKW.png",
```
sergei-maertens, line 167: please don't use/reference real URLs of real environments in publicly available code/repositories.

```suggestion
            "logo_url": "https://example.com/some-logo.png",
```

src/openforms/emails/tests/test_send_mail_html.py
```diff
@@ -102,3 +104,72 @@ def test_oversize_content_raise_suspiciosu_operation(self):
...
+        }
+        html_message = template.render(ctx)
+
+        self.assertIn("height: 150px;", html_message)
+        self.assertIn("width: auto;", html_message)
+        self.assertIn('height="150"', html_message)
+        self.assertNotIn('width="auto"', html_message)
+        self.assertNotIn('width=""', html_message)
```
sergei-maertens, line 175: Please extract the `img` node from the rendered HTML (using pyquery or lxml) and make assertions on that node, rather than just the entire HTML blob. Make tests/assertions as specific as possible.

sergei-maertens (changes requested):

src/openforms/emails/tests/test_send_mail_html.py
```diff
@@ -102,3 +106,75 @@ def test_oversize_content_raise_suspiciosu_operation(self):
...
+        html_message = template.render(ctx)
+
+        img = PyQuery(html_message)("img")
+        img_outer_html = img.outer_html()
+
+        self.assertIn("height: 150px;", img_outer_html)
+        self.assertIn("width: auto;", img_outer_html)
+        self.assertIn('height="150"', img_outer_html)
+        self.assertNotIn('width="auto"', img_outer_html)
+        self.assertNotIn('width=""', img_outer_html)
```
sergei-maertens, line 180: These assertions are still a bit too fuzzy - string comparisons/operations on HTML are usually a Bad Idea :tm: since you can have wrapping content that causes problems. What matters, is that the markup can be parsed correctly and results in the expected outcome, so we need to explicitly validate/test the attributes of the `img` element.

Using your PyQuery approach, we can do that this way:

```suggestion
        img = PyQuery(html_message)("img")

        self.assertEqual(img.attr("height"), "150")
        self.assertIsNone(img.attr("width"))

        img_styles = img.attr("style").replace(" ", "").split(";")
        self.assertIn("width:auto", img_styles)
        self.assertIn("height:150px", img_styles)
```
This checks three attributes explicitly: `width`, `height` and `style`. It abstracts away the potential single/double quote (mis)matches you can have, which the string assertions don't do.

The `style` attribute is then further checked - I noticed a lot of whitespace so cleaned up that string and then check for every expected style rule explicitly.

src/openforms/emails/templates/emails/wrapper.html
```diff
@@ -10,7 +10,9 @@
                     <td align="left" style="padding: 20px; color: {{ style.header.fg }}; background-color: {{ style.header.bg }};">
                         {% if logo_url %}
                             <a style="text-decoration: none;" href="{{ main_website_url }}">
-                                <img style="{% spaceless %}
+                                <img {% if style.logo.width_attr %}width="{{ style.logo.width_attr }}" {% endif %}
+				     {% if style.logo.height_attr %}height="{{ style.logo.height_attr }}" {% endif %}
+				    style="{% spaceless %}
                                     {% if style.logo.width %}width: {{ style.logo.width }}; {% endif %}
                                     {% if style.logo.height %}height: {{ style.logo.height }}; {% endif %}
                                 {% endspaceless %}" src="{{ logo_url }}" />
```
sergei-maertens, line 18: spaceless doesn't work here, as it only removes whitespace _between_ HTML tags. can you put all of this on a single line?

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #2357, opened 2022-11-16
Add a static variable for 'today'


hlmr-erik:
### Thema / Theme

Form designer

### Omschrijving / Description

We'd like to prefill a certain field with today's date. This is possible with the static variable 'now'; however this includes the (dynamic) time as well.
Thus
![image](https://user-images.githubusercontent.com/111565013/202160944-0ba1a603-fa5a-428d-8099-6c627c6ca6f9.png)
Leads to
![image](https://user-images.githubusercontent.com/111565013/202160967-27b8466c-e3af-46d3-a437-6e66542adf8a.png)
Which, as an unwanted side-effect triggers form-validation every second because it keeps updating. 

I tried to use the jsonlogic `{"today": []}` in this place, but it didn't work (returned a nondescript error in the formbuilder when trying to save) and played around to see if I could get the [operators on 'now' used in templates](https://open-forms.readthedocs.io/en/latest/manual/templates.html#formattering-van-variabelen) working here, but to no avail as well.

hlmr-erik:
@SilviaAmAm I did try this, also tried changing the field to a text-field to circumvent the restrictions of a `datetime `field. Just the JSON-logic doesn't support the same syntax as in templates, so I tried to fiddle around a bit like this: 
![image](https://user-images.githubusercontent.com/111565013/202388680-986f2a65-001b-4a7e-8c74-ae837c0772ef.png)
but that didn't work

SilviaAmAm:
Ahh I meant in the default value of the component, but I hadn't realised that there you can only use the date picker to set the default value but not a string with the variables:
![Screenshot from 2022-11-17 09-37-27](https://user-images.githubusercontent.com/19154114/202397601-7ed449cb-b0aa-411a-8757-ab23af512908.png)
And I see that adding it to the json of the component causes a 500 in the backend when starting the form :/

SilviaAmAm:
https://github.com/open-formulieren/open-forms/pull/2383 added a today variable

hlmr-erik:
> #2383 added a today variable

So this can be closed then :) I only recently found out as well that it is possible to achieve with the 'now' variable and Django templating

Issue #2324, opened 2022-11-08
Normalize (variable) data types


sergei-maertens:
See also #2251
See also #2305 (using local type information instead of requiring formio definitions would be very useful)

In short - we need to ensure that all (input) data has the correct type for processing in python and when calculations are done with json-logic, we need to serialize back to json-types and all data needs to be properly normalized.

We have essentially the following flow of information:

```

                                             +------------------+
+------------+      +----------------+       | Logic evaluation |____\ - output data (JSON)
| Input data |____\ | backend        |____\  | (JSON)           |    / - updated variable values (python/JSON)
| (JSON)     |    / | (python types) |    /  +------------------+
+------------+      +----------------+        ^
                                              |
                    +----------------+        |
                    | logic rules    |________+
                    | (JSON)         |
                    +----------------+

```

The 'problem' with JSON is that it only has a number of primitives that are used for richer Python types:

* string <-> date
* string <-> datetime
* string <-> duration
* string <-> time
* number <-> int
* number <-> float
* number <-> decimal
* ...

This is further complicated with the formio component types and the notion of single/multiple values (array vs. primitive).

## Using the python datatypes

Simply just using JSON types (complex & primitives) is not sufficient because we cannot do smart operations on them. We *must* support (non-exhaustive list) the following operations:

* formatting for display using Django's template language filters (e.g. `|date`, `|time`, `|floatformat` )
* calculations, e.g. `timestamp + timedelta`
* private API: introspecting/invoking methods - if we have strategy objects encapsulating data (dataclasses!) we want to use those to the full extent
* deep data extraction/assigment operations (e.g. using `glom` or `jq` to read/write certain data)
* submission data exports (tablib) also benefits from python types for serialization

## Identified boundaries

We can identify "our own code" as the system boundary. This implies:

* data enters this boundary (via the API endpoints) as JSON, once the boundary is being crossed we transform this into the relevant python types. For input variables, we can use the `FormVariable` type information (see `SubmissionValueVariable.to_python`)
* data enters this boundary (from the database adapter) as JSON (JSON logic expresions/rules). We cannot infer a more detailed type here - a string may be a date, datetime, date or simply a literal text.
* data exits the boundary once we pass it to third party libraries (json-logic)
* data exits the boundary once we serialize back for API responses
* data exits the boundary once we store variable values in the database (as JSON)

note: 'as JSON' implies the result of `json.loads(...)` here, so we have python dicts/lists/strings/ints/floats/NoneType.

[...]

PR #2383, opened 2022-11-23
[#2251] Fix date logic issue (attempt 2)

### LINKED ISSUE #2251: Logic with "now" variable not triggering with "bigger than" and "less than"

latest

Using the "now" (datetime) variable as a trigger the bigger/less than don't work as expected. 

This trigger does not work, today is 10/26/2022:
image

The trigger does work with equal and not equal values.

_No response_

_No response_

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

SilviaAmAm:
Fixes #2251 

This is mostly a bandaid fix while we wait for https://github.com/open-formulieren/open-forms/issues/2324

Changes in JSON logic for this PR: https://github.com/maykinmedia/json-logic-py/pull/10

This adds the variable `today` which was also requested here: #2357


src/openforms/js/components/admin/forms/Inputs.js
```diff
@@ -28,6 +29,48 @@ const NumberInput = props => <Input type="number" {...props} />;
...
+      ...extraProps,
+    });
+  }, []);
+
+  const placeHolder =
+    extraProps.placeholder ||
+    defineMessage({
+      description: 'datetime input widget placeholder',
+      defaultMessage: 'Select date/time...',
+    });
```
sergei-maertens, line 58: this means you can't simply pass an already `FormattedMessage` or string to `placeholder`, as it's passed down to `intl.formatMessage`. Maybe it'd be better to use `intl.formatMessage` directly in the default value and then just forward `placeholder`:

```jsx
const placeholder = extraProps.placeholder ?? intl.formatMesage({description: ..., defaultMessage: ...});
...
return (
    <input ref={...} placeholder={placeholder} />
)
```

src/openforms/variables/static_variables/static_variables.py
```diff
@@ -20,6 +20,15 @@ def get_initial_value(self, submission: Optional[Submission] = None):
         return timezone.now()
 
 
+@register_static_variable("today")
+class Today(BaseStaticVariable):
+    name = _("Today")
+    data_type = FormVariableDataTypes.date
+
+    def get_initial_value(self, submission: Optional[Submission] = None) -> str:
+        return timezone.now().date()
```
sergei-maertens, line 29: `timezone.now()` returns a datetime with timezone information in UTC, which means if you execute this at Nov 24th 00:30 AM, the reported date will actually be Nov 23rd -> please add a unit test for that :-)

The correct way would be:

```py
now_utc = timezone.now()
timezone.localtime(now_utc).date()
```

sergei-maertens (changes requested):

src/openforms/js/components/admin/forms/Inputs.js
```diff
@@ -28,6 +29,48 @@ const NumberInput = props => <Input type="number" {...props} />;
...
+      target: {name: name, value: formattedDatetime},
+    };
+    onChange(fakeEvent);
+  };
+
+  useEffect(() => {
+    flatpickr(datetimePickerRef.current, {
+      onChange: wrappedOnChange,
+      ...extraProps,
+    });
```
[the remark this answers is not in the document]
SilviaAmAm: Ahh yes, I had made a mental note of adding the destroy call, but clearly it should have been a physical note :D

src/openforms/variables/static_variables/static_variables.py
SilviaAmAm: :see_no_evil:

sergei-maertens (approved):

Issue #2397, opened 2022-11-25
Vorige-stap-label mentioned twice in Algemene configuratie


CharString:
### Product versie / Product version

0dcd9bf4a6f3c40cb1267cdfbaea9ebe83287577

### Omschrijf het probleem / Describe the bug

`Vorige stap-label[nl]` and `Vorige stap-label[en]` are both mentioned twice in `Admin>Configuratie>Algemene Configuratie`  section `Knop labels`

![image](https://user-images.githubusercontent.com/325643/203995600-eed6286b-9d6b-47ce-a3c2-f7cfedecb3ec.png)

### Verwacht gedrag / Expected behavior

I expect to see them each once, or expect them to have different names if these are different settings; if I give them different values, both are preserved, but the second ones seems to be served as the default form literals.

### Screen resolution

unknown

### Device

Desktop

### OS

Linux

### Browser

Firefox

PR #2404, opened 2022-11-28
[#2397] fix duplicate labels in admin

pi-sigma:
Fixes [#2397]   

 - the labels for previous step in the admin appeared twice
 - the names for the labels are changed because one of them leads back to the previous step within the form, the other leads back to the form after it's been completed
 - the message string with the translation is left empty and needs to be provided on the next release


src/openforms/config/models.py
```diff
@@ -153,7 +153,7 @@ class GlobalConfiguration(SingletonModel):
     )
 
     form_step_previous_text = models.CharField(
-        _("step previous text"),
+        _("previous step"),
```
sergei-maertens, line 156:
```suggestion
        _("previous step text"),
```

consistency with `form_change_text` and `form_confirm_text`

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2408, opened 2022-11-28
Clear data from fields if they are nested within a hidden component

### LINKED ISSUE #2340: Hiding fieldsets doesn't clear their data

2.0.0

Den Haag Taiga issue 334 

If a fieldset is marked as hidden by logic, the values of the components inside the fieldsets are not cleared.

Create a form with:
- A radio button with values `a` and `b`
- A fieldset with a text field inside. Mark the fieldset as "hidden"
- Logic rule: if the radio has value a: mark the fieldset as visible

Fill in the form:
- Select `a` for the radio
- Enter in the text field `This data should be cleared`
- Select `b` for the radio
- Submit the form

Check the submission => The textfield variable was not cleared

Example form on our maykin test environment: slug `test-data-clearing`

The data inside fieldsets should be cleared

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

sergei-maertens:
Bandaid fixes #2340

Note: this causes problems for logic rules that rely on the values of these hidden fields - we can tackle that when we get to #2409

### DIFF
--- a/src/openforms/formio/utils.py
+++ b/src/openforms/formio/utils.py
@@ -228,7 +228,7 @@ def conform_to_mask(value: str, mask: str) -> str:
     return "".join(result)
 
 
-def is_visible_in_frontend(component: JSONObject, data: DataMapping) -> bool:
+def is_visible_in_frontend(component: Component, data: DataMapping) -> bool:
     """Check if the component is visible because of frontend logic
 
     The rules in formio are expressed as:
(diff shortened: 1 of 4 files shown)


SilviaAmAm (approved):

src/openforms/formio/utils.py
```diff
@@ -241,6 +241,8 @@ def is_visible_in_frontend(component: JSONObject, data: DataMapping) -> bool:
             "eq": <compare value>
         }
 
+    .. warning:: this function currently does not take parent components into account
```
joeribekker, line 244: Isn't this the whole point of the ticket?

sergei-maertens: the [method on the `FormioConfigurationWrapper`](https://github.com/open-formulieren/open-forms/pull/2408/files#diff-3082701c8fd1dee09c8388c9d4de636a24ee32cacec78822a0aeecde0f9969feR71) _does_ take it into account and it uses this function. The test proves it works with parent components now :)

PR #2416, opened 2022-11-30
[#2402] Structured data for Object API registration

### LINKED ISSUE #2402: Productaanvraag JSON refactor
Den Haag Taiga issue 332

Currently, all the data from a submission is sent in one object without any structure. Den Haag would like this data to be more structured: the steps and the fieldsets should result in nesting.

**To do**

- [ ] Use a renderer to generate a nested JSON with nested data:
```
"data": {
  "step-slug": {
      "fieldset": {
           "property": "value"
       }
   }
}
```

SilviaAmAm:
Fixes #2402

### DIFF
--- a/docs/configuration/registration/objects.rst
+++ b/docs/configuration/registration/objects.rst
@@ -14,7 +14,7 @@ register form submissions.
 
 Below is an example of the contents in the ``record.data`` attribute in the
 Objects API. The top-level has meta-data about the form submission, and the
-``data`` element holds the submitted form values:
+``data`` element holds the submitted form values, nested within each step (using the step-slug as key):
 
 .. tabs::
 
@@ -24,9 +24,11 @@ Objects API. The top-level has meta-data about the form submission, and the
 
          {
            "data": {
-              "naam": "Jan Jansen",
-              "omschrijving": "Ik heb een vraag over mijn paspoort",
-              "telefoonnummer": "0612345678"
+             "uw-gegevens": {
+               "naam": "Jan Jansen",
+               "omschrijving": "Ik heb een vraag over mijn paspoort",
+               "telefoonnummer": "0612345678"
+           }
            },
            "type": "terugbelnotitie",
            "bsn": "111222333",
--- a/docs/configuration/registration/objects.rst
+++ b/docs/configuration/registration/objects.rst
@@ -28,7 +28,7 @@ Objects API. The top-level has meta-data about the form submission, and the
                "naam": "Jan Jansen",
                "omschrijving": "Ik heb een vraag over mijn paspoort",
                "telefoonnummer": "0612345678"
-           }
+             }
            },
            "type": "terugbelnotitie",
            "bsn": "111222333",
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -36,14 +36,16 @@ class ComponentNode(Node):
     step: SubmissionStep
     depth: int = 0
     is_layout = False
-    path: Path = None
+    path: Path = None  # Path in the data (#TODO rename to data_path?)
+    configuration_path: Path = None  # Path in the configuration tree
 
     @staticmethod
     def build_node(
         step: SubmissionStep,
         component: Component,
         renderer: "Renderer",
-        path: Path = None,
+        path: Path = None,  # Path in the data
+        configuration_path: Path = None,
         depth: int = 0,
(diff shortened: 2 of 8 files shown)


joeribekker (approved):
This is.... exactly how I wanted it :)

docs/configuration/registration/objects.rst
```diff
@@ -24,9 +24,11 @@ Objects API. The top-level has meta-data about the form submission, and the
...
          {
            "data": {
-              "naam": "Jan Jansen",
-              "omschrijving": "Ik heb een vraag over mijn paspoort",
-              "telefoonnummer": "0612345678"
+             "uw-gegevens": {
+               "naam": "Jan Jansen",
+               "omschrijving": "Ik heb een vraag over mijn paspoort",
+               "telefoonnummer": "0612345678"
+           }
```
sergei-maertens, line 31:
```suggestion
             }
```

src/openforms/formio/rendering/nodes.py
```diff
@@ -36,14 +36,16 @@ class ComponentNode(Node):
     step: SubmissionStep
     depth: int = 0
     is_layout = False
-    path: Path = None
+    path: Path = None  # Path in the data (#TODO rename to data_path?)
+    configuration_path: Path = None  # Path in the configuration tree
```
sergei-maertens, line 40: when/if we rename it, we should also take `openforms.formio.utils.flatten_by_path` into account, that's yet another "configuration" path (it's the json path that you can feed to glom given a formio configuration, e.g. `components.0.columns.3` and gives you the component at that location)

src/openforms/formio/rendering/utils.py
```diff
@@ -0,0 +1,53 @@
...
+from openforms.typing import JSONObject
+
+from .default import ColumnsNode, EditGridGroupNode, EditGridNode, FieldSetNode
+from .nodes import ComponentNode
+
+if TYPE_CHECKING:
+    from openforms.submissions.models import Submission
+
+
+def render_nested_json(submission: "Submission") -> JSONObject:
```
sergei-maertens, line 16: maybe name the python module `structured.py` instead of the too-generic `utils.py`?

I also find the `render_nested_json` a bit of a misnomer because it outputs python data structures suitable for json serialization, so maybe `openforms.formio.rendering.structured.render` covers the meaning better?

src/openforms/formio/rendering/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    """Render submission as JSON with nesting
+
+    The data is nested within each submission step (using the form definition slug as key).
+    The data is nested for fieldset components and for column components.
```
sergei-maertens, line 20: maybe add a remark that this is different from how formio itself treats these? This is more operating on the level of Formio's data container component

sergei-maertens (changes requested):
Looks great, couple minor remarks made in the code.

I have one bigger question though - this is quite a big change for systems that may already rely on the format we had before this patch, so this will need some release note warnings.

@joeribekker should we check a mechanism to make the new behaviour opt-in to not break existing systems and/or prevent people from upgrading? Realistically, how many parties are currently using the objects API in (pre) production?

src/openforms/formio/rendering/utils.py
SilviaAmAm: It's true, although I wanted to convey that this can be used to render 'json' (python data suitable for json serialization).

Maybe `openforms.formio.rendering.structured.render_json`?

sergei-maertens: yeah, it's a broader problem where "json" means different things in the python world, so that's a good compromise :P

joeribekker:
> @joeribekker should we check a mechanism to make the new behaviour opt-in to not break existing systems and/or prevent people from upgrading? Realistically, how many parties are currently using the objects API in (pre) production?

So, this whole Objects API route is really specific at the moment and I'm 99% sure that only Den Haag uses it. I think we should even mark the Objects API registration backend as beta as there will be more changes coming and I don't want to drag all the legacy formats we have, before we reach the final state, with us. Marking it as beta allows us to do this with more peace of mind.

As far as the Objects API "productaanvraag" objecttype goes, it will just work. And, applications using this should not make assumptions on the "data" attribute. So, on that side, I also don't see much problems for the ones that use it.

A release note warning is all I want :)

sergei-maertens (approved):

PR #2419, opened 2022-11-30
[#2250] fix broken analytics tools

### LINKED ISSUE #2250: Siteimprove not working

latest

internal: Rot 34

When adding siteimprove analytics: 

- CSP rules are added correctly

image

- Cookies are added, but the "AWSELBCORS" cookie contains the localdomain instead of the siteimprove ID + domain: [accountID].global.siteimproveanalytics.io

subsequently no requests are made to: siteimproveanalytics.com, this behavious does not change when adding the "correct" cookies

pi-sigma:
Fixes #2250
- all analytics tools (except Google Analytics) failed to render because the wrong property was tested in the templates
- fixed the templates
- added integration tests for the rendering of analytics snippets

### DIFF
--- a/src/openforms/analytics_tools/templates/analytics_tools/matomo.html
+++ b/src/openforms/analytics_tools/templates/analytics_tools/matomo.html
@@ -4,9 +4,9 @@
 {% load solo_tags %}
 {% get_solo 'analytics_tools.AnalyticsToolsConfiguration' as analytics_tools_config %}
 
-{% if analytics_tools_config.matomo_enabled %}
+{% if analytics_tools_config.is_matomo_enabled %}
 <!-- Matomo -->
-<script type="text/javascript" nonce="{{ request.csp_nonce }}">
+<script id="matomo-analytics" type="text/javascript" nonce="{{ request.csp_nonce }}">
   var _paq = window._paq = window._paq || [];
   _paq.push(['trackPageView']);
   _paq.push(['enableLinkTracking']);
--- a/src/openforms/analytics_tools/tests/test_analytics_rendering.py
+++ b/src/openforms/analytics_tools/tests/test_analytics_rendering.py
@@ -1,0 +1,123 @@
+from django.urls import reverse
+
+from cookie_consent.models import CookieGroup
+from django_webtest import WebTest
+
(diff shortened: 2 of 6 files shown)


pi-sigma:
I know it's not best practice to assert string inclusion against a large blob of text (as happens in my tests), but I see no way to zoom in on the scripts with Pyquery, Beautifulsoup or the like to ascertain that a particular snippet is rendered on the page. Should the tests be replaced with functional tests using Selenium, should we leave it as is, or is there some option I've overlooked?

src/openforms/analytics_tools/templates/analytics_tools/matomo.html
```diff
@@ -4,7 +4,7 @@
 {% load solo_tags %}
 {% get_solo 'analytics_tools.AnalyticsToolsConfiguration' as analytics_tools_config %}
 
-{% if analytics_tools_config.matomo_enabled %}
+{% if analytics_tools_config.enable_matomo_site_analytics %}
```
sergei-maertens, line 7:
```suggestion
{% if analytics_tools_config.is_matomo_enabled %}
```
this pattern repeats - we have an explicit check for:

* tool is configured correctly
* tool is actually turned on

now you're testing only for the latter

src/openforms/analytics_tools/tests/test_analytics_rendering.py
```diff
@@ -0,0 +1,122 @@
...
+        # Accept cookies
+        form_page = self.app.get(self.url)
+        accept_form = form_page.forms[0]
+        refreshed_form_page = accept_form.submit().follow()
+
+        soup = refreshed_form_page.html
+        script = soup.find("script")
+        script_text = script.get_text()
+
+        self.assertIn("googletagmanager", script_text)
```
sergei-maertens, line 49: to answer your question in the comment - let's give every analytics script an id (e.g. `<script id="analytics-matomo">` and test for that by finding the node with this ID and asserting that it's a script tag. That should be sufficient.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2420, opened 2022-11-30
[#1862] Create ROL for registration medewerker in ZGW API

SilviaAmAm:
Part of #1862


docs/configuration/registration/zgw.rst
```diff
@@ -105,6 +105,7 @@ Configuration
    * **Zaaktype**: Select the default Zaaktype to be used to create the Zaak. *For example:* ``https://example.com/catalogi/api/v1/zaaktypen/1/``
    * **Informatieobjecttype**: Fill in the URL of the Informatieobjecttype to be used to create the Document. *For example:* ``https://example.com/catalogi/api/v1/informatieobjecttypen/1/``
    * **Organisatie RSIN**: Fill the RSIN to be referred to in the created objects. *For example:* ``123456789``
+   * **Medewerker roltype**: Fill in the URL of the Roltype to be used when creating a role for an employee filling in a form for a citizen or company. For example: ``https://example.com/catalogi/api/v1/roltypen/2/``
```
sergei-maertens, line 108: Let's start using user-friendly fields/parameters for this - every time you get a new zaaktype version, the roltype URLs change and this is a maintenance burden. Instead, they should be able to configure the roltype `omschrijving` and our code should match this within the zaaktype.

src/openforms/registrations/contrib/zgw_apis/models.py
```diff
@@ -57,6 +57,14 @@ class ZgwConfig(SingletonModel):
...
         help_text=_("Default RSIN of organization, which creates the ZAAK"),
     )
+    medewerker_roltype = models.URLField(
+        _("medewerker roltype"),
+        blank=True,
+        help_text=_(
+            "URL of the ROLTYPE to use for employees filling in a form for a citizen/company."
+        ),
+        max_length=1000,
+    )
```
sergei-maertens, line 67: I honestly don't see a point in making these global defaults (the same applies to existing fields like `zaaktype` etc.), so let's only handle them in the configuration options serializer for the plugin.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -52,6 +52,12 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
             "Indication of the level to which extend the dossier of the ZAAK is meant to be public."
         ),
     )
+    medewerker_roltype = serializers.URLField(
+        required=False,
+        help_text=_(
+            "URL of the ROLTYPE to use for employees filling in a form for a citizen/company."
+        ),
+    )
```
sergei-maertens, line 60: As mentioned above, make this a regular `CharField` where the omschrijving can be looked up, and validation should then try to retrieve it within the configured zaaktype. If it can't be found -> validation error.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -197,6 +203,24 @@ def register_submission(
             partial(create_rol, zaak, rol_data, options), submission, "intermediate.rol"
         )
 
+        has_registrator = (
+            hasattr(submission, "_registrator") and submission._registrator
+        )
+        if has_registrator:
+            registrator_rol_data = {
+                "betrokkeneType": "medewerker",
+                "roltype": options["medewerker_roltype"],
```
sergei-maertens, line 212: this needs to be looked up based on `omschrijving` within the zaaktype

sergei-maertens (changes requested):

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -197,6 +210,27 @@ def register_submission(
             partial(create_rol, zaak, rol_data, options), submission, "intermediate.rol"
         )
 
+        has_registrator = (
+            hasattr(submission, "_registrator") and submission._registrator
+        )
+        if has_registrator:
+            roltype = retrieve_roltype(
+                options["medewerker_roltype"], **{"zaaktype": options["zaaktype"]}
+            )
```
sergei-maertens, line 219:
```suggestion
            roltype = retrieve_roltype(
                options["medewerker_roltype"], zaaktype=options["zaaktype"]
            )
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -197,6 +210,27 @@ def register_submission(
...
+            registrator_rol_data = {
+                "betrokkeneType": "medewerker",
+                "roltype": roltype["url"],
+                "roltoelichting": "medewerker balie",
```
sergei-maertens, line 223:
```suggestion
                "roltoelichting": _("Employee who registered the case on behalf of the customer."),
```

src/openforms/contrib/zgw/service.py
```diff
@@ -193,7 +193,7 @@ def create_rol(zaak: dict, initiator: dict, options: dict) -> Optional[dict]:
         "zaak": zaak["url"],
         # "betrokkene": initiator.get("betrokkene", ""),
         "betrokkeneType": initiator.get("betrokkeneType", "natuurlijk_persoon"),
-        "roltype": rol_typen["results"][0]["url"],
+        "roltype": initiator.get("roltype", rol_typen["results"][0]["url"]),
```
sergei-maertens, line 196: maybe only fetch `rol_typen` here if there's no `roltype` provided, now we're doing calls to APIs that only slows down things while the result isn't even used.

I'd also rename `initiator` to `betrokkene` here, since the same function is now used in different (non-initiator) contexts.

The logic to find the list of `roltypen` inside of this function should also make use of the new `retrieve_roltype` function - possibly you can refactor that to make the API calls and then provide a `matcher` callback to extract the relevant `roltype`. This keeps things DRY.

src/openforms/registrations/contrib/zgw_apis/tests/test_validators.py
```diff
@@ -0,0 +1,84 @@
...
+
+from ..plugin import ZaakOptionsSerializer
+from .factories import ZgwConfigFactory
+
+
+@requests_mock.Mocker(real_http=False)
+class OmschrijvingValidatorTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        ZgwConfigFactory.create(
```
sergei-maertens, line 16: missing `super()` call

src/openforms/registrations/contrib/zgw_apis/tests/test_validators.py
```diff
@@ -0,0 +1,84 @@
...
+import requests_mock
+from zds_client.oas import schema_fetcher
+from zgw_consumers.test import generate_oas_component
+from zgw_consumers.test.schema_mock import mock_service_oas_get
+
+from ..plugin import ZaakOptionsSerializer
+from .factories import ZgwConfigFactory
+
+
+@requests_mock.Mocker(real_http=False)
```
sergei-maertens, line 12:
```suggestion
@requests_mock.Mocker()
```
`real_http=False` is the default

src/openforms/registrations/contrib/zgw_apis/tests/test_validators.py
```diff
@@ -0,0 +1,84 @@
...
+        data = {
+            "zaaktype": "https://catalogus.nl/api/v1/zaaktypen/111",
+            "medewerker_roltype": "Some description",
+        }
+
+        serializer = ZaakOptionsSerializer(data=data)
+
+        is_valid = serializer.is_valid()
+
+        self.assertFalse(is_valid)
```
sergei-maertens, line 84: please make assertions on at least the error field - this serializer could possibly be invalid because of _other_ fields (if it changes in the future)

src/openforms/registrations/contrib/zgw_apis/validators.py
```diff
@@ -0,0 +1,23 @@
...
+from openforms.contrib.zgw.service import retrieve_roltype
+
+
+class RoltypeOmschrijvingValidator:
+    def __call__(self, data: dict):
+        if not ("medewerker_roltype" in data and "zaaktype" in data):
+            return
+
+        roltype = retrieve_roltype(
+            data["medewerker_roltype"], **{"zaaktype": data["zaaktype"]}
```
sergei-maertens, line 14:
```suggestion
            data["medewerker_roltype"], zaaktype=data["zaaktype"]
```

sergei-maertens (changes requested):

src/openforms/contrib/zgw/service.py
```diff
@@ -173,35 +173,57 @@ def relate_document(zaak_url: str, document_url: str) -> dict:
...
     rol = zrc_client.create("rol", data)
     return rol
 
 
+def match_omschrijving(roltypen: List, **kwargs) -> List:
+    matches = []
+    for roltype in roltypen:
+        if roltype.get("omschrijving") == kwargs["omschrijving"]:
+            matches.append(roltype)
+    return matches
```
sergei-maertens, line 212: can we please avoid using `**kwargs` if we can be explicit instead?

```suggestion
def match_omschrijving(roltypen: List, omschrijving: str) -> List:
    matches = []
    for roltype in roltypen:
        if roltype.get("omschrijving") == omschrijving:
            matches.append(roltype)
    return matches
```
is more readable _and_ easier to type hint

src/openforms/contrib/zgw/service.py
```diff
@@ -173,35 +173,57 @@ def relate_document(zaak_url: str, document_url: str) -> dict:
...
+
+
+def retrieve_roltypen(
+    matcher: Callable = None, query_params: Optional[dict] = None
+) -> List:
```
sergei-maertens, line 217:
```suggestion

def noop_matcher(roltypen: list) -> list:
    return roltypen

def retrieve_roltypen(
    matcher: Callable[[list], list] = noop_matcher, query_params: Optional[dict] = None
) -> List:
```

on python 3.10 you can now also use `list` instead of `List` (and `None | dict` instead of `Optional[dict]`)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2437, opened 2022-12-06
:sparkles: [#2315] FormIO translations in form designer admin

### LINKED ISSUE #2315: Implement saving formio translations

stevenbal:
Closes #2315 

Issues split from this PR:
- https://github.com/open-formulieren/open-forms/issues/2475
- https://github.com/open-formulieren/open-forms/issues/2477
- https://github.com/open-formulieren/open-forms/issues/2478


src/openforms/js/components/form/edit/tabs.js
```diff
@@ -321,28 +321,108 @@ const PREFILL = {
...
 };
 
+const TRANSLATIONS = {
+  key: 'tabs',
+  label: 'Translations',
+  type: 'tabs',
+  components: [
+    {
+      type: 'panel',
+      title: 'NL',
```
stevenbal, line 331: @sergei-maertens I was looking into making this dynamic (based on the languageInfo endpoint), but I'm not sure if it will work in this fashion. I don't think I can rely on async results (because I would have to fetch the languages from endpoint) when exporting the `TRANSLATIONS` tab.

I also tried turning it into an async function `getTranslationTabs` that generates the tabs, but then I would run into the same issue in other places I think, plus that would mean I have to call the endpoint more often then necessary.

Is there another way to do this?

stevenbal: Discussed on Slack, solved by using a `json_script` template filter to pass `settings.LANGUAGES`

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -413,6 +415,15 @@ function reducer(draft, action) {
       // TODO: This could break if a reusable definition is used multiple times in a form
       const step = getFormStep(formDefinition, draft.formSteps, true);
 
+      if (mutationType === 'changed' && schema['of-translations']) {
```
stevenbal, line 418: I'm not sure about this part, because currently this is only to keep track of componentTranslations so they can be shown in the form builder in case validation fails. The actual translations that are sent to the backend are retrieved here: https://github.com/open-formulieren/open-forms/pull/2437/files#diff-88ee85ef9cb073c1e9150e72a526cb11c736468981c6361fa14c2f7bca22c24fR42 (from the formio config `of-translations`)

stevenbal: This will be tackled in https://github.com/open-formulieren/open-forms/issues/2478

src/openforms/conf/base.py
```diff
@@ -287,6 +287,7 @@
                 "django.template.context_processors.request",
                 "django.contrib.auth.context_processors.auth",
                 "django.contrib.messages.context_processors.messages",
+                "openforms.forms.context_processors.language_info",
```
sergei-maertens, line 290: seeing you only access `settings.LANGUAGES`, you can use the built-in context processor: https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#i18n-context-processor

But probably just using `{% get_available_languages as languages %}` will be sufficient

src/openforms/forms/models/form_definition.py
```diff
@@ -72,6 +72,13 @@ class FormDefinition(models.Model):
         help_text=_("The total number of Formio components used in the configuration"),
     )
 
+    component_translations = models.JSONField(
+        verbose_name=_("Component translations"),
+        help_text=_("Translations for literals used in components"),
+        blank=True,
+        null=True,
```
sergei-maertens, line 79: maybe `default=dict` is a better approach to guarantee there's always a dict datastructure returned from there - then the user only needs to check if the language key is available instead of also checking for null

src/openforms/js/components/admin/form_design/FormSteps.js
```diff
@@ -2,6 +2,7 @@ import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import {FormattedMessage} from 'react-intl';
+import {FormStepContext} from './Context';
```
sergei-maertens, line 5: please group/order the imports with the same rules as python - this should go at the end of the imports with the other relative imports

src/openforms/js/components/admin/form_design/data/steps.js
```diff
@@ -2,6 +2,9 @@ import {FORM_DEFINITIONS_ENDPOINT} from 'components/admin/form_design/constants'
 import {DEFAULT_LANGUAGE} from 'components/admin/form_design/LanguageTabs';
 import {FormException} from 'utils/exception';
 import {post, put, ValidationErrors} from 'utils/fetch';
+import cloneDeep from 'lodash/cloneDeep';
```
sergei-maertens, line 5: this import belongs at the top in its own group as it's a third party package

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -413,6 +415,15 @@ function reducer(draft, action) {
...
+        // Story the componentTranslations on the formStep, this is relevant when form
```
sergei-maertens, line 419:
```suggestion
        // Store the componentTranslations on the formStep, this is relevant when form
```

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -156,6 +158,65 @@ const getPathToComponent = (configuration, key) => {
   return '';
 };
 
+// Determine the translations for a configuration, and transform them into the proper shape
+const extractTranslationsFromConfiguration = configuration => {
+  let componentTranslations = {};
+  if (configuration.components) {
```
sergei-maertens, line 164: use the `FormioUtils.eachComponent` - as `components` is not the only key containing nested components, but also `columns`, `rows`...

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -156,6 +158,65 @@ const getPathToComponent = (configuration, key) => {
...
+    componentTranslations = mergeComponentTranslations(
+      componentTranslations,
+      rewriteComponentTranslations(configuration['of-translations'])
+    );
+  }
+  return componentTranslations;
+};
+
+const removeTranslationsFromConfiguration = configuration => {
+  if (configuration.components) {
```
sergei-maertens, line 183: use `FormioUtils.eachComponent`

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -156,6 +158,65 @@ const getPathToComponent = (configuration, key) => {
...
+  let mutatedTranslations = {};
+  for (const [languageCode, translations] of Object.entries(allTranslations)) {
+    for (const entry of translations) {
+      set(mutatedTranslations, `${languageCode}.${entry.literal}`, entry.translation);
+    }
+  }
+  return mutatedTranslations;
+};
+
+const mergeComponentTranslations = (currentTranslations, newTranslations) => {
```
sergei-maertens, line 203: perhaps https://docs-lodash.com/v4/merge/ can simplify this a bit?

src/openforms/js/components/form/edit/tabs.js
```diff
@@ -321,28 +322,82 @@ const PREFILL = {
...
+        components: [
+          {
+            label: 'Literal',
+            key: 'literal',
+            input: true,
+            unique: true,
+            type: 'textfield',
+          },
+          {
+            label: 'Vertaling',
```
sergei-maertens, line 348:
```suggestion
            label: 'Translation',
```
we'll manage that through our own static translations for formio :)

src/openforms/js/components/formio_builder/builder.js
```diff
@@ -1,14 +1,70 @@
...
+
+    values.push(component[fieldName]);
+  }
+  return values;
+};
+
+const injectTranslationsIntoConfiguration = (configuration, componentTranslations) => {
+  let translationMapping = {};
+
+  if (configuration.components) {
```
sergei-maertens, line 42: use `FormioUtils.eachComponent` instead :)

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/FormSteps.js
stevenbal: Ah I thought that `prettier` would order them so I wasn't paying attention it, will do so in the future :stuck_out_tongue:

sergei-maertens: looks like there's a plugin for that: https://dev.to/diballesteros/how-to-quickly-sort-imports-with-prettier-3po7! I created #2487 for this

stevenbal: Oh nice, I'll pick that one up

sergei-maertens (approved):

sergei-maertens:
Let's go! This is something that can be demo'd :)

PR #2438, opened 2022-12-06
[#2428] add dynamic logo label

### LINKED ISSUE #2428: Add dynamic label-name to Logo for accessibility
### Thema / Theme

Frontend

A dynamic title on the Logo, that can be read by screenreaders. is needed. On the aria-label and on the title. Right now the image just reads statically as "terug naar website".
**1.** Goal is to change it to: "terug naar website [**organisatienaam**]"
Where "organisatienaam" should be the dynamic name of the municipality and could be a new config field.
So this issue is a request to add that dynamic name + it should be part of translations (_Back to website organisationname / Terug naar website organisationname_), possibly by use of Blocktrans.

**2.** Additionally the current code is not up to accessibility standards, as only IMG tags and SVG's are allowed to have an "alt".
So the **alt** "_Logo gemeente_" inside the anchor will not be read by screenreaders and should be removed.

**3.** This new config field should be added to the organisation name to both the **aria-label** and the **title** attribute. Both with a translatable "Back to website **organisationname**".

A dynamic title will give an accessible meaning to the organization logo + correct syntax also makes the image-role more accessible for screen-readers.

header-alt

pi-sigma:
Fixes #2428.

- Added a field `organization_name` to `GlobalConfiguration` which is picked up in the header template as part of the logo's `title` and `aria-label` value
- replaced `translate` with `blocktranslate` in the header template to handle translation of strings with variables
- added test for `title` and `aria-label`


src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -4,7 +4,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Open Forms\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-11-29 15:40+0100\n"
+"POT-Creation-Date: 2022-12-06 14:09+0100\n"
```
sergei-maertens, line 7: Please revert all changes to the `.po` file(s) - these do not belong in feature pull requests.

Translations are managed shortly before a release and this is done exclusively by Joeri or me.

src/openforms/config/models.py
```diff
@@ -229,6 +229,14 @@ class GlobalConfiguration(SingletonModel):
             "URL to the main website. Used for the 'back to municipality website' link."
         ),
     )
+    organization_name = models.CharField(
+        _("organization name"),
+        max_length=50,
```
sergei-maertens, line 234: let's make this 100 to be future proof - longest municipality name at the moment seems to be around 30 chars, but if two of those fuse we're at 60+ characters.

src/openforms/ui/tests/test_header.py
```diff
@@ -0,0 +1,28 @@
...
+from django_webtest import WebTest
+
+from openforms.config.models import GlobalConfiguration
+from openforms.forms.tests.factories import FormFactory
+
+
+class HeaderTest(WebTest):
+    @patch("openforms.config.models.GlobalConfiguration.get_solo")
+    def test_logo_label(self, get_solo_mock):
+        get_solo_mock.return_value = GlobalConfiguration(organization_name="Orgy")
```
sergei-maertens, line 14: you might want to use an organization name that's a bit more innocent :grimacing:

src/openforms/ui/tests/test_header.py
```diff
@@ -0,0 +1,28 @@
...
+
+        self.assertEqual(200, response.status_code)
+
+        header_node = response.html.find("header")
+        link_node = header_node.find("a")
+        aria_label = link_node.get("aria-label")
+        title = link_node.get("title")
+
+        self.assertEqual(aria_label, "Back to website of Orgy")
+        self.assertEqual(title, "Back to website of Orgy")
```
sergei-maertens, line 28: you _must_ always use the translatable strings in tests with the correct interpolation format, as these tests will break when we actually enter the translations.

sergei-maertens (changes requested):

src/openforms/config/models.py
```diff
@@ -229,6 +229,15 @@ class GlobalConfiguration(SingletonModel):
             "URL to the main website. Used for the 'back to municipality website' link."
         ),
     )
+    organization_name = models.CharField(
+        _("organization name"),
+        max_length=60,
```
sergei-maertens, line 234:
```suggestion
        max_length=100,
```

src/openforms/ui/tests/test_header.py
```diff
@@ -0,0 +1,30 @@
...
+from django_webtest import WebTest
+
+from openforms.config.models import GlobalConfiguration
+from openforms.forms.tests.factories import FormFactory
+
+
+class HeaderTest(WebTest):
+    @patch("openforms.config.models.GlobalConfiguration.get_solo")
+    def test_logo_label(self, get_solo_mock):
+        get_solo_mock.return_value = GlobalConfiguration(organization_name="Orgee")
```
sergei-maertens, line 15:
```suggestion
        get_solo_mock.return_value = GlobalConfiguration(organization_name="My organization")
```

src/openforms/ui/tests/test_header.py
```diff
@@ -0,0 +1,30 @@
...
+        organization_name = get_solo_mock.return_value.organization_name
```
sergei-maertens, line 16:
```suggestion
```

don't add indirectness in tests, make your assertions as plain as possible

src/openforms/ui/tests/test_header.py
```diff
@@ -0,0 +1,30 @@
...
+
+        self.assertEqual(200, response.status_code)
+
+        header_node = response.html.find("header")
+        link_node = header_node.find("a")
+        aria_label = link_node.get("aria-label")
+        title = link_node.get("title")
+
+        self.assertEqual(aria_label, _("Back to website of %s" % organization_name))
+        self.assertEqual(title, _("Back to website of %s" % organization_name))
```
sergei-maertens, line 30:
```suggestion
        expected_label = _("Back to website of %(name)s") % {"name": "My organization"}
        self.assertEqual(aria_label, expected_label)
        self.assertEqual(title, expected_label)
```

The template has `Back to website of {{ name }}`, so you must use named substitutions in the `gettext` call

The interpolation needs to be done on the `gettext` result, not on the string you feed to `gettext`. `_("foo %s" % "bar")` is _not the same_ as `_("foo %s") % "bar"` - the former will register `"foo bar"` in the message catalog (wrong), the latter will register `"foo %s"` (correct).

sergei-maertens (approved):

PR #2439, opened 2022-12-06
Render PDF reports and confirmation emails in submission language

### LINKED ISSUE #2390: Ensure that submission PDF reports and confirmation e-mails are rendered in the submission language
Epic #2170 - blocked by #2315

When a user made a submission in a particular language, any and all confirmations they receive must be presented in that language.

Notably, this applies to the submission PDF that is generated and the confirmation e-mail that's generated and sent.

**Checklist**

- [x] Form step titles must be correctly localized
- [x] Formio component labels, content... must be correctly localized
- [x] Dropdown/radio/checkbox labels must be correctly localized

CharString:
Partial solution to  #2390

### DIFF
--- a/src/openforms/submissions/models/submission_report.py
+++ b/src/openforms/submissions/models/submission_report.py
@@ -67,7 +67,6 @@ class SubmissionReport(models.Model):
         """
         with override(self.submission.language_code):
             form = self.submission.form
-            # Switch to submission lang TODO
             html_report, pdf_report = render_to_pdf(
                 "report/submission_report.html",
                 context={"report": Report(self.submission)},
--- a/src/openforms/submissions/models/submission_report.py
+++ b/src/openforms/submissions/models/submission_report.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 from django.core.files.base import ContentFile
 from django.db import models
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_lazy as _, override
 
 from celery.result import AsyncResult
 from privates.fields import PrivateMediaFileField
@@ -65,15 +65,17 @@ class SubmissionReport(models.Model):
         :return: string with the HTML used for the PDF generation, so that contents
           can be tested.
         """
-        form = self.submission.form
-        html_report, pdf_report = render_to_pdf(
-            "report/submission_report.html",
-            context={"report": Report(self.submission)},
-        )
-        self.content = ContentFile(
-            content=pdf_report,
-            name=f"{form.name}.pdf",  # Takes care of replacing spaces with underscores
-        )
+        with override(self.submission.language_code):
+            form = self.submission.form
+            # Switch to submission lang TODO
+            html_report, pdf_report = render_to_pdf(
+                "report/submission_report.html",
+                context={"report": Report(self.submission)},
+            )
+            self.content = ContentFile(
+                content=pdf_report,
+                name=f"{form.name}.pdf",  # Takes care of replacing spaces with underscores
+            )
         self.save()
         return html_report
 
--- a/src/openforms/submissions/tests/test_submission_report.py
+++ b/src/openforms/submissions/tests/test_submission_report.py
@@ -10,8 +10,6 @@ from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APITestCase
 
-from openforms.forms.tests.factories import FormStepFactory
-
 from ..models import SubmissionReport
 from ..tasks import generate_submission_report
 from ..tokens import submission_report_token_generator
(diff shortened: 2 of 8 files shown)


src/openforms/submissions/models/submission_report.py
```diff
@@ -65,15 +65,17 @@ def generate_submission_report_pdf(self) -> str:
...
-            "report/submission_report.html",
-            context={"report": Report(self.submission)},
-        )
-        self.content = ContentFile(
-            content=pdf_report,
-            name=f"{form.name}.pdf",  # Takes care of replacing spaces with underscores
-        )
+        with override(self.submission.language_code):
+            form = self.submission.form
+            # Switch to submission lang TODO
```
sergei-maertens, line 70: what's this TODO about?

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -97,6 +99,35 @@ def test_report_generation(self, mock_request):
...
+                "components": [
+                    {
+                        "key": "foo",
+                        "type": "textfield",
+                        "label": "Foo",
+                        "showInEmail": True,
+                    }
+                ],
+            },
+        )
```
sergei-maertens, line 122: I'd consider using the chained `form__generate_minimal_setup` in the `SubmissionFactory` to stuff everything into a single factory call. CTRL+F on the codebase to find some examples.

Additionally, if a submission `language_code` is provided different from `settings.LANGUAGE_CODE`, this should also enforce `form__translations_enabled=True` to generate consistent data. That should be possible with something like:

```py
class SubmissionFactory(...):
    form = factory.SubFactory(
        FormFactory,
        translation_enabled=factory.LazyAttribute(lambda o: o.factory_parent.language_code != settings.LANGUAGE_CODE)
    )
```

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -97,6 +99,35 @@ def test_report_generation(self, mock_request):
...
 
+    def test_report_is_generated_in_same_language_as_submission(self):
+        submission = SubmissionFactory.create(
+            language_code="en",
+            completed=True,
+            form__name_en="Translated form name",
+        )
+        FormStepFactory.create(
+            form=submission.form,
+            form_definition__name="Walsje",
```
sergei-maertens, line 110:
```suggestion
            form_definition__name_nl="Walsje",
```
makes it just that bit more explicit

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -559,6 +560,71 @@ def test_send_confirmation_email_when_appointment_is_changed(self):
...
+                "components": [
+                    {
+                        "key": "foo",
+                        "type": "textfield",
+                        "label": "Foo",
+                        "showInEmail": True,
+                    }
+                ],
+            },
+        )
```
sergei-maertens, line 583: I don't particularly see the benefit of using `make_translated` here, in tests where we are explicitly testing that the correct languages are used. It's much easier to understand what data is being set up by using the explicit `_en`-suffixed-fields, or maybe even do the test data setup within a `translation.override("en")` context manager.

See also the earlier remarks - we definitely favour using the `FormFactory(generate_minimal_setup=True)` trait to ensure consistent data setup over time - this may now be a correct database state, but it may not be in the future and having the trait + centralized post-generation implementation reduces that risk.

src/openforms/submissions/utils.py
```diff
@@ -90,7 +91,11 @@ def remove_submission_uploads_from_session(
 
 def send_confirmation_email(submission: Submission):
     logevent.confirmation_email_start(submission)
+    with translation.override(submission.language_code):
+        _send_confirmation_email(submission)
```
sergei-maertens, line 95: imo the `translation.override` should only happen inside the `_send_confirmation_email` function body, before the `render_email_template` call(s).

Now you're also affecting the resulting language of the `logevent` calls inside the `_send_confirmation_email`, which should use the default language (ideally), otherwise admin users looking at submission logs will get mixed language content.

Probably good to make some assertions about these log events in the tests too.

sergei-maertens (changes requested):

src/openforms/submissions/models/submission_report.py
CharString: That was my breadcrumb that the little birdies forgot to eat. :grimacing: :bird:

src/openforms/submissions/tests/test_submission_report.py
CharString: I expected `submission__form__form_step__form_definition__configuration` to work, but had to change it to `submission__form__formstep__form_definition__configuration`... Where is this inconsistent black wizardry done?

sergei-maertens: `form.formstep_set` is a reverse lookup, and the `generate_minimal_setup` is a Trait that sets up the `formstep` kwarg which generates essentially the m2m through table entry (and exposes the underlying data via that name): https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/tests/factories.py#L26

src/openforms/submissions/utils.py
CharString: I found no clean way to test this in this test.
Looking at a `TimeLineProxy` instance it will always render in the active language at looking time, not at logging time.

Maybe it's an idea to have the logs ensure their output is in settings.LANGUAGE_CODE in a different ticket? I'm guessing production doesn't log to the database and testing on this TimeLineProxy is mute anyway.

sergei-maertens: yeah I was confused - only the structure data that is used as context data is recorded, not the resulting message, that's rendered dynamically according to the active language of the user viewing the logs. Discard my comment!

sergei-maertens (approved):

PR #2443, opened 2022-12-07
[#2436] Copied fieldsets had wrong defaults

### LINKED ISSUE #2436: Error 'Field required' appears too soon

2.0.1, master (1a3520c590a5d62ef4261ef949e31312ce52501a)

Den Haag Taiga Issue 343

Related issues: https://github.com/open-formulieren/open-forms/issues/2054 and https://github.com/open-formulieren/open-forms/issues/2213

Den Haag observed that as they started a form and went to a particular step, the 'field required' error message showed straight away, even before filling anything in.

What seems to be the problem:
The form step in question contains hidden fieldsets with inside a radio component. This fieldset was copied and pasted multiple times. The creation of the variable when copy/pasting this fieldset does not go right and the resulting variable has a default value of `null` instead of `""`. 

Since the default value does not match the expected empty value when filling the form, the backend adds the empty value to the data (and we end up in the situation described in issue #2054)

Create a form with:
- A compulsory radio button with key `radio1` with values `a` and `b`
- A hidden fieldset with frontend logic saying that it should be visible when `radio1` is `a`
- Copy and paste the hidden fieldset

Go to the form:
=> Errors for the compulsory radio show straight away.

The error should only appear if the user tries to submit the step

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2436

### DIFF
--- a/src/openforms/js/components/utils.js
+++ b/src/openforms/js/components/utils.js
@@ -1,0 +1,9 @@
+import FormIOModule from 'formio_module';
+
+const getComponentEmptyValue = component => {
+  const componentClasses = FormIOModule.components;
+  const componentInstance = new componentClasses[component.type]();
+  return componentInstance.emptyValue;
+};
+
+export {getComponentEmptyValue};
--- a/src/openforms/js/components/admin/form_design/variables/utils.js
+++ b/src/openforms/js/components/admin/form_design/variables/utils.js
@@ -2,6 +2,8 @@ import _ from 'lodash';
 import {Utils as FormioUtils} from 'formiojs';
 import {defineMessage} from 'react-intl';
 
+import {getComponentEmptyValue} from 'components/utils';
+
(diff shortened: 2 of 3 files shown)


src/openforms/js/components/utils.js
```diff
@@ -0,0 +1,9 @@
+import FormIOModule from 'formio_module';
+
+const getComponentEmptyValue = component => {
+  const componentClasses = FormIOModule.components;
```
sergei-maertens, line 4: I'd consider looking into formio's component builder itself instead of only our custom module:

* not all component types are registered in our module
* the module only specifies overrides/custom components, formio will fall back to its own components

```js
import Components from 'formio/components/Components';

const instance = Components.create(component);
```

See: https://github.com/formio/formio.js/blob/master/src/components/Components.js#L59

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2455, opened 2022-12-08
[#2398] fix parse_trigger

### LINKED ISSUE #2398: Upgrade to 2.0.1: parse_trigger raises error if component in frontend logic is not present in configuration

2.0.1

If there is a component key in the frontend logic, but this component doesn't exist in the configuration, this step fails (https://github.com/open-formulieren/open-forms/blob/45352ce5d8ff80f66ae758f7c64aa3fdaf2ad06a/src/openforms/forms/utils.py#L305):

```python
if (
            configuration_wrapper
            and configuration_wrapper[trigger_component_key]["type"] == "selectboxes"
        ):
```

In 1.1.8 create a form with:

- A text field with key `textFieldA`
- A fieldset with advanced logic. This has a simple trigger which references `textFieldA`
- Remove the component `textFieldA`

Migrate to 2.0.1

Shouldn't crash

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

pi-sigma:
Fixes #2398

- `openforms.forms.utils.parse_trigger` produced an error when migrating from 1.1.x to 2.0.1 if the migration contained forms which referenced non-existing elements (e.g. because they had been removed)
- the fix adds a check to `parse_trigger` to ensure that elements which trigger logic actions are actually present
- the test only checks that no error occurs on retrieval of formdefinitions after migrating

### DIFF
--- a/src/openforms/forms/tests/test_migrations.py
+++ b/src/openforms/forms/tests/test_migrations.py
@@ -435,3 +435,60 @@ class TestFixRulesWithCheckboxes(TestMigrations):
             self.rule_badly_formatted.json_logic_trigger,
             {"==": [{"var": "selectBoxes.a"}, True]},
         )
+
+
+class TestLogicParsing(TestMigrations):
+    migrate_from = "0002_auto_20210917_1114_squashed_0045_remove_formstep_optional"
+    migrate_to = "0046_convert_advanced_logic"
+    app = "forms"
+
+    def setUpBeforeMigration(self, apps):
+        FormDefinition = apps.get_model("forms", "FormDefinition")
+        Form = apps.get_model("forms", "Form")
+        FormStep = apps.get_model("forms", "FormStep")
+        components = [
+            {
+                "type": "fieldset",
+                "key": "fieldset-1",
+                "logic": [
+                    {
+                        "trigger": {
+                            "type": "simple",
+                            "simple": {
+                                "show": True,
+                                "when": "non-existing-object",
+                                "eq": "trigger value",
+                            },
+                        },
+                        "actions": [
+                            {
+                                "type": "property",
+                                "property": {
+                                    "label": "Hidden",
+                                    "value": "hidden",
+                                    "type": "boolean",
+                                },
+                                "state": False,
+                            }
+                        ],
+                    },
+                ],
+            },
+        ]
+        form_def = FormDefinition.objects.create(
+            slug="form_def", configuration={"components": components}
+        )
+        form = Form.objects.create(slug="form")
+        FormStep.objects.create(form=form, form_definition=form_def, order=0)
+
+    def test_logic_records(self):
+        """Check that form definition is migrated correctly
+
+        The test checks that migrations from 1.1.x to 2.0.1 with form definitions
+        containing references to non-existing objects (a) do not crash and (b) don't
+        trigger any logic actions. The forms themselves are left unmodified (garbage in,
+        garbage out).
+        """
+        FormLogic = self.apps.get_model("forms", "FormLogic")
+        logic_records = FormLogic.objects.all()
+        self.assertFalse(logic_records)
--- a/src/openforms/forms/utils.py
+++ b/src/openforms/forms/utils.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import random
 import string
 import zipfile
@@ -29,6 +30,9 @@ from .api.serializers import (
 from .api.serializers.logic.action_serializers import LogicComponentActionSerializer
 from .constants import EXPORT_META_KEY
 from .models import Form, FormDefinition, FormLogic, FormStep, FormVariable
+
+logger = logging.getLogger(__name__)
+
 
 IMPORT_ORDER = {
     "formDefinitions": FormDefinition,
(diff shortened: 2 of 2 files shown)


src/openforms/forms/tests/test_migrations.py
```diff
@@ -435,3 +435,57 @@ def test_fix_badly_formatted_rules(self):
...
+
+
+class TestLogicParsing(TestMigrations):
+    migrate_from = "0002_auto_20210917_1114_squashed_0045_remove_formstep_optional"
+    migrate_to = "0046_convert_advanced_logic"
+    app = "forms"
+
+    def setUpBeforeMigration(self, apps):
+        self.FormDefinition = apps.get_model("forms", "FormDefinition")
+        self.components = [
```
sergei-maertens, line 447: there's no reason to assign this to `self` here

src/openforms/forms/tests/test_migrations.py
```diff
@@ -435,3 +435,57 @@ def test_fix_badly_formatted_rules(self):
...
         )
+
+
+class TestLogicParsing(TestMigrations):
+    migrate_from = "0002_auto_20210917_1114_squashed_0045_remove_formstep_optional"
+    migrate_to = "0046_convert_advanced_logic"
+    app = "forms"
+
+    def setUpBeforeMigration(self, apps):
+        self.FormDefinition = apps.get_model("forms", "FormDefinition")
```
sergei-maertens, line 446: no reason to assign the model to `self` here - ideally you want the new version of the model in the test anyway

src/openforms/forms/tests/test_migrations.py
```diff
@@ -435,3 +435,57 @@ def test_fix_badly_formatted_rules(self):
...
+                                    "type": "boolean",
+                                },
+                                "state": False,
+                            }
+                        ],
+                    },
+                ],
+            },
+        ]
+        self.FormDefinition.objects.create(
```
sergei-maertens, line 476: no reason to assign to `self` here

src/openforms/forms/tests/test_migrations.py
```diff
@@ -435,3 +435,57 @@ def test_fix_badly_formatted_rules(self):
...
+            name="Definition with textfield",
+            slug="definition-with-textfield",
+            configuration={"components": self.components},
+        )
+
+    def test_parse_trigger(self):
+        """Check that :func:`parse_trigger` does not error out on retrieval of form(definition)
```
sergei-maertens, line 483: this is some confusing phrasing - the function `parse_trigger` will never error out on retrieval of a form definition, since it's not involved in that process at all.

Rather use something like `Assert that the form definition was migrated correctly`

The section below provides sufficient context.

src/openforms/forms/tests/test_utils.py
```diff
@@ -262,7 +262,7 @@ def test_convert_logic_with_selectboxes(self):
         rule = form_step.form.formlogic_set.first()
 
         self.assertEqual(
-            {"==": [{"var": "test1.a"}, True]},
+            {"==": [{"var": "test1.a"}, "a"]},
```
sergei-maertens, line 265: this is wrong. selectboxes have a data value format of:

```json
{
   "a": true,
   "b": false
}
```

and that must be reflected correctly in the json logic trigger

src/openforms/forms/tests/test_migrations.py
```diff
@@ -435,3 +435,57 @@ def test_fix_badly_formatted_rules(self):
...
             {"==": [{"var": "selectBoxes.a"}, True]},
         )
+
+
+class TestLogicParsing(TestMigrations):
+    migrate_from = "0002_auto_20210917_1114_squashed_0045_remove_formstep_optional"
+    migrate_to = "0046_convert_advanced_logic"
+    app = "forms"
+
+    def setUpBeforeMigration(self, apps):
```
sergei-maertens, line 445: you haven't set up a form that uses this form definition, so the migration path that creates `FormLogic` records will never be hit in this test.

Please set up a form with a form step that uses this form definition you create.

src/openforms/forms/tests/test_migrations.py
```diff
@@ -435,3 +435,57 @@ def test_fix_badly_formatted_rules(self):
...
+    def test_parse_trigger(self):
+        """Check that :func:`parse_trigger` does not error out on retrieval of form(definition)
+
+        The test checks that migrations from 1.1.x to 2.0.1 with forms containing
+        references to non-existing objects do not crash. The forms themselves are left
+        unmodified (garbage in, garbage out).
+        """
+
+        form_def = self.FormDefinition.objects.get()
+        self.assertEqual(form_def.configuration["components"][0]["key"], "fieldset-1")
```
sergei-maertens, line 491:
```suggestion
        FormDefinition = self.apps.get_model("forms", "FormDefinition")
        form_def = FormDefinition.objects.get()
        self.assertEqual(form_def.configuration["components"][0]["key"], "fieldset-1")
```

Then, to test the actual migration behaviour, you'll want to query for form logic records and assert none exist (as the creation is skipped due to the bad input logic).

src/openforms/forms/utils.py
```diff
@@ -319,12 +323,19 @@ def parse_trigger(
...
+                    "Logic trigger with non-existing component (%s)",
+                    trigger_component_key,
+                )
+                return
+            else:
+                trigger_component_type = configuration_wrapper[trigger_component_key][
+                    "type"
+                ]
+                if trigger_component_type == "selectboxes":
+                    trigger_component_key = f"{trigger_component_key}.{compare_value}"
```
sergei-maertens, line 338:
```suggestion
                if trigger_component_type == "selectboxes":
                    trigger_component_key = f"{trigger_component_key}.{compare_value}"
                    compare_value = True
```
this bit is missing to pass the existing test

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
backported in 006379ff266db5586644e852a8475017d3e1858e

PR #2457, opened 2022-12-08
Upgrade docker base image to bullseye

### LINKED ISSUE #2312: Upgrade docker images to bullseye
Currently we're based on debian `buster`, while `bullseye` has been out and stable for at least a year now.

sergei-maertens:
Fixes #2312

Verified with `docker-compose` that it all seems to work still.


SilviaAmAm (approved):
After this we need to remember to also upgrade the token-exchange extension: https://github.com/open-formulieren/open-forms-ext-token-exchange/blob/main/Dockerfile

sergei-maertens:
Created https://github.com/open-formulieren/open-forms-ext-token-exchange/pull/12 as well, thanks for the reminder!

PR #2460, opened 2022-12-09
📝 Fix typo and apply tiny stylistic changes.

CharString:
- "To provide" is more specific than "in light of"
 - Down the drain is not the direction we hope to be going. 😉
 - … vs ... 🤷 my editor told me, and I kept it in the commit.


sergei-maertens (approved):
kekw clearly my mind was in the gutter

PR #2468, opened 2022-12-12
:wheelchair: Add autocomplete edit-option

jiromaykin:
fixes open-formulieren/open-forms#2268


jiromaykin:
Autocomplete checkbox is situated here in de Edit dashboard, for both text-field as well as for postcode component:

autocomplete-checkbox

jiromaykin:
Issue: Autocomplete should have a value of 'off'/'on' instead of true/false.

src/openforms/js/components/form/edit/options.js
```diff
@@ -54,6 +54,14 @@ const SHOW_IN_PDF = {
   defaultValue: true,
 };
 
+const AUTOCOMPLETE = {
+  type: 'checkbox',
```
sergei-maertens, line 58: https://formio.github.io/formio.js/app/builder -> this is exposed as a textfield in the vanilla formio builder, which allows specifying on/off as textfield.

So rather than using a checkbox (which results in `true`/`false`), it'd probably be better to use a radio or select dropdown with the values `on`/`off`

src/openforms/js/components/form/edit/options.js
```diff
@@ -54,6 +54,14 @@ const SHOW_IN_PDF = {
...
+  key: 'autocomplete',
+  label: 'Autocomplete',
+  tooltip: 'Display options to fill in the field, based on earlier typed values.',
+  defaultValue: true,
```
sergei-maertens, line 62: I don't see a/the default value in the upstream configuration: https://github.com/formio/formio.js/blob/master/src/components/textfield/editForm/TextField.edit.display.js#L155

I'd reckon we should mostly copy this configuration (but use a radio or dropdown instead for the values). For the full list of values, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values ... maybe it's better to leave this as textfield after all :thinking:

src/openforms/js/components/form/edit/tabs.js
```diff
@@ -28,6 +29,7 @@ const BASIC = {
     LABEL_REQUIRED,
     KEY,
     DESCRIPTION,
+    AUTOCOMPLETE,
```
sergei-maertens, line 32: let's add this to TEXT_BASIC rather than `BASIC`

src/openforms/js/lang/formio/nl.json
```diff
@@ -74,7 +74,8 @@
   "Edit JSON": "Wijzig JSON",
   "Advanced Logic Configured": "Geavanceerde logica geconfigureerd",
   "actions": "acties",
-  "autocomplete": "automatisch aanvullen",
```
sergei-maertens, line 77: I'm afraid this may have broken some frontend translations due to the changed case. Can you add `Autocomplete` as a new item instead of updating this line?

sergei-maertens (changes requested):

src/openforms/js/components/form/edit/options.js
jiromaykin: @sergei-maertens and @joeribekker yes: I guess the idea should be to eventually make a Readme/manual where the user can enter all possibles values for Autocomplete, ranging from on/off to "family-name" etc.

sergei-maertens: users will not do that because it slows them down and distracts them from Getting Shit Done. Optional fields are usually left untouched except by those people that have a passion for getting these sort of things right, but that's only a small subset of people.

jiromaykin: ToDo:
- make autocomplete dropdown with at least these values: https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css-textbox--page
- use Presets (remains a Textfield)
- move 'Postcode' form component to 'depracated' + migrate

jiromaykin:
NB: het password veld zou de autocomplete waarden "password / new-password / current-password" kunnen hebben, maar "new-password" zou nooit vooraf moeten worden ingevuld, lijkt me.

sergei-maertens:
> NB: het password veld zou de autocomplete waarden "password / new-password / current-password" kunnen hebben, maar "new-password" zou nooit vooraf moeten worden ingevuld, lijkt me.

doesn't this cause your browser password manager to generate a new password and store it inside its vault?

jiromaykin:
> doesn't this cause your browser password manager to generate a new password and store it inside its vault?

Ah, I did not know, thanks, I will research.

src/digid_eherkenning_oidc_generics/migrations/0008_auto_20230109_1111.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.2.16 on 2023-01-09 10:11
```
sergei-maertens, line 1: these shouldn't be here - instead you need to fix your local environment by updating the dependencies:

```bash
pip install -r requirements/dev.txt
src/manage.py migrate
```

sergei-maertens (changes requested):

src/openforms/variables/tests/test_rendering.py
```diff
@@ -59,6 +59,7 @@ def test_rendering_user_defined_vars(self):
                         "type": "textfield",
                         "showInEmail": True,
                         "showInPDF": True,
+                        "autocomplete": "on",
```
sergei-maertens, line 62: is this needed?

sergei-maertens (changes requested):

src/openforms/variables/tests/test_rendering.py
jiromaykin: Probably not :) - this was one of the first code lines I wrote in this issue.

sergei-maertens (approved):

PR #2470, opened 2022-12-12
Implement "service fetch" for REST services.

### LINKED ISSUE #2454: Implement the "service fetch" operation
Epic #2169 

Ensure that the/a service fetch configuration is connected to a `FormVariable` so that the variable has sufficient information on how to populate itself.

Initially we focus on JSON-based services - XML will follow later.

* [x] Add an optional FK from `FormVariable` to `ServiceFetchConfiguration`
* [x] Implement a function (to be public API) to evaluate the service fetch, given a `FormVariable` instance and a `values` mapping (`openforms.typing.DataMapping` type)
  This will then be called via a new `openforms.submissions.logic.actions.ActionOperation`
    - This must apply the jq/json-logic mutations
    - This must incorporate the `values` (e.g. for request body templates -> we need to decide on an API/DSL for this, let's discuss)

CharString:
Closes #2454


src/openforms/submissions/logic/rules.py
```diff
@@ -154,19 +159,27 @@ def iter_evaluate_rules(
...
-                # through the rules.
-                if action_details["type"] == LogicActionTypes.variable:
-                    new_value = None
-                    with log_errors(action_details["value"], rule):
-                        new_value = jsonLogic(
-                            action_details["value"], data_container.data
+            for operation in map(compile_action_operation, rule.actions):
+                match operation:
+                    # only assigments to names in the context are done here
+                    case VariableAction(value=expression, variable=name):
```
sergei-maertens, line 165: should we do any assignments at all? I wanted to decouple this as much from the action type itself (right now in master it is too coupled imo) and this introduces even tighter coupling.

It'd maybe make more sense to have the classes/instances expose an attribute `instant_evaluation: bool` to match on and take that information to do the instant evaluation. Then also make the evaluation a method of the action itself so that we keep the implementation as close to the dataclass as possible.

src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -0,0 +1,375 @@
...
+            api_type=APITypes.orc,
+            api_root="https://httpbin.org/",
+            auth_type=AuthTypes.no_auth,
+            oas_file=FileField(
+                from_path=str(Path(__file__).parent.parent / "files" / "openapi.yaml")
+            ),
+        )
+        # prevent parsing the yaml over and over and over
+        cls.service.id = 1  # need pk for __hash__
+        cls.service.build_client = lru_cache(1)(cls.service.build_client)
```
sergei-maertens, line 55: this should be cached in memory by the client though :thinking: See https://github.com/maykinmedia/zgw-consumers/blob/main/zgw_consumers/cache.py#L11

src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -0,0 +1,375 @@
...
+
+# the first test still takes too much time
+# decorating given because I don't want to count on test running order
+given = settings(deadline=500)(given)
+
+
+class ServiceFetchConfigVariableBindingTests(SimpleTestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.service = ServiceFactory.build(
```
sergei-maertens, line 45: I get nervous when the super classmethods are not called... possibly related to your tearDownClass problem?

src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -0,0 +1,375 @@
...
+
+        with requests_mock.Mocker() as m:
+            m.get(requests_mock.ANY)
+            _ = bind(var, context)
+            request = m.last_request
+
+        head, rest = request.url[:26], request.url[26:]
+        self.assertEqual(head, "https://httpbin.org/delay/")
+        # the service should get the value
+        self.assertEqual(unquote(rest), str(field_value))
```
sergei-maertens, line 98: I'm sure there's plenty of value in hypothesis... but I really struggle to read/understand this. What is going on here?

I get really annoyed when I have to execute code in my head to understand assertions instead of simply reading what's supposed to happen.

src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -0,0 +1,375 @@
...
+
+        # I guess if the rest is unmangled, the rest of the assertions are not
+        # really needed; I think
+        # ∄ a way for Bobby Tables to drop a table ∧ become an unmangled entry in it
+        # but asserting is cheaper than a CVE
```
sergei-maertens, line 103: what?

src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -0,0 +1,375 @@
...
+    )
+    def test_it_can_construct_simple_query_parameters(self, field_value, question_mark):
+        # https://swagger.io/docs/specification/describing-parameters/#query-parameters
+        context = {"some_field": field_value}
+
+        var = FormVariableFactory.build(
+            service_fetch_configuration=ServiceFetchConfigurationFactory.build(
+                service=self.service,
+                path="response-headers",
+                query_params=question_mark + "freeform={some_field}",
```
sergei-maertens, line 139: I'm really picturing a UX where users enter key/values and can enter lists as values too, which get serialized down to a mapping/dict in the backend rather than strings. So the question mark in that case becomes irrelevant, especially if we feed the (evaluated/templated out) dict values to something like `furl` or low-level encoding which takes care of all the peculiarities. IIRC the ZDS client can also take params as a dict which is then serialized properly by requests.

src/openforms/variables/tests/factories.py
```diff
@@ -0,0 +1,12 @@
...
+from openforms.registrations.contrib.zgw_apis.tests.factories import ServiceFactory
+
+from ..models import ServiceFetchConfiguration
+
+
+class ServiceFetchConfigurationFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = ServiceFetchConfiguration
+
+    service = factory.SubFactory(ServiceFactory)
```
sergei-maertens, line 12: can we please stick to a consistent style for models/factories... the `Meta` nested class goes below the field declarations, which conforms to django's model code style on https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#model-style

src/openforms/submissions/tests/test_variables/test_fetch_with_logic.py
```diff
@@ -0,0 +1,110 @@
...
+from openforms.registrations.contrib.zgw_apis.tests.factories import ServiceFactory
+from openforms.variables.tests.factories import ServiceFetchConfigurationFactory
+
+from ..factories import SubmissionFactory
+from ..mixins import SubmissionsMixin
+
+
+class ServiceFetchWithActionsTest(SubmissionsMixin, APITestCase):
+    @classmethod
+    def setUpTestData(cls):
```
sergei-maertens, line 20: please call the parent method, mixins are often used to also set up some base test data used in tests

src/openforms/submissions/tests/test_variables/test_fetch_with_logic.py
```diff
@@ -0,0 +1,110 @@
...
+        cls.service = ServiceFactory.create(
+            api_type=APITypes.orc,
+            api_root="https://httpbin.org/",
+            auth_type=AuthTypes.no_auth,
+            oas_file=FileField(
+                from_path=str(Path(__file__).parent.parent / "files" / "openapi.yaml")
+            ),
+        )
+
+    @requests_mock.mock(case_sensitive=True)
```
sergei-maertens, line 30:
```suggestion
    @requests_mock.Mocker(case_sensitive=True)
```
they're the same, but for consistency's sake in the codebase we prefer this variant.

src/openforms/submissions/tests/test_variables/test_fetch_with_logic.py
```diff
@@ -0,0 +1,110 @@
...
+        submission = SubmissionFactory.from_components(
+            [{"type": "number", "key": "someField"}]
+        )
+        fetch_config = ServiceFetchConfigurationFactory.create(
+            service=self.service, path="get"
+        )
+
+        FormLogicFactory.create(
+            form=submission.form,
+            json_logic_trigger={"!": {"var": "someField"}},  # True
```
sergei-maertens, line 41:
```suggestion
            json_logic_trigger=True,
```
also works :)

sergei-maertens (changes requested):

src/openforms/submissions/form_logic.py
```diff
@@ -244,3 +251,61 @@ def check_submission_logic(
...
+                    value = jsonLogic(expression, raw_value)
+                case _:
+                    value = raw_value
+        case _:
+            raise NotImplementedError()
+
+    return value
+
+
+def _request_arguments(fetch_config: ServiceFetchConfiguration, context: DataMapping):
```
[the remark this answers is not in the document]
CharString: :+1:

src/openforms/submissions/form_logic.py
```diff
@@ -244,3 +251,61 @@ def check_submission_logic(
 
     for mutation in mutation_operations:
         mutation.apply(step, {})
+
+
+def bind(var: FormVariable, context: DataMapping) -> JSONValue:
```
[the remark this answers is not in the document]
CharString: Sounds great. I've been hauling this function around a bit.  But it felt "off" everywhere. :+1:

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -109,6 +109,35 @@ const ActionVariableValue = ({action, errors, onChange}) => {
...
+  return (
+    <>
+      <DSLEditorNode errors={errors.variable}>
+        <Select
+          name="variable"
+          choices={getVariableChoices(allVariables)}
+          allowBlank
+          onChange={onChange}
+          value={action.variable}
+        />
```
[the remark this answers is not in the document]
CharString: Well I expect this "variable" choice wil not be needed any more as it should be the FormVariable.

Instead of pre-emptively changing stuff, I'd like to see what @stevenbal comes up with for the modal ui.

src/openforms/submissions/tests/form_logic/test_binding.py
CharString: I think refactoring the Mapping type values a bit tighter maybe to an [ArrayField](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#arrayfield) is also an option... 
If it's okay with you, I'd like to hot potato this decision over to #2694

src/openforms/variables/tests/factories.py
CharString: Woops, Black has made me a bit too lazy in formatting :)

src/openforms/submissions/tests/form_logic/test_binding.py
sergei-maertens: I've instructed Steven to make a key-value UX in the modal for this, where the values may be arrays (especially for querystrings, but only strings for headers). ArrayField cannot store arrays afaik and HStoreField also only takes strings as values, so we'll have to stick to JSONField.

Okay to move this into a different ticket :+1:

src/openforms/submissions/tests/form_logic/test_binding.py
CharString: Yeah, that's what I thought. I've stepped through that line more than I'd like to remember.

src/openforms/submissions/tests/form_logic/test_binding.py
CharString: tl;dr;
I suspect that iff `rest` is well-formed all following assertions are tautologies. But better to be safe than sorry; if being safe is this cheap.


∄ : there does not exist
Bobby Tables: https://xkcd.com/327/
∧: and

src/openforms/submissions/tests/form_logic/test_binding.py
sergei-maertens: perhaps some dummy cache is being setup instead? https://github.com/open-formulieren/open-forms/blob/master/src/openforms/conf/ci.py#L14

src/openforms/submissions/tests/form_logic/test_binding.py
sergei-maertens: I can't even wrap my head around what `head` and `rest` are supposed to be and why all this is happening?

And please stick to plain English in these comments, my mathematical notation knowledge is buried deep in the back of my mind, but I should be able to understand this. That can't be said from colleagues that don't have any compsci or mathematical background at all.

src/openforms/submissions/logic/rules.py
CharString: So this
```py
# If the action type is to set a variable, update the variable state.
# This is the ONLY operation that is allowed to execute while we're looping
# through the rules.
if action_details["type"] == LogicActionTypes.variable:
    ...
```
should be
```py
if rule.instant_evaluation:
    ...
```
?

sergei-maertens: yeah, I feel that makes more sense

Then in the variant for fetching remote service, we can make it an implementation detail that the result is actually saved in the database, whereas for the current "assign to variable" situation we leave them in-memory. That should be a good step towards caching that evaluation result :thinking: 

But probably best to leave that aspect for the dedicated ticket.

src/openforms/submissions/tests/form_logic/test_binding.py
CharString: image

src/openforms/submissions/tests/form_logic/test_binding.py
CharString: Yet it makes sense: the tests are modelling user behaviour. Users can do anything.
image
What puzzles me is the complexity of the implementation: it's a webform.

src/openforms/variables/models.py
```diff
@@ -100,3 +103,36 @@ def clean(self):
...
+        headers = {
+            # no leading spaces and the value into the legal field-value codespace
+            header: value.format(**context).strip().encode("utf-8").decode("latin1")
+            for header, value in (self.headers or {}).items()
+        }
+        # before we go further
+        # assert headers are still valid after we put submmitter data in there
+        HeaderValidator()(headers)
+
+        escaped_values = {k: quote(str(v), safe="") for k, v in context.items()}
```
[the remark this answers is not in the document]
sergei-maertens: additional thought - shouldn't this use `quote_plus` to ensure that spaces in input are also encoded with a `+` char? this also by default sets `safe=""` rather than the `quote` function which has a default `safe="/"`

src/openforms/submissions/logic/actions.py
```diff
@@ -1,14 +1,19 @@
...
 from openforms.formio.service import FormioConfigurationWrapper
 from openforms.forms.constants import LogicActionTypes
-from openforms.forms.models import FormVariable
-from openforms.typing import JSONObject
+from openforms.forms.models import FormLogic, FormVariable
+from openforms.typing import DataMapping, JSONObject
 from openforms.utils.json_logic import ComponentMeta, introspect_json_logic
+from openforms.variables.models import ServiceFetchConfiguration
 
+from ..logic.service_fetching import perform_service_fetch
```
sergei-maertens, line 13:
```suggestion
from .service_fetching import perform_service_fetch
```
I think?

sergei-maertens (changes requested):
I'm happy with the changes!

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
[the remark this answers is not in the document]
CharString: I left the ActionVariable as is; the suggestion changes the semantics of what happens: the select now offers a superset the union of component variables and user defined variables. The ComponentSelection component offers just the components.

src/openforms/variables/models.py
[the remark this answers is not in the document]
CharString: Cleaned it up even more thorough. No cutting corners anymore

src/openforms/submissions/logic/actions.py
CharString: Sure... After lunch and coffee.  I just rebased onto master... Hungry now. :wink:

sergei-maertens (approved):

PR #2472, opened 2022-12-13
[#2362] fix Ogone payment plugin

### LINKED ISSUE #2362: Ogone on Failure/Cancel: Method (POST) not allowed

2.0.0

internal ref: Rid 83
sentry: 318987

When a payment is cancelled or failed Ogone prompts the user to Cancel or go back:
image
When the user selects Back, a POST request is made to Open Forms:
image
which is not allowed:

``` json
{
    "type": "https://formulieren-xxxxxxxxx.nl/fouten/MethodNotAllowed/",
    "code": "method_not_allowed",
    "title": "Methode \"{method}\" niet toegestaan.",
    "status": 405,
    "detail": "Methode \"POST\" niet toegestaan.",
    "instance": "urn:uuid:dd1ce704-e86f-xxxx-xxxx-af9f8e3d740f"
}
```

This exception is raised because plugin.return_method is set to "GET".

pi-sigma:
Fixes #2362

- Allow both GET and POST requests in the return method of the plugin and adjust the corresponding view
- The ticket mentions two further requirements: making the endpoint CSRF-exempt and validating the origin via SHA-IN/SHA-OUT. These are already satisfied (there was no CSRF-protection for Ogone, and the origin of the request is already validated in the `handle_return` method of the plugin (via `client.get_validated_params()`)

### DIFF
--- a/src/openforms/payments/contrib/ogone/plugin.py
+++ b/src/openforms/payments/contrib/ogone/plugin.py
@@ -40,6 +40,7 @@ RETURN_ACTION_PARAM = "action"
 
 @register("ogone-legacy")
 class OgoneLegacyPaymentPlugin(BasePlugin):
+    return_method = ("GET", "POST")
     verbose_name = _("Ogone legacy")
     configuration_options = OgoneOptionsSerializer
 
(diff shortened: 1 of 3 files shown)


pi-sigma:
The test fails at two points:
1. `get_validated_params()` of the `OgoneClient` (`openforms/payments/contrib/ogone/client.py`) raises an `InvalidSignature` since the `SHASIGN` is missing
2. even if the `SHASIGN` is present, the `OgoneLegacyPaymentPlugin` raises a `KeyError` because it expects the presence of a `STATUS` variable when calling `self.apply_status()` (`openforms/payments/contrib/ogone/plugin.py`)

joeribekker:
Refinement: This seems to be a bug in Ogone that we cannot fix. Discisions are pending.

PR #2485, opened 2022-12-15
:bug: [#2484] Disable fallbacks for modeltranslations

### LINKED ISSUE #2484: Global translations of literals do not work for certain values

2.1.0-alpha

If I have a form that does not define translations for literals (such as begin text) that I want to display in English, I cannot fall back on the global literal translations for certain values, instead it shows the Dutch default.

1. Create a Form with no values for `Knopteksten`
2. Make sure in the `Algemene configuratie`, the `Formulier starten-label [en]`  is `Begin form`
3. Load the form using the SDK and switch to Dutch
4. Observe that the start button does not say `Begin form`
5. Change the global translation to anything other than `Begin form`, and reload the SDK to observe that it works

I would expect the default to always work. I'm not sure why it doesn't work for `Begin form`, maybe this is the default of the global default?

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
Fixes #2484

### DIFF
--- a/src/openforms/config/translation.py
+++ b/src/openforms/config/translation.py
@@ -20,3 +20,18 @@ class GlobalConfigurationTranslationOptions(TranslationOptions):
         "form_step_next_text",
         "privacy_policy_label",
     )
+    fallback_undefined = {
+        "submission_confirmation_template": "",
+        "confirmation_email_subject": "",
+        "confirmation_email_content": "",
+        "save_form_email_subject": "",
+        "save_form_email_content": "",
+        "form_previous_text": "",
+        "form_change_text": "",
+        "form_confirm_text": "",
+        "form_begin_text": "",
+        "form_step_previous_text": "",
+        "form_step_save_text": "",
+        "form_step_next_text": "",
+        "privacy_policy_label": "",
+    }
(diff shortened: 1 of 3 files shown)


src/openforms/conf/base.py
```diff
@@ -1119,6 +1119,7 @@
 # Django Modeltranslation
 #
 MODELTRANSLATION_DEFAULT_LANGUAGE = "nl"
+MODELTRANSLATION_ENABLE_FALLBACKS = False
```
stevenbal, line 1122: @sergei-maertens I'm not sure if there is a scenario in which we do want to fallback to another language?

sergei-maertens: I think it's realistic and falling back to NL as a default is sane behaviour. Perhaps https://django-modeltranslation.readthedocs.io/en/latest/usage.html#fallback-undefined is a better, more targeted approach to not treat empty string for those literal fields as a missing translation?

Using a sentinel `empty = object()` seem like valid options here (since `None` would be the non-default field empty value as mt-fields are nullable)

sergei-maertens (approved):

PR #2489, opened 2022-12-15
:heavy_plus_sign: [#2487] Add import sorting plugin for prettier 

### LINKED ISSUE #2487: Add import sorting plugin to prettier
https://dev.to/diballesteros/how-to-quickly-sort-imports-with-prettier-3po7

stevenbal:
Closes #2487


.prettierrc.json
```diff
@@ -11,5 +11,8 @@
   "singleQuote": true,
   "tabWidth": 2,
   "trailingComma": "es5",
-  "useTabs": false
+  "useTabs": false,
+  "importOrder": ["^components/(.*)$", "^[./]"],
```
sergei-maertens, line 15: this is missing the following regexes:

* `^utils/(.*)$`
* `^lang/(.*)$`

sergei-maertens (changes requested):
Can you also add the commit(s) that do the sorting via prettier to the `.git-blame-ignore-revs` file?

I haven't fully vetted this plugin yet, but I'm wondering if it can handle side-effect imports properly (like `initTinymce`) where importing the module results in side effects.

sergei-maertens (approved):

Issue #2475, opened 2022-12-13
Form designer FormIO translations UI: WYSIWYG content field translations


stevenbal:
Related issue: https://github.com/open-formulieren/open-forms/issues/2315

Issue raised in https://github.com/open-formulieren/open-forms/pull/2437

The `vrije tekst` field is a WYSIWYG content field, which will not work properly with the default translations setup implemented in the above PR. It should probably have a different translations tab that only shows one WYSIWYG editor per language, because only the content is translatable.

Question: will this content be picked up by FormIO's translation engine, or should we translate it when serializing the FormStepDefinition?

sergei-maertens:
> Question: will this content be picked up by FormIO's translation engine, or should we translate it when serializing the FormStepDefinition?

It will be picked up, at least according to the default template: https://github.com/formio/formio.js/blob/4.13.x/src/templates/bootstrap/html/form.ejs#L5

```ejs
{{ctx.t(ctx.content)}}
```
the `t` function translates it.

CharString:
Closed by #2544 (don't know why this wasn't automagically picked up)

Issue #2286, opened 2022-11-01
Extend API endpoint `/api/v1/submissions/:uuid/steps/:uuid` to return the `FormDefinition` translations


stevenbal:
Epic #2170 - depends on #2315

The goal of this feature is to relay the component-level translations in the submission step endpoint so that they can be passed to the formio `Form` component _for the active locale_.

Challenges to take into account:

* labels/descriptions... etc. support template evaluation with the submission data -> translations need to do so too
* there may be dirty data used that alters the template evaluation result (during a logic check)

The above challenges require us to process the language information in `SubmissionStepSerializer.to_representation`:

1. Modify the signature of `evaluate_form_logic` - rather than returning the updated configuration, it should return a tuple/datastructure of the configuration and the instantiated `DataContainer` (so that the mutations by logic can be re-used)
2. Define the serializer field `SubmissionStepSerializer.translations` as a `SerializerMethodField`. API documentation should document this as an object with `string` values. The `get_translations` implementation must look up and return `submission_step.form_step.form_definition.component_translations.<language_code>`
3. Update `SubmissionStepSerializer.to_representation` to use the `data_container` return value from `evaluate_form_logic` - this is the input data to template out `representation["translations"]`
4. Use `openforms.template.render_from_string` similarly to `openforms.formio.variables.inject_variables` to evaluate the templates in translation literals. Note that this is not exactly the same - injecting variables processes nested formio configurations, while the translation is a simple flat key-value map.

PR #2501, opened 2022-12-16
Issue/2390 ensure pdf reports and emails render in submission lang

### LINKED ISSUE #2390: Ensure that submission PDF reports and confirmation e-mails are rendered in the submission language

CharString:
Closes #2390

TODO

- [x] Check number of queries made

### DIFF
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -8,7 +8,7 @@ from openforms.submissions.models import SubmissionStep
 from openforms.submissions.rendering.base import Node
 from openforms.submissions.rendering.constants import RenderModes
 
-from ..service import format_value
+from ..service import format_value, translate_function
 from ..typing import Component
 from ..utils import is_layout_component, is_visible_in_frontend, iter_components
 
@@ -73,15 +73,9 @@ class ComponentNode(Node):
         # labels through the translations, before further processing of logic
         # etc.
         if self.renderer.form.translation_enabled and self.step.form_step:
-            lang = self.renderer.submission.language_code
-            translations = (
-                self.step.form_step.form_definition.component_translations.get(lang, {})
+            self.apply_to_labels(
+                translate_function(self.renderer.submission, self.step)
             )
-
-            def translate(string: str) -> str:
-                return translations.get(string) or string
-
-            self.apply_to_labels(translate)
 
     @property
     def is_visible(self) -> bool:
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -57,13 +57,18 @@ class ComponentNode(Node):
         from .registry import register
 
         if (
-            (label := component.get("label"))
-            and step.form_step is not None
+            step.form_step is not None
             and step.form_step.form.translation_enabled
             and (lang := step.submission.language_code)
         ):
+            # Value Formatters have no access to the translations
+            # adjust value labels (for radio, checkbox etc)
             translations = step.form_step.form_definition.component_translations
-            component["label"] = translations.get(lang, {}).get(label, label)
+            for value in component.get("data", component).get("values", []):
+                if value_label := value.get("label"):
+                    value["label"] = translations.get(lang, {}).get(
+                        value_label, value_label
+                    )
 
         node_cls = register[component["type"]]
         nested_node = node_cls(
@@ -216,7 +221,13 @@ class ComponentNode(Node):
         """
         if self.mode == RenderModes.export:
             return self.component.get("key") or "KEY_MISSING"
-        return self.component.get("label") or self.component.get("key", "")
+        label = self.component.get("label") or self.component.get("key", "")
+        return self._localise(label)
+
+    def _localise(self, string):
+        translations = self.step.form_step.form_definition.component_translations
+        lang = self.step.submission.language_code
+        return translations.get(lang, {}).get(string, string)
 
     @property
     def display_value(self) -> Union[str, Any]:
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -56,6 +56,15 @@ class ComponentNode(Node):
         """
         from .registry import register
 
+        if (
+            (label := component.get("label"))
+            and step.form_step is not None
+            and step.form_step.form.translation_enabled
+            and (lang := step.submission.language_code)
+        ):
+            translations = step.form_step.form_definition.component_translations
+            component["label"] = translations.get(lang, {}).get(label, label)
+
         node_cls = register[component["type"]]
         nested_node = node_cls(
             step=step,
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -1,6 +1,6 @@
 import copy
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Iterator, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Iterator, Literal, Optional, Union
 
 from glom import Path, assign, glom
 
@@ -56,20 +56,6 @@ class ComponentNode(Node):
         """
         from .registry import register
 
-        if (
-            step.form_step is not None
-            and step.form_step.form.translation_enabled
-            and (lang := step.submission.language_code)
-        ):
-            # Value Formatters have no access to the translations
-            # adjust value labels (for radio, checkbox etc)
-            translations = step.form_step.form_definition.component_translations
-            for value in component.get("data", component).get("values", []):
-                if value_label := value.get("label"):
-                    value["label"] = translations.get(lang, {}).get(
-                        value_label, value_label
-                    )
-
         node_cls = register[component["type"]]
         nested_node = node_cls(
             step=step,
@@ -81,6 +67,21 @@ class ComponentNode(Node):
             parent_node=parent_node,
         )
         return nested_node
+
+    def __post_init__(self):
+        # Value Formatters have no access to the translations; run all our
+        # labels through the translations, before further processing of logic
+        # etc.
+        if self.renderer.form.translation_enabled and self.step.form_step:
+            lang = self.renderer.submission.language_code
+            translations = (
+                self.step.form_step.form_definition.component_translations.get(lang, {})
+            )
+
+            def translate(string: str) -> str:
+                return translations.get(string) or string
+
+            self.apply_to_labels(translate)
 
     @property
     def is_visible(self) -> bool:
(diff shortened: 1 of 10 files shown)


CharString:
Cherrypicked black and flake8 bump from 2f178a0b24032b79250b762.

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -106,24 +269,73 @@ def test_report_is_generated_in_same_language_as_submission(self):
...
+        self.assertIn("Translated Repeating Group label", html_report)
+        self.assertIn("Translated Repeating Group Item label", html_report)
+        self.assertNotIn("Untranslated Field Set label", html_report)
+        self.assertNotIn("Untranslated Columns label", html_report)
+        self.assertNotIn("Untranslated Repeating Group label", html_report)
+        self.assertNotIn("Untranslated Repeating Group Item label", html_report)
+
+        # assert we've tested every component_type ct in register
+        plugins = set(ct for ct, _ in register.items())
+        plugins.remove("content")  # WYSIWIG seems untranslated in SDK
```
CharString, line 322: #2475

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -98,6 +100,167 @@ def test_report_generation(self, mock_request):
...
+            ("time", "23:50"),
+            ("updatenote", "C#"),
+        ]
+        submission_data = {}
+        translations = {}
+        components = []
+
+        # carthesian products are big, let's loop to fill these
+        for component_type, user_input in fields:
+            # common base
```
sergei-maertens, line 134: perhaps defining some (pure python) factories in `openforms.formio.tests.factories` would help - you can use factory boy to generate/build [dicts](https://til.codeinthehole.com/posts/factory-boy-can-be-used-to-build-dicts/) :-)

That would then also line up better with the data structures used in `openforms.formio`

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -98,6 +100,167 @@ def test_report_generation(self, mock_request):
...
+            }
+            submission_data[f"{component_type}-key"] = user_input
+            translations[
+                f"Untranslated {component_type.title()} label"
+            ] = f"Translated {component_type.title()} label"
+
+            # add the component
+            components.append(component)
+
+            # but massage the weirder components into shape
```
sergei-maertens, line 150: I'm envisioning this a bit more like:

```py


components_and_data = (
    (BSNComponentFactory, "bsn", {"extra": "bsn factory kwargs"}, 111222333),
    (CheckboxComponentFactory, "checkbox", {}, True},
    ...
)
for factory_cls, c_type, extra, user_input in components_and_data:
    component = factory_cls(key=c_type, label=f"Untranslated {c_type} label", showInPDF=True, hidden=false, **extra)
    submission_data[c_type] = user_input

```

so that you can:

* encapsulate component-specific options/configs per factory (while sharing the common attributes
* keep all factory-specifics on a single item
* you can rely on factory details to fill out some irrelevant datastructures while keeping the component definition representative
* simplifies the loop here used and does away with the `match` statement

Also note that checkboxes end up as true/false values, and the user input for a select for example is not the label, but the underlying value. You can review `DefaultFormatterTestCase` which uses json files for definitions _and_ user input values that were created by actually submitting the formio forms :)

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -98,6 +100,167 @@ def test_report_generation(self, mock_request):
...
+            match component:
+                case {"type": "file", "key": key}:
+                    submission_data[key] = [{"originalName": "download(2).pdf"}]
+                case {"type": "radio", "key": key}:
+                    component["values"] = [
+                        {"label": "Untranslated Radio option one", "value": "radioOne"},
+                        {"label": "Untranslated Radio option two", "value": "radioTwo"},
+                    ]
+                    submission_data[key] = "radioOne"  # value of the selected label
+                    translations["Untranslated Radio option one"] = "Radio number one"
```
sergei-maertens, line 160: I think I'd rather see a single literal dict defined outside of the loop that accounts for all the translatable strings - it is closer to Formio's model that you can only have a single key-value map for the translations and that it's _not_ at the component level.

And following my own reasoning, it'd probably make more sense to set up `submission_data` by hand too instead of doing that programmatically.

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -106,24 +269,73 @@ def test_report_is_generated_in_same_language_as_submission(self):
...
+        self.assertIn("June 29, 1911", html_report)
+        self.assertIn("11:50 p.m.", html_report)
+        self.assertIn("1,100,000.00", html_report)
+
+        for component_type, user_input in fields:
+            with self.subTest(f"FormIO label for {component_type}"):
+                if component_type in ["date", "time", "currency"]:
+                    pass  # these are localized
+                else:
+                    self.assertIn(user_input, html_report)
```
sergei-maertens, line 302: now I understand the `user_input` variable, its the formatted user input.

Probably better to keep system input + formatted display together in a dict/datastructure so it's a bit more readable/clear what's happening in the test.

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -106,24 +269,73 @@ def test_report_is_generated_in_same_language_as_submission(self):
...
+                self.assertNotIn(
+                    f"Untranslated {component_type.title()} label", html_report
+                )
+                self.assertIn(f"Translated {component_type.title()} label", html_report)
+
+        # assert structural labels
+        self.assertIn("Translated Field Set label", html_report)
+        self.assertNotIn(
+            "Translated Columns label", html_report
+        )  # 1451 -> never output a label
```
sergei-maertens, line 312: I'd also add the assertion for the untranslated string since a bug in setting up the translations for this label would still pass this assertion.

src/openforms/submissions/tests/test_submission_report.py
sergei-maertens, line 322: it will be actually! see the [`{{ctx.t(ctx.content)}}`](https://github.com/formio/formio.js/blob/4.13.x/src/templates/bootstrap/html/form.ejs#L5) snippet, paired with: https://github.com/formio/formio.js/blob/4.13.x/src/components/content/Content.js#L45

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -560,6 +560,11 @@ def test_send_confirmation_email_when_appointment_is_changed(self):
         self.assertEqual(message.subject, f"Confirmation mail {_('(updated)')}")
 
     def test_template_is_rendered_in_submission_language(self):
+        """A subset of the components, with particularly weird APIs.
+        Translations on all FormIO components are exercised (until exorcised) in
+        .test_submission_report.DownloadSubmissionReportTests:test_report_is_generated_in_same_language_as_submission
+        """
```
sergei-maertens, line 566:
```suggestion
        """
        Assert a subset of the components with particularly weird APIs is translated correctly.
        
        Translations on all FormIO components are exercised (until exorcised) in
        ``DownloadSubmissionReportTests.test_report_is_generated_in_same_language_as_submission``
        (in the ``.test_submission_report`` module).
        """
```

Running tests with `-v 2` or higher prints the docstring of the test intead of the test path (following PEP 257 semantics), so make sure to provide test-specific meaningful summaries.

src/openforms/submissions/tests/test_tasks_confirmation_emails.py
```diff
@@ -573,19 +578,88 @@ def test_template_is_rendered_in_submission_language(self):
...
                             "type": "textfield",
-                            "label": "Foo",
+                            "label": "Label with no translation",
+                            "showInEmail": True,
+                        },
+                        {
+                            "key": "fubar",
+                            "type": "password",
+                            "label": "Label with no translation",
+                            # TODO add a constraint, I don't want this state in my db!
```
sergei-maertens, line 619: not sure I understand, but since this stored in a JSONField and the datastructure itself is recursive/arbitrary nested, I don't think it's possible to express a database constraint for this at all.

sergei-maertens (changes requested):
Can you add a test that combines translations and variabel interpolation in a label?

E.g. a component with label `{{ foo }}` where `foo` is an earlier textfield in the form step (and thus registered as FormVariable).

src/openforms/submissions/tests/test_submission_report.py
CharString: So did I :) It's a few lines down. I'll re-order the assertions so these two are adjacent.

src/openforms/formio/rendering/nodes.py
```diff
@@ -191,7 +205,13 @@ def label(self) -> str:
...
         if self.mode == RenderModes.export:
             return self.component.get("key") or "KEY_MISSING"
-        return self.component.get("label") or self.component.get("key", "")
+        label = self.component.get("label") or self.component.get("key", "")
+        return self._localise(label)
+
+    def _localise(self, string):
+        translations = self.step.form_step.form_definition.component_translations
+        lang = self.step.submission.language_code
+        return translations.get(lang, {}).get(string, string)
```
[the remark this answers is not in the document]
CharString: I messed around with glom, but it got very messy, especially in my head. It was one symbol to many for me to grok when I returned to reading the code and continue work. Maybe I just need more exposure to it. But I opted for extracting it out to a single `translation_function` I put in `formio.services` that can be applied to strings that need translations at times they need to be translated.

CharString:
A final thought.

I was thrown off by the existence of Components and ComponentNodes, resp. the dictionaries used in the form logic code and the dataclasses used in the rendering. And by extension the existence of .registry and ..registry. It's a nice separation of concerns... until it isn't like with the interpolation bit. Since they both operate on the same dependency tree structure, maybe the separation could be: tree and visitor(s). The visitor could apply/trigger logic, emit html for email/pdf or emit export formats. The tree is strictly data, all side-effects and manipulation go in a/the visitor.

CharString:
Rebased I need the translation stuff for #2286

sergei-maertens:
> Since they both operate on the same dependency tree structure...

I don't think this holds entirely true when looking at repeating group components :thinking: 

That being said, I'm in favour of doing the abstractions that way, it *should* be possible to refactor that purely by changing implementation details while keeping public API (mostly) intact. I do want to keep the formio abstraction contained as much as possible though (i.e. avoid it leaking into places where it really shouldn't).

Feel free to create a ticket to put on the backlog, we can then come up with some high level design and can plan for the Here Be Dragons parts.

sergei-maertens (approved):
LGTM

PR #2503, opened 2022-12-19
[#2490] add information about appearance to login logo

### LINKED ISSUE #2490: Add extra meta-information about the login logo to the API endpoint
Required for https://github.com/open-formulieren/open-forms-sdk/issues/322

The login button icon logo information currently expose the URL to the static asset, URL and title. We need an extra prop `appearance` with possible enum values `dark` and `light`. This is to be hardcoded for each plugin and should be set depending on the included static asset (digid -> dark, eherkenning -> light, eidas -> light etc.)

Relevant API documentation: http://localhost:8000/api/v2/docs/#tag/forms/operation/forms_retrieve

pi-sigma:
Fixes #2490


src/openforms/authentication/api/serializers.py
```diff
@@ -29,6 +29,11 @@ class LoginLogoSerializer(serializers.Serializer):
         help_text=_("Information link to the authentication provider"),
         read_only=True,
     )
+    appearance = serializers.URLField(
```
sergei-maertens, line 32: why is this a `URLField`? it's a field that only has to possible choices: `dark` or `light`, and it should be output as `enum` in the API schema documentation.

src/openforms/contrib/digid_eherkenning/utils.py
```diff
@@ -7,11 +7,13 @@ def get_digid_logo(request) -> Dict[str, str]:
...
+        "appearance": "dark",
     }
 
 
 def get_eherkenning_logo(request) -> Dict[str, str]:
     return {
         "image_src": request.build_absolute_uri(static("img/eherkenning.png")),
         "href": "https://www.eherkenning.nl/",
+        "appearance": "light",
     }
```
sergei-maertens, line 19: this should use the `LoginLogo` data class instead of returning a raw dict

pi-sigma: I'm wondering if `utils.py` should be deleted and the functions implemented directly in `openforms/authentication/contrib/digid/plugin.py` and `openforms/authentication/contrib/digid_eherkenning_oidc/plugin.py`. It seems that nothing is gained (except a layer of indirection) by exporting them to a separate file.

src/openforms/authentication/contrib/eherkenning/plugin.py
```diff
@@ -120,4 +120,5 @@ def get_logo(self, request) -> Optional[LoginLogo]:
             title=self.get_label(),
             image_src=request.build_absolute_uri(static("img/eidas.png")),
             href="https://digital-strategy.ec.europa.eu/en/policies/eu-trust-mark",
+            appearance="light",
```
sergei-maertens, line 123: please use the constant `LogoAppearance` you defined

src/digid_eherkenning_oidc_generics/migrations/0007_auto_20221213_1347.py
```diff
@@ -0,0 +1,69 @@
+# Generated by Django 3.2.16 on 2022-12-13 12:47
```
sergei-maertens, line 1: eh... these migration changes don't seem right

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2505, opened 2022-12-19
Fix modal crash

### LINKED ISSUE #1790: Clicking the link to display forms affected by changing a reusable form definition crashes the form builder

latest

Clicking the link in the warning in the screenshot crashes the editor:

image

Error from dev console:

image

_No response_

No crash

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

Brave Version 1.44.112 Chromium: 106.0.5249.119 (Official Build) (64-bit)

sergei-maertens:
Fixes #1790

This PR also fixes an import bug that I noticed while trying to import a form to reproduce the original issue.

### DIFF
--- a/src/openforms/js/components/admin/tables/ChangelistColumn.js
+++ b/src/openforms/js/components/admin/tables/ChangelistColumn.js
@@ -6,6 +6,8 @@ const ChangelistColumn = ({objProp, isBool = false, children}) => {
   return null;
 };
 
+ChangelistColumn.tableRole = 'column';
+
 ChangelistColumn.propTypes = {
   objProp: PropTypes.string.isRequired,
   isBool: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


src/openforms/js/components/admin/tables/ChangelistColumn.js
```diff
@@ -6,6 +6,8 @@ const ChangelistColumn = ({objProp, isBool = false, children}) => {
   return null;
 };
 
+ChangelistColumn.tableRole = 'column';
```
sergei-maertens, line 9: I decided to use the react-tabs approach instead :)

SilviaAmAm (approved):
Does the fix for the import need to be backported?

sergei-maertens:
> Does the fix for the import need to be backported?

I don't think so, the performance updates were post 2.0.x, but I'll make sure to confirm.

PR #2514, opened 2022-12-22
[#2198] Add AOF persistence to Redis container

### LINKED ISSUE #2198: Submission with 'pending' registration are not processed if celery is restarted
### Thema / Theme

Other

Den Haag Taiga user story 288, issue 284

In Den Haag load tests were done and Celery stopped processing tasks because redis didn't have enough resources. So, redis resources were increased and the Celery pod restarted. It was observed that the submissions which still hadn't been registered did not get processed after restarting Celery.

What happened:

-  Registration task is queued (retry flag False), but celery is stuck so doesn't pick it up.
-  We restart celery => Task was not processed by celery, retry flag is still false
-  Registration is not retried

In the admin, the action to retry the registration does not work, because it filters on submissions with status "failed" rather than "pending": https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/admin.py#L315 => So we cant select all the ‘pending’ submissions and make them retry in bulk. 
Note: It is on purpose that we filter only on failed submissions. Pending submissions could still be queued and we would risk registering things twice.

After doubling the resources and redoing a load test, some tasks failed but were not retried. What happened:

- For the tasks that time out (celery timeout) the status of the registration will be still 'pending' and the retry flag false. So it is also not retried.

Den Haag would like:
- To be able to retry the registration task for pending tasks
- Deal with tasks that time out

_No response_

_No response_

SilviaAmAm:
Fixes #2198 

### Concepts important to understand the problem

**Celery late acknowledgement:**
In celery, workers acknowledge messages to signify that a message has been handled. Failing to acknowledge the message will cause it to be redelivered. We use late acknowledgement, which means that the task (delivered as a message) is acknowledged _after_ being executed (so if the worker dies while executing, the message is redelivered).

**Celery prefetch limit:**
The prefetch limit is the maximum number of unacknowledged messages that a worker can prefetch. The setting `worker_prefetch_multiplier` is the number of messages to prefetch per process per worker. We have it set to 1, which means that the worker process can hold at most 1 task in addition to the one that's executing.

**Redis snapshotting:**
Redis by default runs with snapshotting enabled. This means that Redis saves the DB to disk if a certain number of write operations have been performed in a certain period of time. By default, Redis will save the DB:

* After 3600 seconds (an hour) if at least 1 change was performed
* After 300 seconds (5 minutes) if at least 100 changes were performed
* After 60 seconds if at least 10000 changes were performed

### Problem

If Redis runs out of memory and crashes, all tasks that were queued and not picked up by a worker are lost.

### Solution

To solve this, we added AOF (Append Only File) persistence to Redis. This means that every write operation received by the server will be logged. On restarting Redis after it has been killed, its state before the crash will be regenerated and therefore no tasks will be lost.

### DIFF
--- a/docs/installation/redis.rst
+++ b/docs/installation/redis.rst
@@ -4,21 +4,73 @@
 Redis configuration
 ===================
 
-To prevent Celery tasks to be lost in the case where Redis crashes, it it advisable to run Redis with `AOF`_
-(Append Only File) persistence enabled.
+.. note:: The intended audience for this documentation is DevOps engineers
 
-This is because Redis by default runs with snapshotting enabled.
-This means that Redis saves the DB to disk if a certain number of write operations have been performed in a certain
-period of time. By default, Redis will save the DB (see the default `redis.conf file`_ for more details):
+In Open-Forms, `Redis`_ is used as cache, as message broker (for Celery) and as a backend for Celery to store the results
+of the tasks.
+
+Redis is an in-memory data-store, which means that it is susceptible to data loss in the event of abrupt termination or
+power failures unless the default configuration is modified [#]_.
+
+.. _Redis: https://redis.io/
+.. [#] https://docs.celeryq.dev/en/stable/getting-started/first-steps-with-celery.html#redis
+
+Configuring Redis as robust message broker
+==========================================
+
+By default, Redis runs with snapshotting enabled.
+This means that Redis saves the DB (database) to disk if a certain number of write operations have been performed in a
+certain period of time. By default, Redis will save the DB (see the default `redis.conf file`_ for more details):
 
 * After 3600 seconds (an hour) if at least 1 change was performed
 * After 300 seconds (5 minutes) if at least 100 changes were performed
 * After 60 seconds if at least 10000 changes were performed
 
-In Open Forms, Redis is used by Celery as both the broker and the results backend. If Redis crashes (for example because
-it runs out of memory), all tasks that were queued and not picked up by a worker would be lost if no persistence is
-enabled. This means that submissions might not be processed and sent to the registration backend.
-
-
-.. _AOF: https://redis.io/docs/management/persistence/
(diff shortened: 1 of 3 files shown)


docs/installation/redis.rst
```diff
@@ -0,0 +1,24 @@
+.. _installation_redis:
```
sergei-maertens, line 1: I'd expand this documentation section a bit more to provide context.

- [ ] Add a note at the top for the indented audience (DevOps), similar to https://open-forms.readthedocs.io/en/latest/installation/security.html

- [ ] Provide an introduction with high-level context:
    * Redis is used by Open Forms as cache, message broker and result store
    * Redis primarily is an in-memory store which implies data loss is possible unless configuration is tweaked

- [ ] Provide a section "configuring redis as robust message broker"
    * Explain the default config and RDB snapshotting here - point out the risk of data loss with default config
    * Explain AOF mechanism as easiest workaround (and note that the window for data loss is now reduced to max 1s)
    * Mention that alternatives to AOF are available but out of scope for our documentation: HA clustering, replication, sentinel... -> refer to the redis documentation. You could also mention that an alternative broker such as RabbitMQ could be an option (and configured through `CELERY_BROKER_URL`), but is out of scope for our documentation. Refer to the celery docs for this.

- [ ] Provide a section "Deploying multiple redis instances"
    * Point out that we share a single redis instance for cache, message broker and result store because of convenience
    * Point out that it's possible the AOF/clustering configuration *may* impact cache/result store performance and vice-versa (cache/result store use can lead to many writes to the AOF file which you might not want)
    * Document the option to deploy a redis instance for each individual use case and that these can be configured through the envvars `CELERY_BROKER_URL`, `CACHE_<suffix>` and `CELERY_RESULT_BACKEND` (link to the env configuration reference) so you can use different redis configuration files per instance, tuned for the intended usage.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2524, opened 2023-01-02
Describe logic rules in editor

sergei-maertens:
Part of #2450

**Changes**

* Adds model/DB fields for logic rule description
* Implements generating a "human readable" representation of a logic rule
* Data migration populates existing rule descriptions from this generator

Scoped for different PRs:

* Refactor existing rule introspection for logging purposes to use new introspection API
* Add API endpoint to get a rule description given a logic expression
* Allow editing rule descriptions in the logic builder UI
* Revamp logic expression builder:
    * Allow defining common expression "presets" (which inject some `_meta.description` properties)
    * Make logic node/tree expression point-and-click instead of entering JSON
    * Allow power-users to still write JSON directly


src/openforms/utils/json_logic/descriptions.py
```diff
@@ -0,0 +1,325 @@
...
+            if not for_root and wrap:
+                description = f"({description})"
+            return description
+
+        return _get_description
+
+    return decorator
+
+
+def op_var(operation: Operation, for_root=False) -> str:
```
SilviaAmAm, line 115: Id the option `for_root` used anywhere?

src/openforms/utils/json_logic/descriptions.py
```diff
@@ -0,0 +1,325 @@
...
+        return f"-{args[0]}"
+    return " - ".join(args)
+
+
+op_division = JoinArgs(" / ")
+op_min = FunctionLike(_("minimum of ({args})"))
+op_max = FunctionLike(_("maximum of ({args})"))
+
+
+def op_merge(operation: Operation, for_root=False) -> str:
```
SilviaAmAm, line 284: Also here, is `for_root` needed?

SilviaAmAm (approved):

src/openforms/utils/json_logic/descriptions.py
sergei-maertens: not used here, but the signature must match the [protocol](https://github.com/open-formulieren/open-forms/blob/818c34b0e5a038c8f7c5457a0e8c4a9e298acd7c/src/openforms/utils/json_logic/descriptions.py#L18) - other callbacks do require it.

src/openforms/utils/json_logic/descriptions.py
sergei-maertens: Similar here, it's because of the Python callback protocol: https://github.com/open-formulieren/open-forms/blob/818c34b0e5a038c8f7c5457a0e8c4a9e298acd7c/src/openforms/utils/json_logic/descriptions.py#L18

PR #2526, opened 2023-01-03
[#2511] Add Microsoft Graph folder configuration

### LINKED ISSUE #2511: Add Site-ID and folder support for Microsoft Graph registration plugin
Currently, the Microsoft Graph registration plugin does not offer any options:

image

However, we should be able to provide a path, like: `/sites/my-site/open-forms/` and all files should be stored there, in a folder. So, `/sites/my-site/open-forms/OF-12345/`, `/sites/my-site/open-forms/OF-67890/`, etc. (name of the folder should be simply the submission registration identifier that we already have at that point).

Background: https://learn.microsoft.com/en-us/graph/api/driveitem-post-children?view=graph-rest-1.0&tabs=http
Background: https://learn.microsoft.com/en-us/graph/api/resources/sharepoint?view=graph-rest-1.0#sharepoint-api-root-resources

**Tasks**

* [x] Add path configuration field to form configuration (see other registration plugins for examples - and image below)
* [ ] Make sure the configured path is respected when sending Drive-items to Microsoft Graph API
* [x] Add tests for this using mocking Graph API responses
* [ ] Perform live test on Sharepoint test environment
* [x] Bonuspoints: Allow date-element placeholders in the configured path: `/sites/my-site/open-forms/{year}-{month}-{day}/` This folder then also needs to be created if it doesn't exist.

image

SilviaAmAm:
Fixes #2511


src/openforms/registrations/contrib/microsoft_graph/config.py
```diff
@@ -0,0 +1,55 @@
...
+    additional_path = serializers.CharField(
+        label=_("additional path"),
+        help_text=_(
+            "Additional path allowing date-element placeholders, for example {year}-{month}-{day}/. "
+            "It should be a relative path - i.e. it should NOT start with /"
+        ),
+        required=False,
+        validators=[is_relative],
+    )
+    drive_id = serializers.CharField(
```
SilviaAmAm, line 49: Maybe this should be a configuration of the service? :thinking:

src/openforms/contrib/microsoft/client.py
```diff
@@ -83,7 +88,9 @@ def upload_bytes(self, bytes_: bytes, remote_path: Union[str, Path]):
     def upload_stream(self, stream, stream_size: int, remote_path: Union[str, Path]):
         return self.target_folder.upload_file(
             None,
-            remote_path,
+            # upload_file says it accepts strings or Path objects
+            # (https://github.com/O365/python-o365/blob/master/O365/drive.py#L1239), but Path objects give errors
```
SilviaAmAm, line 92: But `urllib.parse.quote` only accepts str or bytes objects (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote)

SilviaAmAm:
After discussion with Joeri: make base path and additional path one single path config

src/openforms/registrations/contrib/microsoft_graph/config.py
SilviaAmAm: Discussed: can stay here

src/openforms/contrib/microsoft/client.py
```diff
@@ -83,7 +88,9 @@ def upload_bytes(self, bytes_: bytes, remote_path: Union[str, Path]):
...
+            remote_path if isinstance(remote_path, str) else str(remote_path),
```
sergei-maertens, line 93:
```suggestion
            str(remote_path),
```
if it's already a string, this is a noop

src/openforms/contrib/microsoft/client.py
```diff
@@ -47,11 +47,16 @@ class MSGraphUploadHelper:
 
     # TODO wrap upload_()-variations in single function and auto-detect object type
 
-    def __init__(self, client: MSGraphClient):
+    def __init__(self, client: MSGraphClient, options: dict):
```
sergei-maertens, line 50:
```suggestion
    def __init__(self, client: MSGraphClient, options: dict):
```
a `TypedDict` or dataclass for the options would be nice(r) for the type hinting

docs/developers/plugins/index.rst
```diff
@@ -61,3 +61,14 @@ Available plugin components
...
+==================
+
+Microsoft Graph
+---------------
+
+Microsoft graph is used in one of the core registration plugins.
+
+Microsoft graph has a web tool that lets you play with the API either with a sample account, or with your own microsoft
+account. This tool can be found `here <https://developer.microsoft.com/en-us/graph/graph-explorer>`_.
+
```
sergei-maertens, line 74: maybe it'd be nicer to set up an additional page under https://open-forms.readthedocs.io/en/latest/developers/backend/index.html#development-and-debug-tooling for External resources/tools and put it there.

When we eventually add documentation for the registration plugins, we can them simply refer to this subtitle with a `:ref:` role.

src/openforms/registrations/contrib/microsoft_graph/config.py
```diff
@@ -0,0 +1,40 @@
...
+from rest_framework import serializers
+
+from openforms.template.validators import DjangoTemplateValidator
+from openforms.utils.mixins import JsonSchemaSerializerMixin
+
+DEFAULT_FOLDER_PATH = "/open-forms/"
+
+
+def is_absolute(value: str) -> None:
+    path = Path(value)
```
sergei-maertens, line 14: using `PurePosixPath` is more appropriate here to enforce forward slashes - on Windows for example this would break since Path resolves to WindowsPath

src/openforms/registrations/contrib/microsoft_graph/plugin.py
```diff
@@ -12,18 +14,42 @@
...
+            options["folder_path"],
+            {
+                "year": "{date:%Y}".format(date=date),
+                "month": "{date:%m}".format(date=date),
+                "day": "{date:%d}".format(date=date),
+            },
+        )
+
+        return (
+            Path(folder_path)
```
sergei-maertens, line 42: this should also be `PurePosixPath` to enforce the correct slashes on different environments.

Additionally, you don't need to wrap the second/third segment in the `Path` constructor:

```py
return PurePosixPath(
    folder_path,
    slugify(submission.form.admin_name,
    submission.public_registration_reference,
)
```
should work and have the same result.

src/openforms/registrations/contrib/microsoft_graph/plugin.py
```diff
@@ -12,18 +14,42 @@
...
+            / Path(slugify(submission.form.admin_name))
+            / Path(submission.public_registration_reference)
+        )
 
-    def _get_folder_name(self, submission: Submission):
-        return f"open-forms/{slugify(submission.form.admin_name)}/{submission.public_registration_reference}"
+    def _get_filename(self, *args: Union[str, "Path"]) -> "Path":
```
sergei-maertens, line 47:
```suggestion
    def _get_filename(self, *args: str | Path) -> Path:
```

src/openforms/registrations/contrib/microsoft_graph/plugin.py
```diff
@@ -12,18 +14,42 @@
...
+        filename = Path("/")
+        for arg in args:
+            part = Path(arg) if not isinstance(arg, Path) else arg
+            filename = filename / part
```
sergei-maertens, line 51:
```suggestion
        filename = Path("/", *args)
```

`Path` takes *args too :-)

```py
>>> Path("/", "foo", "bar")
PosixPath('/foo/bar')
```

src/openforms/registrations/contrib/microsoft_graph/plugin.py
```diff
@@ -34,21 +60,24 @@ def register_submission(self, submission: Submission, options: dict) -> None:
...
-        uploader.upload_json(data, f"{folder_name}/data.json")
+        uploader.upload_json(data, self._get_filename(folder_name, "data.json"))
 
         for attachment in submission.attachments.all():
             uploader.upload_django_file(
-                attachment.content, f"{folder_name}/attachments/{attachment.file_name}"
+                attachment.content,
+                self._get_filename(
+                    folder_name, "attachments", attachment.get_display_name()
+                ),
```
sergei-maertens, line 80:
```suggestion
                folder_name / "attachments" / attachment.get_display_name()

```

src/openforms/registrations/contrib/microsoft_graph/plugin.py
```diff
@@ -70,7 +99,9 @@ def _set_payment(self, uploader, submission, folder_name):
                 content = f"{_('payment received')}: € {submission.price}"
             else:
                 content = f"{_('payment required')}: € {submission.price}"
-            uploader.upload_string(content, f"{folder_name}/payment_status.txt")
+            uploader.upload_string(
+                content, self._get_filename(folder_name, "payment_status.txt")
```
sergei-maertens, line 103:
```suggestion
                content, folder_name / "payment_status.txt"
```

sergei-maertens (changes requested):

src/openforms/registrations/contrib/microsoft_graph/plugin.py
```diff
@@ -1,29 +1,55 @@
...
 @register("microsoft-graph")
 class MSGraphRegistration(BasePlugin):
     verbose_name = _("Microsoft Graph (OneDrive/SharePoint)")
+    configuration_options = MicrosoftGraphOptionsSerializer
+
+    def _get_folder_name(
+        self, submission: Submission, options: MSGraphOptions
+    ) -> "PurePosixPath":
+        now_utc = timezone.now()
+        date = timezone.localtime(now_utc).date()
```
sergei-maertens, line 38: @SilviaAmAm I've tweaked this one because `timezone.now().date()` would return the day before if you run it between 00:00 and 01:00 at night due to timezone issues.

The correct approach is grabbing `now` in UTC, converting that to the localtime (according to `settings.TIMEZONE`) and _then_ grabbing the date portion.

sergei-maertens (approved):

Issue #2377, opened 2022-11-22
Design token: --Link-hover-color not used consistent throughout forms


LaurensBurger:
### Product versie / Product version

latest

### Omschrijf het probleem / Describe the bug

```json
    "link": {
        "color": {
            "value": "#000"
        },
        "hover": {
            "color": {
                "value": "fuchsia"
            }
        }
    }
```
--of-Link-color is (correctly) used for most hyperlinks with the exception of the progress-indicator. The hover token instead is only used in a single component of the forms the a11y-toolbar:
![image](https://user-images.githubusercontent.com/112621537/203314820-38b1fc60-0b98-4fb2-8ce1-faa2332a33bb.png)

This might be intentional but as a form designer i expect the links that originate from either the wysiwyg editor and text fields to also use the --link-hover-color token.

SilviaAmAm:
In the start page, link 1 has the correct colour and hover colour, but link 2 has only the right colour (not hover colour). 
![hover-start-page](https://user-images.githubusercontent.com/19154114/209168673-7988b431-0cdb-4f34-a16f-b73245b292cd.png)

In the step, it's the same. Both link 1 and 2 have the right colour but not the hover colour.
![hover-step](https://user-images.githubusercontent.com/19154114/209168679-5e67cc99-600a-49a6-8f67-faf47ec0590a.png)

I'm don't think that the Progress indicator steps should have the link colour / hover?  :thinking:

PR #2529, opened 2023-01-05
[#2377] More work on links colour

SilviaAmAm:
Part of #2377

Changes:
- The style dictionary does not add the "of" prefix anymore
- Migration to convert design tokens without prefix
- Replace of-link-color and of-link-hover-color with utrecht version
- Update docs
- Modified PDF style

Todo:
- figure out how to update email style => We don't style confirmation email, make issue about removing links from confirmation email.


src/openforms/config/migrations/0035_update_design_tokens.py
```diff
@@ -0,0 +1,58 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("config", "0034_alter_globalconfiguration_form_upload_default_file_types"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            add_prefix_to_design_tokens_and_replace_link_colours,
+            migrations.RunPython.noop,
```
sergei-maertens, line 56: please also implement the reverse operation - migrating forwards, then backwards and then forwards again will lead to `of.of.{tokens}` now

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #2478, opened 2022-12-14
FormIO translations for components refactor


stevenbal:
Currently, FormIO translations are stored globally on the `FormDefinition`, but to display them the FormIO builder injects these translations into the components. The issues with this are:
- we essentially have two states that contain the information
- Because the schema is immutable, any changes to the translations do not show up in other components until the `FormDefinition` is saved

It would be better if the `state.componentTranslations` were leading, and the components parse these to determine which translations are relevant. Additionally, there should be an onChange event to update this `state.componentTranslations`

Additionally, the translatable fields should be prefilled to remove the need for copy pasting

PR #2530, opened 2023-01-05
Issue/2286 add formdefinition translations to submissionstep endpoint

### LINKED ISSUE #2286: Extend API endpoint `/api/v1/submissions/:uuid/steps/:uuid` to return the `FormDefinition` translations

CharString:
Closes #2286

Discussion:

1. I'm always hesitant with changing assertions of existing tests.
`openforms.submissions.tests.test_submit_form_step.FormStepSubmissionTests` fail because they assert the complete configuration which now contains translations. That's the only reason I pop them off iff not Form.translation_enabled, instead of returning an empty object.
2. We're already changing the components in `evaluate_form_logic` for the strings that support template rendering. With a small modification we could apply translation to *all* appropriate strings. And then maybe don't need to pass the translations to the client at all because translations are all done server-side, question mark question mark.  The form builder is still a big blind spot for me. This might touch on #2478. But I think on the client, the _check_logic already triggers re-renders.

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -194,8 +194,9 @@ def evaluate_form_logic(
             new_value = updated_step_data.get(key, empty)
             original_value = initial_data.get(key, empty)
             # Reset the value of any field that may have become hidden again after evaluating the logic
-            component_empty_value = get_component_empty_value(component)
-            if original_value is not empty and original_value != component_empty_value:
+            if original_value is not empty and original_value != (
+                component_empty_value := get_component_empty_value(component)
+            ):
                 if (
                     component
                     and not is_visible_in_frontend(component, data_container.data)
--- a/src/openforms/submissions/tests/test_get_submission_step.py
+++ b/src/openforms/submissions/tests/test_get_submission_step.py
@@ -394,6 +394,32 @@ class IntegrationTests(SubmissionsMixin, APITestCase):
     Integration tests where various subsystems come together.
     """
 
+    def test_it_only_translates_appropriate_string_properties(self):
+        submission = SubmissionFactory.from_data(
+            {"bar": "Korova Milk Bar"},
+            language_code="de",
+        )
+        form_step = submission.steps[0].form_step
+        form_step.form_definition.component_translations = {
+            "de": {"bar": "Kneipe", "Bar": "Kneipe"},
+        }
+        form_step.form_definition.save()
+        self._add_submission_to_session(submission)
+        endpoint = reverse(
+            "api:submission-steps-detail",
+            kwargs={
+                "submission_uuid": submission.uuid,
+                "step_uuid": form_step.uuid,
+            },
+        )
+
+        textfield = self.client.get(endpoint).json()["formStep"]["configuration"][
+            "components"
+        ][0]
+
+        self.assertEqual(textfield["key"], "bar")
+        self.assertEqual(textfield["label"], "Kneipe")
+
     def test_dynamic_date_component_config_based_on_variables(self):
         form = FormFactory.create(
             generate_minimal_setup=True,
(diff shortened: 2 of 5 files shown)


src/openforms/submissions/form_logic.py
```diff
@@ -194,8 +194,9 @@ def evaluate_form_logic(
             new_value = updated_step_data.get(key, empty)
             original_value = initial_data.get(key, empty)
             # Reset the value of any field that may have become hidden again after evaluating the logic
-            component_empty_value = get_component_empty_value(component)
-            if original_value is not empty and original_value != component_empty_value:
+            if original_value is not empty and original_value != (
+                component_empty_value := get_component_empty_value(component)
+            ):
```
CharString, line 199: Using the walrus here saves a bunch of dict lookups and function calls iff `original_value` is empty.

src/openapi.yaml
```diff
@@ -1015,27 +1015,15 @@ paths:
   /api/v2/forms:
     get:
       operationId: forms_list
-      description: |-
```
CharString, line 1018: Why does CI create a Dutch description here? I'm getting Bavikhove vibes here: „Willen julllie de West-Vlaamse OAS alstublieft niet meer ondertitelen? ”

src/openforms/submissions/tests/test_get_submission_step.py
```diff
@@ -394,6 +394,32 @@ class IntegrationTests(SubmissionsMixin, APITestCase):
...
+                "step_uuid": form_step.uuid,
+            },
+        )
+
+        textfield = self.client.get(endpoint).json()["formStep"]["configuration"][
+            "components"
+        ][0]
+
+        self.assertEqual(textfield["key"], "bar")
+        self.assertEqual(textfield["label"], "Kneipe")
```
sergei-maertens, line 421: next time please use `from_components` instead in the factory, I had a hard time figuring out that the `Bar` label is automatically generated from the component key :)

src/openforms/submissions/tests/test_variables/test_variable_injection.py
```diff
@@ -221,3 +221,122 @@ def test_logic_evalution_endpoint(self):
...
+                "step_uuid": self.form_step.uuid,
+            },
+        )
+
+        step_json = self.client.get(endpoint).json()
+        components_by_key = {
+            c["key"]: c for c in step_json["formStep"]["configuration"]["components"]
+        }
+
+        c = components_by_key.get
```
sergei-maertens, line 291: a horrible variable/function name when in `pdb` :sweat_smile:

sergei-maertens (approved):

src/openforms/submissions/tests/test_variables/test_variable_injection.py
CharString: pip install pdbpp and never look back.

src/openforms/submissions/tests/test_get_submission_step.py
CharString: I stumbled upon it's existence by accident :sweat_smile: "Does this work? Yes, it does. :)"

PR #2531, opened 2023-01-06
Implement UI for logic rule descriptions

sergei-maertens:
Related to #2450 - depends on #2524

* Allows explicitly providing your own rule description
* Backend description generation for when it's not explicitly provided
* Supports the `_meta.description` extension to simplify nested parts

https://watch.screencastify.com/v/ry6S1ymjGOArH1IycFoi


src/openapi.yaml
```diff
@@ -2863,6 +2863,112 @@ paths:
...
+
+        This endpoint takes into account (optionally) client-provided sub-expression
+        descriptions via the ``_meta.description`` path. E.g. an input like:
+
+        ```json
+        {
+          "==": [
+            {"var": "foo"},
+            {
+                "+": [{"var": "a", "var": "b"}],
```
SilviaAmAm, line 2880: Shouldn't this be `[{"var": "a"}, {"var": "b"}]`?

src/openforms/js/components/admin/form_design/logic/LogicDescription.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const onChange = event => {
+    const {
+      target: {value},
+    } = event;
+    _onChange(event);
+    setModifiedByHuman(value !== '');
+  };
+
+  const mustBail = isEmpty(logicExpression) || !generationAllowed || modifiedByHuman || hasFocus;
```
SilviaAmAm, line 43: Maybe if the expression is invalid JSON it should also bail?

src/openapi.yaml
sergei-maertens: yes, nice catch!

src/openforms/js/components/admin/form_design/logic/LogicDescription.js
sergei-maertens: this should already the case as the `logicExpression` prop needs to be valid JSON - and it's only updated from the JSONWidget to the parent state when the JSON validates. But it's possible something broke there while I was trying to clean up the code. Is this a theoretical remark or did you see it happen? :P

SilviaAmAm: It was theoretical! :smile:

PR #2533, opened 2023-01-09
Improve UX for editing component translations

### LINKED ISSUE #2478: FormIO translations for components refactor

stevenbal:
Fixes #2478 

TODO:
- [x] implement prefilling
  - [x] prefill labels of `TRANSLATABLE_FIELDS`
    - [x] issue with datefield placeholder?
    - [x] showing datefield defaultValue makes not sense (same for numberfield and others, there should probably a mapping that determines the relevant translatable fields per component? or an override per component?)
  - [x] in case label changes, change it in the translations as well (form-creation-form.js)
    - [x] what if original label is reused in other component, but the label is changed in one of them -> original label shouldn't be changed but new one should be added
    - [x] what if two labels are different and one is changed so they become the same? -> the translation of the changed literal is used
- [x] how to deal with content field, no prefilling needed?
- [x] ensure updates to literals leads to updated translations immediately (before saving, relies on `EDIT_STEP`)
  - seems that edit_step only has the full config (not updated) and triggers way too often 
  - customDefaultValue only runs on opening the modal
  - cannot really update the step.componentTranslations state, because there is a chance the changes aren't saved
- [x] consistent ordering between languages
- [x] make literals readonly in translations? -> doesn't seem to work for datagrid
- [x] when saving, remove empty translations
- ~~implement deleting?~~
- ~~getValuesForField -> why does it return arrays?~~
- [x] general code cleanups
- [x] cleanup unused translations?
- [x] check if layout components are not skipped

**Test/edge cases**

- [x] Entering translation for a literal, then use the same literal in another field
- [x] Enter translations, edit literal -> keep the original translation
- [x] Enter translations -> edit `key` property -> keep the original translation
- [x] Add Selenium integration test to ensure translations can be entered and viewed
    - Explore if we can get drag and drop working (!)
    - Otherwise just start with a configuration and edit that

### DIFF
--- a/src/openforms/forms/forms/form_definition.py
+++ b/src/openforms/forms/forms/form_definition.py
@@ -1,5 +1,4 @@
 from django import forms
-from django.contrib.postgres.forms import JSONField
 
 from ..models import FormDefinition
 from ..validators import validate_formio_js_schema
@@ -7,7 +6,7 @@ from ..widgets import FormBuilderWidget
 
 
 class FormDefinitionForm(forms.ModelForm):
-    configuration = JSONField(
+    configuration = forms.JSONField(
         widget=FormBuilderWidget, validators=[validate_formio_js_schema]
     )
 
@@ -21,5 +20,11 @@ class FormDefinitionForm(forms.ModelForm):
             "login_required",
             "is_reusable",
             "configuration",
+            "component_translations",
         )
-        widgets = {"uuid": forms.TextInput(attrs={"readonly": True})}
+        widgets = {
+            "uuid": forms.TextInput(attrs={"readonly": True}),
+            "component_translations": forms.HiddenInput(
+                attrs={"class": "form-builder__component-translations"}
+            ),
+        }
--- a/src/openforms/forms/tests/selenium_tests/test_form_designer.py
+++ b/src/openforms/forms/tests/selenium_tests/test_form_designer.py
@@ -1,0 +1,257 @@
+import time
+from contextlib import contextmanager
+
+from django.test import override_settings
+from django.urls import reverse
+
+from furl import furl
+from selenium.common import NoSuchElementException
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.wait import WebDriverWait
+
+from openforms.accounts.tests.factories import SuperUserFactory
+from openforms.tests.selenium_tests.base import SeleniumTestCase
+
+from ..factories import FormFactory
+
+
+@contextmanager
+def phase(desc: str):
+    yield
+
+
+class FormDesignerComponentTranslationTests(SeleniumTestCase):
+    @classmethod
+    def getWebdriverOptions(cls):
+        options = super().getWebdriverOptions()
+        options.add_argument("--lang=en-US")
+        return options
+
+    def setUp(self):
+        super().setUp()
+
+        SuperUserFactory.create(username="admin", password="selenium")
+
+        # log the super user in to the admin
+        login_url = furl(self.live_server_url) / reverse("admin:login")
+        self.selenium.get(str(login_url))
+        username_input = self.selenium.find_element(By.ID, "id_username")
+        username_input.send_keys("admin")
+        password_input = self.selenium.find_element(By.ID, "id_password")
+        password_input.send_keys("selenium")
+        self.selenium.find_element(
+            By.XPATH, '//input[@type="submit"][not(@hidden)]'
+        ).click()
+
+    @override_settings(LANGUAGE_CODE="en")
+    def test_editing_translatable_properties(self):
+        # set up a form
+        form = FormFactory.create(
+            name="Selenium test",
+            name_nl="Selenium test",
+            generate_minimal_setup=True,
+            formstep__form_definition__name_nl="Selenium test",
+            formstep__form_definition__configuration={
+                "components": [
+                    {
+                        "type": "textfield",
+                        "key": "field1",
+                        "label": "Field 1",
+                        "description": "Description 1",
+                    },
+                    {
+                        "type": "select",
+                        "key": "field2",
+                        "label": "Field 2",
+                        "description": "Description 2",
+                        "data": {
+                            "values": [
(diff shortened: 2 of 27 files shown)


src/openforms/forms/forms/form_definition.py
```diff
@@ -21,5 +21,6 @@ class Meta:
             "login_required",
             "is_reusable",
             "configuration",
+            "component_translations",
```
sergei-maertens, line 24: TODO -> this isn't going right now in the admin

src/openforms/js/components/form/translations.js
```diff
@@ -0,0 +1,17 @@
+import {Formio} from 'react-formio';
+
+import {DEFAULT_TEXT_TABS} from './edit/tabs';
+
+class TranslationsDataGrid extends Formio.Components.components.datagrid {
```
sergei-maertens, line 5: TODO: check if we can't enforce this behaviour via options

src/openforms/js/components/formio_builder/index.js
```diff
@@ -14,6 +14,7 @@ onLoaded(() => {
   [...FORM_BUILDERS].forEach(node => {
     const configurationInput = BEM.getChildBEMNode(node, BLOCK_FORM_BUILDER, INPUT_ELEMENT);
     const configuration = JSON.parse(configurationInput.value) || {display: 'form'};
+    // FIXME: handle translations?
```
sergei-maertens, line 17: TODO: should be possible to manage component translations from standalone admin page too

src/openforms/js/formio_module.js
```diff
@@ -60,6 +61,7 @@ const FormIOModule = {
     currency: CurrencyField,
     editgrid: EditGrid,
     datamap: Datamap,
+    translationsDataGrid: TranslationsDataGrid,
```
sergei-maertens, line 64: TODO: check if we can avoid this

sergei-maertens (changes requested):
Selenium tests are flaky still - old translations may show up

sergei-maertens:
Ok, all that's left to do is cleaning up the commit history if CI passes

sergei-maertens:
image

sergei-maertens (approved):
(just approving to get rid of the 'changes requested' status)

src/openforms/js/components/formio_builder/translation.js
```diff
@@ -1,3 +1,254 @@
...
+  fieldset: ['legend'],
+  editgrid: ['groupLabel'],
+  // manually entered data values -> labels are translatable
+  select: [/data\.values\[[0-9]+\]\.label/],
+  selectboxes: [/values\[[0-9]+\]\.label/],
+  radio: [/values\[[0-9]+\]\.label/],
+};
+
+const isRegex = item => {
+  return typeof item.test === 'function';
```
CharString, line 47: I wouldn't be surprised if String were function too in javascript xD

src/openforms/js/components/formio_builder/translation.js
```diff
@@ -1,3 +1,254 @@
...
+ * of a particular component.
+ *
+ * @param  {Object} container Key-value mapping, key is language code, values are mappings
+ *                            of literals to their translations *for that language code*.
+ * @param  {Object} component Formio component/schema - it contains the configuration
+ *                            of one particular component in the form builder.
+ * @return {Void}             The container is mutated directly. It is advised to pass
+ *                            in an Immer draft or carefully manage mutable objects.
+ */
+export const persistComponentTranslations = (container, component) => {
```
CharString, line 97: Reading `persistComponentTranslations` elsewhere without context, made me think persistence to the backend. While it's actually moving data from the passed in component to the container... Can't really think of a better name :thinking:

src/openforms/js/components/formio_builder/translation.spec.js
```diff
@@ -0,0 +1,199 @@
...
+    expect(isTranslatableProperty('textfield', 'defaultValue')).toBe(true);
+    expect(isTranslatableProperty('date', 'defaultValue')).toBe(false);
+  });
+
+  it('works with nested paths', () => {
+    expect(isTranslatableProperty('textfield', 'data.values[3].label')).toBe(false);
+    expect(isTranslatableProperty('select', 'data.values[2].value')).toBe(false);
+    expect(isTranslatableProperty('select', 'data.values[1].label')).toBe(true);
+    expect(isTranslatableProperty('radio', 'values[420].label')).toBe(true);
+    expect(isTranslatableProperty('selectboxes', 'values[420].label')).toBe(true);
```
CharString, line 28: Yep. What were they smoking?

src/openforms/scss/components/builder/_builder.scss
```diff
@@ -59,6 +59,9 @@ div.flatpickr-calendar.open {
   .tab-pane.active {
     min-height: 50vh;
   }
+  .formio-component-content .tab-pane.active {
+    min-height: 0;
```
CharString, line 63: I don't even want to know...

src/openforms/forms/tests/selenium_tests/test_form_designer.py
```diff
@@ -0,0 +1,257 @@
...
+
+from furl import furl
+from selenium.common import NoSuchElementException
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.wait import WebDriverWait
+
+from openforms.accounts.tests.factories import SuperUserFactory
+from openforms.tests.selenium_tests.base import SeleniumTestCase
```
CharString, line 15: `base` is a first in the codebase. But "utils" feels off too... Maybe move it to __init__.py. Meh. Base it is...

PR #2535, opened 2023-01-10
[#2143] add option to exclude confirmation page content from PDF

### LINKED ISSUE #2143: (Offer the option to) not include the text from confirmation page in PDF
### Thema / Theme

Other

We'd like to include a text on the confirmation page to highlight the option of downloading the filled form as a PDF. However when this instruction is included in the PDF itself, it isn't appropriate.

On the confirmation page:
image

As of now this text ends up in the PDF as well, and this cannot be set/changed:
image

_No response_

_No response_

pi-sigma:
Fixes #2143 by adding a checkbox to the admin to include or exclude the text from the form confirmation page in the PDF report.


src/openforms/js/components/admin/form_design/Confirmation.js
```diff
@@ -42,6 +42,7 @@ const EMAIL_OPTIONS = [
 
 const Confirmation = ({
   displayMainWebsiteLink = false,
+  includeConfirmationPageContentInPdf: includeConfirmationPageContentInPdf = false,
```
sergei-maertens, line 45:
```suggestion
  includeConfirmationPageContentInPdf = false,
```
not sure why you are aliasing the prop to itself

sergei-maertens (changes requested):
Can you also run `npm run makemessages && npm run compilemessages` so that the frontend translations are baked in and don't give console warnings?

sergei-maertens (approved):

PR #2540, opened 2023-01-11
[#2534] Custom error messages for form components

### LINKED ISSUE #2534: Make the error message customisable for each form field
### Thema / Theme

Frontend

DH Taiga issue 66 / User story 360

Currently, the error message that is displayed to the user when the input fails validation is hardcoded.

It would be good to make these error messages configurable.

In Formio, it is possible to specify a custom error for a form field by adding:

```
"errors": {
    "required": "{{ field }} is required. GNA.",
    "maxLength": "{{ field }} is too long. GNA."
}
```

To the component JSON.

To do:
- Make drop down with available errors for each component and then a field with the custom message
- Figure out how to deal with translations?

_No response_

_No response_

SilviaAmAm:
Fixes #2534 
image


src/openforms/formio/typing.py
```diff
@@ -49,3 +49,5 @@ class Component(TypedDict):
     hidden: bool
     defaultValue: JSONValue
     prefill: PrefillConfiguration
+    errors: JSONObject | None
+    translatedErrors: JSONObject | None
```
sergei-maertens, line 53: this will cause many type "errors" in other places because there's no way (yet, Python 3.11 has it) to specify that keys are optional. So any existing tests without these component keys will  show type errors in IDE's/editors.

src/openforms/submissions/api/serializers.py
```diff
@@ -246,6 +247,12 @@ def to_representation(self, instance):
             instance.submission.data,
             **self.context,
         )
+
+        # Add to each component the custom errors in the current locale
+        new_configuration = get_translated_custom_error_messages(
+            new_configuration, instance.submission
+        )
```
sergei-maertens, line 254: wouldn't it make more sense to hook this into `openforms.formio.service.get_dynamic_configuration`?

This sets up the formio configuration wrapper once (which avoids many `iter_components` calls for performance reasons) and allows you to build on that datastructure. `get_dynamic_configuration` is called by `evaluate_form_logic`.

sergei-maertens (commented):
General approach looks good, couple thoughts added as comments!

sergei-maertens (approved):

PR #2544, opened 2023-01-11
Form designer wysiwyg translations

### LINKED ISSUE #2475: Form designer FormIO translations UI: WYSIWYG content field translations

CharString:
Closes #2475


src/openforms/js/utils/translation.js
```diff
@@ -1,4 +1,8 @@
+import FormioUtils from 'formiojs/utils';
```
CharString, line 1: The form
```javascript
import {Utils as FormioUtils} from 'formiojs';
```

trips jest with `SyntaxError: Unexpected token 'export'`. This phrasing doesn't. [Other workarounds exist](https://www.sobyte.net/post/2022-06/jest/)

src/openforms/js/components/form/content.js
```diff
@@ -26,19 +26,80 @@ const CONTENT_PRESENTATION = {
...
+                  type: 'tabs',
+                  components: LANGUAGES.map(([languageCode, _label]) => {
+                    return {
+                      key: languageCode,
+                      label: languageCode.toUpperCase(),
+                      components: [
+                        {
+                          label: 'Literal',
+                          hideLabel: true,
+                          key: `of-translations.${languageCode}[0].literal`,
```
CharString, line 73: This magnificent key makes it such that it looks like the datagrid used else where and gets picked up into the componentTranslations on PUT and POST.

src/openforms/js/components/form/content.js
```diff
@@ -26,19 +26,80 @@ const CONTENT_PRESENTATION = {
...
+                          input: true,
+                          unique: true,
+                          type: 'hidden',
+                          customDefaultValue: ({instance, data}) =>
+                            instance.options.editComponent.html,
+                        },
+                        {
+                          label: 'Translation',
+                          hideLabel: true,
+                          key: `of-translations.${languageCode}[0].translation`,
```
CharString, line 83: The beauty grows on you. :wink:

src/openforms/js/components/form/content.js
```diff
@@ -26,19 +26,80 @@ const CONTENT_PRESENTATION = {
...
-      rows: 3,
-      tooltip: 'The HTML template for the result data items.',
+      type: 'columns',
+      key: 'content',
+      columns: [
+        {
+          components: [
+            {
+              type: 'tabs',
+              components: [{label: 'Content', components: [CONTENT_EDITOR]}],
```
CharString, line 52: Single tab. The easiest way to make it align and look similar side by side to the translations.

stevenbal (approved):

PR #2545, opened 2023-01-12
:sparkles: [#2519] -- skip upgrade checks in some cases

### LINKED ISSUE #2519: Deploying a version post 2.0 should not execute the 1.1 -> 2.0 pre-checks
*Writing this ticket without 100% understanding what went wrong*

Real life case is where someone imported a 1.x form in Open Forms 2.0.0 Nothing went visibly wrong (except for the warning in the designer).

Migrating from 2.0.0 to 2.0.1 triggered the pre-checks and caused a failure (I must admit I don't see why this would cause data issues but maybe @sjoerdie can elaborate). This was unexpected because Open Forms was already on 2.0.x

The proper solution is to prevent imports/version rollbacks that cause this issue in the first place, and remove the pre-check.

sergei-maertens:
Closes #2519 

* If the current version matches a loose match to the target version, it is implied you already passed those checks before and we don't need to run them again.

### DIFF
--- a/src/openforms/upgrades/tests/test_upgrade_paths.py
+++ b/src/openforms/upgrades/tests/test_upgrade_paths.py
@@ -81,6 +81,26 @@ class UpgradePathTests(TestCase):
             with self.assertRaises(VersionParseError):
                 check_upgrade_path("0b5b0c04ff322bf098872081319decfe558c9f73", "2.0")
 
+    def test_already_on_target(self):
+        """
+        Assert that the checks are skipped if you are already on a version that matches
+        the target version (2.0.1 matches ~2.0).
+        """
+        UPGRADE_PATHS = {
+            "2.0": UpgradeConstraint(
+                valid_ranges={
+                    VersionRange(minimum="1.1"),
+                },
+                # throws exception if checks _are_ being executed
+                scripts=["i-do-not-exist"],
+            ),
+        }
+        with mock_upgrade_paths(UPGRADE_PATHS):
+            with self.subTest(from_version="2.0.1", to="2.0.2"):
+                self.assertTrue(check_upgrade_path("2.0.1", "2.0.2"))
+            with self.subTest(from_version="2.0.1", to="2.1.2"):
+                self.assertTrue(check_upgrade_path("2.0.1", "2.1.2"))
+
 
 class DjangoScriptTests(SimpleTestCase):
     def test_check_management_commands(self):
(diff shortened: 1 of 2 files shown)


SilviaAmAm (approved):

joeribekker (commented):
Will upgrading from 1.x to anything greater than 2.0 (like 2.1, 3.0, etc.) still trigger the constraints, or is it only when upgrading to 2.0.x?

sergei-maertens:
> Will upgrading from 1.x to anything greater than 2.0 (like 2.1, 3.0, etc.) still trigger the constraints, or is it only when upgrading to 2.0.x?

I'm taking the example case of upgrading from 1.1.9 to 2.1.0 and will break down what happens in the checks.

Since we are targetting 2.1.0, the upgrade path to go to 2.1 is encountered:

```
    "2.1": UpgradeConstraint(
        valid_ranges={
            VersionRange(minimum="2.0.0"),
        }
    ),
```

This states that the application must currently be on 2.0.0 _at the least_. The result is that the upgrade is blocked (i.e. going straight to 2.1.0 from 1.1.9 is not supported).

So, in this case, you change your deployment strategy to go in multiple steps: 1.1.9 -> 2.0.y, and then from 2.0.y -> 2.1.0.

Now, when upgrading from 1.1.9 to 2.0.2 (latest patch release), we encounter the relevant upgrade constraint:

```
    "2.0": UpgradeConstraint(
        valid_ranges={
            VersionRange(minimum="1.0.14", maximum="1.1.5"),
            VersionRange(minimum="1.1.6", maximum="1.2.0"),
            VersionRange(minimum="2.0.0-beta.0"),
        },
        management_commands=[
            "check_duplicate_component_keys",
            "check_invalid_field_keys",
        ],
        scripts=[
            "detect_advanced_logic_usage",
        ]
```

The version range 1.1.6-1.2.0 satisfies this. Then the management command and script checks are performed. These pass (or they don't, and you fix the underlying data first).

Now you upgrade to 2.0.2.

The next step is to upgrade 2.0.2 to 2.1.0, so we again encounter the relevant constraint:

```
    "2.1": UpgradeConstraint(
        valid_ranges={
            VersionRange(minimum="2.0.0"),
        }
    ),
```

The version range is satisfied (2.0.2 >= 2.0.0). The 2.0 upgrade checks are **not** executed anymore, as they are not specified in the 2.1 upgrade constraint.

Does this make the flow a bit more clear?

joeribekker:
> This states that the application must currently be on 2.0.0 at the least. The result is that the upgrade is blocked (i.e. going straight to 2.1.0 from 1.1.9 is not supported).

Check!

sergei-maertens:
Backported in 49d4d9491fbd3911b3f8c4e34dc70ebc1e3ae76e

joeribekker:
<3

PR #2547, opened 2023-01-12
[#2396] fix auto_login_backend bug

### LINKED ISSUE #2396: removing "authentication_backends" from a form does not remove "auto_login_authentication_backend" - breaks "saving"

2.0.0 / 2.1.0

When a authentication method is removed the auto_login value remains in the configuration.
image
 
This causes a warning which cannot be solved:
image

After this it is impossible to edit the form in any way, changes are not stored until a authentication method is added again. 

**Tasks**

* [ ] Make sure the "auto_login" value is removed when the corresponding authentication method is disabled.

pi-sigma:
Fixes #2396

One of the API form tests (`test_deleting_a_confirmation_email_template_through_the_api`) had to be modified to avoid breaking as a result of the code change.

### DIFF
--- a/src/openforms/api/utils.py
+++ b/src/openforms/api/utils.py
@@ -23,9 +23,8 @@ def underscore_to_camel(input_: Union[str, int]) -> str:
 def get_from_serializer_data_or_instance(
     field: str, data: dict, serializer: Serializer
 ) -> Any:
-    data_value = data.get(field)
-    if data_value:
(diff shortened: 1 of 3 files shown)


sergei-maertens (approved):

sergei-maertens:
Backport commit: 79267f6963f6de77753a30fae43aee09d7d602ad

Issue #1898, opened 2022-08-22
Refactor DjangoTemplateValidator to use isolated backends/engines


sergei-maertens:
With #1869 we are gaining the ability to use "sandboxed" (to certain degrees) instances of the django template engine. The validator in `openforms.utils.validators.DjangoTemplateValidator` currently uses the "default" django template backend which has all template tags (even private) enabled and accessible. The validator should receive an argument to use a particular backend instance for validation to enforce sandbox behaviour.

PR #2560, opened 2023-01-13
Replace Template(...).render(...) usage with openforms.template.render_from_string

### LINKED ISSUE #1904: Refactor template-from-string rendering
With #1869 we are gaining the ability to use "sandboxed" (to certain degrees) instances of the django template engine.

Currently, we render a number of templates for emails (subjects, confirmation, registration...), relying on `Template(<string>).render(<context)`. This is "old" interface from when Django did not have pluggable template backends.

This can be refactored to use an appropriate template backend in `openforms.template`, which properly sets up the builtins and blocks other libraries from being used/accessed (sandboxing) for this user-contributed template source code.

Possibly we also need a script/management command to check for existing templates in environments that do not meet the restricted backend configuration so these can be fixed.

Related to #1898

sergei-maertens:
Closes #1904

This still uses the Django template engine, but instead uses sandboxes that don't allow using non-public custom tags and/or include files from our source code.

### DIFF
--- a/src/openforms/authentication/contrib/tests/saml_utils.py
+++ b/src/openforms/authentication/contrib/tests/saml_utils.py
@@ -1,11 +1,12 @@
 from base64 import b64encode
 from hashlib import sha1
+from pathlib import Path
 from typing import Optional
-
-from django.template import Context, Template
 
 from digid_eherkenning.models import EherkenningConfiguration
 from onelogin.saml2.utils import OneLogin_Saml2_Utils
+
+from openforms.template import render_from_string
 
 
 def create_test_artifact(service_entity_id: str = "") -> str:
@@ -18,11 +19,8 @@ def create_test_artifact(service_entity_id: str = "") -> str:
 
 
 def get_artifact_response(filepath: str, context: Optional[dict] = None) -> bytes:
-    with open(filepath, "r") as template_source_file:
-        template = Template(template_source_file.read())
-
-    rendered = template.render(Context(context or {}))
-    return rendered.encode("utf-8")
+    template_source = Path(filepath).read_text()
+    return render_from_string(template_source, context or {}).encode("utf-8")
 
 
 def get_encrypted_attribute(attr: str, identifier: str):
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -293,11 +293,11 @@ TEMPLATES = [
             ],
             "loaders": TEMPLATE_LOADERS,
             "builtins": [
+                # TODO: these can be deleted once the validator is done via backend.
                 "openforms.emails.templatetags.appointments",
                 "openforms.emails.templatetags.form_summary",
                 "openforms.emails.templatetags.payment",
                 "openforms.emails.templatetags.products",
-                "openforms.config.templatetags.privacy_policy",
             ],
         },
     },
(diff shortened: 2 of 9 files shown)


src/openforms/conf/base.py
```diff
@@ -293,11 +293,11 @@
             ],
             "loaders": TEMPLATE_LOADERS,
             "builtins": [
+                # TODO: these can be deleted once the validator is done via backend.
```
sergei-maertens, line 296: #1898

src/openforms/config/templatetags/privacy_policy.py
```diff
@@ -7,9 +7,10 @@
 register = template.Library()
 
 
-@register.simple_tag()
-def privacy_policy():
-    conf = GlobalConfiguration.get_solo()
+@register.simple_tag(takes_context=True)
+def privacy_policy(context):
+
+    conf = context.get("global_configuration") or GlobalConfiguration.get_solo()
```
sergei-maertens, line 13: small optimization where we can avoid a cache hit and potentially a DB hit

src/openforms/template/backends/sandboxed_django.py
```diff
@@ -47,3 +50,27 @@ def get_templatetag_libraries(self, custom_libraries: dict) -> dict:
...
+                    "openforms.emails.templatetags.products",
+                    "openforms.config.templatetags.privacy_policy",
+                ],
+            }
+        }
+    )
+
+
+openforms_backend = cast(
+    SandboxedDjangoTemplates, SimpleLazyObject(get_openforms_backend)
```
sergei-maertens, line 72: need to make it lazy because otherwise we get a bunch of circular imports (models.py files import the backend). Apparently instantiating the backend completely tries to load all the builtins.

I think this builtins situation will see some refactoring in the future anyway.

src/openforms/config/templatetags/privacy_policy.py
CharString: My knee-jerk reaction to this circular way of rendering  a link in our own property was why? But setting the correct `rel` attribute on the anchor is *that* important.  :+1:  And I guess anchors aren't allowed in the sandbox?

CharString (commented):
Looks fine, just have the one question.

src/openforms/authentication/contrib/tests/saml_utils.py
```diff
@@ -18,11 +19,8 @@ def create_test_artifact(service_entity_id: str = "") -> str:
 
 
 def get_artifact_response(filepath: str, context: Optional[dict] = None) -> bytes:
-    with open(filepath, "r") as template_source_file:
-        template = Template(template_source_file.read())
-
-    rendered = template.render(Context(context or {}))
-    return rendered.encode("utf-8")
+    template_source = Path(filepath).read_text()
+    return render_from_string(template_source, context or {}).encode("utf-8")
```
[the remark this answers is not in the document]
sergei-maertens: it's not needed in this case, but it's also overkill to set up yet another backend for a test utility :)

The point of making this change is more to prevent people from copy-paste more `Template(...).render(Context(...))` code because they came across it, instead I want people to only find `render_from_string` and use that API.

So this is mostly a cosmetic change.

src/openforms/template/backends/sandboxed_django.py
CharString: Casting this *is* a lie though... It serves for IDE auto-complete purposes. But it breaks type safety. It **should** be a `LazyObject[SandboxedDjangoTemplates]`, but the django annotations are outside of this projects scope.

PR #2561, opened 2023-01-13
Refactor template validators

### LINKED ISSUE #1898: Refactor DjangoTemplateValidator to use isolated backends/engines

sergei-maertens:
Closes #1898 

All the model-level `DjangoTemplateValidator` instances now specify a backend (or use the most-restrictive one by default).

These validators are used for user-submitted Django templates which are by definition untrusted input.


src/openforms/registrations/contrib/email/templates/emails/email_registration.txt
```diff
@@ -1,4 +1,4 @@
-{% load i18n %}{% with form_name=renderer.submission.form.admin_name %}{% autoescape off %}{% if payment_received %}{% blocktrans trimmed %}
+{% load i18n form_summary %}{% with form_name=renderer.submission.form.admin_name %}{% autoescape off %}{% if payment_received %}{% blocktrans trimmed %}
```
CharString, line 1: We dropped more builtins. Was this the form_summary the only one  that tripped a test? Or did we cover them already in the other PR?

src/openforms/emails/models.py
```diff
@@ -38,7 +38,8 @@ class ConfirmationEmailTemplate(models.Model):
                 required_template_tags=[
                     "appointment_information",
                     "payment_information",
-                ]
+                ],
+                backend="openforms.template.openforms_backend",
```
CharString, line 42: ... I feel these validators are a kind of typeguard, and it doesn't sit quite right that on use of the value, we could specify a different backend.

CharString (changes requested):
Needs some clarification in the test. Consider the rest as conversation.

src/openforms/emails/models.py
sergei-maertens: They are here to prevent badly formatted (syntax errors) in templates submitted by form builders to end up in the system which then generates runtime errors and clutters error monitoring.

It's not a bug we (developers) can or should fix, but it is up to us to provide the tooling/feedback to users so they can correct their own mistakes. It's definitely going to cause problems if we render with a stricter subset of the language than that we validate with.

src/openforms/registrations/contrib/email/templates/emails/email_registration.txt
sergei-maertens: we only allow the builtins in sandbox templates because it's hard to document and understand for form builders that they have to `{% load %}` a library before they can use certain tags. That also limits our options to refactor/move these custom tags around.

These templates however are only under our control as developers, so we can use the 'normal' Django template operation and make things more explicit.

src/openforms/template/tests/test_validators.py
```diff
@@ -1,13 +1,18 @@
 from django.core.exceptions import ValidationError
+from django.template import engines
+from django.template.backends.django import DjangoTemplates
 from django.test import SimpleTestCase, override_settings
 from django.utils.translation import gettext as _
 
 from ..validators import DjangoTemplateValidator
 
+django_backend = engines.all()[0]
+assert isinstance(django_backend, DjangoTemplates)
```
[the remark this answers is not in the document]
sergei-maertens: we are testing that the `DjangoTemplateValidator` works correctly in the "generic" sense, without testing any particular backend.

The `{% csrf_token %}` and `{% load i18n %}` template constructs are core django tokens that will not break if we have a bug in our own tags, or won't be affected by unrelated changes (e.g. changing the `{% form_summary %}` tag syntax would break these tests but not because the validator broke).

Before this patch, the `DjangoTemplateValidator` relied on `Template`, which under the hood uses `Engine.get_default()`, which just grabs the first Django templates enginge/configuration from `settings.TEMPLATES`. However, this patch defaults to using our own sandbox as backend for validation (secure defaults), which doesn't allow usage of these built-in django tags.

So, two options to fix these tests:

* replace the tags in there with sandbox-allowed tags
* force the old validator behaviour by using the backend from `Engine.get_default()`

I opted for the second so that the actual template doesn't matter/won't cause issues while still testing the correct validator behaviour. Obtaining this django template instance is a bit convoluted and there's a safe guard in case at some point our first entry in `settings.TEMPLATES` is no longer DTL (but Jinja2 for example). Unlikely to happen, but just in case, this can save some debugging time.

> If not, I'd import the one we expect and assert that it's the correct one

that's essentially what I'm doing - instantiating without duplicating all the settings from settings.TEMPLATES manually is... surprisingly hard and fragile if/when django upgrades change template internals.

The `APP_DIRS` flip is unrelated to this, that's another cleanup of something that used to be required in the past, but what Django now does automatically.

src/openforms/template/validators.py
```diff
@@ -45,10 +53,8 @@ def check_required_tags(self, value):
                 )
 
     def check_syntax_errors(self, value):
-        # code lifted from maykinmedia/mail-editor
-        # https://github.com/maykinmedia/mail-editor/blob/e9ea1762af9a5c7ec0826876cb647dea444b95ba/mail_editor/mail_template.py#L28
-
+        backend = import_string(self.backend)
```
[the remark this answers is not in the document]
sergei-maertens: this is the only place and it's indeed required to the migrations deconstruction. I attempted to add a custom serializer for this so we could pass the instance around everywhere, but that got very ugly very quickly.

`import_string` relies on `sys.modules` cache anyway so I don't foresee any performance regressions here

src/openforms/config/models.py
```diff
@@ -100,7 +101,7 @@ class GlobalConfiguration(SingletonModel):
         help_text=_("Content of the save form email message."),
         default=partial(_render, "emails/save_form/content.html"),
         validators=[
-            DjangoTemplateValidator(),
+            DjangoTemplateValidator(backend="openforms.template.openforms_backend"),
             URLSanitationValidator(),
         ],
     )
```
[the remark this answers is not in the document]
CharString: I agree on the creep, referencing  open-formulieren/security-issues#4 for the future.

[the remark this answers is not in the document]
sergei-maertens: Okay, I also created #2575 for this just now :)

src/openforms/template/tests/test_validators.py
```diff
@@ -1,13 +1,20 @@
...
+# is/should be the regular vanilla Django templates backend.
+#
+# If this assertion fails, it'll most likely be because of changing the template engine
+# to Jinja2 or something else - check your TEMPLATES setting and adapt the index below
+# as needed.
 django_backend = engines.all()[0]
-assert isinstance(django_backend, DjangoTemplates)
+assert (
+    type(django_backend).__name__ == "DjangoTemplates"
+), "Expected the backend to be a plain (non-sandboxed) Django templates backend."
```
CharString, line 17: :gem: crystal clear :man_cook: :kiss:

CharString (approved):

CharString:
I'm sure you'll  fix the migration collision... Re-ask for a review if you feel unsure.

sergei-maertens:
> I'm sure you'll fix the migration collision... Re-ask for a review if you feel unsure.

I shall not let my insecurities affect my work ;)

CharString (approved):
If nobody minds 2 migrations with the prefix '0036_', I don't mind.

PR #2563, opened 2023-01-16
[#2557] Introduce datetime component and compare with var "now"

### LINKED ISSUE #2557: Compare date and time with variable 'now' + some delta
### Thema / Theme

Form designer

DH Taiga issue 359, user story 361

At the moment, there is no way to have a date and a time in a form and combine them to compare them to a datetime.

For example, in DH they have a form to report road works. In this form there are a date and a time field to specify the start of the road works.
In case that the combination date + time are more than 24h in the future, some logic is applied.

To do:
- Figure out how to compare the date + the time fields with a variable
- Figure out if its possible to specify a delta for a datetime in the jsonlogic package.

_No response_

_No response_

SilviaAmAm:
Fixes #2557 

Front end: https://github.com/open-formulieren/open-forms-sdk/pull/343
Json logic: https://github.com/maykinmedia/json-logic-py/pull/13

We possibly will need to make a user friendly version of the logic rules to compare datetimes.


src/openforms/utils/json_logic/descriptions.py
```diff
@@ -323,15 +323,45 @@ def op_today(*args, **kwargs) -> str:
...
+            return gettext(
+                "{years}y, {months}m, {days}d, {hours}h, {minutes}min, {seconds}s"
+            ).format(
+                years=args[0],
+                months=args[1],
+                days=args[2],
+                hours=args[3],
+                minutes=args[4],
+                seconds=args[5],
+            )
```
CharString, line 367: Matching can unpack too:
```suggestion
    match args:
        case [years]:
            return gettext("{years} year(s)").format(**locals())
        case [years, months]:
            return gettext("{years} year(s), {months} month(s)").format(**locals())
        case [years, months, days]:
            return gettext("{years} year(s), {months} month(s), {days} day(s)").format(**locals())
        case [years, months, days, hours]:
            return gettext("{years}y, {months}m, {days}d, {hours}h").format(**locals())
        case [years, months, days, hours, minutes]:
            return gettext(
                "{years}y, {months}m, {days}d, {hours}h, {minutes}min"
            ).format(**locals())
        case [years, months, days, hours, minutes, seconds]:
            return gettext(
                "{years}y, {months}m, {days}d, {hours}h, {minutes}min, {seconds}s"
            ).format(**locals())
        case _:
            raise ValueError("Unexpected amount of arguments. Expected 1-6 got {len(args)}: {args!r}")

```
Should work... 

The error is really needed. If I had to write jsonlogic, I'd want all the help I could get from a compiler.

CharString (changes requested):

CharString:
Should we document the behaviour around daylight savings time changes? I look forward to dismissing DST pain issues with a "won't fix", ask our politicians to stop with this DST nonsense.

SilviaAmAm:
@CharString which behaviour do you mean? 

Thinking about it, maybe I should have included the timezone in the formatter? :thinking:

CharString:
> @CharString which behaviour do you mean?

For example adding a delta > 48h to a datetime to a datetime on a Friday before the switch. Not all weekends have 48 hours. Every year there's one with 47 and one with 49.

Or more in our user's terms: 6 weeks is a common delta in Dutch administrative law. Adding a delta of 6*7 days to 17:00 on a day within 6 weeks of a switch will end up on 16:00 or 18:00 on the target day in the timezone that is in effect then.

> Thinking about it, maybe I should have included the timezone in the formatter? thinking

Well... If we format the time in the same timezone as the start datetime... The above will not become apparent. But on the resulting datetime, the timezone has switched. In the local case between CET and CEST.

But... The tests use fixed offset timezones and frozen now()... So CI won't break, so it's not our problem.
Funny, [the guy that caused all this nonsense](https://en.wikipedia.org/wiki/William_Willett) did so because he liked [having some lit time playing golf](uhttps://archive.org/details/seizedaylightcur0000prer)... Golf with Friends anyone? :eyes:

CharString (approved):

sergei-maertens:
> > @CharString which behaviour do you mean?
> 
> For example adding a delta > 48h to a datetime to a datetime on a Friday before the switch. Not all weekends have 48 hours. Every year there's one with 47 and one with 49.
> 
> Or more in our user's terms: 6 weeks is a common delta in Dutch administrative law. Adding a delta of 6*7 days to 17:00 on a day within 6 weeks of a switch will end up on 16:00 or 18:00 on the target day in the timezone that is in effect then.
> 
> > Thinking about it, maybe I should have included the timezone in the formatter? thinking
> 
> Well... If we format the time in the same timezone as the start datetime... The above will not become apparent. But on the resulting datetime, the timezone has switched. In the local case between CET and CEST.
> 
> But... The tests use fixed offset timezones and frozen now()... So CI won't break, so it's not our problem. Funny, [the guy that caused all this nonsense](https://en.wikipedia.org/wiki/William_Willett) did so because he liked having some lit time playing golf... Golf with Friends anyone? eyes

Using your days/weeks examples - maybe the result should be rounded to a similar precision? if we're talking days/weeks -> don't bother comparing hours/minutes/seconds, but resolve to the day/date instead.

This depends a lot on internal processes though - is the exact time relevant, or are they only working in times of days? -> let's check with the business before making the code more complicated :)

src/openforms/forms/api/validators.py
```diff
@@ -100,10 +100,10 @@ def _is_advanced_logic(
                 return True
         return False
 
-    def _is_date_operand(self, operand: Any) -> bool:
+    def _is_date_or_datetime_operand(self, operand: Any) -> bool:
         return (
             isinstance(operand, Operation)
-            and operand.operator == "date"
+            and (operand.operator == "date" or operand.operator == "datetime")
```
sergei-maertens, line 106: neater would be `operand.operator in ["date", "datetime"]`, but good enough for now

sergei-maertens (approved):

PR #2572, opened 2023-01-17
Truncate logic descriptions that are too long

### LINKED ISSUE #2569: Logic description too long - cannot save rule

latest

Error: 
"name": "0.description",
"code": "max_length",
"reason": "Zorg ervoor dat dit veld niet meer dan 100 karakters bevat."

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Fixes #2569

* Limit the user input to max 100 chars
* Automatically truncate server-side generated description to 100 chars

### DIFF
--- a/src/openforms/js/components/admin/form_design/FormLogic.js
+++ b/src/openforms/js/components/admin/form_design/FormLogic.js
@@ -142,13 +142,18 @@ const Rule = ({
   }
 
   const onDescriptionGenerated = generatedDescription => {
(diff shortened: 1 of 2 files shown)


SilviaAmAm (changes requested):
This prevents the auto generated descriptions from being too long, but I think that in the case where the user manually enters a description that is too long, they wont see the error unless they look in the developer tools.

I think it would be good to make sure this error is visible?

PR #2579, opened 2023-01-18
[#2538] Set radio type to string

### LINKED ISSUE #2538: Radio component shows inconsistently on the summary page

2.0.x

Den Haag Taiga issue 362 / user story 373

If a radio component has integers as the values of the options, then the summary page shows the value instead of the label.

Create a form with:
- A radio component with options with key / values:  `value A` / `valueA` and `value B` / `valueB`
- A radio component with options with key / values:  `value A` / `1` and `value B` / `2`

Fill in the form:
- Select `value A` for both radio buttons
- Go to the summary page
Screenshot from 2023-01-11 11-43-08
See that the second radio button shows they key (`1`) instead of the label (`value A`) of its options.

The label of the options should be shown even if the key is an integer

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2538 

This fix prevents this function https://github.com/formio/formio.js/blob/4.13.x/src/components/radio/Radio.js#L220 from inferring the type of the radio keys and casting them to another type than strings.


sergei-maertens (commented):
This doesn't retroactively fix existing forms/configurations, does it?

sergei-maertens (approved):

PR #2584, opened 2023-01-18
:bug: [#2576] -- fixup the migrations for long prefill_attribute values

sergei-maertens:
The longest value in our code is longer than the column length limit of 50.

The datamigration has been updated to include a schema change, while the actual schema migration is also added to reconcile the DB and model states.

Tested on dump of affected environment and now it migrates cleanly.


CharString (approved):
WAT

PR #2590, opened 2023-01-19
Check submission language relay to registration backends

### LINKED ISSUE #2546: Check how submissions in different languages end up in the registration backends
Epic #2170

- [x] StUF-ZDS: add the language code to `ExtraElementen`
- [x] ZGW API's (upstream ticket: https://github.com/VNG-Realisatie/gemma-zaken/issues/2162)
    - Add (optional) configuration option for which eigenschap (name) holds the language value
    - If no eigenschap is configured, during registration, do nothing
    - If the eigenschap is configured, validate that it exists on the specified zaaktype
    - If the eigenschap is configured, during registration, create a zaakeigenschap resource for the zaak with the value of the language code (let's stick to the same ISO standard as the documenten API here)
- [x] Objects API
- [x] E-mail
- [x] MS Graph API

-> at the very least, the language code metadata should be included.

CharString:
Closes #2546

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -284,7 +284,7 @@ lxml==4.9.1
     #   maykin-python3-saml
     #   xmlsec
     #   zeep
-mail-cleaner==1.0.0
+mail-cleaner==1.1.0
     # via -r requirements/base.in
 mail-parser==3.15.0
     # via django-yubin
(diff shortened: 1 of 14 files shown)


sergei-maertens (commented):
In between review - no remarks so far!

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2593, opened 2023-01-20
[#2541] Handle selectboxes key change in logic

### LINKED ISSUE #2541: Changing selectboxes key breaks logic tab

2.0.1 - latest

When a component's property name has changed, the logic rule(s) related to this component become invalid. This change should only trigger a error on the logic page however when saving the form after changing the property name of a component the PUT request made to /logic-rules returns
```json
{
    "type": "https://open-forms.test.nl/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:725fe779-9d5a-4250-b5b9-d852b0b1e1b9",
    "invalidParams": [
        {
            "name": "0.jsonLogicTrigger",
            "code": "invalid",
            "reason": "De opgegeven variabele is niet aanwezig in het formulier"
        }
    ]
}
```
Which in turn results in a "blank" Logic tab:
image

When Reopening this form the GET request to /logic-rules returns the "old" data that is not valid anymore (**selectBoxes** is now **selectBoxesNew**):
```json 
[
    {
        "uuid": "05bd1ed7-b502-49c8-8475-88a18c406aab",
        "url": "https://open-forms.test.nl/api/v2/forms/05bd1ed7-b502-49c8-8475-88a18c406aab/logic-rules",
        "form": "https://open-forms.test.nl/api/v2/forms/5dc6f0e8-7628-4c7b-aff6-e4c9c1f92491",
        "jsonLogicTrigger": {
            "==": [
                {
                    "var": "selectBoxes.1"
                },
                true
            ]
        },
        "order": 0,
        "triggerFromStep": null,
        "actions": [],
        "isAdvanced": false
    }
]
```
subsequently returning a 400 on every save of the form

1. Create form
2. Create selectboxes component
3. Add option to component
4. Navigate to form logic
5. Add trigger: If "selectBoxes" has value "X" and Save
6. Change property name of "selectBoxes" to "selectBoxesNew"
7. Save form

The logic tab should be broken now

SilviaAmAm:
Fixes #2541 

Tested on:
`{"and":[{"==":[{"var":"selectBoxes.a"},true]},{"==":[{"var":"selectBoxes.b"},true]}]}`
`{"==":[{"var":"selectBoxes.a"},true]}`

### DIFF
--- a/src/openforms/js/components/admin/form_design/utils.js
+++ b/src/openforms/js/components/admin/form_design/utils.js
@@ -65,9 +65,12 @@ const updateKeyReferencesInLogic = (existingLogicRules, originalKey, newKey) => 
     let newRule = {...rule};
     // Replace the key in the JSON trigger
     const stringJsonTrigger = JSON.stringify(rule.jsonLogicTrigger);
-    const compToReplace = JSON.stringify({var: originalKey});
+    const patternToReplace = new RegExp(
+      `\{"var":"${originalKey}(\\.)?([0-9a-zA-Z_\\-]+?)?"\}`,
+      'g'
+    );
     newRule.jsonLogicTrigger = JSON.parse(
-      stringJsonTrigger.replace(compToReplace, JSON.stringify({var: newKey}))
+      stringJsonTrigger.replaceAll(patternToReplace, `\{"var":"${newKey}$1$2"\}`)
     );
     // Replace the key in the actions
     newRule.actions = newRule.actions.map((action, index) => {


src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -65,9 +65,9 @@ const updateKeyReferencesInLogic = (existingLogicRules, originalKey, newKey) =>
     let newRule = {...rule};
     // Replace the key in the JSON trigger
     const stringJsonTrigger = JSON.stringify(rule.jsonLogicTrigger);
-    const compToReplace = JSON.stringify({var: originalKey});
+    const patternToReplace = new RegExp(`\{"var":"${originalKey}(\\.)?(.+?)?"\}`, 'g');
     newRule.jsonLogicTrigger = JSON.parse(
-      stringJsonTrigger.replace(compToReplace, JSON.stringify({var: newKey}))
+      stringJsonTrigger.replaceAll(patternToReplace, `\{"var":"${newKey}$1$2"\}`)
```
sergei-maertens, line 70: how does this behave with multiple vars (since regex is usually greedy)?

if you have a rule with `[{"var": "foo.bar"}, {"var": "foo.baz"}]`, is this going to replace every var individually, or will it consider the string `{"var": "foo.bar"}, {"var": "foo.baz"}`  to be one occurrence?

Perhaps the `$2` group should be limited to `[0-9a-zA-Z_\-]`?

SilviaAmAm: It would give `[{"var":"fooNew.bar"}, {"var":"fooNew.baz"}]` (the `+?` in the 2nd group makes it match as few times as possible). But I think it's a good idea to limit it to the chars you mention.

sergei-maertens (approved):

sergei-maertens:
Backported via 405e3369583144908657ba32846bd237f7dfeaf3

PR #2594, opened 2023-01-20
[#2591] add rate limits to API endpoints for pausing and submitting f…

### LINKED ISSUE #2591: Add limits to submissions/pausing configurable
Currently, there is rate limiting on the API (see API settings). However, we need specific rate limits for certain "final" endpoints to prevent spamming.

There are 2 endpoints that need to be rate limited far more strictly and configurable via env-vars, than the general case. Note that this rate limiting should be relatively easy to do with DRF: https://www.django-rest-framework.org/api-guide/throttling/

* [x] Add Open Forms settings that read from envvars to set rate limits `OPEN_FORMS_API_PAUSE_RATE_LIMIT` (default=3) and `OPEN_FORMS_API_SUBMIT_RATE_LIMIT` (default=10)
* [x] Rate limit the amount of times someone can "submit/finalize" a form. Typically, a user does not submit multiple forms within 1 minute. Base the rate limit on above setting.
* [x] Rate limit the amount of times someone can "pause" a form. This actually allows spamming random people by adding someone else's e-mail address. Hence, we must rate-limit. Base the rate limit on above setting.
* [x] Add tests to see if the rate limits work
* [x] Manually test to see if the rate limit does not prevent normal usage of Open Forms (ie. you rate limited incorrect endpoints)

pi-sigma:
Closes #2591

### DIFF
--- a/src/openforms/submissions/api/viewsets.py
+++ b/src/openforms/submissions/api/viewsets.py
@@ -126,6 +126,15 @@ class SubmissionViewSet(
     lookup_field = "uuid"
     pagination_class = pagination.PageNumberPagination
 
+    @property
+    def throttle_scope(self):
+        # called/checked by the DRF scoped throttle class
+        if self.action == "_complete":
+            return "submit"
+        if self.action == "_suspend":
+            return "pause"
+        return None
+
     def get_object(self):
         if not hasattr(self, "_get_object_cache"):
             submission = super().get_object()
--- a/src/openforms/submissions/tests/test_api_throttling.py
+++ b/src/openforms/submissions/tests/test_api_throttling.py
@@ -1,0 +1,121 @@
+import datetime
+
+from django.conf import settings
+from django.test import override_settings
+
+from freezegun import freeze_time
+from rest_framework import status
+from rest_framework.reverse import reverse
+from rest_framework.test import APITestCase
+
+from openforms.utils.tests.cache import clear_caches
+
+from .factories import SubmissionFactory
+from .mixins import SubmissionsMixin
+
+REST_FRAMEWORK_MODIFIED = settings.REST_FRAMEWORK.copy()
+REST_FRAMEWORK_MODIFIED["DEFAULT_THROTTLE_RATES"]["pause"] = "3/minute"
+REST_FRAMEWORK_MODIFIED["DEFAULT_THROTTLE_RATES"]["submit"] = "10/minute"
+
+
+@override_settings(REST_FRAMEWORK=REST_FRAMEWORK_MODIFIED)
+class APIThrottlingTest(SubmissionsMixin, APITestCase):
+    def setUp(self):
+        self.addCleanup(clear_caches)
+        super().setUp()
+
+    def test_throttling_submit(self):
+        """Assert that the API endpoint for completing submissions is throttled"""
+
+        initial_datetime = datetime.datetime(
+            year=2023, month=1, day=21, hour=11, minute=30, second=0
+        )
+
+        with freeze_time(initial_datetime) as frozen_datetime:
+            # 10 requests per minute allowed
+            for i in range(10):
+                submission = SubmissionFactory.create()
+                self._add_submission_to_session(submission)
+                endpoint = reverse(
+                    "api:submission-complete", kwargs={"uuid": submission.uuid}
+                )
+                response = self.client.post(endpoint)
+
+                assert (
+                    response.status_code == status.HTTP_200_OK
+                ), "Expected: HTTP 200 (OK); Actual: HTTP {status}".format(
+                    status=response.status_code
+                )
+
+            # Request #11 should fail with HTTP 429
+            submission = SubmissionFactory.create()
+            self._add_submission_to_session(submission)
+            endpoint = reverse(
+                "api:submission-complete", kwargs={"uuid": submission.uuid}
+            )
+            response = self.client.post(endpoint)
+
+            assert (
+                response.status_code == status.HTTP_429_TOO_MANY_REQUESTS
+            ), "Expected: HTTP 429 (too many requests); Actual: HTTP {status}".format(
+                status=response.status_code
+            )
+
+            # Wait 1 minute for the next request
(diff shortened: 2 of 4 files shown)


pi-sigma:
A few tests broke after the introduction of rate limits. I cleared the cache on the failing tests, though in some cases I had to clear the cache on the entire suite.

src/openforms/submissions/tests/test_api_throttling.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class APIThrottlingTest(SubmissionsMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        cache.clear()
+
+        throttles = settings.REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"]
+        throttles["submit"] = "10/minute"
+        throttles["pause"] = "3/minute"
```
sergei-maertens, line 23: this kind of setting mutation breaks test isolation, you _have_ to use `django.test.override_settings` so that settings can be restored after a test(case)

src/openforms/submissions/tests/test_api_throttling.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.timezone = ZoneInfo("UTC")
```
sergei-maertens, line 25: what does this do?

src/openforms/submissions/api/viewsets.py
```diff
@@ -126,6 +126,13 @@ class SubmissionViewSet(
     lookup_field = "uuid"
     pagination_class = pagination.PageNumberPagination
 
+    def get_throttles(self):
+        if self.action == "_complete":
+            self.throttle_scope = "submit"
```
sergei-maertens, line 131: is this what DRF recommends to set the throttle scope? looks a bit weird and not-best practices to me

src/openforms/submissions/api/viewsets.py
```diff
@@ -314,7 +321,11 @@ def status(self, request, *args, **kwargs):
         },
     )
     @transaction.atomic()
-    @action(detail=True, methods=["post"], url_name="suspend")
+    @action(
+        detail=True,
+        methods=["post"],
+        url_name="suspend",
+    )
```
sergei-maertens, line 328: I don't see the reason to change the formatting of this code?

src/openforms/submissions/tests/test_disabled_forms.py
```diff
@@ -42,6 +43,11 @@ class InactiveFormTests(SubmissionsMixin, APITestCase):
     Tests for forms that are not active.
     """
 
+    def setUp(self):
+        # clear cache to avoid throttling
+        super().setUp()
+        cache.clear()
```
sergei-maertens, line 49: is this still needed if you fix the setting overrides that broke isolation? If it is, please use the existing utility: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/utils/tests/cache.py#L4

this should also definitely be added as a cleanup in your throttle tests: `self.addCleanup(clear_caches)` which is the _source_ of test isolation breaking (after the tests ran you need some cleanup/cache reset).

You can also isolate those tests/caches by specifying a different cache, like the in memory cache via `@override_settings(CACHES=...)`.

sergei-maertens (changes requested):

src/openforms/submissions/api/viewsets.py
pi-sigma: According to the [DRF docs](https://www.django-rest-framework.org/api-guide/viewsets/#introspecting-viewset-actions)

pi-sigma:
I tried to disable throttling for the failing tests by overriding the relevant settings but that did not work. I now reverted to my original approach of manually clearing the caches for the failing tests.

There is a discussion of this kind of issue on the DRF repo: https://github.com/encode/django-rest-framework/issues/6030

src/openforms/submissions/api/viewsets.py
sergei-maertens: that's about introspecting, not about setting the `trottle_scope`. However, I could figure out from the source code that `view.throttle_scope` is checked (if it exists), so a cleaner implementation would be:

```py
@property
def throttle_scope(self):
    # called/checked by the DRF scoped throttle class
    if self.action == "_complete":
        return "submit"
    if self.action == "_suspend":
        return "pause"
    return None
```

magically setting attributes on `self` is discouraged, as it's hard to track down where they come from and where they are used.

src/openforms/submissions/tests/test_submission_suspension.py
```diff
@@ -32,6 +33,11 @@
 
 
 class SubmissionSuspensionTests(SubmissionsMixin, APITestCase):
+    def setUp(self):
+        # clear caches to avoid problems with throttling; @override_settings not working
+        # see https://github.com/encode/django-rest-framework/issues/6030
+        self.addCleanup(clear_caches)
```
sergei-maertens, line 39: please also add the `super().setUp()` call, as mixins added to the test case may also perform some setup for every test.

src/openforms/submissions/tests/test_api_throttling.py
```diff
@@ -0,0 +1,120 @@
...
+
+REST_FRAMEWORK_MODIFIED = settings.REST_FRAMEWORK.copy()
+REST_FRAMEWORK_MODIFIED["DEFAULT_THROTTLE_RATES"]["pause"] = "3/minute"
+REST_FRAMEWORK_MODIFIED["DEFAULT_THROTTLE_RATES"]["submit"] = "10/minute"
+
+
+@override_settings(REST_FRAMEWORK=REST_FRAMEWORK_MODIFIED)
+class APIThrottlingTest(SubmissionsMixin, APITestCase):
+    def setUp(self):
+        self.addCleanup(clear_caches)
```
sergei-maertens, line 24: See later note about calling the `super`

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2595, opened 2023-01-20
[#1902] Initial work on options for edit form

### LINKED ISSUE #1902: Implement using variables for radio / dropdown / checkbox options
Split from #1708 - closes #2181

It should be possible to specify the data source for dropdowns (and similar components) from a variable. This variable may be populated by fetching data from an external service, set from earlier submitted data, hardcoded...

In the form designer for these components, I should be able to:

- [x] Select the data source
    - Enter values manually at the component level (existing behaviour)
    - From form variable
- [ ] When populating from a form variable, configure:
    - The path to the array containing the options (default to the root, so that would be `{"var": "<varname>"}` in JsonLogic)
    - The path to use for each item's value
    - The path to use for each item's label
    - Value/label paths should default to the same defaults as the current formio data selection behaviour
- [x] Add data migration to update existing form definitions and set the appropriate data source
- [x] Use JsonLogic expressions for the "from form variable" behaviour, e.g.:

Example variable `someVar`:

```json
{
    "nested": {
        "path": [
            {
                "nested": {"value": "value1"},
                "label": "Label 1",
            },
            {
                "nested": {"value": "value2"},
                "label": "Label 2",
            }
        ]
    }
}
```
Example select component definition:

```json
{
    "type": "select",
    "dataSrc": "variable",
    "data": {
        "variable": {
            "variable": {"var": "someVar.nested.path"},
            "itemValue": {"var": "nested.value"},
            "itemLabel": {"cat": ["Item: ", {"var": "label"}]},
        }
    }
}
```

Resulting HTML:

```html
<select>
    <option value="value1">Label 1</option>
    <option value="value2">Label 2</option>
</select>
```

SilviaAmAm:
Fixes #1902


src/openforms/formio/dynamic_config/dynamic_options.py
```diff
@@ -0,0 +1,62 @@
...
+) -> None:
+    if glom(component, data_src_path, default=None) != "variable":
+        return
+
+    items_path = glom(component, "data.itemsExpression")
+
+    # The array of items from which we need to get the values
+    items_array = jsonLogic(items_path, data)
+
+    if not items_array or len(items_array) == 0:
```
CharString, line 25: `not []` is already `True`, the `len` call will never be executed.

src/openforms/formio/dynamic_config/dynamic_options.py
```diff
@@ -0,0 +1,62 @@
...
+    if not value_path:
+        return
+
+    for item in items_array:
+        value = jsonLogic(value_path, item)
+        # value is user input!
+        escaped_value = escape(value)
+        values.append({"label": escaped_value, "value": escaped_value})
+
+    assign(component, options_path, values, missing=dict)
```
CharString, line 62: This is fine, but the is_obj case is only a map away from becoming the exact same as iterating the items_array of primitives.
```py
if is_obj:
    if not (value_path := glom(component, "data.valueExpression", default=None)):
        return  # TODO maybe schedule an email to the form designer?
   
     items_array = (jsonLogic(value_path, item) for item in items_array)

# now items_array is an iterable of strings of user input
escaped_values = map(escape, items_array)

assign(component, options_path, [{"label": v, "value": v} for v in escaped_values], missing=dict)

File: src/openforms/js/components/form/select.js [hunk 3]
```diff
@@ -5,14 +5,37 @@ import {localiseSchema} from './i18n';
...
+  },
   {
     type: 'datagrid',
     input: true,
     label: 'Values',
     key: 'data.values',
     tooltip:
-      'The radio button values that can be picked for this field. Values are text submitted with the form data. Labels are text that appears next to the radio buttons on the form.',
+      'The values that can be picked for this field. Values are text submitted with the form data. ' +
+      'Labels are text that appears next to the radio buttons on the form.',
```
CharString, line 38: I expected this to need a change in some nl.json... but the English string wasn't there. :shrug:

src/openforms/js/components/form/select.js
```diff
@@ -5,14 +5,37 @@ import {localiseSchema} from './i18n';
 
 const Select = Formio.Components.components.select;
 
-const values = [
+const optionsChoices = [
```
CharString, line 8: :+1: I prefer the "Choices" over "values" a lot.

src/openforms/js/components/form/edit/tabs.js
```diff
@@ -87,40 +88,15 @@ const CHOICES_BASIC = {
...
-      ],
-    },
+    LABEL_REQUIRED,
+    KEY,
+    DESCRIPTION,
+    PRESENTATION,
+    MULTIPLE,
+    HIDDEN,
+    CLEAR_ON_HIDE,
+    IS_SENSITIVE_DATA,
```
CharString, line 98: Would people miss the combination "manual" and setting the default value?
I agree leaving it in, as is, is way too confusing.

CharString (changes requested):
The doc text change is the only real required one. The rephrasing of the 2 loops to a series of maps is your choice.

docs/manual/forms/examples/dynamic_options.rst
```diff
@@ -0,0 +1,75 @@
+.. _example_logic_dynamic_options:
+
+Formulier met dynamische opties
+===============================
+
+In dit voorbeeld maken we een deel-formulier bestaande uit 1 stap, waarbij
+de opties van een "radio" component dynamisch worden aangepast afhankelijk van de ingevulde antwoorden van een ander
+veld.
+
+Deze functionaliteit kan in hetzelfde manier worden gebruikt voor de "keuzelijst" component en de "selectievakjes"
```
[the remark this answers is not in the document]
sergei-maertens: > Voor de volledigheid: keuzelijstcomponent, is Nederlands. Houd je van Scrabble™?

These are discussions we've had before - we stick to what is displayed in the form designer (even though that is wrong :( )

src/openforms/formio/components/vanilla.py
```diff
@@ -110,11 +115,23 @@ class Checkbox(BasePlugin):
 class SelectBoxes(BasePlugin):
     formatter = SelectBoxesFormatter
 
+    def mutate_config_dynamically(
```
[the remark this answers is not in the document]
sergei-maertens: there is some mutation happening on the `FormDefinition` endpoints, which is "static" in the sense that there is no submission to provide input date. The `dynamic` here should highlight that the mutations are done in the context of a submission.

src/openforms/formio/dynamic_config/dynamic_options.py
SilviaAmAm: Nice!

Instead of the email, i went for a timeline log that can be seen in the admin!

src/openforms/js/components/form/select.js
SilviaAmAm: This should be added to src/openforms/js/lang/formio/nl.json, but I think that the translations are added at the end of the sprint, that's why I haven't added it :thinking:

src/openforms/js/components/form/edit/tabs.js
SilviaAmAm: It is still possible to add the default value, but only in the manual case. So I moved the field inside the 'options' panel and added a condition to it!

Screenshot from 2023-01-26 10-59-58

SilviaAmAm:
Noticed that if the data source selected is "variable" and the JSON logic expressions are configured, when switching to "manual" the editor doesn't remove straight away the editors.
This requires typing something in the option label field for it to go away:
afbeelding

src/openforms/js/components/form/edit/tabs.js
CharString: Oh, great. I didn't see that.

src/openforms/js/components/form/select.js
CharString: This wasn't towards this PR, but I expected the string to already exist from previous sprints. Guess not.

src/openforms/formio/dynamic_config/dynamic_options.py
```diff
@@ -0,0 +1,56 @@
...
+    if is_obj:
+        value_path = glom(component, "data.valueExpression", default=None)
+        # Case in which the form designer didn't configure the valueExpression by mistake.
+        # Catching this in validation on creation of the form definition is tricky, because the referenced component may
+        # be in another form definition.
+        if not value_path:
+            logevent.form_configuration_error(submission.form, component)
+            return
+
+        items_array = [jsonLogic(value_path, item) for item in items_array]
```
CharString, line 44: This could technically be a generator, but items_array will probably not be so long that it would matter.

CharString (approved):

docs/manual/forms/examples/dynamic_options.rst
```diff
@@ -0,0 +1,75 @@
...
+    #. **Postcode** component met label *Postcode*
+    #. **Tekstveld** component met label *Huisnummer en toevoeging*
+    #. **Tekstveld** component met label *Straatnaam*
+    #. **Tekstveld** component met label *Stad*
+
+#. Sleep een **radio** component op het witte vlak, vul de volgende
+   gegevens in en druk daarna op **Opslaan**:
+
+   * **Label**: Waar ontvangt u graag post?
+   * **Data bron**: *variabel*
```
sergei-maertens, line 45:
```suggestion
   * **Data bron**: *variabele*
```
Translation will need updating I presume?

docs/manual/forms/examples/dynamic_options.rst
```diff
@@ -0,0 +1,75 @@
...
+   gegevens in en druk daarna op **Opslaan**:
+
+   * **label**: IBAN
+   * **Meerdere waarden**: aangevinkt
+
+#. Sleep een **radio** component op het witte vlak, vul de volgende
+   gegevens in en druk daarna op **Opslaan**:
+
+   * **Label**: Waar wilt u de toeslag ontvangen?
+   * **Data bron**: *variabel*
```
sergei-maertens, line 65:
```suggestion
   * **Data bron**: *variabele*
```

src/openforms/formio/dynamic_config/dynamic_options.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # The array of items from which we need to get the values
+    items_array = jsonLogic(items_path, data)
+
+    if not items_array:
+        return
+
+    # Is each item a dict, like for repeating groups, or are they primitives
+    # ready to use?
+    is_obj = isinstance(items_array[0], dict)
```
sergei-maertens, line 33: what happens if they're not a homogenous type?

sergei-maertens (changes requested):

src/openforms/formio/dynamic_config/dynamic_options.py
CharString: Then we ask for that hypothesis lightning talk.

docs/manual/forms/examples/dynamic_options.rst
CharString: Is there some Sphinx plugin that can replace these references to literals in the UI with the actual literals used in the application? I took this as an adjective about `Data bron` which is clearly wrong.

sergei-maertens: Don't think so, and even then, it's coming from JS code dealing with formio & those related translations... Easier to keep track of it manually via PR review.

sergei-maertens:
Based on a quick scan of the code + demo via screenshare: I'm excited about this!

However, I'm parking this until #2533 is merged, which may introduce a couple of merge conflicts in the JS code.

src/openforms/formio/dynamic_config/dynamic_options.py
SilviaAmAm: So far we have the two cases:
1. It's just a string, so there is no need to specify `valueExpression` in the configuration of the component
2. It's an object, so we use `valueExpression` to get the value for the object. 
But if they are different types, we have no way to configure 'for option 1, make the value in way X, for option 2 in way Y.' So for now, non-homogeneous types are not expected in this PR.

Although I should probably add the check that the variable configured is an array :thinking:

sergei-maertens: Created follow up tickets in #2623 and #2624

src/openforms/js/components/form/select.js
sergei-maertens: cool JS-tip: use backticks (`` ` ``) for multiline strings :)

src/openforms/js/components/form/select.js
sergei-maertens, line 8: If I see this correctly, then it's exactly the same as `OPTIONS_CHOICES` except for the `key` being `values` vs. `data.values`, no?

SilviaAmAm: Yes it's true, now that I moved the other parts to the `openForms` attribute, the rest is the same except the `values` component. I will refactor it!

PR #2602, opened 2023-01-24
[#2599] fix typo in registration constants

### LINKED ISSUE #2599: registrations.constants contains a typo
https://github.com/open-formulieren/open-forms/blob/be55e6e8d3034ce22412d9c4a10bd30a10098c35/src/openforms/registrations/constants.py#L37

This doesn't affect functionality in any way. It just looks off when you read the code.
NB To fix this, we also need to change the name accordingly in the database!

pi-sigma:
Fixes #2599


src/openforms/forms/migrations/0067_fix_typo_20230124_1624.py
```diff
@@ -0,0 +1,36 @@
...
+def fix_typo(apps, schema_editor):
+    FormDefinition = apps.get_model("forms", "FormDefinition")
+
+    form_definitions = FormDefinition.objects.all()
+
+    form_definitions_to_update = []
+    for form_definition in form_definitions:
+        updated_form_definition = False
+        for comp in iter_components(configuration=form_definition.configuration):
+            if comp["registration"]["attribute"] == "Strainitiator_straatat":
```
sergei-maertens, line 17: this also needs a `FormVariable.objects.filter(prefill_attribute="...").update(prefill_attribute="...")` call, as a form variable should exist for every component.

See https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/models/form_variable.py#L142

sergei-maertens (changes requested):

src/openforms/forms/migrations/0067_fix_typo_20230124_1624.py
pi-sigma: What should I filter by and update to? The incorrect "Strainitiator_straatat" is not a `prefill_attribute` for any of the plugins. Perhaps I misunderstand the connection between `FormVariable`s and form components.

sergei-maertens (approved):

PR #2603, opened 2023-01-24
🌐 Localise field labels and keys.

### LINKED ISSUE #1242: Component Labels in Form builder in English
Taiga 298

The default value of the component label is not translated.
Screenshot from 2022-01-26 11-51-18

CharString:
Closes #1242

Dead ends:

 - The 'get defaultSchema' is called, and merged, but doesn't propagate to the rendered fields. So that doesn't work.
 - localising just the `static get builderInfo` results in an infinite loop.
 - localising just the `static schema` results in the builder not using it, because it will call the one on the Formio class, not our extension.
 - trying a dynamic import results in flaky translations; the schema is called synchronously, and once a "bad" value is returned while waiting for our import Promise to resolve, it might end up in the render of the form and subsequent schema calls will not trigger a re-render, even though the schema has changed.
 - Setting it on the prototype is painful with babel decorators being an experimental thing; we don't want to replace the original `schema` and `builderInfo`, we want to do something with their return value. Or maybe my ECMAScript-fu is just not good enough to handle decorating static class methods.


src/openforms/js/components/form/i18n.js
```diff
@@ -0,0 +1,29 @@
+import * as en from '../../lang/formio/en.json';
+import * as nl from '../../lang/formio/nl.json';
```
sergei-maertens, line 2: let's use the 'absolute' imports here, webpack is configured appropriately :)

```suggestion
import * as en from 'lang/formio/en.json';
import * as nl from 'lang/formio/nl.json';
```

sergei-maertens (changes requested):
one tiny remark

CharString:
Somehow forgot the family members. #2605 was not introduced by this PR.

sergei-maertens (approved):

Issue #2549, opened 2023-01-12
Accessibility - Cookie banner is overlapping footer links


LaurensBurger:
### Product versie / Product version

latest

### Omschrijf het probleem / Describe the bug

The cookie notice should not overlap the footer as this block accessibility to the list items in the footer. Which in turn causes the focus on these items to not be visible.

### Verwacht gedrag / Expected behavior

Desktop
![image](https://user-images.githubusercontent.com/112621537/212094000-fa5bb3b6-f64c-4d7a-a7f4-4a9dee8dcab4.png)

Mobile
![image](https://user-images.githubusercontent.com/112621537/212094122-1af1c6c7-2521-418f-b168-0dce4727ec11.png)

sergei-maertens:
doesn't work as expected yet on mobile

jiromaykin:
Issue seems to be only occurring in Chrome on Android, possibly due to "URL bar resizing"

PR #2596, opened 2023-01-23
:lipstick: increase bottom banner

jiromaykin:
Increase banner distance on bottom for all screen sizes.
Fixes  open-formulieren/open-forms#2549

PR #2610, opened 2023-01-26
:lipstick: fix cookie banner for chrome on Android

jiromaykin:
Increase banner distance on bottom for all screen sizes, specifically Android.
Fixes  open-formulieren/open-forms#2549

### DIFF
--- a/src/openforms/js/public.js
+++ b/src/openforms/js/public.js
@@ -17,3 +17,35 @@ const init = () => {
 };
 
 onLoaded(init);
+
+// observe whether cookiebanner is in sticky mode or not
+// note: rootMargin must be specified in pixels or percent, not REMs/EMs.
+const observer = new IntersectionObserver(
+  ([e]) => e.target.classList.toggle('stuck', e.intersectionRatio < 1),
+  {root: document.querySelector('body'), rootMargin: '0px 0px 31px 0px', threshold: [1]}
+);
+
+observer.observe(document.querySelector('.cookie-notice'));
+
+// in case page height is less than the full device height
+// (this is a state where the IntersectionObserver will not be triggered, but shadow still needs to be removed)
+const hasScrollbar = document.body.scrollHeight > window.innerHeight;
+
+if (hasScrollbar === false) {
+  document.querySelector('.cookie-notice').classList.add('no-shadow');
+}
+
+// ugly solution that may sent too many requests:
+
+// window.onscroll = function () {
+//   var totalPageHeight = document.body.scrollHeight;
+//   var scrollPoint = window.scrollY + window.innerHeight;
+
+//   // check if we hit the bottom of the page
+//   if (scrollPoint >= totalPageHeight) {
+//     console.log('scrolled to the bottom');
+//     document.querySelector('.cookie-notice').classList.add('stuck');
+//   } else {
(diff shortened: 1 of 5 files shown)


jiromaykin:
Notes to self: "...The resizing behavior of the URL bar is changing in Chrome on Android starting in version 56. Here's what you should know:
Lengths defined in viewport units (i.e. vh) will not resize in response to the URL bar being shown or hidden. Instead, vh units will be sized to the viewport height as if the URL bar is always hidden. That is, vh units will be sized to the "largest possible viewport". This means 100vh will be larger than the visible height when the URL bar is shown...."
"...There is one exception to the above changes and that is for elements that are position: fixed..."

jiromaykin:
Extra whitespace problem occurs in Chrome only on all Android devices.
See also: https://github.com/open-formulieren/open-forms/pull/2596#issuecomment-1401748960

android-chrome-tablets
android-chrome-telefoons

jiromaykin:
@sergei-maertens :
Notes:
-  Since I am not sure if the Footer will have a different height for different users, it's a bit of mystery how much space the footer might take up on mobile views (do all users have a maximum of 3 links?) - so I am assuming an average of 150px here.
-  If the Footer height would be known (or calculated with javascript) this could be subtracted from the max viewheight but for now i do not see a CSS only solution. 
-  Another problem is that on Chrome on Androids there is the "URL bar resizing" behaviour, which means an extra white space is either added or substracted, but I haven't been able to replicate this behaviour in Browserstack, so it might only be viewed on a real native device (which i do not have) - for now I am assuming an average of 5% of the viewheight.

sergei-maertens:
When the page height is less than the full device height, I just expect the cookie bar to be completely above the footer. In the other case, it should/could float above the content - using `position: sticky` should achieve this.

The shadown/overlay styling is not 100% necessary for me, but it would be cool if we can target the sticky vs. static state and style appropriately.

jiromaykin:
@sergei-maertens The IntersectionObserver is very interesting, and I got it working (for a bottom-banner) in my Codepen here: https://codepen.io/jirosworld/pen/YzjOKGZ
but I cannot get it to work inside Open-forms yet; (perhaps because of all sorts of margins/paddings) - and I don't know much yet about Django so I may have done terrible things in the templates.

src/openforms/js/public.js
```diff
@@ -17,3 +18,35 @@ const init = () => {
...
+
+// observe whether cookiebanner is in sticky mode or not
+// note: rootMargin must be specified in pixels or percent, not REMs/EMs.
+const observer = new IntersectionObserver(
+  ([e]) => e.target.classList.toggle('stuck', e.intersectionRatio < 1),
+  {root: document.querySelector('body'), rootMargin: '0px 0px 31px 0px', threshold: [1]}
+);
+
+observer.observe(document.querySelector('.cookie-notice'));
+
```
jiromaykin, line 30: @sergei-maertens I give up 😟; this code worked shortly (where the drop-shadow would disappear neatly when scrolled down) but now it is not working anymore - in this same public.js file I've done some more experimenting (with different results in desktops and mobile). Something is conflicting somewhere, and I'm not sure what it might be? Also: the Master/Base html templates may not be correct.

sergei-maertens:
Closing because I have a POC of a much simpler fix

PR #2611, opened 2023-01-27
[#2568] add template syntax validation to privacy_policy_label

### LINKED ISSUE #2568: config.privacy_policy_label does not have template syntax validator hooked up
Discovered via https://github.com/open-formulieren/open-forms/pull/2560#pullrequestreview-1249905614

Ideally, this HTML field should be validated for valid Django template syntax (with our sandboxed backend). This should probably also enforce that the `{% privacy_policy %}` tag is present. (TBD)

pi-sigma:
Closes #2568


SilviaAmAm (commented):
I think it might be useful to add a test, similar to https://github.com/open-formulieren/open-forms/blob/master/src/openforms/emails/tests/test_confirmation_email.py#L111?

src/openforms/config/models.py
```diff
@@ -360,6 +360,14 @@ class GlobalConfiguration(SingletonModel):
         ),
         default="Ja, ik heb kennis genomen van het {% privacy_policy %} en geef uitdrukkelijk "
         "toestemming voor het verwerken van de door mij opgegeven gegevens.",
+        validators=[
+            DjangoTemplateValidator(
+                required_template_tags=[
+                    "privacy_policy",
+                ],
+                backend="openforms.template.openforms_backend",
```
SilviaAmAm, line 368: I see that the issue mentions the `openforms.template.sandbox_backend` is there a reason why you used this one instead?

sergei-maertens (approved):

PR #2613, opened 2023-01-27
[#2461] expose Form.product as static variable

### LINKED ISSUE #2461: Add product(.price .name .information) as variables
### Thema / Theme

Form designer

This will allow form builders to create forms where product information van be shown, and used with logic.

_No response_

_No response_

pi-sigma:
Closes #2461


src/openforms/products/static_variables/static_variables.py
```diff
@@ -0,0 +1,29 @@
...
+        if not submission or not submission.form.product:
+            return None
+
+        product = submission.form.product
+        data = {
+            "uuid": product.uuid,
+            "name": product.name,
+            "price": product.price,
+            "information": product.information,
+        }
```
SilviaAmAm, line 28: Maybe it would be nice to use a serializer to serialise the data to a dict

src/openforms/products/static_variables/static_variables.py
```diff
@@ -0,0 +1,29 @@
...
+if TYPE_CHECKING:  # pragma: nocover
+    from openforms.submissions.models import Submission
+
+
+@register_static_variable("product")
+class Product(BaseStaticVariable):
+    name = _("Product")
+    data_type = FormVariableDataTypes.object
+
+    def get_initial_value(self, submission: Optional["Submission"]) -> Optional[dict]:
```
SilviaAmAm, line 18: Fyi, since python 3.10 there is an equivalent way of writing `Optional["Submission"]` which is `"Submission" | None` (https://peps.python.org/pep-0604/). I think we tend to use the new syntax more.

src/openforms/products/tests/test_static_variables.py
```diff
@@ -0,0 +1,25 @@
...
+from ..models import Product
+
+
+class TestProductStaticVariables(TestCase):
+    def test_product_static_data(self):
+        test_prod = Product.objects.create(
+            name="test_prod", price=9.99, information="Let Op!"
+        )
+        submission = SubmissionFactory.create()
+        submission.form.product = test_prod
```
SilviaAmAm, line 15: It is also possible to write this only using the factory:
```python
    form__product__name="test_prod",
    form__product__price=9.99,
    form__product__information="Let Op!",
)
```

src/openforms/products/tests/test_static_variables.py
```diff
@@ -0,0 +1,25 @@
+from django.test import TestCase
+
+from openforms.submissions.tests.factories import SubmissionFactory
+from openforms.variables.service import get_static_variables
+
+from ..models import Product
+
+
+class TestProductStaticVariables(TestCase):
+    def test_product_static_data(self):
```
SilviaAmAm, line 10: It may also be good to test the case where there is no submission

src/openforms/products/static_variables/static_variables.py
pi-sigma: I get a `TypeError` when using `|` with a string with `None` (`Submission | None` would work if `Submission` was defined already but raises a `NameError` otherwise)

src/openforms/products/static_variables/static_variables.py
pi-sigma: Isn't serialization the process of converting data types like `dict` to a string (e.g. JSON)? If `dict` is the correct type of return value for `get_initial_value`, what else is there to do?

pi-sigma: It works, but all attributes of `product` are transformed into strings. In particular, `product.price` will be (for example) `'9.99'` instead of `9.99`. Is this intended?

SilviaAmAm: I see that we don't override `COERCE_DECIMAL_TO_STRING`, and looking at the documentation of the decimal field (https://www.django-rest-framework.org/api-guide/fields/#decimalfield), it seems that it is true by default.

sergei-maertens: doing math on floats, especially when money is involved, is a bad idea due to floating point rounding errors.

It'd make more sense to expose the price as an integer, in cents and then divide that back by 100 for the final result.

sergei-maertens (commented):
Can we please add an integration test as well that actually does some calculations with the price to ensure this works as intended?

Ideally I'd just put the `product` model instance in the context rather than a dict with that shape, but that might not work with json logic for the `{"var": "product.price"}` lookups.

docs/manual/forms/pricing_logic.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+From Variable
+=============
+
+The price is determined from a separate user-defined variable, the value of which is turn
+determined by a combination of logic rules and the price fixed in the product configuration.
+
+For example, the "unit price" for a certain service (``product``) provided by the municipality is
+EUR 99.99,-. The total price to be paid by the client (``order``) depends on whether they are
```
SilviaAmAm, line 40: Is the formatting of `EUR 99.99,-` intended?

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1247,8 +1265,12 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl}) => {
               onChange={onFieldChange}
             />
             <PriceLogic
+              pricingLogic={state.form.pricingLogic}
+              formVariables={state.formVariables}
               rules={state.priceRules}
-              onChange={onPriceRuleChange}
+              // TODO: change both PricingLogic and Rule
+              onChange={onPricingLogicChange}
```
SilviaAmAm, line 1272: You can change the `onChange` callback to:
```javascript
event => dispatch({
      type: 'FIELD_CHANGED',
      payload: {name: 'form.pricingLogic', value: event.target.value},
})
```
Then you can remove the event 'CHANGED_PRICING_LOGIC'

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -116,6 +116,7 @@ const initialFormState = {
   submitting: false,
   logicRules: [],
   priceRules: [],
+  pricingLogic: 'static',
```
SilviaAmAm, line 119: This should be nested inside the form (since it is an attribute on the form model). Otherwise it won't automatically be sent to the backend!

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -723,8 +727,13 @@ function reducer(draft, action) {
...
 
     /**
-     * Price rules actions
+     * Price logic actions
      */
+    case 'CHANGED_PRICING_LOGIC': {
+      const {value} = action.payload;
+      draft.pricingLogic = value;
+      break;
+    }
```
SilviaAmAm, line 736: This can be removed (see other comment)

src/openforms/js/components/admin/form_design/PriceLogic.js
```diff
@@ -39,38 +40,51 @@ const PRICING_MODES = [
       defaultMessage: 'Use logic rules to determine the price',
     }),
   ],
+  [
+    'fromVariable',
```
SilviaAmAm, line 44: This should match the value in the backend (so `from_variable` rather than the camelCase).

joeribekker:
Refinement: This hangs on refactor logic/price logic together -> functional design first.

joeribekker:
Refinement: Decided to close this for now, waiting for new stakeholders.

PR #2614, opened 2023-01-27
[#2581] Inconsistent dates / datetimes

### LINKED ISSUE #2581: Date variable (user variable) is stored as date + time

2.0.0 - latest

this causes issues with logic, where a date variable is expected but a datetime variable is used.
https://github.com/open-formulieren/open-forms/blob/c129aa820712f5dda98aea9148d07e9a4ec31745/src/openforms/submissions/models/submission_value_variable.py#L348
Adding .date() to this line solves this issue for form logic.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2581

### DIFF
--- a/src/openforms/formio/dynamic_config/date.py
+++ b/src/openforms/formio/dynamic_config/date.py
@@ -1,6 +1,6 @@
 import operator
 from datetime import date, datetime, time
-from typing import Literal, Optional, TypedDict, Union, cast
+from typing import Literal, Optional, TypedDict, cast
 
 from django.utils import timezone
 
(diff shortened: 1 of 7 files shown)


CharString (changes requested):
So
```
base_value = cast(
        Optional[Union[datetime, str]],
        glom(data, config["variable"], default=None),
    )
```
should include the `date` type in the Union...

I think instead of this casting, a pattern match is a lot clearer:
```python
base_value = glom(data, config["variable"], default=None)
match base_value:
    case datetime.datetime():
        assert ...
        base_date = timezone.localtime(value=base_value).date()
    case datetime.date():
        base_date = base_value
    case str():
        pass # what happened to this case in the Union?
    case None:
        return None
```

> Never send a human to do a machine's job.
> -- Agent Smith

The only thing worse than an Any type is a wrong guess at a cast.
But maybe this is out of scope for the issue...

CharString (approved):

CharString:
@sergei-maertens 
One caveat on second reading:
If `DateConstraintConfiguration` contains information that can help form designers with debugging broken logic. I think we should provide that in the error. But I don't know what helpful information is in scope here.

sergei-maertens:
> @sergei-maertens One caveat on second reading: If `DateConstraintConfiguration` contains information that can help form designers with debugging broken logic. I think we should provide that in the error. But I don't know what helpful information is in scope here.

Clarified in the office - this is not executed at form-design time, but at submission/form-execution runtime, so there's no validation error stuff relevant. This only concerns sanity checks for input.

sergei-maertens (approved):

Issue #2491, opened 2022-12-15
Add document Title for accessibility


jiromaykin:
### Thema / Theme

Frontend

### Omschrijving / Description

Currently the user only sees "openforms" as a `<head><title>` - for accessibility reasons it would be more helpful to have the organization_name here (which will also become visible in the browser history). In case the admin has not assigned an organization_name, there should be a placeholder.

### Added value / Toegevoegde waarde

This will increase accessibility, more ingo here: https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html
"The descriptive title of an HTML Web page is marked up with the <title> element so that it will be displayed in the title bar of the user agent... **_Titles identify the current location without requiring users to read or interpret page content_**..."

### Aanvullende opmerkingen / Additional context

See also this Accessibility issue about meaningful titles:
https://github.com/open-formulieren/open-forms/issues/2267

jiromaykin:
Note to self: possible solution
`<title>{% block title %}{% if config.organization_name %}{{ config.organization_name }} {% else %}Openforms{% endif %}{% endblock %}</title>`

sergei-maertens:
Fixed via ef401cd0b3ee056444cc9f8d4a909db6b115eee6

PR #2616, opened 2023-01-27
Favicon and orgname in page title

LaurensBurger:


sergei-maertens:
@jiromaykin this overlaps with one of your issues I believe

jiromaykin:
This indeed overlaps with one other issue: https://github.com/open-formulieren/open-forms/issues/2491

PR #2618, opened 2023-01-27
Update sdk docs

sergei-maertens:
This completes the documentation improvements - Storybook points to the RTD docs and vice versa.


INSTALL.rst
```diff
@@ -4,9 +4,16 @@ Installation
...
+the `documentation <https://open-forms.readthedocs.io/en/stable/installation/>`_ for
+alternative installation methods.
 
-The project is developed in `Python`_ using the `Django framework`_.
+.. note:: The `Developer documentation`_ on Read the Docs provides working internal
+   links, in case you are reading this on Github or elsewhere.
+
+The (backend) project is developed in `Python`_ using the `Django framework`_.
+
+For the SDK (frontend), see the `Storybook <https://open-formulieren.github.io/open-forms-sdk/>`_
```
svenvandescheur, line 15: Can we elaborate a bit more whether this applies to developing custom application using the sdk?

svenvandescheur (approved):

INSTALL.rst
sergei-maertens: It's not there yet, but we should strive towards that, yes. Plain markdown documentation pages in Storybook should make that possible.

PR #2647, opened 2023-02-03
[#2641] Fix checking domains for links in emails

### LINKED ISSUE #2641: E-mail confirmation template domain check should not include subdomains (www. etc.)

latest

The email template field for forms checks if domains are present in the allowed domains "whitelist". however this whitelist tells the the admin the exclude http://www from these domains. 
When a form designer in turn tries to add a link to the confirmation email they will receive an error telling them the domain they used in the link is not in de allowed list:
Allowed domain entry: exampledomain.com
Link in email: https://www.exampledomain.com/destination
Error: Domain is not present in allowed domain list: www.exampledomain.com

URLSanitationValidator uses netloc but should use tld to check domains in whitelist

SilviaAmAm:
Fixes #2641 

See the note about caching for `tldextract`: https://github.com/john-kurkowski/tldextract#note-about-caching
Questions: 
- Shall we keep the runtime bootstrapping behaviour? 
- How do we deal with updating the file with the suffixes? 
- Should we set the env var `TLDEXTRACT_CACHE`?

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -30,6 +30,7 @@ semantic-version
 tabulate
 weasyprint
 zeep
+tldextract
 # Pinned setuptools to a version lower than 58 to allow pyzmail36 to be
 # installed, as required by django-yubin.
 setuptools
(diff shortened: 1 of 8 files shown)


SilviaAmAm:
Discussed with Chris: let's go for the simpler solution of 
```python
any(parsed.netloc.endswith(domain) for domain in allowlist)
```

PR #2654, opened 2023-02-03
[#2646] Validate privacy policy accepted

### LINKED ISSUE #2646: User's can compete a form by posting /_complete to a form

latest

internal: DH 391

This way they could skip the final agreement checkbox:
image

```json 
{
    "status": "done",
    "result": "success",
    "errorMessage": "",
    "publicReference": "OF-2JC68T",
    "confirmationPageContent": "<p>Bedankt voor uw inzending.</p>",
    "reportDownloadUrl": "https://open-forms.test.nl/api/v2/submissions/327/vd-b040ecasd12c3023312/download",
    "paymentUrl": "",
    "mainWebsiteUrl": "https://www.bla.nl"
}
```

If this is tried with a form that has required fields the api correctly returns:
```json
{
    "incompleteSteps": [
        {
            "formStep": "https://open-forms.test.bla.nl/api/v2/forms/72c5cc2e-68c3-4788-9309-dcd4912afd43/steps/7981ce88-70c0-4f7e-970b-c6409b495872"
        }
    ],
    "submissionAllowed": "yes"
}
```

SilviaAmAm:
Fixes #2646 

SDK PR: https://github.com/open-formulieren/open-forms-sdk/pull/354


src/openforms/submissions/api/serializers.py
```diff
@@ -477,3 +478,9 @@ class SubmissionStepSummarySerialzier(serializers.Serializer):
         required=True,
     )
     data = SubmissionComponentSummarySerializer(many=True)
+
+
+class SubmissionUpdateSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Submission
+        fields = ("privacy_policy_accepted",)
```
sergei-maertens, line 486: I had expected this to be a field for the submission completion serializer and not a separate endpoint with additional call overhead?

sergei-maertens (changes requested):

src/openforms/submissions/api/serializers.py
SilviaAmAm: Ah I wasn't sure, since the `_complete` endpoint doesn't receive data in the request body so I thought it would be weird if it were to receive the `privacy_policy_accepted` field. Also it needs to be persisted to the database, so I thought it'd make more sense to have it in another endpoint than have the `_complete` endpoint update the submission. But indeed there is overhead for an extra call :shrug:

sergei-maertens: it's a POST call already, so it makes perfect sense to add any "finalizers" in the request body.

The serializer can easily be adapted to persist this on the submission instance too, so let's do it like that.

src/openforms/submissions/api/viewsets.py
```diff
@@ -233,6 +234,13 @@ def _complete(self, request, *args, **kwargs):
         ---
         """
         submission = self.get_object()
+
+        update_serializer = SubmissionUpdateSerializer(
+            instance=submission, data=request.data
+        )
+        update_serializer.is_valid(raise_exception=True)
+        update_serializer.save()
```
sergei-maertens, line 242: I don't really understand why we need a separate serializer for this? IMO we can bake this entirely into the existing flow, but we'll need to refactor that a little bit.

The basis should be the `CompletionValidationSerializer` for this entire request-response cycle, in essence:

```py
serializer = SubmissionCompletionSerializer(
    instance=submission,
    data=request.data,
)
...
serializer.save()
return Response(serializer.data)
```

Where `SubmissionCompletionSerializer` is refactored to handle the behaviour of validation + succesful serialization:

1. Validate that privacy policy is accepted
2. Validate that required steps are done etc.

You an do that by marking serializer fields as read/write only so that they are only emitted in response or request body.

This does not need to be a model serializer, there is nothing stopping you from implementing the `save` method yourself and then just adding a line `self.instance.privacy_policy_accepted = True` and `self.instance.save()` to it.

sergei-maertens (changes requested):

src/openforms/submissions/api/viewsets.py
SilviaAmAm: @sergei-maertens I tried implementing this suggestion, I am not confident that it now matches what you had in mind :sweat_smile: 
The coverage is failing I think because `NestedSubmissionRelatedField` is no longer used anywhere. 
If these changes are going in the right direction then I'll remove it

src/openforms/submissions/api/viewsets.py
```diff
@@ -233,13 +232,8 @@ def _complete(self, request, *args, **kwargs):
...
         submission = self.get_object()
-        validation_serializer = validate_submission_completion(
-            submission, request=request
-        )
-        if validation_serializer is not None:
-            return Response(
-                validation_serializer.data, status=status.HTTP_400_BAD_REQUEST
-            )
+
+        validate_submission_completion(submission, request=request)
```
sergei-maertens, line 236: looks good, clever trick with the `MaxLengthValidator` :smile: 

one thing I would change is make this function return the serializer and call the `serializer.is_valid(raise_exception=True)` and `serializer.save()` in the view, rather than lumping that into the validation function (so it also makes sense to rename the function to `get_completion_validation_serializer` to have a more accurate name).

SilviaAmAm:
The coverage seems to complain about lines that have been removed :thinking:

src/openforms/scss/pdfs/_privacy-policy.scss
```diff
@@ -0,0 +1,13 @@
+.submission-privacy-policy {
+  display: flex;
+  align-items: flex-start;
+
+  &__text {
```
sergei-maertens, line 5: for the future:

```scss
@use "microscope-sass/lib/bem";

.submission-privacy-policy {
  ...
  
  @include bem.element('text') { ... }
}
```

-> this spits out the correct strict selector `.submission-privacy-policy .submission-privacy-policy__text` instead of just `.submission-privacy-policy__text`

src/openforms/submissions/api/serializers.py
```diff
@@ -151,6 +151,7 @@ class Meta:
             "is_authenticated",
             "payment",
             "form_url",
+            "privacy_policy_accepted",
```
sergei-maertens, line 154:
```suggestion
```

sergei-maertens (changes requested):

sergei-maertens:
> The coverage seems to complain about lines that have been removed thinking

try rebasing on master again? the codecov config was updated recently as well, maybe it's something to do with that

sergei-maertens (approved):

PR #2659, opened 2023-02-07
:bug: [#2636] Fix Form designer bugs noticed during end2end testing

### LINKED ISSUE #2636: Bugs noticed while end-to-end testing

2.1.0-alpha.2

Noticed a couple bugs while setting up a test form to check the translations feature completeness. This tests only the filling out a form aspect, not the registration/PDF process behind it.

* On the form designer variables tab, creating a user-defined variable of the boolean type and setting the initial value to `False` doesn't seem to work. It always shows 'True' as active value and that's also what it initializes with.
* In the form designer in the advanced options for a logic rule you can select from which step the trigger applies. Setting and saving it works, however, displaying it again shows a dropdown without active value. The value is still set and functionally it works.

Then the next item is a concern/question:

* User defined variables where the value is set dynamically through logic does not persist in the database. This is okay as long as all the logic rules are evaluated when dealing with the submission data, but incorrect if this is not done.

**User defined variable**

1. Create a new form or edit an existing form
2. Navigate to the "variables" tab
3. Navigate to the "gebruikersvariabelen" nested tab
4. Add a variable with any name
5. Select the data type "boolean"
6. Set the initial value to No/Nee
7. Click "Save and edit the form"
8. Navigate back to the variables and inspect the user defined variable
9. Observe that the dropdown is set to Yes/Ja (instead of No/Nee)

**"Trigger step from" display**

1. Create a form with at least one step
2. Navigate to the "Logica" tab
3. Add a logic rule
4. Click the cog icon to the left to open the advanced options
5. Select a step in the dropdown
6. Save the form and reload it
7. Inspect the logic rule advanced configuration
8. Observe that the dropdown has no value set

**Unsaved user-defined variable**

1. Create a new form or edit an existing form with at least one step
2. Navigate to the "variables" tab
3. Navigate to the "gebruikersvariabelen" nested tab
4. Add a variable with any name and an initial value
5. Add a logic rule that sets the value of the variable to a different value than the initial value (trigger can be set to `true` to always run)
6. Save the form
7. Start the form and fill out the details
8. Navigate to the overview (or next step)
9. Check the submissions admin and see that the value of the variable in the database is still the initial value

Note - I haven't tested if the user-defined variable is maybe saved when the form is completed, if that's the case, then this is a non-issue.

* initial value False is saved and displayed correctly
* the selected trigger_from_step is displayed correctly
* not sure :grimacing: 

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
Fix #2636


stevenbal:
@sergei-maertens I think it would be good to add tests for this, but I'm not sure what the status of Selenium is now since you mentioned working on Playwright during standup?

sergei-maertens:
> @sergei-maertens I think it would be good to add tests for this, but I'm not sure what the status of Selenium is now since you mentioned working on Playwright during standup?

it's working locally for me but not on CI :( feel free to try out my branch: https://github.com/open-formulieren/open-forms/pull/2660

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -216,3 +217,201 @@ def setUpTestData():
...
+                await page.locator("[name='initialValue']").select_option(label="No")
+
+                # Verify that the select updated to the selected value
+                await expect(page.locator("[name='initialValue']")).to_have_value(
+                    "false"
+                )
+
+            with phase("Save variable and check state"):
+                await page.get_by_text("Save and continue editing").click()
+                await page.get_by_role("tab", name="Variabelen").click()
```
stevenbal, line 264: @sergei-maertens I noticed that after saving, to locale seems to switch to NL, is this something you noticed too when implementing the Playwright tests?

sergei-maertens: it's because the form designer is (now) hardcoded to use NL, the rest of the page should still be EN

sergei-maertens: discussed - we'll detect in the forms endpoint if it's an admin request and then *not* activate the default language (which is causing this problem)

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -216,3 +217,201 @@ def setUpTestData():
...
+                await page.get_by_role("tab", name="Gebruikersvariabelen").click()
+
+                # Verify that the select still holds the correct value
+                await expect(page.locator("[name='initialValue']")).to_have_value(
+                    "false"
+                )
```
sergei-maertens, line 270: can we also assert that the label of the dropdown is "No"?

src/openforms/js/components/admin/form_design/FormLogic.js
```diff
@@ -56,6 +56,7 @@ const FormLogicRules = ({rules, onAdd, onChange, onDelete}) => {
   // FIXME - getting the validation errors by index breaks if you then delete rules/reorder
   // rules -> they're displayed for the wrong rule. When deleting/reordering rules, the
   // validation errors state needs to be re-ordered the same way.
+  console.log(rules);
```
sergei-maertens, line 59:
```suggestion
```

sergei-maertens (changes requested):

src/openforms/forms/tests/e2e_tests/test_form_designer.py
stevenbal: I tried doing this (with `.to_have_text`), but that seems to just concatenate the text of all the possible options together (e.g. `------YesNo`)

stevenbal:
Possible flakiness detected? :grimacing: https://github.com/open-formulieren/open-forms/actions/runs/4163234717/jobs/7203373922, worked after restarting

sergei-maertens (approved):

PR #2678, opened 2023-02-13
[#2653] update zaaksysteem configuration

### LINKED ISSUE #2653: More Zaaksysteem configuration for FileUploads
DH Taiga: 385

### Thema / Theme

Form designer

We need to fill certain attributes when creating an EIO (some might already be done, not sure)

1. Add fields to ZGW API config AND FileUpload component which overrides the general config value if FileUpload value is empty):

* [x] Default document bronorganisatie (RSIN and can be overriden per FileUpload field). No default value. -> EIO->bronorganisatie
* [x] Default document vertrouwelijkheidsaanduiding (can be overriden per FileUpload field). Default: zaakvertrouwelijk (in general config, empty in FileUpload) -> EIO->vertrouwelijkheid
  _Possible values are:_

    - openbaar - Openbaar
    - beperkt_openbaar - Beperkt openbaar
    - intern - Intern
    - zaakvertrouwelijk - Zaakvertrouwelijk
    - vertrouwelijk - Vertrouwelijk
    - confidentieel - Confidentieel
    - geheim - Geheim
    - zeer_geheim - Zeer geheim

3. Only in FileUpload:

* [x] Add Title field that sets the Title of the document -> EIO->Titel

4. Only in ZGW API confih:

* [x] ZGW/StUF “auteur” field needs to be set to “Aanvrager” -> EIO->auteur

_No response_

_No response_

pi-sigma:
Closes #2653

* added several fields to the ZGW API configuration and to file upload


src/openforms/contrib/zgw/service.py
```diff
@@ -154,7 +155,7 @@ def create_attachment_document(
     base64_body = b64encode(submission_attachment.content.read()).decode()
 
     document_options = {
-        "author": "open-forms",
+        "author": options.get("author"),
```
sergei-maertens, line 158: not an optional field

src/openforms/js/components/form/file.js
```diff
@@ -27,6 +27,32 @@ const REGISTRATION = {
       },
       valueProperty: 'url',
     },
+    {
+      type: 'textfield',
+      key: 'registration.bronorganisatie',
+      label: 'Bronorganisatie',
+      description: 'RSIN of the organization which creates the ZAAK.',
```
sergei-maertens, line 34:
```suggestion
      description: 'RSIN of the organization which creates the ENKELVOUDIGINFORMATIEOBJECT.',
```

src/openforms/js/components/form/file.js
```diff
@@ -27,6 +27,32 @@ const REGISTRATION = {
...
+      label: 'Vertrouwelijkheidaanduiding',
+      widget: 'choicesjs',
+      tableView: true,
+      data: {
+        get values() {
+          return jsonScriptToVar('CONFIDENTIALITY_LEVELS');
+        },
+      },
+      description:
+        'Indication of the level to which extend the dossier of the ZAAK is meant to be public.',
```
sergei-maertens, line 48:
```suggestion
        'Indication of the level to which extent the INFORMATIEOBJECT is meant to be public.',
```

src/openforms/js/components/form/file.js
```diff
@@ -27,6 +27,32 @@ const REGISTRATION = {
...
+    },
+    {
+      type: 'textfield',
+      key: 'registration.titel',
+      label: 'Titel',
```
sergei-maertens, line 53:
```suggestion
      label: 'Title',
```
let's be consistent in our english source strings

src/openforms/js/components/form/file.js
```diff
@@ -27,6 +27,32 @@ const REGISTRATION = {
...
+      description: 'The name under which the INFORMATION OBJECT is formally known.',
```
sergei-maertens, line 54:
```suggestion
      description: 'The name under which the INFORMATIEOBJECT is formally known.',
```
weird mixture of NL and EN but those terms don't translate well.

src/openforms/registrations/contrib/zgw_apis/models.py
```diff
@@ -57,6 +58,20 @@ class ZgwConfig(SingletonModel):
         validators=[validate_rsin],
         help_text=_("Default RSIN of organization, which creates the ZAAK"),
     )
+    vertrouwelijkheidaanduiding = models.CharField(
+        _("vertrouwelijkheidaanduiding"),
+        max_length=24,
+        choices=VertrouwelijkheidsAanduidingen.choices,
+        default=VertrouwelijkheidsAanduidingen.zaakvertrouwelijk,
+        help_text=_(
+            "Indication of the level to which extend the dossier of the ZAAK is meant to be public."
```
sergei-maertens, line 67:
```suggestion
            "Indication of the level to which extend the dossier of the ZAAK is meant to be public."
```
is this actually used for the zaak or only the documents of a zaak?

src/openforms/registrations/contrib/zgw_apis/models.py
```diff
@@ -57,6 +58,20 @@ class ZgwConfig(SingletonModel):
...
+        ),
+    )
+    auteur = models.CharField(
+        _("author"),
+        max_length=200,
+        default="Aanvrager",
+    )
```
sergei-maertens, line 74: maybe make it explicit that this is for the documents registered with the zaak?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -239,8 +239,21 @@ def register_submission(
         )
 
         for attachment in submission.attachments:
+            # attribute names conform to the Open Documenten API specification
             iot = attachment.informatieobjecttype or options["informatieobjecttype"]
-            doc_options = {**options, "informatieobjecttype": iot}
+            bron = attachment.bronorganisatie or options["organisatie_rsin"]
```
sergei-maertens, line 244:
```suggestion
            bronorganisatie = attachment.bronorganisatie or options["organisatie_rsin"]
```
let's stick to the terms of the standard

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -239,8 +239,21 @@ def register_submission(
...
+            vetrouwelik = (
```
sergei-maertens, line 245:
```suggestion
            vetrouwelijkheidaanduiding = (
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -239,8 +239,21 @@ def register_submission(
...
+                attachment.vertrouwelijkheidaanduiding
+                or options["vertrouwelijkheidaanduiding"]
+            )
+            titel = attachment.titel or options.get("titel")
```
sergei-maertens, line 249: `None` is not an accepted type for the API spec, and furthermore a non-empty string is required. So let's keep at least the type consistent:

```suggestion
            titel = attachment.titel or options.get("titel", "")
```

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
```diff
@@ -984,19 +984,26 @@ def test_reference_can_be_extracted(self, m):
...
+    def test_submission_with_zgw_backend_override_fields(self, m):
+        """Assert that override of default values for the ZGW backend works"""
         submission = SubmissionFactory.from_components(
             [
                 {
                     "key": "field1",
                     "registration": {
                         "informatieobjecttype": "https://catalogi.nl/api/v1/informatieobjecttypen/10",
+                        "bronorganisatie": "100000009",
+                        "vertrouwelijkheidaanduiding": "zeer geheim",
```
sergei-maertens, line 996:
```suggestion
                        "vertrouwelijkheidaanduiding": "zeer_geheim",
```
please check against the enum on https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/VNG-Realisatie/gemma-documentregistratiecomponent/1.2.0/src/openapi.yaml#tag/enkelvoudiginformatieobjecten/operation/enkelvoudiginformatieobject_create

Use the available constants to prevent these sort of mistakes please.

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
```diff
@@ -1031,21 +1038,37 @@ def test_submission_with_zgw_backend_override_attachment_iotype(self, m):
...
             document_create_attachment1_body["informatieobjecttype"],
             "https://catalogi.nl/api/v1/informatieobjecttypen/10",
         )
+        self.assertEqual(
+            document_create_attachment1_body["bronorganisatie"],
+            "100000009",
+        )
+        self.assertEqual(
+            document_create_attachment1_body["vertrouwelijkheidaanduiding"],
+            "zeer geheim",
```
sergei-maertens, line 1064: not valid according to API spec

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
```diff
@@ -1062,6 +1085,91 @@ def test_submission_with_zgw_backend_override_attachment_iotype(self, m):
...
         )
+        self.assertEqual(
+            document_create_attachment2_body["bronorganisatie"],
+            "000000000",
+        )
+        self.assertEqual(
+            document_create_attachment2_body["vertrouwelijkheidaanduiding"],
+            "openbaar",
+        )
+        self.assertIsNone(document_create_attachment2_body["titel"])
```
sergei-maertens, line 1096: not valid according to the API spec https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/VNG-Realisatie/gemma-documentregistratiecomponent/1.2.0/src/openapi.yaml#tag/enkelvoudiginformatieobjecten/operation/enkelvoudiginformatieobject_create - this is not nullable and must be a string with length 1-200

This also means a client side validator is needed to prevent titles > 200 chars being entered

src/openforms/submissions/models/submission_files.py
```diff
@@ -199,6 +199,41 @@ def informatieobjecttype(self) -> Optional[str]:
...
+    @property
+    def titel(self) -> str | None:
+        """
+        Get the title for this attachment from the configuration
+        """
+        formdef_config = self.submission_step.form_step.iter_components()
+        for component in formdef_config:
+            if component["key"] == self.form_key:
+                if titel := glom(component, "registration.titel", default=""):
+                    return titel
```
sergei-maertens, line 235: this repetition around `iter_components` is not great for performance, especially because the same processing is done multiple times if all these attributes are accessed (which they are).

Please use the caching datastructure https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/models/form_definition.py#L154

sergei-maertens (changes requested):

src/openforms/contrib/zgw/service.py
```diff
@@ -21,6 +21,7 @@ def create_zaak(
     client = config.zrc_service.build_client()
     today = date.today().isoformat()
     data = {
+        "auteur": options["author"],
```
sergei-maertens, line 24: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/vng-Realisatie/zaken-api/1.2.0/src/openapi.yaml#tag/zaken -> `auteur` does not exist in the request body, where did you pull this from?

src/openforms/contrib/zgw/service.py
```diff
@@ -108,7 +109,7 @@ def create_report_document(
     base64_body = b64encode(submission_report.content.read()).decode()
 
     document_options = {
-        "author": "open-forms",
+        "author": options.get("author", "Aanvrager"),
```
sergei-maertens, line 112: this still breaks if someone assigns `options["author"] = ""`

Wouldn't it be better to ensure that `auteur` is set to a non-empty string in the `options` dict _before_ calling these methods?

src/openforms/contrib/zgw/service.py
```diff
@@ -131,7 +132,7 @@ def create_csv_document(
     base64_body = b64encode(csv_data.encode()).decode()
 
     document_options = {
-        "author": "open-forms",
+        "author": options.get("author", "Aanvrager"),
```
sergei-maertens, line 135: see previous remark

src/openforms/contrib/zgw/service.py
```diff
@@ -154,7 +155,7 @@ def create_attachment_document(
     base64_body = b64encode(submission_attachment.content.read()).decode()
 
     document_options = {
-        "author": "open-forms",
+        "author": options.get("author", "Aanvrager"),
```
sergei-maertens, line 158: see previous remark

src/openforms/contrib/zgw/service.py
```diff
@@ -166,7 +167,9 @@ def create_attachment_document(
 
     options = {**options, **document_options}
 
-    return create_document(name, base64_body, options, get_drc=get_drc)
+    titel = options.get("titel", name)
```
sergei-maertens, line 170: `options["titel"] = ""` breaks this

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -73,6 +73,7 @@ def register_submission(
                 "informatieobjecttype": "informatieobjecttype_submission_report",
                 "organisatie_rsin": "organisatie_rsin",
                 "vertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                "author": "author",
```
sergei-maertens, line 76: why not map this to `auteur` directly so the right key ends up in the API calls?

sergei-maertens (changes requested):

src/openforms/contrib/zgw/service.py
pi-sigma: The only place where the client can assign to `author` is in the ZGW API configuration, where "Author" is a required field. How could `options["author"]` get to be equal to an empty string?

sergei-maertens: existing installations already have a working zgw API configuration, so they will typically not touch this, so the admin form validation will not enforce this. When they upgrade, the new column for the CharField is added to the database and it will not have a value (= empty string) since there is no default specified. That's a way that empty strings can end up here.

I missed that this is sourced from the global configuration, so we can fix this by setting a `default="Aanvrager"` on the model field.

However, I still want this intention/expectation to be made explicit, because the zgw config model is pretty distanced from where the value is actually used, so maybe add an `assert options["auteur"], "auteur must be a non-empty string"` to bake this assumption/required into the code.

src/openforms/contrib/zgw/service.py
```diff
@@ -107,8 +110,10 @@ def create_report_document(
     submission_report.content.seek(0)
     base64_body = b64encode(submission_report.content.read()).decode()
 
+    assert options["auteur"], "auteur must be a non-empty string"
```
sergei-maertens, line 113: I would expect this assertion in the `create_document` function, not here.

And then in the `document_options` it should be:

```
"auteur": options.get("auteur") or "Aanvrager",
```
to provide a fallback for empty or missing values (this requires a test to demonstrate empty auteur is handled! this test  is currently absent because the build would be failing due to the assert above)

src/openforms/contrib/zgw/service.py
```diff
@@ -130,8 +135,10 @@ def create_csv_document(
 ) -> dict:
     base64_body = b64encode(csv_data.encode()).decode()
 
+    assert options["auteur"], "auteur must be a non-empty string"
+
     document_options = {
-        "author": "open-forms",
+        "auteur": options.get("auteur", "Aanvrager"),
```
sergei-maertens, line 141: See above remarks

src/openforms/contrib/zgw/service.py
```diff
@@ -153,8 +160,10 @@ def create_attachment_document(
     submission_attachment.content.seek(0)
     base64_body = b64encode(submission_attachment.content.read()).decode()
 
+    assert options["auteur"], "auteur must be a non-empty string"
+
     document_options = {
-        "author": "open-forms",
+        "auteur": options.get("auteur", "Aanvrager"),
```
sergei-maertens, line 166: see above remarks

src/openforms/contrib/zgw/service.py
```diff
@@ -166,7 +175,12 @@ def create_attachment_document(
 
     options = {**options, **document_options}
 
-    return create_document(name, base64_body, options, get_drc=get_drc)
+    if "titel" in options and options["titel"]:
+        titel = options["titel"]
+    else:
+        titel = name
```
sergei-maertens, line 181:
```suggestion
    titel = options.get("titel") or name
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -63,16 +63,28 @@ class ObjectsAPIRegistration(BasePlugin):
...
+        service functions (e.g. :func:`create_report_document`) and involves a mapping
+        (and in some cases renaming) of variables which would otherwise not be
+        accessible from here. For example, 'vertrouwelijkheidaanduiding' must be named
+        'docVertrouwelijkheidaanduiding' because this is what the ZGW service functions
+        use."""
+
         config = ObjectsAPIConfig.get_solo()
         config.apply_defaults_to(options)
 
+        options["auteur"] = options.get("auteur", "Aanvrager")
```
sergei-maertens, line 78:
```suggestion
        options["auteur"] = options.get("auteur") or "Aanvrager"
```
please add tests where `auteur` is an empty string

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -63,16 +63,28 @@ class ObjectsAPIRegistration(BasePlugin):
...
+
         submission_report = SubmissionReport.objects.get(submission=submission)
         submission_report_options = build_options(
             options,
             {
                 "informatieobjecttype": "informatieobjecttype_submission_report",
                 "organisatie_rsin": "organisatie_rsin",
-                "vertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                "docVertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                "auteur": "auteur",
```
sergei-maertens, line 87: this doesn't seem necessary to me since it's not a rename in any way?

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -63,16 +63,28 @@ class ObjectsAPIRegistration(BasePlugin):
...
+        options["auteur"] = options.get("auteur", "Aanvrager")
+
         submission_report = SubmissionReport.objects.get(submission=submission)
         submission_report_options = build_options(
             options,
             {
                 "informatieobjecttype": "informatieobjecttype_submission_report",
                 "organisatie_rsin": "organisatie_rsin",
-                "vertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                "docVertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
```
sergei-maertens, line 86: here we should probably make this snake_case to be consistent, no?

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -91,7 +103,8 @@ def register_submission(
                 {
                     "informatieobjecttype": "informatieobjecttype_attachment",
                     "organisatie_rsin": "organisatie_rsin",
-                    "vertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                    "docVertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                    "auteur": "auteur",
```
sergei-maertens, line 107: see above

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -127,7 +140,8 @@ def register_submission(
                 {
                     "informatieobjecttype": "informatieobjecttype_submission_csv",
                     "organisatie_rsin": "organisatie_rsin",
-                    "vertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                    "docVertrouwelijkheidaanduiding": "vertrouwelijkheidaanduiding",
+                    "auteur": "auteur",
```
sergei-maertens, line 144: see above

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -859,7 +859,7 @@ def test_submission_with_objects_api_backend_bsn(self, m):
             document_create_body["informatieobjecttype"],
             "https://catalogi.nl/api/v1/informatieobjecttypen/1",
         )
-        self.assertNotIn("vertrouwelijkheidaanduiding", document_create_body)
+        self.assertNotIn("document_vertrouwelijkheidaanduiding", document_create_body)
```
sergei-maertens, line 862: there is no reason why this should ever be in the document create body, it's like asserting that `captain_picard` is not in `document_create_body`

I think this test asserts that if there's no explicit VA specified, that it should not be present in the request body. If a global value is configured, then it should be asserted that this global value is used.

Normally the VA is derived from the zaaktype/documenttype if not specified explicitly, so it should definitely still be possible to not specify any default value and have this key be absent in the document create body so that the API can derive it from the type-information instead.

src/openforms/registrations/contrib/zgw_apis/models.py
```diff
@@ -57,6 +58,32 @@ class ZgwConfig(SingletonModel):
...
+        _("vertrouwelijkheidaanduiding document"),
+        max_length=24,
+        choices=VertrouwelijkheidsAanduidingen.choices,
+        default=VertrouwelijkheidsAanduidingen.zaakvertrouwelijk,
+        help_text=_(
+            "Indication of the level to which extend the document associated with the "
+            "ZAAK is meant to be public. Can be overridden in the file upload "
+            "component of a given form."
+        ),
+    )
```
sergei-maertens, line 81: these should be `blank=True` without a default set, since the default is baked into the `zaaktype` and `informatieobjecttype` for the zaak/document being created. The Zaken API/Documenten API have this business logic baked in.

Checked this with Joeri - the ticket spec is ambiguous/wrong here.

src/openforms/registrations/contrib/zgw_apis/models.py
```diff
@@ -65,6 +92,16 @@ def apply_defaults_to(self, options):
...
         options.setdefault("organisatie_rsin", self.organisatie_rsin)
+        options.setdefault(
+            "zaak_vertrouwelijkheidaanduiding", self.zaak_vertrouwelijkheidaanduiding
+        )
+        # map variable to a key that conforms to Formio naming standards
+        # otherwise the value cannot be changed by the Formio component
+        options.setdefault(
+            "docVertrouwelijkheidaanduiding",
+            self.doc_vertrouwelijkheidaanduiding,
+        )
```
sergei-maertens, line 103: can you elaborate? this is backend code, I don't see how this affects formio :thinking:

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -239,8 +240,25 @@ def register_submission(
         )
 
         for attachment in submission.attachments:
+            # attribute names conform to the Open Documenten API specification
```
sergei-maertens, line 243:
```suggestion
            # attribute names conform to the Documenten API specification
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -239,8 +240,25 @@ def register_submission(
...
+            # `titel` should be a non-empty string
+            # `get_display_name` is used to enforce this
+            titel = attachment.titel or options.get(
+                "titel", attachment.get_display_name()
+            )
+            doc_options = {
+                **options,
+                "informatieobjecttype": iot,
+                "organisatie_rsin": bronorganisatie,
+                "docVertrouwelijkheidaanduiding": vertrouwelijkheidaanduiding,
```
sergei-maertens, line 259: key should only be included if there is a value for it

src/openforms/submissions/models/submission_files.py
```diff
@@ -190,15 +190,65 @@ def informatieobjecttype(self) -> Optional[str]:
         """
         Get the informatieobjecttype for this attachment from the configuration
         """
-        formdef_config = self.submission_step.form_step.iter_components()
-        for component in formdef_config:
+        # use configuration wrapper for caching
+        formdef_config = (
+            self.submission_step.form_step.form_definition.configuration_wrapper.configuration
+        )
+        for component in formdef_config["components"]:
```
sergei-maertens, line 197: this is wrong since the `configuration` is basically an alias of the underlying formio configuration.

Instead you should loop over the `FormioConfigurationWrapper` instance, which properly recurses while using caches.

src/openforms/submissions/models/submission_files.py
```diff
@@ -190,15 +190,65 @@ def informatieobjecttype(self) -> Optional[str]:
...
+    @property
+    def bronorganisatie(self) -> str | None:
+        """
+        Get the bronorganisatie for this attachment from the configuration
+        """
+        # use configuration wrapper for caching
+        formdef_config = (
+            self.submission_step.form_step.form_definition.configuration_wrapper.configuration
+        )
+        for component in formdef_config["components"]:
```
sergei-maertens, line 214: see above

src/openforms/submissions/models/submission_files.py
```diff
@@ -190,15 +190,65 @@ def informatieobjecttype(self) -> Optional[str]:
...
+    @property
+    def docVertrouwelijkheidaanduiding(self) -> str | None:
+        """
+        Get the vertrouwelijkheidaanduiding for this attachment from the configuration
+        """
+        # use configuration wrapper for caching
+        formdef_config = (
+            self.submission_step.form_step.form_definition.configuration_wrapper.configuration
+        )
+        for component in formdef_config["components"]:
```
sergei-maertens, line 230: see above

src/openforms/submissions/models/submission_files.py
```diff
@@ -190,15 +190,65 @@ def informatieobjecttype(self) -> Optional[str]:
...
+
+    @property
+    def titel(self) -> str | None:
+        """
+        Get the title for this attachment from the configuration
+        """
+        formdef_config = (
+            self.submission_step.form_step.form_definition.configuration_wrapper.configuration
+        )
+        for component in formdef_config["components"]:
```
sergei-maertens, line 247: see above

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/plugin.py
pi-sigma: The `create_document` function raises a `KeyError` if `doc_vertrouwelijkheidaanduiding` is not set.

sergei-maertens (approved):

PR #2690, opened 2023-02-13
:sparkles: [#2661] Service fetch configuration modal in Form Designer

### LINKED ISSUE #2661: Modal for `fetch-data` action type in the Form Designer
Epic https://github.com/open-formulieren/open-forms/issues/2169

Should also allow the user to either:
- create a new `service-fetch` configuration
- select an existing `service-fetch` configuration and edit that
- select an existing `service-fetch` configuration and copy it

stevenbal:
Accidentally closed https://github.com/open-formulieren/open-forms/pull/2662 :grimacing: 

Fixes #2661


src/openforms/js/components/admin/forms/MappingArrayInput.js
```diff
@@ -0,0 +1,165 @@
...
+  valueArrayInput = false,
+  ...extraProps
+}) => {
+  const intl = useIntl();
+
+  if (!mapping) {
+    mapping = valueArrayInput ? [['', ['']]] : [['', '']];
+  }
+
+  let inputs = [...mapping];
```
stevenbal, line 26: Initially I had something like this `const [inputs, setInputs] = useState([...mapping])` here, but I ran into the issue that I wanted to be able to manage the state from the `ServiceFetchConfigurationPicker` when I select an existing service fetch configuration

src/openforms/js/components/admin/form_design/logic/actions/Action.js
```diff
@@ -47,6 +54,33 @@ const Action = ({prefixText, action, errors = {}, onChange, onDelete}) => {
             </DSLEditorNode>
 
             <ActionComponent action={action} errors={errors} onChange={onChange} />
+
+            <button
```
stevenbal, line 58: We probably don't want to add this until the backend stuff works as well?

stevenbal:
Created a separate issue to implement the contents of the "try it out" tab: https://github.com/open-formulieren/open-forms/issues/2777

src/openforms/js/components/admin/form_design/logic/actions/Action.js
stevenbal: Currently hidden behind a feature flag

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.js
```diff
@@ -0,0 +1,332 @@
...
+            {selectExisting ? (
+              <>
+                <ActionButton
+                  name="_save_as_new"
+                  text={intl.formatMessage({
+                    description: 'Save as new service fetch configuration button label',
+                    defaultMessage: 'Save as new',
+                  })}
+                  onClick={formik.handleSubmit}
+                />
```
sergei-maertens, line 260: I'd expect a prop `type="submit"` and then in the parent `<form>` tag something like `<form onSubmit={formik.handleSubmit}>`, so that submitting the form with the <kbd>Enter</kbd> key also works.

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.stories.mdx
```diff
@@ -0,0 +1,72 @@
...
+import ServiceFetchConfigurationForm from './ServiceFetchConfigurationForm';
+
+<Meta title="Form design/ServiceFetchConfigurationForm" component={ServiceFetchConfigurationForm} />
+
+# ServiceFetchConfigurationForm
+
+Form to edit/create a ServiceFetchConfiguration
+
+export const Template = args => (
+  <FormLogicContext.Provider
```
sergei-maertens, line 13: see earlier remarks about decorators + code preview.

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.stories.mdx
```diff
@@ -0,0 +1,72 @@
...
+      },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+### Props
+
+<ArgsTable story="Default" />
```
sergei-maertens, line 72:
```suggestion
<ArgsTable of={ServiceFetchCOnfigurationForm} />
```

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
```diff
@@ -0,0 +1,94 @@
...
+      alert(JSON.stringify(values, null, 2));
+    },
+  });
+
+  return (
+    <div className="servicefetchconfiguration-picker">
+      <div className="servicefetchconfiguration-select">
+        <FormRow>
+          <Field
+            name={'fetchConfiguration.existing'}
```
sergei-maertens, line 49:
```suggestion
            name="fetchConfiguration.existing"
```

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
```diff
@@ -0,0 +1,94 @@
...
+            label={
+              <FormattedMessage
+                defaultMessage="Choose existing configuration"
+                description="Service fetch configuration modal select existing field label"
+              />
+            }
+          >
+            <Select
+              name="fetchConfiguration.existing"
```
sergei-maertens, line 58: this *shouldn't* be necessary to repeat because the parent `Field` should inject that, but maybe Formik requires it?

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
```diff
@@ -0,0 +1,94 @@
...
+
+                setSelectExisting(!!event.target.value);
+              }}
+              allowBlank
+            />
+          </Field>
+        </FormRow>
+      </div>
+
+      <div className="servicefetchconfiguration-form">
```
sergei-maertens, line 78: this should probably be a proper `<form onSubmit={formik.handleSubmit)>` tag :)

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
```diff
@@ -0,0 +1,94 @@
...
+          onChange={onChange}
+        />
+      </div>
+    </div>
+  );
+};
+
+ServiceFetchConfigurationPicker.propTypes = {
+  onReplace: PropTypes.func.isRequired,
+};
```
sergei-maertens, line 92: the component has `data`, `onChange`, and `onFormSave` props

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.stories.mdx
```diff
@@ -0,0 +1,83 @@
+import {ArgsTable, Canvas, Meta, Story} from '@storybook/addon-docs';
+
+import {FormLogicContext} from '../Context';
+import ServiceFetchConfigurationPicker from './ServiceFetchConfigurationPicker';
+
+<Meta
+  title="Form design/ServiceFetchConfigurationPicker"
+  component={ServiceFetchConfigurationPicker}
+/>
```
sergei-maertens, line 9: maybe group all these service fetch stories under `Form design/Service Fetch/...`?

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.stories.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+# ServiceFetchConfigurationPicker
+
+Form that is shown when adding a ServiceFetchConfiguration. The user can opt to select an existing
+configuration and copy/update it; or to create a new configuration from scratch.
+
+export const Template = args => (
+  <FormLogicContext.Provider
```
sergei-maertens, line 17: see earlier decorator remarks

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.stories.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+## Presentation
+
+<Canvas>
+  <Story
+    name="Default"
+    args={{
+      onChange: () => {
+        null;
+      },
```
sergei-maertens, line 74: probably `jest.fn(...)` is better

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.stories.mdx
```diff
@@ -0,0 +1,83 @@
...
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+### Props
+
+<ArgsTable story="Default" />
```
sergei-maertens, line 83: document the component props rather than story args please

src/openforms/js/components/admin/forms/MappingArrayInput.js
```diff
@@ -0,0 +1,153 @@
...
+      copy[index][1] = event.target.value;
+    }
+
+    onChange(copy);
+  };
+
+  const onArrayInputChange = (index, value) => {
+    copy[index][1] = value;
+    onChange(copy);
+  };
```
sergei-maertens, line 52: maybe it'd be better to use `produce` from `immer` to deal with immutable/draft stores? Would be nice if we can at some point remove lodash too.

src/openforms/js/components/admin/forms/MappingArrayInput.js
```diff
@@ -0,0 +1,153 @@
...
+
+    onChange(copy);
+  };
+
+  const onInputChange = (index, event) => {
+    if (event.target.name === 'keyInput') {
+      copy[index][0] = event.target.value;
+    } else if (event.target.name === 'valueInput') {
+      copy[index][1] = event.target.value;
+    }
```
sergei-maertens, line 44: I think you can make code flatter with destructuring and switch statements:

```js
const {name, value} = event.target;
switch (name) {
    case 'keyInput': {
        copy[index][0] = value;
        break;
    }
    case 'valueInput': {...}
    default: throw new Error(`unexpected name '${name}'`);
}
```

src/openforms/js/components/admin/forms/MappingArrayInput.js
```diff
@@ -0,0 +1,153 @@
...
+      <button type="button" className="button button--plain" onClick={onAdd}>
+        {addButtonMessage || (
+          <span className="addlink">
+            <FormattedMessage
+              description="Add item to multi-input field"
+              defaultMessage="Add item"
+            />
+          </span>
+        )}
+      </button>
```
sergei-maertens, line 139: looks like we need to break out this plain button into a re-usable component as I'm now seeing markup being duplicated.

src/openforms/js/components/admin/forms/MappingArrayInput.stories.mdx
```diff
@@ -0,0 +1,60 @@
+import {ArgsTable, Canvas, Meta, Story} from '@storybook/addon-docs';
+
+import ArrayInput from './ArrayInput';
+import MappingArrayInput from './MappingArrayInput';
+
+<Meta title="Admin/Custom/MappingArrayInput" component={MappingArrayInput} />
```
sergei-maertens, line 6: there are a bunch of warnings in the console for these stories, please resolve those

src/openforms/js/components/admin/forms/MappingArrayInput.js
```diff
@@ -0,0 +1,153 @@
...
+      </table>
+      <button type="button" className="button button--plain" onClick={onAdd}>
+        {addButtonMessage || (
+          <span className="addlink">
+            <FormattedMessage
+              description="Add item to multi-input field"
+              defaultMessage="Add item"
+            />
+          </span>
+        )}
```
sergei-maertens, line 138: the <kbd>+</kbd> icon is missing when a message is provided:

image

src/openforms/scss/components/admin/_index.scss
```diff
@@ -21,6 +21,7 @@
 @import './logic-trigger';
 @import './logic-action';
 @import './logic-type-selection';
+@import './servicefetchconfiguration_form';
```
sergei-maertens, line 24: can we stick to the existing file name style with dashes rather than underscores?

src/openforms/scss/components/admin/_servicefetchconfiguration_form.scss
```diff
@@ -0,0 +1,3 @@
+.servicefetchconfiguration-form {
+  margin-top: 2em;
+}
```
sergei-maertens, line 3: components should never affect other components! this is a top-level block, so in the CSS box model you are only allowed to touch borders and paddings. Margins push into other siblings which leads to unpredictable layouts.

Instead, the parent container should probably be adapted with flexbox gaps or better spacing logic (possibly through a modifier).

sergei-maertens (changes requested):
Overall I'm very happy with the UX of this modal and the general direction of the code!

That said, I of course have a bunch of things to remarks :sweat_smile:  -  see the inline annotations too please.

* There's a trailing comma somewhere too that's showing up in storybook
  image

* Storybook story templates/args should be a lot more explicit - you have more control than simply mapping story args directly to component props. This also allows you to expose story controls that concern only that story and 'package' it into a larger datastructure suitable for the component props.

* Documentation `ArgsTable` should always document the props of the component rather than the args of the story. These often align, but not necessarily. You may also want to hide some controls, but still document the component props.

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
stevenbal: This field isn't actually managed by Formik, so I think I can just remove the duplication

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
sergei-maertens: I'd put it close to where you'll actually handle the submit event, so probably here. That does also improve composabililty (because you can't put form tags inside other form tags).

src/openforms/js/components/admin/form_design/FormLogic.stories.mdx
```diff
@@ -0,0 +1,139 @@
...
+# FormLogic
+
+Component used to configure the logic for a given Form. Logic is configured by adding Rules, each of
+which specifies either a single condition (simple) or a JsonLogic condition (advanced). In either
+case, this condition can lead to one or more configured actions that should be taken (e.g. changing
+the value of a variable or component).
+
+export const Template = args => (
+  <FeatureFlagsContext.Provider value={{of_service_fetch_enabled: true}}>
+    <FormContext.Provider
```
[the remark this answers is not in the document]
stevenbal: Do we have code preview in our storybook yet? I can't find an example of it

src/openforms/scss/components/admin/_servicefetchconfiguration_form.scss
stevenbal: replaced this with flexbox gaps on the `servicefetchconfiguration-picker` element

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.stories.mdx
stevenbal: When using this for the `MappingArrayInput` it doesn't seem to work :thinking:
image

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.stories.mdx
stevenbal: I changed it to actually accept the arguments that are passed, but I'm not sure what `jest.fn` would do in this case?

src/openforms/js/components/admin/forms/MappingArrayInput.js
```diff
@@ -24,20 +25,24 @@ const MappingArrayInput = ({
...
-    onChange(newMapping);
+    onChange(name, newMapping);
   };
 
   const onDelete = index => {
-    return produce(mapping, draft => draft.filter((_, itemIndex) => itemIndex !== index));
+    const newMapping = produce(mapping, draft =>
+      draft.filter((_, itemIndex) => itemIndex !== index)
+    );
+    onChange(name, newMapping);
```
stevenbal, line 35: @sergei-maertens produce didn't seem to actually lead to anything being deleted, I'm still unsure of how this should all work together, because the state is initialized by Formik in the `ServiceConfigurationFormPicker`, but at the same time it seems as though React components should also keep track of their own state

sergei-maertens (approved):

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -116,6 +127,31 @@ const ActionFetchFromService = ({action, errors, onChange}) => {
...
+        isOpen={isModalOpen}
+        closeModal={closeModal}
+        title={
+          <FormattedMessage
+            description="Service fetch configuration selection modal title"
+            defaultMessage="Service fetch configuration"
+          />
+        }
+        contentModifiers={['with-form', 'large']}
+      >
```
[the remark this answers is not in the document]
sergei-maertens: -> not possible due to the layout

PR #2704, opened 2023-02-14
[#2699] Uploads in repeating group

### LINKED ISSUE #2699: Uploads in repeating group missing from attachments in submission

2.0.3 + latest

Files that are uploaded as part of repeatinggroup are missing from the attachment list in a submission and the submissionattachments.

this example submission only shows 1 attachment:
image

But in reality has muliple that are part of a repeatinggroup(1) that contained more then just the attachment (2,3):
image
image

image

SilviaAmAm:
Fixes #2699

### DIFF
--- a/src/openforms/formio/utils.py
+++ b/src/openforms/formio/utils.py
@@ -45,8 +45,23 @@ def iterate_data_with_components(
     Iterate through a configuration and return a tuple with the component JSON, its value in the submission data
     and the path within the submission data.
 
-    Note: for repeating groups (editgrid component), each component can have multiple values, depending on how many items
-    the user added.
+    For example, for a configuration with components:
+
+    .. code:: json
+
+        [
+            {"key": "surname", "type": "textfield"},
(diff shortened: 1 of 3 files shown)


src/openforms/formio/utils.py
```diff
@@ -35,6 +35,64 @@ def iter_components(
             )
 
 
+def iterate_data_with_components(
```
sergei-maertens, line 38: let's not put this in the utils and create the impression it is public API intended to be used in multiple places. it does the job for the bug that we're dealing with, but it should really be just an implementation detail because the code necessary for this upload handling is not very robust or elegant.

So, let's make it a private function by renaming it to `_iterate_data_with_components` and move it back into the `src/openforms/submissions/attachments.py` module

SilviaAmAm:
New issue: https://github.com/open-formulieren/open-forms/issues/2713

sergei-maertens (approved):

sergei-maertens:
Backported in 667358a7c48fbd7901c2c0df2cdb358a7e89987b

PR #2714, opened 2023-02-15
[#2671] fix QR code

### LINKED ISSUE #2671: QR-code for enabling MFA unreadable in dark mode

2.0.3

A collegue just tried to login to the admin portal for the first time, and noticed she couldn't scan the QR-code in dark mode:
image
(QR code blurred for obvious reasons)

This only goes for the QR-code used for the initial setup: Upon investigating using Windows' native dark mode (no plugins), at first I couldn't reproduce it: The QR-codes in the TOTP tab of the admin panel -do- show with a white background.

1. Set Windows (10) to dark mode
image

2. Delete current (if any) TOTP device from the OF admin-panel

3. Login and try to set the TOTP device anew

### Tasks

* [ ] Enforce white background for QR-code (so also in dark mode)
* [ ] Make sure there is some white padding (10px?) around the QR-code

pi-sigma:
Fixes #2671 

*patched CSS for QR code by adding white background

### DIFF
--- a/src/openforms/scss/_vendor.scss
+++ b/src/openforms/scss/_vendor.scss
@@ -5,6 +5,7 @@
 @import './vendor/react-tabs';
 @import './vendor/rjsf-field';
 @import './vendor/tinymce';
+@import './vendor/django-two-factor-auth';
 
 // this violates a check by postcss-selector-lint, but it's the SDK snippet which has no
 // classnames to target.
--- a/src/openforms/scss/vendor/_django-two-factor-auth.scss
+++ b/src/openforms/scss/vendor/_django-two-factor-auth.scss
@@ -1,0 +1,3 @@
+body.login .centered-image {
+  background-color: white;
+}


src/openforms/scss/_vendor.scss
```diff
@@ -5,6 +5,7 @@
 @import './vendor/react-tabs';
 @import './vendor/rjsf-field';
 @import './vendor/tinymce';
+@import 'vendor/django-two-factor-auth';
```
sergei-maertens, line 8:
```suggestion
@import './vendor/django-two-factor-auth';
```

src/openforms/scss/vendor/_django-two-factor-auth.scss
```diff
@@ -0,0 +1,6 @@
+.centered-image {
```
sergei-maertens, line 1: oof this is a very generic classname, let's scope it to avoid unintended bleeding into other pages

```suggestion
body.login .centered-image {
```

sergei-maertens (changes requested):

src/openforms/scss/vendor/_django-two-factor-auth.scss
```diff
@@ -0,0 +1,6 @@
+body.login.centered-image {
```
sergei-maertens, line 1:
```suggestion
body.login .centered-image {
```
the space is important because the classname is applied to a _child of_ the body, not to the body itself.

src/openforms/scss/_vendor.scss
```diff
@@ -5,9 +5,10 @@
 @import './vendor/react-tabs';
 @import './vendor/rjsf-field';
 @import './vendor/tinymce';
+@import './vendor/django-two-factor-auth';
 
 // this violates a check by postcss-selector-lint, but it's the SDK snippet which has no
-// classnames to target.
+// classnames to target.s
```
sergei-maertens, line 11:
```suggestion
// classnames to target.
```

src/openforms/scss/vendor/_django-two-factor-auth.scss
```diff
@@ -0,0 +1,6 @@
...
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
```
sergei-maertens, line 4: not sure where these lines now suddenly come from, they are already in the 2fa stylesheet itself. The only line that's needed for this patch is the background color line.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2718, opened 2023-02-16
Replace DjangoChoices with models.TextChoices

### LINKED ISSUE #2712: Replace django-choices usage
... with models.TextChoices or appropriate subclasses

-> this works natively with type checkers, while django-choices does not.

### LINKED ISSUE #2721: Upgrade django-simple-certmanager
https://github.com/maykinmedia/django-simple-certmanager/pull/16 was never included in OF

sergei-maertens:
Closes #2712
Closes #2721 

Tasks

- [x] Replace usage in own code
- [x] Update libraries to remove django-choices usage

### DIFF
--- a/src/openforms/payments/constants.py
+++ b/src/openforms/payments/constants.py
@@ -1,39 +1,43 @@
+from django.db import models
 from django.utils.translation import gettext_lazy as _
 
-from djchoices import ChoiceItem, DjangoChoices
+
+class PaymentRequestType(models.TextChoices):
+    get = "get"
+    post = "post"
 
 
-class PaymentRequestType(DjangoChoices):
-    get = ChoiceItem("get")
-    post = ChoiceItem("post")
+class UserAction(models.TextChoices):
+    accept = "accept"
+    exception = "exception"
+    cancel = "cancel"
+    # back = "back"
+    # decline = "decline"
+
+    unknown = "unknown"
 
 
-class UserAction(DjangoChoices):
-    accept = ChoiceItem("accept")
-    exception = ChoiceItem("exception")
-    cancel = ChoiceItem("cancel")
-    # back = ChoiceItem("back")
-    # decline = ChoiceItem("decline")
-
-    unknown = ChoiceItem("unknown")
-
-
-class PaymentStatus(DjangoChoices):
-    # not_required = ChoiceItem("not_required", _("Not required"))
+class PaymentStatus(models.TextChoices):
+    # not_required = "not_required", _("Not required")
 
     # in-progress
-    started = ChoiceItem("started", _("Started by user"))
-    processing = ChoiceItem("processing", _("Backend is processing"))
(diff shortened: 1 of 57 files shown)


src/openforms/payments/contrib/ogone/constants.py
```diff
@@ -1,117 +1,100 @@
...
-        "95", "Payment processed by merchant", PaymentStatus.failed
-    )
-    being_processed = StatusChoice("99", "Being processed", PaymentStatus.processing)
 
-    @classmethod
-    def as_payment_status(cls, ogone_status):
-        choice = cls.get_choice(str(ogone_status))
-        return choice.payment_status
+OGONE_TO_PAYMENT_STATUS: dict[str, str] = {
+    OgoneStatus.invalid_or_incomplete: PaymentStatus.failed,
```
SilviaAmAm, line 62: Same question as above about str vs enum

src/openforms/payments/constants.py
```diff
@@ -1,39 +1,43 @@
...
+    failed = "failed", _("Cancelled or failed")
+    completed = "completed", _("Completed by user")
 
     # flow done
-    registered = ChoiceItem("registered", _("Completed and registered"))
+    registered = "registered", _("Completed and registered")
+
+    @classmethod
+    def get_label(cls, value: str) -> str:
+        return dict(cls.choices)[value]
```
SilviaAmAm, line 36: Why this function instead of for example `PaymentStatus.started.label`?

src/stuf/stuf_bg/tests/test_client.py
```diff
@@ -64,7 +64,7 @@ def test_get_address(self, _mock):
                 ),
             )
             self.client.get_values_for_attributes(
-                "999992314", list(FieldChoices.values.keys())
+                "999992314", [c.value for c in FieldChoices]
```
SilviaAmAm, line 67: Could this not be:
```suggestion
                "999992314", FieldChoices.values
```
?

src/openforms/payments/constants.py
```diff
@@ -1,39 +1,43 @@
...
+
 
-    is_final = {
-        failed.value,
-        completed.value,
-        registered.value,
-    }
+PAYMENT_STATUS_FINAL: set[str] = {
+    PaymentStatus.failed,
```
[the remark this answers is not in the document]
sergei-maertens: yeah they should be the string values, so probably I need `PaymentStatus.failed.value` instead. The Django choices abstraction on top of enum is a bit confusing!

src/openforms/payments/constants.py
sergei-maertens: the values are read somewhere dynamically and then the label needs to be obtained for it for audit logging, so we don't know *which* enum is actually used, only the value from the database (if I understood the calling code correctly).

PR #2722, opened 2023-02-16
[#2699] Fix registration of attachments in repeating groups

### LINKED ISSUE #2699: Uploads in repeating group missing from attachments in submission

SilviaAmAm:
Fixes #2699


src/openforms/submissions/models/submission_files.py
```diff
@@ -192,12 +204,17 @@ def informatieobjecttype(self) -> Optional[str]:
...
-                # Use field-specific override
-                if iotype := glom(
-                    component, "registration.informatieobjecttype", default=""
-                ):
-                    return iotype
+            if (
+                component["key"] != self.form_key
+                and component["key"] != self._component_key
+            ):
+                continue
```
sergei-maertens, line 211: I'd like some comments here that explain why this works:

1. For a file component inside a parent (like editgrid) `component["key"]` will match `self.form_key`
2. However, this will not result in information, because the `registration.informatieobjecttype` is not stored on the editgrid component
3. the next loop iteration will land on either a non-file component (in which case `component["key"]` matches none) or...
4. it will match with the file field component key via `self._component_key`
5. then because it's actually a file component, the `registration.informatieobjecttype` will be read out

However, I think this is still not 100% correct, consider:

```json
[
    {
        "type": "file",
        "key": "attachment"
    },
    {
        "type": "editgrid",
        "key": "editGrid",
        "components": [
            {
                "type": "file",
                "key": "attachment"
            }
        ]
    }
]
```
when processing an upload for the file component inside the editgroup, you will have a database record with `form_key=editGrid` and `_component_key=attachment`.

So, then when looping over the components to find the config, you will come across the following components: `attachment`, `editGrid`, `editGrid.attachment`. Which means that for this upload, the IOType of the non-editgrid upload field would incorrectly be used.

The opposite also happens - if you swap the order of edit group and top-level file upload, if no IOType configuration is determined, it should pick the global configuration. However, because no match is found, the components loop will actually descend into the editgrid and the nested file upload and pick up that configuration (that's probably a bug that exists independently from #2699).

Maybe we should use `openforms.formio.utils.flatten_by_path` which records the JSON path of the `configuration` tree (as apposed to the data path based on the `component["key"]` values). I believe those paths are more stable than the datapaths that have varying lenghts for editgrids, but the drawback is that form configuration changes break the information for existing uploads :thinking:

sergei-maertens (changes requested):
This PR should go to the stable/2.0.x branch too because of the migration and then I'll have to "forward port" it. Probably the easiest is to create a new branch and cherry pick your existing commits onto that?

SilviaAmAm:
Noticed an additional issue:

in the registration email, it doesn't distinguish in which iteration of the repeating group a file was uploaded:
Screenshot from 2023-02-21 11-22-28
Here I had uploaded only test.txt in the first iteration and only test.pdf in the second, but both files appear in both groups

SilviaAmAm:
Now they are linked to the right iteration. Although, it looks like uploading to a component not in the repeating group but with a key matching the key of a component inside the repeating group doesn't work (Attachment component below)
Screenshot from 2023-02-21 14-12-45

PR #2730, opened 2023-02-17
Decouple admin localisation from public UI

### LINKED ISSUE #2685: Switching language in the frontend, also changes language in the admin

2.1.0-alpha.1

The form language switcher also changes language in the admin.

1. Open admin, observe everything is in NL
2. Start multi-language form, in NL, switch to EN
3. Refresh admin, observe everything is in EN

These should be 2 separate things: 1) Language in frontend; 2) Language in admin

* [ ] Taal keuze dropdown toevoegen aan admin
* [ ] Admin user language preference opslaan
* [x] Admin user language preference detecteren en activeren (mbv middleware?)

sergei-maertens:
Closes #2685

This makes it so the admin uses the browser language preferences instead of the language cookie.

I will create a separate PR/feature so that admin users can manage their (language) preferences.

### DIFF
--- a/src/openforms/translations/middleware.py
+++ b/src/openforms/translations/middleware.py
@@ -1,0 +1,54 @@
+import logging
+
+from django.conf import settings
+from django.http import HttpRequest
+from django.utils import translation
+
+from openforms.typing import RequestHandler
+from openforms.utils.urls import is_admin_request
+
+logger = logging.getLogger(__name__)
+
+
+def _remove_language_cookie_from_request(request: HttpRequest) -> None:
+    """
+    Produce a request instance without cookie-based language information.
+
+    settings.LANGUAGE_COOKIE_NAME
+    """
+    logger.debug(
+        "Removing the language cookie (if present) from request %r",
+        request,
+        extra={"request": request},
+    )
+    cookie_name = settings.LANGUAGE_COOKIE_NAME
+    if cookie_name in request.COOKIES:
+        del request.COOKIES[cookie_name]
+        logger.debug(
+            "Cookie %s was present but is now ignored.",
+            cookie_name,
+            extra={"request": request},
+        )
+
+
+class AdminLocaleMiddleware:
+    """
+    Discard the language cookie for locale-information.
+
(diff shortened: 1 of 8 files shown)


src/openforms/typing.py
```diff
@@ -7,3 +10,8 @@
 JSONObject = Dict[str, JSONValue]
 
 DataMapping = Dict[str, Any]  # key: value pair
+
+
+class RequestHandler(Protocol):
+    def __call__(self, request: HttpRequest) -> HttpResponseBase:  # pragma: no cover
+        ...
```
stevenbal, line 17: Didn't know this worked! Any reason to use this over `pass`?

src/openforms/translations/middleware.py
```diff
@@ -0,0 +1,54 @@
...
+    this applies to the public UI and not the admin. For the admin UI, we either use
+    the browser Accept-Language header or look up user preferences stored in the user
+    account itself.
+    """
+
+    def __init__(self, get_response: RequestHandler):
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest):
+        # we do not language-code prefixed URLs
```
stevenbal, line 48: I'm not entirely sure what is meant by "prefixed URLs"?

EDIT: judging from the changes to `is_admin_request`, I'm assuming prefixed with the admin base path

stevenbal (approved):

src/openforms/typing.py
sergei-maertens: it's convention to use this for typing/overloading :-) see also https://python.land/python-ellipsis

src/openforms/translations/middleware.py
sergei-maertens: django has a [`i18n_patterns`](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#django.conf.urls.i18n.i18n_patterns) which essentially comes down to a `/nl/` or `/en/` prefix to have localized URLs. That's the sort of URLs this refers to :) We don't use them or aren't planning on doing so, so that's left out of scope for this feature.

Issue #2683, opened 2023-02-13
Add services and servicefetchconfigurations to Form export/import


CharString:

sergei-maertens:
Note that this depends on the service/service fetch serializers to exist to be able to export to/from JSON.

CharString:
#2693 adds drf servializers

joeribekker:
Refinement: Decided not to implement this right now because Rotterdam is not sponsoring this at the moment. We do need to resolve any export crashes if they happen and document that service fetch is not exported/imported -> new issue created.

joeribekker:
Refinement: This could be handled by adding export/import of services so they match between environments.

PR #2740, opened 2023-02-21
Allow inline service fetch configuration on form variable serializer

### LINKED ISSUE #2693: Allow inline ServiceFetchConfiguration on the the FormVariable serializer
For easier validation

CharString:
closes #2693

### DIFF
--- a/src/openforms/api/serializers.py
+++ b/src/openforms/api/serializers.py
@@ -1,15 +1,10 @@
-from dataclasses import dataclass
-from functools import total_ordering
-from itertools import groupby
-from typing import Any, Iterable, List, MutableMapping, Type
+from typing import List
 
-from django.core.exceptions import ValidationError as DjangoValidationError
-from django.db.models import Model
+from django.db import transaction
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
-from rest_framework.exceptions import ValidationError
 
 from openforms.api.utils import underscore_to_camel
 
@@ -75,27 +70,6 @@ class ValidationErrorSerializer(ExceptionSerializer):
     invalid_params = FieldValidationErrorSerializer(many=True)
 
 
-@dataclass
-@total_ordering
-class BatchItem:
-    priority: int
-    instance: Model
-    source: str  # index.field_name of the source data
-
-    @property
-    def model(self) -> Type[Model]:
-        return type(self.instance)
-
-    def __post_init__(self):
-        # high priority first, grouped by type
-        self._sorting_key = (-self.priority, self.model.__name__)
-
-    def __lt__(self, other: object) -> bool:
-        if not isinstance(other, BatchItem):  # pragma: no cover
-            raise TypeError(f"Can't compare BatchItem to {type(other)}")
-        return self._sorting_key < other._sorting_key
-
-
 class ListWithChildSerializer(serializers.ListSerializer):
     child_serializer_class = None  # class or dotted import path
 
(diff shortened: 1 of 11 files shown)


src/openforms/api/serializers.py
```diff
@@ -69,6 +75,27 @@ class ValidationErrorSerializer(ExceptionSerializer):
...
+    @property
+    def model(self) -> Type[Model]:
+        return type(self.instance)
+
+    def __post_init__(self):
+        # high priority first, grouped by type
+        self._sorting_key = (-self.priority, self.model.__name__)
+
+    def __lt__(self, other: object) -> bool:
+        if not isinstance(other, BatchItem):  # pragma: no cover
```
CharString, line 94: Not worth a unit test. I guess removing this check is also an option. Raising `AttributeError` when there is no `_sorting_key` works too.

src/openforms/api/serializers.py
```diff
@@ -85,15 +112,79 @@ def get_child_serializer_class(self):
...
+        data: Iterable[MutableMapping[str, Any]],
+    ) -> Iterable[BatchItem]:
+        # Helps create() and validate() deal with nested serializers in O(1) db
+        # queries. i.e. one insert per model type.
+        #
+        # It helps by yielding BatchItems of unsafed Model instances. Nested
+        # models should be safed first, so parent gets saved with the new
+        # foreign keys. The order of BatchItems is defined such that high
+        # priority comes first, and models of the same type are grouped
+        # together for a single bulk_create.
```
CharString, line 126: Is this clear?
I initially had this as recursive function, so children could have lists of children too. But we don't need that (yet/ever?) and this is a lot simpler to follow.

src/openforms/api/serializers.py
```diff
@@ -85,15 +112,79 @@ def get_child_serializer_class(self):
...
+                field = child_fields[field_name]
+                if isinstance(field, serializers.ModelSerializer):
+                    # Turn sub_data into *unsaved* instance
+                    # if only drf Serializers had a "build" method that just
+                    # instantiated without saving
+                    sub_model = field.Meta.model
+                    sub_instance = (
+                        sub_data
+                        if isinstance(sub_data, sub_model)
+                        else sub_model(**sub_data)
```
CharString, line 147: DRF turns the data into instances on the validate pass, so in the create pass, no sub_model call is needed anymore.

src/openforms/api/serializers.py
```diff
@@ -69,6 +75,27 @@ class ValidationErrorSerializer(ExceptionSerializer):
     invalid_params = FieldValidationErrorSerializer(many=True)
 
 
+@dataclass
+@total_ordering
+class BatchItem:
```
CharString, line 80: Maybe `_BatchItem` is better, this should not be used elsewhere, and should be kept as an implementation detail.

src/openforms/api/serializers.py
```diff
@@ -85,15 +112,79 @@ def get_child_serializer_class(self):
...
+            try:
+                item.instance.clean()
+            except DjangoValidationError as e:
+                # reconstruct the full source path
+                raise ValidationError(
+                    {
+                        f"{item.source}.{field}": errors
+                        for field, errors in e.message_dict.items()
+                    }
+                )
```
CharString, line 173: Should we collect all errors? For the whole batch?
This is already collecting all errors from this child item. And it's complex as it is.

src/openforms/api/serializers.py
SilviaAmAm, line 147: I think this would now crash if service_fetch_configuration on a variable is None?

CharString: :+1: It did. I thought that was the same case as not setting it.

CharString:
Needs to solidify to something with #2836 in order to be able to pick up #2683

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -118,6 +135,28 @@ class Meta:
         # applying # this validation on the whole collection.
         validators = []
 
+    def validate_service_fetch_configuration(self, value):
+        if value is None:
+            return value
+        config_instance = None
+        if config_id := value.get("id"):
+            try:
+                config_instance = ServiceFetchConfiguration.objects.get(id=config_id)
```
sergei-maertens, line 144: `ServiceFetchConfiguration.objects.filter(id=config_id).exists()` is more idiomatic (it returns a bool and does a `SELECT 1 FROM` query

src/openforms/variables/validators.py
```diff
@@ -114,5 +120,39 @@ def __call__(self, value: Mapping[str, str] | JSONValue) -> None:
             raise ValidationError(errors)
 
 
+class ServiceFetchConfigurationValidator:
```
sergei-maertens, line 123: I'd break this out into more validators that can be composed:

1. `ServiceFetchMethodValidator`
2. `ServiceFetchMappingExpressionValidator`

so that you can properly use the `code` for the validation error and let the errors be combined by DRF.

https://www.django-rest-framework.org/api-guide/validators/#writing-custom-validators

This should also make them more unit-testable.

sergei-maertens (changes requested):

src/openforms/forms/api/serializers/form_variable.py
CharString: But don't we need the instance to pass to the `ServiceFetchConfigurationSerializer` in line 155?

sergei-maertens: yes, I'm an idiot :grimacing:

sergei-maertens (approved):

PR #2747, opened 2023-02-21
:sparkles: [#2697] Variable injection for service fetch JSON request body 

### LINKED ISSUE #2697: Service fetch with body needs to support template evaluation/variable injection
The (JSON) body should be interpreted as a Django template where the `DataMapping` can be injected into.

This requires JSON/JS escaping on the input variables so that double quotes don't result in a broken body or a body with "script" injection. Possibly we need a separate django template instance for this with different auto-escape semantics/implementation rather than the HTML escaping it's doing out of the box.

stevenbal:
Fixes #2697 

Based on #2746, that should probably be merged first

### DIFF
--- a/src/openforms/submissions/tests/form_logic/test_fetching_form_varaiable_values_from_services.py
+++ b/src/openforms/submissions/tests/form_logic/test_fetching_form_varaiable_values_from_services.py
@@ -386,8 +386,8 @@ class ServiceFetchConfigVariableBindingTests(SimpleTestCase):
         self.assertEqual(request.body, b'{"foo": {"nested": "bar"}}')
 
     @requests_mock.Mocker()
-    def test_it_sends_the_body_as_json_with_variables_escape_quotes(self, m):
-        m.get("https://httpbin.org/anything", json={"foo": "&quot;bar"})
+    def test_it_sends_the_body_as_json_with_variables_escape_characters(self, m):
+        m.get("https://httpbin.org/anything", json={})
 
         var = FormVariableFactory.build(
             service_fetch_configuration=ServiceFetchConfigurationFactory.build(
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -326,6 +326,60 @@ def test_it_sends_the_body_as_json(self, m):
...
+        )
+
+        _ = bind(var, {"some_variable": "bar"})
+        request = m.last_request
+
+        self.assertIn(("Content-Type", "application/json"), request.headers.items())
+        self.assertEqual(request.body, b'{"foo": {"nested": "bar"}}')
+
+    @requests_mock.Mocker()
+    def test_it_sends_the_body_as_json_with_variables_escape_quotes(self, m):
```
stevenbal, line 366: I'm not sure if there are other examples of values that could potentially break the JSON body

src/openforms/submissions/tests/form_logic/test_binding.py
```diff
@@ -326,6 +326,60 @@ def test_it_sends_the_body_as_json(self, m):
...
+
+        var = FormVariableFactory.build(
+            service_fetch_configuration=ServiceFetchConfigurationFactory.build(
+                service=self.service,
+                path="anything",
+                body={"foo": "{{some_variable}}"},
+            )
+        )
+
+        _ = bind(var, {"some_variable": '"bar'})
```
stevenbal, line 377: I'm not sure if there are other examples of values that could potentially break the JSON body

src/openforms/submissions/tests/form_logic/test_fetching_form_varaiable_values_from_services.py
```diff
@@ -349,6 +349,60 @@ def test_it_sends_the_body_as_json(self, m):
...
+                path="anything",
+                body={"foo": "{{some_variable}}"},
+            )
+        )
+
+        _ = perform_service_fetch(var, {"some_variable": '"bar'})
+        request = m.last_request
+
+        self.assertIn(("Content-Type", "application/json"), request.headers.items())
+        self.assertEqual(request.body, b'{"foo": "&quot;bar"}')
```
CharString, line 404:
```suggestion
        self.assertEqual(request.body, rb'{"foo": "\"bar"}')
```
Maybe check the other values that need to be escaped too:
https://www.json.org/img/string.png

src/openforms/variables/models.py
```diff
@@ -15,6 +15,25 @@
...
+        else:
+            interpolated[key] = (
+                render_from_string(value, context, backend=sandbox_backend)
+                .strip()
+                .encode("utf-8")
+                .decode("latin1")
+            )
+    return interpolated
+
+
```
CharString, line 36: Does this handle bodies like this?
```
{"naughty_list": ["de broer van Henk de Vries", "{{some_variable}}"]}
```

CharString (changes requested):

sergei-maertens (changes requested):

joeribekker:
Closed due to refinement.

PR #2767, opened 2023-02-24
Document additional admin components

sergei-maertens:
& separate standard + custom components a bit better.

Logical grouping:

image

### DIFF
--- a/src/openforms/js/components/admin/forms/FormRow.stories.mdx
+++ b/src/openforms/js/components/admin/forms/FormRow.stories.mdx
@@ -5,11 +5,11 @@ import Fieldset from './Fieldset';
 import FormRow from './FormRow';
 import {TextInput} from './Inputs';
 
-<Meta title="Django admin/FormRow" component={FormRow} />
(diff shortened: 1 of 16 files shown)


src/openforms/js/components/admin/forms/FormRow.stories.mdx
```diff
@@ -19,14 +19,43 @@ export const Template = args => (
...
+export const FormDecorator = Story => (
+  <form>
+    <Fieldset>
+      <Story />
+    </Fieldset>
+  </form>
+);
+
+<Canvas>
+  <Story name="Multiple fields" decorators={[FormDecorator]} args={{}}>
```
SilviaAmAm, line 41: I think the fact that the single field is not wrapped in the fieldset, while the multiple fields story is wrapped, makes it look a bit inconsistent?
Screenshot from 2023-02-28 10-42-40

sergei-maertens: right - I added the fieldset wrapper to get the Django styling to apply, but completely looked over the misaligned single field errors! I'll add the decorator too.

PR #2772, opened 2023-02-24
[#2096] Prevent submission of form with blocked steps

### LINKED ISSUE #2096: As a user i can bypass "block continuing to next step"

latest

In cases where a user changes their mind and go back a step and if this step triggers the logic that disables the continuation of the form they are still able to click the final step  of the form (overzicht) and complete the form.

This user has previously selected: nee, but changed their mind and returned to the step and selects: ja
image
The next step is now disabled. But this does't prevent them from clicking "overzicht" and submitting the old data:
image

When the next step is disabled, the submit and overzicht steps should also be disabled.

SilviaAmAm:
Fixes #2096 

TODO:
Frontend


src/openforms/submissions/api/validation.py
```diff
@@ -81,6 +90,12 @@ def get_submission_completion_serializer(
             "incomplete_steps": incomplete_steps,
             "submission_allowed": submission.form.submission_allowed,
             "privacy_policy_accepted": request.data.get("privacy_policy_accepted"),
+            "contains_blocked_steps": any(
+                [
+                    not submission_step.can_submit
+                    for submission_step in state.submission_steps
+                ]
+            ),
```
CharString, line 98:
```suggestion
            "contains_blocked_steps": any(
                    not submission_step.can_submit
                    for submission_step in state.submission_steps
            ),
```
not sure what black says about the formatting, but the list creation is needless.

src/openforms/submissions/api/validation.py
```diff
@@ -49,6 +50,14 @@ def validate_submission_allowed(self, submission_allowed):
...
             )
 
+    def validate_contains_blocked_steps(self, value):
+        if value:
+            raise serializers.ValidationError(
+                _(
+                    "Submission of this form is not allowed due to the answers submitted in a step."
+                )
+            )
+
```
CharString, line 60: This can't be triggered by accident, right? Because of
https://github.com/open-formulieren/open-forms-sdk/pull/370
Should this raise `SuspiciousOperation`

CharString (changes requested):

src/openforms/submissions/api/validation.py
SilviaAmAm: It shouldn't be, since if the user navigates to the overview (even by changing the URL) the form wouldn't have persisted the step data that causes the form to prevent continuing to the next step. So the user would have to manually make a PUT to persist the step data and then navigate to the overview to submit the form

CharString (approved):
Edit: I missed a comment.

src/openforms/submissions/api/validation.py
sergei-maertens: I interpreted this as a safeguard indeed - an HTTP 400 in this case is better than HTTP 500 because the generic error handling in the SDK *should* display more context so the end-user can contact the municipality with actual information rather than a generic error that's hard to trace.

sergei-maertens (approved):

Issue #2775, opened 2023-02-27
API for retrieving the forms that the user has yet to complete


alextreme:
### Thema / Theme

API

### Omschrijving / Description

Dimpact / @robwagelaar has requested a functionality to show in OIP which not-yet-completed forms a user can continue with, as an expansion of the MijnAanvragen/MijnZaken module.

In cooperation with Joeri I wrote up the following API spec, this to allow external systems to make use of the OpenFormulieren or eSuite API to determine which forms a (with DigiD logged-in) user can continue with:

https://redocly.github.io/redoc/?url=https://gist.githubusercontent.com/alextreme/61dbd2605ea7f714744d6a36f786dd8e/raw/209538fb28617c0ae3d9ab2a9046d9876260e0e1/FormulierenAPI&nocors

https://taiga.maykinmedia.nl/project/open-inwoner/us/975?milestone=445

Atos has developed a proof-of-concept API and an initial integration is being worked on in OIP. I'd like to request the same API to become available from Open Formulieren.

### Added value / Toegevoegde waarde

Het tonen van de openstaande formulieren waar de inwoner met Open Formulieren is gestart in de Mijn-omgeving

### Aanvullende opmerkingen / Additional context

Eerste OF PoC mag ontwikkeld worden door @Bartvaderkin , na Joeri's vakantie kunnen we de integratie verder bespreken.

sergei-maertens:
See also #2301 and #2785

sergei-maertens:
This is blocked by #2301 - but there is no planning or prio for this at the moment.

sergei-maertens:
Refinement: unclear if/what needs to be done by the development team.

Functionally: "Mijn"-environments like OIP need to be able to query for a given user (BSN, KVK) all the "open"/unfinished form submissions to be able to continue them from the portal. This includes:

* the form submissions that have been explicitly paused (which have hashed authentication details)
* the form submissions that have been abonded (e.g. if their session timed out after 15 minutes, which don't have hashed auth details)

What this means for Open Forms is that:

* we need an endpoint to query unfinished submissions, which takes some kind of identifying property for the user (citizen or company)
    - bsn
    - kvk
    - kvk + vestigingsnummer (branch number)
    - OIDC user ID (UUID, e-mail address...)
    - whatever else we support in `AuthAttribute` :-)
[...]

PR #2782, opened 2023-02-27
[#2775] First pass at implementing suspended-submission endpoint

Bartvaderkin:
PR for POC exploration and discussion.


src/openforms/submissions/api/views.py
```diff
@@ -106,3 +109,32 @@ def get(self, request, *args, **kwargs):
...
+        qs = super().get_queryset()
+        bsn = self.request.query_params.get("bsn")
+        if not bsn:
+            return qs.none()
+        else:
+            qs = qs.filter(
+                auth_info__value=bsn,
+                auth_info__attribute=AuthAttribute.bsn,
+                auth_info__attribute_hashed=False,
+            )
```
sergei-maertens, line 139: you need to compare/filter by hash since suspended submissions will have these values hashed.

src/openforms/submissions/api/views.py
```diff
@@ -106,3 +109,32 @@ def get(self, request, *args, **kwargs):
...
+    serializer_class = SuspendedSubmissionSerializer
+    pagination_class = PageNumberPagination
+
+    queryset = Submission.objects.filter(
+        completed_on__isnull=True,
+        suspended_on__isnull=False,
+    )
+
+    # TODO authentication? permissions?
+    # TODO we need something here to verify current requests is allowed to access this BSN
```
sergei-maertens, line 127: from the API spec I saw the external API is expected to use TokenAuthentication, so we could connect this with a (custom) user permission in the backend.

Since it's an API client and not a real user, afaik there's no way to limit which BSNs are valid. Note that the end-user needs to re-authenticate for the form anyway, so even with a mismatching BSN this will be caught during the auth fase.

src/openforms/submissions/tests/test_submission_suspension.py
```diff
@@ -218,3 +219,98 @@ def test_resume_url_does_not_work_after_submission_has_been_completed(self):
...
+                auth_info__value=bsn,
+                auth_info__attribute=AuthAttribute.bsn,
+            ),
+            SubmissionFactory.create(
+                form__name="hashed BSN",
+                suspended_on=None,
+                completed_on=None,
+                auth_info__value="<!>hashed-bsn<!>",
+                auth_info__attribute=AuthAttribute.bsn,
+                auth_info__attribute_hashed=True,
```
sergei-maertens, line 290: shortcut is available via `SubmissionFactory.create(auth_info__with_hashed_identifying_attributes=True)` trait

joeribekker:
Refinement: This hangs on #2775 requiring a functional design for encryption/hashing.

joeribekker:
Refinmenent: Closing since no-one asked for this and it requires work outside the scope of this PR.

Bartvaderkin:
Open Inwoner asked for this, see @alextreme's ticket https://github.com/open-formulieren/open-forms/issues/2775

joeribekker:
@Bartvaderkin Open Inwoner is not a client, it's a project. I need someone with budget to sign off on it to pick it up further.

PR #2783, opened 2023-02-28
[#2781] Fix hiding nested components

### LINKED ISSUE #2781: Hiding nested components doesn't clear their data

2.0.x, 2.1.x

If there is a component with key `nested.textfield`, if it is hidden by backend logic then its value is not cleared.

Create a form with:
- A radio component with values `a` and `b`
- A text field with key `nested.field`
- A logic rule that says: if radio is equal to `a` then change `nested.field` to hidden

Fill the form:
- Fill in "test" in the textfield
- Check the radio `a`

See how in the backend the submission still contains the value "test" for the text field

The variable for the textfield should have an empty value

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2781 

Needs backporting to stable 2.0.x and 1.1.x

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from django.utils.functional import empty
 
 import elasticapm
+from glom import assign, glom
 
 from openforms.formio.service import (
     get_dynamic_configuration,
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):

sergei-maertens:
Backport to 2.0: d9f4c671e94662f85c7a30cc76b39d0f9d9d9043

For the 1.1.x series there's no backport, the original fix wasn't backported either, see https://github.com/open-formulieren/open-forms/issues/1755#issuecomment-1236817987

PR #2800, opened 2023-03-01
[#2798] Update options values automatically

### LINKED ISSUE #2798: Auto-generation of Select (and radio/checkbox) values from labels is broken

latest

When you add a Select component to the form, you can define the values for the component in a tabular form.

This used to generate a (slugified) value from the label entered, reducing the amount of typing needed by form designers. This no longer works - the value column stays empty.

This probably applies to other values/data components too like the radio/checkboxes components.

1. Go to the form designer in the admin
2. Drag a "Keuzelijst" component into the dropzone
3. Fill out a label "Optie 1" for the first choice
4. Observe that the 'value' column stays empty

Values are derived from labels as default suggestion.

### Screen resolution

None

### Device

None

### OS

None

### Browser

Chrome

SilviaAmAm:
Fixes #2798

### DIFF
--- a/src/openforms/forms/tests/e2e_tests/test_form_designer.py
+++ b/src/openforms/forms/tests/e2e_tests/test_form_designer.py
@@ -351,6 +351,51 @@ class FormDesignerComponentTranslationTests(E2ETestCase):
             # the modal should close
             await expect(page.locator("css=.formio-dialog-content")).to_be_hidden()
 
+    async def test_key_automatically_generated_for_select_options(self):
+        @sync_to_async
+        def setUpTestData():
+            # set up a form
+            form = FormFactory.create(
+                name="Playwright test",
+                name_nl="Playwright test",
+                generate_minimal_setup=True,
+                formstep__form_definition__name_nl="Playwright test",
+                formstep__form_definition__configuration={
+                    "components": [],
+                },
+            )
+            return form
+
+        await create_superuser()
+        form = await setUpTestData()
+        admin_url = str(
+            furl(self.live_server_url)
+            / reverse("admin:forms_form_change", args=(form.pk,))
+        )
+
+        async with browser_page() as page:
+            await self._admin_login(page)
+            await page.goto(str(admin_url))
+            await page.get_by_role("tab", name="Steps and fields").click()
+
+            # Drag and drop a component
+            await page.get_by_text("Keuzelijst").hover()
+            await page.mouse.down()
+            await page.locator('css=[ref="-container"]').hover()
+            await page.mouse.up()
+
+            # Check that the modal is open
+            await expect(page.locator("css=.formio-dialog-content")).to_have_count(1)
+
+            # Update the label
+            value_label_input = page.locator('css=[name="data[data.values][0][label]"]')
+            await value_label_input.click()
+            await value_label_input.fill("Test")
+
+            # Check that the key has been updated
+            value_key_input = page.locator('css=[name="data[data.values][0][value]"]')
+            await expect(value_key_input).to_have_value("test")
+
 
 class FormDesignerRegressionTests(E2ETestCase):
     async def test_user_defined_variable_boolean_initial_value_false(self):
--- a/src/openforms/forms/tests/e2e_tests/test_form_designer.py
+++ b/src/openforms/forms/tests/e2e_tests/test_form_designer.py
@@ -414,10 +414,7 @@ class FormDesignerComponentTranslationTests(E2ETestCase):
             await page.get_by_role("tab", name="Steps and fields").click()
 
             # Drag and drop a component
-            await page.get_by_text("Keuzelijst").hover()
-            await page.mouse.down()
-            await page.locator('css=[ref="-container"]').hover()
-            await page.mouse.up()
+            await drag_and_drop_component(page, "Keuzelijst")
 
             # Check that the modal is open
             await expect(page.locator("css=.formio-dialog-content")).to_be_visible()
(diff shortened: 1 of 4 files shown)


src/openforms/js/components/formio_builder/builder.js
```diff
@@ -368,15 +368,18 @@ const FormIOBuilder = ({
...
+          data: {
+            ...newComponentConfiguration,
+            openForms: {
+              ...(newComponentConfiguration?.openForms || {}),
+              translations: translations,
+            },
           },
         },
-      };
+        {fromSubmission: false}
```
SilviaAmAm, line 381: It seems that here https://github.com/formio/formio.js/blob/v4.13.13/src/components/_classes/component/Component.js#L2632
formio does this check on `fromSubmission`. Before we added the custom `onChange` for the translations, the `fromSubmission` flag was undefined. The place where this flag gets set is the `setSubmission` function https://github.com/formio/formio.js/blob/v4.13.13/src/Webform.js#L783. 
Not really sure yet what this `fromSubmission` function means, but this seems to solve our problem

sergei-maertens: I think it tracks if there was existing data that was loaded (like when you resume a paused form in our SDK), so then it won't override the submission data with calculated data because it was persisted already at some point.

I can't find any references to it in the documentation, only in the source code, but it seems that setting it to false here would be correct usage because we're doing programmatic stuff behind the scenes.

SilviaAmAm: The problem with setting the flag here to false is that then I think the component is not updated, so the translations are no longer updated.

The `setValue` works better than the `setSubmission`, but any new translations are not added. I think it may be a matter of findingin which order the different updates have to happen so that the component is rendered with all the updates :see_no_evil:

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -351,6 +351,48 @@ async def test_enable_translations_and_create_new_step(self):
...
+        async with browser_page() as page:
+            await self._admin_login(page)
+            await page.goto(str(admin_url))
+            await page.get_by_role("tab", name="Steps and fields").click()
+
+            # Drag and drop a component
+            await drag_and_drop_component(page, "Keuzelijst")
+
+            # Check that the modal is open
+            await expect(page.locator("css=.formio-dialog-content")).to_have_count(1)
```
sergei-maertens, line 385: I found out about `.to_be_visible()` which is semantically closer than `to_have_count`, but I'll clean that up in a separate PR for all the existing tests I think :thinking:

sergei-maertens (commented):
Couple findings while testing :(

* ~~I added a test that shows there's a regression in tracking the translations. We kept a previousLiteralsMap so that when we detect changes in the component options, the translation is not lost. Now it is.~~ fixed in d669333bfee7ad767af7eb852359f84664f66bef
* With a textfield component I'm not getting infinite loops, but I am getting them with a Select component that has the options sourced from a variable.

I will push my console logging so this is observable.

src/openforms/js/components/form/edit/options.js
```diff
@@ -208,6 +208,7 @@ const OPTIONS_CHOICES = [
         },
       },
     ],
+    clearOnHide: true,
```
sergei-maertens, line 211: I don't believe this contributed anything to breaking the infinite change event loop, but it seems more correct either way.

src/openforms/js/components/formio_builder/builder.js
```diff
@@ -411,12 +391,43 @@ const FormIOBuilder = ({
...
+    const {instance} = changed;
+
+    if (!flags.fromSubmission) {
+      instance.root.triggerChange(flags, changed, modifiedByHuman);
+    }
+
+    // we only need to sync translations when an actual human changed some literal
+    if (!modifiedByHuman) {
+      return;
+    }
```
sergei-maertens, line 409: partially fixes some of the infinite change event triggers, but in generally this seems a bit of a safer approach - we only translate literals that are entered by the form designer, and don't even bother with automatically updated 'literals' since those aren't translated anyway.

Aborting here reduces the risk of change trigger cascades.

src/openforms/js/components/formio_builder/builder.js
```diff
@@ -450,7 +461,6 @@ const FormIOBuilder = ({
       form={formRef.current}
       options={builderOptions}
       onChange={formSchema => onChange(cloneDeep(formSchema))}
-      onUpdateComponent={() => (previousLiteralsRef.current = {})}
```
sergei-maertens, line 453: this fires more often than expected/anticipated and it seems to work without :shrug:

src/openforms/js/components/formio_builder/translation.js
```diff
@@ -241,7 +242,9 @@ export const handleComponentValueLiterals = (
       return null;
   }
   let translations = {};
-  values.forEach(({label}, index) => {
+  const nonEmptyValues = values.filter(item => !isEmpty(item));
+  if (!nonEmptyValues.length) return null;
+  nonEmptyValues.forEach(({label = ''}, index) => {
```
sergei-maertens, line 247: this is the crucial fix for the infinite loops - looks like formio for its data values has a default value of `[{}`], so the `label` inside the object is undefined. It uses that even if we have the data source set to variable rather than literal options.

It looks like the `undefined` label then caused an infinite cycle of change triggers. Filtering out these empty objects (there's no literal to translate anyway) stopped the infinite loops.

src/openforms/js/components/formio_builder/builder.js
sergei-maertens: this broke the translations update completely again so I removed it :grimacing:

sergei-maertens (approved):

PR #2807, opened 2023-03-02
Add Service API endpoint

### LINKED ISSUE #2680: Add an API endpoint to expose zgw_consumers.Service
Add an API endpoint to expose zgw_consumers.Service, read-only, fields api_root, label and api_type. Only form-designers have access to this (question should we create a special permission for this)

- [x] define ModelSerializer
- [x] Decide on URL
- [x] Decide on permission

CharString:
Closes #2680


src/openapi.yaml
```diff
@@ -3311,6 +3311,32 @@ paths:
               $ref: '#/components/headers/X-Is-Form-Designer'
             Content-Language:
               $ref: '#/components/headers/Content-Language'
+  /api/v2/services/:
```
CharString, line 3314: I considered adding it under the config. Because that's roughly where it lives in the admin interface. But moved it here: flat is better than nested.


TODO this endpoint should be flagged private/internal/subject to change. But I couldn't find out how to do that.

sergei-maertens: there is a `openforms.api.utils.mark_experimental` decorator :-)

src/openforms/services/api/views.py
```diff
@@ -0,0 +1,27 @@
...
+    class Meta:
+        model = Service
+        fields = (
+            "label",
+            "api_root",
+            "api_type",
+        )
+
+
+class ServiceListView(views.APIView, ListMixin):
```
sergei-maertens, line 17: if possible, we prefer using viewsets:

```py
from rest_framework.viewsets import GenericViewSet
from rest_framework import mixins

class ServiceViewSet(mixins.ListModelMixin, GenericViewSet):
    ...
```
docs: https://www.django-rest-framework.org/api-guide/viewsets/

note that mixins should always appear left of the base classes due to how Python's MRO works. Please also stick to the patterns in other apps: a `api/viewsets.py` file and a `api/serializers.py` file + document the view/viewset using `drf_spectacular.utils.extend_schema(_view)`

src/openforms/services/api/viewsets.py
```diff
@@ -0,0 +1,32 @@
...
+    list=extend_schema(
+        summary=_("List available services"),
+        description=_(
+            "Return a list of available (JSON) services/registrations configured "
+            "in the backend.\n\n"
+            "Note that this endpoint is **EXPERIMENTAL**."
+        ),
+    )
+)
+@mark_experimental
```
sergei-maertens, line 22: @CharString I've marked the entire viewset as experimental since we might even still tweak the endpoints/filters etc.

sergei-maertens (approved):

sergei-maertens:
Approved, but will be merged after the release branch for 2.1.0-rc.0 is merged.

PR #2811, opened 2023-03-02
Fix date/time min/max validation configuration not saving correctly

### LINKED ISSUE #2769: max/min dates/months/years are ignored in validation tab

sergei-maertens:
Fixes #2769

The layout component keys are not used as intermediate steps like with data containers, so the full namespaced path needs to be provided to the keys that actually write some configuration.

### DIFF
--- a/src/openforms/js/components/form/edit/date-edit-form.js
+++ b/src/openforms/js/components/form/edit/date-edit-form.js
@@ -14,14 +14,15 @@ const MESSAGES = {
 };
 
 const getMinMaxValidationEditForm = (fieldName, componentType) => {
+  const namespace = `openForms.${fieldName}`;
   const panel = {
     type: 'panel',
-    key: `openForms.${fieldName}`,
+    key: `${fieldName}ValidationPanel`,
     title: MESSAGES.panel[fieldName],
     components: [
       {
         type: 'select',
-        key: `openForms.${fieldName}.mode`,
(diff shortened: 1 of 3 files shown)


src/openforms/js/components/form/edit/date-edit-form.js
```diff
@@ -62,34 +63,31 @@ const getMinMaxValidationEditForm = (fieldName, componentType) => {
           ) {
             return false;
           }
-          const upperCasedFieldName = fieldName.charAt(0).toUpperCase() + fieldName.substring(1);
-          const attrName = `enable${upperCasedFieldName}Input`;
-          const isFixedValue = data?.openForms[fieldName]?.mode === 'fixedValue';
-          return isFixedValue && !data[attrName];
```
sergei-maertens, line 68: @SilviaAmAm you seem to have touched this most recently, I grepped the codebase for `enable.*input` (case insensitive) and couldn't find any uses and it seems to work without. Do you remember if/why this was needed? From the git history it seemed like a leftover from my experiments/prototyping that should've been cleaned up before.

SilviaAmAm: No, i don't remember why this was needed :S

sergei-maertens: good riddance then :)

pi-sigma (approved):

PR #2815, opened 2023-03-03
:sparkles: Add management command to check component usage

sergei-maertens:
We can use this on instances to quickly get some information about how often certain component types are used.


CharString (approved):
Looks useful. Why not run the types through `collections.Counter`? It :heart: its `most_common` function.

sergei-maertens:
> Looks useful. Why not run the types through `collections.Counter`? It heart its `most_common` function.

I always forget it exists - could be useful in the future, but for now the simplicity of this code probably makes it easiest to follow. I have some PTSD/flashbacks to crazy collections/itertools code from another project :sweat_smile:

PR #2816, opened 2023-03-03
Improve password field

SilviaAmAm:
Frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/375


sergei-maertens:
https://github.com/open-formulieren/open-forms/pull/2886 moves the password field to the "deprecated" group as checking with customers reveals that one uses this in production nor have any concrete examples why/what they would use it for. It will be removed at some point. This also means that we should not waste any resources on this component for could-be's or maybes.

PR #2817, opened 2023-03-03
[#2769] add end-to-end test for datetime validation

### LINKED ISSUE #2769: max/min dates/months/years are ignored in validation tab

pi-sigma:
Closes #2769 (complement to #2811)


sergei-maertens:
This is being included in #2811, thanks!

PR #2818, opened 2023-03-03
[#2804] Static vars in submission confirmation  template

### LINKED ISSUE #2804: Using empty auth_kvk / auth_bsn as conditionals in the confirmation page/mail doesn't work

2.0.3

Setting a empty string as conditional leads to:
image

This doesn't work: 
Beste {% if auth_bsn != "" %} {{ achternaam }} {% endif %}{% if auth_kvk !=  "" %} {{ achternaam1 }} {% endif %},

For taking the values from their respective fields ('achternaam' is the ID of the last name field in the BSN version, 'achternaam1' that of the KvK-version). 

However, the same effect can be achieved with this workaround:
Beste {% if achternaam != null %} {{ achternaam }} {% endif %}{% if achternaam1 != null %} {{ achternaam1 }} {% endif %},

where the conditions are taken from prefill values that only get filled when logged in with kvk or bsn. So in all, its not a big issue; but would be more elegant if it could be based on the login options directly.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2804

### DIFF
--- a/src/openforms/submissions/models/submission.py
+++ b/src/openforms/submissions/models/submission.py
@@ -448,7 +448,7 @@ class Submission(models.Model):
         return state
 
     def render_confirmation_page(self) -> str:
-        from openforms.variables.service import get_static_variables
+        from openforms.variables.utils import get_variables_for_context
 
         if not (template := self.form.submission_confirmation_template):
             config = GlobalConfiguration.get_solo()
@@ -461,11 +461,7 @@ class Submission(models.Model):
             "_submission": self,
             "_form": self.form,  # should be the same as self.form
             "public_reference": self.public_registration_reference,
-            **self.data,
-            **{
-                variable.key: variable.initial_value
-                for variable in get_static_variables(submission=self)
-            },
+            **get_variables_for_context(submission=self),
         }
         return render_from_string(template, context_data, backend=openforms_backend)
 
--- a/src/openforms/submissions/models/submission.py
+++ b/src/openforms/submissions/models/submission.py
@@ -448,6 +448,8 @@ class Submission(models.Model):
         return state
 
     def render_confirmation_page(self) -> str:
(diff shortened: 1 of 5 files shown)


src/openforms/submissions/models/submission.py
```diff
@@ -460,6 +462,10 @@ def render_confirmation_page(self) -> str:
             "_form": self.form,  # should be the same as self.form
             "public_reference": self.public_registration_reference,
             **self.data,
+            **{
+                variable.key: variable.initial_value
+                for variable in get_static_variables(submission=self)
+            },
```
sergei-maertens, line 468: note that the order matters here! now the static variabels overwrite any variables possibly added by form fields. This shouldn't happen due to the unique key validation and checks that you're not using static variables, but it is possible new static variables in the future cause problems with existing forms.

I'm also surprised this happened in the first place - clearly we're building template context differently in different places. So I'd rather see a refactor/cleanup that defines a function `get_variables_for_context(submission)` somewhere that encapsulates this for anything using the django template language for rendering (confirmation text, confirmation e-mail, registration e-mail, dynamic formio labels/etc.).

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens (approved):

sergei-maertens:
Backports

* `stable/2.0.x`: 84f06ce5bb0bceda779522c44604d1880ac4a01a and 93c53af65e34fb4647a4976ba551fe3a5693c6d1
* `stable/2.1.x`: 503780c855eb7b6554acaa02260f254cd12ca6f9, 709a40c20a3a2c4faca2df2873dd41bf615fc0cd and 4f1e444135066f0527ad531066065175851961ab

PR #2824, opened 2023-03-04
Use design token editor in admin

### LINKED ISSUE #2957: Add design editor for design-tokens
Fixed by #2824 

* [x] Add description to field on what it does (for now, only the curated set is fine)
* [x] Adding a searchfield (that also searches in the description)
* [x] Adding categories or curated set, so users can easily see the things they most likely want to change

estimated in the order of a week.

sergei-maertens:
Closes #2957 

The JSON field in the admin is now enhanced with a button to open a user-friendlier editor in a modal.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "UNLICENSED",
       "dependencies": {
         "@fortawesome/fontawesome-free": "^6.1.1",
-        "@open-formulieren/design-tokens": "^0.23.0",
+        "@open-formulieren/design-tokens": "^0.25.0",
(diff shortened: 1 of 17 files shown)


joeribekker:
Refinement: Discussed the benefits of this (no JSON errors by users, more intuitive UI). Discussed things still to do:

* [ ] Add description to field on what it does (for now, only the curated set is fine)
* [ ] Adding a searchfield (that also searches in the description)
* [ ] Adding categories or curated set, so users can easily see the things they most likely want to change

estimated in the order of a week.

SilviaAmAm (approved):

PR #2825, opened 2023-03-06
Reorganize and document modals

sergei-maertens:
I noticed some styling issues with the (form) modals, so decided to document them in SB first before we tackle those issues.

Modals are particularly hard to document in storybook, there are a number of bugs where the story/canvas is not unmounted, so open modals then overlay the docs page and there's no way to get rid of them.

My workaround is to set a portal selector node that mounts it into the `#root` element which gets hidden in docs view mode, which in turns hides the overlay too. Other attempts were disabling the docs page and forcing story view, but I couldn't get that to work properly.


src/openforms/js/components/admin/modals/FormModal.js
```diff
@@ -10,12 +10,14 @@ const FormModal = ({
   onFormSubmit,
   children,
   extraModifiers = [],
+  ...props
```
CharString, line 13: How does this work with the defined PropTypes?
Any undefined proptype behaves like an optional `PropTypes.any`?

.storybook/main.js
```diff
@@ -63,9 +62,10 @@ module.exports = {
               // sourceMap: argv.sourcemap,
             },
           },
-        ],
-      },
-    ]);
+        ].filter(Boolean),
```
CharString, line 65: Clever, I like it a lot. Maybe a ternary on `isEnvProduction` is clearer, though. Both for us and the JIT of the VM. :)

.storybook/main.js
```diff
@@ -63,9 +62,10 @@ module.exports = {
...
+        sideEffects: true,
```
CharString, line 66: This is because we are setting the innerHTML?

I had a look for other workarounds in [other react modal storybooks](https://storybook.js.org/showcase/tag/modal/react). Most are just broken. But the [carbon (IBM's design system) modals](https://react.carbondesignsystem.com/?path=/story/components-composedmodal--default) seem to work around it with some sort of manager:

https://github.com/carbon-design-system/carbon/tree/36685a0be5e6/packages/react/src/components/ComposedModal

I don't completely understand it, but maybe it's useful.

CharString (changes requested):
The ternary...
The Carbon thing may be creep. I don't know. Maybe you read it and understand it better. Looks like a way to keep our book side effect free.

src/openforms/js/components/admin/modals/FormModal.js
sergei-maertens: yeah I think it just does nothing about it, so won't complain or validate anything.

.storybook/main.js
sergei-maertens: I've taken it from CRA actually - it's prevalent in their webpack.config.js

.storybook/main.js
sergei-maertens: taken it from CRA sass setup, but I think it might not be needed. It was a "throw stuff at the wall and see what sticks" thing. I'll revert it and see if it breaks anything.

I'll take a look at the Carbon setup.

sergei-maertens: The Twilio one has the same issue at least :stuck_out_tongue: 

image

my eyes...

PR #2829, opened 2023-03-06
[#2822] fix form variable datatype info

### LINKED ISSUE #2822: Fix form variable data type information
There's been some ping-ponging with date/datetime components and the data types of the underlying variables.

Sentry 321980 occurred because a Date component was trying to parse its value as a `datetime` type.

-> create a data migration, loop over all components in every form and check all the `type: "date"` component variables that they have the correct type.

workaround: editing the form in the admin and saving it again, this re-creates the variables with the correct type.

pi-sigma:
Fixes #2822


src/openforms/forms/migrations/0072_check_form_variable_datatype.py
```diff
@@ -0,0 +1,36 @@
...
+        data_type=FormVariableDataTypes.datetime
+    )
+
+    form_variables_to_update = []
+    for form_var in form_variables:
+        updated_form_var = False
+
+        form_def = form_var.form_definition
+        for comp in iter_components(configuration=form_def.configuration):
+            if comp["key"] == "date":
```
sergei-maertens, line 21: this does not seem to correctly correlate form variables and the components from the form definition within the same form.

Try setting up test data like this:

1. form 1, has a variable with `key=field1`, `type=datetime`
2. form definition 1, used in form 1, has a component `{"type": "datetime", key: "field1"}`
3. form 2, has a variable with `key=field1`, `type=datetime`
4. form definition 2, used in form 2, has a component `{"type": "date", key: "field1"}`

After migrating, variable one should still be type `datetime`, while variable 2 should be type `date`

sergei-maertens: This works okay because of the `FormVariable.form_definition` field, however, what will make it crash is setting up a user defined variable (non-component) with `type="datetime"`, since `form_def.configuration` will throw `AttributeError` (`form_def` will be None` in that case.

src/openforms/forms/tests/test_data_migrations.py
```diff
@@ -73,3 +74,60 @@ def test_open_when_empty_is_false(self):
...
+            key="unrelated",
+            form=form,
+            data_type=FormVariableDataTypes.datetime,
+        )
+
+    def test_form_variable_datatype(self):
+        FormVariable = self.apps.get_model("forms", "FormVariable")
+        form_var1 = FormVariable.objects.filter(key="field1").first()
+        form_var2 = FormVariable.objects.filter(key="field2").first()
+        form_var3 = FormVariable.objects.filter(key="unrelated").first()
```
sergei-maertens, line 124: for the future - since there's only one expected match, you can do `FormVariable.objects.get(key="...")` instead.

sergei-maertens (approved):

sergei-maertens:
Backported in a1907b5805e98622019bf590cf45a6d145a04a1c

Issue #2834, opened 2023-03-06
API endpoint for ServiceFetchConfiguration


stevenbal:
This endpoint will be used by the ServiceFetchConfiguration modal to select existing configs

TODO:
- [x] Fields:
  - [x] url
  - [x] readable name
- [x] pagination
- [x] Use this API in the modal

PR #2836, opened 2023-03-07
:sparkles: [#2834] API endpoint for ServiceFetchConfigurations 

stevenbal:
Closes https://github.com/open-formulieren/open-forms/issues/2834

~~Requires https://github.com/open-formulieren/open-forms/pull/2690 to be merged first~~


src/openapi.yaml
```diff
@@ -5309,6 +5495,19 @@ paths:
             X-Is-Form-Designer:
               $ref: '#/components/headers/X-Is-Form-Designer'
             Content-Language:
+    ApiTypeEnum:
```
stevenbal, line 5498: Not sure where these came from :thinking:

src/openforms/js/components/admin/form_design/FormLogic.js
```diff
@@ -64,8 +65,18 @@ const FormLogic = ({logicRules = [], onChange, onDelete, onAdd}) => {
 
   const backendDataToLoad = [
     {endpoint: SERVICES_ENDPOINT, stateVar: 'services'},
-    // TODO load ServiceFetchConfigurations
-    {endpoint: '/api/v2/service-fetch-configurations', stateVar: 'serviceFetchConfigurations'},
+    {
+      endpoint: SERVICE_FETCH_CONFIG_ENDPOINT,
+      stateVar: 'serviceFetchConfigurations',
+      // TODO endpoint itself should probably do this?
```
CharString, line 71: ...  these `backendDataToLoad` entries are conceptually our own `serviceFetchConfiguration`s. We could apply the same naming and patterns here. Maybe not the same class names (that could be confusing), but the functions and attributes.

CharString (changes requested):
:+1:  "Just" needs to reconcile the conflicts with master. And maybe that bit of API doc polish. I don't know when we will get to taking this out of "experimental".

On that. I like the choice for urls as resource references over magic ids.

src/openforms/services/api/viewsets.py
```diff
@@ -20,7 +20,7 @@
     )
 )
 @mark_experimental
-class ServiceViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+class ServiceViewSet(viewsets.ReadOnlyModelViewSet):
```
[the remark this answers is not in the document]
stevenbal: I think it does the same as having `mixins.ListModelMixin, viewsets.GenericViewSet`, except it also adds `mixins.RetrieveModelMixin` (which is required for hyperlinked serializers)

src/openforms/js/components/admin/form_design/FormLogic.js
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
stevenbal: seems that is isn't, I'll remove it

src/openforms/js/components/admin/form_design/FormLogic.js
stevenbal: I think it might actually be alright to keep this reshaping here, because the only reason this has to happen is to handle the data in the underlying React components. It would be strange to have the API return headers/queryParams as lists of tuples, while in the database it is still stored as json blobs, imo.

src/openapi.yaml
stevenbal: Seems to have been resolved after rebasing on master and regenerating the schema :man_shrugging:

CharString (approved):
IIRC your `ServiceFetchConfigurationSerializer` had `url` instead of `id`. I liked yours better than mine. Did it break the `ListWithChildSerializer` of FormVariables, or is this just the simplest thing that works?

CharString:
> IIRC your `ServiceFetchConfigurationSerializer` had `url` instead of `id`. I liked yours better than mine. Did it break the `ListWithChildSerializer` of FormVariables, or is this just the simplest thing that works?

It did:
```diff
+@mark_experimental
+class ServiceFetchConfigurationSerializer(serializers.HyperlinkedModelSerializer):
+    class Meta:
+        model = ServiceFetchConfiguration
+        fields = (
+            "url",
+            "name",
+            "service",
+            "path",
+            "method",
+            "headers",
+            "query_params",
+            "body",
+            "data_mapping_type",
+            "mapping_expression",
+        )
+        extra_kwargs = {
+            "url": {
+                "lookup_field": "pk",
+                "view_name": "api:servicefetchconfiguration-detail",
+            },
+            "service": {
+                "lookup_field": "pk",
+                "view_name": "api:service-detail",
+            },
+        }
```

stevenbal:
> IIRC your ServiceFetchConfigurationSerializer had url instead of id. I liked yours better than mine. Did it break the ListWithChildSerializer of FormVariables, or is this just the simplest thing that works?

It did indeed, I kept it with `ids` because I think that is the simplest way to make it work. If we choose to use URLs, an extra step is required in the serializer `.to_internal_value` to turn the URLs into `ServiceFetchConfiguration` objects again, which seems a bit unnecessary

sergei-maertens:
> > IIRC your ServiceFetchConfigurationSerializer had url instead of id. I liked yours better than mine. Did it break the ListWithChildSerializer of FormVariables, or is this just the simplest thing that works?
> 
> It did indeed, I kept it with `ids` because I think that is the simplest way to make it work. If we choose to use URLs, an extra step is required in the serializer `.to_internal_value` to turn the URLs into `ServiceFetchConfiguration` objects again, which seems a bit unnecessary

This should be done automatically by the `HyperlinkedIdentityField` - that's exactly what serializer fields do, serialize and deserialize between JSON primitives and Django-domain objects like model instances.

src/openforms/variables/views.py
sergei-maertens, line 1: this file is not consistent with other apps - please split it into `views.py` and `viewsets.py` and ensure both are in the `variables/api/` package instead of `variables` package.

src/openforms/variables/models.py
```diff
@@ -119,12 +124,13 @@ def request_arguments(self, context: DataMapping) -> dict:
...
 
-        query_params = self.query_params.format(**escaped_for_query)
-        query_params = (
-            query_params[1:] if query_params.startswith("?") else query_params
+        query_params = urlencode(
+            {
+                param: value.format(**context)
+                for param, value in (self.query_params or {}).items()
+            }
         )
```
sergei-maertens, line 134: afaik we don't even need the `urlencode` because you can pass a dict for `params` to requests, see https://requests.readthedocs.io/en/latest/api/#requests.request

so this would be sufficient:

```py
query_params = {param: value.format(**context) for param, value in (self.query_params or {}).items()}
```

src/openforms/variables/migrations/0004_migrate_query_params.py
```diff
@@ -0,0 +1,37 @@
...
+            config._query_params = {}
+        config.save()
+
+
+def migrate_query_params_to_textfield(apps, _):
+    ServiceFetchConfiguration = apps.get_model("variables", "ServiceFetchConfiguration")
+
+    for config in ServiceFetchConfiguration.objects.all():
+        config.query_params = json.dumps(config._query_params)
+        config.save()
```
sergei-maertens, line 24: this should do the `urlencode` bit instead, I believe, but may not escape the curly braces syntax.

src/openforms/services/api/viewsets.py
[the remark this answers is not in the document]
sergei-maertens: for completeness sake - `viewsets.GenericViewSet` does not expose any actions at all, so before this would only expose the `list` action. `ReadOnlyModelViewSet` exposes the `list` and `retrieve` actions (through their respective mixins).

sergei-maertens (changes requested):

src/openforms/variables/tests/test_validators.py
```diff
@@ -184,3 +185,80 @@ def test_it_invalidates_field_names_invalid_characters(
...
+
+    def test_validate_none_success(self):
+        value = None
+
+        self.validate(value)
+
+    def test_validate_success(self):
+        value = {
+            "foo": ["bar"],
+            "bar": ["baz", "qux"],
```
CharString, line 201: What does requests do when you pass a dict of lists of multiple strings? I couldn't find it in the docs, and I only remember passing it `list[tuple[str, str]]` or `dict[str, str]`.

The RFC standard doesn't prescribe any semantics. I know php parses a querysting "foo[]=1,foo[]=2" as an array foo. But I've seen Java services that do something with semi colons.

src/openforms/variables/migrations/0009_servicefetchconfiguration_name.py
```diff
@@ -0,0 +1,21 @@
...
+from django.db import migrations
+
+
+def populate_names(apps, _):
+    ServiceFetchConfiguration = apps.get_model("variables", "ServiceFetchConfiguration")
+
+    for config in ServiceFetchConfiguration.objects.all():
+        if not config.name:
+            config.name = f"{config.method} - {config.service.api_root}"
+            config.save()
```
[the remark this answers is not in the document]
stevenbal: Looks neat, didn't know that was possible!

src/openforms/variables/tests/test_validators.py
```diff
@@ -184,3 +185,80 @@ def test_it_invalidates_field_names_invalid_characters(
...
+        }
+
+        self.validate(value)
```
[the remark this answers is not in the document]
stevenbal: `self.fail` and `raise self.failureException from e` seem to give the same information if a `ValidationError` occurs, so I'll stick to `self.fail`

src/openforms/variables/validators.py
```diff
@@ -123,6 +123,54 @@ def __call__(self, value: Mapping[str, str] | JSONValue) -> None:
             raise ValidationError(errors)
 
 
+@deconstructible
+class QueryParameterValidator:
+    """Validates if headers are well-formed according to RFC 3986.
+
+    ...
+    """
```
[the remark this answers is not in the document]
stevenbal: I guess since this validator is fairly generic, I could just write `Validates if the value is a mapping of strings to lists of strings`?

src/openforms/variables/tests/test_validators.py
stevenbal: Tested this real quick, it seems that doing `requests.get(url, params={"foo": ["bar", "baz"]})` results in a querystring like this: `?foo=bar&foo=baz`

src/openforms/variables/migrations/0011_migrate_interpolation_format.py
```diff
@@ -0,0 +1,71 @@
...
+
+import re
+
+from django.db import migrations
+
+MATCH_SINGLE_BRACKETS = r"{([A-z0-9_]*)}"
+REPL_SINGLE_BRACKETS = r"{{\1}}"
+
+MATCH_DOUBLE_BRACKETS = r"{{([A-z0-9_]*)}}"
+REPL_DOUBLE_BRACKETS = r"{\1}"
```
CharString, line 11: It think `r"\."` should be in the character class too. For components nested in layout components or repeated groups. 

(sorry for suggesting regexp changes in text, GitHub for some reason doesn't give me the suggestion button)

src/openforms/variables/migrations/0011_migrate_interpolation_format.py
```diff
@@ -0,0 +1,71 @@
...
+            config.query_params = apply_regex_nested(
+                config.query_params, MATCH_SINGLE_BRACKETS, REPL_SINGLE_BRACKETS
+            )
+
+        config.save()
+
+
+def backwards(apps, _):
+    ServiceFetchConfiguration = apps.get_model("variables", "ServiceFetchConfiguration")
+
```
CharString, line 47: This backwards migration is correct when applied immediately after the forward migration... But after that, people may use template filters in the interpolations. We can either drop these, to get syntactically correct values, but it would change the meaning and results. So I think we should give up on the backwards migration or give a warning that it may break.

CharString (changes requested):

src/openforms/variables/migrations/0011_migrate_interpolation_format.py
stevenbal: Hmm that's true, I guess I'll just drop it then

src/openforms/variables/migrations/0011_migrate_interpolation_format.py
```diff
@@ -0,0 +1,51 @@
...
+import re
+
+from django.db import migrations
+
+MATCH_SINGLE_BRACKETS = r"{([A-z0-9_\.]*)}"
+REPL_SINGLE_BRACKETS = r"{{\1}}"
+
+MATCH_DOUBLE_BRACKETS = r"{{([A-z0-9_\.]*)}}"
+REPL_DOUBLE_BRACKETS = r"{\1}"
```
sergei-maertens, line 11: these are unused constants now

src/openforms/variables/validators.py
```diff
@@ -123,6 +123,51 @@ def __call__(self, value: Mapping[str, str] | JSONValue) -> None:
...
+                    ).format(
+                        parameter=field_name,
+                        value=field_value,
+                    )
+                )
+            elif len(field_value):
+                if not all(isinstance(sub_value, str) for sub_value in field_value):
+                    errors.append(
+                        _(
+                            "{parameter!s}: value '{value!s}' should be a list of strings, but isn't."
```
sergei-maertens, line 156: the exact messages here defined are not the same as the ones used in tests, so after entering translations the tests will break. Please use the exact same messages.

sergei-maertens (changes requested):
FYI - zgw-consumers 0.26.0 has an `uuid` field now for `Service` that we can use in the API endpoints :)

sergei-maertens (approved):

PR #2847, opened 2023-03-09
[#2826] validate date value in logic rules

### LINKED ISSUE #2826: Date in logic rules only work when formatted: yyyy-mm-dd, otherwise form breaks "Could not parse date '10-10-2020'" (low)

pi-sigma:
Fixes #2826

* values for date variables in logic rules should be specified in ISO 8601fomat
* removed `JsonLogicActionValueValidator` since it's no longer needed and replaced with basic `JsonLogicValidator`

### DIFF
--- a/src/openforms/forms/api/serializers/logic/action_serializers.py
+++ b/src/openforms/forms/api/serializers/logic/action_serializers.py
@@ -1,4 +1,5 @@
 import warnings
+from datetime import date
 
 from django.urls import resolve
 from django.utils.translation import ugettext_lazy as _
@@ -6,10 +7,13 @@ from django.utils.translation import ugettext_lazy as _
 from drf_polymorphic.serializers import PolymorphicSerializer
 from drf_spectacular.utils import extend_schema_serializer
 from furl import furl
+from json_logic.typing import Primitive
 from rest_framework import serializers
 
 from openforms.api.serializers import DummySerializer
 from openforms.submissions.api.fields import URLRelatedField
+from openforms.utils.json_logic.api.validators import JsonLogicValidator
+from openforms.variables.constants import FormVariableDataTypes
 
 from ....constants import (
     LOGIC_ACTION_TYPES_REQUIRING_COMPONENT,
@@ -18,7 +22,6 @@ from ....constants import (
     PropertyTypes,
 )
 from ....models import FormStep
-from ...validators import JsonLogicActionValueValidator
 from .fields import ActionFormStepUUIDField
 
 
@@ -61,7 +64,7 @@ class LogicValueActionSerializer(serializers.Serializer):
             "A valid JsonLogic expression describing the value. This may refer to "
             "(other) Form.io components."
         ),
-        validators=[JsonLogicActionValueValidator()],
+        validators=[JsonLogicValidator()],
     )
 
 
@@ -151,9 +154,11 @@ class LogicComponentActionSerializer(serializers.Serializer):
 
     def validate(self, data: dict) -> dict:
         """
-        Check that the component is supplied depending on the action type.
+        1. Check that the component is supplied depending on the action type.
+        2. Check that the value for date variables has the right format
         """
         action_type = data.get("action", {}).get("type")
+        action_value = data.get("action", {}).get("value")
         component = data.get("component")
         form_step = data.get("form_step")
 
@@ -188,6 +193,27 @@ class LogicComponentActionSerializer(serializers.Serializer):
                 {"variable": self.fields["variable"].error_messages["blank"]},
                 code="blank",
             )
+
+        # validate format of value for date variable
+        if action_type == LogicActionTypes.variable and isinstance(
+            action_value, Primitive
+        ):
+            form_var = self.context["form_variables"].variables[variable]
+
+            if form_var.data_type == FormVariableDataTypes.date:
+                try:
+                    date.fromisoformat(action_value)
+                except (ValueError, TypeError) as ex:
+                    raise serializers.ValidationError(
+                        {
+                            "action": {
+                                "value": _(
+                                    "Value for date variable must be a string in the "
+                                    "format yyyy-mm-dd (e.g. 2023-07-03)"
+                                ),
+                            }
+                        },
+                    ) from ex
 
         if (
             action_type
--- a/src/openforms/forms/api/validators.py
+++ b/src/openforms/forms/api/validators.py
@@ -1,4 +1,4 @@
-from typing import Any, Union, cast
+from typing import Any, cast
 
 from django.utils.translation import gettext as _
 
@@ -15,24 +15,6 @@ from openforms.utils.json_logic.api.validators import JsonLogicValidator
 from openforms.variables.service import get_static_variables
 
 from ..validation.registry import register as formio_validators_registry
-
-
-class JsonLogicActionValueValidator(JsonLogicValidator):
-    code = "invalid"
-    message = _("This field needs to refer to a form component")
-
-    def __call__(self, value: Union[dict, str]) -> None:
-        if isinstance(value, (str, int, bool)):
-            return
-
-        # ensure that the expression itself is valid
-        super().__call__(value)
-
-        expression = JSONLogicExpression.from_expression(value)
-        tree = expression.as_tree()
-        assert isinstance(tree, Operation)
-        if tree.arguments[0] == "":
-            raise serializers.ValidationError(self.message, code=self.code)
 
 
 class JsonLogicTriggerValidator(JsonLogicValidator):
--- a/src/openforms/forms/tests/test_serializers.py
+++ b/src/openforms/forms/tests/test_serializers.py
@@ -1,0 +1,81 @@
+from django.test import TestCase
+
+from hypothesis import given
+from hypothesis.extra.django import TestCase as HypothesisTestCase
+from json_logic.typing import Primitive
+
+from openforms.forms.api.datastructures import FormVariableWrapper
+from openforms.forms.api.serializers.logic.action_serializers import (
+    LogicComponentActionSerializer,
+)
+from openforms.forms.tests.factories import FormFactory, FormVariableFactory
+from openforms.tests.search_strategies import json_values
+from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
+
+json_values_primitive = json_values().filter(lambda j: isinstance(j, Primitive))
+
+
+class LogicComponentActionSerializerPropertyTest(HypothesisTestCase):
+    @given(json_values_primitive)
+    def test_date_format_validation_against_primitive_json_values(self, json_value):
+        """Assert that serializer is invalid for random Primitive json data types
+        (str, int, etc.) as logic action values for date variables
+        """
+        form = FormFactory.create()
+        FormVariableFactory.create(
+            source=FormVariableSources.user_defined,
+            key="date_var",
+            form=form,
+            data_type=FormVariableDataTypes.date,
+        )
+        form_vars = FormVariableWrapper(form=form)
+
+        serializer = LogicComponentActionSerializer(
+            data={
+                "variable": "date_var",
+                "action": {
+                    "type": "variable",
+                    "value": json_value,
+                },
+            },
+            context={"request": None, "form_variables": form_vars},
+        )
+        self.assertFalse(serializer.is_valid())
+
+
+class LogicComponentActionSerializerTest(TestCase):
+    def test_date_variable_with_non_trivial_json_trigger(self):
+        """Check that valid JSON expressions are not ruled out by date format validation"""
+        form = FormFactory.create()
+        FormVariableFactory.create(
+            source=FormVariableSources.user_defined,
+            key="someDate",
+            form=form,
+            data_type=FormVariableDataTypes.date,
+        )
+        FormVariableFactory.create(
+            source=FormVariableSources.user_defined,
+            key="anotherDate",
+            form=form,
+            data_type=FormVariableDataTypes.date,
+        )
+        form_vars = FormVariableWrapper(form=form)
+
+        serializer = LogicComponentActionSerializer(
+            data={
+                "json_logic_trigger": {
+                    "==": [
+                        {"var": "someDate"},
+                        {},
+                    ]
+                },
+                "variable": "anotherDate",
+                "action": {
+                    "type": "variable",
+                    "value": {"var": "anotherDate"},
+                },
+            },
+            context={"request": None, "form_variables": form_vars},
+        )
+
+        self.assertTrue(serializer.is_valid())
--- a/src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
+++ b/src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
@@ -81,7 +81,6 @@ class VariablesInLogicBulkAPITests(APITestCase):
         FormVariableFactory.create(
             source=FormVariableSources.user_defined, key="testUserDefined", form=form
         )
-
         form_logic_data = [
             {
                 "form": f"http://testserver{reverse('api:form-detail', kwargs={'uuid_or_slug': form.uuid})}",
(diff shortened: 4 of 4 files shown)


pi-sigma:
I included the validation for date values in the `LogicComponentActionSerializer`. Not sure if this is the right place.

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +190,19 @@ def validate(self, data: dict) -> dict:
                 code="blank",
             )
 
+        if action_type and action_type == "variable":
+            form_var = FormVariable.objects.get(key=variable)
```
SilviaAmAm, line 194: This will look through all FormVariables (not only form variables related to this particular form). So if you have another form with a variable with the same key, this will give an error.

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +190,19 @@ def validate(self, data: dict) -> dict:
...
+
+            if form_var.data_type == "date":
```
SilviaAmAm, line 196: I think here and on line 193 (`action_type == "variable"`) you can use `FormVariableDataTypes.date` and  `LogicActionTypes.variable` respectively.

This helps in case we ever change the string, then we don't have to change strings everywhere but just in one place.

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +190,19 @@ def validate(self, data: dict) -> dict:
...
+                try:
+                    date.fromisoformat(data.get("action").get("value"))
```
SilviaAmAm, line 198: I think you can use `format_date_value` (from `openforms.utils.date`) instead of `date.fromisoformat`.

Also if `data.get("action")` returns `None`, the next get would give an error. So you would have to do `data.get("action", {}).get("value")`.

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +190,19 @@ def validate(self, data: dict) -> dict:
...
+                except ValueError:
+                    raise serializers.ValidationError(
+                        {
+                            "action.value": "Value for date variable must be in isoformart (yyyy-mm-dd)"
```
SilviaAmAm, line 202:
```suggestion
                            "action.value": "Value for date variable must be in isoformat (yyyy-mm-dd)"
```

SilviaAmAm: Although maybe this could be simplified to `Value for date variable must be in format yyyy-mm-dd? :thinking:

src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
```diff
@@ -117,6 +116,57 @@ def test_invalid_user_defined_variable_in_trigger(self):
...
+        FormVariableFactory.create(
+            source=FormVariableSources.user_defined,
+            key="date_var",
+            form=form,
+            data_type=FormVariableDataTypes.date,
+        )
+        FormStepFactory.create(
+            form=form,
+            form_definition__configuration={
+                "components": [
```
SilviaAmAm, line 131: Maybe you don't need to set the configuration. Since in this case it doesn't really matter what the form is?

src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
```diff
@@ -117,6 +116,57 @@ def test_invalid_user_defined_variable_in_trigger(self):
...
+                        "key": "textfield",
+                    },
+                ]
+            },
+        )
+        form_logic_data = [
+            {
+                "form": f"http://testserver{reverse('api:form-detail', kwargs={'uuid_or_slug': form.uuid})}",
+                "order": 0,
+                "json_logic_trigger": {"==": [{"var": "textfield"}, "some test value"]},
```
SilviaAmAm, line 143: Then, if you remove the textfield, here you can use some trigger `{"!!": [True]}` which always evaluates to true.

src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
```diff
@@ -117,6 +116,57 @@ def test_invalid_user_defined_variable_in_trigger(self):
...
+                "actions": [
+                    {
+                        "variable": "date_var",
+                        "action": {
+                            "type": "variable",
+                            "value": "07-03-2023",  # not isoformart
```
SilviaAmAm, line 149:
```suggestion
                            "value": "07-03-2023",  # not isoformat
```

SilviaAmAm (changes requested):
I think having the check for the value in the `validate` of the `LogicComponentActionSerializer` is okay, although it would be good to change the description of the function (line 155) since now it's doing more than just checking the component.

It would be nice to have it in the `LogicValueActionSerializer` but I don't know if there is a way to pass down enough context for it to do the check :thinking:

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +192,20 @@ def validate(self, data: dict) -> dict:
                 code="blank",
             )
 
+        # validate format of value for date variable
+        if action_type and action_type == LogicActionTypes.variable:
+            # get related form variable from context of parent serializer
+            form_var = self.parent.parent.context["form_variables"].variables[variable]
```
sergei-maertens, line 198: I believe the context is the same for the entire tree, meaning `self.context is self.parent.context is self.parent.parent.context`. This is also kinda required, since a serializer itself should not be concerned with how deeply it is nested (as that would break the ability to compose serializers).

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +192,20 @@ def validate(self, data: dict) -> dict:
                 code="blank",
             )
 
+        # validate format of value for date variable
+        if action_type and action_type == LogicActionTypes.variable:
```
sergei-maertens, line 196:
```suggestion
        if action_type == LogicActionTypes.variable:
```
tautology - if action type is `None` or falsy, it will never be equal to the literal string so you can shortcut this.

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +192,20 @@ def validate(self, data: dict) -> dict:
...
+            # get related form variable from context of parent serializer
+            form_var = self.parent.parent.context["form_variables"].variables[variable]
+            action_value = data.get("action", {}).get("value")
+
+            if form_var.data_type == FormVariableDataTypes.date:
+                if _ := format_date_value(action_value) == "":
+                    raise serializers.ValidationError(
```
sergei-maertens, line 203: I know Silvia said otherwise, but I'm not a fan of this flow control - much rather see a `ValueError` being caught and then `raise ValidationError(...) from exc` for full context.

`format_date_value` is written with the intent that a valid date structure was passed in, this is abusing it outside of its intended usage.

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +192,20 @@ def validate(self, data: dict) -> dict:
...
+                        {
+                            "action.value": "Value for date variable must be in the format yyyy-mm-dd"
```
sergei-maertens, line 205: this error message structure should be a nested structure, the formatting with dotted paths is something that is done in the generic validation error handler -> I expected a nested dict here instead.

src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
```diff
@@ -117,6 +116,46 @@ def test_invalid_user_defined_variable_in_trigger(self):
...
+
+        self.assertEqual(status.HTTP_400_BAD_REQUEST, response.status_code)
+
+        error = response.json()
+
+        self.assertEqual("invalid", error["code"])
+        self.assertEqual(
+            "Value for date variable must be in the format yyyy-mm-dd",
+            error["invalidParams"][0]["reason"],
+        )
```
sergei-maertens, line 157: I'd like to see an assertion on the param `name` here too, to ensure we have a valid data path for the frontend to render.

sergei-maertens (changes requested):

src/openforms/forms/tests/variables/test_variables_in_logic_trigger.py
```diff
@@ -117,6 +116,47 @@ def test_invalid_user_defined_variable_in_trigger(self):
             error["invalidParams"][0]["reason"],
         )
 
+    def test_invalid_date_format_for_user_defined_variable(self):
```
sergei-maertens, line 119: you have the error flow tested now, but where is the test for the happy flow?

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -189,6 +192,23 @@ def validate(self, data: dict) -> dict:
...
             )
 
+        # validate format of value for date variable
+        if action_type == LogicActionTypes.variable:
+            form_var = self.context["form_variables"].variables[variable]
+            action_value = data.get("action", {}).get("value")
+
+            if form_var.data_type == FormVariableDataTypes.date:
+                try:
+                    date.fromisoformat(action_value)
```
sergei-maertens, line 202: try writing a test with `{"action": {"value": 42}}` and see what happens :)

-> this would be a good testcase for a hypothesis test with all JSON types, but do it on the serializer itself rather than the entire endpoint.

sergei-maertens (changes requested):

src/openforms/forms/api/serializers/tests/test_action_serializers.py
sergei-maertens, line 1: Please put this file with the other tests in `openforms.forms.tests` instead of creating a new test directory structure in `openforms.forms.api.serializers.tests`

src/openforms/forms/api/serializers/tests/test_action_serializers.py
```diff
@@ -0,0 +1,169 @@
+from django.utils.translation import ugettext_lazy as _
+
+from hypothesis import given
+from hypothesis.extra.django import TestCase as HypothesisTestCase
+from hypothesis_jsonschema import from_schema
```
sergei-maertens, line 5: hypothesis-jsonschema is designed to work with json-schema, which is a (draft) standard for describing json schema's in JSON. It is used internally by OpenAPI speficitation, for example.

This does do what you think it does in terms of "throw whatever valid JSON at the serializer and see if it explodes", as json-schema is much more strict and structured.

So, please remove this dependency, and instead write the test using existing constructs we have:

1. We have a JSON search strategy here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/tests/search_strategies.py#L29
2. Example usage: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/variables/tests/test_validators.py#L71

Then finally, strings are the only type that we do allow/expect, so we need to exclude those, you can use `filter` for that.

This should result in testcode roughly looking like:

```py
json_without_strings = json_values.filter(lambda c: not isinstance(c, str))

@given(json_without_strings)
def test_date_validation_arbitrary_json_input(self, json_value):
    serializer = LogicComponentActionSerializer(...)
    
    is_valid = serializer.is_valid()

    self.assertFalse(is_valid)
    self.assertEqual(
            serializer.errors["action"]["value"],
            _(
                "Value for date variable must be a string in the format "
                "yyyy-mm-dd (e.g. 2023-07-03)"
            ),
        )
```

Hypothesis will take care of generating numbers, booleans, objects, arrays... for that test and run it with a number of combinations. The end result of anything that is not a string should be an invalid serializer.

src/openforms/forms/api/validators.py
```diff
@@ -20,8 +20,9 @@
 class JsonLogicActionValueValidator(JsonLogicValidator):
     code = "invalid"
     message = _("This field needs to refer to a form component")
+    requires_context = True
 
-    def __call__(self, value: Union[dict, str]) -> None:
+    def __call__(self, value: Union[dict, str], context) -> None:
```
sergei-maertens, line 25: why is the context added but not used?

sergei-maertens (changes requested):

src/openforms/forms/api/serializers/tests/test_action_serializers.py
CharString: If you don't need the strategy elsewhere you could also use [assume](https://hypothesis.readthedocs.io/en/latest/details.html#hypothesis.assume) in the single test that needs some cases filtered:
```py
@given(json_values)
def test_something(self, json_value):
    assume(not isinstance(json_value, str))
    ...  # assertions....
```

`assume` is like `assert` that doesn't fail the test, but "fails the generated value" so to speak.

pi-sigma: This test doesn't work. The validator on action values in `openforms/forms/api/validators.py` will throw an `AssertionError` for arrays, lists, etc.

joeribekker:
Refinement: The `expression.as_tree` operation might be too strict on what is valid JSON Logic. If it's a list, it will return a list of trees and otherwise an operation. So, there are cases that the unexpected is returned but this seem valid anyway.

src/openforms/forms/api/validators.py
```diff
@@ -30,8 +30,8 @@ def __call__(self, value: Union[dict, str]) -> None:
 
         expression = JSONLogicExpression.from_expression(value)
         tree = expression.as_tree()
-        assert isinstance(tree, Operation)
-        if tree.arguments[0] == "":
+        assert isinstance(tree, Operation | float | list)
+        if isinstance(tree, Operation) and tree.arguments[0] == "":
```
sergei-maertens, line 34: the float should really go above with the other primitives, and you could re-use the type hint for this:

```py
from openforms.typing import JSONPrimitive

if isinstance(value, JSONPrimitive):
    return

super().__call__(value)

expression = SONLogicExpression.from_expression(value)
tree = expression.as_tree()
```
I'm trying to figure out where this validation comes from, from the commits it seems that #1472 is the reason:

> Validate that component can't be empty (both in the trigger and the actions) 

via #1499

But it seems that `test_cant_have_empty_component_value_in_value_action` is also a wrong promise, since `{"var": ""}` is a valid JSON logic expression referring to the identity of what you put in:

```py
>>> from json_logic import jsonLogic
>>> foo = {"bar": "baz"}
>>> jsonLogic({"var": ""}, foo)
{'bar': 'baz'}
```
So perhaps this validator doesn't even have a reason anymore to exist :thinking: And if it does exist, we should validate that the `Var` leaf nodes point to valid form variables, but that can in turn cause dead references (e.g. `{"var": {"var": "foo"}}` that can't be validated due to the dynamic nature. Probably those complex scenarios should be caught via static analysis for data types (big topic #2324 ).

So I'm leaning towards getting rid of this validator alltogether.

sergei-maertens: @pi-sigma let's replace this validator with the base `JSONLogicValidator` and remove the associated test(s). It's no longer relevant.

src/openforms/forms/tests/test_serializers.py
```diff
@@ -0,0 +1,36 @@
...
+from openforms.forms.api.datastructures import FormVariableWrapper
+from openforms.forms.api.serializers.logic.action_serializers import LogicComponentActionSerializer
+from openforms.forms.tests.factories import FormFactory, FormVariableFactory
+from openforms.tests.search_strategies import json_values
+from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
+
+
+class LogicComponentActionSerializerPropertyTest(HypothesisTestCase):
+    @given(json_values())
+    def test_date_format_validation_against_arbitrary_json_values(self, json_value):
```
sergei-maertens, line 13: can you add a happy flow test so that a date type variable still works with a non-trivial json logic expression like `{"var": "anotherDate"}`?

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2867, opened 2023-03-10
[#2844] improve error message for confirmation email options

### LINKED ISSUE #2844: Errors for confirmationEmailOptions are not shown

2.0.x, 2.1.x

When a user selects form-specific confirmation email options but doesn't fill in a template, the fact that saving gives an error is unclear. 

- Create a form with a step with a random field.
- Under Bevestiging, at the bottom select 'formulierspecifieke email'
- Save the form

It looks like all went well (except there is no green banner at the top saying 'Success...'). But if you look at the Bevestigings tab you see:
Screenshot from 2023-03-08 16-25-18

This is because the onderwerp / inhoud fields are not filled in. 

- [ ] There should be an error message at the top that something went wrong
- [ ] The error message should be clearer (i had to look at the code to understand what it meant)

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

pi-sigma:
Fixes #2844 

* The error was shown on the confirmation email dropdown but the message was too unspecific and therefore unclear
* There was already a test in place that checked for empty templates, but both the title of the function and the docstring were misleading, hence easy to miss
* Refactored the function that checks if a template is empty (by refactoring `ConfirmationEmailTemplate.is_usable` to `ConfirmationEmailTemplate.is_empty` and adjusting the relevant tests)


sergei-maertens (commented):
How does this behave when you do configure a subject and/or body for the template, but leave the dropdown set to "use the global configuration"?

src/openforms/emails/models.py
```diff
@@ -65,5 +65,5 @@ def __str__(self):
         return _("Confirmation email template - {form}").format(form=form_str)
 
     @property
-    def is_usable(self) -> bool:
-        return bool(self.subject and self.content)
+    def is_empty(self) -> bool:
```
SilviaAmAm, line 68: Im not sure if `is_empty` makes more sense, since it concerns 2 properties of `ConfirmationEmailTemplate`. I think I would leave `is_usable` and maybe add a docstring saying that the `ConfirmationEmailTemplate` is only usable if both the subject / content are present?

src/openforms/forms/api/serializers/form.py
```diff
@@ -248,16 +248,18 @@ def validate(self, attrs):
...
-                                "The form specific confirmation email template is not set up correctly and "
-                                "can therefore not be selected."
+                                "The form-specific confirmation email template "
+                                "(subject + content) should not be empty."
                             ),
                             code="invalid",
-                        )
+                        ),
+                        # raise vacuous error b/c the problem concerns email_option + template
+                        "confirmation_email_template": ErrorDetail(""),
```
SilviaAmAm, line 262: I'm not sure if this is useful, because the user doesn't see it. I think that the updated error is clear. Maybe you can make it even more explicit and say: "In order to use an email template specific to this form, both the subject and the content of the template need to be filled in" ... or something along those lines :P

SilviaAmAm (changes requested):
It would be good to fix the badge for the `confirmation_email_option` error in the front end. I think you can do this by adding `confirmationEmailOption: 'submission-confirmation',` here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/form-creation-form.js#L153

src/openforms/forms/api/serializers/form.py
pi-sigma: I should have added a more useful comment for this. The `ValidationError` on `confirmation_email_template` creates the error display on the confirmation tab, the `ValidationError` on `confirmation_email_option` creates the error display on the email option dropdown.

sergei-maertens (approved):

PR #2868, opened 2023-03-10
[#2851] Fix translation import

### LINKED ISSUE #2851: Importing form in browser with EN language preference gives empty names

2.1.x

DH Taiga issue 441 / user story 444

You can have a look at form 16 on the dev environment for DH.

The form is missing all the form name / step names

I think these step reproduce at least part of what is going wrong with that form:

- Create a form with a step in 2.0.1 on on DH test environment
- Export the form (single export) (here the language preference of the browser does not matter) 
- Import the form locally from a browser with english language preference
=> observe that the names of the form and the steps are empty in dutch. 

Saving the form gives errors for the blank names, but updating the names doesn't seem to fix the errors

Note: if you import the form multiple times, delete the form and the form definitions (!!) before re-importing

I am not sure, but I guess the dutch names should be filled so that if you save the form you don't get an error?

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2851

### DIFF
--- a/src/openforms/forms/utils.py
+++ b/src/openforms/forms/utils.py
@@ -9,6 +9,7 @@ from uuid import uuid4
 from django.conf import settings
 from django.db import transaction
 from django.utils import timezone
+from django.utils.translation import override
 
 from rest_framework.exceptions import ValidationError
 from rest_framework.test import APIRequestFactory
@@ -139,6 +140,7 @@ def import_form(import_file, existing_form_instance=None):
 
 
 @transaction.atomic
+@override(language=settings.LANGUAGE_CODE)
 def import_form_data(
     import_data: dict, existing_form_instance: Form = None
 ) -> List[FormDefinition]:
--- a/src/openforms/config/models.py
+++ b/src/openforms/config/models.py
@@ -25,8 +25,9 @@ from openforms.emails.validators import URLSanitationValidator
 from openforms.payments.validators import validate_payment_order_id_prefix
 from openforms.template import openforms_backend, render_from_string
 from openforms.template.validators import DjangoTemplateValidator
+from openforms.translations.utils import ensure_default_language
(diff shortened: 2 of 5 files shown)


src/openforms/forms/utils.py
```diff
@@ -138,9 +140,25 @@ def import_form(import_file, existing_form_instance=None):
...
 
 
-@transaction.atomic
+@contextmanager
+def default_language(language: str, *args, **kwargs):
+    activate(language)
+    try:
+        yield
+    finally:
+        deactivate()
```
sergei-maertens, line 149: we now have some duplication and slightly different ways of doing things, see `openforms.forms.api.viewsets.FormViewSet.retrieve` where the following code is used:

```py
if not form.translation_enabled and not is_admin_request(request):
    translation.activate(settings.LANGUAGE_CODE)
```

IMO `settings.LANGUAGE_CODE` is the correct setting to use rather than `MODELTRANSLATION_DEFAULT_LANGUAGE`, as it means this is our default language in the application. The model translation setting is used internally in their fallback logic for when a translation for the active language is missing.

Additionally, this should be a utility that abstracts away the implementation details so that i can be used in both the import and viewset, something like:

```py
def activate_default_language() -> None:
    ...

@contextmanager
def force_default_language():
    ...
```

P.S. also look into `contextlib.ContextDecorator`, so you can use the utility both as context manager and as decorator, this would allow you to do:

```py
@transaction.atomic
@force_default_language
def import_form_data(...):
    ...
```

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
I'll merge and backport this on Monday

sergei-maertens:
Backported in 750823e848f18a03cc7b30765fb18160a11bdd51

PR #2883, opened 2023-03-14
[#2872] Change the flow of submission reference generation

### LINKED ISSUE #2872: Adjust mail to behandelaar
We willen de:

- mail naar behandelaar
- onderwerpregel

kunnen aanpassen zodat we bijvoorbeeld component/gebruikersvariabelen kunnen toevoegen.

From Sivlia:

Dit heeft te maken met de email registratie backend en niet de Object registratie backend in combinatie met een email notificatie. 

DH zou graag de template / subject van de email die gestuurd wordt door de email registratie backend (zoals bij de bevestigings email)

From Sergei:

Before sending registration email, generate registration reference and include it in the context and reuse it during reference extraction (extract public reference task)

SilviaAmAm:
Part of #2872 

TODO:
Update diagram with new flow => To be done once the flow is agreed


sergei-maertens (changes requested):

SilviaAmAm:
I indeed forgot to remove those extra `set_submission_reference`!

And I was unsure what to do with the abstract method => whether to have one that just passes in the parent so that you don't have to add it to all the classes that inherit from it. I went for the abstract method just because all the other methods on the plugin were abstract methods.

src/openforms/submissions/tasks/registration.py
```diff
@@ -1,30 +1,81 @@
...
+
+    if not registration_plugin:
+        return
+
+    # The registration options were validated in the pre-registration task AND in the registration task, so not
+    # validating again. Possible edge case: the options changed between the registration task and the obtain_reference
+    # task?
+    registration_plugin.obtain_submission_reference(
+        submission, submission.form.registration_backend_options
+    )
```
sergei-maertens, line 66: this is going a bit in circles now, can we not leave this original implementation?

`set_submission_reference` wraps around `get_reference_for_submission` which loads the registration plugin for the submission and calls `plugin.get_reference_from_result(result)`, so you now have plugin methods `get_reference_from_result` and `obtain_submission_reference` that just call each other (in a way). Additionally, `obtain_submission_reference` does not return a submission reference but mutates the provided submission, which is unexpected.

Preferably functions/methods have no side-effects and return an output for a given input.

Since `set_submission_reference` is idempotent and takes care of calling the correct plugin, I don't see why this `obtain_submission_reference` task needs changing.

sergei-maertens (changes requested):

PR #2885, opened 2023-03-14
[#2872] Make template configurable

### LINKED ISSUE #2872: Adjust mail to behandelaar

SilviaAmAm:
Fixes #2872 

Still to do (next PRs):
- Improve the UI for adding the custom templates?
- Documentation
- Help tool in wysiwyg widget for designers to help them with all the variables that are available in the templates


src/openforms/forms/api/serializers/form.py
```diff
@@ -191,6 +196,14 @@ class Meta:
                 "lookup_url_kwarg": "uuid_or_slug",
             },
         }
+        validators = [
+            AllOrNoneTruthyFieldsValidator(
+                "registration_email_subject",
+                "registration_email_payment_subject",
+                "registration_email_content_html",
+                "registration_email_content_text",
```
SilviaAmAm, line 204: I went for the same behaviour as for the confirmation email: if you have a custom subject, then you should also have a custom content. Im not sure this is really nice in this case (maybe it's okay to want to only overwrite the payment subject or only the content). :thinking: 

I think it would be nicer to only have:
``` python
AllOrNoneTruthyFieldsValidator("registration_email_content_html", "registration_email_content_text")
```
since the html and text content should have the same info

joeribekker: > (maybe it's okay to want to only overwrite the payment subject or only the content). 🤔

I like this. Simple but nice improvement!

However, I value consistent behaviour. Otherwise we get questions about this.
So, I agree by having it this way if we change it for the other emails as well. If not, then we should require both subject and content to be filled.

SilviaAmAm: Ok! I will make an issue to change the confirmation emails to have the same behaviour and tackle it in a different PR

CharString: :+1: I remember struggling with the first time I tried to configure a custom conformation email.

src/openforms/forms/tests/test_api_form_plugin_options.py
```diff
@@ -151,3 +151,89 @@ def test_payment_backend_options(self):
...
+        self.assertEqual(data["registrationEmailSubject"], "Custom subject")
+        self.assertEqual(data["registrationEmailPaymentSubject"], "")
+        self.assertEqual(
+            data["registrationEmailContentHtml"],
+            "",
+        )
+        self.assertEqual(
+            data["registrationEmailContentText"],
+            "",
+        )
```
CharString, line 183: We haven't set the asserted values explicitly to `""`, that's a behaviour from a Factory/model, right?
Then it's saver to check that they are the same to whatever the factory built.

src/openforms/forms/tests/test_api_form_plugin_options.py
```diff
@@ -151,3 +151,89 @@ def test_payment_backend_options(self):
...
+                "registration_email_content_text": "Custom text template {% payment_information %}",
+            },
+        )
+
+        self.assertEqual(response.status_code, 200)
+
+        data = response.json()
+
+        self.assertEqual(data["registrationEmailSubject"], "")
+        self.assertEqual(data["registrationEmailPaymentSubject"], "")
```
CharString, line 206: Same here

src/openforms/registrations/contrib/email/utils.py
```diff
@@ -0,0 +1,33 @@
...
+            submission.form.registration_email_content_html,
+            submission.form.registration_email_content_text,
+        )
+
+    return (
+        subject_template,
+        payment_subject_template,
+        config.registration_email_content_html,
+        config.registration_email_content_text,
+    )
```
CharString, line 33: Maybe define an explicit return type instead of a generic 4-tuple. It's clearer and more flexible to change. A simple `namedtuple` or `dataclass` will do. I think Sergei prefers the latter.

A 3-tuple, like from `EmailRegistration.render_registration_email`, is the borderline of my taste.

src/openforms/registrations/contrib/email/plugin.py
```diff
@@ -77,51 +73,64 @@ def render_registration_email(submission, extra_context=None):
         }
         base_context = {
             "public_reference": submission.public_registration_reference,
-            "datetime": timezone.localtime(submission.completed_on).strftime(
+            "completed_on": timezone.localtime(submission.completed_on).strftime(
```
CharString, line 76: Love this name change. Does it need a data migration?

CharString (changes requested):
With all these variables people are required to add to their templates, should we consider adding an insertion tool to the WYSIWYG toolbar that provides easy access?

src/openforms/forms/tests/test_api_form_plugin_options.py
SilviaAmAm: Yes indeed, it's the behaviour of the model, since these attributes are by default empty! I will change the assertion.

src/openforms/registrations/contrib/email/plugin.py
SilviaAmAm: I don't think so, since up to now the templates were not customisable? :thinking:

SilviaAmAm:
It would definitely be helpful for the designers to have a help tool for the variables! I will put a note about it for the things that still need to happen in the next PRs.

sergei-maertens (changes requested):
There are some aspects here that conflict with the idea of registration backends being pluggable and self-contained:

* `src/openforms/template/backends/sandboxed_django.py` needed updating to add a plugin template tag
* the global defaults for templates are added as model fields on the `GlobalConfiguration` model
* the form-specific overrides are added as model fields on the `Form` model.

So even if you don't use or enable this registration plugin, the data is there and may even confuse users. 

Architecturally, I think we should stick to the idea that plugins are self-contained:

* Add a model `openforms.registrations.contrib.email.models.Defaults` or something like that where you track the global configuration fields, instead of on `config.GlobalConfiguration`, as this is plugin-specific configuration and not Open Forms wide configuration.
* Bring the form-specific overrides into the plugin options via `openforms.registration.contrib.email.config.EmailOptionsSerializer`. The plugin-specific config is saved in the form JSONField and can later be accessed through the `options` in the registration plugin methods.
* Add/implement a hook for plugins so they can specify which additional built in template tags should be included, so the `get_openforms_backend` can then consult the registry of enabled registration plugins and add the specified import paths for template tags.

SilviaAmAm:
Todo:
- [ ] Move GlobalConfig fields to EmailConfig model
- [ ] Move Form fields to EmailOptionsSerializer
- [ ] Add hook for templates

src/openforms/registrations/contrib/email/templatetags/data_summary.py
```diff
@@ -0,0 +1,15 @@
+from django import template
+from django.template.loader import get_template
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def data_summary(context):
```
sergei-maertens, line 8:
```suggestion
def email_data_summary(context):
```
I'd consider namespacing this templatetag too - it's intended for e-mail, I can imagine there being a `data_summary` for different contexts too :thinking: 

We do already have the `{% summary %}` tag which is used for the confirmation e-mail - maybe that should be renamed/aliased to `confirmation_summary` and this one named `registration_summary`?

src/openforms/template/backends/sandboxed_django.py
```diff
@@ -43,7 +43,7 @@ def get_templatetag_libraries(self, custom_libraries: dict) -> dict:
         This prevents user-content from loading and using the libraries that are
         private API to Open Forms.
         """
-        return {}
+        return custom_libraries
```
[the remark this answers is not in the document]
SilviaAmAm: I don't think so? Since it is initialised with `backend = SandboxedDjangoTemplates({})` and so the params are `{}` which then results in `libraries` to be empty?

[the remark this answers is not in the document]
sergei-maertens: I've added an explicit test to confirm this - it works because we set the default for `libraries` in the init params to empty dict and don't specify this anywhere else & because the method override basically removes Django's template tag discovery. At least now there is an explicit test :)

I'm not sure why this change was needed though, because we don't expose them via `libraries`, but via `builtins` instead :thinking:

sergei-maertens (approved):

PR #2887, opened 2023-03-15
Add logging to service fetch action

### LINKED ISSUE #2681: Add logging to Service Fetch Action
Service fetch logic action needs to log when fetches are performed.

Ensure that the full logic evaluation logs with context are visible in the submission admin > logs as well. Currently they show that the logic rule was triggered, but no additional information is visible. This should log the resulting extracted value at minimum. Also add python level logging

CharString:
Closes #2681

### DIFF
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -7,7 +7,7 @@ from openforms.formio.service import FormioConfigurationWrapper
 from openforms.forms.constants import LogicActionTypes
 from openforms.forms.models import FormLogic, FormVariable
 from openforms.typing import DataMapping, JSONObject, JSONValue
-from openforms.utils.json_logic import ComponentMeta, introspect_json_logic
+from openforms.utils.json_logic import ComponentMeta
 from openforms.variables.models import ServiceFetchConfiguration
 
 from ..models import SubmissionStep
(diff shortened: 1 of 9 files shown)


src/openforms/submissions/logic/actions.py
```diff
@@ -225,16 +226,37 @@ class ServiceFetchAction(ActionOperation):
...
         # XXX perform_service_fetch is expressed in terms of FormVariables
         dummy_var = FormVariable(
             name=self.variable,
             service_fetch_configuration=ServiceFetchConfiguration.objects.get(
                 pk=self.fetch_config
             ),
         )
-        with log_errors({}, self.rule):  # TODO proper logging/error handling
-            return {dummy_var.name: perform_service_fetch(dummy_var, context)}
+        with log_errors({}, self.rule):  # TODO proper error handling
```
CharString, line 241: This context manager is a bit to concrete.
I'm not sure whether I should change it, or turn this instance into a try-except.

We might even do a `log()` call on exception and log the error to the history in the admin.

src/openforms/submissions/logic/actions.py
```diff
@@ -225,16 +226,37 @@ class ServiceFetchAction(ActionOperation):
...
+            result = perform_service_fetch(dummy_var, context)
+            log(asdict(result))
```
CharString, line 243: This is the most general thing that gives useful debugging information. I'm not sure if extra markup is wanted in the History view in  the admin.

src/openforms/submissions/logic/actions.py
sergei-maertens: the admin `log` call makes sense, but it should be generic, something along the lines of "Service fetch action crashed (contact technical support)."

It kind of depends if it's a crash caused by bad logic rules, or by bugs by developers. The logic log page is meant for form designers so they can test their forms and see that their logic rules are behaving as they expect.

src/openforms/submissions/logic/actions.py
```diff
@@ -190,29 +194,26 @@ class VariableAction(ActionOperation):
...
-            action_logic_introspection = introspect_json_logic(
-                self.value, component_map, initial_data
-            )
-            value = action_logic_introspection.as_string()
-
         return {
             "key": self.variable,
             "type_display": LogicActionTypes.get_label(LogicActionTypes.variable),
-            "value": value,
+            "value": log_data["value"],
```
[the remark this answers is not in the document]
sergei-maertens: I'll spin up or check a 2.0.x instance to check how this was supposed to look so we can get that fixed in master then.

sergei-maertens (approved):

PR #2898, opened 2023-03-16
Fix logic crashes in celery task

### LINKED ISSUE #2893: Crash on logic rule logging

2.1.0 / latest

Celery traceback:

```
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.10/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/bbt/code/open-forms/src/openforms/logging/tasks.py", line 43, in log_logic_evaluation
    log_entry = _log_logic_evaluation(
  File "/home/bbt/code/open-forms/src/openforms/logging/logic.py", line 57, in log_logic_evaluation
    input_data += rule_introspection.get_input_components(
  File "/home/bbt/code/open-forms/src/openforms/utils/json_logic/datastructures.py", line 57, in get_input_components
    component_meta = components_map[key]
KeyError: 'today'
```

Logic rule:

image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #2893

Good news: this doesn't crash the actual logic evaluation

Bad news: crashes do happen in the celery tasks and not only for non-primitive JSON types.

### DIFF
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -213,7 +213,7 @@ class VariableAction(ActionOperation):
         return {
             "key": self.variable,
             "type_display": LogicActionTypes.get_label(LogicActionTypes.variable),
-            "value": log_data["value"],
+            "value": log_data.get("value", ""),  # error caught by log_errors in `eval`
         }
 
 
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -1,13 +1,13 @@
-from dataclasses import dataclass
-from typing import Any, Dict, Mapping, Optional, Type, TypedDict
+from dataclasses import asdict, dataclass
+from typing import Any, Callable, Dict, Mapping, Optional, Type, TypedDict
 
 from json_logic import jsonLogic
 
 from openforms.formio.service import FormioConfigurationWrapper
 from openforms.forms.constants import LogicActionTypes
 from openforms.forms.models import FormLogic, FormVariable
-from openforms.typing import DataMapping, JSONObject
-from openforms.utils.json_logic import ComponentMeta, introspect_json_logic
+from openforms.typing import DataMapping, JSONObject, JSONValue
+from openforms.utils.json_logic import ComponentMeta
 from openforms.variables.models import ServiceFetchConfiguration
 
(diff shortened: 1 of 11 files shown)


src/openforms/utils/json_logic/datastructures.py
```diff
@@ -54,6 +56,11 @@ def get_input_components(
 
             # TODO: this *may* be nested var.var expressions
             key = cast(str, node.arguments[0])
+            if key not in components_map:
+                inputs.append(
+                    InputVar(key=key, value=context[key], step_name="", label="")
```
SilviaAmAm, line 61: I think this fails for selectboxes components. For example, if you have a selectboxes component with values A and B, the value of this component in the data is `{"a": False, "b": False}`. In the logic rule you would have the trigger contain `{"var": "selectBoxes.a"}`, but as a key `"selectBoxes.a"` doesn't correspond to any component.

src/openforms/submissions/logic/actions.py
```diff
@@ -190,29 +194,26 @@ class VariableAction(ActionOperation):
...
-            action_logic_introspection = introspect_json_logic(
-                self.value, component_map, initial_data
-            )
-            value = action_logic_introspection.as_string()
-
         return {
             "key": self.variable,
             "type_display": LogicActionTypes.get_label(LogicActionTypes.variable),
-            "value": value,
+            "value": log_data["value"],
```
SilviaAmAm, line 216: When calling `get_action_log_data` from `get_targeted_components` there is this:
```python
log_data=action_log_data.get(str(i), {})
```
So if the `log_data={}` I think this will give a key error.

SilviaAmAm (changes requested):

src/openforms/submissions/logic/actions.py
CharString: Oh yeah... It's when `eval` fails and the error is caught by the `log_errors`

SilviaAmAm (approved):

PR #2904, opened 2023-03-17
Add tests for openforms.api.validators

sergei-maertens:
* Fixes drop in test coverage due to the `ModelValidator`
* Added explicit tests for the `AllOrNoneRequiredFieldsValidator` which was drive-by coverage from other tests

### DIFF
--- a/src/openforms/api/validators.py
+++ b/src/openforms/api/validators.py
@@ -10,21 +10,28 @@ from rest_framework.fields import get_error_detail
 from .utils import get_model_serializer_instance
 
 
-class AllOrNoneRequiredFieldsValidator:
+class AllOrNoneTruthyFieldsValidator:
     """
-    Validate that the set of fields is present as soon as one field is provided.
+    Validate that either none or all fields have a 'non-empty' value.
 
-    Field values are checked to be truthy to determine if they are provided or not.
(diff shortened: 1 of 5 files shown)


src/openforms/api/validators.py
```diff
@@ -10,21 +10,24 @@
 from .utils import get_model_serializer_instance
 
 
-class AllOrNoneRequiredFieldsValidator:
+class AllOrNoneTruthyFieldsValidator:
```
sergei-maertens, line 13: I went the wrong way refactoring this and then the functional tests for the confirmation email template broke, so decided to rename the serializer class instead in hopes of making it clearer what the intent is.

src/openforms/api/validators.py
```diff
@@ -10,21 +10,24 @@
...
     """
-    Validate that the set of fields is present as soon as one field is provided.
+    Validate that either none or all fields have a 'non-empty' value.
```
SilviaAmAm, line 15: Maybe good to add a note about the bool behaviour too? The name of the validator already clarifies that this validates for truthy/falsy values, but maybe good to make it explicit also in the doc string?

SilviaAmAm (approved):

PR #2910, opened 2023-03-21
Fix focus shift/flicker in form builder

### LINKED ISSUE #2819: Key of password field doesn't automatically update

2.1.x

And the translations also seem to be broken

- Create a form with a password field
- Update the label
=> Notice how the key doesn't change and if you go to the translations tab, there are no translations

### LINKED ISSUE #2840: Dismissing form builder modal breaks translations tab

2.1.0-rc.0, latest

When dismissing the modal in the form builder by clicking the cross or clicking outside of the modal to close it, the translations tab data is no longer visible upon the next open of the modal. Clicking any of the buttons (Opslaan / Annuleren) behaves as expected.

Note that the underlying data is not gone - it appears to be purely a presentation issue. Also when the translations are not visible, if you close the modal by clicking Opslaan / Annuleren, upon the next open they are visible again.

1. Go to the form builder and drag a component into the drop zone
2. In the modal, click the translations tab and see that there are key-value pairs present
3. Save the component
4. Edit the component again and check the translations tab - they are still present
5. Close the modal by clicking outside of it or on the cross in the top right
6. Edit the component again and check the translations tab
7. Observe that the entries are no longer visible, but the table header is

Translations are still visible after closing the modal by any means other than clicking a button.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

### LINKED ISSUE #2889: Vrije tekst veld - focus shifts after change

2.1.0

Entering tekst in this component is made very difficult because after each "change" the field will lose focus, and the user needs to refocus to continue typing - just to lose focus again.

Peek 2023-03-16 11-15

sergei-maertens:
Fixes #2819 (better fix)
Fixes #2889
Fixes #2840

Essentially this moves the translation hooks into the custom WebformBuilder class which seems to solve a number of problems:

* Initialize the translations before any change event fires when the modal to edit the component opens
* Do tap into the change events to keep track of literals (which is now done on the `editForm` instance instead of a React ref)
* Fix a bug in formio where the `key` configuration/component is not re-rendered after setting a unique value (original issue is #2819)
* This patch *accidentally* fixes the form builder dismissal removing the translations state (#2840)

We are no longer relying on obscure `onChange` options of the formio form class (and having to call the original parent methods etc.), but instead limit this to pure builder functionality.

Hopefully this makes the whole approach a bit more stable, but it would still be beneficial to completely replace this form builder with something React-based where you don't have to reason as much about re-rendering after value/prop changes.

### DIFF
--- a/src/openforms/js/components/formio_builder/WebformBuilder.js
+++ b/src/openforms/js/components/formio_builder/WebformBuilder.js
@@ -2,29 +2,25 @@ import {Formio} from 'formiojs';
 import FormioUtils from 'formiojs/utils';
 import BuilderUtils from 'formiojs/utils/builder';
 import cloneDeep from 'lodash/cloneDeep';
+import get from 'lodash/get';
 import isEmpty from 'lodash/isEmpty';
+import set from 'lodash/set';
 
 import {getComponentEmptyValue} from 'components/utils';
+
+import {
+  addTranslationForLiteral,
+  handleComponentValueLiterals,
+  isTranslatableProperty,
+} from './translation';
 
 const WebformBuilderFormio = Formio.Builders.builders.webform;
 
 class WebformBuilder extends WebformBuilderFormio {
   constructor() {
-    let element, options;
-    if (arguments[0] instanceof HTMLElement || arguments[1]) {
-      element = arguments[0];
-      options = arguments[1];
-    } else {
-      options = arguments[0];
-    }
+    super(...arguments);
 
-    if (element) {
-      super(element, options);
-    } else {
-      super(options);
-    }
-
-    const defaultRequiredValue = options.evalContext.requiredDefault;
+    const defaultRequiredValue = this.options.evalContext.requiredDefault;
 
     // Update the component schema (used by the builder) to change the default value of the 'required' checkbox.
     // This cannot be done directly in the src/openforms/js/components/form/edit/tabs.js file, because the defaultValue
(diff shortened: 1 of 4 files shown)


src/openforms/js/components/formio_builder/WebformBuilder.js
```diff
@@ -72,9 +68,94 @@ class WebformBuilder extends WebformBuilderFormio {
...
+        Object.entries(translations).forEach(([languageCode, translationsForLanguage]) => {
+          if (!translationsForLanguage.length) return;
+          const {literal, translation} = translationsForLanguage[0];
+
+          // Can't get this to work with FormioUtils.getComponent... it's just skipping
+          // over the columns/hidden fields for some reason :(
+          let literalComponent;
+          let translationComponent;
+          this.editForm.everyComponent(_component => {
+            const prefix = `openForms.translations.${languageCode}[0].`;
```
SilviaAmAm, line 110: Since this trick of checking the prefix 'openForms.translations.` to make sure we are dealing with the translation component of the edit form is used multiple times, I wonder if it would be worth adding a custom property in the schema of the translation component along the lines of 'isTranslationComponent'?

src/openforms/js/components/formio_builder/WebformBuilder.js
```diff
@@ -218,33 +337,27 @@ class WebformBuilder extends WebformBuilderFormio {
...
-    if (!window.sessionStorage) {
-      return console.warn('Session storage is not supported in this browser.');
-    }
-    this.addClass(this.refs.form, 'builder-paste-mode');
-    window.sessionStorage.setItem(
-      'formio.clipboard',
-      JSON.stringify(this.fixDefaultValues(component))
-    );
+    const fixedComponent = {...component, schema: fixDefaultValues(component)};
+    return super.copyComponent(fixedComponent);
```
SilviaAmAm, line 341: I'm not sure anymore, but I think I had avoided calling super because it does:
`window.sessionStorage.setItem('formio.clipboard', JSON.stringify(component.schema));`
And the `component.schema` then overwrites the fixes that we make here?

sergei-maertens: I did break things in earlier versions, but while fixing that I think your concern is accounted for. The original version put the `fixedSchema` on the clipboard, this is still happening now because I mutate the component that's passed to the super call:

```js
const fixedComponent = {...component, schema: fixDefaultValues(component)};
```
so what gets put on the clipboard via `super.copyComponent(fixedComponent)` is essentially `JSON.stringify(fixDefaultValues(component))`

SilviaAmAm: Ahh yes, makes sense

Issue #2884, opened 2023-03-14
Documentation in manual for Service Fetch


stevenbal:
Currently there is no documentation in the manual yet for service fetch. I would expect it to be somewhere near `Manual > Formulieren > Dynamische formulieren > Acties`.

Example case: select a product to make and appointment and then select a timeslot.

Two examples, one simple one with a predefined config, one to create a new config

- [x] Show how to select service fetch as action
- [x] show images of modal and briefly describe
  - [ ] describe selecting existing config (copy vs update)
  - [x] describe creating config from scratch
  - [x] describe variable interpolation options
    - [x] for headers and query parameters, only values can contain variables 
- [x] Describe the experimental nature of this feature and that many features are still. missing "Due to the agile development, these features are currently not on the roadmap."
- [x] Check docs on import/export behaviour. Make sure the docs describe what is or is not exported and/or does not work.
- [x] Examples with dummy json (https://dummyjson.com/):
  - list of open Open Zaak zaken.
  - example get list of things from list endpoint and fill selectbox with IDs and display values
  - example from sprint demo (2 fields that depend on eachother)

PR #2914, opened 2023-03-21
[Draft] Feature/2835 service fetch modal save

stevenbal:


sergei-maertens:
> The things I tried to propagate the error into the modal were problematic; there is not enough information in the ValidationError to be able to mark the right spots. I think the validation error handling deserves its own ticket and shouldn't keep up this PR any more. It touches on the logging and the "try out" tab in terms of providing the form designer with the information they need in order to construct a complex form.

Since I've been using Formik, this whole form designer desperately needs a solution like that instead of rolling our own. Can you create a follow up ticket for the validation error display if there's none yet?

src/openforms/forms/api/drf_camel_case.py
```diff
@@ -17,7 +17,13 @@ def json_underscoreize(self):
...
            registration backend. Future improvement!
         """
-        ignore_fields = []
+        ignore_fields = [
+            # ServiceFetchConfiguration attributes
+            "body",
+            "headers",
+            "mapping_expression",
+            "query_params",
+        ]
```
sergei-maertens, line 26: I'm not a fan of this global approach for the form parser due to the risk of key collision in the future - I think it'd be better to define a separate parser/renderer for the form variables related endpoints, similarly to `IgnoreConfigurationFieldCamelCaseJSONParser` and then specify that parser class in the `parser_classes` argument for the `action` [decorator](https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing).

src/openforms/forms/admin/form_variable.py
```diff
@@ -33,4 +33,5 @@ class FormVariableAdmin(admin.ModelAdmin):
         "data_format",
         "is_sensitive_data",
         "initial_value",
+        "service_fetch_configuration",
```
sergei-maertens, line 36: can you add this field to the `raw_id_fields` as well? In general, FK fields should be listed there if you expect > 10 database records, or your admin page will get slower and slower over time to populate the dropdown.

src/openforms/forms/tests/e2e_tests/test_service_fetch.py
```diff
@@ -0,0 +1,587 @@
...
+
+    if not await page.locator('div.form-row.field-headers input[name="key"]').count():
+        await page.locator("div.form-row.field-headers span.addlink").click()
+    await page.locator('div.form-row.field-headers input[name="key"]').last.fill(
+        list(data["headers"].items())[0][0]
+    )
+    await page.locator('input[name="value"]').fill(list(data["headers"].items())[0][1])
+
+    # FIXME doesn't seem to work with `get_by_label`
+    await page.locator('textarea[name="body"]').fill(data["request_body"])
```
sergei-maertens, line 77: I think this works via the HTML `id` attribute on the textarea and the label `for` attribute. The textarea is missing the ID attribute.

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -247,6 +247,20 @@ const saveVariables = async (state, csrftoken) => {
       // in the backend
       if (variable.dataType === 'boolean' && typeof variable.initialValue != 'boolean')
         variable.initialValue = variable.initialValue === 'true';
+
+      if (variable.serviceFetchConfiguration) {
+        if (Array.isArray(variable.serviceFetchConfiguration.headers)) {
+          variable.serviceFetchConfiguration.headers = Object.fromEntries(
+            variable.serviceFetchConfiguration.headers || []
```
sergei-maertens, line 254:
```suggestion
            variable.serviceFetchConfiguration.headers
```
this code path is only entered when it's proven to be an array via the `Array.isArray` check, so it can't be a falsy value (can't be null, undefined, false or empty string)

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -247,6 +247,20 @@ const saveVariables = async (state, csrftoken) => {
...
+          );
+        }
+
+        if (Array.isArray(variable.serviceFetchConfiguration.queryParams)) {
+          variable.serviceFetchConfiguration.queryParams = Object.fromEntries(
+            variable.serviceFetchConfiguration.queryParams || []
```
sergei-maertens, line 260: similar remark as above

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -105,28 +107,62 @@ const ActionFetchFromService = ({action, errors, onChange}) => {
...
 
+  const formContext = useContext(FormContext);
+
+  let serviceFetchConfigFromVar =
+    _.cloneDeep(formContext.formVariables.find(element => element.key === action.variable))
+      ?.serviceFetchConfiguration || undefined;
+
+  if (serviceFetchConfigFromVar) {
+    if (!Array.isArray(serviceFetchConfigFromVar.headers)) {
+      serviceFetchConfigFromVar.headers = Object.entries(serviceFetchConfigFromVar.headers || []);
```
sergei-maertens, line 118:
```suggestion
      serviceFetchConfigFromVar.headers = Object.entries(serviceFetchConfigFromVar.headers || {});
```
this is supposed to convert records into two-tuples, no?

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -105,28 +107,62 @@ const ActionFetchFromService = ({action, errors, onChange}) => {
...
+    }
+
+    if (!Array.isArray(serviceFetchConfigFromVar.queryParams)) {
+      serviceFetchConfigFromVar.queryParams = Object.entries(
+        serviceFetchConfigFromVar.queryParams || []
```
sergei-maertens, line 123:
```suggestion
        serviceFetchConfigFromVar.queryParams || {}
```
same as above

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -105,28 +107,62 @@ const ActionFetchFromService = ({action, errors, onChange}) => {
     setIsModalOpen(false);
   };
 
+  const formContext = useContext(FormContext);
+
+  let serviceFetchConfigFromVar =
+    _.cloneDeep(formContext.formVariables.find(element => element.key === action.variable))
```
sergei-maertens, line 113: this should also filter the variables down to user defined variables like the dropdown/variable selection, no?

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -105,28 +107,62 @@ const ActionFetchFromService = ({action, errors, onChange}) => {
...
+              defaultMessage="Fetch configuration:"
+            />
+          </b>
+          &nbsp;
+          {serviceFetchConfigFromVar?.name ||
+            intl.formatMessage({
+              description: 'No service fetch configuration configured yet message',
+              defaultMessage: '(not configured yet)',
+            })}
+        </span>
```
sergei-maertens, line 164: let's give this a proper class name and do the styling with CSS instead of relying on `<b>` or `<strong>`. Additionally, the text and button are not properly vertically aligned (centered) because the span gets some bottom and right margins that shouldn't be there.

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.js
```diff
@@ -62,6 +62,27 @@ const ServiceFetchConfigurationForm = ({formik, selectExisting = false}) => {
...
+                    description="Service fetch configuration modal form name field label"
+                  />
+                }
+              >
+                <TextInput
+                  id="name"
+                  value={formik.values.name}
+                  maxLength="1000"
+                  {...formik.getFieldProps('name')}
+                />
```
sergei-maertens, line 82:
```suggestion
                <TextInput
                  id="name"
                  maxLength="1000"
                  {...formik.getFieldProps('name')}
                />
```
the `value` prop should already be included with `formik.getFieldProps`

Even cleaner would be:

```jsx
import {Field as FormikField} from 'formik';

<FormikField name="name" as={TextInput} id="name" maxLength="1000" />
```

The `as` prop allows specifying your own components, while any additional props are forwarded to the `as`-component. The `name` is taken and under the hood formik grabs all the relevant props.

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.js
```diff
@@ -62,6 +62,27 @@ const ServiceFetchConfigurationForm = ({formik, selectExisting = false}) => {
...
+              </Field>
+            </FormRow>
+
             <FormRow>
               <Field
                 name={'method'}
```
sergei-maertens, line 88:
```suggestion
                name="method"
```
style convention with react (string) props -> just use double quotes

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationPicker.js
```diff
@@ -114,8 +125,10 @@ const ServiceFetchConfigurationPicker = ({onChange, onFormSave}) => {
 };
 
 ServiceFetchConfigurationPicker.propTypes = {
+  variableName: PropTypes.string.isRequired,
   onChange: PropTypes.func.isRequired,
   onFormSave: PropTypes.func.isRequired,
+  initialValues: PropTypes.object,
```
sergei-maertens, line 131:
```suggestion
  initialValues: PropTypes.shape({...}),
```
we know ahead of time what the shape of this object is, right? So we should be able to make a strict and explicit proptype

src/openforms/submissions/logic/actions.py
```diff
@@ -231,17 +231,19 @@ def eval(
...
         log: Callable[[JSONValue], None],
     ) -> DataMapping:
-        # XXX perform_service_fetch is expressed in terms of FormVariables
-        dummy_var = FormVariable(
-            name=self.variable,
-            service_fetch_configuration=ServiceFetchConfiguration.objects.get(
-                pk=self.fetch_config
-            ),
-        )
+        if self.fetch_config:  # the old way
```
sergei-maertens, line 234: ~~can't we make a data migration so this branch of code can be deleted?~~

... we need to support old version export/import... Is there a test for old format imports & ensuring service fetch still works? I'd prefer having the compatibility layer in the import code/serializer rather than "runtime" code if that's possible (but that should be a separate issue/PR).

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.js
CharString: That `FormikField` seems to need some `Formik` component and a `FormikContext`. I went with just removing the `value` attribute.

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
CharString: No, this gets the ServiceFetchConfig from the variable set on the action passed in. The configs are an attribute on the variable of the action... The initial release, the action could change any compontent value and the backend created a faux variable, just to evaluate the fetch action. That's changed now, as the configs are now actually bound to a user defined variable as originally specced out... So yes; filtering is done elsewhere to only allow picking of user defined variables such that `action.variable` here is always in that set.

:thinking: I need to test that case when writing #2884: in the current release create a fetch action that sets the choices of a selectbox or other component directly, and see how it migrates and behaves in the new release.

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
CharString: This works too because `Object.prototype.isPrototypeOf([]) === true;` :wink:

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
sergei-maertens: I think I understand :sweat_smile:

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
sergei-maertens: I get why it works, but I don't particularly like relying on quirks like `typeof [] === 'object'`. I'm more concerned about the way this code can be read:

1. default value is an array
2. so this could be read as the implication that `serviceFetchConfigFromVar.headers` is expected to be an array
3. which doesn't hold up with the check above that says it's not an array

this is actually what triggered me to make this remark in the first place - the if-branch looked out of place because if `serviceFetchConfigFromVar.headers` already always is an array, then this would be a dead code branch that never evaluates.

So following the logic where the variable is NOT an array, then what is it expected to be? -> I can only think of an object (excluding arrays and null) and I'd like to see that type consistency in the fallback value too.

src/openforms/js/components/admin/form_design/variables/ServiceFetchConfigurationForm.js
sergei-maertens: ah right, this works differently then because this case uses the `useFormik` hook.

I don't like that formik offers three ways to do the same thing and that we now use all three smeared out over this repo, the SDK repo and the formio-builder/formio-renderer repo's :thinking: 

I'll keep it in the back of my head to clean up at some point and review more strictly. Removing the value attribute is sufficient for this PR :+1:

src/openforms/submissions/logic/actions.py
CharString: Yeah, fixing this cleanly felt like a rabbit hole and I went for the `if`. I didn't expect codecov to find the `else` path uncovered. I remember changing the code to pass tests and leaving the `if` because of exports and databases containing the old scheme. I've already linked  #2884  to this for a testing a similar "old style" use.

sergei-maertens: Can you then create an issue to move this shim to the import code/serializer and add a `FIXME` with a link to the issue?

src/openforms/js/components/admin/forms/Select.js
```diff
@@ -48,15 +48,15 @@ const Message = PropTypes.shape({
 });
 
 const Label = PropTypes.oneOfType([PropTypes.string, Message]);
+const Value = PropTypes.oneOfType([PropTypes.string, PropTypes.number]);
+
+// Not excatly the 2-tuple we mean, but hey...
```
sergei-maertens, line 53: Typescript will save us when we eventually get to refactoring all this :)

Issue #2911, opened 2023-03-21
Problems uploading .heic files


SilviaAmAm:
### Product versie / Product version

2.1.x

### Omschrijf het probleem / Describe the bug

DH Taiga issue 317

There seem to be issues with uploading `.heic` files in the file upload component. Files provided by den haag (see DH taiga)
- sample1.heic 
- sample2.heic

=> In a file component where all file types are allowed for upload, only `sample1.heic` is accepted (`sample2.heic` is seen as a `.heif` file)
=> In a file component where only `.heic` file types are allowed, neither file is accepted. It looks like in the formio widget the mime type for both files is guessed to be `image/heif` and so both fail the validation.

Extra note:
The error messages are escaped when they shouldn't be?
![Screenshot from 2023-03-21 10-22-25](https://user-images.githubusercontent.com/19154114/226567241-82fb139e-34f5-413b-a286-af6ce7f6bf99.png)

### Stappen om te reproduceren / Steps to reproduce

Create a form with
- An upload component with only `.heic` files accepted
- An upload component with all file types accepted

Fill in the form with the files provided above

### Screen resolution

None

### Device

None

### OS

None

CharString:
Refinement:
What does DH need exactly to get the Ooievaarspas life?

From what I read from [wiki](https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format), heif is a container which may contain an image that uses heic compression.
Both provided files are HEIF files, and only from sample2.heic, libmagic can determine that the contents uses heic compression... 

```py
In [1]: import magic

In [2]: sample2 = open("sample2.heic", "rb").read()

In [3]: sample1 = open("sample1.heic", "rb").read()

In [4]: magic.detect_from_content(sample1)
Out[4]: FileMagic(mime_type='image/heif', encoding='binary', name='ISO Media, HEIF Image')

In [5]: magic.detect_from_content(sample2)
Out[5]: FileMagic(mime_type='image/heic', encoding='binary', name='ISO Media, HEIF Image HEVC Main or Main Still Picture Profile')
```

Asked differently, can they process any heif file? And is it enough to let any heif file with a .heic extension through?

CharString:
To be clear: I haven't looked into the double escaping of the error message.

CharString:
["There is no elegant solution"](https://stackoverflow.com/questions/58269428/i18next-react-trans-component-escaping-characters-correctly) fixing the escaping dives through FormIO into i18next. I vote for "wontfix"

PR #2915, opened 2023-03-21
Assume image/heif and image/heic are the same.

CharString:
Part of a fix for #2911 

heif is a container format. heic is a heif file with a certain type of compressed file in there.

libmagic doesn't recognize all heics as such, nor can PIL/pillow in dialogue with the product owner, this solution suffices for the usecase.

Alternative would be to do all of the following:
- add heif explicitly to the allowed filetype options which isn't great UX;
- install extra dependency (like wand or a pillow file reader) to inspect the whole file.

### DIFF
--- a/src/openforms/formio/tests/test_validators.py
+++ b/src/openforms/formio/tests/test_validators.py
@@ -140,6 +140,30 @@ class MimeTypeValidatorTests(SimpleTestCase):
         ):
             validator(self.CORRECT_GIF)
 
+    def test_mif1_brand_heif_files_are_acceptable_heic(self):
+        # lib magic has a hard time recognizing the HEVC is used and a heif container actutally is heic
+        validator = validators.MimeTypeValidator({"image/heic"})
+        sample_1 = SimpleUploadedFile(
+            "sample1.heic", b"\x00\x00\x00\x18ftypmif1", content_type="image/heif"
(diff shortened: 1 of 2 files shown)


src/openforms/formio/tests/test_validators.py
```diff
@@ -164,6 +164,16 @@ def test_heic_brand_heif_files_are_recognized_as_heic(self):
...
+    def test_heic_brand_heif_files_are_not_recognized_as_png(self):
+        # lib magic has a hard time recognizing the HEVC is used and a heif container actutally is heic
+        validator = validators.MimeTypeValidator({"image/png"})  # accept any
+        sample_2 = SimpleUploadedFile(
+            "sample2.heic", b"\x00\x00\x00\x18ftypheic", content_type="image/heif"
+        )
+
+        with self.assertRaises(ValidationError):
+            validator(sample_2)
+
```
CharString, line 176: I expected this to return a 400 on TemporaryFileUpload. But it doesn't.
So right now you can upload a heic in a "image/png" field and on going to the next step it will mark the field as invalid. :shrug: Works. But I had the wrong expectation myself.

SilviaAmAm (approved):

sergei-maertens (approved):

PR #2932, opened 2023-03-24
[#2716] include co-sign identifier (bsn) in submission report

### LINKED ISSUE #2716: Allow fields to be filled with co-sign information
### Thema / Theme

Plugins

SED: 34

These points were not completed in the PVE: https://github.com/open-formulieren/open-forms/issues/296#:~:text=Add%20Co%2DSign,TODO%3A%20how%20exactly%3F

co-sign data is available in the submission and could be available to use as prefill values.
This will allow the form builder to show the co-signers information in the summary and registration systems.

_No response_

_No response_

pi-sigma:
Closes #2716 

* the identifier (bsn) of form co-signers is included in the submission report, alongside first + last name

### DIFF
--- a/src/openforms/submissions/report.py
+++ b/src/openforms/submissions/report.py
@@ -6,7 +6,9 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 from django.utils.safestring import SafeString
+from django.utils.translation import gettext_lazy as _
 
+from openforms.authentication.constants import AuthAttribute
 from openforms.forms.models import Form
 
 if TYPE_CHECKING:
@@ -36,15 +38,26 @@ class Report:
 
     @property
     def co_signer(self) -> str:
-        if not self.submission.co_sign_data:
+        """Retrieve and normalize data about the co-signer of a form"""
+
+        if not (co_sign_data := self.submission.co_sign_data):
             return ""
 
-        if not (co_signer := self.submission.co_sign_data.get("representation", "")):
+        representation = co_sign_data.get("representation") or ""
+        identifier = co_sign_data["identifier"]
+        co_sign_auth_attribute = co_sign_data["co_sign_auth_attribute"]
+        auth_attribute_label = AuthAttribute[co_sign_auth_attribute].label
+
+        if not representation:
             logger.warning(
                 "Incomplete co-sign data for submission %s", self.submission.uuid
             )
 
-        return co_signer
+        return _("{representation} ({auth_attribute}: {identifier})").format(
+            representation=representation,
+            auth_attribute=auth_attribute_label,
+            identifier=identifier,
+        )
 
     @property
     def confirmation_page_content(self) -> SafeString:
--- a/src/openforms/submissions/serializers.py
+++ b/src/openforms/submissions/serializers.py
@@ -4,6 +4,7 @@ type model fields.
 """
 from rest_framework import serializers
 
+from openforms.authentication.constants import AuthAttribute
 from openforms.authentication.registry import register
 
 
@@ -11,6 +12,7 @@ class CoSignDataSerializer(serializers.Serializer):
     plugin = serializers.ChoiceField(choices=())
     identifier = serializers.CharField()
     representation = serializers.CharField(required=False, allow_blank=True, default="")
+    co_sign_auth_attribute = serializers.ChoiceField(choices=AuthAttribute.choices)
     # TODO: validate fields shape depending on value of plugin (polymorphic serializer)
     fields = serializers.DictField(
         child=serializers.CharField(allow_blank=True),
--- a/src/openforms/submissions/tests/test_tasks_pdf.py
+++ b/src/openforms/submissions/tests/test_tasks_pdf.py
@@ -222,6 +222,7 @@ class SubmissionReportCoSignTests(TestCase):
             submission__co_sign_data={
                 "plugin": "digid",
                 "identifier": "123456782",
+                "co_sign_auth_attribute": "bsn",
                 "fields": {
                     "voornaam": "Tina",
                     "geslachtsnaam": "Shikari",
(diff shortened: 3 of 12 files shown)


src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -243,4 +245,74 @@ def test_cosign_data_included(self):
...
+            <div class="submission-step-row">
+                <div class="submission-step-row__label">{key}</div>
+                <div class="submission-step-row__value">123456782</div>
+            </div>
+            """,
+            key=_("Identifier of co-signer"),
+        )
+        self.assertInHTML(expected_identifier, rendered, count=1)
+
+    def test_incomplete_cosign_data(self):
```
CharString, line 262: I would split this into two tests... But first check with @joeribekker or @LaurensBurger if this is what the user story describes. I read the ticket as: require a BSN and show it + name if we know it.

SilviaAmAm (changes requested):

pi-sigma:
A few tests needed refactoring after the changes. Most are straightforward, but there's one about which I'm not entirely sure: `test_co_sign_flow` in [src/openforms/authentication/tests/test_views](https://github.com/open-formulieren/open-forms/pull/2932/files#diff-c21b699bcaa2edeb83aac742a4e940ea88a4ff0a813f6cf58502d7403cfedb1d).

In order to make the test pass, I changed the `provides_auth` attribute of `authentication.tests.mocks.Plugin` to "bsn" instead of "dummy". I also removed the mock. The test passes with or without it, and I didn't see the point of it.

src/openforms/submissions/report.py
```diff
@@ -15,6 +16,35 @@
...
+    def representation(self) -> str:
+        return self.co_sign_data.get("representation", "")
+
+    @property
+    def identifier(self) -> str:
+        return self.co_sign_data.get("identifier", "")
+
+    @property
+    def auth_attribute(self) -> str:
+        return self.co_sign_data.get("co_sign_auth_attribute", "")
```
sergei-maertens, line 33: this should use the human readable label of the `AuthAttributes` choices instead of the raw string value:

```
class AuthAttribute(models.TextChoices):
    """
    core identifying attributes retrieved from the authentication plugins
    """

    bsn = "bsn", _("BSN")
    kvk = "kvk", _("KvK number")
    pseudo = "pseudo", _("Pseudo ID")
    employee_id = "employee_id", _("Employee ID")
```

there's some sort of `AuthAttribute.labels` helper available, check the django docs :)

src/openforms/submissions/report.py
```diff
@@ -15,6 +16,35 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class CoSigner:
+    co_sign_data: dict
+
+    @property
+    def representation(self) -> str:
+        return self.co_sign_data.get("representation", "")
```
sergei-maertens, line 25: in general the patern `self.co_sign_data.get("...") or ""` is better to guarantee the type is correct if the key is absent or empty-ish (i.e. present but the value is `None`)

maybe a better approach would be a complete dataclass and have the `Report.co_signer` perform this normalization

src/openforms/submissions/report.py
```diff
@@ -35,11 +65,13 @@ def show_payment_info(self) -> bool:
...
 
     @property
-    def co_signer(self) -> str:
-        if not self.submission.co_sign_data:
-            return ""
+    def co_signer(self) -> CoSigner | None:
+        if not (co_sign_data := self.submission.co_sign_data):
+            return None
+
+        co_signer = CoSigner(co_sign_data)
```
sergei-maertens, line 72: this should be something more along the lines of:

```py
co_signer = CoSigner(identifier=co_sign_data.get('identifier") or "", representation=co_sign_data.get("representatin") or "", ...)
```
-> do the normalization in this central entrypoint to make the dataclass strongly and completely typed instead of dealing with flexible types like dicts inside of them.

src/openforms/submissions/report.py
```diff
@@ -15,6 +16,35 @@
...
+        if not self.identifier:
+            return _(
+                "{representation} (missing {auth_attribute} for co-signer)"
+            ).format(
+                representation=self.representation, auth_attribute=self.auth_attribute
+            )
+        return f"{self.representation} ({self.auth_attribute}: {self.identifier})"
+
+    def __repr__(self):
+        return f"CoSigner(co_sign_data={self.co_sign_data}, auth_attribute={self.auth_attribute})"
```
sergei-maertens, line 45: these are typically generated by the dataclass and there should not be a need to override them.

If you need a custom/particular string representation, implement them as a custom method and call that in the template

sergei-maertens (changes requested):

pi-sigma:
I got rid of the dataclass for the co-signer, after all (it's not used elsewhere).

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -222,25 +222,56 @@ def test_cosign_data_included(self):
...
+
         expected = format_html(
             """
             <div class="submission-step-row">
                 <div class="submission-step-row__label">{key}</div>
-                <div class="submission-step-row__value">T. Shikari</div>
+                <div class="submission-step-row__value">T. Shikari (BSN: 123456782)</div>
             </div>
             """,
             key=_("Co-signed by"),
```
sergei-maertens, line 245: this value should also use the `gettext(...)`call otherwise if the shape changes in translations, this test will fail

src/openforms/submissions/serializers.py
```diff
@@ -21,3 +23,6 @@ class CoSignDataSerializer(serializers.Serializer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["plugin"].choices = register.get_choices()
+
+    # def get_co_sign_auth_attribute(self, obj):
+    #     return obj.get_co_sign_auth_attribute_display()
```
sergei-maertens, line 28:
```suggestion
```

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Backported via 050db5356252d5624f4209c16cbbefd4dec9aded

PR #2943, opened 2023-03-29
[#2942] Fix undefined auth attributes

### LINKED ISSUE #2942: Auth attribute undefined in edit form for Cosing component

2.1.x

auth_attributes_undefined

Create a form with:
- A cosign component => When selecting the auth plugin notice the "provides: undefined"

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2942 

Now it looks like this (I added a KvK to the first demo plugin to test how it looks when there are multiple auth attributes):
Screenshot from 2023-03-29 12-10-44

I don't really understand why the .join() syntax wasn't working, or why the list gets automatically formatted now :thinking:

### DIFF
--- a/src/openforms/js/components/form/coSign.js
+++ b/src/openforms/js/components/form/coSign.js
@@ -34,7 +34,7 @@ const EDIT_FORM_TABS = [
               url: getFullyQualifiedUrl('/api/v2/authentication/plugins'),
             },
             valueProperty: 'id',
-            template: `<span>{{ item.label }}, provides: {{ item.providesAuth.join(', ') }}</span>`,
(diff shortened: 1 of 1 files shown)


CharString (approved):
I try not to try to hard to understand behaviour in files with a filename that ends in ".js". :wink:

SilviaAmAm:
This is indeed an array of strings. 

I also don't like the lack of space, but we currently don't have any plugins that return multiple identifiers and I don't really understand what is going on with the syntax, so I went for just fixing the `undefined` for now :shrug:

PR #2948, opened 2023-03-30
[#2945] Prevent updating logic if the variable had an empty string

### LINKED ISSUE #2945: Adding a user variable After adding logic rule removes variable from logic rule

2.x.x

create form with step add a field
create simple rule and save form
image
go to variables > user variables and add random variable:
image
result:
image

After refreshing the logic seems to be back to the original state (before adding the user variable)

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #2945 

This is really a bandaid, because it is still possible to leave a variable with empty label/key and then create a logic rule with the empty variable (and then if you change its label, the corresponding rule won't be updated). So I guess at some point we need to fix this too, but I think this is less urgent


CharString (approved):

SilviaAmAm:
backported in 14dfaaeb8f66297000c979c97127f30ee6b950d5 (2.0.x) and a0e72edc7b18b7bffb3329b1a1390c40bf916300 (2.1.x)

PR #2954, opened 2023-03-31
[#2947] Fix missing custom errors in number components

### LINKED ISSUE #2947: Number component is missing custom errors

2.1.x

Since https://github.com/open-formulieren/open-forms/issues/2534 it was possible to configure custom errors for each component.

However, it looks like these are no longer being populated with the right errors.

Create a form with:
-  A number component with a max value
- Go to the "validation" tab and click on the 'custom errors'
- The 'max' validator is not present and for some reason there are 2x required fields:
Screenshot from 2023-03-30 14-20-29

It should be like here: https://github.com/open-formulieren/open-forms/pull/2540 where there are multiple validation messages

### Workaround

The custom error can be added directly to the JSON of the component instead of through the UI

SilviaAmAm:
Fixes #2947 

Should be backported to 2.1.x


sergei-maertens (approved):

sergei-maertens:
Backport: 7198947e4a52c6d6b01198d7d052cbd48f21b29b

PR #2960, opened 2023-04-03
Implement type inference

### LINKED ISSUE #2959: Prototype the static analysis
* [x] Implement a type inference algorithm preferably M as it was [proven](https://dl.acm.org/doi/10.1145/291891.291892) to terminate earlier than 𝒲 and give more pinpointed errors
* [x] Implement a function `parse` that takes a JsonLogic expression and returns an expression that `M()` can take.
* [x] Implement `type_check`
* [ ] support the variadic versions of some JsonLogic operators (e.g. "<", "+", "in")
* [x] support return type polymorphism*

*return type polymorphism in python notation:
```python
T1 = TypeVar("T1")
T2 = TypeVar("T2")
def if_operator(condition: Bool, then: T1, else_:T2) -> Union[T1, T2]:
    ...
```

CharString:
Closes #2959


sergei-maertens:
Cleaning up clutter from open PRs - sorry!

PR #2964, opened 2023-04-05
[#2900] Frontend logic checkboxes

### LINKED ISSUE #2900: Component made visible by checkbox being true are emptied on _check_logic

2.1.0
Reported by: Utrecht, Horst a/d Maas

Only occurs for when made visible by checkbox (still looking into Choices)

Peek 2023-03-17 09-21
Peek 2023-03-17 09-28

If the component that was shown by checkbox is in a different group, this bug does not occur:
Peek 2023-03-17 10-00

SilviaAmAm:
Fixes #2900 (I think)

### DIFF
--- a/src/openforms/formio/utils.py
+++ b/src/openforms/formio/utils.py
@@ -318,11 +318,6 @@ def is_visible_in_frontend(component: Component, data: DataMapping) -> bool:
     trigger_component_value = glom(data, trigger_component_key, default=None)
     compare_value = conditional.get("eq")
 
-    # Issue #2900 - The values of checkboxes are bools, but in the Formio frontend logic, the compare value is a
-    # string ('true'/'false'). In Formio they cast it to a bool.
-    if isinstance(trigger_component_value, bool):
-        compare_value = compare_value == "true"
-
     if (
         isinstance(trigger_component_value, dict)
         and compare_value in trigger_component_value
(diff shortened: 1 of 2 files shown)


CharString (commented):
What about the fieldgroup aspect mentioned in the bug report?

I don't see how there could be a difference in behaviour...

I think this fixes it too. But what when a checkbox hides the whole fieldgroup? Before this fix, it would accidentally do the right thing, but this fix would break that case? :thinking:

SilviaAmAm:
I added a comment at the bottom of the issue with the behaviour I saw, but I couldn't reproduce the difference between the fields inside/outside the fieldset. I don't think that it did the right thing before this fix: in the case of the checkbox hiding a whole fieldset when the checkbox is checked, the frontend would think that the group is visible, but the backend thinks that it is still hidden, so clears all the fields inside the group. This then results in the answers disappearing in the frontend.

src/openforms/formio/utils.py
```diff
@@ -318,6 +318,11 @@ def is_visible_in_frontend(component: Component, data: DataMapping) -> bool:
     trigger_component_value = glom(data, trigger_component_key, default=None)
     compare_value = conditional.get("eq")
 
+    # Issue #2900 - The values of checkboxes are bools, but in the Formio frontend logic, the compare value is a
+    # string ('true'/'false'). In Formio they cast it to a bool.
+    if isinstance(trigger_component_value, bool):
+        compare_value = compare_value == "true"
```
sergei-maertens, line 324: for readability sake, I think this should go under the next if clause because that one is doing a similar isinstance check, but with dict - so these two if branches now look like they can both be executed, but that's not the case! Only one can ever match.

The compare value can also be `false` to display/hide something when the checkbox is unchecked - which is confusing me in reading this line. So, maybe we can make this a bit more easy to reason about with something like:

```py
compare_value = {"true": True, "false": False}.get(compare_value, False)
```
which imo highlights the nature of mapping string booleans to actual booleans.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #2979, opened 2023-04-11
:whale: [#2978] -- Fix podman incompatibility.

### LINKED ISSUE #2978: Dockerfile does not build in rootless docker alternative "Podman"

2.1.1

The Dockerfile does not work with Podman, an alternative to Docker that is rootless and used at a customer of Open Forms. The main issues seems to be that certain `VOLUME`/`COPY` instructions result into permissions being set to the root user, which can be mitigated by setting the `USER` earlier in te process and passing the `--chown=maykin` flag to the `COPY` statement copying the source.

 - Install `Podman`
 - Clone repository.
 - Run `podman build .` in the project directory.

Should build.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

svenvandescheur:
Closes #2978

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,6 +283,18 @@ jobs:
       target_env: ${{ matrix.target.env }}
       extensions: ${{ matrix.target.extensions }}
 
+  podman_build:
+    name: Build podman image
+    runs-on: ubuntu-latest
+    container:
+      image: mattermost/podman:1.8.0
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build the Podman image
+        run: |
+          podman build .
+
   oas-up-to-date:
     needs: tests
     name: Check for unexepected OAS changes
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):
If we're going to support podman, this needs a CI job that proves the failure and fix, otherwise we have no guarantees this won't be broken again accidentally in the future

joeribekker:
> If we're going to support podman, this needs a CI job that proves the failure and fix, otherwise we have no guarantees this won't be broken again accidentally in the future

See my comment in the ticket.

In essence, we can make it podman compatible but we will not provide support on it. If desired, we can explicitely mention this in the documentation.

sergei-maertens:
> > If we're going to support podman, this needs a CI job that proves the failure and fix, otherwise we have no guarantees this won't be broken again accidentally in the future
> 
> See my comment in the ticket.
> 
> In essence, we can make it podman compatible but we will not provide support on it. If desired, we can explicitely mention this (works but not supported) in the documentation.

agree - but I still want some automated guarantee that we can build with podman if we implement this patch. We do periodically restructure the dockerfiles according to new insights and this is definitely something that would break.

sergei-maertens:
Closing this as we currently don't have any need for it, nor does the stakeholder anymore.

PR #2985, opened 2023-04-13
Fix broken StUF registration for some integrations

### LINKED ISSUE #2983: stuf_zds registration broken: ValueError not enough values to unpack (expected 2, got 1)

2.1.1
Sentry: 324640 
Haarlemmermeer: 17

create_zaak_identificatie
create_document_identificatie 
Both expect 2 values: _, xml
while "execute_call" only returns 1 "xml"
https://github.com/open-formulieren/open-forms/blob/ba53c5554fe5a08695c68db4d318123a7f563c5a/src/stuf/stuf_zds/client.py#L174 
https://github.com/open-formulieren/open-forms/blob/ba53c5554fe5a08695c68db4d318123a7f563c5a/src/stuf/stuf_zds/client.py#L246

Which breaks stuf_zds registration

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Fixes #2983

Short summary: our tests (and other integrations apparently) do return the following structure:

```
Envelope
  Header
  Body
```

which the destructuring accepts, as the root element only has two children.

However, for this particular bug, the response has the structure:

```
Envelope
  Body
```

which breaks unpacking.

Unpacking was never intentional here, so that that's being fixed and a regression test is added without the header element.

### DIFF
--- a/src/stuf/stuf_zds/client.py
+++ b/src/stuf/stuf_zds/client.py
@@ -10,7 +10,7 @@ from django.utils.translation import gettext_lazy as _
 
 import requests
 from lxml import etree
-from lxml.etree import Element
+from lxml.etree import _Element
 from requests import RequestException
 
 from openforms.config.models import GlobalConfiguration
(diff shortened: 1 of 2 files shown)


sergei-maertens:
Needs backporting to 2.1.x where the refactors happened, 2.0.x and older are not affected.

SilviaAmAm (approved):

sergei-maertens:
Backported via: d1167fa19bea2a5bb8209a234617aafba58a2a08

PR #2989, opened 2023-04-14
Issue/2963 config templates

### LINKED ISSUE #2963: Updating to 2.1.1 resets the general config templates to their defaults

2.1.1

From @LaurensBurger:

> in general config their customized confirmation templates were replaced with the default OF variant (NL) and the buttons were all EN.
> The buttons is not a big issue, but the confirmations being replaced is.

First thoughts by @sergei-maertens:

> it could be a data-migration thing that we missed since a number of database columns got "replaced" with adding translations

Not reproduced yet.

Existing custom templates/button names are kept after migration.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

pi-sigma:
Fixes #2963


sergei-maertens (changes requested):
For completeness sake, discussed in office:

1. no need to insert the default config, you can just create an instance in memory (probably)
2. no need to loop over all config instances, you can just use `.first()` as there will only ever be one record
3. add some comments to explain the logic between comparing defaults
4. add a comment explaining the original field (from before model-translation) is still in the DB and this is picked up by the migrations without the aliasing from modeltranslation

pi-sigma:
Closing because we apply to patch to stable/2.1.x (see #2990)

PR #2990, opened 2023-04-14
[#2963] fix global configuration templates

### LINKED ISSUE #2963: Updating to 2.1.1 resets the general config templates to their defaults

pi-sigma:
Fixes #2963


sergei-maertens (approved):

sergei-maertens:
Applied to master via 1c7bbc50319cb2a45d52033540d202b9991263b9

PR #2992, opened 2023-04-14
[#2946] Make confirmation email consistent

### LINKED ISSUE #2946: Make it possible to overwrite the confirmation email subject/content independently
### Thema / Theme

Plugins

Currently, there is a global configuration for the subject and the content of the confirmation email and it is possible to overwrite them at the form level.

It is only possible to overwrite both the subject AND the content at the form level, but not only the content or only the subject.

For registration emails, it is possible to overwrite them independently, so it would be good to have the same behaviour for confirmation emails.

_No response_

_No response_

SilviaAmAm:
Fixes #2946 

Changes:
- Removed form attribute `confirmation_email_option` and replaced it with `send_confirmation_email` (added migration to deal with the change)
- Removed validator from serializer that requires subject/content to both be filled
- Updated frontend


src/openforms/emails/models.py
```diff
@@ -19,12 +19,7 @@ def set_for_form(self, form: Form, data: Optional[dict]):
 
 
 class ConfirmationEmailTemplate(models.Model):
-    """Template of email to be sent on completion of submission
-
-    Subject and content fields are validated independently. If the confirmation email
-    is specific to a particular form, both fields must be filled in or `is_usable` will
-    return `False`.
-    """
+    """Template of email to be sent on completion of submission"""
```
sergei-maertens, line 22: the `is_usable` method can be removed, no?

src/openforms/forms/migrations/0076_move_confirmation_email_setting.py
```diff
@@ -0,0 +1,55 @@
...
+    Form = apps.get_model("forms", "Form")
+
+    forms = Form.objects.all()
+
+    for form in forms:
+        form.send_confirmation_email = (
+            form.confirmation_email_option != ConfirmationEmailOptions.no_email
+        )
+
+    Form.objects.bulk_update(forms, fields=["send_confirmation_email"])
```
sergei-maertens, line 19:
```py
from django.db.models import Case, When, Value
Form.objects.update(
    send_confirmation_email=Case(
        When(confirmation_email_option=Value(ConfirmationEmailOptions.no_email), then=False), 
        default=True
    )
)
```
single query to do the update at the database level rather than python, see https://docs.djangoproject.com/en/3.2/ref/models/conditional-expressions/#advanced-queries

src/openforms/forms/migrations/0076_move_confirmation_email_setting.py
```diff
@@ -0,0 +1,55 @@
...
+        if hasattr(form, "confirmation_email_template") and bool(
+            form.confirmation_email_template.subject
+            and form.confirmation_email_template.content
+        ):
+            form.confirmation_email_option = (
+                ConfirmationEmailOptions.form_specific_email
+            )
+            continue
+
+        form.confirmation_email_option = ConfirmationEmailOptions.global_email
```
sergei-maertens, line 42: the reverse can also be done with a single Case/When query I think

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -442,7 +442,6 @@ async def test_editing_content_translations_are_saved(self):
             await expect(page.locator("css=.formio-dialog-content")).to_be_visible()
 
             default_string = page.locator("css=.ck-editor__editable").nth(0)
-            # import bpdb; bpdb.set_trace()
```
sergei-maertens, line 445: :see_no_evil:

sergei-maertens (changes requested):
Nice, very minor remarks :)

src/openforms/emails/models.py
SilviaAmAm: I removed it and updated some test that used it

src/openforms/forms/migrations/0076_move_confirmation_email_setting.py
SilviaAmAm: Oooh I had tried to do it in one query with the `F()` but had not succeeded! Nice

src/openforms/forms/migrations/0076_move_confirmation_email_setting.py
SilviaAmAm: I couldn't get this one to work with one query :( I added an alternative, not sure if it's better

sergei-maertens (approved):

src/openforms/forms/migrations/0076_move_confirmation_email_setting.py
sergei-maertens: worksforme

PR #3001, opened 2023-04-17
[#2895] fix admin inline colors

### LINKED ISSUE #2895: Inline colours for content field doesn't work if you are logged in as admin

2.1.0

Explanation:
 - This happens only if you are logged in as admin. 
 - [This code](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/api/serializers/form_definition.py#L117) is ran multiple times if you are logged in as admin:
```python
rewrite_formio_components_for_request(
      instance.configuration_wrapper,
      request=self.context["request"],
)
```
 -  This results in the inline style being converted to the `style` tag (with CSP nonce), and in the subsequent calls the `style` tag is removed from the HTML (resulting in no colour).
 - This is because when we are admin, we have the extra the `admin_only_fields` which include `translations` (see code [here](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/api/serializers.py#L139 )). This then causes [this super call](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/api/serializers/form_definition.py#L108) to run [this code](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/translations/api/serializer_helpers.py#L11) from the `TranslationsFieldMixin`. The `FormDefinitionTranslationsSerializer` calls the `to_representation` of the `FormDefinition` serializer and so the `rewrite_formio_components_for_request` is called multiple times.

- Create a form with a content field with colour red
- Fill in the form when you are an admin
- Fill in the form when you are not logged in as admin
- Observe the difference

Both admin and non-admin should see the same thing

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

pi-sigma:
Fixes #2895

### DIFF
--- a/src/openforms/formio/dynamic_config/__init__.py
+++ b/src/openforms/formio/dynamic_config/__init__.py
@@ -44,7 +44,10 @@ def rewrite_formio_components_for_request(
     Loop over the formio configuration and inject request-specific configuration.
     """
     for component in configuration_wrapper:
-        register.update_config_for_request(component, request=request)
+        # prevent multiple overwrites of content components (bandaid for #2895)
+        html = component.get("html", None)
+        if component["type"] != "content" or (html and "nonce" not in html):
+            register.update_config_for_request(component, request=request)
     return configuration_wrapper
 
 
--- a/src/openforms/formio/dynamic_config/__init__.py
+++ b/src/openforms/formio/dynamic_config/__init__.py
@@ -44,10 +44,7 @@ def rewrite_formio_components_for_request(
     Loop over the formio configuration and inject request-specific configuration.
     """
     for component in configuration_wrapper:
-        # prevent multiple overwrites of content components (bandaid for #2895)
-        html = component.get("html", None)
-        if component["type"] != "content" or (html and "nonce" not in html):
-            register.update_config_for_request(component, request=request)
+        register.update_config_for_request(component, request=request)
     return configuration_wrapper
 
 
--- a/src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
+++ b/src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
@@ -1,9 +1,20 @@
-from django.test import TestCase, override_settings
+from django.test import TestCase, override_settings, tag
+from django.urls import reverse
 
+from pyquery import PyQuery as pq
+from rest_framework.test import APITestCase
+
+from openforms.accounts.tests.factories import SuperUserFactory
 from openforms.formio.datastructures import FormioConfigurationWrapper
 from openforms.formio.dynamic_config import rewrite_formio_components
+from openforms.forms.tests.factories import (
+    FormDefinitionFactory,
+    FormFactory,
+    FormStepFactory,
+)
 from openforms.logging.models import TimelineLogProxy
 from openforms.submissions.tests.factories import SubmissionFactory
+from openforms.submissions.tests.mixins import SubmissionsMixin
 
 
 @override_settings(LANGUAGE_CODE="en")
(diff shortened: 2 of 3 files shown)


src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
```diff
@@ -743,3 +753,42 @@ def test_different_label_key_options(self):
...
+        )
+        user = SuperUserFactory.create()
+        self.client.force_authenticate(user=user)
+
+        response = self.client.get(endpoint, HTTP_X_CSP_NONCE="dGvsa==")
+        formio_component = response.json()["formStep"]["configuration"]["components"][0]
+        html = formio_component["html"]
+
+        self.assertIn("nonce", html)
+        self.assertIn("color:#e64c4c", html)
```
SilviaAmAm, line 794: I think it might be good to make the test and the fix a bit more specific.

For example, if you change the component in this test to be `<p><span style="color:#e64c4c;">Test nonce</span></p>` you see that at the end the inline style is not properly changed into a `<style>` tag and the test still succeeds.

So maybe in the fix check that the nonce is in the inline style and in the test check that the nonce is in the id and there is a `style` tag present?

SilviaAmAm (commented):
As we discussed in the office, this bandaid seems fine to fix this specific problem, but it doesn't address the bigger issue of code being ran multiple times when logged in as admin due to the `translations` field. 

@sergei-maertens I think this is fine for now?

sergei-maertens:
> As we discussed in the office, this bandaid seems fine to fix this specific problem, but it doesn't address the bigger issue of code being ran multiple times when logged in as admin due to the `translations` field.
> 
> @sergei-maertens I think this is fine for now?

bandaid fix for now is fine yeah

sergei-maertens:
I've backported the fix to stable/2.1.x in 9ed42ce935b6a89557ab3b92858e0495ad35c842, but @pi-sigma, please do clean up the tests a bit per Silvia's suggestions before we merge this into master.

src/openforms/formio/dynamic_config/__init__.py
```diff
@@ -44,7 +45,17 @@ def rewrite_formio_components_for_request(
...
+        if html_string := component.get("html", None):
+            soup = BeautifulSoup(html_string, "lxml")
+            tags = soup.find_all()
+            # find style attributes (not tags)
+            styles = [tag.get("style", None) for tag in tags]
+            nonce_missing = any(styles) and not all(
+                getattr(style, "nonce", None) for style in styles
+            )
+        if component["type"] != "content" or nonce_missing:
+            register.update_config_for_request(component, request=request)
```
sergei-maertens, line 58: a component definition like `{"type": "content", "key": "content", "html": ""}` will cause this to crash because `nonce_missing` is an undefined variable

src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
```diff
@@ -743,3 +754,82 @@ def test_different_label_key_options(self):
...
+        )
+        html1 = component1["html"]
+
+        doc = pq(html1)
+        style = doc.find("style")
+        id = doc.find("span").attr("id")
+
+        self.assertIn("color:#e64c4c", style.html())
+        self.assertIsNotNone(style.attr("nonce"))
+        self.assertIn("nonce", id)
```
sergei-maertens, line 820: can you please wrap each component in `with self.subTest():`

src/openforms/formio/dynamic_config/__init__.py
sergei-maertens, line 58: what happens now if I have a component with malicious HTML?

```html
<div>
  <style nonce="my-malicious-and-bad-nonce"></style>
  <script>alert('xss')</script>
</div>
```

If I read the code correctly, this means the style tag will be found, and all found style tags have a nonce (even though this is a spoofed nonce), so `nonce_missing` will be `False`, causing the component not to be post processed even though it should have been.

1. I'm not sure what `soup.find_all()` does without arguments - keep in mind that style tags can in theory be nested deeply
2. you need to compare against the server-side known value of the nonce, this is set by the CSP middleware on the request object. You can only leave the HTML untouched if all style tags have a nonce value that's equal to the one set on the request. This prevents malicious actors providing hardcoded nonces.

This is not the right place to do this sanitation - you should make this specific to the Content component: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/components/vanilla.py#L173

sergei-maertens (changes requested):
I'd like to see the malicious inputs as test cases too please

src/openforms/formio/dynamic_config/__init__.py
SilviaAmAm: Wouldn't it be enough though for the current issue to make sure that there are no longer tags with inline styles? 
Since this is a bandaid trying to figure out if the function has already run or not, i think that checking whether the nonce are consistent is not really that necessary?

CharString: > Wouldn't it be enough though for the current issue to make sure that there are no longer tags with inline styles?

No, while having run the code once does imply absence of inline styles, the absence of inline styles doesn't imply the code has already run before. I haven't checked what `register.update_config_for_request` does, but if I understand correctly it should be called at least (and preferably at most) once. Absence of inline styles does not guarantee that.

SilviaAmAm: but for the content component,  `register.update_config_for_request` calls https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/components/vanilla.py#L173 (which, if I understand correctly, checks if there are inline styles and if they are present they are transformed in style tags with a nonce). If there are no inline styles, the `rewrite_for_request` for the content shouldn't run again. So I think that the check should be moved to the `rewrite_for_request` but then checking there if there are inline styles should be enough?

sergei-maertens:
The problem is that you can't tell the difference between legitimate style
tags or style tags maliciously added by a form designer. While it's an edge
case as we consider those trusted, this should still work in the secure,
zero trust way purely because there's a risk the code is copied/reused in
untrusted contexts

The code may also be called without the CSP nonce request header, and in
that case we can't even properly extract let alone detect these situations.
Remember that the backend serves the CSP nonce and passes it to the SDK,
which in turn tells the API which nonce to use. This is more secure as the
potentially malicious inline style cannot predict (statically) what the
nonce will be each time the SDK fetches the configuration.

So to be correct, all style tags must have the specified nonce from the
request header, if present. If it's absent, the styles stay inline and
will/should be blocked by the CSP of the page serving the SDK/form.

src/openforms/formio/dynamic_config/__init__.py
```diff
@@ -44,7 +45,25 @@ def rewrite_formio_components_for_request(
...
+            tags_with_inline_style = [tag for tag in tags if tag.get("style")]
+            tags_with_nonce = [tag for tag in tags if tag.get("nonce")]
+            html_nonce = tags_with_nonce[0].get("nonce") if tags_with_nonce else None
+
+        if component["type"] != "content":
+            register.update_config_for_request(component, request=request)
+        elif tags_with_inline_style:
+            register.update_config_for_request(component, request=request)
+        elif request_nonce and html_nonce and html_nonce != request_nonce:
+            register.update_config_for_request(component, request=request)
```
sergei-maertens, line 66: Can you move this patch into the correct place - the Formio component plugin at `src/openforms/formio/components/vanilla.py`, in the `Content` class

sergei-maertens (changes requested):

src/openforms/formio/dynamic_config/__init__.py
```diff
@@ -44,7 +45,25 @@ def rewrite_formio_components_for_request(
...
     """
     for component in configuration_wrapper:
-        register.update_config_for_request(component, request=request)
+        # prevent multiple overwrites of content components (bandaid for #2895)
+        # postprocess (a) non-content components, (b) components with inline style
+        # (c) components with nonce that does not match the nonce on the request
+
+        # get HTML (if any) from component and parse it
+        if html_string := component.get("html", None):
+            request_nonce = request.headers.get("X-Csp-Nonce", None)
```
sergei-maertens, line 54: You can use `NONCE_HTTP_HEADER` from `src/csp_post_processor/constants.py`, and the `.get` returns `None` by default already so you can leave the second argument off

src/openforms/formio/dynamic_config/__init__.py
```diff
@@ -44,7 +45,25 @@ def rewrite_formio_components_for_request(
...
+            html = lxml.html.fromstring(html_string)
+            tags = list(html.iter())
+            tags_with_inline_style = [tag for tag in tags if tag.get("style")]
+            tags_with_nonce = [tag for tag in tags if tag.get("nonce")]
```
sergei-maertens, line 58: shouldn't this be limited to the `tags_with_inline_style` instead of all `tags`?

src/openforms/formio/dynamic_config/__init__.py
sergei-maertens, line 66: wouldn't this check be better formulated as:

```py
tags_with_correct_nonce = [tag for tag in tags_with_nonce if tag.get("nonce") == request_nonce]
is_suspicious = len(tags_with_correct_nonce) != (num_style_tags := len(tags_with_inline_style)) and num_style_tags > 0
```

sergei-maertens (changes requested):

sergei-maertens (changes requested):
A simpler approach would probably to NOT do this lxml parsing at all, and in the `post_process_html` keep track of the return value's status by setting a private attribute.

```py
def post_process_html(html: str | SafeString, request: Union[HttpRequest, Request]) -> str:
    if getattr(html, "_csp_post_processed", False):
        return html

    ...

    result = ... # do post processing here
    result._csp_post_processed = True  # type:ignore
    return result

```

sergei-maertens (approved):

PR #3006, opened 2023-04-18
[#3004] - Allow disabling "Save current information" for forms.

### LINKED ISSUE #3004: Allow disabling "Save current information" for forms.
### Thema / Theme

Other

"Save current information" is undesirable in some deployments of Open Forms. I propose adding a Boolean field `save_allowed` (default=`True`) on the `Form` model and exposing in the admin and `FormSerializer`.

- The SDK can then respond by not showing  the "Save current information" button.
- TBD: How to handle submission of data if save_allowed=`False`?

Some organizations don't want to use forms with features like this or may have other approaches to dealing with these flows. Making this configurable provides for this scenario.

_No response_

svenvandescheur:
Depends on: https://github.com/open-formulieren/open-forms-sdk/pull/399

Closes #3004


src/openforms/submissions/api/serializers.py
```diff
@@ -276,19 +276,27 @@ class SubmissionSuspensionSerializer(serializers.ModelSerializer):
             "The email address where the 'magic' resume link should be sent to"
         ),
     )
+    suspend_allowed = serializers.BooleanField()
```
sergei-maertens, line 279: this isn't right - this is not a field that the client may control. It needs to be looked up from the serializer instance for validation via `self.instance.form.suspend_allowed`

src/openforms/submissions/api/serializers.py
```diff
@@ -276,19 +276,27 @@ class SubmissionSuspensionSerializer(serializers.ModelSerializer):
...
             "suspended_on": {
                 "read_only": True,
             }
         }
 
+    # TODO: Move validators to view?
+
+    def validate_suspend_allowed(self, suspend_allowed):
+        if not suspend_allowed:
+            raise exceptions.PermissionDenied(_("Suspending this form is not allowed."))
```
sergei-maertens, line 298: as discussed - this should raise a `ValidationError`, not a permission denied, if we bring it in the serializer instead of the view.

This is also not a serializer marked as experimental, so now I'm even more convinced that this should happen in the view instead of the serializer (where it should indeed raise a permission denied).

src/openforms/submissions/api/validation.py
```diff
@@ -34,6 +34,8 @@ class CompletionValidationSerializer(serializers.Serializer):
     privacy_policy_accepted = serializers.BooleanField()
     contains_blocked_steps = serializers.BooleanField()
 
+    # TODO: Move validators to view?
```
sergei-maertens, line 37:
```suggestion
    # TODO: Move permission check-like validators to view?
```

src/openforms/submissions/api/viewsets.py
```diff
@@ -350,6 +350,7 @@ def retrieve(self, request, *args, **kwargs):
         description=_("Retrieve the data to display in the submission summary page."),
         responses={
             200: SubmissionStepSummarySerialzier(many=True),
+            403: ExceptionSerializer,
```
sergei-maertens, line 353: where is this change coming from?

src/openforms/submissions/api/viewsets.py
```diff
@@ -334,7 +334,7 @@ def _suspend(self, request, *args, **kwargs):
         submission = self.get_object()
         serializer = SubmissionSuspensionSerializer(
             instance=submission,
-            data=request.data,
+            data={**request.data, "suspend_allowed": submission.form.suspend_allowed},
```
sergei-maertens, line 337: this is the wrong approach, better would be to do:

```py
if not submission.form.submission_allowed:
    raise PermissionDenied(...)
serializer = SubmissionSuspensionSerializer(...)
```

src/openforms/submissions/api/serializers.py
```diff
@@ -276,19 +276,27 @@ class SubmissionSuspensionSerializer(serializers.ModelSerializer):
...
+
```
CharString, line 299: Yeah, I think this check belongs in the `_suspend` view, right after loading the submission and before de-serialising the user input. The fact that it doesn't raise a `ValidationError` is a tell tail sign.

src/openforms/submissions/tests/test_submission_suspension.py
```diff
@@ -69,6 +69,34 @@ def test_suspended_submission(self):
...
+    @freeze_time("2020-12-11T10:53:19+01:00")
+    @override_settings(LANGUAGE_CODE="en")
+    def test_suspended_submission_not_allowed(self):
+        form = FormFactory.create(suspend_allowed=False)
+        FormStepFactory.create(form=form)
+        step2 = FormStepFactory.create(form=form)
+        submission = SubmissionFactory.create(form=form)
+        SubmissionStepFactory.create(
+            submission=submission, form_step=step2, data={"foo": "bar"}
+        )
```
sergei-maertens, line 81:
```suggestion
        submission = SubmissionFactory.create(
            form__generate_minimal_setup=True,
            form__suspend_allowed=False,
        )
```

src/openforms/submissions/tests/test_submission_suspension.py
```diff
@@ -69,6 +69,34 @@ def test_suspended_submission(self):
...
+            response_data["type"], "http://testserver/fouten/PermissionDenied/"
+        ),
+        self.assertEqual(response_data["code"], "permission_denied"),
+        self.assertEqual(
+            response_data["title"], "You do not have permission to perform this action."
+        ),
+        self.assertEqual(response_data["status"], 403),
+        self.assertEqual(
+            response_data["detail"], "Suspending this form is not allowed."
+        ),
```
sergei-maertens, line 98:
```suggestion
        response = self.client.post(endpoint, {"email": "hello@open-forms.nl"})
        
        self.assertEqual(response.status_code, 403)
        error = response.json()["invalidParams"][0]
        self.assertEqual(error["code"], "permission_denied")
```

1. The error structure you're testing does not adhere to the documented error structure in the API spec
2. The status code and error code are most important, the error detail message is "fluff", no need to test for that and deal with translation machinery.

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -308,6 +309,25 @@ const FormConfigurationFields = ({
           />
         </Field>
       </FormRow>
+      <FormRow>
+        <Checkbox
+          name="form.suspendAllowed"
```
sergei-maertens, line 314: please add the default value of `True` to `form-creation-form.js`, in the `initialFormState` constant.

src/openforms/submissions/api/validation.py
```diff
@@ -34,6 +34,8 @@ class CompletionValidationSerializer(serializers.Serializer):
...
+
```
CharString, line 38: :thinking:  SubmissionAllowed is very much like the SuspensionAllowed. But accepting the privacy policy isn't. @sergei-maertens ?

sergei-maertens (changes requested):

CharString (dismissed):
To keep the naming scheme the same as `submissionAllowed`, I'd use "suspension allowed" instead of "suspend allowed"

src/openforms/submissions/api/validation.py
sergei-maertens: yeah we discussed this yesterday - cleaning this up right is out of scope (and the submit/complete serializer is still marked experimental so we can get away with it), but the privacy policy is something the client can fix/control, hence an HTTP 400 is appropriate, while the submission being blocked is more of a permission thing they have little control over, hence an HTTP 403 would be more appropriate.

HOWEVER - the steps being blocked CAN be derived from user input and logic, so a 400 may be appropriate if/when giving different answers unblocks submission. So it's a bit of a grey area :grimacing: Probably this should be broken up with the _complete endpoint checking if its outright blocked on the form -> HTTP 403 and otherwise if it's dynamically blocked -> HTTP 400 and leave it to the serializer.

sergei-maertens (changes requested):
Can you please:

* Get rid of the merge migration by rebasing on master and re-generating the migration, so we get a linear migration history from new features

* Clean up your commit history by rebasing on master - there are now a number of commits without ticket reference and you can incorporate all the feedback in their relevant "feature" commits.

Other than that this looks fine to merge.

re: the migrations - we already end up with plenty of merge migrations from forward-ported fixes to older stable branches, so for the master branch we really want to keep a linear migration history as much as possible. You can roll back the migration on your dev machine first

sergei-maertens:
I've cleaned up the commit history, should be good to merge now when CI completes.

sergei-maertens (approved):

Issue #2999, opened 2023-04-17
Ensure the test suite passes when ran in reverse order


sergei-maertens:
@bart-maykin

- [x] Fix broken test isolation
- [x] Set up CI to run in both normal and reverse order

PR #3019, opened 2023-04-21
Chore/2999 fixed broken isolated tests

bart-maykin:
#2999 

Fixing unit tests that are failing when running tests in reverse

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,26 @@ jobs:
           DB_USER: postgres
           DB_PASSWORD: ''
 
+      - name: Run tests reverse order
+        run: |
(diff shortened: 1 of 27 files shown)


sergei-maertens (changes requested):
Quick couple of remarks for Monday:

* Cleaning up the state from earlier tests at the start of some tests is not the right approach - we need to find where those TimelineLogProxy records are created and why, and ensure that they are cleaned up after the test that caused them to be created
* TransactionTestCase should really only be used if there is no other option, as they slow down the test suite considerably.
* We prefer *not* hitting the database or cache at all with django-solo models, and instead tend to `unittest.patch.mock` the `SoloModel.get_solo()` call by returning an in-memory instance.

PR #3033, opened 2023-04-26
[#1879] display only relevant validation plugins

### LINKED ISSUE #1879: Only show relevant validation plugins in each component
### Thema / Theme

Form designer

Currently, all components show the dropdown to chose a validation plugin.

For example, in the date component one can choose the KvK valiator and the phone number validator.
Screenshot from 2022-08-18 16-31-30

Components should show only the relevant validators.

_No response_

_No response_

pi-sigma:
Closes #1879 

* updated existing tests to reflect changes

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -4589,6 +4589,12 @@ paths:
       operationId: validation_plugins_list
       description: List all prefill plugins that have been registered.
       summary: List available validation plugins
+      parameters:
+      - in: query
+        name: component
+        schema:
+          type: string
+        description: The formio component for which the validation plugin is intended.
       tags:
       - validation
       security:
@@ -8254,7 +8260,13 @@ components:
         label:
           type: string
           description: The human-readable name for a plugin.
+        components:
+          type: array
+          items:
+            type: string
+          description: The components for which the plugin is relevant.
       required:
+      - components
       - id
       - label
     ValidationResult:
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -4591,10 +4591,11 @@ paths:
       summary: List available validation plugins
       parameters:
       - in: query
-        name: component
+        name: componentType
         schema:
           type: string
-        description: The formio component for which the validation plugin is intended.
+        description: Only return validators applicable for the specified component
+          type.
       tags:
       - validation
       security:
@@ -8260,13 +8261,14 @@ components:
         label:
           type: string
           description: The human-readable name for a plugin.
-        components:
+        forComponents:
           type: array
           items:
             type: string
+          title: Components
           description: The components for which the plugin is relevant.
       required:
-      - components
+      - forComponents
       - id
       - label
     ValidationResult:
--- a/src/openforms/contrib/kvk/validators.py
+++ b/src/openforms/contrib/kvk/validators.py
@@ -13,7 +13,7 @@ from openforms.validations.registry import register
 @deconstructible
 class NumericBaseValidator:
     value_size: int = NotImplemented
-    value_label: str = NotImplemented
+    value_label: int = NotImplemented
     error_messages = {
         "too_short": _("%(type)s should have %(size)i characters."),
     }
@@ -77,7 +77,7 @@ class KVKRemoteBaseValidator:
                 return True
 
 
-@register("kvk-kvkNumber", verbose_name=_("KvK number"), component="textfield")
+@register("kvk-kvkNumber", verbose_name=_("KvK number"), for_components=("textfield",))
 @deconstructible
 class KVKNumberRemoteValidator(KVKRemoteBaseValidator):
     query_param = "kvkNummer"
@@ -88,7 +88,7 @@ class KVKNumberRemoteValidator(KVKRemoteBaseValidator):
         super().__call__(value)
 
 
-@register("kvk-rsin", verbose_name=_("KvK RSIN"), component="textfield")
+@register("kvk-rsin", verbose_name=_("KvK RSIN"), for_components=("textfield",))
 @deconstructible
 class KVKRSINRemoteValidator(KVKRemoteBaseValidator):
     query_param = "rsin"
@@ -100,7 +100,9 @@ class KVKRSINRemoteValidator(KVKRemoteBaseValidator):
 
 
 @register(
-    "kvk-branchNumber", verbose_name=_("KvK branch number"), component="textfield"
+    "kvk-branchNumber",
+    verbose_name=_("KvK branch number"),
+    for_components=("textfield",),
 )
 @deconstructible
 class KVKBranchNumberRemoteValidator(KVKRemoteBaseValidator):
--- a/src/openforms/contrib/kvk/validators.py
+++ b/src/openforms/contrib/kvk/validators.py
@@ -77,7 +77,7 @@ class KVKRemoteBaseValidator:
                 return True
 
 
-@register("kvk-kvkNumber", verbose_name=_("KvK number"))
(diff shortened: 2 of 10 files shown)


src/openforms/validations/api/views.py
```diff
@@ -29,7 +29,11 @@ class ValidatorsListView(ListMixin, APIView):
     serializer_class = ValidationPluginSerializer
 
     def get_objects(self):
-        return list(register.iter_enabled_plugins())
+        param_component = self.request.query_params.get("component", None)
```
SilviaAmAm, line 32: It might be good to use a serializer here to validate the query params (similarly to, for example https://github.com/open-formulieren/open-forms/tree/master/src/openforms/appointments/api/views.py#L88)

src/openforms/validations/api/views.py
SilviaAmAm, line 1: I think you should extend the API schema to document that you can use a query param to filter. (see https://github.com/open-formulieren/open-forms/tree/master/src/openforms/appointments/api/views.py#L63 as an example)

src/openforms/validations/api/views.py
```diff
@@ -29,7 +29,11 @@ class ValidatorsListView(ListMixin, APIView):
...
+        filtered = filter(
+            lambda x: x.component == param_component, register.iter_enabled_plugins()
+        )
+        return [item for item in filtered]
```
SilviaAmAm, line 36: Maybe more readable? :smile: 

```suggestion
        return [plugin for plugin in register.iter_enabled_plugins() if plugin.component == param_component]
```
But also:
The query string should not be compulsory, if you don't specify it, I would expect that all the validators are returned.

src/openforms/validations/tests/test_api.py
```diff
@@ -40,22 +58,44 @@ def setUp(self):
     def test_auth_required(self):
         self.client.logout()
         url = reverse("api:validators-list")
-        response = self.client.get(url)
+        component = "textfield"
+        response = self.client.get(url, args=component)
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
     def test_validations_list(self):
```
SilviaAmAm, line 65: I think it might be better to have this test check that if no querystring is used, then all validators are returned, then in a separate test check the filters for a specific component.

SilviaAmAm (changes requested):
One more thought:
I can imagine in the future that a validator might be relevant for more than one component. Maybe each validator should have a tuple of components, instead of just one. Then the filter should check "query string in components" instead of "query string == component" :thinking:

src/openapi.yaml
```diff
@@ -8233,7 +8239,12 @@ components:
         label:
           type: string
           description: The human-readable name for a plugin.
+        components:
+          type: array
+          items: {}
+          description: The components for which the plugin is relevant.
```
sergei-maertens, line 8245: this doesn't seem right - shouldn't this be a proper schema, my first impression is that it each item should be a string?

src/openforms/contrib/kvk/validators.py
```diff
@@ -12,7 +12,7 @@
 
 @deconstructible
 class NumericBaseValidator:
-    value_size: int = NotImplemented
+    value_size: str = NotImplemented
```
sergei-maertens, line 15: uh, this should be `int`, no?

src/openforms/contrib/kvk/validators.py
```diff
@@ -48,6 +48,7 @@ class KVKBranchNumberValidator(NumericBaseValidator):
 class KVKRemoteBaseValidator:
     query_param = NotImplemented
     value_label = NotImplemented
+    components: tuple = ("textfield",)
```
sergei-maertens, line 51: maybe `relevant_for_components` to better cover the meaning? The type hint should also be fully specified.

```py
relevant_for_components: tuple[str] = ("textfield",)
```

Looking at the remaining code, I think we should not specify this as a class attribute, but instead make it a required argument to the `register` decorator so you'd register a validator like this:

```py
@register("kvk-kvkNumber", verbose_name=_("KvK number"), for_components=("textfield",))
class KVKNumberRemoteValidator(...):
     ...
```

This information is not relevant for the validator implementation itself but only for the plugin/registry metadata.

src/openforms/validations/api/serializers.py
```diff
@@ -9,6 +9,10 @@ class ValidationInputSerializer(serializers.Serializer):
     )
 
 
+class ValidationRequestParamSerializer(serializers.Serializer):
+    component = serializers.CharField()
```
sergei-maertens, line 13: typically you'd use a filter set/backend for this instead of a serializer, see https://django-filter.readthedocs.io/en/stable/guide/rest_framework.html

This is then also properly picked up by the API schema generation.

src/openforms/validations/api/serializers.py
```diff
@@ -33,3 +37,7 @@ class ValidationPluginSerializer(serializers.Serializer):
         label=_("Label"),
         help_text=_("The human-readable name for a plugin."),
     )
+    components = serializers.ListField(
+        label=_("Components"),
+        help_text=_("The components for which the plugin is relevant."),
+    )
```
sergei-maertens, line 43: I think that `serializers.CharField(many=True)` is more appropriate if that's possible, otherwise you want to specify the `child` argument of the `ListField` to properly type the items. That should fix the OpenAPI schema issues.

src/openforms/validations/api/views.py
```diff
@@ -11,13 +11,25 @@
...
+        summary=_("List available validation plugins"),
+        parameters=[
+            OpenApiParameter(
+                "component",
+                OpenApiTypes.STR,
+                description=_(
+                    "The formio component for which the validation plugin is intended."
+                ),
+            )
+        ],
```
sergei-maertens, line 31: should not be needed if you specify the filter set for the view.

src/openforms/validations/api/views.py
```diff
@@ -29,7 +41,19 @@ class ValidatorsListView(ListMixin, APIView):
...
+        param_component = self.request.query_params.get("component", None)
+
+        if param_component is None:
+            return plugins
+
+        serializer = ValidationRequestParamSerializer(data=self.request.query_params)
+        if not serializer.is_valid():
+            return []
+
+        # import pdb;pdb.set_trace()
```
sergei-maertens, line 54: This should use the filter mechanism, for a reference check DRFs base views: https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L144

This would be useful to add to the `ListMixin` so that it acts more similar to the DRF `ListView`.

src/openforms/validations/registry.py
```diff
@@ -27,6 +27,7 @@ class RegisteredValidator:
     identifier: str
     verbose_name: str
     callable: ValidatorType
+    components: tuple
```
sergei-maertens, line 30:
```suggestion
    components: tuple[str]
```

src/openforms/validations/registry.py
```diff
@@ -84,6 +87,7 @@ def decorator(validator: Union[Type, ValidatorType]):
                 identifier=identifier,
                 verbose_name=verbose_name,
                 callable=call,
+                components=getattr(validator, "components", ()),
```
sergei-maertens, line 90: I would enforce that validators must specify this at all times and get rid of the `getattr` stuff.

sergei-maertens (changes requested):

src/openforms/validations/api/serializers.py
sergei-maertens: The library requires models, so this suggestion is plain wrong

PR #3042, opened 2023-05-01
clean up type annotations

pi-sigma:
Cleanup type annotations


joeribekker:
Refinement: @pi-sigma rebases this and marks as ready for review.

PR #3047, opened 2023-05-02
:art: [#3031] made object api registry plugin fields wider

### LINKED ISSUE #3031: Increase size of Object API fields
### Thema / Theme

Frontend

DH Taiga issue 487

The fields to configure the URLs of the various fields of the Object API registration backends are a bit small, so you can't see the full URL when looking at the configuration.
objectapi-fields

This makes debugging easier since you can see the UUIDs of the various resources

_No response_

bart-maykin:
Fixes #3031 

Added css to increase specific input fields based on class and id

### DIFF
--- a/src/openforms/scss/admin/_app_overrides.scss
+++ b/src/openforms/scss/admin/_app_overrides.scss
@@ -202,7 +202,6 @@ $djai-border-width: 8px;
     color: #999; // same color as django forms.css
   }
 
-
   select.form-control {
     $properties: display, width, font-size, font-weight, line-height, color, background-clip,
       transition;
--- a/src/openforms/scss/admin/_app_overrides.scss
+++ b/src/openforms/scss/admin/_app_overrides.scss
@@ -202,14 +202,6 @@ $djai-border-width: 8px;
     color: #999; // same color as django forms.css
   }
 
-  input#root_objecttype.form-control,
-  input#root_productaanvraagType.form-control,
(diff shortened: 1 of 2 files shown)


src/openforms/scss/admin/_app_overrides.scss
```diff
@@ -202,6 +202,15 @@ $djai-border-width: 8px;
...
   }
 
+  input#root_objecttype.form-control,
+  input#root_productaanvraagType.form-control,
+  input#root_informatieobjecttypeSubmissionReport.form-control,
+  input#root_informatieobjecttypeSubmissionCsv.form-control,
+  input#root_informatieobjecttypeAttachment.form-control,
+  input#root_organisatieRsin.form-control {
+    width: 25em;
+  }
```
sergei-maertens, line 212: generally we avoid ID-based and HTML element-based selectors, but instead we favour classnames only

Can you check if we can leverage the `uiSchema` feature of [react-json-schema-form](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/uiSchema/)? This is hooked up from serializer data here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/registrations/index.js#L10

sergei-maertens (changes requested):

sergei-maertens (changes requested):
Fix is alright, just make sure to format the code with prettier & get CI to pass

sergei-maertens (approved):

PR #3048, opened 2023-05-02
:bug: [#2970] fixed bug that would ignore design token for confirmation and suspension mail

### LINKED ISSUE #2970: Footer in website and confirmation/pausing email use different design tokens

2.1.x

Reported by DH

Currently, to set the footer of OF to a different bg/fg colour, we can configure it in the admin with these tokens:
```
{
  "of": {
    "page-footer": {
      "bg": {
        "value": "#EDBF07"
      },
      "fg": {
        "value": "#000000"
      }
    }
  }
}
```
However, this results in the default blue for the footer of the confirmation and the 'form pausing' email. To have the right colours, these design tokens have to be added:
```
{
  "of": {
    "page-footer": {
      "bg": {
        "value": "#EDBF07"
      },
      "fg": {
        "value": "#000000"
      }
    }
  },
  "page-footer": {
    "bg": {
      "value": "#EDBF07"
    },
    "fg": {
      "value": "#000000"
    }
  }
}
```

This is because [here](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/emails/context.py#L66) we retrieve the tokens with path `page-footer.bg.value` instead of `of.page-footer.bg.value`.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

bart-maykin:
Fixes #2970

- Changed glom spec to match of token spec.
- Added unit tests to test confirmation and suspension mail css with and without design tokens


src/openforms/emails/tests/test_confirmation_email.py
```diff
@@ -797,3 +797,128 @@ def test_templatetag_alias(self):
         ).lstrip()
 
         self.assertEquals(expected_text, text)
+
+    def test_email_with_default_design_token(self):
+        with patch(
+            "openforms.config.models.GlobalConfiguration.get_solo",
+            return_value=GlobalConfiguration(logo="https://logo.png"),
```
SilviaAmAm, line 804: Maybe good to make it explicit that here no design tokens are specified:
```suggestion
            return_value=GlobalConfiguration(logo="https://logo.png", design_token_values={}),
```

src/openforms/submissions/tests/test_submission_suspension.py
```diff
@@ -262,3 +262,99 @@ def test_identifying_attributes_hashed_on_suspend(self):
         # Assert that identifying attribute is hashed
         self.assertNotEqual(submission.auth_info.value, "123456789")
         self.assertTrue(submission.auth_info.attribute_hashed)
+
```
SilviaAmAm, line 265: I wonder if these new tests are really needed, since both confirmation/suspension email call the same function `get_wrapper_context` to take care of putting the design tokens in the context? Maybe just a test of `_get_design_token_values` could have been enough?

SilviaAmAm (commented):
Fix looks good! Just some comments on the tests :slightly_smiling_face:

SilviaAmAm (approved):

sergei-maertens (approved):

PR #3053, opened 2023-05-03
[#1424] Add reference to PDF report

### LINKED ISSUE #1424: Include registration reference in PDF
Follow up from #1423

We cannot simply include the reference since the PDF is attached while registering the submission and obtaining the reference. We need to figure out how to make this possible.

SilviaAmAm:
Fixes #1424


src/openforms/submissions/tests/test_tasks_registration.py
```diff
@@ -143,200 +139,3 @@ def obtain_reference():
         submission2.refresh_from_db()
         self.assertEqual(submission2.public_registration_reference, "OF-UNIQUE")
         self.assertEqual(submission1.public_registration_reference, "OF-OTHER")
-
-
-class GenerateSubmissionReferenceTests(TestCase):
```
SilviaAmAm, line 148: Moved to the `test_pre_registration.py` file

src/openforms/submissions/tasks/__init__.py
```diff
@@ -34,7 +34,6 @@ def on_completion(submission_id: int) -> None:
     pre_registration_task = pre_registration.si(submission_id)
     generate_report_task = generate_submission_report.si(submission_id)
     register_submission_task = register_submission.si(submission_id)
-    obtain_submission_reference_task = obtain_submission_reference.si(submission_id)
```
SilviaAmAm, line 37: I don't think having this task makes much sense now?

CharString (changes requested):
Just some formatting. When viewing at the `test_backend` changes I thought it was missing some failure modes. But those are covered in the narrower tests.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -327,3 +316,38 @@ def get_config_actions(self) -> List[Tuple[str, str]]:
...
+            submission, self.zaak_mapping, REGISTRATION_ATTRIBUTE
+        )
+
+        _create_zaak = partial(
+            create_zaak,
+            options,
+            payment_required=submission.payment_required,
+            existing_reference=submission.public_registration_reference,
+            **zaak_data,
+        )
```
CharString, line 341: Hmm. So `execute_unless_result_exists` never calls this partial with any other arguments... Technically still partial application of the function, but do you think this is clearer?

```suggestion
        _create_zaak = lambda: create_zaak(
            options,
            payment_required=submission.payment_required,
            existing_reference=submission.public_registration_reference,
            **zaak_data,
        )
```

CharString (approved):
:ok_hand: No need to change the partial to a lambda, I'm just interested in your opinions. I saw the `execute_unless_results_exists` also returns an `Any` type. But it should return some type T which is the return type of the `callback` and `default` should be of that same type.

But that's beyond the scope of this PR. glom isn't type hinted anyway.

src/openforms/registrations/contrib/zgw_apis/plugin.py
SilviaAmAm: Yes, maybe it is a bit clearer! Although I am not sure if there was any rationale for using the partial vs the lambda when this code was written :thinking:

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -365,3 +361,20 @@ def get_config_actions(self):
                 ),
             ),
         ]
+
+    def pre_register_submission(self, submission: "Submission", options: dict) -> None:
```
sergei-maertens, line 365: I'd move this method to be above `register_submission` so that it reads a bit more naturally.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -188,7 +173,7 @@ def register_submission(
             "intermediate.documents.report.document",
         )
         execute_unless_result_exists(
-            partial(relate_document, zaak["url"], document["url"]),
+            partial(relate_document, result["zaak"]["url"], document["url"]),
```
sergei-maertens, line 176: `zaak` is defined on line 162 already as `zaak = result["zaak"]`

```suggestion
            partial(relate_document, zaak["url"], document["url"]),
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -327,3 +316,38 @@ def get_config_actions(self) -> List[Tuple[str, str]]:
                 ),
             ),
         ]
+
+    @wrap_api_errors
+    def pre_register_submission(self, submission: "Submission", options: dict) -> None:
```
sergei-maertens, line 321: Same as StUF-ZDS - let's define this method above the main registration method for more natural reading.

src/openforms/registrations/contrib/zgw_apis/plugin.py
CharString: Oh yes. I hate that rule. It hurts readability.

src/openforms/registrations/tasks.py
```diff
@@ -16,6 +17,56 @@
...
+    submission = Submission.objects.get(id=submission_id)
+    if submission.pre_registration_completed:
+        return
+
+    registration_plugin = submission.get_registration_plugin()
+    if not registration_plugin:
+        submission.set_submission_reference()
+        submission.pre_registration_completed = True
+        submission.save()
+        return
```
sergei-maertens, line 31: I think this should be wrapped in a `with transaction.atomic()` block? There is a non-zero chance that generated submission references are not unique

src/openforms/registrations/tasks.py
```diff
@@ -16,6 +17,56 @@
...
+        if not submission.registration_result:
+            submission.registration_result = {}
+
+        assign(
+            submission.registration_result,
+            "temporary_internal_reference",
+            submission.public_registration_reference,
+            missing=dict,
+        )
+        submission.save()
```
sergei-maertens, line 52: I'm also not sure if this should or should not be in an atomic transaction block :thinking:

src/openforms/registrations/tests/test_registration_hook.py
```diff
@@ -47,6 +47,7 @@ def setUpTestData(cls):
 
         cls.submission = SubmissionFactory.create(
             completed=True,
+            pre_registration_completed=True,
```
sergei-maertens, line 50: imo this `pre_registration_completed` should be made part of the `completed` trait.

src/openforms/submissions/models/submission.py
```diff
@@ -182,6 +182,13 @@ class Submission(models.Model):
             "Indication whether the registration in the configured backend was successful."
         ),
     )
+    pre_registration_completed = models.BooleanField(
+        _("pre-registration completed"),
+        default=False,
+        help_text=_(
+            "Indicates whether the pre-registration task completed successfully."
+        ),
+    )
```
sergei-maertens, line 191: we'll need a data migration to update existing records that have a succesful registration I think? for data consistency sake.

I think we can also lock this in at the DB level with a constraint that prevents the `registration_status=completed, pre_registration_completed=False` combination as it functionally does not make any sense, right?

src/openforms/submissions/models/submission.py
```diff
@@ -642,3 +649,13 @@ def get_prefilled_data(self):
                 variable.key: variable.value for variable in prefill_vars
             }
         return self._prefilled_data
+
+    def get_registration_plugin(self):
+        from ..tasks.registration import get_registration_plugin
+
+        return get_registration_plugin(self)
```
sergei-maertens, line 656: I don't like this - celery tasks should really be leaf nodes that use other utilities and not the other way. additionally, you're now leaking registration app/module details in the core submissions app: see https://open-forms.readthedocs.io/en/latest/developers/architecture.html#core

I'd move this `get_registration_plugin` utility into `openforms.registrations.service` instead

src/openforms/submissions/models/submission.py
```diff
@@ -642,3 +649,13 @@ def get_prefilled_data(self):
...
+
+    def set_submission_reference(self):
+        from ..tasks.registration import set_submission_reference
```
sergei-maertens, line 659: I think you could move this to `openforms.submissions.public_references` instead and contain the implementation details in there. Fine to also expose this as a public method then, but importing utilities from celery tasks is a bit of a no-go.

src/openforms/submissions/tests/test_on_completion_retry_chain.py
```diff
@@ -37,6 +37,7 @@ def test_payment_status_update_now_succeeds(self):
             # set by :func:`openforms.payments.tasks.update_submission_payment_status`
             needs_on_completion_retry=True,
             registration_success=True,
+            pre_registration_completed=True,
```
sergei-maertens, line 40: could use factory traits?

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -212,6 +212,26 @@ def test_confirmation_page_content_not_included_in_pdf(self):
...
+            with_report=True,
+            public_registration_reference="OF-12345",
+        )
+
+        html = submission.report.generate_submission_report_pdf()
+
+        doc = pq(html)
+        reference_node = doc(".metadata").children()[1]
+
+        self.assertEqual(reference_node.text, "Your reference is: OF-12345")
```
sergei-maertens, line 233: this will break when we add the NL translations - either mark the test with `@override_settings` or use the translatable string in the test.

src/openforms/submissions/tests/test_tasks_registration.py
```diff
@@ -31,39 +24,12 @@ def test_source_reference_from_registration_result(self, mock_generate):
             registration_result={"zaak": {"identificatie": "AEY64"}},
         )
 
-        obtain_submission_reference(submission.id)
+        set_submission_reference(submission)
```
sergei-maertens, line 27: can't you use the method here instead?

src/openforms/submissions/tests/test_tasks_registration.py
```diff
@@ -85,7 +51,7 @@ def get_random_string(*args, **kwargs):
             "openforms.submissions.tasks.registration.get_random_string",
             new=get_random_string,
         ):
-            obtain_submission_reference(submission.id)
+            set_submission_reference(submission)
```
sergei-maertens, line 54: method?

src/openforms/submissions/tests/test_tasks_registration.py
```diff
@@ -127,7 +93,7 @@ def obtain_reference():
                 "openforms.submissions.tasks.registration.generate_unique_submission_reference",
                 new=generate_unique_submission_reference,
             ):
-                obtain_submission_reference(submission1.id)
+                set_submission_reference(submission1)
```
sergei-maertens, line 96: method?

sergei-maertens (changes requested):
Thanks for fixing all my spelling mistakes :P

src/openforms/registrations/tasks.py
SilviaAmAm: The `set_submission_reference` function that gets called by `submission.set_submission_reference()` already wraps the saving of the submission reference in a transaction, and it already tries to save submission (so if there is a duplicate reference, it is raised here). If that fails, no other changes have been performed yet. 
So I wonder if wrapping it again is needed?

sergei-maertens: yes, but the subsequent `submission.save` call is not covered and if that would be (implicitly) doing other write queries there's a risk you get partial saves committed to the DB, I think? or if additional steps were added here? IMO it can't hurt to wrap it in an explicit transaction block.

src/openforms/submissions/tests/test_tasks_registration.py
SilviaAmAm: I removed the method (because of the other comment), so I'm leaving these!

sergei-maertens (approved):

PR #3069, opened 2023-05-05
Optimize JS bundle sizes

sergei-maertens:
By relying on browserslist information to figure out which JS features are natively supported in browsers, we can remove some polyfills and cut down the bundle size about 11% (from 3.46 MiB to 3.07 MiB).

---

Command used to emit production bundle.

```bash
npm run build
```

## Results

| Bundle       | baseline    | optimized (0.2% share)   | optimized (0.5% share) |
|--------------|-------------|--------------------------|------------------------|
| core-js.js   | 3.46 MiB    | 3.47 MiB                 | 3.07 MiB               |
| core-css.css | 224 KiB     | 231 KiB                  | 223 KiB                |
| public.js    | 4 KiB       | 4 KiB                    | 4 KiB                  |

## Conclusions:

* targetting 0.2% browser share as minimum makes bundle sizes slightly worse

  -> amongst others, this includes some browser without async/await syntax, requiring
  the regenerator runtime being included

* targetting 0.5%+ browser share targets all browsers supporting async/await

* For the admin environment, we can be quite aggressive in terms of supported browsers,
  it's a controlled environment. Using supported, modern browsers should be the default
  in the kind of environments where OF is used.

## Other thoughts

Splitting the JS bundles might yield benefits, but organizing the entrypoints for
different domains would *definitely* help (e.g. everything in the admin except form
designer would likely be a smaller bundle).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,6 +60,7 @@
         "autoprefixer": "^10.2.4",
         "babel-loader": "^8.2.2",
         "babel-plugin-formatjs": "^10.3.4",
+        "browserslist": "^4.21.5",
         "copy-webpack-plugin": "^6.4.1",
         "css-loader": "^5.0.2",
         "cssnano": "^5.0.17",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,6 @@
         "bem.js": "^1.0.10",
         "classnames": "^2.3.1",
         "copy-to-clipboard": "^3.3.1",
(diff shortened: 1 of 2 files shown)


sergei-maertens:
> Seems fine. Given our user base and this is backend code, I expect the browsers they use to by pretty uniform. Do we have User-Agent stats of deployments?

We'd need either analytics insights for that, or maybe extract it from APM metadata, but it's not readily available.

PR #3073, opened 2023-05-10
[#2845] add option to include content component in summary

### LINKED ISSUE #2845: Content component missing option to include it in summary page

2.0.x, 2.1.x

DH Taiga issue 83 / user story 88

All components have a check box in the edit form to specify whether it should be shown in the summary. However, the content component is missing this checkbox.
afbeelding
afbeelding

pi-sigma:
Closes #2845


src/openforms/js/components/form/content.js
```diff
@@ -25,7 +33,7 @@ const CUSTOM_CSS_CLASS = {
 
 const CONTENT_PRESENTATION = {
   ...PRESENTATION,
-  components: [SHOW_IN_EMAIL, SHOW_IN_PDF],
+  components: [SHOW_IN_SUMMARY(false), SHOW_IN_EMAIL, SHOW_IN_PDF],
```
SilviaAmAm, line 36: In other places we did `{...SHOW_IN_SUMMARY, defaultValue: false}`. For example here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/form/edit/tabs.js#L50

Maybe the above is a less invasive solution? Also in general we expect these components to be 'dicts' rather than functions, so maybe it's good to remain consistent?

SilviaAmAm (changes requested):
In order to get the content component to appear in the summary confirmation page, this logic in the renderer also needs to be changed (otherwise, even if you select in the form editor that the component should be shown, it will never be rendered):
- https://github.com/open-formulieren/open-forms/tree/master/src/openforms/formio/rendering/default.py#L145
- https://github.com/open-formulieren/open-forms/tree/master/src/openforms/formio/rendering/nodes.py#L88

SilviaAmAm (changes requested):

src/openforms/formio/rendering/tests/test_vanilla_formio_components.py
```diff
@@ -340,7 +340,7 @@ def test_wysiwyg_component(self):
             RenderModes.confirmation_email: False,
             RenderModes.pdf: True,
             RenderModes.cli: True,
-            RenderModes.summary: False,
+            RenderModes.summary: True,
```
sergei-maertens, line 343: Is this right? I don't see any hint that `showInSummary` is enabled for this component, so then I'd expect it to *not* be displayed in the summary unless it's explicitly opted-in to?

I am looking at `RENDER_CONFIGURATION` in `src/openforms/formio/rendering/conf.py`, which basically states that if the option is not explicitly provided, then the default is to display it:

```py
    RenderModes.summary: RenderConfiguration(
        RenderConfigurationOptions.show_in_summary, default=True
    ),
```

However, our desired default behaviour for content component is that it is not displayed unless explicitly enabled (as per the checkbox), which is the inverse.

So I'd expect these tests:

* content component with explicit opt-out (`"showInSummary": False`) -> not visible
* content component with explicit opt-in (`"showInSummary": True`) -> visible
* content component with implicit opt-out (`showInSummary` absent) -> not visible

I'm not sure how realistic the missing key variant is - when using the form.io builder it does seem to always include the key with its default value. So maybe the tests just need to be updated to not have implicit/missing keys, as that doesn't reflect real life?

@SilviaAmAm please double check if my interpretation of the ticket/beviour is right

SilviaAmAm: Yes, I also thought this. I had suggested above that to make sure the default is included a migration should be added (then there shouldn't be content components without the `showInSummary` key). And then for the case where there is no key (although it shouldn't happen) use the same hardcoded default as other components specified in `RENDER_CONFIGURATION`

sergei-maertens: okay, so then all that is missing is basically just the on/off cases in tests, as we now only have the edge case that shouldn't happen anyway because of the data migration, right?

SilviaAmAm: Ahh yes, indeed

sergei-maertens: @pi-sigma can you please update these tests so we can merge this?

pi-sigma: Changes had been made already, but I forgot to push the latest code. Sorry about that!

PR #3074, opened 2023-05-10
[#2762] update settings: log outgoing requests

### LINKED ISSUE #2762: Add logging in the admin to see outgoing requests/responses
### Thema / Theme

Admin

By far, most discussions arises who sends what to where. Therefore, we should log ALL outgoing requests and make them viewable in the admin. This includes the full URL, headers and body. For now, it's also not needed to link these logs to specific actions in Open Forms. They are mostly used for testing purposes and can be related based on date and time!

To tackle the basic privacy and security concernns: There should be a (Django) permission for viewing logs AND logs should be cleaned up with a regular interval. Other concerns can be tackled in other issues and are not part of this ticket.

I prefer to have a library added for this if it exists, otherwise a library can created for this.
See: https://github.com/maykinmedia/django-log-outgoing-requests

Shorten the time to analyse and fix issues with external systems.

Additional requests I made on the library that didn't make it into the initial version, and thus are tasks for us:

* [x] Can we make it so this (meaning: save or not save to the database) can be enabled / disabled from the admin (singleton class)?
* [x] Can we add an autoclean command/task so that entries are deleted after X days (default=7, so I can call clean_request_logs --keep=3)
* [x] I really prefer a view of the raw request and response (so simply what goes over the line), in addition to the formatted/table view
* [x] In the list view, I basically want to see 3 things: the request URL, the http response code and the timestamp. The URL can be limited to 100 chars maybe with elipises right after the domain name.
* [x] List view: Filter on http status code, http method, time hierarchy
* [x] Logging post body opt in? Just do it

You currently have:

``LOG_OUTGOING_REQUESTS_DB_SAVE``

Can we change this to have:

```
LOG_OUTGOING_REQUESTS_ALWAYS_LOG_TO_DB = None  # True, False -both mean ignore admin-, None -use admin setting-

LOG_OUTGOING_REQUESTS_LOG_BODY_TO_DB = True  # only has effect when saving to db

LOG_OUTGOING_REQUESTS_LOG_BODY_TO_STDOUT = False
```

pi-sigma:
Closes #2762 (partly)
~~Depends on [maykinmedia/django-log-outgoing-requests](https://github.com/maykinmedia/django-log-outgoing-requests/pull/4)~~

- add settings for saving request & response body
- add celery task + schedule for cleaning up logs

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,6 +11,7 @@ import sys
 import django
 
 sys.path.insert(0, os.path.abspath("../src"))
+os.environ["LOG_REQUESTS"] = "false"
 
 import openforms  # noqa isort:skip
 
--- a/docs/developers/backend/tests.rst
+++ b/docs/developers/backend/tests.rst
@@ -107,3 +107,29 @@ Example custom command:
     NO_E2E_HEADLESS=1 E2E_DRIVER=firefox python src/manage.py test src --tag=e2e
 
 .. note:: Only the presence of the ``NO_E2E_HEADLESS`` is checked, not the value
+
+Known issues
+============
+
+**AssertionError: Database queries to 'default' are not allowed in SimpleTestCase subclasses.**
+
+These are often caused by django-solo ``SingletonModel`` sucblasses that are being
+called somewhere, e.g. ``GlobalConfiguration.get_solo``. Sometimes they fetch from
+cache, sometimes there is a cache miss and a database query is needed (e.g. when running
+tests in reverse).
+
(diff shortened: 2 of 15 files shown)


pi-sigma:
The error in `log_outgoing_requests/log_requests.py`, line 24 is caused by a name conflict that's fixed by the complement PR in [maykinmedia/django-log-outgoing-requests](https://github.com/maykinmedia/django-log-outgoing-requests/pull/4)

sergei-maertens:
Styling for JSON without newlines is not great yet:

image

sergei-maertens (approved):

PR #3075, opened 2023-05-10
[#2808] fix overflowing filenames

### LINKED ISSUE #2808: long filename in fileupload that causes error -> overflows container

2.1.0.rc0

image

pi-sigma:
Fixes #2808

- overflow is prevented by removing the second occurrence of the filename from the error message (the filename is already displayed, hence the second occurrence is redundant)
- existing tests modified to reflect changes

### DIFF
--- a/src/openforms/formio/api/validators.py
+++ b/src/openforms/formio/api/validators.py
@@ -73,9 +73,7 @@ class MimeTypeValidator:
             or mimetype_allowed(mime_type, self._regular_mimes, self._wildcard_mimes)
         ):
             raise serializers.ValidationError(
-                _("The file '{filename}' is not a valid file type.").format(
-                    filename=value.name
-                ),
+                _("The provided file is not a valid file type.")
             )
 
         # Contents is allowed. Do extension or submitted content_type agree?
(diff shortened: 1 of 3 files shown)


src/openforms/formio/api/validators.py
```diff
@@ -96,9 +96,7 @@ def __call__(self, value: UploadedFile) -> None:
                 return
 
             raise serializers.ValidationError(
-                _("The file '{filename}' is not a {file_type}.").format(
-                    filename=value.name, file_type=f".{ext}"
-                )
+                _("The provided file is not a {file_type}.").format(file_type=f".{ext}")
```
SilviaAmAm, line 99: I think there is also an occurrence on line 76 that uses the filename in the error?

SilviaAmAm (commented):
I am wondering if this is the right approach :thinking: 

I agree that the error message with the filename is a bit redundant (since the file name is just above it):
Screenshot from 2023-05-11 12-05-18
But for example for mobile:
Screenshot from 2023-05-11 12-10-06
The name is not really visible :thinking: I don't know if for accessibility this is a problem.

Also: with the Formio errors, the form designers can specify custom errors for the various components. With these backend errors we haven't added this possibility yet, but I can imagine it will be requested in the future. 

So maybe this could be solved with a CSS fix to break the word and wrap?

joeribekker:
Refinement: On mobile: Move the bin-icon and file size to the next line AND show the whole filename using css break-word (this should wrap the filename).

public/mockServiceWorker.js
SilviaAmAm, line 1: I don't think these changes are related to the issue? Maybe good to do it in a separate PR?

SilviaAmAm (commented):
Maybe remove the changes from the mockServiceWorker file? But otherwise this part of the fix seems good to me!

PR #3080, opened 2023-05-10
:green_heart: [#2999] mock GlobalConfiguration for test_disabled_plugins_are_skipped

bart-maykin:
fixes test_disabled_plugins_are_skipped issolation #2999

### DIFF
--- a/src/openforms/config/tests/test_config_check.py
+++ b/src/openforms/config/tests/test_config_check.py
@@ -41,35 +41,37 @@ class ConfigCheckTests(TestCase):
             self.assertEqual(response.status_code, 200)
 
     def test_disabled_plugins_are_skipped(self):
-        config = GlobalConfiguration.get_solo()
-        config.plugin_configuration = {
-            "registrations": {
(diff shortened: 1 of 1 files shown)


src/openforms/config/tests/test_config_check.py
```diff
@@ -41,35 +41,37 @@ def test_access_permission(self):
...
-            },
-            "prefill": {
-                "kvk-kvknumber": {"enabled": False},
-            },
-        }
-        config.save()
-
-        user = StaffUserFactory(user_permissions=["configuration_overview"])
-        self.client.force_login(user)
+        with patch(
```
sergei-maertens, line 44: nice - however we can avoid the indentation level by moving the patch call to a test method decorator since the configuration values are fixed anyway:

```py
@patch("openforms.config.models.GlobalConfiguration.get_solo", return_value=GlobalConfiguration(...))
def test_disabled_plugins_are_skipped(self, mock_get_solo):
    ...
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3085, opened 2023-05-11
[#1530] Part 3 - Change on_completion chain

### LINKED ISSUE #1530: Co-signing does not work with a DigiD Gateway (e.g. OIDC) or DigiD SSO

1.1

When using OpenID Connect, clicking on the co-sign option redirects directly to the form and doesnt give the chance to sign in as someone else with DigiD.

### Proposed solution

**Update: Den Haag agrees with the Open Forms flow and abandons their own mixed flow. With this, the BSN check is also removed and can later be added.**

Use out-of-band email, form code. This solves co-signing when DigiD SSO or DigiD via a gateway is used. Note that the email that is sent never contains a link, but contains instructions to find a fixed "co-sign form" where the co-sign user can fill in their code.

Untitled (3)

[Latest sequence diagram](https://sequencediagram.org/index.html#initialData=C4S2BsFMAIGEHsC0BnEBzAdiDbregPIAOkG0AYvAE4C2yAUPUQIZWgDGILGwcSqmaAFVkkKoxFjEAPmKkK1OgC5oAEXQhV0ADLw02enLKVayGZKoqTNaADMQkcABNk0YPDshw4PBnoWZIwVTKy8ffCdmYGZDEmNFM2kLK0VoDEhIF2h2fnQsHH9RKkC44OVoAFEaZi9oZicnKkhkZAAdDHhbbNzMMUKpWVLrZBUAZQBXACMaMFj5YZL5hJUAcVIxKJhbVJynGHdPDCdoduAACxBXbdoAemQpmZaQeD8ghcGlkOhR9yboe+ml1QLzs1G6NCIUFAIPwAA1oJEAJ4MN4JcxFMYPIHPMiXaBxJzYXA5FB5ImMVGmGQIUmCZLfUjHSDVWoHEkCMgAdzAZ1BtG6e3oNI5wiKMgA6pBJqhgJBQkdurSyNcbCDOVKZZBGMK8qKBpTyup9FpdPo-Dq6WKPvEvuQwr4+TZdlq5ja6NSemR6aMzvBOY7-uMaNUqIi6gr2bqAY9gebPXritayiNvljeJHMK7kzIAEqQfTIYBUKI4vgQl6kYAqABazGYAGsbgAFKjwJzjdjAACCzAwADdi8w0DcqjVwDcAHRT+hAA)

[Original sequence diagram link to refine further](https://sequencediagram.org/index.html#initialData=C4S2BsFMAIGEHsC0BnEBzAdiDbregPIAOkG0AYvAE4C2yAUPUQIZWgDGILGwcSqmaAFVkkKoxFjEAPmKkK1OgC5oAEXQhV0ADLw02enLKVayGZKoqTNaADMQkcABNk0YPDshw4PBnoWZIwVTKy8ffCdmYGZDEmNFM2kLK0VoDEhIF2h2fnQsHH9RKkC44OVoAFEaZi9oZicnKkhkVwAKAGpoACEAZQA5PFs0jKdMgEoAHQx4IZyUPLFCqVlS62QVHoBXACMaMFj5NZLDhJUAcVIxKJhbVJzRtw97DCdoKeAACxBXW9oAemQOz2LRA8D8QSOKxOIWgPXcTWggN231QYLs1Gy8BoRCgoDR+AA7NBIgBPBgQhLmIobIEo0Fkb7QOJObC4OYCfJoRiQAAe3Ei2ygnQppjs4HgAHcDvFTDIEPNBMlYaRXpBqrV3JiFWQJWAPujaJjRvR5RzhEUZABZVQACVCLy1Zt+NjREsg21QwEg9BV3L5zBezEFkEASERqeQ25jMXC2cVSkV0GRnABaAEFYCoAApUeBOTbsaIBgBuVCjaCmrRom2QvGw7HAmwevQGM0dCyoY3oKfTMgAKsxkABrVypzMASRUsCa1zbgmiQ6mv1nZGrYgr2wHmWgaOb29muUwYk73dgcoPZCVPRV0DVNR8mvZeTcA8HJvP5uW1rtOj0Bi-fZfEdx3OSBeHnYcDW6fo92XRZ+yHICxytW0VAAJWaIgwVEeh-2kU0n0vD5JWfBC321D9imkL8VCCR85xfAAaaAmhZJoC2gVcqEeA0bCXOjOR9F4yLNAIoRlcp1H0LRdH0MhWnYANoG2GBmE2TUQCGHoegITt8MVC0xLKdYKDCXweKNGBWgxFgWi3IsQGYaBwGwQdOwTRJ3MnD5IHYQcoJbfdtTEaBWg0xF4XGaUjLPcjCOI51EU2GhqioEk6gdfjtyIPE-D0i8DM8vhtSiyETyzHM8wLKMMBLMtoF1T5oAAOhaxhGCAA)

### Tasks

* [x] Refactor co-sign component to be a component that has 1 field: E-mailaddress of co-signer (so it no longer allows you to actually co-sign). This can be required or not.
* [x] Once a submission is submitted with co-signing email filled in, sent an email to the co-signer emailadress.
* [x] Generate a form/submission code/hash (can be typically the internal reference number) upon submitting
* [x] Make sure sending the submission to a registration backend is blocked (and stored for X days)
* [ ] Make the email sent to the co-signer configurable in the admin config. Make sure the email contains the form/submission code and the number of days its possible to fill in.
* [x] Create fixed co-sign form view (in SDK)
* [x] Step 1: Login to co-sign form with DigiD/e-Herkenning/eIDAS
* [x] Step 2: Fill in form/submission code
* [x] Step 3: Retrieve and show the summary page of the original form submission
* [x] Submit and thus complete so-sign.
* [x] Sent the blocked submission to the registration backend
* [ ] Sent reminder after configurable Y days (Y < X) to the co-signer.
* [ ] Sent an email to the user and co-signer once the submission is processed OR the the user if it expired.

### More specific tasks

Form builder: 
- [x] Add new co-sign component (special email component)
- [x] Add warning about having no more than 1 co-sign component per form
- [x] Add migration to convert old co-sign components to the new version https://github.com/open-formulieren/open-forms/pull/3079
- [x] Add a warning or an error in case the plugin specified for the cosign is not enabled on the form.

Backend:
- [x] Add method to model to extract co-sign component
- [x] Add attribute to submission "cosign_complete"
- [x] Update the on_completion tasks: the registration happens only if co-sign has been completed or if it's not required.
- [x] Add a task to send an email to the co-signer
- [x] Modify confirmation email template to mention that the form will be processed after cosign https://github.com/open-formulieren/open-forms/pull/3085
- [ ] Add an admin filter for forms with cosign and custom confirmation templates => Then the form designers can easily find which templates should be updated to mention the cosign. 
- [ ] Make a note of manual intervention in the CHANGELOG about updating the confirmation template to mention cosign. https://github.com/open-formulieren/open-forms/issues/3122
- [ ] Add a task for a confirmation email after co-sign to the main person with the co-signer in CC => Needs https://github.com/maykinmedia/mail-cleaner/issues/2 first
- [x] Expose that a form is a co-sign form
- [x] Add to auth module a signal for when a co-signer logs in
- [x] Implement backend view to search for the submission https://github.com/open-formulieren/open-forms/pull/3090
- [x] Implement endpoint to persist co-sign https://github.com/open-formulieren/open-forms/pull/3118
- [ ] Make co-sign email template configurable (Add a dev view to visualise it). It should include the deadline for cosigning.
- [x] Add 'log out' button to the 'search submission' view. https://github.com/open-formulieren/open-forms/pull/3134
- [ ] Add a reminder email when approaching the deadline for cosigning
- [ ] Add a notification if the cosigner failed to cosign on time.

Frontend:
- [x] Add co-sign component 
- [x] Move old co-sign component https://github.com/open-formulieren/open-forms-sdk/pull/411
- [x] Add co-sign button to a form https://github.com/open-formulieren/open-forms-sdk/pull/423 + https://github.com/open-formulieren/open-forms/pull/3107
- [x] Submission summary view: add privacy notice, log-out button https://github.com/open-formulieren/open-forms-sdk/pull/427
- [x] Confirmation view: download PDF report
- [x] Add route to cosign summary page

More things that came to mind:
- [ ] If a form has a cosign component, maybe that step should have 'login required' checked?
- [x] The cosign component should have the `authPlugin` attribute as required?
- [ ] Refactor the `cosign_data` attribute on the submission to also use Auth info model

Clean up:
- [ ] Remove the old co-sign component (Frontend/Backend)
- [ ] Remove co-sign flow from all plugins
- [ ] Following this discussion: https://github.com/open-formulieren/open-forms/pull/3085#discussion_r1206547012 remove the try/except clauses around the send email
- [ ] Use json path query to get form definitions: https://github.com/open-formulieren/open-forms/pull/3085#discussion_r1206543366
- [ ] Deal with the case where an MTA sends us a report that the co-sign email was undeliverable https://github.com/open-formulieren/open-forms/issues/3126

See additional comment: https://github.com/open-formulieren/open-forms/issues/1530#issuecomment-1586858384

SilviaAmAm:
Fixes #1530


src/openforms/submissions/models/submission.py
```diff
@@ -642,3 +648,31 @@ def get_prefilled_data(self):
...
+
+        # Cosign not complete, but required
+        if self.form.cosigning_required:
+            return True
+
+        # Co-sign not complete, not required, but the optional cosign component was filled in with an email
+        if self.get_cosigner_email():
+            return True
+
+        return False
```
CharString, line 678: Doesn't this read like English already?

```suggestion
    def waiting_on_cosign(self) -> bool:
        return (not self.cosign_complete 
            and self.form.cosigning_required
            or self.get_cosigner_email() != ""
        )
```

We are waiting when we're not done and it's required or we have an email set.

CharString (changes requested):
What happens if an MTA sends us a report that `test@test.nl` was undeliverable? Is that in the Parts to come? :crossed_fingers: :wink:

src/openforms/submissions/models/submission.py
```diff
@@ -642,3 +648,31 @@ def get_prefilled_data(self):
...
+            for (
+                component,
+                upload_info,
+                data_path,
+                configuration_path,
+            ) in iterate_data_with_components(
+                form_step.form_definition.configuration, self.data
+            ):
+                if component["type"] == "cosign":
+                    return glom(self.data, data_path, default=None)
```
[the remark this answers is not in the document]
CharString: Whatever it becomes, we want to `@functools.lru_cache(maxsize=1)` it too; this function is called a few times directly or indirectly via `waiting_on_cosign`

src/openforms/submissions/models/submission.py
```diff
@@ -642,3 +648,31 @@ def get_prefilled_data(self):
...
+
+    @property
+    def waiting_on_cosign(self) -> bool:
+        if self.cosign_complete:
+            return False
+
+        # Cosign not complete, but required
+        if self.form.cosigning_required:
+            return True
```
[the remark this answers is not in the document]
SilviaAmAm: I was thinking that in situations where a form needs co-signing not in all situations there would be a second person (the example I had in mind is if it's a form about something children-related, then there aren't necessarily 2 peoples that need to confirm the data). But maybe this should be handled by making the cosign field hidden or in a step not applicable :thinking:

If we go down this route I need to change the co-sign component to always be required.

src/openforms/submissions/models/submission.py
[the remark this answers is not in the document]
SilviaAmAm: It's true, getting the data from the `SubmissionValueVariable` once we have the key could be a better approach!

So this could become:
- jsonpathquery to get the form definition with the cosign => from here we get the key of the cosign component
- retrieve the `SubmissionValueVariable` to get the value.

src/openforms/submissions/models/submission.py
SilviaAmAm: Heh I always struggle to parse in my head multiple bool conditions at once, so I tend to spread them out :sweat_smile:  I can make it a bit more compact.

src/openforms/submissions/models/submission.py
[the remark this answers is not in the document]
SilviaAmAm: Since we discussed that the json-path query shouldn't be used, for now I added the lru cache to the function. If this function causes performance problems, we can look into saving the key of the cosign component on the form or something.

SilviaAmAm:
> What happens if an MTA sends us a report that test@test.nl was undeliverable? Is that in the Parts to come?

Mhmmm :thinking: Atm if the `send_mail_html` raises an error, then the task will raise an error. But I don't think that this covers the case of receiving a report that the email was undeliverable? I am not sure how to handle this case, I will add it to the list of things to do.

CharString (approved):
If you think this is the most readable `waiting_on_cosign`. I'm fine with it.
My kneejerk to the if-else statements that returns bool literals is rewrite them to single return of disjunctive or conjunctive normal form. But I'm different. :grimacing:. That's what maths homework does to you. :)

src/openforms/forms/models/form.py
```diff
@@ -345,6 +346,16 @@ def get_authentication_backends_display(self):
         "authentication backend(s)"
     )
 
+    def get_cosign_component(self) -> JSONObject | None:
```
sergei-maertens, line 349: perhaps we can define a more strict `TypedDict` for the co-sign component shape in `openforms.formio.components.custom`, similar to how we have `openforms.formio.components.vanilla.FileComponent` to accurately reflect the shape of the object returned?

src/openforms/registrations/tasks.py
```diff
@@ -103,6 +105,13 @@ def register_submission(submission_id: int) -> None:
         # if it's already successfully registered, do not overwrite that.
         return
 
+    if submission.waiting_on_cosign:
+        logger.debug(
+            "Skipping registration for submission '%s' as it hasn't been co-signed yet.",
+            submission,
+        )
+        return
```
sergei-maertens, line 113: Perhaps we should also have a `logevent` record here to show in the admin for a given submission?

src/openforms/submissions/attachments.py
```diff
@@ -99,74 +100,6 @@ class UploadContext:
     num_uploads: int
 
 
-def _iterate_data_with_components(
```
sergei-maertens, line 102: ah well this explains why this code looked familiar :)

src/openforms/submissions/models/submission.py
```diff
@@ -14,6 +15,7 @@
 
 from openforms.config.models import GlobalConfiguration
 from openforms.formio.datastructures import FormioConfigurationWrapper
+from openforms.formio.utils import iterate_data_with_components
```
sergei-maertens, line 18: please expose this function via `openforms.formio.service` which is the place for public API. It's okay to just re-export it from the utils module, but we should avoid importing from other app modules other than the `service` module.

src/openforms/submissions/models/submission.py
```diff
@@ -656,3 +663,28 @@ def get_prefilled_data(self):
                 variable.key: variable.value for variable in prefill_vars
             }
         return self._prefilled_data
+
+    @lru_cache
+    def get_cosigner_email(self) -> str | None:
```
sergei-maertens, line 668: I believe `lru_cache` is not thread-safe and as we deploy uwsgi with multiple threads, this cache would be shared by different requests (each request gets its own thread) in the same process. In this case, it's mitigated because it's on an Submission instance which is typically created per thread, but I think I'd rather see [`django.utils.functional.cached_property` ](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.functional.cached_property)for this.

src/openforms/submissions/models/submission.py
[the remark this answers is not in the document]
sergei-maertens: the json-path query is still a viable option, we just need to properly implement it as a [custom database function](https://dev.to/idrisrampurawala/writing-custom-django-database-functions-4dmb), which allows us to use the ORM and all the protections it has and avoid the `.raw` call.

That is also the proper way to support different database vendors in the future.

src/openforms/submissions/models/submission.py
```diff
@@ -656,3 +663,28 @@ def get_prefilled_data(self):
...
+            ) in iterate_data_with_components(
+                form_step.form_definition.configuration, self.data
+            ):
+                if component["type"] == "cosign":
+                    return glom(self.data, data_path, default=None)
+
+    @property
+    def waiting_on_cosign(self) -> bool:
+        if self.cosign_complete:
+            return False
```
sergei-maertens, line 684: line not covered by tests :grimacing:

src/openforms/submissions/tasks/co_sign.py
```diff
@@ -0,0 +1,53 @@
...
+        send_mail_html(
+            subject=f"Co-sign request for {submission.form.name}",
+            html_body=content,
+            from_email=settings.DEFAULT_FROM_EMAIL,
+            recipient_list=[recipient],
+            text_message=content,
+        )
+    except Exception:
+        logevent.cosigner_email_failure(submission)
+        raise
```
sergei-maertens, line 51: this doesn't work how you expect it - because we use django-yubin, any `send_email*` calls actually just queue the e-mail which will be effectively sent via celery beat and another celery task.

So any SMTP borking will not show up here, and the logevent will never report failure even though it may fail.

I think logging SMTP failures is a whole different topic/task and should probably be removed from this PR.

src/openforms/emails/templates/emails/confirmation/content.html
```diff
@@ -12,6 +12,10 @@
 Your reference is: {{ tt_openvariable }} public_reference {{ tt_closevariable }}<br>
 <br>
 
+{{ tt_openblock }} if waiting_on_cosign {{ tt_closeblock }}
+This form will not be processed until it has been co-signed. A co-sign request was sent to {{ tt_openvariable }} cosigner_email {{ tt_closevariable }}.
+{{ tt_openblock }} endif {{ tt_closeblock }}
```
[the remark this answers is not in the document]
SilviaAmAm: I linked this issue with the checkbox I had in the main ticket

src/openforms/submissions/models/submission.py
SilviaAmAm: The documentation says it is threadsafe: https://docs.python.org/3/library/functools.html#functools.lru_cache.
Should I still change it to a [django.utils.functional.cached_property](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.functional.cached_property)?

sergei-maertens: I think it's thread safe in the sense that multiple threads can use the cache, but the contents of the cache is shared across threads, but I'm not 100% sure about this!

src/openforms/submissions/tasks/co_sign.py
SilviaAmAm: Do you mean that the all the log events here should be removed?
I had been following how the confirmation email was doing it, should we remove it from there too?
```
    try:
        send_mail_html(
            subject,
            html_content,
            settings.DEFAULT_FROM_EMAIL,  # TODO: add config option to specify sender e-mail
            to_emails,
            text_message=text_content,
        )
    except Exception as e:
        logevent.confirmation_email_failure(submission, e)
        raise

    submission.confirmation_email_sent = True
    ```

sergei-maertens: Hrm, in this case it's probably just logging that the confirmation e-mail was scheduled correctly, so if we do the same for the cosign email, that's at least consistent. However, the wording of those log events is not correct as it doesn't prove that the e-mail was effectively sent successfully to an SMTP server :thinking: 

maybe `sent` should be reworded in the `scheduled` in the actual log messages? But ideally that should then be a separate issue + PR to keep the scope narrow here.

PR #2583, opened 2023-01-18
🧪 Runs tests in reverse order in CI

CharString:
Closes #2999

This fails some tests because they or for some reason not isolated.

Unit tests should be able to run in random order.

PR #3088, opened 2023-05-12
Test reverse order

bart-maykin:
draft pr to see if the test issolation problems are fixed.


sergei-maertens:
I'm going to close this - instead we should rebase the exsting PR #2583 to check if it's fixed (and we can then update that PR to add the additional CI job)

PR #3090, opened 2023-05-12
[#1530] Part 4 - Backend view for finding submission

### LINKED ISSUE #1530: Co-signing does not work with a DigiD Gateway (e.g. OIDC) or DigiD SSO

SilviaAmAm:
Fixes #1530 

Depends on https://github.com/open-formulieren/open-forms/pull/3085

### DIFF
--- a/src/openforms/authentication/utils.py
+++ b/src/openforms/authentication/utils.py
@@ -1,5 +1,7 @@
 from typing import Literal, Optional, TypedDict
 
+from rest_framework.request import Request
+
 from openforms.submissions.models import Submission
 
 from .constants import FORM_AUTH_SESSION_KEY, AuthAttribute
@@ -46,12 +48,8 @@ def store_registrator_details(
     )
 
 
-def has_valid_authentication_for_cosign(request, expected_plugin) -> bool:
-    if FORM_AUTH_SESSION_KEY not in request.session:
+def is_authenticated_with_plugin(request: Request, expected_plugin: str) -> bool:
+    try:
+        return request.session[FORM_AUTH_SESSION_KEY]["plugin"] == expected_plugin
+    except KeyError:
         return False
-
-    auth_data = request.session[FORM_AUTH_SESSION_KEY]
-    if auth_data["plugin"] == expected_plugin:
-        return True
-
-    return False
--- a/src/openforms/authentication/utils.py
+++ b/src/openforms/authentication/utils.py
@@ -2,7 +2,7 @@ from typing import Literal, Optional, TypedDict
 
 from openforms.submissions.models import Submission
 
-from .constants import AuthAttribute
+from .constants import FORM_AUTH_SESSION_KEY, AuthAttribute
 from .models import AuthInfo, RegistratorInfo
 
 
@@ -44,3 +44,14 @@ def store_registrator_details(
     RegistratorInfo.objects.update_or_create(
         submission=submission, defaults=registrator_auth
     )
+
+
+def has_valid_authentication_for_cosign(request, expected_plugin) -> bool:
+    if FORM_AUTH_SESSION_KEY not in request.session:
+        return False
+
+    auth_data = request.session[FORM_AUTH_SESSION_KEY]
+    if auth_data["plugin"] == expected_plugin:
+        return True
+
+    return False
--- a/src/openforms/submissions/forms.py
+++ b/src/openforms/submissions/forms.py
@@ -1,0 +1,40 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy as _
+
+from .models import Submission
+
+
+class SearchSubmissionForCosignForm(forms.Form):
+    code = forms.CharField(
+        label=_("Code"),
+        help_text=_(
+            "You should have received this code in the email requesting you to cosign a form."
+        ),
+        required=True,
+        widget=forms.TextInput(
+            attrs={
+                "class": "openforms-input",
+            }
+        ),
+    )
+    form_slug = forms.CharField(
+        label=_("form slug"),
+        help_text=_("The slug of the form to which the submission corresponds."),
+        required=True,
+        widget=forms.HiddenInput(),
+        disabled=True,
+    )
+
+    def clean(self):
+        if not Submission.objects.filter(
+            public_registration_reference=self.cleaned_data["code"],
+            form__slug=self.initial["form_slug"],
+            cosign_complete=False,
+        ).exists():
+            raise ValidationError(
+                _(
+                    "Could not find submission corresponding to this code that requires co-signing"
+                )
+            )
+        return self.cleaned_data
--- a/src/openforms/submissions/forms.py
+++ b/src/openforms/submissions/forms.py
@@ -18,23 +18,23 @@ class SearchSubmissionForCosignForm(forms.Form):
             }
         ),
     )
-    form_slug = forms.CharField(
-        label=_("form slug"),
-        help_text=_("The slug of the form to which the submission corresponds."),
-        required=True,
-        widget=forms.HiddenInput(),
-        disabled=True,
-    )
+
+    def __init__(self, instance, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.form = instance
 
     def clean(self):
-        if not Submission.objects.filter(
+        submission = Submission.objects.filter(
             public_registration_reference=self.cleaned_data["code"],
-            form__slug=self.initial["form_slug"],
+            form=self.form,
             cosign_complete=False,
-        ).exists():
+        ).first()
+        if not submission:
             raise ValidationError(
                 _(
-                    "Could not find submission corresponding to this code that requires co-signing"
+                    "Could not find a submission corresponding to this code that requires co-signing"
                 )
             )
+        self.cleaned_data["submission"] = submission
         return self.cleaned_data
--- a/src/openforms/submissions/templates/submissions/find_submission_for_cosign.html
+++ b/src/openforms/submissions/templates/submissions/find_submission_for_cosign.html
@@ -17,12 +17,9 @@
         <div class="card__body">
             {% block card_body %}
                 <p class="body">
-                    <form method="post">{% csrf_token %}
+                    <form method="post" action=".">{% csrf_token %}
                         {% include "includes/forms/errorlist.html" with errors=form.non_field_errors only %}
 
-                        {% for hidden in form.hidden_fields %}
-                            {{ hidden }}
-                        {% endfor %}
                         {% include "includes/forms/field_wrapper.html" with field=form.code type='text' only %}
 
                         <div class="openforms-toolbar openforms-toolbar--reverse openforms-toolbar--bottom">
--- a/src/openforms/submissions/templates/submissions/find_submission_for_cosign.html
+++ b/src/openforms/submissions/templates/submissions/find_submission_for_cosign.html
@@ -1,0 +1,38 @@
+{% extends 'ui/views/abstract/detail.html' %}
+{% load i18n %}
+
+{% block extra_css %}
+    {% include "forms/sdk_css_snippet.html" %}
+    {{ block.super }} {# Called after SDK CSS for custom design overrides #}
+{% endblock %}
+
+{% block card %}
+    <div class="card">
+        <header class="card__header">
+            <h1 class="title">
+                {% trans "Retrieve submission for co-signing" %}
+            </h1>
+        </header>
+
+        <div class="card__body">
+            {% block card_body %}
+                <p class="body">
+                    <form method="post">{% csrf_token %}
+                        {% include "includes/forms/errorlist.html" with errors=form.non_field_errors only %}
+
+                        {% for hidden in form.hidden_fields %}
+                            {{ hidden }}
+                        {% endfor %}
+                        {% include "includes/forms/field_wrapper.html" with field=form.code type='text' only %}
+
+                        <div class="openforms-toolbar openforms-toolbar--reverse openforms-toolbar--bottom">
+                            <button class="openforms-button openforms-button--primary" type="submit">
+                                <span class="openforms-button__label">{% trans "Submit" %}</span>
+                            </button>
+                        </div>
+                    </form>
+                </p>
+            {% endblock %}
+        </div>
+    </div>
+{% endblock %}
--- a/src/openforms/submissions/tests/test_views.py
+++ b/src/openforms/submissions/tests/test_views.py
@@ -1,0 +1,371 @@
+from unittest.mock import patch
+
+from django.http import HttpRequest
+from django.test import TestCase, override_settings
+from django.urls import reverse
+
+from furl import furl
+
+from openforms.accounts.tests.factories import UserFactory
+from openforms.authentication.constants import FORM_AUTH_SESSION_KEY
+from openforms.submissions.tests.factories import SubmissionFactory
+
+
+class SearchSubmissionForCosignView(TestCase):
+    def test_successfully_get_form(self):
+        SubmissionFactory.from_components(
+            components_list=[
+                {
+                    "key": "cosign",
+                    "type": "cosign",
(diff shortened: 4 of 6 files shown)


CharString (changes requested):
Any reason why we're going with a POST? Can't we email a link and handle this with a get so a cosigner doesn't have to cut and paste? It removes the user clipboard from the flow...

How long are the codes we're generating? Are they easy to type in? I'm thinking of public libraries that are acting as [Informatiepunt Digitale Overheid](https://www.informatiepuntdigitaleoverheid.nl/?lang=en)

(conflicting concerns, I know :grimacing: I'm just thinking out loud)

SilviaAmAm:
> Any reason why we're going with a POST? Can't we email a link and handle this with a get so a cosigner doesn't have to cut and paste? It removes the user clipboard from the flow...

This is because the municipalities have a new approach where "you shouldn't send links in emails because it encourages clicking on links in emails"...

> How long are the codes we're generating? Are they easy to type in? 

If they are internal references, they are something like `OF-` + 5 or 6 characters. If they are references from external registration backends we don't have much control over them (I think for ZGW apis they are like `ZAAK-` + some numbers. Not sure about StUff-ZDS).

CharString (approved):

src/openforms/authentication/utils.py
```diff
@@ -44,3 +46,10 @@ def store_registrator_details(
     RegistratorInfo.objects.update_or_create(
         submission=submission, defaults=registrator_auth
     )
+
+
+def is_authenticated_with_plugin(request: Request, expected_plugin: str) -> bool:
+    try:
+        return request.session[FORM_AUTH_SESSION_KEY]["plugin"] == expected_plugin
+    except KeyError:
+        return False
```
sergei-maertens, line 55: Missing test coverage

src/openforms/submissions/views.py
```diff
@@ -235,3 +240,41 @@ def get_sendfile_opts(self) -> dict:
...
+    cosign_url = None
+
+    def has_permission(self):
+        """
+        The user should have authenticated with the auth plugin specified on the form for the co-sign component
+        """
+        form = get_object_or_404(Form, slug=self.kwargs["form_slug"])
+        cosign_component = form.get_cosign_component()
+        expected_auth_plugin = cosign_component["authPlugin"]
+        return is_authenticated_with_plugin(self.request, expected_auth_plugin)
```
sergei-maertens, line 259: I think using `UserPassesTestMixin` instead of `PermissionRequiredMixin` is more appropriate here - the latter really typically refers to codename permissions like `"foo.can_bar"`. Given that you need to override the main implementation (`has_permission`) to make it work, I think the `def test_func` of the former mixin is more appropriate.

src/openforms/authentication/utils.py
```diff
@@ -44,3 +46,10 @@ def store_registrator_details(
     RegistratorInfo.objects.update_or_create(
         submission=submission, defaults=registrator_auth
     )
+
+
+def is_authenticated_with_plugin(request: Request, expected_plugin: str) -> bool:
```
sergei-maertens, line 51: I can't shake the feeling that we have more/similar checks that would benefit from refactoring to this utility, e.g. the submission resume flow. Have you checked those?

src/openforms/submissions/forms.py
```diff
@@ -0,0 +1,40 @@
...
+                "class": "openforms-input",
+            }
+        ),
+    )
+    form_slug = forms.CharField(
+        label=_("form slug"),
+        help_text=_("The slug of the form to which the submission corresponds."),
+        required=True,
+        widget=forms.HiddenInput(),
+        disabled=True,
```
sergei-maertens, line 26: doesn't `disabled` result in the slug not actually being submitted in the `POST` request?

The `clean` method also doesn't use this data, but instead uses the `self.initial["form_slug"]` which doesn't seem 100% correct either.

src/openforms/submissions/tests/test_views.py
```diff
@@ -0,0 +1,371 @@
...
+from django.urls import reverse
+
+from furl import furl
+
+from openforms.accounts.tests.factories import UserFactory
+from openforms.authentication.constants import FORM_AUTH_SESSION_KEY
+from openforms.submissions.tests.factories import SubmissionFactory
+
+
+class SearchSubmissionForCosignView(TestCase):
```
sergei-maertens, line 14: wouldn't it make more sense to use `WebTest` here instead of Django's bare test case?

Also - why are there `self.client.login(...)` calls here - shouldn't this be used by non-staff users so anonymous Django users?

src/openforms/submissions/views.py
```diff
@@ -235,3 +240,41 @@ def get_sendfile_opts(self) -> dict:
...
+
+    def get_initial(self):
+        return {"form_slug": self.kwargs["form_slug"]}
```
sergei-maertens, line 262: probably makes more sense to:

1. Look up the form already with `get_object_or_404`
2. Pass the form instance to the form via `def get_form_kwargs()`
3. Extend the form `__init__` to take the `form` instance and store it in `self.form`. That way, you don't need the form slug hidden input at all, you just get the form from the URL and perform the access control checks.

src/openforms/submissions/views.py
```diff
@@ -235,3 +240,41 @@ def get_sendfile_opts(self) -> dict:
...
+
+    def form_valid(self, form):
+        submission = Submission.objects.get(
+            public_registration_reference=form.cleaned_data["code"]
+        )
```
sergei-maertens, line 267: You can avoid this additional query because you're already doing it in the form clean method - so look up this submission (with all the relevant validation) and if all is find, you can store it in `self.cleaned_data["submission"]`

Then here in the view, it's just a `submission = form.cleaned_data["submission"]` access and continue business as usual.

src/openforms/submissions/templates/submissions/find_submission_for_cosign.html
```diff
@@ -0,0 +1,38 @@
...
+        <header class="card__header">
+            <h1 class="title">
+                {% trans "Retrieve submission for co-signing" %}
+            </h1>
+        </header>
+
+        <div class="card__body">
+            {% block card_body %}
+                <p class="body">
+                    <form method="post">{% csrf_token %}
```
sergei-maertens, line 20:
```suggestion
                    <form method="post" action=".">{% csrf_token %}
```

src/openforms/submissions/views.py
```diff
@@ -235,3 +240,41 @@ def get_sendfile_opts(self) -> dict:
...
+        add_submmission_to_session(submission, self.request.session)
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return self.cosign_url
+
+    def dispatch(self, request, *args, **kwargs):
+        next_url = self.request.GET.get("next")
```
sergei-maertens, line 275: isn't this parameter lost on the form submit `POST` call?

So you'd have to include that with a hidden input, and add it to the form as well so you get proper URL validation instead of using raw GET/POST args?

src/openforms/submissions/forms.py
```diff
@@ -0,0 +1,40 @@
...
+        super().__init__(*args, **kwargs)
+
+        self.form = instance
+
+    def clean(self):
+        submission = Submission.objects.filter(
+            public_registration_reference=self.cleaned_data["code"],
+            form__slug=self.form.slug,
+            cosign_complete=False,
+        ).first()
```
sergei-maertens, line 32:
```suggestion
        submission = Submission.objects.filter(
            public_registration_reference=self.cleaned_data["code"],
            form=self.form,
            cosign_complete=False,
        ).first()
```

sergei-maertens (approved):

PR #3091, opened 2023-05-12
Feature/2999 msgraph issues

bart-maykin:
Fixes test issolation problems in test_backend.py  for #2999

### DIFF
--- a/src/openforms/registrations/contrib/microsoft_graph/tests/test_backend.py
+++ b/src/openforms/registrations/contrib/microsoft_graph/tests/test_backend.py
@@ -38,13 +38,13 @@ class MockFolder:
 
 
 @temp_private_root()
+@patch(
+    "openforms.registrations.contrib.microsoft_graph.models.MSGraphRegistrationConfig.get_solo",
+    return_value=MSGraphRegistrationConfig(service=MSGraphServiceFactory.create()),
+)
 class MSGraphRegistrationBackendTests(TestCase):
     @classmethod
     def setUpTestData(cls):
-        config = MSGraphRegistrationConfig.get_solo()
-        config.service = MSGraphServiceFactory.create()
-        config.save()
-
         cls.options = dict(folder_path="/open-forms/")
 
     @classmethod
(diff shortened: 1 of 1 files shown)


src/openforms/registrations/contrib/microsoft_graph/tests/test_backend.py
```diff
@@ -38,13 +38,13 @@ def upload_file(self, *args, **kwargs):
 
 
 @temp_private_root()
+@patch(
+    "openforms.registrations.contrib.microsoft_graph.models.MSGraphRegistrationConfig.get_solo",
+    return_value=MSGraphRegistrationConfig(service=MSGraphServiceFactory.create()),
```
sergei-maertens, line 43: this now performs a database query at import-time and is not the correct approach, any data created for tests should be done so in the transaction.

The decorator approach is not ideal here - more robust would be starting the mock in the `setUp` method, but it's a bit of a multi-step process:

In the `setUpTestData` you want to call the factory so that the dummy service is only set up once (and rolled back when the test transaction is rolled back):

```py
@classmethod
def setUpTestData(cls):
    super().setUpTestData()
    
    cls.msgraph_config_patcher = patch(
        "openforms.registrations.contrib.microsoft_graph.models.MSGraphRegistrationConfig.get_solo",
        return_value=MSGraphRegistrationConfig(service=MSGraphServiceFactory.create())
    )
    cls.options = dict(folder_path="/open-forms/")
```

then, in the `setUp` (which runs before every tests), you can start the mock:

```py
def setUp(self):
    super().setUp()

    self.msgraph_config_patcher.start()
    self.addCleanup(self.msgraph_config_patcher.stop)
```

For a similar flow, see: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/api/tests/test_endpoints.py#L48

Under the hood this is exactly wat happens in the decorator and context manager - they essentially wrap the function/class/codeblock in a block that calls `.start` before and `.stop` after.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3098, opened 2023-05-16
Introduce FormioData data structure

### LINKED ISSUE #2784: Create smarter data structure for Formio data
Thought occurred after seeing https://github.com/open-formulieren/open-forms/pull/2783/files

We often have some assumptions about the component keys not being nested, e.g. `component["key"] in submission_data`, which is too simplistic. Formio allows specifying nested paths by using `.`-characters as separators, e.g. `nested.key` which results in a nested submission data structure `{"nested": {"key": "value"}}`.

A smarter datastructure could apply this lookup behaviour with simple python operations and make it more intuitive to do the right thing.

Known places that should use this:

* `openforms.submissions.logic.datastructures.DataContainer.get_updated_step_data`
* `openforms.submissions.models.submission_value_variable.SubmissionValueVariablesState`
* `openforms.submissions.form_logic`
* `openforms.variables.utils.get_variables_for_context`
* `openforms.submissions.viewsets.SubmissionStepViewSet.logic_check`
* ...

sergei-maertens:
Closes #2784

Adds a datastructure to deal with the formio keys that can contain periods (`.`) which expand into nested objects.

There are now some chained calls that convert between `FormioData` <-> `dict` that would probably benefit from changing their signature to accept `FormioData`, but I'm not sure how troublesome that will be if we need to do backportable fixes in those areas :thinking: Let's discuss?

### DIFF
--- a/src/openforms/formio/datastructures.py
+++ b/src/openforms/formio/datastructures.py
@@ -107,6 +107,8 @@ class FormioData(UserDict):
     details (such as using ``glom`` for this).
     """
 
+    data: dict[str, JSONValue]
+
     def __getitem__(self, key: Hashable):
         return cast(JSONValue, glom(self.data, key))
 
--- a/src/openforms/formio/datastructures.py
+++ b/src/openforms/formio/datastructures.py
@@ -89,6 +89,24 @@ class FormioConfigurationWrapper:
 
 
 class FormioData(UserDict):
+    """
+    Handle formio (submission) data transparently.
+
+    Form.io supports component keys in the format 'topLevel.nested' which get converted
+    to deep-setting of object properties (using ``lodash.set`` internally). This
+    datastructure mimicks that interface in Python so we can more naturally perform
+    operations like:
+
+    .. code-block:: python
+
+        data = FormioData()
+        for component in iter_components(...):
+            data[component["key"]] = ...
+
+    without having to worry about potential deep assignments or leak implementation
+    details (such as using ``glom`` for this).
+    """
+
     def __getitem__(self, key: Hashable):
         return cast(JSONValue, glom(self.data, key))
 
--- a/src/openforms/formio/datastructures.py
+++ b/src/openforms/formio/datastructures.py
@@ -84,3 +84,7 @@ class FormioConfigurationWrapper:
             component = cast(Component, glom(self.configuration, path))
             nodes.append(component)
         return all(is_visible_in_frontend(node, values) for node in nodes)
+
+
+class FormioData(dict):
+    pass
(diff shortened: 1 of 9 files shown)


src/openforms/formio/datastructures.py
```diff
@@ -84,3 +86,38 @@ def is_visible_in_frontend(self, key: str, values: DataMapping) -> bool:
...
+            data[component["key"]] = ...
+
+    without having to worry about potential deep assignments or leak implementation
+    details (such as using ``glom`` for this).
+    """
+
+    data: dict[str, JSONValue]
+
+    def __getitem__(self, key: Hashable):
+        return cast(JSONValue, glom(self.data, key))
```
sergei-maertens, line 113: I *know* we don't like the casts, but the premise really is json-serializable data when working with formio submission data. Though the datatypes thing might complicate this a bit more?

src/openforms/submissions/form_logic.py
```diff
@@ -188,29 +190,17 @@ def evaluate_form_logic(
...
-                component_empty_value := get_component_empty_value(component)
-            ):
-                if (
-                    component
-                    and not is_visible_in_frontend(component, data_container.data)
-                    and component.get("clearOnHide")
-                ):
-                    assign(data_diff, key, component_empty_value, missing=dict)
-                    continue
-
```
SilviaAmAm, line 207: I'm confused by how it still works :thinking: Where do we check the `clearOnHide` now?

SilviaAmAm (commented):
I like this new wrapper! 

Regarding the discussion points: it's probably good to use the wrapper consistently and then when we have to deal with backports we do a separate PR to the stable branches with the fixes with glom instead?

src/openforms/submissions/form_logic.py
sergei-maertens: it's a couple lines above actually! close the initialization of `data_diff`. The exact same code but written differently.

I tried adding a test for the missing coverage, but then it turned out that this line of code was just never hit (dead code) because of the code above. CTRL + F on `clearOnHide` and you'll find it.

src/openforms/formio/datastructures.py
```diff
@@ -84,3 +86,38 @@ def is_visible_in_frontend(self, key: str, values: DataMapping) -> bool:
...
+
+    def __setitem__(self, key: Hashable, value: JSONValue):
+        assign(self.data, key, value, missing=dict)
+
+    def __contains__(self, key: Hashable) -> bool:
+        try:
+            self[key]
+        except PathAccessError:
+            return False
```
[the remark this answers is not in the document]
sergei-maertens: > I like this new wrapper!
> 
> Regarding the discussion points: it's probably good to use the wrapper consistently and then when we have to deal with backports we do a separate PR to the stable branches with the fixes with glom instead?

I'm okay with this approach

SilviaAmAm (approved):

PR #3101, opened 2023-05-17
Flesh out SDK embedding configuration documentation

### LINKED ISSUE #2940: Improve documentation of `IS_HTTPS ` setting.

2.1.0

Background
---

When an Open Forms installation is deployed in a setting where a SDK runs on a different domain. Various (Django) settings need to be configured. A documentation page ([Embedding the SDK in a page](https://open-forms.readthedocs.io/en/latest/developers/sdk/embedding.html)) exists to point an administrator to the relevant settings.

 - `CSRF_COOKIE_SAMESITE` should be set to` "None"` to make the CSRF cookie cross origin.
 - `CSRF_COOKIE_SECURE` should be set the `True` when the cookie to allow the CSRF cookie to be cross origin.

 - `SESSION_COOKIE_SAMESITE` should be set to` "None"` to make the session cross origin.
 - `SESSION_COOKIE_SECURE` should be set the `True` when the cookie to allow the session cookie to be cross origin.

These settings (among others) can be controlled using `IS_HTTPS` setting, which Is documented in "other settings" group on the [Enviornment configuration reference](https://open-forms.readthedocs.io/en/latest/installation/config.html?highlight=IS_HTTPS) as 

 > Used to construct absolute URLs and controls a variety of security settings. Defaults to the inverse of DEBUG.

The [Embedding the SDK in a page](https://open-forms.readthedocs.io/en/latest/developers/sdk/embedding.html) documentation does not mention this setting.

 The implementation for the `<TYPE>_COOKIE_SAMESITE` is to change the default value (allowing it to be overridden) while there is a tight coupling between `IS_HTTPS` and `<TYPE>_COOKIE_SECURE`, preventing it from being overridden.

The [Environment configuration reference](https://open-forms.readthedocs.io/en/latest/installation/config.html#cross-origin-resource-sharing-cors-settings) links to the Django documentation in the context of cross origin settings. Pointing an administrator to the variables mentions above.

Configuring Django according to its default settings may lead to broken configurations/installations as the `<TYPE>_COOKIE_SECURE` settings cannot be overridden. 

Suggestions
---

- Improve the explanation of the `IS_HTTPS` setting with what is does and when it should be configured.
- Mention `IS_HTTPS` on the [Embedding the SDK in a page](https://open-forms.readthedocs.io/en/latest/developers/sdk/embedding.html) documentation.
- Allow `<TYPE>_COOKIE_SECURE` to be overridden.

1. Navigate to the Open Forms documentation.
2. Navigate to the "Embedding the SDK in a page" page.
3. Navigate to the "Cross-Origin Resource Sharing (CORS) settings" section.
4. Navigate to the Django documentation.
5. Configure the `<TYPE>_COOKIE_SECURE=True` settings according to the Django documentation.
6. Configure the `<TYPE>_COOKIE_SAMESITE="None"` settings according to the Django documentation.

The expected result would be an Open Form installation working able to respond to cross origin requests. However; since `SameSite=None` cookies require `Secure` it may not.

The exact behavior is inconsistent among browsers. Leading to confusion about the actual issue (the setting not being picked up).

### Screen resolution

unknown

### Device

Desktop

### OS

MacOS

### Browser

Chrome

sergei-maertens:
Closes #2940

I am correcting some factual mistakes in the ticket itself too, there really isn't more to do than documenting the behaviour here.

* Elaborate on how IS_HTTPS is used
* Make it clear that embedding is not possible in an HTTP context
* Break up CORS settings in recommended/dangerous sections
* Document which settings should be typically specified to enable embedding

### DIFF
--- a/docs/installation/config.rst
+++ b/docs/installation/config.rst
@@ -83,11 +83,27 @@ Cross-Origin Resource Sharing (CORS) settings
 
 See: https://github.com/adamchainz/django-cors-headers
 
+* ``CORS_ALLOWED_ORIGINS``: A list of origins that are authorized to make
+  cross-site HTTP requests. Defaults to ``[]``. An example value would be:
+  ``https://cms.example.org,https://forms.example.org``
+
+* ``CSRF_TRUSTED_ORIGINS``: the list of trusted CSRF origins, e.g. ``cms.example.com``.
+  When embedding forms on third party sites, these third party domains need to be added
+  to the allowlist. The default value is taken from the ``CORS_ALLOWED_ORIGINS`` setting.
+  See also `the Django documentation <https://docs.djangoproject.com/en/3.2/ref/settings/#csrf-trusted-origins>`_.
+
+It is recommended to configure CORS explicitly by setting just ``CORS_ALLOWED_ORIGINS``.
+
+**Other CORS-settings**
+
+The settings below are available but should only be specified/used when you have
+intricate knowledge of how they are used. While wildcards and regexes may seem
+attractive options, we are currently limited by our framework in how
+``CSRF_TRUSTED_ORIGINS`` can be specified (which must be an explicit list without
+patterns or wildcards).
+
 * ``CORS_ALLOW_ALL_ORIGINS``: If ``True``, all origins will be allowed. Other
   settings restricting allowed origins will be ignored.Defaults to ``False``.
-
-* ``CORS_ALLOWED_ORIGINS``: A list of origins that are authorized to make
-  cross-site HTTP requests. Defaults to ``[]``.
 
 * ``CORS_ALLOWED_ORIGIN_REGEXES``: A list of strings representing regexes that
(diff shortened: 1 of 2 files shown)


docs/installation/config.rst
```diff
@@ -184,6 +196,29 @@ Other settings
 * ``IS_HTTPS``: Used to construct absolute URLs and controls a variety of
   security settings. Defaults to the inverse of ``DEBUG``.
 
+  .. warning::
+
+     We strongly recommended setting IS_HTTPS=False in development environments
```
SilviaAmAm, line 201: I know what you mean here, but I wonder if it could be misinterpreted when municipalities have 'dev environments' :thinking: So maybe you could add 'locally' in this sentence somewhere?

sergei-maertens: good ditinction - I've updated it, can you check if this is better?

SilviaAmAm (approved):

PR #3102, opened 2023-05-17
[#3040] Improve error messages when allowed types are configured in the General config

### LINKED ISSUE #3040: File upload component gives user-unfriendly error if the wrong file type is uploaded

2.1.x, 2.0.x

DH Taiga issue 465 / user story 466

If you have a component where only certain types of files are allowed and you upload the wrong type of file you get an error `Bestand is van het verkeerde type; het moet zijn application&#x2F;pdf`.
The mime type should ideally be the name of the extension, e.g.: `Bestand is van het verkeerde type; het moet zijn '.pdf'`
 

Create a form with:
- A file upload component that only accepts PDF files

Fill in the form:
- Upload a .txt file => Observe the error

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #3040

### DIFF
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -4,12 +4,13 @@ Implement backend functionality for core Formio (built-in) component types.
 Custom component types (defined by us or third parties) need to be organized in the
 adjacent custom.py module.
 """
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Literal, TypedDict
 
 from rest_framework.request import Request
 from rest_framework.reverse import reverse
 
 from csp_post_processor import post_process_html
+from openforms.config.constants import UploadFileType
 from openforms.config.models import GlobalConfiguration
 from openforms.typing import DataMapping
 from openforms.utils.urls import build_absolute_uri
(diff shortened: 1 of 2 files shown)


src/openforms/formio/components/vanilla.py
```diff
@@ -89,6 +95,14 @@ def rewrite_for_request(component: FileComponent, request: Request):
         if component.get("useConfigFiletypes", False):
             config = GlobalConfiguration.get_solo()
             component["filePattern"] = ",".join(config.form_upload_default_file_types)
+            component["file"].update(
+                {
+                    "allowedTypesLabels": [
+                        dict(UploadFileType.choices)[mimetype]
```
CharString, line 101:
```suggestion
                        getattr(UploadFileType.choices, mimetype)
```
getattr boils down to a single dict lookup in the `choices.__dict__`, instead of creating a new dict and then doing the lookup on each iteration.

CharString (changes requested):
Do you think it needs an e2e test? The others had, iirc.

SilviaAmAm:
> Do you think it needs an e2e test? The others had, iirc.

In the previous PR I had added an e2e test because it was the form designer JS (formio) that was modifying the configuration of the component. In this case, it is the backend that dynamically modifies the configuration of the component based on the global configuration. So I thought it would be enough to check that the method that dynamically updates the component config actually modifies the file component.

I could add a test that calls the get endpoint for a form definition and check that the configuration contains the expected changes (but I dont think an e2e test is needed).

src/openforms/formio/components/vanilla.py
SilviaAmAm: `UploadFileType.choices` is actually a list of tuples `(key, value)`, so i'm not sure if there is a way to get around transforming it to a dict :thinking:

CharString:
>  (but I dont think an e2e test is needed).
You're right. I was distracted by the screenshot in the issue.

sergei-maertens (approved):

Issue #3105, opened 2023-05-19
Refactor global configuration


pi-sigma:
### Thema / Theme

API

### Omschrijving / Description

Currently only the privacy policy of the global configuration is exposed via our API (`api/v2/config/privacy_policy_info`). During discussion of #2809, the possibility of refactoring the global configuration came up. Should we expose the entire configuration?

PR #3108, opened 2023-05-22
[#2809] make link title for downloading submission report configurable

### LINKED ISSUE #2809: Allow changing of "Download uw inzending als pdf document" string
### Thema / Theme

Form designer

DH: 403
This text is unclear for some users, as "pdf" and "inzending" are terms not used in any previous steps of the form

_No response_

_No response_

pi-sigma:
Closes #2809 
Front-end: open-formulieren/open-forms-sdk#424

- modified `src/openforms/config/migrations/0043_globalconfiguration_templates.py` to use snapshot of translatable fields in order to avoid trying to access `submission_report_download_link_title`, which doesn't exist at the time the migration is run.


src/openforms/submissions/api/serializers.py
SilviaAmAm, line 1: I wonder if it would make more sense to expose this on the form rather than the submission? 

Since the literals, the confirmation email template are on the form?

pi-sigma: I thought about that too, but I feel there is something arbitrary about how parts of the general configuration get exposed via different parts of our API. I created an issue ([link](https://github.com/open-formulieren/open-forms/issues/3105)) about the possibility of exposing the entire global configuration via its own endpoint, which would put an end to this. This was originally part of the present ticket but I separated it out (discussed with @joeribekker). I'll bring this issue up during standup.

CharString: I agree with it being on the form unless the strings and configs can change as a function of submissions.
So the question is, can we imagine configurations that differ depending on submitted values? e.g. will the title of the file contain the name of the submitter, or something similar. Do we want to allow `{{variable}}` interpollation in these strings?

SilviaAmAm (approved):

src/openforms/config/models.py
```diff
@@ -64,6 +64,12 @@ class GlobalConfiguration(SingletonModel):
         default=runtime_gettext(_("Thank you for submitting this form.")),
         validators=[DjangoTemplateValidator()],
     )
+    submission_report_download_link_title = models.CharField(
+        max_length=128,
+        verbose_name=_("Submission download link title"),
+        help_text=_("The title of the link to download the report of a submission."),
+        default=runtime_gettext(_("Download PDF")),
+    )
```
sergei-maertens, line 72: shouldn't this field be marked translatable? Admins/form designers should be able to enter the NL and EN text, would be weird to have an entire form in EN and then suddenly this download link is NL.

You should then also be able to drop the `runtime_gettext` with just the normal gettext as the default language is NL anyway.

sergei-maertens (changes requested):

src/openforms/submissions/api/serializers.py
sergei-maertens: this is not about the file name/title of the file, just about the text of the download link, right?

src/openforms/config/models.py
pi-sigma: I thought we use `runtime_gettext` to defer evaluation of the translatable strings. Is that not required in the present case?

sergei-maertens (approved):

Issue #2651, opened 2023-02-03
Cookies added by analytics: localhost instead of TLD


LaurensBurger:
### Product versie / Product version

2.0.3 - latest

### Omschrijf het probleem / Describe the bug

![image](https://user-images.githubusercontent.com/112621537/216619881-23dd4a0b-48e7-4443-a2d4-83a101d9b868.png)

Not sure if this is just visual issue - don't see errors in the requests 

### Screen resolution

None

### Device

None

### OS

None

joeribekker:
I guess it picks whats configured in Sites?

sergei-maertens:
`contrib.sites` is not enabled/installed, must be something else

joeribekker:
Workaround can be done: Manually change localhost to the proper TLD.

PR #3111, opened 2023-05-22
Refactor some analytics tools code

sergei-maertens:
Some refactoring following #2651

~~Depends on #3110 which needs to be merged first~~

---

I didn't particularly like the repeat `if` branches and checks in the `save` method, or the `isinstance` checks in the utilities + broken type annotations.


setup.cfg
```diff
@@ -1,7 +1,7 @@
 [pycodestyle]
 [pep8]
 [flake8]
-ignore=E203,E261,E501,F405,W293,W291,W503
+ignore=E203,E261,E501,E731,F405,W293,W291,W503
```
sergei-maertens, line 4: this allows assigning lambda's to variables

not needed anymore in this PR, but multiple people have been annoyed by that restriction, leaving it up to the reviewer to decide if we do/don't want this

src/openforms/config/migrations/0030_piwik_pro_support_squashed_0037_globalconfiguration_form_upload_default_file_types.py
```diff
@@ -12,149 +12,6 @@
...
-    update_analytical_cookies,
-    update_csp,
-)
-
-# Functions from the following migrations need manual copying.
-# Move them and any dependencies into this file, then update the
-# RunPython operations to refer to the local versions:
-
-
-def move_analytics_config(apps, _):
```
sergei-maertens, line 29: this datamigration has executed during the 2.0 -> 2.1 upgrade, and upgrading to 2.2 from 2.0 is blocked so we can safely remove this

SilviaAmAm (approved):

PR #3115, opened 2023-05-24
Fixes for bugs noticed while re-testing the SDK

sergei-maertens:
No ticket for these, but I ran into some issues while testing the React upgrade in the SDK going through the appointment flow.

### DIFF
--- a/src/openforms/appointments/views.py
+++ b/src/openforms/appointments/views.py
@@ -2,8 +2,6 @@ import logging
 
 from django.views.generic import RedirectView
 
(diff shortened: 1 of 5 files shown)


SilviaAmAm (commented):
Code cov is complaining that one line is not covered by tests, but apart from that it looks good!

sergei-maertens:
> Code cov is complaining that one line is not covered by tests, but apart from that it looks good!

yeah, it's because the demo plugin wasn't tested to begin with :(

PR #3119, opened 2023-05-25
 copy the "Bevestingsmailsjabloon" when a form is copied

### LINKED ISSUE #3070: "Bevestingsmailsjabloon" gets removed after copy

2.1.x + 2.2.x

internal: LV72

The mail confirmation template is always empty after copying a form.
original form:
image
New form
image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

foarsitter:
This pull request fixes #3070 by erasing `self.confirmation_email_template.pk` and updating the relation to `copy` so its added as a new row with the correct foreign key.

### DIFF
--- a/src/openforms/forms/models/form.py
+++ b/src/openforms/forms/models/form.py
@@ -1,4 +1,5 @@
 import uuid as _uuid
+from contextlib import suppress
 from copy import deepcopy
 from typing import List, Optional
 
@@ -363,6 +364,7 @@ class Form(models.Model):
 
     @transaction.atomic
     def copy(self):
+        from ...emails.models import ConfirmationEmailTemplate
         from .form_variable import FormVariable
 
         form_steps = self.formstep_set.all().select_related("form_definition")
@@ -407,6 +409,11 @@ class Form(models.Model):
             variable.pk = None
             variable.form = copy
             variable.save()
+
+        with suppress(ConfirmationEmailTemplate.DoesNotExist):
+            self.confirmation_email_template.pk = None
+            self.confirmation_email_template.form = copy
+            self.confirmation_email_template.save()
 
         return copy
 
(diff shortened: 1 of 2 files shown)


src/openforms/forms/models/form.py
```diff
@@ -408,6 +409,13 @@ def copy(self):
             variable.form = copy
             variable.save()
 
+        try:
+            self.confirmation_email_template.pk = None
+            self.confirmation_email_template.form = copy
+            self.confirmation_email_template.save()
+        except ObjectDoesNotExist:
```
sergei-maertens, line 416: Generally it's best practice to catch the model-specific subclass of this exception:

```py
```

and then:

```suggestion
        except ConfirmationEmailTemplate.DoesNotExist:
```

sergei-maertens (approved):
Hi @foarsitter, thanks for this contribution!

This patch looks great and thanks for adding a regression test! I have one small remark.

I see that CI also has opinions, you can resolve those by running:

```bash
isort src
black src
```

src/openforms/forms/models/form.py
foarsitter: The avoid circular imports I never import models from other packages in `models.py`. Of course I can place the import at function level but not very fond of the either.

sergei-maertens: A local import is okay here - though you are right in that it's a bit of a code smell 

IMO this model should either just be model fields of the Form model or be moved into the forms app, but that's a topic for another improvement 😅

sergei-maertens:
> @sergei-maertens, ah rookie mistake, fixed the formatting.
> 
> Is it an idea to add a [.pre-commit-config.yaml](https://github.com/cookiecutter/cookiecutter-django/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/.pre-commit-config.yaml)?
> 
> Thanks for looking into it!

I believe a long time ago we had some performance issues or issues with it not using the venv isort/black, but I'm definitely open to reconsider it! I think most of us rely on local git hooks and editor plugins to run the formatters on save.

src/openforms/forms/models/form.py
foarsitter: Pushed with a local import and decided to use `with suppress` instead of the `try/except`.

sergei-maertens (approved):

PR #3124, opened 2023-05-26
Support hosting on subpaths

annashamray:


sergei-maertens:
@annashamray please create an issue for this and relate PR and issue

CharString (changes requested):
Can you change the tests for the middlewares you fixed so they pass a  `SCRIPT_NAME` to `django.test.Client` or `django.test.RequestFactory` or their method calls?

annashamray:
@CharString unit tests for subpath are added

src/openforms/templates/includes/cookie-notice.html
```diff
@@ -4,7 +4,7 @@
 
 {% url 'cookie_consent_cookie_group_list' as manage_url %}
 
-{% if cookie_groups and request.path != manage_url %}
+{% if cookie_groups and request.path_info != manage_url %}
```
sergei-maertens, line 7: shouldn't `manage_url` (which is just a `reverse` call) have applied the `SCRIPT_NAME` prefix already? So in that case, the `request.path` branch would be correct.

annashamray: `url` template tag doesn't use `request` in its implementation. Even when I test it with `SUBPATH=/of`, `manage_url` still equals `/cookies/`

src/openforms/conf/base.py
```diff
@@ -558,6 +558,17 @@
...
 
+# Deal with being hosted on a subpath
+SUBPATH = config("SUBPATH", None)
+if SUBPATH:
+    if not SUBPATH.startswith("/"):
+        SUBPATH = f"/{SUBPATH}"
+
+if SUBPATH and SUBPATH != "/":
+    STATIC_URL = f"{SUBPATH}{STATIC_URL}"
+    MEDIA_URL = f"{SUBPATH}{MEDIA_URL}"
```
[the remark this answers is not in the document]
annashamray: The issue for tests - https://github.com/open-formulieren/open-forms/issues/3138

src/openforms/templates/includes/cookie-notice.html
sergei-maertens: huh that's odd, because the django `url` tag or underlying `reverse` don't need this request object either. It relies on `get_script_prefix()` https://github.com/django/django/blob/40a2c811e3ed7fdb26ab4443e39e113c2fcf2aa9/django/urls/base.py#L109 which is a (thread) local.

```py
>>> from django.urls import get_script_prefix, set_script_prefix, reverse
>>> get_script_prefix()
'/'
>>> reverse("cookie_consent_cookie_group_list")
'/cookies/'
>>> set_script_prefix("/of")
>>> reverse("cookie_consent_cookie_group_list")
'/of/cookies/'

```

sergei-maertens (approved):

PR #3125, opened 2023-05-31
:sparkles: [2606] added haal centraal v2 api support

### LINKED ISSUE #2606: Add support: Haal Centraal BRP v2.0
### Thema / Theme

Plugins

Internal: Utr 41, Enschede 9

Currently, version 1.3.1 is available in OF but everyone is migrating to v2.0 which will also be the national standaard (and I'm super happy that they didn't call it 1.0.0 again because it moved to a different department on paper!).

More info on HC BRP 2.0: https://brp-api.github.io/Haal-Centraal-BRP-bevragen/v2/getting-started

**Tasks**

* [x] In `haalcentraalconfig` rename existing label (Haal Centraal API) to "BRP Bevragen Personen service"
* [x] In `haalcentraalconfig` add a version dropdown: `BRP Bevragen Personen version`: 1.3 and 2.0
* [x] Make a data migration and make sure the default is set to 1.3 (currently supported version)
* [x] Change the existing plugin to support 2.0 behaviour (mostly about doing a POST instead of a GET)
* [x] Change the existing plugin to switch to 1.3 and 2.0 based on the `haalcentraalconfig`.
* [ ] Check if there any attribute changes in the 2.0 API compared to 1.3. There is a script or management command (created by @Bartvaderkin ) to generate the mapping from API attributes to choices.
* [x] If there are any changes, we need different mappings for both API versions (please keep a base mapping and deviate only where needed)
* [x] Restructure tests to differentiate between 1.3 and 2.0 mocks/tests
* [x] Add specific 2.0 tests

bart-maykin:
fixes #2606

- Added version dropdown in HaalCentraalConfig singleton
- Added version support for v2 api
- Added new atributes for the v2 api
- Made the HaalCentraal plugin work with v1 and v2 of the api


src/openforms/prefill/contrib/haalcentraal/tests/test_plugin.py
```diff
@@ -107,8 +101,224 @@ def test_get_prefill_values_http_404(self, m):
         expected = {}
         self.assertEqual(values, expected)
 
-    def test_get_available_attributes(self):
+
+class HaalCentraalPrefillApiV2Test(TestCase):
```
sergei-maertens, line 105: See the earlier remark about adding a base class for tests, as it looks like the actual tests are identical.

It's easy to add a regression test for V1 and then forget to add it for V2 or vice versa :thinking:

sergei-maertens (changes requested):
The actual implementation looks okay in terms of functionality, I mostly have a bunch of remarks about the code organization and keeping things small/unit-testable.

sergei-maertens (changes requested):
Please also look at the Codecov annotations for missing test coverage.

Furthermore, the haal centraal plugin `check_config` method also needs updating - it currently doesn't adapt itself to the GET vs. POST of each respective client.

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -5,61 +5,57 @@
...
     requires_auth = AuthAttribute.bsn
 
-    request_kwargs = dict(headers={"Accept": "application/hal+json"})
-
     @staticmethod
     def get_available_attributes() -> list[tuple[str, str]]:
-        return Attributes.choices
+        config = get_config()
+        if not config:
+            return Attributes.choices
```
bart-maykin, line 40: Because the service of haal centraal config is allowed to be empty this will make that get_available_attributes always returns the default value.

sergei-maertens (approved):

PR #3129, opened 2023-05-31
Added image scanning and updated base image

joeribekker:


Dockerfile
```diff
@@ -9,9 +9,11 @@ FROM openformulieren/open-forms-sdk:${SDK_RELEASE} as sdk-image
 
 # Stage 1 - Backend build environment
 # includes compilers and build tooling to create the environment
-FROM python:3.10.9-slim-bullseye AS backend-build
+FROM python:3.10.11-slim-bullseye AS backend-build
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
+RUN apt-get update
+RUN apt-get upgrade -y
+RUN apt-get install -y --no-install-recommends \
```
sergei-maertens, line 16: you really want to have this in a single `RUN` statement because each RUN instruction creates a separate image layer which needlessly increases the image size

sergei-maertens (changes requested):

joeribekker:
Refinement: Not adding this because of its limited value and there will be a ticket to create "dailies" or the like.

PR #3141, opened 2023-06-06
[#3114] Permission check messes submission state up

### LINKED ISSUE #3114: Blocked 'next' button does not become active if user not logged in as admin

2.1.x, latest (d6ac4d1b60199b02c81179d3825adb3bb862b108)

DH Taiga issue 503

If there is logic that blocks/unblocks the 'next' button of a form step, this seems to not work properly if the user is not logged in as admin.

Create a form with:
1. A step with a text field called `Field A`
2. A step with a text field called `Field B`
3. A logic rule saying: If `Field A` is equal to `Field B` and the step `step 2` has been reached, block going to the next step.
Screenshot from 2023-05-24 16-05-40

Fill in the form as follows (:warning:  Make sure you are not logged in as admin!!):
1. Fill `A` in the first step
2. Fill `B` in the second step and go to the next step (overview page)
3. Now go back to the first step and change `A` to `B`, then go to the next step.
4. In the second step, the 'next' button is blocked. Change `B` to `A` => The button does not become active again

Note: This doesn't happen if you are logged in as admin. Then the button becomes active as expected.

The button should not stay blocked.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #3114 

Needs backport to 2.0.x and 2.1.x

### DIFF
--- a/src/openforms/submissions/tests/test_models.py
+++ b/src/openforms/submissions/tests/test_models.py
@@ -610,3 +610,10 @@ class SubmissionTests(TestCase):
 
         with self.subTest("nested"):
             self.assertEqual(submission3.cosigner_email, "test@test.nl")
+
+    def test_clear_execution_state_without_execution_state(self):
+        submission = SubmissionFactory.create()
+
+        self.assertFalse(hasattr(submission, "_execution_state"))
+
+        submission.clear_execution_state()


src/openforms/submissions/tests/test_models.py
```diff
@@ -610,3 +610,10 @@ def test_get_cosigner_email(self):
 
         with self.subTest("nested"):
             self.assertEqual(submission3.cosigner_email, "test@test.nl")
+
+    def test_clear_execution_state_without_execution_state(self):
+        submission = SubmissionFactory.create()
+
+        self.assertFalse(hasattr(submission, "_execution_state"))
+
+        submission.clear_execution_state()
```
CharString, line 619: As from the previous comment, the situation is a lot hairier.

The notion of staleness on the `SubmissionState` is broken too: the `load_execution_state` has a `refresh` and disregards references to the current state object it has returned previously and hands out a new one. That's unsafe. It probably hasn't resulted in problems, because most callers, get one, do some work and let it go out of scope and not keep it around.  I haven't checked. (I'm not a Rust compiler :wink:)

CharString (changes requested):
The issue is bigger than this patch, let's clean this up proper.

PS, but keep this patch around, maybe squashed. If the backport of the proper solution is too cumbersome to backport.

src/openforms/submissions/models/submission.py
```diff
@@ -492,6 +492,15 @@ def load_execution_state(self, refresh: bool = False) -> SubmissionState:
...
 
+    def clear_execution_state(self) -> None:
+        if not hasattr(self, "_execution_state"):
+            return
+
+        for submission_step in self._execution_state.submission_steps:
+            submission_step._can_submit = True
+
+        del self._execution_state
+
```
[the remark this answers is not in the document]
CharString: Hmm.. `__del__` is problematic. It isn't guaranteed to be called and CPython 3.11 doesn't call them in the face of circular references... So it the cleanup should be explicitly done.

src/openforms/submissions/tests/test_models.py
sergei-maertens: Can we create a new issue/big topic for this? This state (submission but also variables) has grown way beyond the initial scope/ideas and I think it's definitely up for refactoring. We should now have a clearer picture of what it needs to do and should probably be able to break this out into a proper service layer.

It's important to decouple the usage throughout the stack from the backing database models, but we also need to implement the DB aspect properly for performance reasons. We should be able to collectively come up with a set of (functional) requirements. I think the test suite is implementation detail agnostic enough to be able to achieve this.

SilviaAmAm:
Backported to:
- 2.0.x: https://github.com/open-formulieren/open-forms/pull/3152
- 2.1.x: https://github.com/open-formulieren/open-forms/pull/3153

PR #3146, opened 2023-06-09
[#2859] Enable configuring which ZGW API set to use

### LINKED ISSUE #2859: ZGW component - Per formulier kunnen bepalen waar (welke ZGW API) het document moet worden weggeschreven
### Thema / Theme

Admin

Userstory
**Als Beheerder** van  Open Formulieren
**Wil ik** kunnen instellen per formulier waar het aanvraagdocument (pdf) en eventuele bijlagen worden weggeschreven (DRC)
**Zodat** e-Suite gemeenten kunnen bepalen in welke registratie ((OpenZaak/eSuite) of anders systeem) de documenten worden weggeschreven en de zaak kan worden afgehandeld door medewerkers.

NB. Dit is belangrijk voor Gemeenten die het eSuite zaaksysteem gebruiken om Open Formulieren te kunnen inzetten voor zowel gedurende de migratie tijd waarin zowel de eSuite als het nieuwe Zaak afhandel component van de Podium D gebruikt wordt.

27d1f130-4364-4b81-9c25-024508e3e14c

_No response_

_No response_

### LINKED ISSUE #2922: When configuring the Object API registration backend there is no mention of the ZGW API configuration
### Thema / Theme

Documentation

If you use the Object API registration backend, then the configuration tells you to create a service for the document API and the Object API and then to select them here: `/admin/registrations_objects_api/objectsapiconfig/1/change/`.

When designing a form, if the form designer wants to override the informatieobjecttype per attachment, there is a tab 'registration' inside the component:
Screenshot from 2023-03-22 13-37-40

This uses the the Catalogi API to show the available informatieobjecttypen. However, in the Object API registration documentation, it doesn't mention that this should be configured under `/admin/zgw_apis/zgwconfig/`

_No response_

_No response_

SilviaAmAm:
Fixes #2859 
Fixes #2922


src/openforms/contrib/zgw/service.py
```diff
@@ -52,22 +55,22 @@ def create_zaak(
...
     config = ZgwConfig.get_solo()
-    return config.drc_service
+    zgw_config = config.default_zgw_api_group
+    return zgw_config.drc_service
 
 
-def partial_update_zaak(zaak_url: str, data: dict) -> dict:
-    config = ZgwConfig.get_solo()
-    client = config.zrc_service.build_client()
+def partial_update_zaak(zaak_url: str, data: dict, zrc_service: Service) -> dict:
```
sergei-maertens, line 62: The order of arguments is a bit inconsistent - `create_zaak` takes the service as first argument, this function as last. Can you normalize that?

I'm also wondering if it wouldn't make more sense to pass the `client` instead of the service? This would unlock some connection pooling with constructs like:

```py
with zrc_service.build_client() as client:
    partial_update_zaak(client, ...)
    ... # other operations with same client, (eventually) sharing a connection pool
```

src/openforms/contrib/zgw/service.py
```diff
@@ -52,22 +55,22 @@ def create_zaak(
...
+    client = zrc_service.build_client()
     zaak = client.partial_update("zaak", data, url=zaak_url)
     return zaak
 
 
-def set_zaak_payment(zaak_url: str, partial: bool = False) -> dict:
+def set_zaak_payment(zaak_url: str, service: Service, partial: bool = False) -> dict:
```
sergei-maertens, line 68: see above remarks about passing `client` instead of `service` and order of arguments

src/openforms/forms/tests/e2e_tests/test_registration_backend_conf.py
```diff
@@ -0,0 +1,184 @@
...
+
+class FormDesignerRegistrationBackendConfigTests(E2ETestCase):
+    async def test_configuring_zgw_api_group(self):
+        """
+        The flow in this test is:
+        - Configure the Registration backend to use set 1 of the configured ZGW apis
+        - Go to the file upload component and check that the query parameter in the request to the informatieobjecttype endpoint is the PK of the right group
+        - Go back to the registration tab and change which ZGW API group should be used
+        - Go to the file component and check that the query parameter in the request changes
+        """
```
sergei-maertens, line 23: Docstrings are parsed in the test runner in verbosity mode, so for this test it would output `The flow in this test is:`, which isn't particularly useful. PEP-257 was written about docstrings, but the summary is to stick to the format:

```py
"""
Describe in imperative form the summary
{blank line}
Provide an elaborate description with all the necessary information, and
even ``RST``/Sphinx syntax.
"""
```

src/openforms/registrations/contrib/zgw_apis/api/filters.py
```diff
@@ -0,0 +1,24 @@
...
+            "in this set will be returned. Otherwise, the Catalogi API in the default ZGW API set will be returned."
+        ),
+        label=_("ZGW API set"),
+        required=False,
+    )
+    registration_backend = serializers.CharField(
+        help_text=_("The ID of the registration backend to use."),
+        label=_("Registration backend ID"),
+        required=False,
+    )
```
sergei-maertens, line 24: I think you should be able to dynamically set `choices` as well to fetch them from the registration plugin register?

src/openforms/registrations/contrib/zgw_apis/api/views.py
```diff
@@ -3,35 +3,59 @@
...
 from rest_framework.views import APIView
+from zds_client import Client
 from zgw_consumers.api_models.base import factory
 from zgw_consumers.api_models.catalogi import Catalogus, InformatieObjectType
 from zgw_consumers.service import get_paginated_results
 
 from openforms.api.views import ListMixin
 from openforms.registrations.contrib.zgw_apis.models import ZgwConfig
 
+from ...objects_api.models import ObjectsAPIConfig
```
sergei-maertens, line 14: cross-app relative import, but the fact that this code is now relevant for both ZGW and Objects API makes me think the view itself should rather be moved to `openforms.contrib.catalogi` or something like that

src/openforms/registrations/contrib/zgw_apis/api/views.py
```diff
@@ -3,35 +3,59 @@
...
+        if registration_backend == "zgw-create-zaak" and zgw_api_group is not None:
+            return zgw_api_group.ztc_service.build_client()
+        elif registration_backend == "objects_api":
+            config = ObjectsAPIConfig.get_solo()
+            if config.catalogi_service:
+                return config.catalogi_service.build_client()
+
+        config = ZgwConfig.get_solo()
+        if config.default_zgw_api_group:
+            return config.default_zgw_api_group.ztc_service.build_client()
```
sergei-maertens, line 47: I think this makes more sense as a method on the query params serializer?

src/openforms/registrations/contrib/zgw_apis/migrations/0007_move_singleton_data.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("zgw_apis", "0006_zgwapigroupconfig"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            move_singleton_data_to_zgw_api_set_model, migrations.RunPython.noop
```
sergei-maertens, line 37: Now migrating forwards and then backwards again leads to data loss. Can you implement the reverse operation where you grab the attached group config put that config back in the singleton?

sergei-maertens (commented):
Can you please address the missing coverage annotations from Codecov?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -71,6 +83,18 @@ class Meta:
             RoltypeOmschrijvingValidator(),
         ]
 
+    def validate(self, attrs):
```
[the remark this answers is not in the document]
SilviaAmAm: I was thinking that since `zgw_api_group` is not required, if it is not present, then `validate_zgw_api_group` will not run?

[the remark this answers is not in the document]
sergei-maertens: Ah that could be, I'm not sure. Give it a try and see if it runs or not? If it doesn't, it'd make more sense to attach the error to this field instead of `non_field_errors` though :)

src/openforms/contrib/zgw/service.py
```diff
@@ -46,37 +48,32 @@ def create_zaak(
...
-    return partial_update_zaak(zaak_url, data)
+    return partial_update_zaak(zrc_client, zaak_url, data)
 
 
 def create_document(
     name: str,
     base64_body: str,
     options: dict,
-    get_drc=default_get_drc,
+    get_drc: Callable,
```
sergei-maertens, line 74: FYI, you can also specify the types of args and return type:

```suggestion
    get_drc: Callable[[], ZGWClient],
```
-> no args, `ZGWClient` return type

src/openforms/js/components/form/file.js
```diff
@@ -18,12 +19,37 @@ const REGISTRATION = {
...
+        custom(context) {
+          const instance = context.instance;
+          const registrationInfo =
+            instance?.options?.openForms?.registrationBackendInfoRef?.current;
+          if (!registrationInfo) return [];
+
+          // if the url starts with '/', then formio will prefix it with the formio
+          // base URL, which is of course wrong. We there explicitly use the detected
+          // host.
+          let url = new URL(getFullyQualifiedUrl('/api/v2/registration/informatieobjecttypen'));
```
sergei-maertens, line 33:
```suggestion
          const url = '/api/v2/registration/informatieobjecttypen';
```
we don't need this workaround anymore because there's no formio config sitting in the way anymore (so you can also drop the comment above). I think you can then also just use the literal string in the `get` call instead of assigning to an intermediate variable.

PR #3155, opened 2023-06-13
Appointments rework

### LINKED ISSUE #2471: Change appointments flow

In summary:

### Tasks

**Optional**

* [x] #3193

### Mockup

**Step 1**

**Step 2**

Alternative:

image

**Step 3**

**Step 4**

sergei-maertens:
(Partially) closes #2471

* [x] Put new appointments configuration behind a feature flag
* [x] Implement (optionally) configuring the single appointment location (globally)

For a separate PR:

* ~~Implement filtering possible locations/products by the configured global location~~
* ~~Support multiple products in endpoints parameters~~

### DIFF
--- a/src/openforms/appointments/tests/test_admin.py
+++ b/src/openforms/appointments/tests/test_admin.py
@@ -1,5 +1,4 @@
 from contextlib import contextmanager
-from typing import cast
 from unittest.mock import patch
 
 from django.test import tag
@@ -26,9 +25,8 @@ from .factories import AppointmentInfoFactory
 
 @contextmanager
 def patch_registry(register):
-    field = cast(
-        AppointmentBackendChoiceField, AppointmentsConfig._meta.get_field("plugin")
-    )
+    field = AppointmentsConfig._meta.get_field("plugin")
+    assert isinstance(field, AppointmentBackendChoiceField)
     validators = [
         validator
         for validator in field.validators
@@ -36,10 +34,11 @@ def patch_registry(register):
     ]
     assert len(validators) == 1, "Expected only one PluginValidator"
     validator = validators[0]
-    with patch("openforms.appointments.admin.register", register), patch(
-        "openforms.appointments.utils.register", register
-    ), patch.object(field, "registry", register), patch.object(
-        validator, "registry", register
+    with (
+        patch("openforms.appointments.admin.register", register),
+        patch("openforms.appointments.utils.register", register),
+        patch.object(field, "registry", register),
+        patch.object(validator, "registry", register),
     ):
         yield
 
@@ -169,7 +168,8 @@ class AppointmentsConfigAdminTests(WebTest):
                     _("Please configure the plugin first"),
                 )
 
-            config = cast(AppointmentsConfig, AppointmentsConfig.get_solo())
+            config = AppointmentsConfig.get_solo()
+            assert isinstance(config, AppointmentsConfig)
             config.plugin = "test"
             config.save()
 
--- a/src/openforms/forms/admin/mixins.py
+++ b/src/openforms/forms/admin/mixins.py
@@ -1,5 +1,3 @@
-from typing import cast
-
 from django.utils.encoding import force_str
 from django.utils.translation import gettext as _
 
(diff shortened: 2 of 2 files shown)


src/openforms/appointments/admin.py
```diff
@@ -10,18 +13,37 @@
...
 @admin.register(AppointmentsConfig)
 class AppointmentsConfigAdmin(SingletonModelAdmin):
     def formfield_for_dbfield(self, db_field, request, **kwargs):
         if isinstance(db_field, AppointmentBackendChoiceField):
-            assert not db_field.choices
-            _old = db_field.choices
-            db_field.choices = register.get_choices()
-            field = super().formfield_for_dbfield(db_field, request, **kwargs)
-            db_field.choices = _old
-            return field
```
sergei-maertens, line 24: this monkey patching suffers from race conditions, making a copy instead does not

src/openforms/appointments/admin.py
```diff
@@ -10,18 +13,37 @@
...
+            except ValueError:
+                return super().formfield_for_dbfield(
+                    db_field,
+                    request,
+                    disabled=True,
+                    widget=AdminTextInputWidget(
+                        attrs={"placeholder": _("Please configure the plugin first")}
+                    ),
+                    **kwargs,
+                )
```
sergei-maertens, line 39: dynamic selects using JS/React would be a better UX, but also takes a lot more time.

src/openforms/appointments/tests/test_admin.py
```diff
@@ -11,13 +14,36 @@
...
+        for validator in field.validators
+        if isinstance(validator, PluginExistsValidator)
+    ]
+    assert len(validators) == 1, "Expected only one PluginValidator"
+    validator = validators[0]
+    with patch("openforms.appointments.admin.register", register), patch(
+        "openforms.appointments.utils.register", register
+    ), patch.object(field, "registry", register), patch.object(
+        validator, "registry", register
+    ):
```
CharString, line 43:
```suggestion
    with (
        patch("openforms.appointments.admin.register", register),
        patch("openforms.appointments.utils.register", register),
        patch.object(field, "registry", register),
        patch.object(validator, "registry", register),
    ):

```

Sometimes you have to hold black's hand a bit; let it know we love it.

src/openforms/forms/admin/mixins.py
```diff
@@ -20,7 +22,7 @@ class FormioConfigMixin:
     def render_change_form(
         self, request, context, add=False, change=False, form_url="", obj=None
     ):
-        config = GlobalConfiguration.get_solo()
+        config = cast(GlobalConfiguration, GlobalConfiguration.get_solo())
```
CharString, line 25: Same here. If we don't want to trip up production with an `AssertionError`, we can run with -O and wait for the bug report with weird behaviour. Or we can catch it and log it to Sentry and then let it go in with its unexpected behaviour.

I'm not saying the cast is wrong here. But code has a tendency to get copied. I've caught faulty casts before.
I'd rather take the performance hit of an assert, than have a cheap cast to make the type checker give me false confidence.

I do love the goal btw... just not the means.

CharString (changes requested):

src/openforms/appointments/admin.py
sergei-maertens: yeah but django's `ModelAdmin` instances are not thread safe and cause all these problems in the first place :( the alternative is re-implementing half of django's own admin formfield_for_dbfield, ModelForm fields from DB fields with custom admin widgets and a bunch of other things just to make the dynamic options be registered in a dropdown properly.

The Django admin code is an unfixable mess :(

src/openforms/appointments/tests/test_admin.py
```diff
@@ -11,13 +14,36 @@
...
 from ..models import AppointmentsConfig
 from ..registry import Registry
 from .factories import AppointmentInfoFactory
 
 
+@contextmanager
+def patch_registry(register):
+    field = cast(
+        AppointmentBackendChoiceField, AppointmentsConfig._meta.get_field("plugin")
+    )
```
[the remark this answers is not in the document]
sergei-maertens: ah that's more elegant indeed!

src/openforms/appointments/tests/test_admin.py
sergei-maertens: I *knew* there was a trick, I just couldn't remember what it was, thanks!

src/openforms/forms/admin/mixins.py
sergei-maertens: if it trips up production with `AssertionError`, we have bigger fish to fry!

PR #3158, opened 2023-06-15
Appointments: get all locations an filter products by location

### LINKED ISSUE #2471: Change appointments flow

sergei-maertens:
(Partially) closes #2471

Implemented getting all appointment locations (without any products provided) and filtering products based on location.

* Getting all locations is required to populate the options in the admin to hard-pin the location
* Filtering products based on a location is required to take the admin configuration into account

### DIFF
--- a/src/openforms/appointments/api/views.py
+++ b/src/openforms/appointments/api/views.py
@@ -21,7 +21,11 @@ from openforms.submissions.api.permissions import (
 )
 from openforms.submissions.models import Submission
 
-from ..api.serializers import (
+from ..base import AppointmentLocation, AppointmentProduct
+from ..exceptions import AppointmentDeleteFailed, CancelAppointmentFailed
+from ..models import AppointmentsConfig
+from ..utils import delete_appointment_for_submission, get_plugin
+from .serializers import (
     CancelAppointmentInputSerializer,
     DateInputSerializer,
     DateSerializer,
@@ -31,9 +35,6 @@ from ..api.serializers import (
     TimeInputSerializer,
     TimeSerializer,
 )
-from ..base import AppointmentLocation, AppointmentProduct
-from ..exceptions import AppointmentDeleteFailed, CancelAppointmentFailed
-from ..utils import delete_appointment_for_submission, get_plugin
 
 logger = logging.getLogger(__name__)
 
@@ -52,10 +53,15 @@ class ProductsListView(ListMixin, APIView):
 
     def get_objects(self):
         plugin = get_plugin()
+        config = AppointmentsConfig().get_solo()
+        assert isinstance(config, AppointmentsConfig)
+        kwargs = {}
+        if location_id := config.limit_to_location:
+            kwargs["location_id"] = location_id
         with elasticapm.capture_span(
             name="get-available-products", span_type="app.appointments.get_products"
         ):
-            return plugin.get_available_products()
+            return plugin.get_available_products(**kwargs)
 
 
 # The serializer + @extend_schema approach for querystring params is not ideal, the
--- a/src/openforms/appointments/contrib/jcc/client.py
+++ b/src/openforms/appointments/contrib/jcc/client.py
@@ -1,7 +1,9 @@
+from zeep.client import Client
+
 from .models import JccConfig
 
 
-def get_client():
+def get_client() -> Client:
     service = JccConfig.get_solo().service
     assert service is not None
     return service.build_client()
--- a/src/openforms/appointments/contrib/jcc/plugin.py
+++ b/src/openforms/appointments/contrib/jcc/plugin.py
@@ -7,7 +7,9 @@ from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 
 from requests.exceptions import RequestException
+from zeep.client import Client
 from zeep.exceptions import Error as ZeepError
+from zgw_consumers.concurrent import parallel
 
 from openforms.plugins.exceptions import InvalidPluginConfiguration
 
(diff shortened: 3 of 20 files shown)


src/openforms/appointments/api/views.py
```diff
@@ -52,10 +53,15 @@ class ProductsListView(ListMixin, APIView):
 
     def get_objects(self):
         plugin = get_plugin()
+        config = AppointmentsConfig().get_solo()
+        assert isinstance(config, AppointmentsConfig)
```
SilviaAmAm, line 57: In what situation would this not be a `AppointmentsConfig`?

src/openforms/appointments/contrib/jcc/client.py
```diff
@@ -1,7 +1,9 @@
+from zeep.client import Client
+
 from .models import JccConfig
 
 
-def get_client():
+def get_client() -> Client:
     service = JccConfig.get_solo().service
     assert service is not None
     return service.build_client()
```
SilviaAmAm, line 9: I remember this comment from Chris about the asserts: https://github.com/open-formulieren/open-forms/pull/3024#discussion_r1177632873

Maybe the client should be checked and in case of `None` raise an informative error like 'service not configured' or something?

src/openforms/appointments/contrib/jcc/plugin.py
```diff
@@ -71,12 +80,47 @@ def get_available_products(
...
+        locations = [
+            AppointmentLocation(
+                identifier=identifier,
+                name=entry["locationDesc"],
+                address=entry["address"],
+                postalcode=entry["postalcode"],
+                city=entry["city"],
+            )
+            for identifier, entry in zip(location_ids, details)
+        ]
```
SilviaAmAm, line 109: Should this also be wrapped in a try/except like the `getGovLocations` calls so in case of errors in also raises the appropriate errors?

src/openforms/appointments/contrib/jcc/tests/test_plugin.py
```diff
@@ -200,3 +242,98 @@ def test_get_appointment_details(self, m):
...
+    @requests_mock.Mocker()
+    def test_get_locations_unexpected_exception(self, m):
+        m.post(requests_mock.ANY, exc=IOError("tubes are closed"))
+        product = AppointmentProduct(identifier="k@pu77", name="Kaputt")
+
+        with self.assertRaises(AppointmentException):
+            self.plugin.get_locations()
+
+        with self.assertRaises(AppointmentException):
+            self.plugin.get_locations(products=[product])
```
SilviaAmAm, line 295: Should there maybe also be a test in which the call for getting location details fail?

src/openforms/appointments/contrib/qmatic/client.py
```diff
@@ -13,17 +13,21 @@ class QmaticClient(Session):
...
     """
 
-    _config = None
+    _config: QmaticConfig | None = None
 
     def request(self, method: str, url: str, *args, **kwargs):
         if not self._config:
-            self._config = QmaticConfig.get_solo()
+            config = QmaticConfig.get_solo()
+            assert isinstance(config, QmaticConfig)
```
SilviaAmAm, line 21: same question here about when this is not a `QmaticConfig`

src/openforms/appointments/contrib/qmatic/tests/factories.py
```diff
@@ -5,6 +5,7 @@
 from ..models import QmaticConfig
 
 
+# TODO: consilidate this in src/openforms/contrib or src/zgw_consumers_ext
```
SilviaAmAm, line 8: consilidate? :thinking:

src/openforms/appointments/contrib/qmatic/tests/test_api_endpoints.py
```diff
@@ -1,30 +1,44 @@
...
+from .test_plugin import mock_response
 
-from ...constants import AppointmentDetailsStatus
-from ...contrib.qmatic.client import QmaticException
-from ...contrib.qmatic.tests.factories import QmaticConfigFactory
-from ...contrib.qmatic.tests.test_plugin import mock_response
-from ...models import AppointmentsConfig
-from ...tests.factories import AppointmentInfoFactory
 
+class MockConfigMixin:
```
SilviaAmAm, line 21: Could this use the mixin in src/openforms/appointments/contrib/qmatic/tests/utils.py?

src/openforms/appointments/api/views.py
CharString: @SilviaAmAm It's always an AppointmentsConfig, but the assertion makes that clear to type checkers. Strictly speaking a `typing.cast` is sufficient here, but I prefer asserts, as assert will catch mistakes where cast does nothing during runtime, and gives you false confidence during static analysis.

joeribekker: Really? An assert makes this clear to type checkers.

If so, then it's useful and should be patterniz d but in all other cases I think this is really redundant

src/openforms/appointments/contrib/jcc/client.py
sergei-maertens: in principle we have the config check page for this to guard against it. Whether this crashes in production with an `AssertionError` or an `AttributeError` (NoneType has no attribute `build_client`) has the same result in the end - it crashes.

src/openforms/appointments/contrib/jcc/plugin.py
sergei-maertens: ah frick, it should... this exception handling is very un-elegant :(

I'll add a test for it tomorrow and exception handling.

src/openforms/appointments/contrib/qmatic/tests/factories.py
sergei-maertens: consolidate typo, oops!

src/openforms/appointments/api/views.py
CharString: And to people too. :slightly_smiling_face:  From just reading this:
```python
    @classmethod
    def get_solo(cls):
        cache_name = getattr(settings, 'SOLO_CACHE', solo_settings.SOLO_CACHE)
        if not cache_name:
            obj, created = cls.objects.get_or_create(pk=cls.singleton_instance_id)
            return obj
        cache = get_cache(cache_name)
        cache_key = cls.get_cache_key()
        obj = cache.get(cache_key)
        if not obj:
            obj, created = cls.objects.get_or_create(pk=cls.singleton_instance_id)
            obj.set_to_cache()
        return obj
```
I can't infer the exact type of `obj`. It's the union of `Self` (because I know the contract of `cls.object.get_or_create`) and `Any` because we read "something" from some memory.  I think at `cache.get`, the library ought to make the assertion what we read from cache is an instance of `cls`, or, alternatively, `get_cache` should return a `Mapping[..., Self]` with the return type of `cls.get_cache_key()` at the ellipsis (probably `str`). In principle the cache is mutable by anything; and assertions guarantee our assumptions.

BTW this is exactly the use for the `assert` keyword: asserting the obvious, that ought to be true and may be optimized out during runtime with `python -O`

SilviaAmAm (approved):

PR #3160, opened 2023-06-15
Feature/2852 productaanvraag data field

### LINKED ISSUE #2852: Productaanvraag - Integratie Open Formulieren met de eSuite (Productaanvraag)
### Thema / Theme

API

**Specificatie en taken**

[Zie comments.](https://github.com/open-formulieren/open-forms/issues/2852#issuecomment-1571534979)

**Oorspronkelijke ticket**

Als Dimpact gemeente wil ik Open Formulieren kunnen integreren met de eSuite van ATOS zodat ik op langer termijn het eSuite formulieren component kan uitfaseren door het te vervangen door Open Formulieren omdat we de transitie maken van de eSuite naar het open source "PodiumD" van dimpact.

Om Open Formulieren met de eSuite te kunnen integreren willen we gebruik maken van "Object type" Productaanvraag voor een eenduidige gegevensuitwisseling. In een refinementsessie met Dimpact/ATOS (eSuite)/Maykin(Open formulieren)/ en Rotterdam is een aanpassing van de "Productaanvraag" gerefined waarbij het JSON schema is aangevuld.

Maykin heeft een voorstel gestuurd welke ATOS heeft bekeken/aangevuld en ter review heeft teruggestuurd naar Maykin.
Joeri B weet hiervan.

https://github.com/joeridiederen/objecttypes/blob/patch-1/community-concepts/productaanvraag/productaanvraag-dimpact

Voordeel van het gebruik van deze "productaanvraag" is een gemakkelijkere integratie tussen OF en eSuite. 

Eigenlijk is dit de "epic" waar de volgende userstories onderdeel van uitmaken.
(Wordt verder aangevuld). 

**Tasks**

* [x] #2853
* [x] #2854
* [ ] #2855
* [ ] #2856
* [ ] #2857
* [x] #2858

bart-maykin:
fixed #2852 

- Added object api template to the model and serializer
- added template tag to render form field data
- replaced the data we send to the object api from form field data to the template


src/openforms/registrations/contrib/objects_api/templatetags/objects_api.py
```diff
@@ -0,0 +1,18 @@
...
+
+from openforms.formio.rendering.structured import render_json
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def json_summary(context):
+    submission = context.get("_submission")
+    if submission:
```
SilviaAmAm, line 14: Maybe nicer to do:

```
if not submission:
    return {}
```
And then the rest (so that the 'important' part is not nested).

src/openforms/registrations/contrib/objects_api/utils.py
```diff
@@ -0,0 +1,10 @@
+from .models import ObjectsAPIConfig
+
+
+def get_registration_json_templates(submission) -> dict:
+    config = ObjectsAPIConfig.get_solo()
+    form_options = submission.form.registration_backend_options
```
SilviaAmAm, line 6: Maybe good to only retrieve the global if there is no template specified on the form? Then you would save a query.

Also, in [src/openforms/registrations/contrib/objects_api/models.py](https://github.com/open-formulieren/open-forms/pull/3160/files#diff-554ceb9c5db4b300f57e6a76cc6247c1482f0052f904f8ceae12e44d0cddb2c2) you set the default on the form in case it's not specified. So if you keep that, maybe this is not needed?

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -90,6 +104,17 @@ class ObjectsAPIConfig(SingletonModel):
...
     )
+    content_json = models.TextField(
+        _("Json field"),
+        validators=[
+            DjangoTemplateValidator(
+                backend="openforms.template.openforms_backend",
+            ),
+            JsonTemplateValidator(max_length=300),
+        ],
+        blank=True,
```
SilviaAmAm, line 115: I think that the global default should be required :thinking:

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -90,6 +104,17 @@ class ObjectsAPIConfig(SingletonModel):
         validators=[validate_rsin],
         help_text=_("Default RSIN of organization, which creates the INFORMATIEOBJECT"),
     )
+    content_json = models.TextField(
+        _("Json field"),
+        validators=[
+            DjangoTemplateValidator(
+                backend="openforms.template.openforms_backend",
+            ),
+            JsonTemplateValidator(max_length=300),
```
SilviaAmAm, line 113: This validation on the length does not make too much sense here, since once the `json.loads` happens, there will be all the extra data from the form/vairables that will make it longer. The check should happen after you have rendered the template and before you are about to convert it to JSON. 

I discussed with Sergei that this should then raise a `django.core.exceptions.SuspiciousOperation` if the rendered template larger is than a configurable size. To determine a reasonable default, we could time how long it takes with inputs of increasing size (1mb, ...) and decide what seems appropriate

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -128,15 +133,9 @@ def register_submission(
...
 
-        record_data = {
-            "data": render_json(submission),
-            "type": options["productaanvraag_type"],
-            "submission_id": str(submission.uuid),
-            "language_code": submission.language_code,
-            "attachments": attachments,
-            "pdf_url": document["url"],
-        }
+        templates = get_registration_json_templates(submission)
```
SilviaAmAm, line 136:
```suggestion
        template = get_registration_json_template(submission)
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +160,29 @@ def register_submission(
                 get_drc=get_drc,
                 language=language_code_2b,
             )
-            record_data["csv_url"] = submission_csv_document["url"]
+            csv_url = submission_csv_document["url"]
+
+        base_render_context = {
+            "_submission": submission,
+            **get_variables_for_context(submission),
+            "type": options["productaanvraag_type"],
```
SilviaAmAm, line 168: Maybe better?
```suggestion
            "productaanvraag_type": options["productaanvraag_type"],
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +160,29 @@ def register_submission(
...
+            "submission": {
+                "uuid": submission.uuid,
+                "language_code": submission.language_code,
+                "attachments": attachments,
+                "pdf_url": document["url"],
+                "csv_url": csv_url,
+            },
```
SilviaAmAm, line 175: why this nesting? I think it becomes a bit confusing to have both `submission` and `_submission`. I am not sure if we need the uuid at all, but I think that the attachments, language code and the urls can appear in the top level?

src/openforms/registrations/contrib/objects_api/templates/registrations/contrib/objects_api/content_json.txt
SilviaAmAm, line 1: I guess this is # Todo ? :P

src/openforms/registrations/contrib/objects_api/validators.py
```diff
@@ -0,0 +1,64 @@
...
+
+@deconstructible
+class JsonTemplateValidator:
+    """Validate that the converted django template is legal JSON"""
+
+    def __init__(
+        self,
+        max_length: int | None = None,
+        backend: str = openforms_backend,
+        error_type: JsonTemplateValidatorErrorTypes.choices = JsonTemplateValidatorErrorTypes.model,
```
SilviaAmAm, line 24: The max_length is not needed (see previous comment), and the error type also not.

Also see other comments about the backend

src/openforms/registrations/contrib/objects_api/constants.py
```diff
@@ -0,0 +1,6 @@
+from django.db import models
+
+
+class JsonTemplateValidatorErrorTypes(models.TextChoices):
+    model = "models", "Models"
+    api = "api", "Api"
```
SilviaAmAm, line 6: I think that this also will no longer be needed

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +160,29 @@ def register_submission(
...
+        }
+
+        record_data = {
+            "data": render_from_string(
+                templates,
+                context={**base_render_context},
+                disable_autoescape=True,
+                backend=openforms_backend,
+            ),
+        }
```
SilviaAmAm, line 185: There should be no need to nest the rendered json in the "data" because on line 194 you already have the key data (but I guess this is still work in progress)

src/openforms/registrations/contrib/objects_api/plugin.py
joeribekker: I asked for the nesting, as namespacing, to prevent conflicts but I'll leave this up to you and Sergei to decide.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +160,29 @@ def register_submission(
                 get_drc=get_drc,
                 language=language_code_2b,
             )
-            record_data["csv_url"] = submission_csv_document["url"]
+            csv_url = submission_csv_document["url"]
+
+        base_render_context = {
+            "_submission": submission,
```
SilviaAmAm, line 166: After discussing with DH this morning, it is important that the public reference is included here. For the confirmation email we have in the context:
```
...
"public_reference": submission.public_registration_reference
...
```
so it would be good to add that here too.

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -1,11 +1,19 @@
+from functools import partial
+
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
 from solo.models import SingletonModel
 from zgw_consumers.constants import APITypes
 
+from openforms.registrations.contrib.email.models import render_with_language
```
CharString, line 9: Pulling this out of `contrib.email` feels  off. Is setting the language critical? Also see my remark on the template itself. I'm not sure i18n is needed during render. But I might be wrong.

CharString (changes requested):
@sergei-maertens I feel not well versed into the objecten api to be able to judge the changes to `src/openforms/registrations/contrib/objects_api/tests/test_backend.py` Could you have a quick look?

src/openforms/conf/base.py
```diff
@@ -1193,3 +1193,7 @@
 
 if OPEN_FORMS_EXTENSIONS:
     INSTALLED_APPS += OPEN_FORMS_EXTENSIONS
+
+
+# Objects API registration backedend maxinum template size in bytes
+OBJECTS_API_DATA_SIZE_LIMIT = config("OBJECTS_API_DATA_SIZE_LIMIT", 1000000)
```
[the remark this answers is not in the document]
bart-maykin: The Default is 1mb, maybe I should add a extra comment about it?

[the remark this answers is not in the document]
sergei-maertens: please do, makes it easier to understand the value. You can also use the underscores like in Chris' comment to make the value more readable.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +160,59 @@ def register_submission(
...
+            **get_variables_for_context(submission),
```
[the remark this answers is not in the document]
sergei-maertens: Let's use the `variables` naming here, because that's what they are conceptually *and* technically. So a namespace `variables` (and not `submission.variables`).

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +162,59 @@ def register_submission(
...
+            disable_autoescape=False,
+            backend=openforms_backend,
+        )
+
+        if sys.getsizeof(record_data) > settings.OBJECTS_API_DATA_SIZE_LIMIT:
+            logger.error(
+                "Content JSON field exceeded size of %s with total size: %s",
+                settings.OBJECTS_API_DATA_SIZE_LIMIT,
+                sys.getsizeof(record_data),
+            )
```
sergei-maertens, line 199: I don't think we need the `logger` call, raising the exception is sufficient to make it end up in Sentry.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +162,59 @@ def register_submission(
...
+            raise SuspiciousOperation(
+                _(
+                    "Content JSON field exceeded size of %(max)s with total size: %(total)s.",
+                )
+                % {
+                    "max": settings.OBJECTS_API_DATA_SIZE_LIMIT,
+                    "total": sys.getsizeof(record_data),
+                }
+            )
```
sergei-maertens, line 208: * we don't translate exception messages, as they're not user facing.
* you can also make the file sizes a bit more human-readable using Django's `filesizeformat` filter
* finally, using the walrus operator (`:=`) avoids re-calculating the same thing

```suggestion
from django.template.defaultfilters import filesizeformat
        
        if (data_size := sys.getsizeof(record_data)) > settings.OBJECTS_API_DATA_SIZE_LIMIT:
            formatted_size = filesizeformat(data_size)
            max_size = filesizeformat(settings.OBJECTS_API_DATA_SIZE_LIMIT)
            raise SuspiciousOperation(
                f"Templated out content JSON exceeds the maximum size {max_size} ("
                f"it is {formatted_size})."
            )
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +162,59 @@ def register_submission(
...
+
+        try:
+            record_data = json.loads(record_data)
+        except (TypeError, json.decoder.JSONDecodeError) as err:
```
sergei-maertens, line 212: when does this raise `TypeError`? `record_data` should always be a string because it's coming from the template engine, what am I missing?

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -161,7 +162,59 @@ def register_submission(
...
+            logger.error(
+                "Template object doesn't have valid JSON format",
+                exc_info=err,
+            )
+            raise ValidationError(_("Invalid JSON"), code="invalid") from err
```
sergei-maertens, line 217: I don't think that `ValidationError` is appropriate here, at this point it's not something that can be corrected anymore by anyone, it can only crash. So I'd probably replace this with a `raise RuntimeError("Template evaluation did not result in valid JSON") from err`

src/openforms/registrations/contrib/objects_api/templates/registrations/contrib/objects_api/content_json.txt
```diff
@@ -0,0 +1,16 @@
+{% load registrations.contrib.objects_api.json_summary capture_tags %}
+{% capture as tt_openblock silent %}{% templatetag openblock %}{% endcapture %}
+{% capture as tt_closeblock silent %}{% templatetag closeblock %}{% endcapture %}
+{% capture as tt_openvariable silent %}{% templatetag openvariable %}{% endcapture %}
+{% capture as tt_closevariable silent %}{% templatetag closevariable %}{% endcapture %}
+{% autoescape off %}
```
sergei-maertens, line 6: you can remove this and disable auto-escaping in the `plugin.py` as mentioned before

src/openforms/registrations/contrib/objects_api/templates/registrations/contrib/objects_api/content_json.txt
```diff
@@ -0,0 +1,16 @@
...
+{
+  "data": {{ tt_openblock }} json_summary {{ tt_closeblock }},
+  "type": "terugbelnotitie",
```
sergei-maertens, line 9: this should definitely be configurable too if we support templating here now

1. Add django-solo model default
2. Make it an optional field in the options serializer too

src/openforms/registrations/contrib/objects_api/templates/registrations/contrib/objects_api/content_json.txt
```diff
@@ -0,0 +1,16 @@
...
+  "bsn": "{{ tt_openvariable }} auth_bsn {{ tt_closevariable }}",
+  "pdf_url": "{{ tt_openvariable }} submission.pdf_url {{ tt_closevariable }}",
+  "attachments": {{ tt_openvariable }} submission.attachments {{ tt_closevariable }},
```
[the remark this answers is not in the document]
sergei-maertens: Is that so? As far as I can see `attachments` is a list of strings, of which the `str(attachments)` representation happens to be valid JSON. If it's outputting the right thing, then I chalk it up to pure coincidence :thinking:

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_summary.py
```diff
@@ -0,0 +1,18 @@
...
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def json_summary(context):
+    submission = context.get("_submission")
+    if not submission:
+        return {}
+
+    get_json_renderer = render_json(submission)
```
sergei-maertens, line 17: please avoid variable names that look like function names

```suggestion
    json_data = render_json(submission)
```

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -1849,3 +2162,142 @@ def match_attachment(request):
             document_create_attachment_body["vertrouwelijkheidaanduiding"], "geheim"
         )
         self.assertEqual(document_create_attachment_body["titel"], "A Custom Title")
+
+    @override_settings(OBJECTS_API_DATA_SIZE_LIMIT=1000000)
+    def test_submission_with_objects_api_content_json_exceed_max_file_limit(self, m):
+        # generates template of 1000007 bytes
+        self.config.content_json = {"oversized_list": list(range(0, 138875))}
+        self.config.save()
+        self.config.refresh_from_db()
```
sergei-maertens, line 2171: this is making your data big and therefore the test slow

It's far easier to just use a stupid small file size limit: `@override_settings(OBJECTS_API_DATA_SIZE_LIMIT=10)` -> 10 bytes, that'll trip everything.

You also really want to avoid making actually database/cache calls to singleton models, for those `unittest.mock.patch` is appropriate. There are plenty of other tests where this happens, so please apply the same pattern.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -1849,3 +2162,142 @@ def match_attachment(request):
...
+            str(cm.exception),
+        )
+
+    def test_submission_with_objects_api_content_json_not_valid_json(self, m):
+        # trailing comma on last item
+        self.config.content_json = {
+            "key": "value",
+        }
+        self.config.save()
+        self.config.refresh_from_db()
```
sergei-maertens, line 2243: see config mocking remark above

src/openforms/registrations/service.py
```diff
@@ -49,7 +49,6 @@ def extract_submission_reference(submission: Submission) -> str:
 
 
 def get_registration_plugin(submission: Submission) -> BasePlugin | None:
-
```
sergei-maertens, line 52: I don't see why this change was necessary - please refrain from changing files that are not related to your ticket, as they needlessly clutter the diff for reviewers.

sergei-maertens (changes requested):

sergei-maertens:
I haven't properly looked at the test code yet, but let's clean up the implementation first.

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+        submission_step = submission.steps[0]
+        step_slug = submission_step.form_step.form_definition.slug
+
+        # Loads template tag and uses template tag
+        template = Template(
+            """{% load registrations.contrib.objects_api.json_summary %}{% json_summary %}"""
+        )
+
+        rendered = template.render(Context({"_submission": submission}))
```
sergei-maertens, line 50: We have `openforms.template.render_from_string` for rendering templates from strings, see the documentation: https://open-forms.readthedocs.io/en/latest/developers/backend/core/templating.html

Using `django.template.Template` directly is a backwards compatibility thing from Django, but it can't handle (possible future) other template engines. It's best avoided.

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -0,0 +1,68 @@
...
+
+        expected = (
+            '{"%s": {"voornaam": "Foo", "achternaam": "Bar", "tussenvoegsel": "de"}}'
+            % step_slug
```
sergei-maertens, line 54: set up your test data with the factories accordingly with a fixed slug, and then you can make "dumb" assertions without any dynamic stuff, which ultimately makes tests easier to understand/read/follow.

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_render_json_summary_no_submission(self):
+        # Loads template tag and uses template tag
+        template = Template(
+            """{% load registrations.contrib.objects_api.json_summary %}{% json_summary %}"""
+        )
+
+        rendered = template.render(Context())
+        expected = "{}"
+
+        self.assertEqual(rendered, expected)
```
sergei-maertens, line 68:
```suggestion
        rendered = render_from_string("{% json_summary %}")

        self.assertEqual(rendered, "{}")
```

The loading of template tag libraries is not something that we want to burden our form designers with, that's why they are implicitly loaded in the template engine by us for them.

Next you want to make sure that these templates can be loaded with the template backend that will actually be used, and you can still use that to test the behaviour.

With simple assertions, please just use the literals directly which makes the test shorter and easier to follow.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -37,6 +38,29 @@ def setUpTestData(cls):
...
+                    }\n
+                ],\n
+                "pdf": "{{ submission.pdf_url }}",\n
+                "csv": "{{ submission.csv_url }}",\n
+                "bijlagen": [\n
+                    {% for attachment in submission.attachments %}\n
+                    "{{ attachment }}"{% if not forloop.last %},{% endif %}\n
+                    {% endfor %}\n
+                ]\n
+            }""",
```
sergei-maertens, line 63: You can use `textwrap.dedent` to handle the indentation of the multiline string. Additionally, I don't think you need the `\n` in there, they should be included when using `"""`

sergei-maertens: If we're providing a template here, how are we testing that our default shipped template works?

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -564,19 +600,261 @@ def test_submission_with_objects_api_backend_override_defaults_do_not_upload_csv
...
+                "taal": "{{ submission.language_code  }}"\n
+            }""",
+            zaak_vertrouwelijkheidaanduiding="geheim",
+            doc_vertrouwelijkheidaanduiding="geheim",
+        )
+
+        expected_result = {
+            "url": "https://objecten.nl/api/v1/objects/1",
+            "uuid": "095be615-a8ad-4c33-8e9c-c7612fbf6c9f",
+            "type": objects_form_options["objecttype"],
```
sergei-maertens, line 710: please avoid references/cleverness in tests and include the literal value if you know it

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -564,19 +600,261 @@ def test_submission_with_objects_api_backend_override_defaults_do_not_upload_csv
...
+            "record": {
+                "index": 0,
+                "typeVersion": objects_form_options["objecttype_version"],
                 "data": {
                     "data": {
                         f"{step_slug}": {
```
sergei-maertens, line 716: `f"{variable}"` is the same as `variable`

see also the earlier remark about setting up your test data with a known slug so you can avoid using variables in your assertions

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -0,0 +1,78 @@
...
+        rendered = render_from_string(
+            "{% json_summary %}",
+            context={},
+            backend=openforms_backend,
+            disable_autoescape=True,
+        )
+
+        expected = "{}"
+
+        self.assertEqual(rendered, expected)
```
sergei-maertens, line 78:
```suggestion
        self.assertEqual(rendered, "{}")
```
simple is better :)

src/openforms/submissions/tests/factories.py
```diff
@@ -122,6 +122,7 @@ def from_components(
         cls,
         components_list: List[dict],
         submitted_data: dict = None,
+        form_definition_slug: str = "",
```
sergei-maertens, line 125: how about `form_definition_kwargs = None`, and then in the body:

```py
form_definition_kwargs = form_definition_kwargs or {}
form_definition_kwargs.setdefault("name", ...)
form_definition_kwargs.setdefault("configuration", ...)
```

and now you can pass generic factory kwargs while still being functional without any kwargs

docs/manual/templates.rst
```diff
@@ -424,3 +424,89 @@ Voorbeeld
...
+=====================================  ===========================================================================
+Variabele                              Beschrijving
+=====================================  ===========================================================================
+``{{ productaanvraag_type }}``         Het productaanvraag type.
+``{{ submission.public_reference }}``  De publieke referentie van de inzending.
+``{{ submission.kenmerk }}``           De inzending UUID.
+``{{ submission.language_code }}``     De taal van het formulier die werd ingezonden, bijvoorbeeld 'nl'.
+``{{ submission.pdf_url }}``           De URL van het inzending rapport (in PDF formaat).
+``{{ submission.csv_url }}``           De URL van het inzending rapport (in CSV formaat).
+``{% json_summary %}``                 JSON met ``"eigenschapsnaam": "ingevulde waarde"`` van alle formulier velden.
```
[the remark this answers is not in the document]
SilviaAmAm: This only contains form field variables.

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -8,6 +8,16 @@
...
 
 
+@register.simple_tag(takes_context=True)
+def uploaded_attachment_urls(context):
+    """
+    Output a sequence of attachment URLs as a JSON-serialized list.
+    """
+    # attachments is a list of URLs
+    attachments = context.get("submission", {}).get("uploaded_attachment_urls", [])
+    return SafeString(SafeString(json.dumps(attachments)))
```
sergei-maertens, line 18: accidental double safestring?

sergei-maertens (approved):

PR #3165, opened 2023-06-16
Fix styling Django pages after NLDS migration

SilviaAmAm:
Screenshot from 2023-06-16 15-36-06
Screenshot from 2023-06-16 15-36-24

### DIFF
--- a/src/openforms/scss/components/_form-choices.scss
+++ b/src/openforms/scss/components/_form-choices.scss
@@ -1,5 +1,5 @@
 @import 'microscope-sass/lib/bem';
-@import '~microscope-sass/lib/typography';
+@import 'microscope-sass/lib/typography';
 
 .openforms-form-choices {
   @include modifier('inline') {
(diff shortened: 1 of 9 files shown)


src/openforms/scss/components/_form-choices.scss
```diff
@@ -11,6 +11,7 @@
     }
 
     @include element('choice') {
+      @include margin($margin: true, $value-mobile: $typography-margin-list);
```
SilviaAmAm, line 14: Added as otherwise the last child didn't have a margin.

src/openforms/scss/components/_form-choices.scss
```diff
@@ -1,5 +1,5 @@
 @import 'microscope-sass/lib/bem';
-@import 'microscope-sass/lib/responsive';
+@import '~microscope-sass/lib/typography';
```
sergei-maertens, line 2: the ~ construct is a legacy thing and is no longer needed

```suggestion
@import 'microscope-sass/lib/typography';
```

src/openforms/scss/components/_form-choices.scss
sergei-maertens: yeah we'll have to refactor this CSS and markup too at some point :thinking:

sergei-maertens (approved):

PR #3167, opened 2023-06-19
Prepare release v2.2.0

sergei-maertens:


CHANGELOG.rst
```diff
@@ -2,6 +2,222 @@
...
 
+
+2.2.0 "TBD" (2023-06-??)
+========================
+
+.. epigraph::
+
+   **TDB** *TODO*
+
+   -- TODO
```
sergei-maertens, line 13: TODO: complete these

src/openforms/js/lang/nl.json
```diff
@@ -65,7 +65,7 @@
     "originalDefault": "Incomplete Submissions Removal Method"
   },
   "20K9To": {
-    "defaultMessage": "Service fetch configuration",
+    "defaultMessage": "Servicebevraging instellen",
```
jiromaykin, line 68: Isn't 'fetch' an acceptable Dutch term instead of bevraging?

src/openforms/js/lang/nl.json
```diff
@@ -755,7 +770,7 @@
     "originalDefault": "Component"
   },
   "MXKqVO": {
-    "defaultMessage": "Below you can specify a custom template for the configuration email. If not specified, the globally configured template will be used.",
+    "defaultMessage": "Je kan hieronder een formulierspecifiek sjabloon opstellen voor de bevestigingsmail. Indien leeg, dan worden de algemene instellingen gebruikt.",
```
jiromaykin, line 773: "Je kan" is informal - if all language in the project is formal then use "Je kunt".

src/openforms/js/lang/nl.json
```diff
@@ -1295,7 +1320,7 @@
     "originalDefault": "Object"
   },
   "fpX4ei": {
-    "defaultMessage": "Mapping expression language",
+    "defaultMessage": "Soort mappingexpressie",
```
jiromaykin, line 1323: Not sure but use "taal" instead of "soort"?

src/openforms/js/lang/en.json
```diff
@@ -464,6 +464,11 @@
     "description": "Manage process vars component column title",
     "originalDefault": "Field"
   },
+  "ENyg8G": {
+    "defaultMessage": "There are multiple co-sign components specified in this form. At most one should be present.",
+    "description": "Too many cosign components",
```
jiromaykin, line 469: Inconsistent spelling of 'cosign' and 'co-sign' in project :-) if this isn't lingo, then make consistent.

jiromaykin (changes requested):
I've added some considerations as part of the Translations review. Feel free to implement or mark as resolved so I can approve.

src/openforms/js/lang/nl.json
sergei-maertens: I stuck to this term as that's what we've communicated so far about this feature.

src/openforms/js/lang/nl.json
sergei-maertens: I've updated it to not address the reader at all, but rather give a description which is more consistent with other texts (no precedent for "U kunt", "U kan", "Je kan" or "Je kunt" at all).

src/openforms/js/lang/nl.json
sergei-maertens: I don't think we want users to really think about "programming languages" or get confused with language as in English vs. Dutch - ultimately the feature is about configuring an expression to extract data, which happens to be implemented in a "language". The important information however is designating what kind of expression it is (the original source string is not ideal either in that regard).

src/openforms/js/lang/en.json
sergei-maertens: yeah, luckily this is only in the message description which is internal (and based on that, the message ID is generated). I've checked an all public-facing terms seem to be spelled `co-sign`.

Both spellings appear to be correct though, making it even more confusing :sweat_smile:

jiromaykin (approved):
Approval for translations only :-)

Issue #3139, opened 2023-06-06
Can't start form in maintenance mode as form designer/admin


LaurensBurger:
### Product versie / Product version

2.1.3 - latest
reported by: Nopolder, Vught, SED, Horst

### Omschrijf het probleem / Describe the bug

Being logged in as a admin/form designer does not allow me to start a form that is set to maintenance mode.
![image](https://github.com/open-formulieren/open-forms/assets/112621537/10dbeb96-9317-41dc-b7db-4498ad2669eb)
![image](https://github.com/open-formulieren/open-forms/assets/112621537/1bec89c2-a627-4892-94d0-040935b6841b)

This should be possible and is causing interruptions in the "flow" of form designing

sergei-maertens:
Can confirm, I ran into this and recently and thought I was doing something wrong myself

PR #3199, opened 2023-06-29
:label: Refactor formio types in backend

sergei-maertens:
The Form.io base component type is now more realistic with the totality set to False, meaning that any key in the dict may be absent. This matches Form.io's own typescript definitions.

Additionally, the types are centralized in a typing package where (relative) imports don't cause circular import issues - they are now truly standalone. The runtime code that had type definitions now imports them from the top-level entry point.

This change allows us to more strictly type individual component types, while having the type checker help accounting for possible Form.io JSON weirdness.

Needed for the #3139 PR which makes use of these types.

### DIFF
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -9,15 +9,11 @@ from openforms.submissions.models import Submission
 from openforms.typing import DataMapping
 from openforms.utils.date import format_date_value
 
-from ..dynamic_config.date import (
-    FormioDateComponent,
-    FormioDatetimeComponent,
-    mutate as mutate_min_max_validation,
-)
+from ..dynamic_config.date import mutate as mutate_min_max_validation
 from ..formatters.custom import DateFormatter, DateTimeFormatter, MapFormatter
 from ..formatters.formio import DefaultFormatter, TextFieldFormatter
 from ..registry import BasePlugin, register
-from ..typing import Component
+from ..typing import Component, DateComponent, DatetimeComponent
 from ..utils import conform_to_mask
 from .np_family_members.constants import FamilyMembersDataAPIChoices
 from .np_family_members.haal_centraal import get_np_children_haal_centraal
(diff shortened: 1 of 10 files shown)


CharString (approved):
The impression I get from FormIO structures is that they always have all keys  present, but they may be `None`, which is slightly different from how I understand `total=False`: keys not present at all.

Also, it's not very ergonomic. As it still forces you to validate upon every use of a value, whether it's present. Resulting in shot gun validation; type checks all over the code.
I'd prefer dealing with quirks at the edges of the system, and have narrower types flow through. For instance: parsing in the `Serializer` to `NewType`s would have caught my misuse of `var.name` where `var.key` was meant.

But I guess dealing with `Optional[str]`, `Optional[int]` etc is a step up from dealing with `Optional[Any]`.

sergei-maertens:
> The impression I get from FormIO structures is that they always have all keys present

this is most definitely not the case - as absent/optional keys in typescript resolve to `undefined`. You can definitely get formio working with partial schema's. The fact that "full" schema's are exchanged over the API is an implementation details of the formio-builder, which records the full schema from parent/defaults behind the scenes.

If you play around on https://formio.github.io/formio.js/ and only enter the minimal schema in the JSON, then things appears to still be fully functional.

> Also, it's not very ergonomic. As it still forces you to validate upon every use of a value, whether it's present. Resulting in shot gun validation; type checks all over the code.

IMO the silver bullet is a pragmatic approach with Python 3.11's `NotRequired` for TypedDict, as the current (`master`) situation results in `type: ignore` being littered all over the place, which unfortunately also surpresses valid type checker errors (e.g. assuming something is `int` while in reality it is a `str`).

Because of the `total=True` default, we can't add additional/situational keys to the base component type because they're not always present, but we do make use of those keys in certain situations.

> I'd prefer dealing with quirks at the edges of the system, and have narrower types flow through. For instance: parsing in the Serializer to NewTypes would have caught my misuse of var.name where var.key was meant.

I just don't see how we make that work with this code base where the interface is "return list of formio components" and then explicitly defining the components that should satisfy this type hint, with specific component types. At least, not without slapping `type: ignore` on my constant definition, and at that point using type hints serves no benefit at all anymore :(

PR #3200, opened 2023-06-29
[#3182] Import forms without cosign template tag

### LINKED ISSUE #3182: importing form from pre 2.2.0 impossible due to missing "{% cosign_information %} "

2.2.0

importing a form that has a template cannot be done in 2.2.0 because these never have "{% cosign_information %}". 
image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #3182 

Needs backport to 2.2

### DIFF
--- a/src/openforms/emails/api/serializers.py
+++ b/src/openforms/emails/api/serializers.py
@@ -1,6 +1,9 @@
+import re
+
 from rest_framework import serializers
 
 from openforms.translations.api.serializers import ModelTranslationsSerializer
+from openforms.utils.migrations_utils.regex import add_cosign_info_templatetag
 
 from ..models import ConfirmationEmailTemplate
 
@@ -15,3 +18,23 @@ class ConfirmationEmailTemplateSerializer(serializers.ModelSerializer):
             "content",
             "translations",
         )
+
+    def to_internal_value(self, data):
+        if content := data.get("content"):
+            data["content"] = self._add_cosign_templatetag(content)
+
+        return super().to_internal_value(data)
+
+    def _add_cosign_templatetag(self, content):
+        is_import = self.context.get("is_import", False)
+
+        if not is_import:
+            return content
+
+        pattern = re.compile(r"\{%\s?cosign_information\s?%\}")
+        match = pattern.search(content)
+
+        if not match:
+            content = add_cosign_info_templatetag(content)
+
+        return content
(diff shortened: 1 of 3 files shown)


src/openforms/emails/api/serializers.py
```diff
@@ -15,3 +18,23 @@ class Meta:
             "content",
             "translations",
         )
+
+    def to_internal_value(self, data):
+        if content := data.get("content"):
+            data["content"] = self._add_cosign_templatetag(content)
```
sergei-maertens, line 24: this doesn't seem to handle the `translations` nested structure, right?

src/openforms/emails/api/serializers.py
```diff
@@ -15,3 +18,23 @@ class Meta:
...
+
+        return super().to_internal_value(data)
+
+    def _add_cosign_templatetag(self, content):
+        is_import = self.context.get("is_import", False)
+
+        if not is_import:
+            return content
+
+        pattern = re.compile(r"\{%\s?cosign_information\s?%\}")
```
sergei-maertens, line 34: you typically want to put that at the module level so that the regex is compiled once and then the cache is used, now it's being compiled every time this method runs so you might as well just use `re.search(r"\{%\s?cosign_information\s?%\}", content)`

src/openforms/forms/tests/test_import_export.py
```diff
@@ -768,3 +768,66 @@ def test_export_with_filetype_information(self, m_get_solo):
...
+                    "internal_name": "Test Form",
+                    "product": None,
+                    "show_progress_indicator": True,
+                    "slug": "test-form",
+                    "url": "http://testserver/api/v2/forms/324cadce-a627-4e3f-b117-37ca232f16b2",
+                    "uuid": "324cadce-a627-4e3f-b117-37ca232f16b2",
+                    "confirmation_email_template": {
+                        "content": "Some content {% appointment_information %}\n{% payment_information %}",
+                        "subject": "A subject",
+                    },
```
sergei-maertens, line 792: the test should include a realistic `translations` structure here and make assertions on that

sergei-maertens (changes requested):

sergei-maertens:
Backported with ef21f66620dc84f14c03a6b6af2d2f73264d6857

PR #3204, opened 2023-06-30
[#1882] Configure min/max on currency component

### LINKED ISSUE #1882: Currency fields should allow setting a max- or minimum value
### Thema / Theme

Form designer

When setting up a currency field, the form-editor as of yet doesn't allow for limits to be defined.

When applying for a government grant, it would be useful to make it clear that there is a maximum (or minimum) amount to be requested.

This is the way it is set up in a regular 'number' field, which seems appropriate for currency fields as well:

image

Als formulierbouwer
Wil ik in een valuta-veld ook max en min waarden kunnen definiëren, net zoals dat nu bij een getal veld kan.
Zodat het voor een eindgebruiker bij het invullen van een subsidie aanvraag duidelijk te maken is dat er een maximaal (of minimaal) bedrag aan te vragen is.

Related: Dimpact PF-75

SilviaAmAm:
Fixes #1882

### DIFF
--- a/src/openforms/formio/typing/base.py
+++ b/src/openforms/formio/typing/base.py
@@ -17,6 +17,8 @@ from .dates import DateConstraintConfiguration
 class Validate(TypedDict, total=False):
     required: bool
     maxLength: int
+    min: int | float
+    max: int | float
 
 
 class OpenFormsConfig(TypedDict, total=False):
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):
Just for completeness sake - we are aware that using floats for currency is... not recommended, but at the moment we are limited to Form.io.

PR #3212, opened 2023-07-04
Add file upload name template

### LINKED ISSUE #1890: Allow for pre- and suffixes in renaming uploads
### Thema / Theme

Form designer

Currently, it is possible to rename uploads to a pre-defined value or keep the original name. To pre- or suffix the original filenames has its benefits.

At times, a person who submitted a form might refer to a certain file by the name they uploaded it with (DCIM04564.jpg, Scan081.pdf) in subsequent contact - and at the same time, the municipality would like a neatly ordered set of uploads - it would be great if the filename of uploads could be set to get a pre- or suffix as well (voorkant_DCIM04564.jpg, toestemmingsverklaring_Scan081.pdf).

_No response_

Related: Dimpact PF-69

*Template/Suffix/Prefix moet optioneel zijn*

CharString:
Closes #1890


src/openforms/submissions/tests/test_submission_attachment.py
```diff
@@ -1617,3 +1617,48 @@ def test_combination_attachment_and_repeating_group_with_numbers(self):
...
+
+        submission = SubmissionFactory.from_components(components, data)
+
+        result = attach_uploads_to_submission_step(submission.steps[0])
+        self.assertEqual(len(result), 1)
+
+        attachment_basename = Path(result[0][0].file_name).stem
+        upload_basename = Path(upload.file_name).stem
+
+        self.assertEqual(attachment_basename, f"prefix_{upload_basename}_postfix")
```
sergei-maertens, line 1664: Can you please instead use literals in the test and factory call so that we can make plain assertions instead of evaluating python code in our head?

I don't really want to have to reason about the datastructure returned by `attach_uploads_to_submission_step` in a test that deals with the filename templating :grimacing:

sergei-maertens (approved):

PR #3217, opened 2023-07-04
:technologist: Mitigate flakiness due to IO side-effects

sergei-maertens:


CharString (approved):
Nice hack :black_heart:

PR #3224, opened 2023-07-05
[#3130] Fix copying form definitions with long names

### LINKED ISSUE #3130: Form definitions with long names cause 500 when copied

2.1.0 / 2.2.0

When a form definition has  a long name, the new copy of this form definition will have a even longer name because the new copy will include 
(copy)
Original name: **This is a pretty looong title for a formdefinition**
New name: **This is a pretty looong title for a formdefinition (copy)**

Which results in a 500 for the user: value too long for type character varying(50)

Create a form definition with the title:
This is a pretty looong title for a formdefinition
Try to copy this form definition:
image

vaszig:
Fixes #3130

### DIFF
--- a/src/openforms/forms/models/form_definition.py
+++ b/src/openforms/forms/models/form_definition.py
@@ -15,6 +15,7 @@ from django.utils.translation import gettext_lazy as _
 from autoslug import AutoSlugField
 
 from openforms.formio.utils import iter_components
+from openforms.utils.helpers import truncate_str_if_needed
 
 from ..models import Form
 from ..tasks import detect_formiojs_configuration_snake_case
@@ -102,6 +103,9 @@ class FormDefinition(models.Model):
             callback = partial(detect_formiojs_configuration_snake_case.delay, self.id)
         transaction.on_commit(callback)
 
+    def get_charfield_max_length(self, field_name: str) -> int:
+        return self._meta.get_field(field_name).max_length
+
     @transaction.atomic
     def copy(self):
         copy = deepcopy(self)
@@ -114,6 +118,17 @@ class FormDefinition(models.Model):
             else ""
         )
         copy.slug = _("{slug}-copy").format(slug=self.slug)
+
+        # truncate name and internal name if needed
+        copy.name = truncate_str_if_needed(
+            self.name, copy.name, self.get_charfield_max_length("name")
+        )
+        copy.internal_name = truncate_str_if_needed(
+            self.internal_name,
+            copy.internal_name,
+            self.get_charfield_max_length("internal_name"),
+        )
+
         copy.save()
         return copy
 
--- a/src/openforms/forms/models/form_definition.py
+++ b/src/openforms/forms/models/form_definition.py
@@ -15,7 +15,7 @@ from django.utils.translation import gettext_lazy as _
 from autoslug import AutoSlugField
 
 from openforms.formio.utils import iter_components
-from openforms.utils.helpers import truncate_str_if_needed
+from openforms.utils.helpers import get_charfield_max_length, truncate_str_if_needed
 
 from ..models import Form
 from ..tasks import detect_formiojs_configuration_snake_case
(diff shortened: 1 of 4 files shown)


src/openforms/forms/models/form_definition.py
```diff
@@ -114,6 +118,17 @@ def copy(self):
...
+
+        # truncate name and internal name if needed
+        copy.name = truncate_str_if_needed(
+            self.name, copy.name, self.get_charfield_max_length("name")
+        )
+        copy.internal_name = truncate_str_if_needed(
+            self.internal_name,
+            copy.internal_name,
+            self.get_charfield_max_length("internal_name"),
+        )
```
SilviaAmAm, line 130: I am thinking: the slug also has a max length of 100 char, and since it is derived (usually) from the name, it also has a chance of being too long :thinking:

src/openforms/utils/helpers.py
```diff
@@ -0,0 +1,12 @@
...
+    Given the original string and the new string, return either the original one if
+    it doesn't exceed the given max length, or the new one truncated and by adding
+    the ellipsis to the original one.
+    """
+    if len(new_str) <= max_length:
+        return new_str
+
+    new_str_length = len(new_str) - len(original_str)
+    str_max_length = max_length - new_str_length
+    return original_str[: str_max_length - 1] + "\u2026" + new_str[-new_str_length:]
```
SilviaAmAm, line 12: I wonder if it doesn't make more sense to have a method that takes in the original string and the max length, then does something along the lines of:
```python
import textwrap

def copy_name(original_name, max_length):
     length_without_copy = max_length - 7
     return textwrap.shorten(original_name, width=length_without_copy, placeholder="\u2026") + " (copy)"
```
And this is used directly in src/openforms/forms/models/form_definition.py on lines 114/115?

(note that `shorten` breaks in between the words rather than in the words, but I don't think this is a problem for this case? :thinking: )

src/openforms/utils/tests/test_helpers.py
SilviaAmAm, line 1: I think there should also be a test that mimics importing a form where the copied name/internal name/slug exceed the limit number of characters. You can look here for inspiration https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/tests/test_import_export.py#L850 :slightly_smiling_face:

SilviaAmAm (changes requested):

src/openforms/forms/models/form_definition.py
vaszig: Yes, it was my first thought but then didn't touch it as it can be too long only if the user explicitly defines something like 90 chars as a slug. But yeah, just to be safe, I will add the slug as well.

src/openforms/utils/helpers.py
vaszig: As Sergei mentioned as well, we should avoid explicit numbers (the actual length of the additional str) that's why I did it this way. Additionally the length will be `7` ( " (copy)") in english but `8` in dutch (" (kopie)"), and this is the reason I did it more generic, like a function that could be used by anything. What is your opinion?

SilviaAmAm: Ahh makes sense!

src/openforms/utils/tests/test_helpers.py
SilviaAmAm: As discussed on slack: I meant copy instead of import!

src/openforms/forms/tests/test_formdefinition_admin.py
```diff
@@ -26,3 +30,33 @@ def test_formdefinition_list_for_user_with_all_perms(self):
...
+        self.assertEqual(FormDefinition.objects.count(), 1)
+
+        url = reverse("admin:forms_formdefinition_changelist")
+        response = self.app.get(url, user=user)
+        form = response.forms["changelist-form"]
+        form["action"] = "make_copies"
+        form["_selected_action"].checked = True
+        response = form.submit("submit", value="0")
+
+        form_definition_copy = FormDefinition.objects.last()
```
SilviaAmAm, line 51: I think this could cause flaky tests when the database doesn't return the items in the same order. Usually we fix it by adding `.order_by("pk")` and then taking the last item.

SilviaAmAm (approved):

sergei-maertens (approved):

sergei-maertens (approved):

PR #3229, opened 2023-07-07
⬆️  [#3049] -- upgrade to django-hijack 3.4.1

ErhanCitil:
Ref #3049


sergei-maertens (approved):
Checked the changelog and I didn't spot anything that would cause problems.

Issue #3161, opened 2023-06-15
It's not possible to revert "Data removal" settings in a form


LaurensBurger:
### Product versie / Product version

2.1.x

### Omschrijf het probleem / Describe the bug

lv 85
When you set custom data removal in a form, it is impossible to remove them and in turn  making the form default back to the data removal settings in the general config.
The fields in this tab can not be saved when Null.
![image](https://github.com/open-formulieren/open-forms/assets/112621537/48166d13-e01b-40c1-8eae-fe81107d3223)

### Screen resolution

None

### Device

None

### OS

None

joeribekker:
Refinement: Annoying for editor, reported by real life use. Seems like an easy to tackle fix.

PR #3234, opened 2023-07-07
[#3161] Revert data removal settings in a form

vaszig:


SilviaAmAm (commented):
This fix works, I hesitate on whether the front end should have been fixed to send the right type of data rather than the backend be more lenient :thinking:

vaszig:
> This fix works, I hesitate on whether the front end should have been fixed to send the right type of data rather than the backend be more lenient thinking

Yes, I didn't think about it. I have a question though, since we have an API shouldn't we validate the values at this point as well?

sergei-maertens (changes requested):
Let's apply this fix in the frontend indeed, the API definition is correct:

* it accepts integers or `null`
* DRF already handles string -> number conversion

So, if the value is empty string in the frontend, it should actually send `null` to the backend instead. You can make this patch here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/data/complete-form.js#L59

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -62,6 +62,26 @@ const saveForm = async (state, csrftoken) => {
       // FIXME - should not be required in backend for form designer
       name: draft.form?.translations?.[DEFAULT_LANGUAGE]?.name,
       authenticationBackends: draft.selectedAuthPlugins,
+
+      // make sure empty str for PositiveIntegerField is converted to null.
```
sergei-maertens, line 66: maybe create a helper function in this file for this?

```js
const normalizeLimit = (draft, field) => {
    const removalOptions = draft.form?.submissionsRemovalOptions;
    if (!removalOptions) return;
    const currentValue = removalOptions?.[field];
    if (currentValue === '') {
      removalOptions[field] = null;
    }
};
```

and then call it using:

```js
normalizeLimit(draft, 'successfulSubmissionsRemovalLimit');
// ...
```

sergei-maertens (approved):

PR #3235, opened 2023-07-07
⬆️  [#3049] -- updated to django-tinymce 3.6.1

ErhanCitil:
Ref #3049

### DIFF
--- a/src/openforms/js/initTinymce.js
+++ b/src/openforms/js/initTinymce.js
@@ -43,14 +43,16 @@
         },
       };
 
-      const id = el.id;
-      if ('elements' in mce_conf && mce_conf['mode'] == 'exact') {
-        mce_conf['elements'] = id;
+      // replace default prefix of 'empty-form' if used in selector
+      if (mce_conf.selector && mce_conf.selector.includes('__prefix__')) {
+        mce_conf.selector = `#${el.id}`;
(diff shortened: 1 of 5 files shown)


sergei-maertens (changes requested):
@ErhanCitil can you incorporate the init file changes that Silvia mentions above? You can test this by setting your browser in dark mode and then go to `Admin > Configuratie > Algemene configuratie`

The WYSIWYG widget in the screenshot is tinymce:

image

src/openforms/js/initTinymce.js
```diff
@@ -34,12 +34,12 @@
 
       // TODO Django 4.2: use explicit theme names rather than the media query approach:
       // https://github.com/django/django/blob/main/django/contrib/admin/static/admin/css/dark_mode.css#L36
-      const useDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches;
+      let dark = 'dark';
```
sergei-maertens, line 37: we can only make this change when open-forms itself updates to Django 4.2! For now, we're still on 3.2

sergei-maertens: @ErhanCitil can you please revert this change?

sergei-maertens (approved):
Applied the necessary changes and tested, seems to work properly.

PR #3236, opened 2023-07-07
⬆️  [#3049] -- upgrade to elastic-apm 6.17.0

ErhanCitil:
Ref #3049

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -230,7 +230,7 @@ drf-polymorphic==1.0.0
     # via -r requirements/base.in
 drf-spectacular==0.25.1
     # via -r requirements/base.in
-elastic-apm==6.7.2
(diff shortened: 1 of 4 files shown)


SilviaAmAm (commented):
I don't know from this changelog if it will break anything that we use: https://github.com/elastic/apm-agent-python/blob/main/CHANGELOG.asciidoc ... @CharString ? :stuck_out_tongue:

CharString (approved):
Seems fine at first glance. And since it's not as critical as Sentry, I'm okay with this update. We just need to be careful that metrics from pre and post update may be due to changes in instrumentation. But AFAIK we're not inspecting anything in particular atm.

sergei-maertens:
Changelog doesn't look like it'd cause problems for us

PR #3237, opened 2023-07-07
⬆️  [#3049] -- upgrade to sentry-sdk[django] 1.27.1

ErhanCitil:
Ref #3049


CharString (commented):
If I understand the changelog correctly, there have been some changes to Celery beat monitoring.

@sergei-maertens it's unclear to me if this has any repercussions; the one test I see looks very implementation specific.

sergei-maertens:
Celery beat monitoring is new and opt-in, it looks like. We are currently not doing any beat/cron monitoring, but it's nice to have the option to enable that at some point!

PR #3241, opened 2023-07-10
[#3215] Prefill with bewindvoering/machtigen 

### LINKED ISSUE #3215: Prefill data when using eHerkenning Bewindvoering
### Thema / Theme

Plugins

When someone logs in with eHerkenning bewindvoering, then we receive not only the `KvK` of the company, but also the BSN of the aanvrager and the BSN of the gemachtigde.

DH would like to be able to prefill data based on the BSN of the aanvrager/gemachtigde.

This has high prio for DH.

### Backend

- [x] Modify `prefill_variables` (src/openforms/prefill/__init__.py#132). Currently, this function gives to `_fetch_prefill_values` only the prefill plugin identifier and the attribute to retrieve in this shape:

```json
{
	"haalcentraal": ["name", "surname"],
	"kvkapi": ["postcode", "company name"]
}
```

- [x] Update the digid_eherkenning_oidc implementation so that when they add data to the `machtigen` key of the `AuthInfo`, the keys are predictable. (authoriser/authorisee => according to google translate 'endorsee or authorised person' :stuck_out_tongue: ). Currently, the keys correspond to what is configured in the admin for the machtigen/bewindvoering OIDC plugins.

We need to change this so that it contains the name of the attribute to use:

```json
{
	"haalcentraal": {
		"authoriser": ["name", "surname"],
		"authorisee": ["name", "surname"]
	},
	"kvkapi": {
		"main": ["postcode", "company name"]
	}
}
```

- [x] Modify the `get_prefill_values` of the various prefill plugins so that they can deal with this structure and retrieve the values for the correct identifier. Prefill plugins using the BSN need to be able to retrieve values for 2x BSNs while there should always be at most 1 kvk.

### Form Builder

- [x] Update the prefill tab of the components so that the form designer can configure which identifier to use for retrieving prefill data. This should be a required select field with default value 'main'. The other values should be 'authoriser' (vertegenwordigde/aanvrager) and (something along the lines of) 'autorisee' (gemachtigde, the person actually filling the form in).

- [x] Add the prefill tab to the formio-builder repo?

SilviaAmAm:
Fixes #3215 

Related PR in the new builder: https://github.com/open-formulieren/formio-builder/pull/26


src/openforms/js/components/admin/form_design/variables/VariablesTable.js
```diff
@@ -111,13 +112,6 @@ const EditableVariableRow = ({index, variable, onDelete, onChange}) => {
   const prefillPluginChoices = availablePrefillPlugins.map(plugin => [plugin.id, plugin.label]);
   const [prefillAttributeChoices, setPrefillAttributeChoices] = useState([]);
 
-  const formSteps = formContext.formSteps;
-  const formStepsChoices = formSteps.map(step => {
-    if (step.formDefinition) return [step.formDefinition, step.name];
-
-    return [step._generatedId, step.name];
-  });
-
```
SilviaAmAm, line 120: This seemed not to be used anywhere :shrug:

CharString: Doesn't look like it had side-effects, so removal is probably ok.

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
```diff
@@ -89,6 +89,7 @@ const VariableRow = ({index, variable}) => {
       <td>{getFormDefinitionName(variable.formDefinition)}</td>
       <td>{variable.prefillPlugin}</td>
       <td>{variable.prefillAttribute}</td>
+      <td>{variable.prefillIdentifierRole || 'main'}</td>
```
CharString, line 92: I first thought: shouldn't 'main' be translated?
But I suspect the attribute always has a proper value, if I look at the proptypes of the component

src/openforms/prefill/__init__.py
```diff
@@ -139,11 +145,16 @@ def prefill_variables(submission: "Submission", register=None) -> None:
...
-            variable.form_variable.prefill_attribute
-        )
+        plugin_id = variable.form_variable.prefill_plugin
+        attribute_name = variable.form_variable.prefill_attribute
+        identifier_role = variable.form_variable.prefill_identifier_role
+
+        if not grouped_fields[plugin_id].get(identifier_role):
+            grouped_fields[plugin_id][identifier_role] = []
+
+        grouped_fields[plugin_id][identifier_role].append(attribute_name)
```
CharString, line 157:
```suggestion
        grouped_fields[plugin_id].setdefault(identifier_role, []).append(attribute_name)
```

src/openforms/prefill/tests/test_prefill_hook.py
```diff
@@ -116,6 +116,105 @@ def apply_prefill(configuration: dict, submission: "Submission", register=None)
 
 
 class PrefillHookTests(TransactionTestCase):
+    @patch(
+        "openforms.prefill.contrib.haalcentraal.plugin.HaalCentraalPrefill.get_prefill_values",
+        return_value={"naam.voornamen": "John", "naam.geslachtsnaam": "Dodo"},
+    )
+    def test_fetch_values_with_multiple_bsn(self, m_bsn):
```
CharString, line 123: There's no bsn handling here, right?

Maybe call it multiple_people

CharString (changes requested):
I find the codecov messages in the plugin weird. I'd expect there would already have been tests for these codepaths.

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
SilviaAmAm: Mhmm yes it's true, maybe all of them should use the translatable label instead of the value :thinking:

SilviaAmAm: On second thoughts, it looks like this atm:
Screenshot from 2023-07-11 10-45-42
So I will just add the label for the identifier role in this PR.

docs/manual/forms/form_fields.rst
```diff
@@ -132,6 +132,13 @@ Registratie
   naar het achterliggende registratie systeem, dan kunt u hier een attribuut
   kiezen dat beschikbaar is in het achterliggende registratie systeem.
 
+Prefill
+-------
+
+* **Plugin**: Welke prefill plugin te gebruiken om data op te halen over de persoon/het bedrijf die ingelogd is.
+* **Pluginattribuut**: Welke attribuut van de prefill-backend data te gebruiken om deze component te prefillen.
+* **Identifier rol**: In het geval dat de gebruiker met DigiD Machtigen en eHerkenning Bewindvoering heeft ingelogd, moeten de gegevens gerelateerd aan het 'hoofd identifier' worden gebruikt, of die gerelateerd aan de gemachtigde persoon?
```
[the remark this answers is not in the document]
sergei-maertens: maybe something like this would be clearer?

```
**Identifier rol**: Bij DigiD Machtigen of eHekerkenning zijn meerdere rollen actief: de gemachtigde (persoon die het formulier invult), of de 'geauthenticeerde' (persoon of bedrijf voor wie het formulier ingevuld wordt). Deze optie stelt in
voor welke van deze rollen de relevante gegevens voor het geselecteerde attribuut opgevraagd worden.
```

I think "Geauthenticeerde" is a better description to refer to the person or company and we should provide a small explanation so that it's clear who's who.

src/openforms/authentication/contrib/digid_eherkenning_oidc/plugin.py
```diff
@@ -152,14 +152,17 @@ class DigiDMachtigenOIDCAuthentication(OIDCAuthentication):
...
-                "plugin": self.identifier,
-                "attribute": self.provides_auth,
-                "value": claim[config.vertegenwoordigde_claim_name],
-                "machtigen": request.session[DIGID_MACHTIGEN_OIDC_AUTH_SESSION_KEY],
-            }
+        if not claim or CO_SIGN_PARAMETER in request.GET:
+            return
+
+        config = self.config_class.get_solo()
+        machtigen_data = request.session[DIGID_MACHTIGEN_OIDC_AUTH_SESSION_KEY]
```
sergei-maertens, line 159: this could be

```suggestion
        machtigen_data = request.session[self.session_key]
```
to be consistent with the `self.config_class` above?

src/openforms/authentication/contrib/digid_eherkenning_oidc/plugin.py
```diff
@@ -179,16 +182,22 @@ class EHerkenningBewindvoeringOIDCAuthentication(OIDCAuthentication):
...
-                    EHERKENNING_BEWINDVOERING_OIDC_AUTH_SESSION_KEY
-                ],
-            }
+        if not claim or CO_SIGN_PARAMETER in request.GET:
+            return
+
+        config = self.config_class.get_solo()
+        machtigen_data = request.session[
+            EHERKENNING_BEWINDVOERING_OIDC_AUTH_SESSION_KEY
+        ]
```
sergei-maertens, line 191:
```suggestion
        machtigen_data = request.session[self.session_key]
```

src/openforms/authentication/contrib/digid_eherkenning_oidc/plugin.py
```diff
@@ -179,16 +182,22 @@ class EHerkenningBewindvoeringOIDCAuthentication(OIDCAuthentication):
...
+        request.session[FORM_AUTH_SESSION_KEY] = {
+            "plugin": self.identifier,
+            "attribute": self.provides_auth,
+            "value": claim[config.vertegenwoordigde_company_claim_name],
+            "machtigen": {
+                # TODO So far the only possibility is that this is a BSN.
+                "value": machtigen_data.get(config.gemachtigde_person_claim_name)
```
sergei-maertens, line 198: Do I then understand it correctly you always have two attributes?

* DigiD machtigen: BSN of the machtiger and BSN of the gemachtigde
* eHerkenning bewindvoering: KVK number of the company and BSN of the gemachtigde?

I thought that at some point there was also 3 attributes possible: company KVK, BSN of person responsible for company and then BSN of a gemachtigde acting on behalve of the first BSN?

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
```diff
@@ -111,13 +114,6 @@ const EditableVariableRow = ({index, variable, onDelete, onChange}) => {
   const prefillPluginChoices = availablePrefillPlugins.map(plugin => [plugin.id, plugin.label]);
   const [prefillAttributeChoices, setPrefillAttributeChoices] = useState([]);
 
-  const formSteps = formContext.formSteps;
-  const formStepsChoices = formSteps.map(step => {
-    if (step.formDefinition) return [step.formDefinition, step.name];
-
-    return [step._generatedId, step.name];
-  });
```
sergei-maertens, line 119: dead code?

src/openforms/js/components/admin/form_design/variables/constants.js
```diff
@@ -91,10 +92,47 @@ const EMPTY_VARIABLE = {
...
+  [
+    'authorised_person',
+    defineMessage({
+      description: 'Choices (main/authorised person) label',
+      defaultMessage: 'Authorised person',
+    }),
+  ],
+];
+
+const IDENTIFIER_ROLE_LABELS = {
```
sergei-maertens, line 119: I don't understand the duplication here?

Can't you re-use `IDENTIFIER_ROLE_CHOICES` (using `Object.fromEntries(IDENTIFIER_ROLE_CHOICES)`) and then apply `intl.formatMessage(options[value])`, with `intl` coming from `const intl = useIntl()`

src/openforms/js/components/admin/form_design/variables/types.js
```diff
@@ -8,6 +8,7 @@ const Variable = PropTypes.shape({
   source: PropTypes.string,
   prefillPlugin: PropTypes.string,
   prefillAttribute: PropTypes.string,
+  prefillIdentifierRole: PropTypes.string,
```
sergei-maertens, line 11:
```suggestion
  prefillIdentifierRole: PropTypes.oneOf(['main', 'authorised_person']),
```
we can make this stricter, likely we also should include the empty string for existing components.

src/openforms/prefill/__init__.py
```diff
@@ -56,25 +56,31 @@ def _fetch_prefill_values(
     )
 
     @elasticapm.capture_span(span_type="app.prefill")
-    def invoke_plugin(item: Tuple[str, List[str]]) -> Tuple[str, Dict[str, Any]]:
-        plugin_id, fields = item
-        plugin = register[plugin_id]
+    def invoke_plugin(
+        item: Tuple[str, dict[str, list[str]]]
+    ) -> Tuple[str, Dict[str, Any]]:
```
sergei-maertens, line 61:
```suggestion
    def invoke_plugin(
        item: tuple[str, dict[str, list[str]]]
    ) -> tuple[str, dict[str, Any]]:
```

src/openforms/prefill/__init__.py
```diff
@@ -56,25 +56,31 @@ def _fetch_prefill_values(
...
-            values = plugin.get_prefill_values(submission, fields)
-        except Exception as e:
-            logger.exception(f"exception in prefill plugin '{plugin_id}'")
-            logevent.prefill_retrieve_failure(submission, plugin, e)
-            return plugin_id, {}
-        else:
-            if values:
-                logevent.prefill_retrieve_success(submission, plugin, fields)
+        retrieved_results = {}
+        for identifier_role, fields in field_groups.items():
```
sergei-maertens, line 69: this for-loop now runs inside the thread pool executor, so this means that if you're fetching for two different BSNs with the same plugin that the requests happen sequentially, while we should be able to do these requests in parallel.

Essentially, the processing of the `field_groups` should happen outside of this function and be fed to the `with parallel()` context manager so that we minimize the amount of sequential HTTP requests for better performance.

I think that way the actual `invoke_plugin` would not see many changes

src/openforms/prefill/__init__.py
```diff
@@ -139,11 +145,13 @@ def prefill_variables(submission: "Submission", register=None) -> None:
...
 
-    grouped_fields = defaultdict(list)
+    grouped_fields = defaultdict(dict)
     for variable in variables_to_prefill:
-        grouped_fields[variable.form_variable.prefill_plugin].append(
-            variable.form_variable.prefill_attribute
-        )
+        plugin_id = variable.form_variable.prefill_plugin
+        attribute_name = variable.form_variable.prefill_attribute
+        identifier_role = variable.form_variable.prefill_identifier_role
```
sergei-maertens, line 152:
```suggestion
        identifier_role = variable.form_variable.prefill_identifier_role or "main"
```
for existing components that don't have this configuration option?

src/openforms/prefill/__init__.py
```diff
@@ -139,11 +145,13 @@ def prefill_variables(submission: "Submission", register=None) -> None:
     state = submission.load_submission_value_variables_state()
     variables_to_prefill = state.get_prefill_variables()
 
-    grouped_fields = defaultdict(list)
+    grouped_fields = defaultdict(dict)
```
sergei-maertens, line 148:
```suggestion
    grouped_fields = defaultdict(defaultdict(list)
```
and then you'd be able to avoid the `setdefault` call later on?

src/openforms/prefill/base.py
```diff
@@ -44,11 +48,28 @@ def get_co_sign_values(
...
+        Given a submission and the role of the identifier, return the value of the identifier.
+
+        The role of the identifier has to do with whether it is the 'main' identifier or an identifier
+        of someone logging in on behalf of someone/something else.
+
+        :param submission: an active :class:`Submission` instance
+        :param identifier_role: A string with one of the choices in :class:`IdentifierRoles`
+        :return: The value for the identifier
+        """
+        raise NotImplementedError(
```
sergei-maertens, line 73: can we not provide a default implementation that has the old behaviour? This seems like a bit of an unnecessary breaking change.

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -72,25 +72,35 @@ def _get_values_for_bsn(
 
         return values
 
-    @classmethod
-    def get_prefill_values(
-        cls, submission: Submission, attributes: Sequence[str]
-    ) -> dict[str, Any]:
+    def get_identifier_value(
+        self, submission: Submission, identifier_role: str
```
sergei-maertens, line 76: I think that `identifier_role` should be a strict type hint: `Literal["main", "authorised_person"]`

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -72,25 +72,35 @@ def _get_values_for_bsn(
...
-            or submission.auth_info.attribute != AuthAttribute.bsn
+            identifier_role == IdentifierRoles.main
+            and submission.auth_info.attribute == self.requires_auth
         ):
-            #  If there is no bsn we can't prefill any values so just return
-            logger.info("No BSN associated with submission, cannot prefill.")
+            return submission.auth_info.value
+
+        if identifier_role == IdentifierRoles.authorised_person:
+            return submission.auth_info.machtigen.get("value")
```
sergei-maertens, line 88: should we rename `value` to `identifier`? I'm worried that additional data might need to be stored in the future and "value" is too generic of a name.

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -72,25 +72,35 @@ def _get_values_for_bsn(
...
+
+    def get_prefill_values(
+        self,
```
sergei-maertens, line 91: I don't really see a reason why this can't be a classmethod still? None of the `self` instance attributes seem to be used.

sergei-maertens (commented):
Can you also use the constant/choices you defined instead of using the magic string `"main"` for defaults/fallbacks?

We should also properly document in the release notes that this contains breaking changes for the prefill plugin developers, as the public API now requires additional fields (and method signatures have changed).

src/openforms/prefill/contrib/haalcentraal/plugin.py
CharString: Should turn `get_identifier_value` into a classmethod too, in that case.

src/openforms/authentication/contrib/digid_eherkenning_oidc/plugin.py
SilviaAmAm: I think I told you that would get 2 BSNs in the eHErkenning bewindvoering case, but I was wrong! We get only 1 BSN back (I checked on the ACC environment of DH where bewindvoering works).

src/openforms/js/components/admin/form_design/variables/VariablesTable.js
SilviaAmAm: Yes, it didn't seem to be used anywhere!

src/openforms/prefill/__init__.py
SilviaAmAm: the `prefill_identifier_role` attribute on the variable has a default, so I think it should always have a value?

src/openforms/prefill/__init__.py
SilviaAmAm: Ahh I forgot that the argument must be a callable

src/openforms/prefill/contrib/haalcentraal/plugin.py
SilviaAmAm: I liked `value`, because in the `AuthInfo` model we have the fields `plugin`, `attribute`, `value`. So I thought it would make sense to use `value` also for the machtigen data. But I can change it if you think `identifier` is better

src/openforms/prefill/contrib/haalcentraal/plugin.py
SilviaAmAm: In `get_identifier_value` we need `self.requires_auth` :S

src/openforms/prefill/__init__.py
sergei-maertens: you're right, that should work!

src/openforms/prefill/contrib/haalcentraal/plugin.py
sergei-maertens: auth info has value because it only makes sense in that context of the entire auth info (it's basically a description of identifier), but the machtigen data is not well structured so I'd prefer the additional clarity.

src/openforms/prefill/contrib/haalcentraal/plugin.py
sergei-maertens: but that's a class attribute so should be available through `cls.requires_auth`, no?

src/openforms/prefill/base.py
SilviaAmAm: Do you mean that to change the base implementation with:
```python
        if not submission.is_authenticated:
            return

        if (
            identifier_role == IdentifierRoles.main
            and submission.auth_info.attribute == self.requires_auth
        ):
            return submission.auth_info.value
```
? 
(This method did not exist before the change, we would get the identifier inside `get_prefill_values`)

src/openforms/prefill/contrib/haalcentraal/plugin.py
SilviaAmAm: Oh yes, it's a class attribute... I'm a silly :facepalm:

src/openforms/prefill/base.py
sergei-maertens: I think that would work.

It's still highly dependent on the authentication plugin that's used though, but let's see if we can make things more agnostic or more formal in the next iteration(s) 🤔

src/openforms/prefill/types.py
```diff
@@ -0,0 +1,7 @@
+from typing import Literal
+
+from .constants import IdentifierRoles
+
+IdentifierRole = (
+    Literal[IdentifierRoles.main] | Literal[IdentifierRoles.authorised_person]
```
sergei-maertens, line 6: does that work? I don't think this can be extracted from static analysis 🤔

sergei-maertens (approved):
Ship it!

PR #3243, opened 2023-07-10
⬆️  [#3049] -- upgrade to django-treebeard 4.7

ErhanCitil:


CharString (commented):
From https://django-treebeard.readthedocs.io/en/latest/changes.html:
> Change the return value of delete() for all node classes to be consistent with Django, and return a tuple of the number of objects deleted and a dictionary with the number of deletions per object type.

Dit you check our codebase for calls to this function?

CharString (approved):
I've checked. Doesn't seem the case.

PR #3248, opened 2023-07-11
Upgrade pytest

CharString:
A QOL upgrade; the old pytest in the requirements is fighting with my
pytest-plugins. And we're in the process of updating real requirements,
so I keep breaking my test runner.


sergei-maertens (approved):
I didn't even realize we had it in the dependencies

PR #3261, opened 2023-07-17
[#3219] Remove uploads from registration if deleted

### LINKED ISSUE #3219: Uploads that have been removed are still present in registration

2.1.3

When a user uploads a file and contrinues to the next step, and later returns to a step to delete the uploaded file/upload another file, all files will be included in the registration, even the ones that were removed by the user.

upload file:
image

Continue the form to "overzicht" and click "wijzigen" to return tot he step with the upload

Remove current upload 

Add new upload:
image

"Overzicht" will show only the new upload:
image

But in the registration Both files are present:
image

The deleted upload should not be included in the registration.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Fixes #3219

### DIFF
--- a/src/openforms/submissions/api/views.py
+++ b/src/openforms/submissions/api/views.py
@@ -106,8 +106,7 @@ class TemporaryFileView(DestroyAPIView):
     def perform_destroy(self, instance):
         # delete files from disc as well if they had been already
         # saved when trying to access the next form step
-        if instance.attachments.exists():
-            instance.attachments.all().delete()
+        instance.attachments.all().delete()
 
         remove_upload_from_session(instance, self.request.session)
         instance.delete()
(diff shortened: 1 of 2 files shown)


SilviaAmAm (approved):

src/openforms/submissions/api/views.py
```diff
@@ -104,5 +104,10 @@ def get(self, request, *args, **kwargs):
         )
 
     def perform_destroy(self, instance):
+        # delete files from disc as well if they had been already
+        # saved when trying to access the next form step
+        if instance.attachments.exists():
+            instance.attachments.all().delete()
```
sergei-maertens, line 110: I think you can avoid a query by removing the `.exists`, or does that crash?

sergei-maertens (approved):

src/openforms/submissions/api/views.py
vaszig: I have understood that without calling any method the value will be always truthy. Tested locally and indeed it enters the if statement even if there are no attachments.

vaszig: Discussed with Sergei, the if statement here can be removed.

PR #3263, opened 2023-07-17
Add endpoint to create an appointment

### LINKED ISSUE #2471: Change appointments flow

sergei-maertens:
Partly closes #2471

Tasks

- [x] Set up models to store appointment data
- [x] Ensure that creating an appointment completes the submission
- [x] Register which plugin was active at the time of appointment creation
- [x] Ensure that the correct Celery task chain is kicked off
- [x] Ensure that the submission status endpoint is functional (and the URL is communicated in the response)
- [x] Validate the input against the plugin (products, location, date, datetime, contact detail fields)


src/openforms/formio/tests/test_validation.py
```diff
@@ -0,0 +1,31 @@
+from django.test import SimpleTestCase
+
+from rest_framework.serializers import ValidationError
+
+from ..validation import validate_formio_data
+
+
+class FormioValidationTests(SimpleTestCase):
+    def test_textfield_required_validation(self):
```
SilviaAmAm, line 9: I think that codecov also wants a test where there is no `validate` in the component

src/openforms/formio/validation.py
```diff
@@ -0,0 +1,131 @@
...
+
+    """
+    wrapper = FormioConfigurationWrapper({"components": components})
+    paths = wrapper.reverse_flattened
+    data = FormioData(values)
+
+    errors = {}
+
+    for component in wrapper:
+        if "key" not in component:
```
SilviaAmAm, line 112: Are there components without key?

src/openforms/submissions/api/mixins.py
```diff
@@ -0,0 +1,58 @@
...
+
+class SubmissionCompletionMixin:
+    request: Request
+
+    def _complete_submission(self, submission: Submission) -> str:
+        """
+        Mark the submission as completed.
+
+        This encapsulates the logic of what it means to 'complete' a submission,
+        ensuring that the relevant metadata is set and post-completation hooks trigger,
```
SilviaAmAm, line 28:
```suggestion
        ensuring that the relevant metadata is set and post-completion hooks trigger,
```

src/openforms/submissions/api/mixins.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+
+        # dispatch signal for modules to tap into
+        submission_complete.send(sender=self.__class__, request=self.request)
+
+        submission.calculate_price(save=False)
+        submission.completed_on = timezone.now()
+        submission.save()
+
+        persist_user_defined_variables(submission, self.request)
```
SilviaAmAm, line 39: Does this cause problems because the rules are evaluated twice? I think both `calculate_price` and `persist_user_defined_variables` may evaluate them and I have some memories of this causing problems :thinking:

src/openforms/appointments/api/views.py
```diff
@@ -341,3 +346,38 @@ def post(self, request, *args, **kwargs):
...
+            description=_("Invalid or missing submission in session."),
+        ),
+        "5XX": OpenApiResponse(
+            response=ExceptionSerializer,
+            description=_("Unable to cancel appointment."),
+        ),
+    },
+)
+class CreateAppointmentView(SubmissionCompletionMixin, CreateAPIView):
+    authentication_classes = ()
```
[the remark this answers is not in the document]
sergei-maertens: it should have it! I can't remember why this was copy-pasted this way...

src/openforms/formio/validation.py
sergei-maertens: I will turn it into an assert to satisfy the type checker!

src/openforms/submissions/api/mixins.py
sergei-maertens: the fuck did I do there

src/openforms/submissions/api/mixins.py
sergei-maertens: I don't think so, as I just moved this code out of the API endpoint into a re-usable mixin?

PR #3265, opened 2023-07-17
[#586] Move SOAP services to own module

### LINKED ISSUE #586: Suwinet prefill
Als burger van de gemeente wil ik graag dat mijn aanvraagformulieren zoveel mogelijk voor ingevuld worden. Voor het ophalen van data bij de Belastingdienst of bij Suwinet, moet ik hiervoor een juist subOIN meegeven met de serviceaanroep. De aangeroepen servive kan dan controleren of dit verzoek om uitwisseling van data gehonoreerd mag worden. 

There are many different APIs in Suwinet:

* Bijstandsregelingen
* BRPDossierPersoonGSD
* DUODossierPersoonGSD
* DUODossierStudiefinancieringGSD
* GSDDossierReintegratie
* IBVerwijsindex
* KadasterDossierGSD
* RDWDossierDigitaleDiensten
* RDWDossierGSD
* SVBDossierPersoonGSD
* UWVDossierAanvraagUitkeringStatusGSD
* UWVDossierInkomstenGSDDigitaleDiensten
* UWVDossierInkomstenGSD
* UWVDossierQuotumArbeidsbeperktenGSD
* UWVDossierWerknemersverzekeringenGSDDigitaleDiensten
* UWVDossierWerknemersverzekeringenGSD
* UWVWbDossierPersoonGSD

The responses can contain many nested keys. For example, a person can have multiple salaries + benefits (so the responses would correspond to "arrays of objects" if turned into JSON).

This is problematic because the form designer does not know in advance how much data will come back.

To discuss:
- How will the form designer configure how a component will be prefilled? 
- How are we going to deal with "prefilling repeating groups"?
- How do we deal with the many APIs of Suwinet? 
- Do we need to do extra processing of the data? For example, convenience method to calculate the total income or something.

SilviaAmAm:
Partly Fixes #586 

Moving the SOAP services out of the Stuf app.


src/soap/admin.py
```diff
@@ -0,0 +1,40 @@
...
+                    "client_certificate",
+                    "server_certificate",
+                ]
+            },
+        ),
+    )
+
+    class Meta:
+        model = SoapService
+        fields = "__all__"
```
sergei-maertens, line 40: that's not a thing with the django admin, is it?

src/soap/admin.py
```diff
@@ -0,0 +1,40 @@
+from django.contrib import admin
+from django.utils.translation import gettext_lazy as _
+
+from soap.models import SoapService
```
sergei-maertens, line 4:
```suggestion
from .models import SoapService
```
let's use relative imports inside the same app, in case it needs to be moved around more

sergei-maertens (approved):

PR #3267, opened 2023-07-18
Allow changing registration plugin and zaaktype with logic

### LINKED ISSUE #3179: Allow to change registration plugin and or change "zaaktype" with logic
Als formulierbouwer wil ik binnen formulieren de mogelijkheid om o.b.v. vragen/validatie/logica, binnen één formulier kan sturen/flowen naar:

* één van meerdere mogelijke zaaktypen
* mailverwerking van het formulier

Zodat ik voor formulieren die o.b.v. een nuance verschil, op input van de burger/of het bedrijf, moeten resulteren in een verschillend zaaktype, of mailverwerking niet meerdere formulieren hoef te bouwen.

**Argumenten**

* In theorie kun je ook flowen middels een beslisboom, maar volgens Haarlemmermeer haalt dit vragen uit de context.
* het is niet wenselijk om meerdere formulieren te hosten met dezelfde naam met een nuanceverschil.

Related: Dimpact PF-82

**Tasks**

* [x] Mogelijkheid tot toevoegen meerdere registratieplugins op een formulier 
* [ ] Mogelijkheid om in logica te kiezen tussen registratieplugin
* [ ] Mogelijkheid om voor bepaald zaaktype te kiezen op basis van logica.
* [ ] Omschrijving kan afwijken op basis van logica regels (maar hier is een ander ticket op)

CharString:
Closes #3179


src/openforms/forms/api/serializers/form.py
```diff
@@ -217,21 +260,62 @@ def create(self, validated_data):
...
+        new_data = {}
+        if "regisration_backend" in validated_data:
+            new_data["backend"] = validated_data.pop("registration_backend")
+        if "registration_backend_options" in validated_data:
+            new_data["options"] = validated_data.pop("registration_backend_options")
+        if any(v is not None for v in new_data.values()):
+            backend = instance.registration_backends.first()
+            FormRegistrationBackendSerializer(
+                instance=backend, context=self.context
+            ).update(backend, new_data)
```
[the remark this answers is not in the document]
sergei-maertens: also if you're going to use the serializer anyway, I think it'd make more sense to do:

```py
backend = instance.registration_backends.first()
_serializer = FormRegistrationBackendSerializer(instance=backend, data=new_data, context=self.context, partial=self.partial)
assert _serializer.is_valid(raise_exception=False)
_serializer.save()  # update or create the instance depending on whether one existed or not
```
I'm not sure what the desired behaviour is for create, I'd reckon if you PATCH or PUT registration backend (+ options) to the form and none exists yet, it should be created, so then the additional required fields for a new instance should be provided too (or you create a new instance using `instance = backend or FormRegistrationBackend(form=self.instance, key="default", name=_("Default"))`

You can also pass additional kwargs to `save` like `save(form=self.instance)` if that could help with things.

src/openforms/forms/api/serializers/form.py
```diff
@@ -217,21 +260,62 @@ def create(self, validated_data):
...
+
         instance = super().update(instance, validated_data)
         ConfirmationEmailTemplate.objects.set_for_form(
             form=instance, data=confirmation_email_template
         )
+        if registration_backends is None:
+            return instance
+
+        with transaction.atomic():
```
sergei-maertens, line 312: I don't think this is needed because the entire method is already decorated.

src/openforms/forms/api/serializers/form.py
```diff
@@ -245,16 +329,79 @@ def get_fields(self):
...
                 ("", "")
             ] + payment_register.get_choices()
+
+        # add fields for v2 backwards compatibility
+        if "registration_backends" in fields:
+            fields["registration_backend"] = fields[
+                "registration_backends"
+            ].child.get_fields()["backend"]
+            fields["registration_backend"].required = False
+            fields["registration_backend"].allow_null = True
```
sergei-maertens, line 339: this was not originally nullable in v2 - is this deliberate?

src/openforms/forms/api/serializers/form.py
```diff
@@ -245,16 +329,79 @@ def get_fields(self):
             fields["payment_backend"].choices = [
                 ("", "")
             ] + payment_register.get_choices()
+
+        # add fields for v2 backwards compatibility
+        if "registration_backends" in fields:
+            fields["registration_backend"] = fields[
```
sergei-maertens, line 335: have you checked if this returns the original instance or a copy of it? I think it's the latter, but it would suck if this causes the validation to be relaxed in the new serializer :thinking:

src/openforms/forms/api/serializers/form.py
```diff
@@ -245,16 +329,79 @@ def get_fields(self):
...
+                "registration_backends"
+            ].child.get_fields()["backend"]
+            fields["registration_backend"].required = False
+            fields["registration_backend"].allow_null = True
+            fields["registration_backend"].allow_blank = True
+            fields["registration_backend_options"] = fields[
+                "registration_backends"
+            ].child.get_fields()["options"]
```
sergei-maertens, line 343: we could lift this up:

```py
child_fields = fields["registration_backends"].child.get_fields()
```

to avoid the repetition

src/openforms/forms/api/serializers/form.py
```diff
@@ -245,16 +329,79 @@ def get_fields(self):
...
+                                    code="invalid",
+                                ),
+                                "registration_backend_options": ErrorDetail(
+                                    _(
+                                        "registration_backend is deprecated, please use just registration_backends",
+                                    ),
+                                    code="invalid",
+                                ),
+                            }
+                        )
```
sergei-maertens, line 395: Can we simplify our life and make this mutual exclusive? Just always say use one or the other.

Old exports should only have `registration_backend` and `registration_backend_options`. New exports should only have `registration_backends` (so make sure that the old fields are dropped in new exports?). We don't support importing newer formats into older versions anyway, backwards compatibility is already hard enough without handling forwards compatibility :stuck_out_tongue:

src/openforms/forms/api/serializers/form.py
```diff
@@ -355,6 +502,8 @@ def get_fields(self):
             del fields["payment_options"]
         if "authentication_backends" in fields:
             fields["authentication_backends"].write_only = False
+        if "registration_backend" in fields:
+            del fields["registration_backend"]
```
sergei-maertens, line 506: shouldn't this also drop `registration_backend_options`?

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -225,4 +235,17 @@ def validate(self, data: dict) -> dict:
...
+        #             {"action": {"value": _("Invalid registration backend")}},
+        #             code="blank",
+        #         )
+        #     # form is not in scope here :(
+        #     # if not form.registration_backends.filter(key=action_value).exists():
+        #     #     raise serializers.ValidationError(
+        #     #         {"action": {"value": _("Invalid registration backend")}},
+        #     #         code="invalid",
+        #     #     )
+        #
```
sergei-maertens, line 250: have you tried getting it from `self.context["form"]` or adding it to the context if it's not there yet? But from `def logic_rules_bulk_update` it seems like it should be there.

src/openforms/forms/migrations/0088_enable_mutiple_registration_backends.py
```diff
@@ -0,0 +1,111 @@
...
+import openforms.registrations.fields
+from openforms.registrations.registry import register
+
+
+def forward(apps, _):
+    Form = apps.get_model("forms", "Form")
+    FormRegistrationBackend = apps.get_model("forms", "FormRegistrationBackend")
+    for form in Form.objects.exclude(
+        models.Q(registration_backend="") | models.Q(registration_backend__isnull=True)
+    ):
```
sergei-maertens, line 14: I think it's possible to have registration backends without options? Like ZGW with default options? It doesn't look particularly realistic, *but* it could also be partial configuration and I wonder if there's any harm in copying those over?

src/openforms/forms/migrations/0088_enable_mutiple_registration_backends.py
```diff
@@ -0,0 +1,111 @@
...
+        backend = form.registration_backend
+        # getting the plugin for a human readable name
+        # as form.get_registration_backend_display doesn't exist
+        plugin = register[backend]
```
sergei-maertens, line 18: I think this can lead to interesting crashes if people have stale data with custom registration plugins that have since been removed. Perhaps this can be done in a more robust way:

```py
backend_label = register[backend].get_label() if backend in register else "Surprise!"
```

src/openforms/forms/migrations/0088_enable_mutiple_registration_backends.py
```diff
@@ -0,0 +1,111 @@
...
+            backend=backend,
+            options=form.registration_backend_options or {},
+        )
+
+
+def reverse(apps, _):
+    # This will lose data, but preserves the simple cases of just one backend
+    # No attempt at determinism for the multi-backend case is done; there will
+    # probably be logic rules that will break anyway. Ops need to use db backups,
+    # this is for this dev, that has "Update demo fixutures" at the end of this ticket's
```
sergei-maertens, line 32:
```suggestion
    # this is for this dev, that has "Update demo fixtures" at the end of this ticket's
```

src/openforms/forms/models/form_registration_backend.py
```diff
@@ -0,0 +1,43 @@
...
+    options = models.JSONField(
+        _("registration backend options"),
+        default=dict,
+        blank=True,
+    )
+
+    class Meta:
+        unique_together = ("form", "key")
+
+    def get_backend_display(self):
```
sergei-maertens, line 36: please also implement `def __str__` as that is used in the admin, plus the `Meta.verbose_name` and `Meta.verbose_name_plural`, again for the admin.

src/openforms/forms/tests/factories.py
```diff
@@ -1,12 +1,30 @@
...
 from ..utils import form_to_json
 
 
+def random_registration_plugin():
+    return random.choice(
+        [p.identifier for p in registration_registry.iter_enabled_plugins()]
+    )
+
+
+class FormRegistrationBackendFactory(factory.django.DjangoModelFactory):
```
sergei-maertens, line 19: this is missing the `form` subfactory, which is a required FK

src/openforms/forms/tests/test_api_import_export.py
```diff
@@ -183,7 +186,14 @@ def test_form_import(self):
         self.assertNotEqual(imported_form.pk, form1.pk)
         self.assertNotEqual(imported_form.uuid, str(form1.uuid))
         self.assertEqual(imported_form.active, False)
-        self.assertEqual(imported_form.registration_backend, form1.registration_backend)
+        for imported, original in zip(
+            imported_form.registration_backends.all(),
+            form1.registration_backends.all(),
+        ):
```
sergei-maertens, line 192: this is a bit fragile because there's no explicit ordering specified, not in the qs here nor in the model meta. So I'd add an `order_by("id")` here for test robustness.

It's now up to the DB engine to determine in which order results are returned.

src/openapi.yaml
```diff
@@ -7149,6 +7141,29 @@ components:
...
+          description: A recognisable name for this backend configuration.
+          maxLength: 255
+        backend:
+          type: string
+          title: Registration backend
+          maxLength: 100
+        options:
+          type: object
+          additionalProperties: {}
+          title: Registration backend options
```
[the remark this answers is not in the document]
CharString: That was a bit of a lie:

 - There are no plugins with `allow_null` `configuration_options`
 - `not plugin.configuration_options` in `validate_backend_options` is hard to be true because
```python
    def check_plugin(self, plugin):
        if not plugin.configuration_options:
            raise ValueError(...
```

But yes, it now handles it and tests for it.

src/openforms/forms/api/serializers/form.py
```diff
@@ -217,21 +260,62 @@ def create(self, validated_data):
         confirmation_email_template = validated_data.pop(
             "confirmation_email_template", None
         )
+        registration_backends = validated_data.pop("registration_backends", [])
+
+        registration_backend_options = validated_data.pop(
+            "registration_backend_options", {}
+        )
+        if "registration_backend" in validated_data:
```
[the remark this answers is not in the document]
CharString: It is. There are api PATCH tests that change single values in the options dict.

src/openforms/forms/api/serializers/form.py
CharString: The admin UI was sending nulls, but I've since fixed that.

src/openforms/forms/api/serializers/form.py
CharString: It isn't the same instance as `FormRegistrationBackendSerializer().fields['backend']`, but I'm not sure if that's enough. I'll rewrite this in a way I'm sure it's not the same instance.

src/openforms/forms/api/serializers/logic/action_serializers.py
CharString: `form = self.context["form_variables"].form`, but turns out you already get a "invalid_choice" error as is.

src/openforms/js/components/admin/form_design/RegistrationFields.js
```diff
@@ -76,10 +89,51 @@ const RegistrationFields = ({
...
+        //         defaultMessage="The key to use to refer to this configuration in form logic."
+        //         description="Backend key help text"
+        //       />
+        //     }
+        //     disabled
+        //   >
+        //   <TextInput onChange={onChange} value={backendKey} />
+        //   </Field>
+        // </FormRow>
+        //
```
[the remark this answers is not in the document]
CharString: Yes initially I had this as an input, because I had the action interface use logic that resolved to the key. So having it user defined was nicer. But no more...

src/openforms/payments/tests/test_services.py
```diff
@@ -54,7 +55,16 @@ def create_payment(self, **kwargs):
...
-        return SubmissionPaymentFactory.create(**good_kwargs)
+        registration_backend_kwargs = {
+            "backend": good_kwargs.pop("submission__form__registration_backend"),
+            "options": good_kwargs.pop("submission__form__registration_options", {}),
+        }
+
+        submission = SubmissionPaymentFactory.create(**good_kwargs)
+        FormRegistrationBackendFactory.create(
+            form=submission.form, **registration_backend_kwargs
+        )
```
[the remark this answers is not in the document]
CharString: I tried a few times and failed. And not having it ended up showing me the extent of the change.

src/openforms/registrations/contrib/email/utils.py
```diff
@@ -11,25 +14,32 @@ class RegistrationEmailTemplates:
     content_text: str
 
 
-def get_registration_email_templates(submission) -> RegistrationEmailTemplates:
-    config = EmailConfig.get_solo()
+def get_registration_email_templates(
+    submission: Submission,
+) -> RegistrationEmailTemplates:
+    global_config = EmailConfig.get_solo()
```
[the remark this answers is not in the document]
CharString: Or a .pyi for get_solo..

src/openforms/submissions/models/submission.py
```diff
@@ -727,3 +737,25 @@ def waiting_on_cosign(self) -> bool:
...
+    @property
+    def registration_backend(self) -> FormRegistrationBackend:
+        if self.finalised_registration_backend:
+            return self.finalised_registration_backend
+
+        match self._registration_backend:
+            case str(key):
+                self._registration_backend = self.form.registration_backends.get(
+                    key=key
+                )
```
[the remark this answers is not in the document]
CharString: Right, and then what? fallback to the default? Raise?

src/openforms/submissions/models/submission.py
```diff
@@ -727,3 +737,25 @@ def waiting_on_cosign(self) -> bool:
...
+            case None:
+                self._registration_backend = self.form.registration_backends.first()
```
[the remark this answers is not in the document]
CharString: No even for new cases: if I specify just one backend, no logic rules are needed.

Forgetting to set one or setting a wrong one was specced as undefined behaviour.

I think the correct behaviour would be to skip registration. But as per your other comment, I'll revert to the first in a defined order, and also log debug info to the timeline of the submission.

src/openforms/submissions/logic/actions.py
```diff
@@ -263,10 +266,33 @@ def get_action_log_data(
...
+    registration_backend_key: str
+
+    @classmethod
+    def from_action(cls, action: ActionDict) -> "SetRegistrationBackendAction":
+        return cls(registration_backend_key=action["action"]["value"])
+
+    def apply(
+        self, step: SubmissionStep, configuration: FormioConfigurationWrapper
+    ) -> None:
+        step.submission.registration_backend = self.registration_backend_key
```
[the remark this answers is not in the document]
CharString: `str` is a FormRegistrationBackendKey
since (form_id, backend_key) is a compound key, it's enough for form logic to do a cheap operation of setting such a "str"

On finalisation it's resolved to a FormRegistrationBackend.

I've changed the cryptic `str` to a NewType to express the intent. With algebraic data types, the match would be much cleaner, but on runtime it's still just a `str`.

src/openapi.yaml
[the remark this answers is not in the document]
sergei-maertens: `plugin.configuration_options` in this case expects a `serializers.Serializer` subclass, I don't think it says anything about that serializer being `allow_null` :thinking: 

So checking things - the serializer points to the model field, which has `allow_null`, so DRF treats this as a nullable JSONField (which explains the old schema). Next, the serializer validation runs:

```py
self.validate_backend_options(
            attrs,
            "registration_backend",
            "registration_backend_options",
            registration_register,
        )
```
which results in the actual options validation via:

```py
serializer = plugin.configuration_options(data=options)
```
there are code paths here where `options is None`, but since there is no `allow_null=True` kwarg, I think it was effectively indeed blocked and would result in validation errors.

src/openforms/forms/api/serializers/form.py
```diff
@@ -217,21 +269,87 @@ def create(self, validated_data):
...
             form=instance, data=confirmation_email_template
         )
+        if registration_backends is None:
+            return instance
+
+        instance.registration_backends.all().delete()
+        FormRegistrationBackend.objects.bulk_create(
+            FormRegistrationBackend(form=instance, **backend)
+            for backend in registration_backends
+        )
```
sergei-maertens, line 351: So edge case:

1. form with registration backend is configured
2. submission is created and completed for this registration backend
3. submission registration attempt fails
4. form is updated with new configuration (possibly to fix registration issue)
5. submission registration attempt retries

The FK from submission to registration backend is broken due to the `models.SET_NULL` off the top of my head - which makes the registration retry break as there is no longer any registration meta-information available.

I'm not sure what the best approach is here - maybe just store the `key` of the registration backend used? Since copying/freezing the information doesn't allow fixing the registration issues after the fact :thinking:

src/openforms/forms/migrations/0088_enable_mutiple_registration_backends.py
sergei-maertens: agreed that it's pretty much only possible directly in the admin/ORM, the API doesn't seem to allow it.

src/openforms/forms/tests/factories.py
```diff
@@ -32,6 +41,37 @@ class Params:
...
+    def registration_backend(form, create, extracted, **kwargs):
+        if extracted is None:
+            return
+        (
+            FormRegistrationBackendFactory.create
+            if create
+            # Does build even make sense? Is it just to keep the garbage collector busy?
+            # Let's anyway; maybe the constructor has side-effects; everything is possible.
+            else FormRegistrationBackendFactory.build
+        )(form=form, backend=extracted, **kwargs)
```
sergei-maertens, line 64: just FYI - I think doing `SomeFactory()` will invoke the same build strategy as the parent/root, but i'm not 100% sure. Either way I prefer the explicit approach anyway.

src/openforms/forms/tests/test_factories.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_registration_backend_creates_a_config(self):
+        form = FormFactory.create(registration_backend="demo")
+        self.assertEqual(form.registration_backends.count(), 1)
+        self.assertTrue(FormRegistrationBackend.objects.get(form=form, backend="demo"))
+
+    def test_registration_backend_options_get_passed(self):
+        form = FormFactory.create(
+            registration_backend="email",
+            registration_backend_options={"to_emails": ["me@example.com"]},
+        )
```
sergei-maertens, line 17: TIL how to make that work! Very cool

src/openforms/forms/tests/test_serializers.py
```diff
@@ -83,6 +89,33 @@ def test_date_variable_with_non_trivial_json_trigger(self):
...
+            context=context,
+        )
+
+        with self.assertRaises(ValidationError) as raised:
+            serializer.is_valid(raise_exception=True)
+
+        self.assertEqual(
+            raised.exception.args[0]["action"]["type"][0].code,
+            "invalid_choice",
+        )
```
sergei-maertens, line 117: why not:

```py
is_valid = serializer.is_valid()

self.assertFalse(is_valid)
# assert on serializer.errors["action"]["type"]
```
the exception flow is DRF internals (as the conversion of the exception to an error response is handled entirely in DRF itself even and *could* be subject to change).

src/openforms/forms/tests/test_serializers.py
```diff
@@ -83,6 +89,33 @@ def test_date_variable_with_non_trivial_json_trigger(self):
 
         self.assertTrue(serializer.is_valid())
 
+    def test_registration_backend_action_errors_no_value(self):
```
sergei-maertens, line 92: test name makes me think this should be a validation error because the `value` is missing, but in fact what is being tested is that a valid `type` is specified. I think we need both tests?

src/openforms/forms/tests/test_serializers.py
```diff
@@ -135,3 +168,173 @@ def test_form_without_cosign(self):
...
+            backend="email",
+            options={"to_emails": ["me@example.com"]},
+        )
+        context = {"request": None}
+        data = FormSerializer(context=context).to_representation(form)
+        # remove v2 data
+        del data["registration_backend"]
+        del data["registration_backend_options"]
+
+        # delelte the first line
```
sergei-maertens, line 194:
```suggestion
        # delete the first line
```

src/openforms/payments/services.py
```diff
@@ -23,7 +23,7 @@ def update_submission_payment_registration(submission: Submission):
         return
 
     try:
-        plugin = register[submission.form.registration_backend]
+        plugin = register[submission.registration_backend.backend]
```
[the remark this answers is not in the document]
sergei-maertens: I'm not sure if catching `AttributeError` here is the best option? It would silently fail typo's or other unintended `AttributeErrors`. I think I'd rather see something like:

```py
if (registration_backend := submission.registration_backend) is None:
    logevent.registration_payment_update_failure(submission, error=e)
    return
try:
    plugin = register[registration_backend.backend]
except KeyError as e:
    ...
```
the duplicated logevent line is annoying and I hope to introduce `structlog` which can de-duplicate this at some point.

src/openforms/registrations/contrib/email/utils.py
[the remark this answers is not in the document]
sergei-maertens: yeah I've tried that and could *not* get it to work...

src/openforms/submissions/logic/actions.py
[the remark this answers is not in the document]
sergei-maertens: related to the delete and `models.SET_NULL`, I'm thinking it might be the best option to just use a CharField on the model and store the key, exactly because the combination of `(form_id, key)` uniquely identifies the backend from the model. Do keep in mind that it's possible the key no longer exists (so we don't get referential integrity with this approach).

src/openforms/tests/test_x_forwarded_host.py
```diff
@@ -11,6 +11,7 @@ class XForwardedHostTests(SimpleTestCase):
 
     factory = RequestFactory()
 
+    @override_settings(USE_X_FORWARDED_HOST=False)
```
sergei-maertens, line 14: this tests that our settings set it to disabled by default, the intent is coupling the test to the relevant setting and this change decouples it.

src/openforms/submissions/models/submission.py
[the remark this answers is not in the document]
sergei-maertens: this was mostly talking about existing data - it seems weird that you can have:

1. an existing submission that only has max 1 registration backend
2. you update the form for that submission and add another registration backend
3. looking at the existing submission now suddenly points to the new registration backend, which would mess up registration retries

So using defined order here should now be fine if the order lines up with the data migration that copies form registration configuration into the new model. But I think it'd have removed this implicit behaviour and recorded the registration backend used explicitly on the submission model after the new field was/is added (and I think in particular if we switch to a CharField instead of FK this is easily achieved, even more so if we add the charfield to the submission first, copy the information over and only then update the `Form` schema :thinking: , this would remove all ambiguity)

src/openforms/tests/test_x_forwarded_host.py
CharString: I guess I popped off 1 to many of my stash. This was tripping my local test runs, from when I was using mitmproxy to test the Signicat DigiD interactions.

src/openforms/forms/tests/test_factories.py
sergei-maertens: it's good for your character development :wink:

src/openforms/tests/test_x_forwarded_host.py
sergei-maertens: could also be because of your `.env` configuration? commenting that out might help

src/openforms/submissions/models/submission.py
```diff
@@ -284,6 +285,14 @@ class Submission(models.Model):
         ),
     )
 
+    finalised_registration_backend_key = models.CharField(
+        _("final registration backend key"),
+        max_length=50,  # FormRegistrationBackend.key.max_length
+        null=True,
```
sergei-maertens, line 291: we don't normally make text-based fields nullable, as the empty string is the empty value

src/openforms/submissions/models/submission.py
```diff
@@ -737,3 +746,58 @@ def waiting_on_cosign(self) -> bool:
...
+            case _:
+                default = backends.first()
+                registration_debug(
+                    self,
+                    extra_data={
+                        "message": _("Multiple backends defined on form"),
+                        "backend": {"key": default.key, "name": default.name},
+                    },
+                )
+                return default.key
```
sergei-maertens, line 777: There's some queries that can be avoided here, now you do queries for:

1. getting the count
2. query to get the first backend

you can make this `O(1)` by using `len(backends)` and `backends[0]`

CharString: Ackchyually....

It was already `O(1)`; the number of operations was a constant factor `2`  :wink:
It was `Ө(2)`; big theta of 2.

sergei-maertens (approved):

PR #3269, opened 2023-07-18
Disable more (irrelevant) form configuration for appointment forms

### LINKED ISSUE #2471: Change appointments flow

sergei-maertens:
Partly fixes #2471

A bunch of fields/tabs are not relevant for appointment fields, such as registration and payment but even the individual form steps as they are hard-coded.

### DIFF
--- a/src/openforms/js/components/admin/form_design/form-creation-form.js
+++ b/src/openforms/js/components/admin/form_design/form-creation-form.js
@@ -1124,6 +1124,8 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl}) => {
   // dev/debug helper
   const activeTab = new URLSearchParams(window.location.search).get('tab');
 
+  const {isAppointment = false} = state.form.appointmentOptions;
+
   return (
     <ValidationErrorsProvider errors={state.validationErrors}>
       <FormObjectTools isLoading={loading} historyUrl={formHistoryUrl} formUrl={formUrl} />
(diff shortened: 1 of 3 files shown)


src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1169,53 +1171,63 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl}) => {
             <Tab hasErrors={state.tabsWithErrors.includes('form')}>
               <FormattedMessage defaultMessage="Form" description="Form fields tab title" />
             </Tab>
-            <Tab hasErrors={state.tabsWithErrors.includes('form-steps')}>
-              <FormattedMessage
-                defaultMessage="Steps and fields"
-                description="Form design tab title"
-              />
-            </Tab>
+            {!isAppointment && (
```
SilviaAmAm, line 1174: Why is this condition needed? I would expect that since that an appointment form doesn't have steps/variables/registration backend, it wouldn't have errors?

sergei-maertens: This is hiding the tab for steps and fields, not the "non form field"-errors. Or am I confused here?

SilviaAmAm: Ohhh my bad, I understand now.

PR #3270, opened 2023-07-19
Add endpoint for address geo search

### LINKED ISSUE #2174: Kaartmateriaal - Zoeken op adres
Als burger wil ik binnen de kaart-interface mijn adres kunnen zoeken, zodat de kaartserver op basis van mijn zoekopdracht het adres en de juiste coordinaten toont en verwerkt.

Voorbeeld: PDOK locatieserver: [Introductie - PDOK](https://www.pdok.nl/introductie/-/article/pdok-locatieserver-1)

**Tasks**

* [x] Invoerveld toevoegen voor adres
* [x] Automatisch meer inzoomen als je dit doet.
* [x] Ook gemeente ophalen uit Locatieserver

Oorsprong: https://github.com/Dimpact-Samenwerking/Open-Formulieren/issues/41 RTM: 7.10 Story points: 5 Business value: 3

bart-maykin:
fixes #2174 

- base setting for the pdok locatie server url
- added api endpoint to fetch and massage bag data for the map component
- added unit tests for the new api endpoint

[leaflet-map-recording.webm](https://github.com/open-formulieren/open-forms/assets/101265650/2b994ac6-a24a-4b1b-b3e6-81bf4dfc976a)


src/openforms/formio/api/serializers.py
```diff
@@ -51,3 +51,19 @@ def get_url(self, instance) -> str:
...
+
+
+class MapSearchLatLng(serializers.Serializer):
+    lat = serializers.FloatField(label=_("Latitude"))
+    lng = serializers.FloatField(label=_("Longitude"))
+
+
+class MapSearchRD(serializers.Serializer):
+    x = serializers.FloatField(label=_("X"))
+    y = serializers.FloatField(label=_("Y"))
```
CharString, line 63: It this the RijksDriehoek projection of the same location?
Do we use it?

From the mock in  open-formulieren/open-forms-sdk#486 it doesn't seem we use it (doest leaflet even support it?), yet in the processing of the pdok response, we require them to be present...

I ask because:

Since they are both different representations a point in space, what does it mean when they point to different points? So I'd prefer serving 1 representation and applying a transform on the client.

src/openforms/conf/base.py
```diff
@@ -1182,6 +1182,10 @@
     "LOG_OUTGOING_REQUESTS_MAX_AGE", default=7 * 24
 )  # number of hours
 
+# Pdok api locatie server urldeddddddd77487
+PDOK_LOCATIE_SERVER_URL = config(
+    "PDOK_LOCATIE_SERVER_URL", "https://api.pdok.nl/bzk/locatieserver/search/v3_1/"
+)
```
CharString, line 1188: "urldeddddddd77487" ?

Locatieserver is one word in Dutch. Please correct that in all locations. I like that you stuck with the name of the service.

CharString (changes requested):

src/openforms/formio/api/views.py
```diff
@@ -86,3 +95,81 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+                            }
+                        )
+
+        return Response(
+            self.serializer_class(
+                instance=locations,
+                context={"request": request},
+                many=True,
+            ).data
+        )
```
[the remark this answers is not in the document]
CharString: Oh and are the docs in the response guaranteed to be sorted on descending score?

src/openforms/conf/base.py
bart-maykin: I have absolutely no clue how this happend and how I missed it.

src/openforms/formio/api/serializers.py
bart-maykin: @sergei-maertens asked me to expose it, because Leaflet works with RijksDriehoek although doesn't support the usage of it at the moment.

src/openforms/formio/api/views.py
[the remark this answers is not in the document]
bart-maykin: https://www.pdok.nl/restful-api/-/article/pdok-locatieserver-1#operation/free the default value of the sort param is on "score desc" so it should be.

src/openforms/formio/api/serializers.py
sergei-maertens: ideally we'd use the RD coordinates only at some point because there's always precision loss with projections. I don't want to get into the complexities of Leaflet on that aspect right now because it's very confusing, but it shouldn't hurt to expose both WSG:84 and RD coordinates _if we have them_.

src/openforms/formio/api/urls.py
```diff
@@ -10,4 +10,9 @@
         TemporaryFileUploadView.as_view(),
         name="temporary-file-upload",
     ),
+    path(
+        "mapsearch",
+        MapSearchView.as_view(),
+        name="map-search",
+    ),
```
sergei-maertens, line 17: this functionality has nothing to do with Form.io (it just happens to be used in a custom form.io component), so I'd move this to a different place in the codebase.

IMO `src/openforms/contrib/kadaster` is a good candidate - you can implement the `api` package there and wire it up in the main URL routes under `/api/v2/geo/address-search`.

I want to "mask" the fact that we use PDOK/Kadaster APIs for this a bit (it's not a secret), but structuring the API and code this way makes it easier to plug in alternative backends if that need ever becomes relevant.

sergei-maertens (changes requested):

src/openforms/conf/base.py
```diff
@@ -1182,6 +1183,10 @@
     "LOG_OUTGOING_REQUESTS_MAX_AGE", default=7 * 24
 )  # number of hours
 
+# Pdok api locatieserver url
+PDOK_LOCATIE_SERVER_URL = config(
+    "PDOK_LOCATIE_SERVER_URL", "https://api.pdok.nl/bzk/locatieserver/search/v3_1/"
+)
```
sergei-maertens, line 1189: 1. When adding settings, can you please respect the grouping of settings in this file? It now looks like this setting belongs to the django-log-outgong-requests library, which is not the case.

2. I think I'd also rather have this service specified in a run time `zgw_consumers.Service` model (with a default for this URL maybe?) so that it can be changed - many organizations funnel all their API traffic through their gateway, and maybe this also needs to go over the NLX network. We have an existing pattern of configuring external services in the admin, I don't think this warrants deviating from it.

src/openforms/contrib/kadaster/api/serializers.py
```diff
@@ -0,0 +1,19 @@
...
+    lng = serializers.FloatField(label=_("Longitude"))
+
+
+class MapSearchRijksDriehoekSerializer(serializers.Serializer):
+    x = serializers.FloatField(label=_("X"))
+    y = serializers.FloatField(label=_("Y"))
+
+
+class MapSearchSerializer(serializers.Serializer):
+    label = serializers.CharField(label=_("weergave naam"))
```
sergei-maertens, line 17: please use an english text for the label

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+from openforms.submissions.tests.factories import SubmissionFactory
+from openforms.submissions.tests.mixins import SubmissionsMixin
+
+
+@override_settings(LANGUAGE_CODE="en")
+class MapSearchApiTests(SubmissionsMixin, APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.submission = SubmissionFactory.create()
```
sergei-maertens, line 18: please always call the super method so you don't get unexpected behaviour when adding/removing mixins

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def tearDown(self):
+        self._clear_session()
```
sergei-maertens, line 21: why is this needed? I don't think I've seen this in other test cases?

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_query_parameter_utrecht(self, mock_requests):
+        # mock the response
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {
```
sergei-maertens, line 28: this is not how we mock API calls, we use the `requests_mock` library for that

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+        self.assertEqual(
+            body,
+            [
+                {
+                    "label": "Gemeente Utrecht",
+                    "latLng": {"lat": 52.09113798, "lng": 5.0747543},
+                    "rd": {"x": 133587.182, "y": 455921.594},
+                },
+            ],
+        )
```
sergei-maertens, line 72: can you please break up your test itself in blocks my removing whitespace for the setup and assert phase and ensure there's a blank line between each phase?

```
# setup phase
setup1
setup2

# call phase
call_the_thing_you_test

# assert phase
assert1
assert2
```

Sticking to this structure helps in reading tests and figuring out what's actually being tested.

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def test_call_with_no_active_submission(self):
+        url = reverse("api:geo:address-search")
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```
sergei-maertens, line 78:
```suggestion
    def test_call_with_no_active_submission(self):
        url = reverse("api:geo:address-search")

        response = self.client.get(url)

        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```
minimal example of the test structure that makes things more obvious when scanning code

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def test_call_with_no_query_parameter(self):
+        url = reverse("api:geo:address-search")
+
+        self._add_submission_to_session(self.submission)
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
sergei-maertens, line 86:
```suggestion
    def test_call_with_no_query_parameter(self):
        url = reverse("api:geo:address-search")
        self._add_submission_to_session(self.submission)

        response = self.client.get(url)
        
        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_api_get_bag_data_exception(self, mock_requests):
+        mock_requests.side_effect = requests.exceptions.RequestException
```
sergei-maertens, line 90: please use the proper requests mocking library

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+        body = response.json()
+        self.assertEqual(body, [])
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_api_return_something_other_then_200(self, mock_requests):
+        # mock the response
+        mock_response = MagicMock()
+        mock_response.status_code = 400
+        mock_response.json.return_value = {
```
sergei-maertens, line 106: please use the proper requests mocking library

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+        body = response.json()
+        self.assertEqual(body, [])
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_api_return_data_not_having_response(self, mock_requests):
+        # mock the response
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {}
```
sergei-maertens, line 130: please use the proper requests mocking library

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+        body = response.json()
+        self.assertEqual(body, [])
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_api_return_data_not_having_docs(self, mock_requests):
+        # mock the response
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {
```
sergei-maertens, line 148: please use the proper requests mocking library

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+            "response": {
+                "numFound": 205959,
```
sergei-maertens, line 150: non-zero num found but no results? is that a representative test case?

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+        body = response.json()
+        self.assertEqual(body, [])
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_api_return_data_not_having_centroide_ll(self, mock_requests):
+        # mock the response
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {
```
sergei-maertens, line 173: please use the proper requests mocking library

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,248 @@
...
+
+        body = response.json()
+        self.assertEqual(body, [])
+
+    @patch("openforms.contrib.kadaster.api.views.requests")
+    def test_call_with_api_return_data_not_having_centroide_rd(self, mock_requests):
+        # mock the response
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {
```
sergei-maertens, line 214: please use the proper requests mocking library

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+            OpenApiTypes.STR,
+            OpenApiParameter.QUERY,
+            description=_("The search query we send to the pdok locatieserver api."),
+            required=True,
+        )
+    ],
+    responses={
+        200: MapSearchSerializer,
+        status.HTTP_400_BAD_REQUEST: ValidationErrorSerializer,
+        status.HTTP_401_UNAUTHORIZED: ExceptionSerializer,
```
sergei-maertens, line 39: I think we get 403 rather than 401?

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+    },
+)
+class MapSearchView(GenericAPIView):
+    serializer_class = MapSearchSerializer
+    permission_classes = [AnyActiveSubmissionPermission]
+    renderer_classes = [CamelCaseJSONRenderer]
```
sergei-maertens, line 45: why the override from the defaults? Now the browsable API is not functional for development.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def get(self, request: Request, *args, **kwargs):
+        query = request.GET.get("q")
+        if not query:
+            return HttpResponseBadRequest(_("Missing query parameter 'q'"))
```
sergei-maertens, line 50: this does not adhere to the documented HTTP 400 error format - you need to raise a `serializers.ValidationError` for that.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+
+        url = f"{settings.PDOK_LOCATIE_SERVER_URL}free"
+        data = {"q": query}
+
+        try:
+            bag_data = requests.get(url, params=data)
+        except requests.exceptions.RequestException as e:
+            logger.exception(f"couldn't retrieve pdok locatieserver data: {e}")
```
sergei-maertens, line 58: doesn't this also cover connection errors and SSL errors etc.? which would lead to misconfiguration, so you'd want to properly see the necessary information.

You also don't want to interpolate information in error messages like that, especially not with f-strings. the `logger.execption` call already includes the exception information which we can see in Sentry.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+    def get(self, request: Request, *args, **kwargs):
+        query = request.GET.get("q")
+        if not query:
+            return HttpResponseBadRequest(_("Missing query parameter 'q'"))
+
+        url = f"{settings.PDOK_LOCATIE_SERVER_URL}free"
+        data = {"q": query}
+
+        try:
+            bag_data = requests.get(url, params=data)
```
sergei-maertens, line 56:
```suggestion
            bag_response = requests.get(url, params=data)
```
it's not data, it's a requests Response which holds the data + additional metadata

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+            and (response := bag_data.json().get("response"))
+            and (docs := response.get("docs"))
+        ):
+            # no response with docs: return empty
+            return Response(
+                self.serializer_class(
+                    instance=[],
+                    context={"request": request},
+                    many=True,
+                ).data
```
sergei-maertens, line 73: you should use `self.get_serializer(instance=[])` for this, which will call the `get_serializer_context` and `get_serializer_class` methods internally.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+        OpenApiParameter(
+            "q",
+            OpenApiTypes.STR,
+            OpenApiParameter.QUERY,
+            description=_("The search query we send to the pdok locatieserver api."),
+            required=True,
+        )
+    ],
+    responses={
+        200: MapSearchSerializer,
```
sergei-maertens, line 37:
```suggestion
        200: MapSearchSerializer(many=True),
```

The API documentation now makes it look like a single object is returned instead of an array of objects.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+                    f"Malformed centroide_ll in pdok locatieserver response: {doc}"
+                )
+                return
+            try:
+                x, y = re.findall(r"\d+\.\d+", doc.get("centroide_rd", ""))
+            except ValueError:
+                logger.info(
+                    f"Malformed centroide_rd in pdok locatieserver response: {doc}"
+                )
+                return
```
sergei-maertens, line 91: There has to be a better way to obtain this information than rolling our own regexes...?

What's the name of the format called for their geo data? Did you check if there are python gis libraries to parse that? I'd even expect it to be built into django in `django.contrib.gis`.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+            return {
+                "label": doc.get("weergavenaam", ""),
+                "latLng": {"lat": lat, "lng": lng},
+                "rd": {"x": x, "y": y},
```
sergei-maertens, line 95: lat, lng, x and y can be undefined, I think?

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+        locations = [loc for doc in docs if (loc := parse_doc(doc))]
+        return Response(
+            self.serializer_class(
+                instance=locations,
+                context={"request": request},
+                many=True,
+            ).data
+        )
```
sergei-maertens, line 105: see earlier remark

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,105 @@
...
+            required=True,
+        )
+    ],
+    responses={
+        200: MapSearchSerializer,
+        status.HTTP_400_BAD_REQUEST: ValidationErrorSerializer,
+        status.HTTP_401_UNAUTHORIZED: ExceptionSerializer,
+    },
+)
+class MapSearchView(GenericAPIView):
```
sergei-maertens, line 42: I think you can considerably remove boilerplate by using https://open-forms.readthedocs.io/en/stable/developers/backend/core/utils.html#openforms.api.views.mixins.ListMixin so you only have to implement `get_objects` and don't have to deal with the serializer/response boilerplate

src/openforms/formio/api/views.py
sergei-maertens, line 1: These changes can be reverted.

sergei-maertens (changes requested):

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,111 @@
...
+    authentication_classes = ()
+    permission_classes = [AnyActiveSubmissionPermission]
+    serializer_class = MapSearchSerializer
+
+    def get_objects(self):
+        config = KadasterApiConfig.get_solo()
+        if not config.kadaster_service:
+            raise ValidationError(
+                {"invalid_params": _("No service defined for the Kadaster service.")}
+            )
```
sergei-maertens, line 51: We have generic tooling that converts DRF error structures into the format with `invalidParams`, so you never have to worry about that aspect. DRF validation errors take the form of `{field_name: error}`, and that is automatically processed.

In this case, the error is not related to any particular field, so you can do:

```py
raise serializers.ValidationError(_("No backend service defined to retrieve search results."), code="invalid-configuration")
```
DRF should then automatically assign that to `api_settings.NON_FIELD_ERRORS_KEY` (which is `"non_field_errors"`) and it will be handled properly.

Additionally, please don't import `ValidationError` from serializers, but import the serializers module and use `serializers.ValidationError`. This is convention with DRF to not confuse it with Django's built-in `django.core.exceptions.ValidationError`.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,111 @@
...
+
+        query = self.request.GET.get("q")
+        if not query:
+            raise ValidationError({"invalid_params": _("Missing query parameter 'q'")})
```
sergei-maertens, line 55:
```suggestion
            raise ValidationError({"q": _("Missing query parameter 'q'")}, code="required")
```

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,111 @@
...
+
+        kadaster_client = config.kadaster_service.build_client()
+        query_params = {"q": f"{query}"}
```
sergei-maertens, line 58:
```suggestion
        query_params = {"q": query}
```
this f-string is not needed

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,111 @@
...
+                operation_id="free",
+                data=None,
+                url="v3_1/free",
+                method="GET",
+                request_kwargs=dict(
+                    params=query_params,
+                ),
+            )
+        except (ClientError, requests.RequestException):
+            logger.exception("couldn't retrieve pdok locatieserver data")
```
sergei-maertens, line 72:
```suggestion
        try:
            bag_response = kadaster_client.operation(
                operation_id="free",
                data=None,
                url="v3_1/free",
                method="GET",
                request_kwargs=dict(
                    params=query_params,
                ),
            )
        except (ClientError, requests.RequestException):
            bag_response = {}
            logger.exception("Couldn't retrieve PDOK locatieserver data")
```
I'm not a fan of re-assigning variables, it's error-prone when more/other cases need to be handled.

src/openforms/contrib/kadaster/api/views.py
```diff
@@ -0,0 +1,111 @@
...
+                )
+                return
+
+            try:
+                x, y = fromstr(doc.get("centroide_rd"))
+            except (ValueError, TypeError):
+                logger.info(
+                    f"Malformed centroide_rd in pdok locatieserver response: {doc}"
+                )
+                return
```
sergei-maertens, line 97: This can be cleaned up substantially:

1. There is no need for this function to be defined runtime - move it to the module level
2. The GEO data parsing is a repetition that can be its own function

```py
def parse_coordinates(doc: dict, key: str) -> tuple[float, float] | None:
    if not (wkt_value := doc.get(key)):
        return None
    try:
        coord0, coord1 = fromstr(wkt_value)  # we expect a two-tuple, and not any arbitrary shape
    except (ValueError, TypeError):
        logger.info("Malformed %s in location server response: %r", key, doc)
        return None
    return coord0, coord1


@dataclass
class RD:
    x: float
    y: float

    @classmethod
    def fromtuple(cls, value: tuple[float, float]):
        x, y = value
        return cls(x=x, y=y)


@dataclass
class LatLng:
    lat: float
    lng: float

    @classmethod
    def fromtuple(cls, value: tuple[float, float]):
        lat, lng = value
        return cls(lat=lat, lng=lng)


@dataclass
class Location:
    label: str
    lat_lng: LatLng
    rd: RD
    

def process_search_result(doc: dict) -> Location | None:
    label = doc.get("weergavenaam", "")
    lat_lng = parse_coordinates(doc, "centroide_ll")
    x_y = parse_coordinates(doc, "centroid_rd")
    if not lat_lng or not x_y:
        return None
    return Location(
        label=label, 
        lat_lng=LatLng.fromtuple(lat_lng),
        rd=RD.fromtuple(x_y),
    )
```

I have some some doubts about filtering out search results that are missing RD though :thinking: why would those not be displayed?

src/openforms/contrib/kadaster/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class KadasterAPIConfig(AppConfig):
+    name = "openforms.contrib.kadaster"
+    label = "contrib_kadaster"
+    verbose_name = _("Kadaster API")
```
sergei-maertens, line 8:
```suggestion
    verbose_name = _("Kadaster")
```
it's possible we'll do more things than just API stuff

src/openforms/contrib/kadaster/models.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def get_default_kadaster_service():
+    default_service = Service.objects.create(
+        label="Locatieserver",
+        api_type=APITypes.orc,
+        api_root="https://api.pdok.nl/bzk/locatieserver/search/",
+        oas="https://api.pdok.nl/bzk/locatieserver/search/v3_1/openapi.json",
+        auth_type=AuthTypes.no_auth,
+    )
```
sergei-maertens, line 16: this should be a `get_or_create` based on `api_root` - if a service already exists with this API root, this will crash with `IntegrityError` due to a database unique constraint being violated.

src/openforms/contrib/kadaster/models.py
```diff
@@ -0,0 +1,36 @@
...
+    return default_service
+
+
+class KadasterApiConfig(SingletonModel):
+    """
+    global configuration and defaults
```
sergei-maertens, line 22:
```suggestion
    Global configuration and defaults.
```

src/openforms/contrib/kadaster/models.py
```diff
@@ -0,0 +1,36 @@
...
+        label="Locatieserver",
+        api_type=APITypes.orc,
+        api_root="https://api.pdok.nl/bzk/locatieserver/search/",
+        oas="https://api.pdok.nl/bzk/locatieserver/search/v3_1/openapi.json",
+        auth_type=AuthTypes.no_auth,
+    )
+    return default_service
+
+
+class KadasterApiConfig(SingletonModel):
```
sergei-maertens, line 20: please add a custom manager that does a `select_related` on the `kadaster_service` field. Now we always get two queries to check for the service, we can reliably make this a single query :zap:

src/openforms/contrib/kadaster/models.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    global configuration and defaults
+    """
+
+    kadaster_service = models.OneToOneField(
```
sergei-maertens, line 25: This is too broad - I would rename this to `search_service` instead. Kadaster/PDOK has many more services that serve other functionality.

src/openforms/contrib/kadaster/models.py
```diff
@@ -0,0 +1,36 @@
...
+        "zgw_consumers.Service",
+        verbose_name=_("Kadaster API"),
+        on_delete=models.PROTECT,
+        limit_choices_to={"api_type": APITypes.orc},
+        null=True,
```
sergei-maertens, line 30: null=True can be removed because we provide a `default`.

sergei-maertens (changes requested):
Can you do the rename globally? Everything "map-search" should basically become "address-search"

src/openforms/contrib/kadaster/api/views.py
bart-maykin: What do you exactly mean by boiler plating? Because the only thing I can think of when I see that is the spectacular code, which I think is good to have for documentation purposes.

src/openforms/contrib/kadaster/api/views.py
bart-maykin: How can we differentiate between misconfiguration and connection errors? Because I don't see how we can tell a service that is down apart from a service that doesn't exists.  

Is the code that we have now sufficient? Because we are doing the same thing now as in the haalcentraal/client.py

src/openforms/contrib/kadaster/api/tests/test_api_map_search.py
```diff
@@ -0,0 +1,345 @@
...
+        url = reverse("api:geo:address-search")
+
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    @patch(
+        "openforms.contrib.kadaster.api.views.KadasterApiConfig.get_solo",
+        return_value=KadasterApiConfig(kadaster_service=None),
+    )
```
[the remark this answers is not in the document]
bart-maykin: I tried doing this but it didn't work. Thats why I ended up overwriting the patch.

I ended up deleting the test and related code because the model field is required and has a default value so the code would never be hit.

sergei-maertens (approved):

PR #3272, opened 2023-07-19
[#3166] Add tests for HC connection for v1 and v2

### LINKED ISSUE #3166: Implement Haal Centraal V2 config check + tests
From coverage report: https://app.codecov.io/gh/open-formulieren/open-forms/blob/feature%2F2606-add-support/src%2Fopenforms%2Fprefill%2Fcontrib%2Fhaalcentraal%2Fplugin.py#L142

https://github.com/open-formulieren/open-forms/blob/7620a926a4035bb3c2be83aa8d10d1be2cf9613b/src/openforms/prefill/contrib/haalcentraal/plugin.py#L129

This config check method appears to not be properly tested.

It tries to do a `retrieve` operation on an expected 404 endpoint (which was the case in v1), but on V2 the behaviour is unknown at the moment: it could be a 404 still (because the endpoint does not exist), but it could also be a HTTP 405 because V2 uses POST instead of GET.

- [x] Add tests
- [x] Check behaviour in reference implementation/demo environment for v2
- [x] Ensure it supports Haal Centraal personen V2

vaszig:
Fixes #3166

### DIFF
--- a/src/openforms/prefill/contrib/haalcentraal/plugin.py
+++ b/src/openforms/prefill/contrib/haalcentraal/plugin.py
@@ -152,18 +152,21 @@ class HaalCentraalPrefill(BasePlugin):
         )
 
     def check_config(self):
-        try:
-            config = HaalCentraalConfig.get_solo()
-            assert isinstance(config, HaalCentraalConfig)
-            if not config.service:
-                raise InvalidPluginConfiguration(_("Service not selected"))
+        """
+        The purpose of this fuction is to simply check the connection to the
+        service, so we are using dummy data and an endpoint which does not exist.
+        We want to avoid calls to the national registration by using bsn numbers.
+        """
+        config = HaalCentraalConfig.get_solo()
+        assert isinstance(config, HaalCentraalConfig)
+        if not config.service:
+            raise InvalidPluginConfiguration(_("Service not selected"))
 
-            client = config.service.build_client()
-            # FIXME: haal centraal v2 client does not support GET methods
-            client.retrieve("test", "test")
+        client = config.build_client()
+        assert client is not None
+        try:
+            client.make_config_test_request()
         except ClientError as e:
-            if e.args[0].get("status") == 404:
-                return
             raise InvalidPluginConfiguration(
                 _("Client error: {exception}").format(exception=e)
             )
--- a/src/openforms/prefill/contrib/haalcentraal/tests/test_config_check.py
+++ b/src/openforms/prefill/contrib/haalcentraal/tests/test_config_check.py
@@ -1,0 +1,136 @@
+from typing import Literal
+from unittest.mock import patch
+
+from django.test import SimpleTestCase
+from django.utils.translation import gettext as _
+
+import requests_mock
+from zds_client.oas import schema_fetcher
+from zgw_consumers.models import Service
+from zgw_consumers.test import mock_service_oas_get
+
+from openforms.plugins.exceptions import InvalidPluginConfiguration
+from openforms.registrations.contrib.zgw_apis.tests.factories import ServiceFactory
+
+from ....registry import register
+from ..constants import HaalCentraalVersion
+from ..models import HaalCentraalConfig
+
+plugin = register["haalcentraal"]
+
+
+class ConfigCheckTests:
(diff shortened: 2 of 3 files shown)


SilviaAmAm (approved):

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -152,15 +153,29 @@ def get_co_sign_values(
...
-            config = HaalCentraalConfig.get_solo()
-            assert isinstance(config, HaalCentraalConfig)
-            if not config.service:
-                raise InvalidPluginConfiguration(_("Service not selected"))
-
-            client = config.service.build_client()
-            # FIXME: haal centraal v2 client does not support GET methods
-            client.retrieve("test", "test")
+            if config.version == HaalCentraalVersion.haalcentraal13:
+                client.retrieve("ingeschrevenpersonen", "999990676")
```
sergei-maertens, line 165: this really should try to retrieve a resource that doesn't exist with an invalid "BSN".

The problem if you start using "real requests" is that those result in calls to the national registration, which may incur a cost but at least is frowned upon because they are unnecessary calls and there's no legal basis to request the details of a "random" person.

So the intent here is that you check the connection to the service - if you get a 404 that means it's working because the service replied with something useful (rather than a 403, 5XX... which would imply issues in the configuration).

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -152,15 +153,29 @@ def get_co_sign_values(
...
+            else:
+                client.operation(
```
sergei-maertens, line 167: similar remarks as above - you should use a non-existent resource/endpoint here to ensure that no real data is returned, ever.

sergei-maertens (changes requested):

src/openforms/prefill/contrib/haalcentraal/tests/test_config_check.py
```diff
@@ -0,0 +1,138 @@
...
+    def setUp(self):
+        self.service = ServiceFactory.build(
+            api_root="https://personen/api/",
+            oas="https://personen/api/schema/openapi.yaml",
+        )
+        super().setUp()
+
+    def test_undefined_service_raises_exception(self):
+        self.config.service = None
+        self.config.save()
```
sergei-maertens, line 90: you don't need the `save` call I think, because we mock the calls to get the instance and return the in-memory object.

This should in turn allow you to use `SimpleTestCase` instead of `TestCase` as base class, as we don't expect any database access.

src/openforms/prefill/contrib/haalcentraal/tests/test_config_check.py
```diff
@@ -0,0 +1,138 @@
...
```
sergei-maertens, line 122: see above

sergei-maertens (approved):

src/openforms/prefill/contrib/haalcentraal/plugin.py
```diff
@@ -152,17 +153,40 @@ def get_co_sign_values(
...
+                    data={
+                        "type": "RaadpleegMetBurgerservicenummer",
+                        "burgerservicenummer": ["test"],
+                        "fields": [AttributesV2.burgerservicenummer],
+                    },
+                    url="test",
+                    request_kwargs={
+                        "headers": {"Content-Type": "application/json; charset=utf-8"}
+                    },
+                )
```
sergei-maertens, line 187: While working on a suggestion with `match / case`, it dawned on me that we can make this a lot simpler and rely on type checking to protect against future versions:

```suggestion
            client.make_config_test_request()
```

and then implement that method for both versions of the client + add it to `HaalCentraalClient` protocol to enforce future versions to imlement this from the get-go.

I would then implement that for the V1.3 client as:

```python
def make_config_test_request(self):
    try:
        self.retrieve("test", "test")
    except ClientError as e:
        if e.args[0].get("status") == 404:
            return
        raise
```
-> so for the expected 404 the test passes, otherwise it bubbles up to the plugin config check.

sergei-maertens (approved):

PR #3276, opened 2023-07-20
[#3189] Clean up provides_auth types

### LINKED ISSUE #3189: Clean up provides_auth types
https://github.com/open-formulieren/open-forms/pull/3168/files#diff-39c390e8d7412756adc2bb538f768077a055aec6da62873b61c8821d957cc72eR40 introduced a type hint:

```py
provides_auth: list[AuthAttribute] | AuthAttribute | None = None
```
but every plugin only assigns a scalar `AuthAttribute` value. There is some `isinstance` code that checks for string vs. anything else, which can presumably be cleaned up.

It should also be possible to drop the `= None` assignment and use just the type declaration - type checks will then complain about missing assignments in subclasses.

vaszig:
Fixes #3189

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -5970,7 +5970,7 @@ components:
           type: string
           description: The human-readable name for a plugin.
         providesAuth:
-          type: array
+          type: string
           items:
             type: string
             title: Authentication attribute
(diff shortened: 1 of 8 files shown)


SilviaAmAm (commented):
Looks good. I think you might need to regenerate again the OAS? I see that the CI job checking the OAS is failing.

sergei-maertens (approved):

PR #3298, opened 2023-07-26
Implement inp.heeftAlsKinderen StUF-BG request

### LINKED ISSUE #3084: "inp.heeftAlsKinderen" missing from request for npFamilyMembers component

any

Internal: Utr 76, Nopolder 42, Enschede 10

Request to stuf-bf is missing the inp.heeftAlsKinderen in scope. This causes these requests to return nothing, or just the BSN as is indicated in scope of the template.
```xml
    <ns:scope>
        <ns:object StUF:entiteittype="NPS">
            <ns:inp.bsn xsi:nil="true"/>
        </ns:object>
    </ns:scope>
```

CharString:
Closes #3084

### DIFF
--- a/src/stuf/stuf_bg/tests/test_client.py
+++ b/src/stuf/stuf_bg/tests/test_client.py
@@ -1,6 +1,8 @@
 import uuid
+from functools import lru_cache
 from io import BytesIO
 from pathlib import Path
+from random import shuffle
 from unittest.mock import patch
 
 from django.conf import settings
@@ -24,6 +26,15 @@ from stuf.tests.factories import StufServiceFactory
 
 PATH_XSDS = (Path(settings.BASE_DIR) / "src" / "stuf" / "stuf_bg" / "xsd").resolve()
 STUF_BG_XSD = PATH_XSDS / "bg0310" / "vraagAntwoord" / "bg0310_namespace.xsd"
+
+
+@lru_cache
+def _stuf_bg_xmlschema_doc():
+    # don't add the etree.XMLSchema construction to this; it's an object
+    # that contains state like error_log
+
+    with STUF_BG_XSD.open("r") as infile:
+        return etree.parse(infile)
 
 
 class StufBGConfigTests(TestCase):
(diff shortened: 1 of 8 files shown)


src/openforms/formio/components/custom.py
```diff
@@ -106,12 +107,18 @@ class NPFamilyMembers(BasePlugin):
...
-        }
-        config = FamilyMembersTypeConfig.get_solo()
-        return handlers[config.data_api]
+        match FamilyMembersTypeConfig.get_solo().data_api:
+            case FamilyMembersDataAPIChoices.haal_centraal:
+                return get_np_children_haal_centraal
+            case FamilyMembersDataAPIChoices.stuf_bg:
+                return get_np_children_stuf_bg
+            case "":
+                logger.critical("No Family Members Data API configured")
```
sergei-maertens, line 116:
```suggestion
                logger.warning("No Family Members Data API configured")
```
I'd rather see this being checked in the configuration overview page, these logger statements are only informative for us and I don't have the illusion that anyone will pro-actively act on them. So let's not create unhandled noise in Sentry, as it'll take away attention from real crashes/errors.

src/openforms/formio/components/custom.py
```diff
@@ -106,12 +107,18 @@ class NPFamilyMembers(BasePlugin):
 
     @staticmethod
     def _get_handler() -> Callable[[str], list[tuple[str, str]]]:
-        handlers = {
-            FamilyMembersDataAPIChoices.haal_centraal: get_np_children_haal_centraal,
-            FamilyMembersDataAPIChoices.stuf_bg: get_np_children_stuf_bg,
-        }
-        config = FamilyMembersTypeConfig.get_solo()
-        return handlers[config.data_api]
+        match FamilyMembersTypeConfig.get_solo().data_api:
```
sergei-maertens, line 110: This requires backporting to Open Forms 2.0.x which is on Python 3.8, for which this syntax won't work.

I think a more robust approach here would be to use StUF-BG as fallback if nothing is explicitly configured, as that is what the bulk of the organizations use (only a few large municipalities are early adopters of Haal Centraal).

So to address the key error: `handlers.get(config.data_api, get_np_children_stuf_bg)`

I don't see how any other values than one of the `FamilyMembersDataAPIChoices` and the empty value `""` would end up in the database. And to ensure that every choice is present in the mapping, I'd rather see a module-level developer helper-check along the lines of:

```python
for value in FamilyMembersDataAPIChoices.values:
    assert value in HANDLER_MAPPING
```

Personally I also don't find that the match statement helps with readability, a mapping lookup is much flatter and simpler to process IMO.

src/openforms/formio/components/custom.py
```diff
@@ -106,12 +107,18 @@ class NPFamilyMembers(BasePlugin):
...
+            case FamilyMembersDataAPIChoices.haal_centraal:
+                return get_np_children_haal_centraal
+            case FamilyMembersDataAPIChoices.stuf_bg:
+                return get_np_children_stuf_bg
+            case "":
+                logger.critical("No Family Members Data API configured")
+            case unknown:  # catch all
+                logger.critical(
+                    f'Invalid Family Members Data API "{unknown}" configured'
```
sergei-maertens, line 119: don't use f-strings for log records, but use a proper interpolation format (and generally this implies old-style string formatting). That way, Sentry will group the errors by log message template instead of creating separate items for every one of them.

src/stuf/stuf_bg/tests/test_client.py
```diff
@@ -185,3 +185,40 @@ def test_errors_are_not_swallowed(self):
...
+            doc.getroot()
+            .xpath(
+                "soap:Body",
+                namespaces={"soap": "http://schemas.xmlsoap.org/soap/envelope/"},
+            )[0]
+            .getchildren()[0]
+        )
+
+        # validate soap message with xsd
+        xmlschema.assert_(soap_body)
```
sergei-maertens, line 212: This would assert the order of elements is correct, right?

CharString: I've added the test and reworded things so these things can become apparent from reading the test suite.

PR #3301, opened 2023-07-26
Handle DigiD SAML Artefact with missing sector code

### LINKED ISSUE #3278: DigiD sector code is not provided by Signicat

2.2.0

I noticed that the response processed in this line: https://github.com/open-formulieren/open-forms/blob/e3291c302bb5082d67151649a76cb05433bf9b0b/src/openforms/authentication/contrib/digid/views.py#L67

Throws an exception because Signicat did not pass the sectorial code. Let's make this more resillient: If it's not splittable, we assume its the BSN.

In #3135 I reported this already but dismissed it since Signicat indicated its not a problem in "real" test and production environments. Testing with organizations, I'm not so sure about this. So, let's accommodate for this common scenario.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

CharString:
Closes #3278

The Signicat borker [sic!] has shown to send nameid's in artefact
responses that don't contain the sector code for BSN (even in their production
environment)

We assume BSN there... I don't know what this means for the 500 people
named F. Vogelpest in Heerlen that share the same SOFI.

### DIFF
--- a/src/openforms/authentication/contrib/digid/views.py
+++ b/src/openforms/authentication/contrib/digid/views.py
@@ -86,13 +86,15 @@ class DigiDAssertionConsumerServiceView(
             )
             return HttpResponseRedirect(failure_url)
 
-        sector_code, sectoral_number = name_id.split(":")
-
-        # We only care about users with a BSN.
-        if sector_code != SectorType.bsn:
-            raise BSNNotPresentError
-
-        bsn = sectoral_number
+        match name_id.split(":"):
+            case [SectorType.bsn, bsn]:
+                pass
+            case [bsn]:
+                # Sectorcode missing; assume BSN as SOFI aren't issued since
+                # Aanpassingswet Brp in 2014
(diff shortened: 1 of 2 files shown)


src/openforms/authentication/contrib/digid/views.py
```diff
@@ -86,13 +86,15 @@ def get(self, request):
...
-        bsn = sectoral_number
+        match name_id.split(":"):
+            case [SectorType.bsn, bsn]:
+                pass
+            case [bsn]:
+                # Sectorcode missing; assume BSN as SOFI aren't issued since
+                # Aanpassingswet Brp in 2014
+                pass
+            case _:
+                raise BSNNotPresentError()
```
sergei-maertens, line 97: for the record - this line was not originally covered to begin with

sergei-maertens (approved):

PR #3303, opened 2023-07-26
Fixed race condition between celery worker and DB transaction

### LINKED ISSUE #3302: Uploaded image resizing fails due to race condition
Sentry: 311035
Taiga: HodM 56

The celery task is scheduled in the regular python code and the worker picks up the task before the transaction is comitted.

It looked like #507 for a while, but there's a different cause.

sergei-maertens:
Fixes #3302

Prevent the celery worker from picking up the attachment image resize task before the DB transaction is committed. Celery workers run in different processes and thus different database transactions. The database record must be committed before the worker can look it up to start processing.

### DIFF
--- a/src/openforms/submissions/attachments.py
+++ b/src/openforms/submissions/attachments.py
@@ -11,6 +11,7 @@ from urllib.parse import urlparse
 from django.conf import settings
 from django.core.files.temp import NamedTemporaryFile
 from django.core.files.uploadedfile import UploadedFile
+from django.db import transaction
 from django.urls import Resolver404, resolve
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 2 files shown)


CharString (approved):
I think I'd use `lambda: ...` for the delayed application; `partial` is reserved in my mind when it's intended to receive more parameters.

sergei-maertens:
Backport commits:

* 2.2.x: 408ea99fe37b7699355600a86a4a31865eca8c07
* 2.1.x: d8f81abb842722cae9048b5dc985b31e57a2717c
* 2.0.x: 3aa653ddca2e66d74a97478e7314acbcc4eb3060

PR #3305, opened 2023-07-26
Upgrade base image

sergei-maertens:
Builds fine locally

Depends on #3304


CharString (approved):
Can you upgrade my machine too? I'm running a kernel from backports :wink:

Issue #3242, opened 2023-07-10
Figure out test flakiness


sergei-maertens:
Tests on Github CI (and sometimes) locally fail in unpredictable ways (because they exceed the hypothesis deadline).

This issue tracks the ideas and information to resolve it.

Known problematic tests:

* [ ] ...

CharString:
Note: forgetting a `temp_private_root` decorator, will leave the file system in a dirty state, which could be a source of flakeyness.

Maybe we can run some checks in CI after the test suite passed to see if the tests as  whole left remnants.

sergei-maertens:
Seems to mostly been resolved now. We can create a new issue for new tests that pop up.

PR #3307, opened 2023-07-26
[#3242] Profile more failing tests

sergei-maertens:
And tweak the profiling output to declutter the stdout output in CI.

### DIFF
--- a/src/openforms/tests/utils.py
+++ b/src/openforms/tests/utils.py
@@ -4,6 +4,8 @@ import io
 import os
 import pstats
 import socket
+import sys
+from pathlib import Path
 from pstats import SortKey
 
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


src/openforms/tests/utils.py
```diff
@@ -55,14 +57,29 @@ def supress_output(stdchannel, dest_filename):
...
+    # On github CI, write to the step summary
+    github_summary = os.getenv("GITHUB_STEP_SUMMARY")
+    if not github_summary:
+        yield sys.stdout
+    else:
+        with Path(github_summary).open("a") as outfile:
+            outfile.write("```\n")
+            yield outfile
+            outfile.write("```")
+            outfile.write("\n\n")
```
SilviaAmAm, line 71: It is not too clear to me how we see the output data. If we set the `GITHUB_STEP_SUMMARY` var, it will be written to a file, otherwise it would endup in the standard output. But how do we see this file? I am probably missing something :smile:

sergei-maertens: if you click on the "summary" in github actions in the top left, it's on the page!

image

PR #3314, opened 2023-07-27
Cleanup StUF-BG tests a bit

CharString:
- assert validity of mock response
- 🔥 remove test_address, moved useful assertions elsewhere
  - it wasn't testing getting an address
  - it was patching and mocking things it wasn't asserting
  - it was asserting the same thing twice
- test the Client.get_values which is the actual api of the class

TODO:

- do the same for the other xml responses in tests/responses
- test the failure modes of get_values
- rename get_values_for_attributes to \_get_values_for_attributes
  if it's really just a private helper

### DIFF
--- a/src/openforms/prefill/contrib/stufbg/plugin.py
+++ b/src/openforms/prefill/contrib/stufbg/plugin.py
@@ -119,6 +119,13 @@ class StufBgPrefill(BasePlugin):
             if value and "@noValue" not in value:
                 response_dict[attribute] = value
 
+        # postcodes in StUF BG responses have this form "[1-9][0-9]{3}[A-Z]{0,2}"
+        # Our prefill tests expect roughly "[1-9][0-9]{3} [A-Z]{0,2}"
+        if FieldChoices.postcode in response_dict and " " not in (
+            postcode := response_dict[FieldChoices.postcode]
+        ):
+            response_dict[FieldChoices.postcode] = postcode[:4] + " " + postcode[4:]
+
         return response_dict
 
     def get_identifier_value(
--- a/src/stuf/stuf_bg/templates/stuf_bg/tests/responses/StufBgResponse.xml
+++ b/src/stuf/stuf_bg/templates/stuf_bg/tests/responses/StufBgResponse.xml
@@ -20,9 +20,10 @@
                 </StUF:ontvanger>
                 <StUF:referentienummer>{{ referentienummer }}</StUF:referentienummer>
                 <StUF:tijdstipBericht>{{ tijdstip_bericht }}</StUF:tijdstipBericht>
+                <StUF:entiteittype>NPS</StUF:entiteittype>
             </ns:stuurgegevens>
             <ns:parameters>
-                <StUF:indicatorAfnemerIndicatie>false</StUF:indicatorAfnemerIndicatie>
+                <StUF:indicatorVervolgvraag>false</StUF:indicatorVervolgvraag>
             </ns:parameters>
             <ns:antwoord>
                 <ns:object StUF:entiteittype="NPS">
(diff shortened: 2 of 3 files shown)


src/openforms/prefill/contrib/stufbg/plugin.py
```diff
@@ -119,6 +119,13 @@ def _get_values_for_bsn(
             if value and "@noValue" not in value:
                 response_dict[attribute] = value
 
+        # postcodes in StUF BG responses have this form "[1-9][0-9]{3}[A-Z]{0,2}"
+        # Our prefill tests expect roughly "[1-9][0-9]{3} [A-Z]{0,2}"
+        if FieldChoices.postcode in response_dict and " " not in (
+            postcode := response_dict[FieldChoices.postcode]
+        ):
+            response_dict[FieldChoices.postcode] = postcode[:4] + " " + postcode[4:]
```
sergei-maertens, line 127: There is a bit more nuance to this - I have no idea if it conflicts or not, but we run normalizers for the Formio data fields - see `openforms.formio.components.custom.Postcode.normalizer`

Though that also makes me think - we deprecated the custom postcode component in favour of a pre-configured text field, and that one probably does not run this normalizer. The functionality there is generic enough to conform the value to the mask. I'll have to create a ticket!

src/stuf/stuf_bg/templates/stuf_bg/tests/responses/StufBgResponse.xml
```diff
@@ -20,9 +20,10 @@
                 </StUF:ontvanger>
                 <StUF:referentienummer>{{ referentienummer }}</StUF:referentienummer>
                 <StUF:tijdstipBericht>{{ tijdstip_bericht }}</StUF:tijdstipBericht>
+                <StUF:entiteittype>NPS</StUF:entiteittype>
             </ns:stuurgegevens>
             <ns:parameters>
-                <StUF:indicatorAfnemerIndicatie>false</StUF:indicatorAfnemerIndicatie>
+                <StUF:indicatorVervolgvraag>false</StUF:indicatorVervolgvraag>
```
sergei-maertens, line 26: I can't judge the correctness of this and also don't see more details in your commit message. Did you get this from the XSDs?

src/stuf/stuf_bg/tests/test_client.py
```diff
@@ -37,98 +34,73 @@ def _stuf_bg_xmlschema_doc():
...
-        self.config.service = self.service
-        self.config.save()
-        self.client = self.config.get_client()
-
-    @freeze_time("2020-12-11T10:53:19+01:00")
-    @patch(
-        "stuf.client.uuid.uuid4",
-        return_value=uuid.UUID("38151851-0fe9-4463-ba39-416042b8f406"),
+def _extract_soap_body(full_doc: str):
+    doc = etree.parse(BytesIO(full_doc))
```
sergei-maertens, line 38: note that there's also `stuf.xml.fromstring` which uses a defused parser and might make it easier to deal with the BytesIO stuff? It takes str and bytes

src/stuf/stuf_bg/tests/test_client.py
```diff
@@ -37,98 +34,73 @@ def _stuf_bg_xmlschema_doc():
...
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+        cls.config = StufBGConfig.get_solo()
+        cls.config.service = StufServiceFactory.create(soap_service__soap_version="1.1")
+        cls.config.save()
+
+    def setUp(self):
+        super().setUp()
+        self.client = self.config.get_client()
```
sergei-maertens, line 100: looking at the `StufBGClient` class itself, I don't see any use of `StufBGConfig` in there. It seems to me that we can bypass that entirely?

```py
def setUp(self):
    service = StufServiceFactory.build(soap_service__soap_version="1.1")
    self.client = StufBGClient(service)
```

combining that with a separate test that asserts that `StufBGConfig().get_client()` returns a client instance seems sufficient to me (you already have a test for the misconfiguration case)

src/stuf/stuf_bg/templates/stuf_bg/tests/responses/StufBgResponse.xml
CharString: From stuf0301.xsd
```xml
	<complexType name="ParametersAntwoordSynchroon">
		<annotation>
			<documentation>Type voor gebruik in La01, La07, La09, La11 en La13 berichten.</documentation>
		</annotation>
		<complexContent>
			<restriction base="StUF:ParametersAntwoordSynchroonFormeel">
				<sequence>
					<element name="indicatorVervolgvraag" type="boolean"/>
					<element name="indicatorAfnemerIndicatie" type="boolean" default="false" minOccurs="0"/>
					<element name="aantalVoorkomens" type="StUF:AantalVoorkomens" minOccurs="0"/>
				</sequence>
			</restriction>
		</complexContent>
	</complexType>
```

src/stuf/stuf_bg/tests/test_client.py
```diff
@@ -142,71 +114,97 @@ def test_order_of_attributes_in_template_is_correct(self):
...
+            m.post(self.client.service.soap_service.url)
+            # perform request
+            self.client.get_values_for_attributes(test_bsn, FieldChoices.values)
+
+        self.assertEqual(
+            TimelineLogProxy.objects.filter(
+                template="logging/events/stuf_bg_request.txt"
+            ).count(),
+            1,
+        )
```
sergei-maertens, line 152: perhaps we should make a mixin/helper for this in `openforms.utils.tests.logging`, but I've been writing these a bit different lately:

```py
log_events = {log.event for log in TimelineLogProxy.objects.all()}
self.assertIn("stuf_bg_request", log_events)
```
In many of these assertions the exact count doesn't really matter too much, but you could easily turns this into a list comprehension with count checks too. Some helpers like `self.assertLogEvent("stuf_bg_request", count=None)` could be useful?

CharString: Yes or something with a similar API to TestCase.assertLogs
```
with self.assertLogsToTimeLine(tag=..., event=..., count=...):
   ...
```

CharString: Or maybe the count is implied, and the context manager records the count on `__enter__` and asserts it is greater on `__exit__`.

CharString: Let's let this one simmer a bit. I've had a quick glance of the assertions made across tests, but I'm still a bit hazy about how timelinelogger does its thing. I don't want to slow down our suite with queries for assertions that belong in the library. DontTestThePlatform™

sergei-maertens: sure - simmering is good

w/r to the implied count - not necessarily, e.g. the registration task retry machinery would explicitly test for multiple log entries since every retry needs to log its attempt

sergei-maertens (approved):

PR #3317, opened 2023-07-28
[#3127] Generate Prefill attributes from OAS specifications

### LINKED ISSUE #3127: Update generate_prefill_from_spec command
Update the generate_prefill_from_spec command to be fully compatible with oas 3.0.0 and 3.0.3

SilviaAmAm:
Partly Fixes #3127 

I didn't re-implement yet all the features that the old management command (`generate_prefill_from_spec`) implements. 
What still should be done:
- Specify a service and then generate the attributes from the OAS of this service
- Specify a path and then only generate the attributes for this path
- If no path/schema are specified, print all paths and schemas in the OAS
- Option to add '_embedded.' in paths for $refs
- Keep track of the titles in addition to the properties names in the schema

Note: without https://github.com/maykinmedia/openapi3-parser/pull/1, this will fail for OAS that have properties not specifying their type

### DIFF
--- a/src/openforms/prefill/attributes_generator.py
+++ b/src/openforms/prefill/attributes_generator.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
+from typing import Iterator
 
 from openapi_parser import parse
+from openapi_parser.specification import Object, Property, Specification
 
 from openforms.template import render_from_string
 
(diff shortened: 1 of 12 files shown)


sergei-maertens:
Blocked on https://github.com/manchenkoff/openapi3-parser/issues/61

SilviaAmAm:
Discussed:
- Move all the code in a different class
- Add option to the old command to chose which parser you want to use
- Remove new command

src/openforms/prefill/attributes_generator.py
```diff
@@ -0,0 +1,122 @@
...
+    def get_specifications(self):
+        if self.url:
+            # The content types supported in the openapi3-parser are these:
+            # https://github.com/manchenkoff/openapi3-parser/tree/master/src/openapi_parser/home/silvia/repositories/openapi3-parser/src/openapi_parser/enumeration.py
+            # However, in the haal centraal OAS we find content types such as "application/hal+json". So with
+            # strict_enum=False, the parser just checks that the content type is a string
+            return parse(uri=self.url, strict_enum=False)
+
+        raise AttributeGeneratorException(
+            "No URL to the OAS specifications was provided."
```
sergei-maertens, line 51:
```suggestion
            "No URL to the OAS specification was provided."
```
is there a reason the plural form is used everywhere?

src/openforms/prefill/attributes_generator.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+    def get_attributes(self, property_schema):
```
sergei-maertens, line 54:
```suggestion
    def get_attributes(self, property_schema):
```
I'd appreciate type hints to make it easier to reason about what `property_schema` is exactly :grimacing:

src/openforms/prefill/attributes_generator.py
```diff
@@ -0,0 +1,122 @@
...
+                            [attribute_bit for attribute_bit in attribute]
+                        ),
+                        "attribute_label": " > ".join(
+                            [attribute_bit.capitalize() for attribute_bit in attribute]
+                        ),
+                    }
+                )
+        return formatted_attributes
+
+    def generate_attributes(self):
```
sergei-maertens, line 106: it looks like this is the only public method, so I'd mark the other ones private with an underscore.

src/openforms/prefill/attributes_generator.py
```diff
@@ -0,0 +1,122 @@
...
+
+            return [
+                self.get_attributes(schema_property)
+                for schema_property in parsed_schemas[self.schema].properties
+            ]
+
+        raise AttributeGeneratorException("No schema specified")
+
+    def format_attributes(self, attributes):
+        """Turn each attribute in ``{"attribute_variable_name": "bla_bla", "attribute_value": "bla.bla", "attribute_label": "Bla > Bla"}``"""
```
sergei-maertens, line 85: Can you pay a bit more attention to the docstring hygiene, following [PEP 257](https://peps.python.org/pep-0257/)

Essentially, the summary should fit one line (so keeping a max length of ~ 80 chars is a good guideline), followed by a more elaborate explanation.

```suggestion
        """
        Prepare attributes for code generation.
        
        Each attribute is prepared by:
        
        * using snake_case to create valid variable names/identifiers
        * creating a string value with periods as path separators
        * capitalize each part for the human readable label
       """
```

src/openforms/prefill/tests/files/personen.yaml
```diff
@@ -0,0 +1 @@
+../../contrib/haalcentraal/tests/files/personen.yaml
```
sergei-maertens, line 1: symlink?

src/openforms/prefill/tests/test_attributes_generation.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        generator = OpenApi3AttributesGenerator(
+            schema="Datum",
+            url="https://personen/api/schema/openapi.yaml?v=3",
+            command="test",
+        )
+
+        output = generator.generate_attributes()
+
+        expected_output = """
```
sergei-maertens, line 72: you could leverage `textwrap.dedent` to keep the code nicely indented?

src/openforms/prefill/tests/test_management_commands.py
```diff
@@ -38,3 +41,45 @@ def test_list_plugins(self, *mocks):
...
+        call_command(
+            "generate_prefill_from_spec",
+            parser="openapi3-parser",
+            schema="Datum",
+            url="https://personen/api/schema/openapi.yaml?v=3",
+            stdout=stdout,
+        )
+
+        stdout.seek(0)
+        output = stdout.read()
```
sergei-maertens, line 65: I think you can do `stdout.getvalue()` which is equivalent

sergei-maertens (commented):
I think you can make this code a bit simpler and better typed with a different approach using iterators and data. You can pass an iterator to the template engine and loop over it as usual, that'll work without issues.

Untested alternative implementation (without docstrings or any of that :P )

```py
from dataclasses import dataclass
from openapi_parser import parse
from openapi_parser.specification import Specification, Object, Property
from typing import Iterator


class AttributeGeneratorException(Exception):
    pass


@dataclass
class Attribute:
    parts: list[str]

    @property
    def variable_name(self) -> str:
        return "_".join(part.lower() for part in self.parts).lstrip("_")

    @property
    def value(self) -> str:
        return ".".join(self.parts)

    def label(self) -> str:
        return " > ".join(part.capitalize() for part in self.parts)


@dataclass
class OpenApi3AttributesGenerator:
    url: str = ""
    schema: str = ""

    def generate_attributes(self) -> str:
        spec = self.load_specification()
        attributes = self._get_schema_attributes(spec)
        return render_from_string(...)

    def load_specification(self) -> Specification:
        if not self.url:
            raise AttributeGeneratorException("No URL to the OAS specification was provided.")
        return parse(uri=self.url, strict_enum=False)

    def _get_schema_attributes(self, spec: Specification) -> Iterator[Attribute]:
        if not self.schema:
            raise AttributeGeneratorException("No schema specified")
        
        if (schema := spec.schemas.get(self.schema)) is None:
            raise AttributeGeneratorException(
               f'The given schema name "{self.schema}" could not be found in the specifications'
           )
        assert isinstance(schema, Object), f"Expected an object schema, got {type(schema)}."
        for schema_property in schema.properties:
            yield from self._get_attributes(schema_property)

    def _get_attributes(self, property: Property, prefix: str = "") -> Iterator[Attribute]:
        if isinstance(property.schema, Object):
            for nested in property.schema.properties:
                yield from self._get_attributes(nested, prefix=nested.name)
        else:
            parts = ([prefix] if prefix else []) + [property.name]
            yield Attribute(parts)
```

The template would then loop over the `Attribute` instances from the generator and you can do:

```django
{{ attribute.variable_name }} = "{{ attribute.value }}", _("{{ attribute.label }}")
```

src/openforms/prefill/tests/files/personen.yaml
SilviaAmAm: Yes, to reuse the same OAS that we use for testing in Haal Centraal... I thought it might be better than copying it over in this folder?

sergei-maertens: yeah it was a question to see if I interpreted this correctly, this is fine!

src/openforms/prefill/attributes_generator.py
```diff
@@ -0,0 +1,130 @@
...
+        if (schema := oas_specification.schemas.get(schema_name)) is None:
+            raise AttributeGeneratorException(
+                f'The given schema name "{schema_name}" could not be found in the specifications'
+            )
+
+        assert isinstance(
+            schema, Object
+        ), f"Expected an object schema, got {type(schema)}."
+        for schema_property in schema.properties:
+            yield from self._get_property_attributes(schema_property)
```
sergei-maertens, line 119: looking at this method and also `_get_property_attributes`, none of them seem to actually be using `self`, so I'm inclined to make them `@staticmethod`, but that breaks the `self._get_property_attributes` invocation.

Shouldn't we just use plain functions instead of methods then?

src/openforms/prefill/tests/test_attributes_generation.py
```diff
@@ -0,0 +1,130 @@
...
+        ):
+            generator.generate_attributes()
+
+    @Mocker()
+    def test_wrong_schema_provided(self, m_request_oas):
+        m_request_oas.get(
+            "https://personen/api/schema/openapi.yaml?v=3",
+            status_code=200,
+            content=load_binary_mock("personen.yaml"),
+        )
```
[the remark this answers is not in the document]
SilviaAmAm: I also realised that if you use the imported `load_binary_mock`, it uses the path in the `haalcentraal` folder. I added it to some util file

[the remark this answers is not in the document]
sergei-maertens: I don't understand? If you use `zgw_consumers.test.mock_service_oas_get`, you don't need any `load_binary_mock`. This `load_binary_mock` is already duplicated twice in master, with this PR this would be a third time. In this situation, for downloading the Open API spec, there is already existing tooling in the zgw-consumers package that does exactly that. You then also don't need the symlink, because it's configured in the settings which folders to scan for the file.

And actually, reasoning more about it - can't we just use file paths directly when testing the generator itself, rather than hitting the network and/or dealing with network mocks?

```py
parse("./src/openforms/prefill/contrib/haalcentraal/tests/files/personen.yaml")
```
works fine for me locally.

Even for the management command you can pass in the filepath as URI and things should work as expected :)

sergei-maertens (approved):

PR #3329, opened 2023-08-01
:lock: [#3328] Use mtls when certs are configured for Qmatic

### LINKED ISSUE #3328: Qmatic plugin does not use mtls

2.3.0-alpha.0

Certificate can be provided but is not sent along to the server.

Qmatic itself does not use this but when an organization uses a gateway, it can require a certificate.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

joeribekker:
Fixes #3328

Still requires a test, maybe move cert to generic REST-service feature, and maybe also requires server side validation...

### DIFF
--- a/src/openforms/appointments/contrib/qmatic/client.py
+++ b/src/openforms/appointments/contrib/qmatic/client.py
@@ -15,6 +15,20 @@ class QmaticClient(Session):
 
     _config: QmaticConfig | None = None
 
+    def _get_cert(self) -> None | str | tuple[str, str]:
+        if not self._config:
+            self._config = QmaticConfig.get_solo()
+
+        certificate = self._config.service.client_certificate
+        if not certificate:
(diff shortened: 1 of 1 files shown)


sergei-maertens:
This should indeed be tackled in a generic way in one of the libraries, this code already lives in `stuf.client` and soap/stuf related services, but it applies to any service we're accessing using `requests`.

Some utility that abstracts away this method should be used, so you can just feed the kwargs to requests:

```py
    def _get_auth_kwargs(self):
        cert = self.service.get_cert()
        auth = self.service.get_auth()
        verify = self.service.get_verify()
        return {"cert": cert, "auth": auth, "verify": verify}
```

joeribekker:
Refinement: Closed in favor of newer PR.

PR #3330, opened 2023-08-01
Update confirmation email sending for new appointments

sergei-maertens:
Closes #3321

- The form summary rendering is now empty for appointment forms (there are no steps or formio forms to render)
- The contact details submitted (dynamic, depending on the selected product(s)) are now included
- The setting to not send a confirmation email is respected
- Restructured some code to move more responsibility into the appointments app/module

Issue #3321, opened 2023-07-31
Ensure the appointment confirmation email contains all the necessary data and is sent out


joeribekker:
- [x] Figure out correct email address
- [x] Ensure that appointment details are (still) present
- [x] Ensure that the contact details are shown
- [x] Generic form summary block may not output anything
- [x] Ensure that the correct amount of products is shown

PR #3331, opened 2023-08-01
Correctly report product amount when fetching JCC appointment details

sergei-maertens:
Depends on #3330, related to #3321

Ensure that we correctly interpret the product count when retrieving appointment details from JCC.

Implemented based on documentation, as I'm still waiting for support/help from JCC to figure out the appointment booking issues.

### DIFF
--- a/src/openforms/appointments/contrib/jcc/plugin.py
+++ b/src/openforms/appointments/contrib/jcc/plugin.py
@@ -360,7 +360,7 @@ class JccAppointment(BasePlugin):
             for product_id, count in product_ids.items():
                 _product = client.service.getGovProductDetails(productID=product_id)
                 product = Product(
-                    identifier=product_id, name=_product.description, amount=count
+                    identifier=product_id, name=_product.description or "", amount=count
                 )
                 app_products.append(product)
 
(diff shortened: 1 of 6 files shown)


src/openforms/appointments/contrib/jcc/plugin.py
```diff
@@ -340,10 +356,13 @@ def get_appointment_details(self, identifier: str) -> AppointmentDetails:
...
             app_products = []
-            for pid in details.productID.split(","):
-                product = client.service.getGovProductDetails(productID=pid)
-
-                app_products.append(Product(identifier=pid, name=product.description))
+            product_ids = Counter(details.productID.split(","))
+            for product_id, count in product_ids.items():
+                _product = client.service.getGovProductDetails(productID=product_id)
+                product = Product(
+                    identifier=product_id, name=_product.description, amount=count
```
SilviaAmAm, line 363: I see in the WSDL in the mocks, that the description can appear also "zero" times. Would this cause an error in this case?

sergei-maertens: I'll try it!

sergei-maertens: Added and fallback to empty string, it returned `None` otherwise (which doesn't fit the type definitions)

Issue #3332, opened 2023-08-02
Minor appointment issues


joeribekker:
Feedback

* [x] #3333
* [x] Retrieve 4 weeks instead of 2 weeks  -- Taiga HadM 59
* [x] If you choose 2 products, you can no longer choose a location? -- Taiga HadM 61
* [x] Timezone information is lost in JCC. Appointments for 11.50 are registered as 9.50 -- Taiga HadM 62

PR #3334, opened 2023-08-02
Use naive datetimes for JCC

sergei-maertens:
Bullet 4 in #3332

### DIFF
--- a/src/openforms/appointments/contrib/jcc/plugin.py
+++ b/src/openforms/appointments/contrib/jcc/plugin.py
@@ -276,6 +276,11 @@ class JccAppointment(BasePlugin):
             FIELD_TO_XML_NAME[key]: value for key, value in client.details.items()
         }
 
+        # ensure start_at is naive, as JCC does not handle ISO-8601 datetimes with TZ
+        # information.
+        if not timezone.is_naive(start_at):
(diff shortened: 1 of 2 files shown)


src/openforms/appointments/contrib/jcc/plugin.py
```diff
@@ -276,6 +276,11 @@ def create_appointment(
             FIELD_TO_XML_NAME[key]: value for key, value in client.details.items()
         }
 
+        # ensure start_at is naive, as JCC does not handle ISO-8601 datetimes with TZ
+        # information.
+        if not timezone.is_naive(start_at):
+            start_at = timezone.make_naive(start_at, timezone=TIMEZONE_AMS)
```
CharString, line 282: Shouldn't this be set to `settings.TIME_ZONE`?

sergei-maertens: that is the (implicit) default in Django, but we've had projects where we configured this django setting to be UTC which would yield the wrong results in this case.

JCC itself communicates all datetimes in the Dutch timezone, which is `Europe/Amsterdam`, so our code interfacing with JCC must be pinned to that value. settings.TIME_ZONE also being `Europe/Amsterdam` is coincidence.

PR #3340, opened 2023-08-04
Make the base registry class a generic type

sergei-maertens:
This uncovered a couple of bugs too where plugins were being yielded rather than their identifiers.

### DIFF
--- a/src/openforms/authentication/registry.py
+++ b/src/openforms/authentication/registry.py
@@ -1,3 +1,4 @@
+import logging
 from typing import TYPE_CHECKING, Iterator, List, Optional
 
 from django.http import HttpRequest
@@ -8,6 +9,8 @@ if TYPE_CHECKING:  # pragma: no cover
     from openforms.forms.models import Form
 
     from .base import BasePlugin, LoginInfo  # noqa: F401
+
+logger = logging.getLogger(__name__)
 
 
 def _iter_plugin_ids(form: Optional["Form"], registry: "Registry") -> Iterator[str]:
@@ -31,6 +34,11 @@ class Registry(BaseRegistry["BasePlugin"]):
         options = list()
         for plugin_id in _iter_plugin_ids(form, self):
             if plugin_id not in self._registry:
+                logger.warning(
+                    "Plugin %s not found in registry, ignoring it.",
+                    plugin_id,
+                    extra={"plugin_id": plugin_id, "form": form.pk if form else None},
+                )
                 continue
             plugin = self._registry[plugin_id]
             info = plugin.get_login_info(request, form)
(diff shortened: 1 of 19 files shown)


src/openforms/authentication/registry.py
```diff
@@ -1,28 +1,40 @@
...
+    from .base import BasePlugin, LoginInfo  # noqa: F401
 
-class Registry(BaseRegistry):
+
+def _iter_plugin_ids(form: Optional["Form"], registry: "Registry") -> Iterator[str]:
+    if form is not None:
+        yield from form.authentication_backends
+    else:
+        for plugin in registry.iter_enabled_plugins():
+            yield plugin.identifier
```
sergei-maertens, line 18: coverage

src/openforms/authentication/registry.py
```diff
@@ -1,28 +1,40 @@
...
+    ) -> List["LoginInfo"]:
         options = list()
-        plugins = form.authentication_backends if form else self.iter_enabled_plugins()
-        for plugin_id in plugins:
-            if plugin_id in self._registry:
-                plugin = self._registry[plugin_id]
-                info = plugin.get_login_info(request, form)
-                options.append(info)
+        for plugin_id in _iter_plugin_ids(form, self):
+            if plugin_id not in self._registry:
```
sergei-maertens, line 33: coverage

src/openforms/formio/rendering/registry.py
```diff
@@ -1,19 +1,23 @@
-from typing import Type
+from typing import Callable, TypeAlias
 
 from openforms.plugins.registry import BaseRegistry
 
 from .nodes import ComponentNode
 
+TComponentNode: TypeAlias = type[ComponentNode]
```
CharString, line 7: Maybe we should be more generic: `class Registry(BaseRegistry[TComponentNode])`?

And the binding to AbstractBasePlugin and iter_enabled_plugins, should live on some BasePluginRegistry? :thinking: 

Also the docstring of `__call__` doesn't make sense to me: all Registry's I've seen are used as class decorators.
The only differences I see in the override is the unique_identifier length isn't checked, and check_plugin isn't called, no reference to the registry is set on the registrant, and a runtime check on top of the type check.

CharString (approved):
I have a thought, but this PR improves. So LGTM.

src/openforms/formio/rendering/registry.py
CharString: PS plugins.BaseRegistry is a fine name for such a "BasePluginRegistry"; no need to break api.

sergei-maertens: quickly discussed on slack - `ComponentNode` should just conform to `AbstractBasePlugin`, I don't see a reason why it shouldn't, but I deliberately left it out of scope for this PR because it was getting too big.

There's some other formio rendering refactoring that needs to be done, so perhaps I can group all those things together (there's a draft ticket in my column on the project board)

PR #3342, opened 2023-08-04
Check fieldSettings for what to do with optional fields

### LINKED ISSUE #3333: Could it be that some fields are not provided by the "fields" endpoint but are still required. Notably: Email-address and initials should always be required -- Taiga HadM 58 (turns out these fields were marked as "Visible" not "Required" but I guess we only show required fields? I'll make a new, less prio, ticket for this)

sergei-maertens:
Closes #3333

Now interpreted as:

* Hidden: only display required fields
* Visible: display all fields, mark required field names as required
* Required: all fields are displayed and required

Unclear right now is what the standaardvelden means.

TODO:

- [x] Add admin configuration to make field required despite being optional in JCC
- [ ] Figure out the phone numbers situation


joeribekker:
Refinement: Closed because the ticket is blocked and no stakeholder. API documentation of JCC unclear and no response yet from them.

PR #3361, opened 2023-08-14
Add current_year static variable

### LINKED ISSUE #1889: Enable 'maximum mode' in the datepicker component for choosing years only
### Thema / Theme

Form designer

A few of our forms need the functionality to be able to block when a certain calculation based on years is made, for example %currentyear% - 4 = block. 

However, the date-component - which does allow to make calculations based on %today% - only allows for full dates including months and days.
It would be nice if the date component could be adapted to offer setting it to 'year only', 'month and year only' and 'full date'.

These validations could be set statically (as we did for now), but then the forms would have to be adapted again every new year.

--

Als formulierbouwer
Wil ik een nieuw component dat zich gedraagt als een datum component maar dan alleen met jaren.
Zodat ik kan valideren op jaar en enkel jaar.

**Acceptatiecriteria**

Een datum validatie die controleert op [vandaag] [+ of -] [X jaar]. Reeds beschikbaar voor datum component dus op 1-1-2023, ipv 2023)

Case

Haarlemmermeer heeft een formulier waarin zij willen valideren voor welk Jaar een aanvraag wordt ingediend. Alleen het jaar is dan relevant, niet de dag.

To offer dynamic validations based on other time-units than full date.

This seems possble with form.io's [maximum mode](https://cloud.githubusercontent.com/assets/13321142/13097253/30722118-d4e5-11e5-9a59-4363e16abbed.png) in the datepicker settings

To illustrate, the validations possible for a date component:
image

And the more limited ones for a regular 'number' component (which we've used now as a stand-in for a year-field):
image

Related: Dimpact PF-68

CharString:
Closes #1889

Having current_year available as a variable solves the problem of
Haarlemmermeer of not being able to "block"/validate based on an integer
relative to the current year. They can adapt their current forms to use
this, so they won't need to change the forms on each January 1st
anymore.

I think the wish for a picker is a distraction from this core problem.
But should one really feel the need for a year picker, with this
`current_year` one can construct a one with a Select component with
variable options with an option expression like this:

```json
{
  "merge": [
    { "-": [{ "var": "current_year" }, 3] },
    { "-": [{ "var": "current_year" }, 2] },
    { "-": [{ "var": "current_year" }, 1] },
    { "var": "current_year" },
    { "+": [{ "var": "current_year" }, 1] },
    { "+": [{ "var": "current_year" }, 2] },
    { "+": [{ "var": "current_year" }, 3] }
  ]
}
```

This solution evades all kinds of combinations FormIO fun and
JavaScript's ill-conceived Date API.

PS: use `{{ current_year|stringformat: "d" }}` to diplay the value in templates
without thousands separator.

### DIFF
--- a/src/openforms/variables/static_variables/static_variables.py
+++ b/src/openforms/variables/static_variables/static_variables.py
@@ -30,6 +30,16 @@ class Today(BaseStaticVariable):
         return timezone.localtime(now_utc).date()
 
 
+@register_static_variable("current_year")
+class CurrentYear(BaseStaticVariable):
+    name = _("Current year")
+    data_type = FormVariableDataTypes.int
+
+    def get_initial_value(self, submission: Optional[Submission] = None) -> int:
+        now_utc = timezone.now()
+        return timezone.localtime(now_utc).year
+
+
 @register_static_variable("environment")
 class Environment(BaseStaticVariable):
     name = _("Environment")
--- a/src/openforms/variables/tests/test_static_variables.py
+++ b/src/openforms/variables/tests/test_static_variables.py
@@ -29,6 +29,21 @@ class NowTests(TestCase):
         self.assertEqual(variable.initial_value, timezone.now())
 
 
+@freeze_time("2022-08-29T17:10:00+02:00")
+class CurrentYearTests(TestCase):
+    def test_with_submission(self):
+        submission = SubmissionFactory.build()
+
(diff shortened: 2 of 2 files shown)


sergei-maertens (approved):
Approving and merging this, however I'll reopen the ticket so for the recipe/documentation aspect.

PR #3363, opened 2023-08-14
Update appointment cancel/change links

### LINKED ISSUE #3322: Check cancel and change links and view functions
With the new flow, these actions that are shown in the email, don't work. A blank page is shown, with just the title element.

sergei-maertens:
Closes #3322

* removed new style appointment change links as it's deemed too complicated in the frontend to solve
* cancel situation is handled in the SDK PR


src/openforms/appointments/admin.py
```diff
@@ -73,17 +74,22 @@ def get_list_display(self, request):
...
 
         if obj.start_time and obj.start_time <= timezone.now():
             return "-"
 
-        actions = [
-            ("change", _("Change")),
-            ("cancel", _("Cancel")),
+        actions: list[tuple[Literal["cancel", "change"], str]] = [
+            ("cancel", gettext("Cancel")),
         ]
```
sergei-maertens, line 86: type checker was complaining as it infers the `str` of `tuple[str, ...]` to be a `Literal` :(

CharString (approved):
Are the ZDS removals from the po related?

sergei-maertens:
> Are the ZDS removals from the po related?

No, it's a drive-by dead code removal

PR #3374, opened 2023-08-16
Rework statement checkboxes configuration

### LINKED ISSUE #3372: Allow configuring checkboxes (privacy, statement of truth) per form
Discussed with the stakeholder.

Currently the privacy and declaration of truth checkboxes are configured globally, but not every checkbox is equally applicable to every form. We've decided that the actual texts remain global configuration, and whether they're enabled or not can be configured globally (the default) or have per-form overrides.

**Tasks**

- [x] Add checkboxes configuration part to form designer, dropdown with 3 options: use global config, enable, disable
    - [x] This must be done for each of both checkboxes
- [x] Update validation serializers to take form-specific config into account
- [x] Include checkboxes configuration as part of the form serializer
- [x] With the above, the API endpoint for global configuration can be deleted, as we should always fetch this from the form configuration.

sergei-maertens:
Closes #3372

The SDK needs a PR to handle these changes too.

- Information is now exposed as part of the form detail endpoint/serializer
- Configuration on the form specifies whether it's enabled or not (or defer to global configuration, the default)
- Removed the newly added endpoint since it requires form context


src/openforms/forms/api/serializers/form.py
```diff
@@ -168,6 +172,14 @@ class FormSerializer(PublicFieldsSerializerMixin, serializers.ModelSerializer):
     )
     hide_non_applicable_steps = serializers.SerializerMethodField(read_only=True)
     submission_report_download_link_title = serializers.SerializerMethodField()
+    submission_statements_configuration = serializers.SerializerMethodField(
```
SilviaAmAm, line 175: This name confused me a bit. I wonder if `formio_checkbox_statements`, or `config_checkbox_statements` would be clearer :thinking:

src/openforms/js/lang/nl.json
```diff
@@ -114,6 +114,11 @@
     "description": "Button text to add extra item",
     "originalDefault": "Add another"
   },
+  "3ZtB10": {
+    "defaultMessage": "Vraag toestemming om gegevens te verwerken",
```
SilviaAmAm, line 118: I think this translation is not the right one?

src/openforms/submissions/api/validators.py
```diff
@@ -67,16 +66,17 @@ def __call__(self, form: Form, field: serializers.RelatedField):
...
         self.ask_statement_field_name = ask_statement_field_name
         self.message = message or self.message
 
-    def __call__(self, value: bool):
-        config = GlobalConfiguration.get_solo()
-        assert isinstance(config, GlobalConfiguration)
-
-        should_statement_be_accepted = getattr(config, self.ask_statement_field_name)
+    def __call__(self, value: bool, field: serializers.BooleanField):
+        form = field.context["submission"].form
```
SilviaAmAm, line 76: With the appointments, the `submission` in the `context` can be `None` in case the that the user does not have the right permissions for the submission. Could that be a problem here?

src/openforms/forms/api/serializers/form.py
sergei-maertens: I'm not 100% sure - we already have `submissionAllowed` which controls some parts of the summary page related to completing/submitting the form, so I wanted to have that `submission` part in there.

I'm also not sure about mentioning `formio` in there, as it's an implementation detail?

Maybe `submission_completion_checkboxes` would work, but if we ever get a different component in there than a checkbox, then that's also no longer an accurate statement. I'm kind of anticipating clients wanting to put whatever fields in there, e.g. they could request a signature component at some point too. Maybe `submission_completion_configuration` then? Which aligns with the `formStep.configuration` field which holds a formio definition.

SilviaAmAm: Ahh fair, it was the 'submission' part that confused me :smile: Then it can remain as it is!

PR #3387, opened 2023-08-18
Ensure customer detail data is normalized

### LINKED ISSUE #3286: Incorporate data normalization during appointment registration
See `openforms.appointments.utils.get_formatted_phone_number`, probably we need to configure fields to have a sanitizer mapping that get passed the values?

sergei-maertens:
Closes #3286

Generic solution for normalization of data, which is plugin specific

### DIFF
--- a/src/openforms/appointments/base.py
+++ b/src/openforms/appointments/base.py
@@ -2,7 +2,7 @@ import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import date, datetime
-from typing import Generic, List, TypeVar
+from typing import Callable, Generic, List, TypeAlias, TypeVar
 
 from django.db.models import TextChoices
 from django.urls import reverse
@@ -74,6 +74,10 @@ class CustomerDetails(Generic[F]):
     details: dict[F, JSONPrimitive]
 
 
+TInput = TypeVar("TInput", bound=JSONPrimitive)
+Normalizer: TypeAlias = Callable[[TInput], TInput]
+
+
 @dataclass()
 class AppointmentDetails:
     identifier: str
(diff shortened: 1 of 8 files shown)


src/openforms/appointments/base.py
```diff
@@ -91,13 +95,14 @@ def __str__(self):
...
 
-class BasePlugin(ABC, AbstractBasePlugin):
+class BasePlugin(Generic[F], ABC, AbstractBasePlugin):
     """
     Base Appointment plugin.
     """
 
     configuration_options = EmptyOptions
     supports_multiple_products: bool = False
+    normalizers: dict[TextChoices, list[Normalizer]] = {}
```
sergei-maertens, line 105:
```suggestion
    normalizers: dict[F, list[Normalizer]] = {}
```

src/openforms/appointments/base.py
```diff
@@ -185,13 +190,26 @@ def get_required_customer_fields(
...
+    ) -> dict[F, JSONPrimitive]:
+        # XXX: We could consider using openforms.formio.service.normalize_value_for_component,
+        # but there currently is no phone normalizer and adding it could potentially break this
+        # component...
+        normalized = {}
+        for key, value in data.items():
+            for normalizer in self.normalizers.get(key, []):
+                value = normalizer(value)
+            normalized[key] = value
+        return normalized
```
CharString, line 204: For this beauty, Haskell has the `.` operator and PureScript `<<<`. You could read it as "after".
So `f . g` is another function that applies `f` to the result of applying `g` to its argument.  Instead of a list of functions to loop through, the normalizer would be 1 function to apply... And the default not `[]` but the identity function. And this checks, because we defined `Normalizer` as a function from `a` to `a`; a functor... more specifically an endofunctor... You know... the kind of functors from the definition of an monad. :wink:

CharString (changes requested):
The demo normalizer is confusion. I'm fine with removing it.

sergei-maertens:
> The demo normalizer is confusion. I'm fine with removing it.

I'll think of something else, removing it is not an option since I want this behaviour tested at the base plugin level.

CharString (approved):

PR #3390, opened 2023-08-21
Ensure QMatic works with new-style appointments

### LINKED ISSUE #3299: Support multiple products/people in Qmatic appointments
Left over from open-formulieren/open-forms#2471

- ~~Check the documentation on whether multi-person appointments in the simple form are supported~~ it is possible
- ~~Implement the multi-product client methods if they're supported~~ it is possible

sergei-maertens:
Closes #3299 - note that there is an SDK PR to make the "amount" field readonly when multiple products is not supported

This ensures that Qmatic works in single-product, single-customer mode with the new appointment flow. It also deprecates the entire plugin as it appears that this particular plugin is not currently used.

We can revert on the deprecation if a stakeholder steps up.

### DIFF
--- a/src/openforms/appointments/contrib/qmatic/tests/test_plugin.py
+++ b/src/openforms/appointments/contrib/qmatic/tests/test_plugin.py
@@ -1,6 +1,7 @@
-from datetime import date, datetime, timezone
+from datetime import date, datetime
 
 from django.test import SimpleTestCase, TestCase
+from django.utils import timezone
 from django.utils.translation import gettext as _
 
 import requests_mock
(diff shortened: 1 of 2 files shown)


src/openforms/appointments/contrib/qmatic/tests/test_plugin.py
```diff
@@ -1,6 +1,7 @@
-from datetime import date, datetime, timezone
+from datetime import date, datetime
```
CharString, line 1: Was timezone an unused import?

CharString (approved):

src/openforms/appointments/contrib/qmatic/tests/test_plugin.py
sergei-maertens: NO, it was a wrong import. It is replaced with `django.utils.timezone`

PR #3392, opened 2023-08-21
[#3310] Cramped pdf report approval and statement of truth addition

### LINKED ISSUE #3310: PDF contains cramped [X] approval

2.1.4

image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

### LINKED ISSUE #3388: Add [X] for statement of truth
Should look the same like #3310 

The relevant submission data is in https://github.com/open-formulieren/open-forms/blob/b14b07be7fa9e206594840d83baf6e184c339d4f/src/openforms/submissions/models/submission.py#L230

Note that this is not always required - similar logic applies as the privacy config. Also note that the privacy config/statement of truth have been made configurable on the form level: 

https://github.com/open-formulieren/open-forms/blob/b14b07be7fa9e206594840d83baf6e184c339d4f/src/openforms/forms/models/form.py#L201

vaszig:
Fixes #3310
Fixes #3388


src/openforms/config/models.py
```diff
@@ -624,6 +624,13 @@ def render_privacy_policy_label(self) -> str:
             backend=openforms_backend,
         )
 
+    def render_statement_of_truth_label(self) -> str:
+        return render_from_string(
+            self.statement_of_truth_label,
+            context={"global_configuration": self},
+            backend=openforms_backend,
+        )
```
sergei-maertens, line 632: We currently don't support template expressions in this field, so this is not necessary. You can just use `config.statement_of_truth_label` in the template.

src/openforms/scss/pdfs/_privacy-policy.scss
```diff
@@ -1,11 +1,13 @@
...
-.submission-privacy-policy {
+.submission-consent {
   display: flex;
   align-items: flex-start;
 
   @include bem.element('text') {
-    padding-left: 5px;
+    flex: 1;
+    padding-left: 8px;
+    min-width: 0;
```
sergei-maertens, line 10: could we try `gap: 8px` to get the spacing between elements? Or is there a particular reason why this doesn't work?

So it would become:

```scss
.submission-consent {
  display: flex;
  align-items: flex-start;
  gap: 8px;

  @include bem.element('text') {
    p { ... }
  }
}
```

src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -101,11 +101,19 @@ <h2 class="subtitle">{{ submission_step_node.render }}</h2>
...
+                <section class="submission-consent">
+                    <div class="submission-consent__check">[X]</div>
+                    <div class="submission-consent__text">{{ config.render_privacy_policy_label }}</div>
                 </section>
             {% endif %}
+
+            {% if config.ask_statement_of_truth and report.submission.statement_of_truth_accepted %}
+                <section class="submission-consent">
+                    <div class="submission-consent__check">[X]</div>
+                    <div class="submission-consent__text">{{ config.render_statement_of_truth_label }}</div>
```
sergei-maertens, line 113:
```suggestion
                    <div class="submission-consent__text">{{ config.statement_of_truth_label }}</div>
```
I'd like to move away of template syntax for this, the privacy policy currently needs it because of a custom template tag, but we should clean that up at some point too.

src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -101,11 +101,19 @@ <h2 class="subtitle">{{ submission_step_node.render }}</h2>
             {% endif %}
 
             {% if config.ask_privacy_consent and report.submission.privacy_policy_accepted %}
```
sergei-maertens, line 103: this now only looks at the global configuration (via `config`) but doesn't take the form configuration into account. There is a `Form.get_statement_checkbox_required` method that you should use for this (it's also what the validators in the serializer use, see https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/api/validators.py#L77)

sergei-maertens (changes requested):

src/openforms/scss/pdfs/_privacy-policy.scss
vaszig: Yes, I used `gap` at the beginning but it doesn't seem to work in the pdf, that's why I left the `padding-left` as it was originally.

src/openforms/config/models.py
vaszig: And how can we avoid the `<p>` tags in the result?If I just use the attribute it renders these as well.

src/openforms/config/tests/test_consent.py
```diff
@@ -0,0 +1,48 @@
+from django.test import TestCase, override_settings
+
+from openforms.config.models import GlobalConfiguration
+
+
+@override_settings(LANGUAGE_CODE="en")
+class PrivacyPolicyTests(TestCase):
```
sergei-maertens, line 7: this seems to duplicate `test_privacy_policy.py`, no?

src/openforms/config/tests/test_consent.py
```diff
@@ -0,0 +1,48 @@
...
+        label = conf.render_privacy_policy_label()
+
+        self.assertEqual(
+            "I read the  and agree.",
+            label,
+        )
+
+
+@override_settings(LANGUAGE_CODE="en")
+class StatementOfTruthTests(TestCase):
```
sergei-maertens, line 37: This test doesn't really add anything - it's just setting a value for a Django model field and asserting that the value was saved - so you're testing that Django works, which is not needed.

src/openforms/submissions/report.py
```diff
@@ -32,6 +32,18 @@ def __post_init__(self):
     def form(self) -> Form:
         return self.submission.form
 
+    @property
+    def needs_privacy_content(self) -> bool:
```
sergei-maertens, line 36:
```suggestion
    def needs_privacy_consent(self) -> bool:
```

src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -100,12 +100,20 @@ <h2 class="subtitle">{{ submission_step_node.render }}</h2>
                 </section>
             {% endif %}
 
-            {% if config.ask_privacy_consent and report.submission.privacy_policy_accepted %}
-                <section class="submission-privacy-policy">
-                    <div class="submission-privacy-policy__check">[X]</div>
-                    <div class="submission-privacy-policy__text">{{ config.render_privacy_policy_label }}</div>
+            {% if report.needs_privacy_content and report.submission.privacy_policy_accepted %}
```
sergei-maertens, line 103:
```suggestion
            {% if report.needs_privacy_consent and report.submission.privacy_policy_accepted %}
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3396, opened 2023-08-22
Fix integrity error when retrying to submit appointment

sergei-maertens:
Partly closes open-formulieren/open-forms-sdk#506


src/openforms/appointments/api/views.py
```diff
@@ -388,6 +390,13 @@ def get_serializer_context(self):
         context.update({"submission": submission})
         return context
 
+    @transaction.atomic
+    def create(self, request: Request, *args, **kwargs):
+        # ensure any previous attempts are deleted before creating a new one
+        submission = self.extract_submission()
+        Appointment.objects.filter(submission=submission).delete()
```
CharString, line 397: Can there be other, existing appointments we don't want to delete?

sergei-maertens: No, it's a `OneToOneField`, which is also where the unique validation error is coming from.

CharString (approved):

PR #3405, opened 2023-08-23
Remove service fetch and appointment feature flags

sergei-maertens:
We completely forgot to turn these on by default :)


CharString (approved):
image

sergei-maertens:
> image

Users be like:

gif

PR #3412, opened 2023-08-24
🐛[#3320] Add appointment info to summary pdf

### LINKED ISSUE #3320: The summary PDF for appointments does not contain appointment data
The summary is most likely empty due to special form for appointments.

CharString:
Closes #3320

### DIFF
--- a/src/openforms/submissions/models/submission_report.py
+++ b/src/openforms/submissions/models/submission_report.py
@@ -7,6 +7,7 @@ from django.utils.translation import gettext_lazy as _, override
 from celery.result import AsyncResult
 from privates.fields import PrivateMediaFileField
 
+from openforms.appointments.models import AppointmentInfo
 from openforms.utils.pdf import render_to_pdf
 
 from ..report import Report
@@ -65,11 +66,21 @@ class SubmissionReport(models.Model):
         :return: string with the HTML used for the PDF generation, so that contents
           can be tested.
         """
+        try:
+            appointment_id = self.submission.appointment_info.appointment_id
+        except AppointmentInfo.DoesNotExist:
+            appointment_id = None
+
         with override(self.submission.language_code):
             form = self.submission.form
             html_report, pdf_report = render_to_pdf(
                 "report/submission_report.html",
-                context={"report": Report(self.submission)},
+                context={
+                    "report": Report(self.submission),
+                    # appointment_information tag needs _submission and _appointment_id
+                    "_submission": self.submission,
+                    "_appointment_id": appointment_id,
+                },
             )
             self.content = ContentFile(
                 content=pdf_report,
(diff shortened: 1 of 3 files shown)


src/openforms/submissions/models/submission_report.py
```diff
@@ -65,11 +66,21 @@ def generate_submission_report_pdf(self) -> str:
         :return: string with the HTML used for the PDF generation, so that contents
           can be tested.
         """
+        try:
+            appointment_id = self.submission.appointment_info.appointment_id
```
vaszig, line 70: Shouldn't we check if the `appointment_info` exists first and then access the `appointment_id`?

vaszig (commented):
I haven't been able to test this via a form, but concerning the code I only have one question, because I haven't used a try except this way.

sergei-maertens (approved):
Reluctantly approving this because it is a release blocker, but as discussed in the office:

Adding extra information to the template context to make the `{% appointment_information %}` tag work is not the desired approach, because:

* right now it breaks `openforms.submissions.dev_views.SubmissionPDFTestView` which does not receive this context
* this tag is really meant for confirmation emails, not PDFs. The markup with inline styles is quite horrible and we'd like to prevent it from escaping it's "email context" containment.
* it may behave funky with old-style appointments, because those have the form data as part of the formio form definitions/form steps, which renders through the generic Renderer mechanism. That could cause the appointment information to be displayed twice.

So, please create a follow up PR to address these things, some ideas/guidelines:

* Check if we can build on top of the existing `openforms.submissions.rendering.renderer` and let it render an `AppointmentNode` (to be defined and implemented in `openforms.appointments`). I think this would then be the only import of the appointments app to the submissions app (to stick to loose coupling principles).
* Move the test checking for appointment data in the PDF to the appointments app. Appointments is a open forms module making use of the core submissions module, not the other way around.
* Set up proper HTML and CSS for the appointments data in the PDF.

src/openforms/submissions/models/submission_report.py
sergei-maertens, line 70: Can you add an extra check to only target new style appointments:

See https://github.com/open-formulieren/open-forms/blob/master/src/openforms/appointments/core.py#L84

```py
    if not submission.form.is_appointment:
        raise NoAppointmentForm("Not an appointment form")
```

CharString: `.` operator is left associative, meaning it's evaluated from left to right like `((self.submission).appointment_info).appointment_id`. So at the second braces `Submission.__get_attribute__("appointment_info")` (or `__getattr__`) will raise `AppointmentInfo.DoesNotExist` (not an `AttributeError` 🙹) before it gets to the last `.`, which is explicitly caught with an `except` block.

If you want to learn more about operator overloading in Python, [Expecter Gadget](https://github.com/garybernhardt/expecter) is a more approachable code base. Or https://github.com/garybernhardt/dingus, which I was using before unittest.mock was in the standard library,

PR #3424, opened 2023-08-25
[#3051] Add datetime fields to automatically activate/deactivate a form

### LINKED ISSUE #3051: Add datetime fields to automatically activate/deactivate a form
### Thema / Theme

Admin

DH Taiga issue 438 / user story 433

Add 2 fields to automatically set a form to active / inactive

_No response_

_No response_

vaszig:
Closes #3051


src/openapi.yaml
SilviaAmAm, line 1: Before generating the OAS, run the compile messages management command. This avoids changing the other fields descriptions back to english.

(see docs note here: https://open-forms.readthedocs.io/en/latest/developers/i18n.html?highlight=OAS#backend)

src/openforms/forms/admin/tasks.py
```diff
@@ -102,3 +103,33 @@ def process_forms_import(import_file: str, user_id: int) -> None:
...
     call_command("delete_export_files")
+
+
+@app.task()
+def activate_forms():
+    """Activate all the forms that should be activated by the specific date and time."""
+    now = timezone.now().minute
+    forms = Form.objects.filter(
+        active=False, _is_deleted=False, activate_on__minute=now
+    )
```
SilviaAmAm, line 114: Shouldn't this also check that the date is correct?

src/openforms/forms/admin/tasks.py
```diff
@@ -102,3 +103,33 @@ def process_forms_import(import_file: str, user_id: int) -> None:
...
+        form.save()
+        logger.debug(f"Activated form {form.admin_name}")
+        logevent.form_activated(form)
+
+
+@app.task()
+def deactivate_forms():
+    """Deactivate all the forms that should be deactivated by the specific date and time."""
+    now = timezone.now().minute
+    forms = Form.objects.live().filter(deactivate_on__minute=now)
```
SilviaAmAm, line 128: Same remark as for activate_on

src/openforms/forms/models/form.py
```diff
@@ -228,6 +228,18 @@ class Form(models.Model):
         ),
     )
     _is_deleted = models.BooleanField(default=False)
+    activate_on = models.DateTimeField(
+        _("activate on"),
+        blank=True,
+        null=True,
+        help_text=_("Date and time that the form should be activated."),
```
SilviaAmAm, line 235:
```suggestion
        help_text=_("Date and time on which the form should be activated."),
```

src/openforms/forms/tests/admin/test_tasks.py
```diff
@@ -122,3 +127,115 @@ def test_import_form_failure(self, m_import_form):
...
+        activate_forms()
+
+        log_entry = TimelineLogProxy.objects.last()
+        message = log_entry.get_message()
+
+        self.assertEqual(log_entry.content_object.id, form.id)
+        self.assertEqual(
+            message.strip().replace("&quot;", '"'),
+            f"{log_entry.fmt_lead}: Form was activated.",
+        )
```
SilviaAmAm, line 161: I think this can be tested in the previous test? It feel that it doesn't need a separate test

src/openforms/forms/tests/admin/test_tasks.py
```diff
@@ -122,3 +127,115 @@ def test_import_form_failure(self, m_import_form):
...
+        form._is_deleted = True
+        form.save(update_fields=["_is_deleted"])
+
+        activate_forms()
+
+        form.refresh_from_db()
+
+        self.assertFalse(form.active)
+        self.assertIsNotNone(form.activate_on)
+
```
SilviaAmAm, line 186: I think you also need a test where you have a form that should be activated at `2023-10-11T21:15:00Z` (not the same date, but the same minute) and make sure that at `2023-10-10T21:15:00Z` it is not activated.

src/openforms/forms/tests/admin/test_tasks.py
```diff
@@ -122,3 +127,115 @@ def test_import_form_failure(self, m_import_form):
...
+        form1.refresh_from_db()
+        form2.refresh_from_db()
+
+        self.assertFalse(form1.active)
+        self.assertIsNone(form1.deactivate_on)
+        self.assertFalse(form2.active)
+        self.assertIsNone(form2.deactivate_on)
+
+    @freeze_time("2023-10-10T21:15:00Z")
+    def test_form_deactivation_is_logged(self):
```
SilviaAmAm, line 205: Same remarks as for the activation tests :smile:

src/openforms/conf/base.py
```diff
@@ -699,6 +699,14 @@
         "task": "openforms.forms.admin.tasks.clear_forms_export",
         "schedule": crontab(hour=0, minute=0, day_of_week="sunday"),
     },
+    "activate-forms": {
+        "task": "openforms.forms.admin.tasks.activate_forms",
+        "schedule": crontab(minute="*"),
```
SilviaAmAm, line 704: I wonder if it is necessary to make it check every minute? :thinking: Would every hour not be sufficient? Although then we would have to make corresponding changes in the frontend to make it clear that it is only checked every hour :sweat_smile:

src/openforms/forms/models/form.py
```diff
@@ -228,6 +228,18 @@ class Form(models.Model):
...
+    )
+    deactivate_on = models.DateTimeField(
+        _("deactivate on"),
+        blank=True,
+        null=True,
+        help_text=_("Date and time that the form should be deactivated."),
+    )
```
SilviaAmAm, line 242: I think that there should be some validation that these fields are in the future?

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -290,6 +292,44 @@ const FormConfigurationFields = ({
...
+            />
+          }
+          helpText={
+            <FormattedMessage
+              defaultMessage="When the form should be activated."
+              description="Form activation field help text"
+            />
+          }
+        >
+          <DateTimeInput value={activateOn} onChange={onChange} enableTime={true} />
```
SilviaAmAm, line 311: Not sure how easy to do, but I think that the widget should grey out datetimes in the past?

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -60,6 +60,16 @@ const normalizeLimit = (draft, field) => {
...
 
+/**
+ * Convert empty str value for a datetime field to null.
+ */
+const normalizeDateTime = (draft, field) => {
+  const form = draft.form;
+  if (form[field] === '') {
+    form[field] = null;
+  }
+};
```
SilviaAmAm, line 71: Is this specific to a datetime field?

src/openforms/scss/components/admin/_datetime.scss
```diff
@@ -0,0 +1,9 @@
+.datetime-input {
+  display: inline-flex;
+  gap: 3px;
+  padding-right: 2px;
+
+  i {
+    cursor: pointer;
+  }
```
SilviaAmAm, line 8: These icons look a bit misaligned? (tried on Firefox)
Screenshot from 2023-08-30 10-56-40

SilviaAmAm (changes requested):

src/openforms/forms/admin/tasks.py
vaszig: You are totally right, thanks for this!

src/openforms/conf/base.py
vaszig: IMO since the user can define the minutes as well, it should be every minute. @sergei-maertens can share his thoughts with us.

src/openforms/forms/models/form.py
vaszig: Already discussed with Sergei (it was my first thought as well), but there are some use cases which would not be better if we had this validation.

src/openforms/js/components/admin/form_design/data/complete-form.js
vaszig: As I have implemented this, no. I will fix this and give it a more generic name.

src/openforms/conf/base.py
sergei-maertens: Let's just do every minute and see if that load causes issues. If it does, I think I would rather tap into the `post_save` of the form and do a `activate_form.apply_async(eta=form.activate_on)` or run the beat command every hour and have that schedule the actual activation in x minutes.

For now I thought about not doing this because the task could be scheduled to run in 3 months, and that task/key will definitely expire in redis or has too much risk to be lost anyway.

SilviaAmAm (approved):

src/openforms/forms/admin/tasks.py
```diff
@@ -102,3 +104,40 @@ def process_forms_import(import_file: str, user_id: int) -> None:
...
+        activate_on__gt=now - timedelta(minutes=5),
+    )
+
+    for form in forms:
+        form.active = True
+        form.activate_on = None
+        logger.debug(f"Activated form {form.admin_name}")
+        logevent.form_activated(form)
+
+    Form.objects.bulk_update(forms, fields=["active", "activate_on"])
```
sergei-maertens, line 126:
```suggestion
    for form in forms:
        form.active = True
        form.activate_on = None
        logger.debug("Activating form %s", form.admin_name)
        transaction.on_commit(lambda: logevent.form_activated(form))

    Form.objects.bulk_update(forms, fields=["active", "activate_on"])
```
& then also decorate the task with `@transaction.atomic`

So, a couple things here:

1. The logging - never do the interpolation directly in the log message, since Sentry aggregates logs by their message template. My code suggestion ensures that there's only a single Sentry record for this activation, while having many events each with their own context and timestamp for every form.
2. the transactions - it's now possible (with autocommit, the default) that a log event is written claiming activation, but the actual `bulk_update` query failing. So now we have a lying log record, because the form failed to activate.

I don't expect failures now, but I'm also not ruling them out, weird things can happen.

I don't really like wrapping the entire task in a transaction.atomic either (as I suggest) - one broken form would then possibly prevent other forms from being activated. So probably the most robust way to handle this is to activate each form individually and wrap each one in its own transaction:

```py
for form in forms:
    form.active = True
    form.activate_on = None
    with transaction.atomic():
        form.save(update_fields=["active", "activate_on"])
        logevent.form_activated(form)
```

I would then make that even nicer by putting all that activation logic in a model method, so your task becomes:

```py
for form in forms:
    with transaction.atomic():
         try:
            form.activate()
        except DatabaseError as exc:
            logger.error("Form activation of form %s failed", form.admin_name, exc_info=exc, extra={"pk": form.pk})
```

and then implement `def activate` to set the appropriate fields + write to the event log.

src/openforms/forms/tests/admin/test_tasks.py
```diff
@@ -122,3 +127,175 @@ def test_import_form_failure(self, m_import_form):
...
+        # make sure the activation is logged as well
+        log_entries = TimelineLogProxy.objects.all()
+        self.assertEqual(log_entries.count(), 2)
+
+        for log_entry in log_entries:
+            message = log_entry.get_message()
+
+            self.assertEqual(
+                message.strip().replace("&quot;", '"'),
+                f"{log_entry.fmt_lead}: Form was activated.",
```
sergei-maertens, line 157: Is this going to break when we update translations? Tests run with NL locale by default iirc.

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
sergei-maertens, line 1: Can you please run `./bin/makemessages_js.sh` and `./bin/compilemessages_js.sh` to get rid of the console warnings about missing translations?

src/openforms/js/components/admin/forms/Inputs.js
```diff
@@ -66,7 +68,22 @@ const DateTimeInput = ({name, value, formatDatetime, onChange, ...extraProps}) =
 
   return (
     <div className="datetime-input">
-      <input ref={datetimePickerRef} placeholder={placeHolder} />
+      <input
+        ref={datetimePickerRef}
+        placeholder={placeHolder}
+        value={value ? value : ''}
```
sergei-maertens, line 74:
```suggestion
        value={value ?? ''}
```
this will use the RHS if `value` is `undefined` or `null`, and otherwise use `value`. Small stylistic thing, not an essential change!

src/openforms/scss/components/admin/_datetime.scss
```diff
@@ -0,0 +1,10 @@
+.datetime-input {
+  display: inline-flex;
+  gap: 3px;
+  padding-right: 2px;
+
+  i {
```
sergei-maertens, line 6: we generally avoid targetting HTML elements with selectors. In this case, `.fa` is more appropriate as it indicates it's a Font-Awesome icon.

src/openforms/scss/components/admin/_datetime.scss
```diff
@@ -0,0 +1,10 @@
...
+    margin-top: 2px;
```
sergei-maertens, line 7: perhaps applying `align-items: center` on the parent works better to vertically center things?

sergei-maertens (changes requested):

src/openforms/scss/components/admin/_datetime.scss
vaszig: Yes, this is what I tried first but it isn't shown correctly

image

src/openforms/forms/tasks.py
```diff
@@ -82,3 +85,57 @@ def recouple_submission_variables_to_form_variables(form_id: int) -> None:
...
+        activate_on__lte=now,
+        activate_on__gt=now - timedelta(minutes=5),
+    )
+
+    for form in forms:
+        with transaction.atomic():
+            try:
+                form.activate()
+            except DatabaseError as exc:
+                logger.error(
```
sergei-maertens, line 108: Can you address the missing coverage? :grimacing:

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3427, opened 2023-08-28
[#3389] Do not build test images anymore

### LINKED ISSUE #3389: Remove test docker images
### Thema / Theme

Other

Currently, OF provides Docker images with the test dependencies installed, so that the extensions modules can run the tests.

One of the issues with this approach is that the Docker images don't contain the files inside the `tests` directories (since these are specified in the `.dockerignore`).

One of the suggested approaches would be to use git submodules to do the testing of the custom extensions.

- [x] Haal centraal HR extension
- [x] Token exchange extension
- [x] Demo extension
- [x] Documentation

This is needed for #3185

_No response_

_No response_

SilviaAmAm:
Fixes #3389 

Depends on:
https://github.com/open-formulieren/demo-extension/pull/5
https://github.com/open-formulieren/open-forms-ext-token-exchange/pull/13

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -11,7 +11,6 @@ on:
         # options:
         #   - production
         #   - extensions
-        #   - ci
       image_name:
         required: true
         type: string # image name without tag, e.g. 'openforms/open-forms'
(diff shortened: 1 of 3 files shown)


sergei-maertens (commented):
https://github.com/open-formulieren/open-forms/blob/master/docker/ci/config.json#L28 can be removed too I think. Please check if that script itself also needs updating due to test images not being relevant at all anymore.

sergei-maertens (approved):

PR #3431, opened 2023-08-29
♻[#3416] Refactor rendering appointment data

### LINKED ISSUE #3416: Refactor rendering of appointment data in pdf.
Reluctantly approving this because it is a release blocker, but as discussed in the office:

Adding extra information to the template context to make the `{% appointment_information %}` tag work is not the desired approach, because:

* right now it breaks `openforms.submissions.dev_views.SubmissionPDFTestView` which does not receive this context
* this tag is really meant for confirmation emails, not PDFs. The markup with inline styles is quite horrible and we'd like to prevent it from escaping it's "email context" containment.
* it may behave funky with old-style appointments, because those have the form data as part of the formio form definitions/form steps, which renders through the generic Renderer mechanism. That could cause the appointment information to be displayed twice.

So, please create a follow up PR to address these things, some ideas/guidelines:

* Check if we can build on top of the existing `openforms.submissions.rendering.renderer` and let it render an `AppointmentNode` (to be defined and implemented in `openforms.appointments`). I think this would then be the only import of the appointments app to the submissions app (to stick to loose coupling principles).
* Move the test checking for appointment data in the PDF to the appointments app. Appointments is a open forms module making use of the core submissions module, not the other way around.
* Set up proper HTML and CSS for the appointments data in the PDF.

_Originally posted by @sergei-maertens in https://github.com/open-formulieren/open-forms/pull/3412#pullrequestreview-1593234684_

CharString:
Closes #3416

The idea is that a renderer should behave a like a Django Form does.
i.e. render the default representation when used as `{{ appointment }}`

No effort has been put (yet) into the getattr side of that API.
So within the submission_report template you can't access the
appointment attributes, the way you can access Django form fields.

### DIFF
--- a/src/openforms/appointments/renderer.py
+++ b/src/openforms/appointments/renderer.py
@@ -1,10 +1,14 @@
 from typing import Iterator
 
+from django.template.loader import render_to_string
+from django.utils.safestring import SafeString
+
 from openforms.formio.rendering.nodes import ComponentNode
 from openforms.formio.typing import Component
 from openforms.submissions.rendering import Renderer
 
 from .service import get_appointment
+from .utils import get_plugin
 
 
 class AppointmentRenderer(Renderer):
@@ -26,5 +30,24 @@ class AppointmentRenderer(Renderer):
                 renderer=self,
                 configuration_path=f"{index}",
             )
-            yield child_node
             yield from child_node
+
+    def __str__(self) -> SafeString:
+        # Like a Django Form renderer, render to the default representation
+
+        if (
+            not (appointment := get_appointment(self.submission))
+            or not appointment.plugin
+        ):
+            return ""
+
+        template_name = (
+            f"appointments/appointment__{self.mode}.{'html' if self.as_html else 'txt'}"
+        )
+        plugin = get_plugin(plugin=appointment.plugin)
+
+        ctx = {
+            "appointment": plugin.get_appointment_details(appointment.pk),
+            "contact_details": self.get_children(),  # todo: a bit of a wart
+        }
+        return render_to_string(template_name, ctx)
--- a/src/openforms/appointments/renderer.py
+++ b/src/openforms/appointments/renderer.py
@@ -7,8 +7,7 @@ from openforms.formio.rendering.nodes import ComponentNode
 from openforms.formio.typing import Component
 from openforms.submissions.rendering import Renderer
 
-from .service import get_appointment
-from .utils import get_plugin
+from .utils import get_appointment, get_plugin
 
 
 class AppointmentRenderer(Renderer):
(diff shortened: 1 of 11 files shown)


src/openforms/submissions/report.py
```diff
@@ -77,3 +77,13 @@ def confirmation_page_content(self) -> SafeString:
         # has been applied by ``Template.render``
         content = self.submission.render_confirmation_page()
         return content
+
+    @property
+    def appointment(self):
+        from openforms.appointments.renderer import AppointmentRenderer
```
sergei-maertens, line 83: Can you please expose this in `appointments/service.py` and import it from there? That's how it's documented that different django apps interface with each other.

src/openforms/appointments/renderer.py
```diff
@@ -26,5 +30,24 @@ def get_children(self) -> Iterator[ComponentNode]:
                 renderer=self,
                 configuration_path=f"{index}",
             )
-            yield child_node
             yield from child_node
+
+    def __str__(self) -> SafeString:
+        # Like a Django Form renderer, render to the default representation
```
sergei-maertens, line 36: I'm not sure I agree - there is the `render_report` management command which loops over all nodes: `for node in renderer:` and then outputs the submission data/information to the CLI (that's the CLI render mode). I would expect to also see the appointment information there.

This implementation now assumes that you'll always be rendering a Django HTML template, but that's not the case. The render nodes implementation abstracts this - output HTML when HTML is requested, output plain text otherwise (which can be for the CLI output mode or the confirmation email...).

I think that `yield (from) AppointmentNode(...)` in the main submission renderer is the better approach here, and then `AppointmentNode` and `AppointmentRenderer` (which is a wart for the e-mails) can share some functionality.

sergei-maertens (changes requested):

src/openforms/appointments/templates/appointments/appointment__pdf.html
```diff
@@ -0,0 +1,61 @@
+{% load i18n %}
+<section class="appointment">
+<h2>{% trans "Appointment information" %}</h2>
+
+<table>
```
sergei-maertens, line 5: Is there a reason this markup is different from the other markup (and class names) in `submission_report.html`? I would expect this information to just be formatted the same as a normal submission key-value with hardcoded titles since there are no form step names available.

src/openforms/appointments/utils.py
```diff
@@ -309,3 +309,10 @@ def get_confirmation_mail_suffix(submission: Submission) -> str:
         return ""
 
     return _("(updated)")
+
+
+def get_appointment(submission: Submission) -> Appointment | None:
```
sergei-maertens, line 314: I don't like that this `utils` module is now a mess of old, deprecated stuff and new things that are okay to re-use, but I suppose it's the best we can do for now.

sergei-maertens (approved):
I don't mind putting renderer semantics/cleanup on the backlog, because it is indeed more used as a node visitor capable of outputting to various "frontends". Now we have a good picture of what it needs to be able to do and possible future refactors we want to do, we should be able to draw up a proper specification.

PR #3454, opened 2023-09-04
[#3422] Prevent broken logic rule to break admin

### LINKED ISSUE #3422: Removing variables/fields can cause Logic tab to crash

2.3.0

Sentry: 337920
Utr: 112

Removing fields, groups or steps that are in use in the logic tab can cause the logic tab to crash, in turn breaking the form. 
The only way to fix it is to export the form, remove logic rules that contain the unused variables and import again.

When trying to import a form in this state you see the error: 
`Er is iets foutgegaan bij het importeren van formulier: {'json_logic_trigger': [ErrorDetail(string='De opgegeven variabele is niet aanwezig in het formulier', code='invalid')]}`

unsure about conditions on how to reproduce

SilviaAmAm:
Fixes #3422 

- [x] Add error boundary per rule showing the JSON of the rule for debugging purposes

Screenshot from 2023-09-07 15-11-49

- [x] Fix root cause

Screenshot from 2023-09-07 16-15-37

### DIFF
--- a/src/openforms/js/components/admin/form_design/FormLogic.js
+++ b/src/openforms/js/components/admin/form_design/FormLogic.js
@@ -13,6 +13,7 @@ import Loader from 'components/admin/Loader';
 import ButtonContainer from 'components/admin/forms/ButtonContainer';
 import Fieldset from 'components/admin/forms/Fieldset';
 import {ValidationErrorContext} from 'components/admin/forms/ValidationErrors';
+import ErrorBoundary from 'components/errors/ErrorBoundary';
 
 import {FormLogicContext} from './Context';
 import StepSelection, {useFormStep} from './StepSelection';
@@ -177,41 +178,24 @@ FormLogicRules.propTypes = {
   onAdd: PropTypes.func.isRequired,
 };
 
-const Rule = ({
+const RuleBody = ({
   isCreate,
-  _logicType,
-  description,
-  _mayGenerateDescription,
-  order,
-  jsonLogicTrigger,
-  triggerFromStep: triggerFromStepIdentifier,
-  actions,
   isAdvanced,
+  jsonLogicTrigger,
+  actions,
+  displayAdvancedOptions,
+  description,
+  triggerFromStepIdentifier,
   onChange,
-  onDelete,
-  errors = {},
+  errors,
+  mayGenerateDescription,
 }) => {
   const intl = useIntl();
-  const [displayAdvancedOptions, setDisplayAdvancedOptions] = useState(false);
   const [expandExpression, setExpandExpression] = useState(isCreate);
   const [resetRequest, setResetRequest] = useState(0);
   const triggerFromStep = useFormStep(triggerFromStepIdentifier);
 
-  const deleteConfirmMessage = intl.formatMessage({
-    description: 'Logic rule deletion confirm message',
-    defaultMessage: 'Are you sure you want to delete this rule?',
-  });
-
-  // if no logicType has been set yet, we first present the type selection before the
-  // actual rule can be set up.
-  if (!_logicType) {
-    return (
-      <LogicTypeSelection
-        onChange={selectedType => onChange({target: {name: '_logicType', value: selectedType}})}
-        onCancel={onDelete}
-      />
-    );
-  }
+  const TriggerComponent = isAdvanced ? AdvancedTrigger : Trigger;
 
   const onDescriptionGenerated = generatedDescription => {
     let newDescription = intl.formatMessage(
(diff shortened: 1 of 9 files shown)


src/openforms/js/components/admin/form_design/FormLogic.js
```diff
@@ -364,6 +303,23 @@ const Rule = ({
...
+                      description="Warning missing trigger step"
+                      defaultMessage="The selected trigger step could not be found in this form!
+                      Please change it by clicking on the cog icon here on the left."
+                    />
+                  ),
+                  level: 'warning',
+                },
+              ]}
+            />
+          )}
```
sergei-maertens, line 321: Is there a way we can pass this in so that the cog icon is automatically expanded and the warning shown with the relevant field, instead of having this indirect message?

src/openforms/js/components/admin/form_design/FormLogic.js
```diff
@@ -385,6 +341,137 @@ const Rule = ({
...
+      </div>
+
+      <div className="logic-rule-error__json">
+        {JSON.stringify({
+          jsonLogicTrigger,
+          triggerFromStepIdentifier,
+          order,
+          actions,
+          isAdvanced,
+        })}
```
sergei-maertens, line 410: wrapping this in `<code><pre>...</pre></code>` might make it look a tad better?

src/openforms/js/components/admin/form_design/FormLogic.js
```diff
@@ -385,6 +341,137 @@ const Rule = ({
...
+  jsonLogicTrigger,
+  triggerFromStep: triggerFromStepIdentifier,
+  actions,
+  isAdvanced,
+  onChange,
+  onDelete,
+  errors = {},
+}) => {
+  const intl = useIntl();
+  const [displayAdvancedOptions, setDisplayAdvancedOptions] = useState(false);
```
sergei-maertens, line 376: I think we could do the broken-trigger-from-step detection here and then you can immediately call `setDisplayAdvancedOptions(true)` if that's the case (or use that state as initial state value :thinking: )

src/openforms/js/components/errors/ErrorMessage.js
```diff
@@ -0,0 +1,23 @@
...
+        <FAIcon icon="exclamation-circle" title="error" />
+      </span>
+      <div className="error-message__children">{children}</div>
+    </div>
+  );
+};
+
+ErrorMessage.propTypes = {
+  children: PropTypes.node,
+  modifiers: PropTypes.arrayOf(PropTypes.string),
```
sergei-maertens, line 20: unused :)

src/openforms/scss/components/admin/_index.scss
```diff
@@ -38,3 +39,6 @@
 
 // Submission logs
 @import './logic-evaluated-table';
+
+// Errors
+@import 'error-message';
```
sergei-maertens, line 44:
```suggestion
@import './error-message';
```
bit of consistency :)

src/openforms/js/components/admin/form_design/FormLogic.js
SilviaAmAm: Mhmmm I wanted to have as much logic as possible inside the error boundary, so that if something unexpected goes wrong it doesn't break the page but the action to delete is still visible :thinking: Maybe I can pass `setDisplayAdvancedOptions` to `RuleBody`?

SilviaAmAm: Screenshot from 2023-09-08 12-08-39

src/openforms/js/components/admin/form_design/FormLogic.js
SilviaAmAm: Screenshot from 2023-09-08 12-15-04
:tada: 
(I left out the `<pre>` because it was making it go off the page)

src/openforms/js/components/admin/form_design/FormLogic.js
sergei-maertens: is this warning also done with error boundary then?

> Maybe I can pass setDisplayAdvancedOptions to RuleBody?

It's an option, but then you'll have to rely on `useEffect` I think and it may conflict with the user click actions that update the state :thinking: 

Error boundaries are really a bit of a last resort btw, it's good to have them because you don't know what will break, but if you can detect error situations and prevent crashes that way, that's always preferred. So that's why I suggested this approach :)

SilviaAmAm: > is this warning also done with error boundary then?

No, this error is done by checking if `triggerFromStepIdentifier` is set but `triggerFromStep` is null! 

But what I meant is: the errors tend to come from the logic that is now moved into the `RuleBody` component and wrapped in an error boundary.
If I move the code that checks for the trigger step back to `Rule`, if it breaks unexpectedly it will break the whole page because it will be outside of the error boundary.

sergei-maertens: I think I understand.

Could we maybe perform the check inside the advanced config (the component behind the cog) and force that to be open regardless of the toggle state/prop if there's an error? So then the actual checks are all contained inside the error boundary but we get nicer feedback.

sergei-maertens (approved):

PR #3459, opened 2023-09-05
Implement multi-product endpoints for Qmatic plugin

### LINKED ISSUE #3456: Implement multi-product/multi-person appointments for Qmatic
Will close the epic #2265

Qmatic is currently running in single-product mode, but it supports multi-product and multi-customer.

- [x] Implement multiple products and filtering by service group
- [x] Implement booking multi-product appointment

sergei-maertens:
Partly closes #3456

* Fixed `limit_to_location` config field being too short for Qmatic (got an ID with length 64)
* Refactored the exception handling in the Qmatic plugin to be consistent with JCC
* Refactored client to support both `v1` and `v2` API endpoints (added datamigration + updated docs)
* Implement retrieving/returning products based on location ID provided or not and whether any other products are already selected which limits the space of possible options due to service groups being used
* Tests are set up in an attempt to be robust, using VCR to be assured we have "mocks" generated from a real service & testing logical/semantic behaviour.


sergei-maertens:
Got some new insights, so please don't review yet.

src/openforms/appointments/contrib/qmatic/client.py
```diff
@@ -41,3 +82,36 @@ def request(self, method: str, url: str, *args, **kwargs):
...
+        # the shape depends on whether we hit v1 or v2
+        response_data = response.json()
+
+        # v2 API returns a list of service groups (expected to only have one item)
+        if location_id:
+            return response_data
+
+        # v1 API returns a list of branches
+        branches: list[BranchDict] = response_data
+        service_groups = sum((branch["serviceGroups"] for branch in branches), [])
```
CharString, line 116: :) I end up writing the sum over empty list too. I'm always unsure how to write a pythonic flatmap.
Whether it was inside out or outside in... I think it's
`[group for branch in branches for group in branch["serviceGroups"]]`

src/openforms/appointments/contrib/qmatic/plugin.py
```diff
@@ -40,6 +44,36 @@
...
+        @wraps(func)
+        def wrapper(*args, **kwargs) -> T:
+            try:
+                return func(*args, **kwargs)
+            except GracefulQmaticException:
+                return default
+
+        return wrapper
+
+    return decorator
```
CharString, line 74: Nice. I was thinking about something like this when I heard about algebraic effects in Unison. And how they use them to do recoverable error handling by throwing/raising with passing a continuation: https://www.unison-lang.org/learn/fundamentals/abilities/error-handling/

src/openforms/appointments/contrib/qmatic/client.py
```diff
@@ -1,10 +1,48 @@
...
+
 from requests import Session
 
+from .exceptions import QmaticException
 from .models import QmaticConfig
 
+# API DATA DEFINITIONS
+
+
+class ServiceDict(TypedDict):
```
CharString, line 11: Is this ever referenced other than as the base class of `FullServiceDict`?

src/openforms/appointments/contrib/qmatic/tests/data/vcr_cassettes/ListAvailableProductsTests/ListAvailableProductsTests.test_listing_products_with_location_and_product_constraint.yaml
```diff
@@ -0,0 +1,440 @@
+interactions:
+- request:
+    body: null
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Authorization:
+      - Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiIiLCJpYXQiOjE2OTM5MjU5OTgsImNsaWVudF9pZCI6IiIsInVzZXJfaWQiOiIiLCJ1c2VyX3JlcHJlc2VudGF0aW9uIjoiIn0.mqpHvBBHEsCM_8A9CnBnDLrcgqeq6Ees-lNVXnvdFRU
```
CharString, line 10: Are these disposable tokens? Or should they have been [filtered](https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-sensitive-data-from-the-request) from the cassette?

src/openforms/appointments/contrib/qmatic/client.py
sergei-maertens: Those sort of nested list comprehensions in python are an abomination

src/openforms/appointments/contrib/qmatic/plugin.py
sergei-maertens: Took me a bit of frustration when I first tried with a context manager, but that can't affect the return value of the wrapped block, and using a decorator by itself doesn't allow capturing any args you want in the log message (and hence my request to add `structlog` to OF). It only dawned later on me that I can combine them to achieve cleaner error handling.

src/openforms/appointments/contrib/qmatic/client.py
sergei-maertens: Yeah, the service group `services` is a list of `ServiceDict` which lacks the information of the full service dict. Would've been nice if that returned `FullServiceDict` as then I'd only need to make a single API call but now we need two.

src/openforms/appointments/contrib/qmatic/tests/data/vcr_cassettes/ListAvailableProductsTests/ListAvailableProductsTests.test_listing_products_with_location_and_product_constraint.yaml
sergei-maertens: They are a mistake in the service factory usage, it should be set up with no auth instead of zgw auth. I reckon it's now producing a token for client id `""` and secret `""`. I checked them on jwt.io and realized that they're not an issue.

I'd prefer making a separate PR to address this so this can get merged, is that okay?

PR #3460, opened 2023-09-06
:sparkles: [#3187] prevent selecting duplicate form definitions

### LINKED ISSUE #3187: Prevent adding the same form definition to the same form multiple times

2.2.x

When selecting reusable form definitions, the form designer can add the same form definition multiple times to the same form.

We should prevent this.
too-many

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

bart-maykin:
fixes #3187 

- added line of code in the filtering of the selectbox options to filterout selected form definitions
- added playwright tests

### DIFF
--- a/src/openforms/forms/tests/e2e_tests/test_form_designer.py
+++ b/src/openforms/forms/tests/e2e_tests/test_form_designer.py
@@ -1369,3 +1369,278 @@ class AppointmentFormTests(E2ETestCase):
             self.assertIsNone(form.product)
 
         await assertState()
+
+
+class SelectReuseableFormDefinitionsTests(E2ETestCase):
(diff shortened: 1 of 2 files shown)


src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -1369,3 +1369,193 @@ def assertState():
             self.assertIsNone(form.product)
 
         await assertState()
+
+
+class SelectReuseableFormDefinitionsTests(E2ETestCase):
+    async def test_all_reuseable_form_definition_options_are_available(self):
```
sergei-maertens, line 1375: reading this test name I'd expect all form definitions to be visible in the dropdown, but the assertions check the opposite?

sergei-maertens (approved):

PR #3471, opened 2023-09-07
Support multi product / multi customer appointments in Qmatic

### LINKED ISSUE #3456: Implement multi-product/multi-person appointments for Qmatic

sergei-maertens:
Closes #3456

Customer details are duplicated for the product with the largest amount. Qmatic considers customers independently from their services (where JCC simply duplicates the product ID for every occurrence).

This means that 2 different products, one with amount 1 and the other amount 3 makes 3 customers in total, and the 3 customers is also applied to product 1. This can lead to some undesired wasted time, for which I will create a Taiga issue with the stakeholder.

### DIFF
--- a/src/openforms/appointments/contrib/qmatic/client.py
+++ b/src/openforms/appointments/contrib/qmatic/client.py
@@ -190,7 +190,7 @@ class QmaticClient(Session):
         self, location_id: str, service_ids: list[str], day: date, num_customers: int
     ) -> list[datetime]:
         """
-        Get list of available dates for multiple services and customers.
+        Get list of available times for multiple services and customers.
 
         ``num_customers`` is the total number of customers, which affects the
         appointment duration in Qmatic (using
--- a/src/openforms/appointments/contrib/qmatic/client.py
+++ b/src/openforms/appointments/contrib/qmatic/client.py
@@ -1,5 +1,8 @@
+from datetime import date
 from typing import TypedDict
 
+import pytz
+from dateutil.parser import isoparse
 from requests import Session
 
 from .exceptions import QmaticException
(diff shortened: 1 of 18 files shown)


src/openforms/appointments/contrib/qmatic/client.py
```diff
@@ -115,3 +141,85 @@ def list_service_groups(
...
+            isoparse(value).replace(tzinfo=branch_timezone).date()
+            for value in response.json()["dates"]
+        ]
+        return dates
+
+    def list_times(
+        self, location_id: str, service_ids: list[str], day: date, num_customers: int
+    ) -> list[datetime]:
+        """
+        Get list of available dates for multiple services and customers.
```
SilviaAmAm, line 193:
```suggestion
        Get list of available times for multiple services and customers.
```
?

src/openforms/appointments/contrib/qmatic/plugin.py
```diff
@@ -201,25 +242,37 @@ def get_dates(
...
-        date. The `start_at` and `end_at` arguments are ingored.
-        """
-        if len(products) != 1:
-            return []
+        From the documentation:
 
-        client = QmaticClient()
-        product_id = products[0].identifier
+            numberOfCustomers will be used on all services when calculating the
+            appointment duration For example, a service with Duration of 10 minutes and
```
SilviaAmAm, line 248:
```suggestion
            appointment duration. For example, a service with Duration of 10 minutes and
```

SilviaAmAm (approved):

src/openforms/appointments/contrib/qmatic/client.py
sergei-maertens: yes, got caught copy pasting :P

PR #3472, opened 2023-09-07
Catch errors in the admin for fields making outgoing requests

### LINKED ISSUE #3279: If ANY Catalog API is down, you can not configure a ZGW API set

2.2.0

Sentry ID: 335396

The ZGW API set wants to retrieve a list of all zaaktypen, from all catalogs. Maybe we should not do this at all, or make it asynchronous, and move it to a dropdown.

I see this pattern quite often. Realtime connections that, if they fail, crash the app. Maybe we need something like `admin_requests` that handles errors gracefully with an error message? Ofcourse it still is case-by-case implementation aspect on how to proceed further.

This is also true for the Appointment configuration > location field.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

bart-maykin:
fixes #3279 

Made a decorator that wraps a try, catch around a function and either returns it normally or returns a default value with a django message error

todo:
- write a nice explanation what the decorator does and should be used
- look into where we need to use it for now
- look into which exceptions the function should handle per usecase
- look into nice, understandable and none incriminating error messages

### DIFF
--- a/src/openforms/appointments/contrib/jcc/tests/test_admin.py
+++ b/src/openforms/appointments/contrib/jcc/tests/test_admin.py
@@ -1,0 +1,117 @@
+from unittest.mock import patch
+
+from django.urls import reverse
+from django.utils.translation import gettext as _
+
+import requests
+import requests_mock
+from django_webtest import WebTest
+
+from openforms.accounts.tests.factories import SuperUserFactory
+from openforms.appointments.contrib.jcc.models import JccConfig
+from openforms.appointments.models import AppointmentsConfig
+from openforms.utils.tests.cache import clear_caches
+from openforms.utils.tests.logging import disable_logging
+from soap.tests.factories import SoapServiceFactory
+
+from .utils import WSDL
+
+
+@disable_logging()
+class ApointmentConfigAdminTests(WebTest):
+    def setUp(self):
+        super().setUp()
+
+        self.addCleanup(clear_caches)
+
+    @requests_mock.Mocker()
+    def test_admin_while_service_is_down(self, m):
+        m.register_uri(
+            requests_mock.ANY, requests_mock.ANY, exc=requests.RequestException
+        )
+
+        appointments_config = AppointmentsConfig.get_solo()
+        appointments_config.plugin = "jcc"
+        appointments_config.limit_to_location = "test"
+        appointments_config.save()
+
+        config = JccConfig.get_solo()
+        config.service = SoapServiceFactory.create(url=str(WSDL))
+        config.save()
+
+        superuser = SuperUserFactory.create()
+        response = self.app.get(
+            reverse(
+                "admin:appointments_appointmentsconfig_change",
+                args=(AppointmentsConfig.singleton_instance_id,),
+            ),
+            user=superuser,
+        )
+        self.assertEqual(response.status_code, 200)
+
+        form = response.forms["appointmentsconfig_form"]
+
+        self.assertEqual(form["plugin"].value, "jcc")
+        self.assertEqual(form["limit_to_location"].value, "test")
+
+        error_node = response.pyquery(
+            ".field-limit_to_location .openforms-error-widget .openforms-error-widget__column small"
+        )
+        self.assertNotIn(
+            _(
+                "Could not load data - enable and check the request logs for more details."
+            ),
+            error_node.text(),
+        )
+
+        form.submit()
+
+        self.assertEqual(form["limit_to_location"].value, "test")
+
+    @requests_mock.Mocker()
+    @patch(
+        "openforms.appointments.contrib.jcc.client.build_client",
+        side_effect=requests.RequestException,
+    )
+    def test_admin_while_wsdl_is_down(self, m, mock_client):
+        m.register_uri(
+            requests_mock.ANY, requests_mock.ANY, exc=requests.RequestException
+        )
+
+        appointments_config = AppointmentsConfig.get_solo()
+        appointments_config.plugin = "jcc"
+        appointments_config.limit_to_location = "test"
+        appointments_config.save()
+
+        config = JccConfig.get_solo()
+        config.service = SoapServiceFactory.create(url=str(WSDL))
+        config.save()
+
+        superuser = SuperUserFactory.create()
+        response = self.app.get(
+            reverse(
+                "admin:appointments_appointmentsconfig_change",
+                args=(AppointmentsConfig.singleton_instance_id,),
+            ),
+            user=superuser,
+        )
+        self.assertEqual(response.status_code, 200)
+
+        form = response.forms["appointmentsconfig_form"]
+
+        self.assertEqual(form["plugin"].value, "jcc")
+        self.assertEqual(form["limit_to_location"].value, "test")
+
+        error_node = response.pyquery(
+            ".field-limit_to_location .openforms-error-widget .openforms-error-widget__column small"
+        )
+        self.assertIn(
+            _(
+                "Could not load data - enable and check the request logs for more details."
+            ),
+            error_node.text(),
+        )
+
+        form.submit()
+
+        self.assertEqual(form["limit_to_location"].value, "test")
--- a/src/openforms/appointments/contrib/qmatic/tests/test_admin.py
+++ b/src/openforms/appointments/contrib/qmatic/tests/test_admin.py
@@ -1,11 +1,18 @@
 from django.urls import reverse
+from django.utils.translation import gettext as _
 
+import requests
+import requests_mock
 from django_webtest import WebTest
 
 from openforms.accounts.tests.factories import SuperUserFactory
+from openforms.appointments.contrib.qmatic.models import QmaticConfig
+from openforms.appointments.models import AppointmentsConfig
 from openforms.utils.tests.cache import clear_caches
+from openforms.utils.tests.logging import disable_logging
 
 from ..constants import CustomerFields
+from .factories import ServiceFactory
 
 
 class QmaticConfigAdminTests(WebTest):
@@ -26,3 +33,53 @@ class QmaticConfigAdminTests(WebTest):
         for value in checked_field_values:
             with self.subTest(f"{value=}"):
                 self.assertIn(value, CustomerFields)
+
+
+@disable_logging()
+class ApointmentConfigAdminTests(WebTest):
+    def setUp(self):
+        super().setUp()
+
+        self.addCleanup(clear_caches)
+
+    @requests_mock.Mocker()
+    def test_admin_while_service_is_down(self, m):
+        appointments_config = AppointmentsConfig.get_solo()
+        appointments_config.plugin = "qmatic"
+        appointments_config.limit_to_location = "test"
+        appointments_config.save()
+
+        config = QmaticConfig.get_solo()
+        config.service = ServiceFactory.create()
+        config.save()
+
+        m.get(requests_mock.ANY, exc=requests.RequestException)
+
+        superuser = SuperUserFactory.create()
+        response = self.app.get(
+            reverse(
+                "admin:appointments_appointmentsconfig_change",
+                args=(AppointmentsConfig.singleton_instance_id,),
+            ),
+            user=superuser,
+        )
+        self.assertEqual(response.status_code, 200)
+
+        form = response.forms["appointmentsconfig_form"]
+
+        self.assertEqual(form["plugin"].value, "qmatic")
+        self.assertEqual(form["limit_to_location"].value, "test")
+
+        error_node = response.pyquery(
+            ".field-limit_to_location .openforms-error-widget .openforms-error-widget__column small"
+        )
+        self.assertNotIn(
+            _(
+                "Could not load data - enable and check the request logs for more details."
+            ),
+            error_node.text(),
+        )
+
+        form.submit()
+
+        self.assertEqual(form["limit_to_location"].value, "test")
--- a/src/openforms/registrations/contrib/zgw_apis/tests/test_admin.py
+++ b/src/openforms/registrations/contrib/zgw_apis/tests/test_admin.py
@@ -1,0 +1,66 @@
+from django.urls import reverse
+from django.utils.translation import gettext as _
+
+import requests
+import requests_mock
+from django_webtest import WebTest
+from privates.test import temp_private_root
+from zgw_consumers.api_models.constants import VertrouwelijkheidsAanduidingen
+
+from openforms.accounts.tests.factories import SuperUserFactory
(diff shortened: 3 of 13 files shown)


bart-maykin:
We can use this as a nice generic function to handle the error handling and messaging. Or we add a try, catch around the super call with the messaging.

src/openforms/utils/decorators.py
```diff
@@ -65,3 +67,36 @@ def _wrapped_view_func(request, *args, **kwargs):
         return _wrapped_view_func
 
     return decorator
+
+
+def mixin_exception_handler(
```
bart-maykin, line 72: Move to utils/admin.py

sergei-maertens: I checked that file and it's actually registering models for real, so perhaps we shouldn't pollute it with utilities. Instead, let's put this in `openforms.admin.decorators` :)

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+from django.test.client import RequestFactory
+from django.utils.translation import gettext_lazy as _
+
+import requests
+import zds_client
+
+from openforms.utils.decorators import dbfields_exception_handler
+
+
+class TestDbFieldsExceptionHandler(TestCase):
```
CharString, line 15:
```suggestion
class DbFieldsExceptionHandlerTests(TestCase):
```
consistent with the rest of the codebase.

src/openforms/utils/decorators.py
```diff
@@ -65,3 +68,44 @@ def _wrapped_view_func(request, *args, **kwargs):
...
+                    messages.error(request, error_message)
+
+                # return form.field instance for crashing field
+                # TODO: Dynamicly add widget to form.field with decorated ModelAdmin formfield_overrides
+                return db_field.formfield(
+                    help_text=db_field.help_text,
+                    validators=db_field.validators,
+                    max_length=db_field.max_length,
+                    disabled=True,
+                )
```
CharString, line 107: I like the addition of `disabled`. But do all fields have a `max_length`?

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+            return forms.CharField(
+                label="db field",
+                max_length=1000,
+                help_text=_("help text"),
+            )
+
+        request = self.factory.get("/test")
+
+        # setup django.messages
+        setattr(request, "session", "session")
```
CharString, line 37: Is this right? I'd expect `request.session` to be a `MutableMapping` like a dict. But I might be wrong.

I would consider pulling this piece of setup to `setUp`; have it create self.request instead of self.factory.

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,80 @@
+from django import forms
+from django.contrib.messages import get_messages
+from django.contrib.messages.storage.fallback import FallbackStorage
+from django.db import models
+from django.test import TestCase
```
CharString, line 5: We're not touching the database, `SimpleTestCase` is enough, and cheaper.

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+from django.test.client import RequestFactory
+from django.utils.translation import gettext_lazy as _
```
CharString, line 7: Why the translations for the test models?

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+        messages = FallbackStorage(request)
+        setattr(request, "_messages", messages)
+
+        form_field = formfield_for_dbfield(self.db_field, request)
+
+        messages = [m.message for m in get_messages(request)]
+        self.assertNotIn(
+            "Could not load data for field 'db field' - enable and check the request logs for more details",
+            messages,
+        )
```
CharString, line 47: This is an isolated unittest. We call the function with no side-effects, so nothing else should be writing messages.
I think expecting `get_messages(request) == []` is fine here.

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def test_raising_not_given_exception(self):
+        @dbfields_exception_handler(exceptions=(requests.exceptions.RequestException,))
+        def formfield_for_dbfield(db_field, request):
+            raise zds_client.ClientError()
+
+        request = self.factory.get("/test")
+
+        with self.assertRaises(zds_client.ClientError):
+            formfield_for_dbfield(self.db_field, request)
```
CharString, line 80: With the setUp change, we can assert messages is empty after this context.

CharString (changes requested):
A bit of polish. But nice overall.

I'm not sure if I want an extra test that actually tests the regression of the issue in the ticket.
@SilviaAmAm What do you think about that?

src/openforms/utils/decorators.py
```diff
@@ -65,3 +68,43 @@ def _wrapped_view_func(request, *args, **kwargs):
...
     return decorator
+
+
+def dbfields_exception_handler(
+    exceptions: tuple[Type[Exception]],
+):
+    """
+    A decorator that wraps a try catch around the formfields_for_dbfields function.
+    If the decorator catches an exception it will return a base formfield with
+    an error message discribing that an error has occured.
```
SilviaAmAm, line 79:
```suggestion
    A decorator that wraps a try catch around the formfields_for_dbfields function.
    If the decorator catches an exception it will return a base formfield with
    an error message describing that an error has occurred.
```

src/openforms/utils/decorators.py
```diff
@@ -65,3 +68,43 @@ def _wrapped_view_func(request, *args, **kwargs):
...
+                ).format(verbose_name=db_field.verbose_name)
+
+                # filter out duplicated messages that get generated because of the returning form.field
+                if error_message not in [
+                    x.message for x in messages.get_messages(request)
+                ]:
+                    messages.error(request, error_message)
+
+                # return form.field instance for crashing field
+                # TODO: Dynamicly add widget to form.field with decorated ModelAdmin formfield_overrides
```
SilviaAmAm, line 101:
```suggestion
                # TODO: Dynamically add widget to form.field with decorated ModelAdmin formfield_overrides
```

SilviaAmAm:
> I'm not sure if I want an extra test that actually tests the regression of the issue in the ticket.
SilviaAmAm What do you think about that?

I think it would be good to add it!

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,72 @@
...
+
+        form_field = TestClass.formfield_for_dbfield(
+            TestClass, db_field=self.db_field, request=self.request
+        )
+
+        self.assertIsInstance(form_field, forms.CharField)
+        self.assertNotIn(
+            "Could not load data - enable and check the request logs for more details",
+            form_field.widget.render("db field", None),
+        )
```
sergei-maertens, line 41: I don't know how I feel about this artificially constructed classes that pretend to be an admin class.

Can't we create real model admin classes and test those?

Something like:

```py
from django.contrib.admin import AdminSite
test_site = AdminSite(name="test")

@suppress_requests_errors(fields=["some_field"])
class TestAdmin(admin.ModelAdmin):
    pass

# bind a model to it
model_admin = TestAdmin(SomeModel, test_site)

# now you can test this using:
formfield = model_admin.formfield_for_db_field(Model._meta.get_field("some_field"), ...)
```

This would at least accurately describe the intended usage.

sergei-maertens (changes requested):

src/openforms/utils/decorators.py
```diff
@@ -65,3 +70,65 @@ def _wrapped_view_func(request, *args, **kwargs):
...
+
+                return html
+
+        kwargs["widget"] = CustomWidget
+        return db_field.formfield(**kwargs)
+
+    def _decorator(admin_class: type[admin.ModelAdmin]):
+        original_func = admin_class.formfield_for_dbfield
+        for field in fields:
+            assert admin_class.model._meta.get_field(field)
```
sergei-maertens, line 117: I don't like that this requires the `model = ...` to be set on the admin class. I understand why it's done, because moving it inside the `def _wrapped` is too late as a development aid and code coverage cannot help there either.

So I'd propose to pass the `model` as an argument to the `suppress_requests_errors` decorator itself:

```py
@suppress_requests_errors(AppointmentsConfig, fields=["limit_to_location"])
```
which makes performing this sanity check a lot more straightforward, without indirection and polluting the admin class itself.

So, why is this a bad approach? because you can actually register and admin class for multiple models. This is valid:

```py
@admin.register(SOAPService, StUFService, Service)
class MyAdmin(admin.ModelAdmin):
    search_fields = ("label",)
```
i.e. registering the same admin class for different models which have some shared functionality.

sergei-maertens (changes requested):

src/openforms/appointments/contrib/jcc/tests/test_admin.py
```diff
@@ -0,0 +1,104 @@
...
+            ),
+            user=superuser,
+        )
+        self.assertEqual(response.status_code, 200)
+
+        plugin = response.context["adminform"].form["plugin"]
+        self.assertEqual(plugin.value(), config.plugin)
+
+        limit_to_location = response.context["adminform"].form["limit_to_location"]
+        self.assertEqual(limit_to_location.value(), config.limit_to_location)
```
sergei-maertens, line 44: preferably you use `response.forms["..."]` which will give you a webtest response

e.g. assertion:

```py
self.assertEqual(form["plugin"], "jcc")
```

src/openforms/appointments/contrib/jcc/tests/test_admin.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_admin_while_service_is_down(self, m):
+        m.register_uri(
+            requests_mock.ANY, requests_mock.ANY, exc=requests.RequestException
+        )
+
+        config = AppointmentsConfig.get_solo()
+        assert isinstance(config, AppointmentsConfig)
+        config.plugin = "jcc"
+        config.limit_to_location = "test"
+        config.save()
```
sergei-maertens, line 28: I'm not sure it's wise to make use of `MockConfigMixin` here, since the test itself is asserting database state and effect of reading/writing config changes.

src/openforms/appointments/contrib/qmatic/tests/test_admin.py
```diff
@@ -26,3 +33,53 @@ def test_customer_fields_are_listed(self):
         for value in checked_field_values:
             with self.subTest(f"{value=}"):
                 self.assertIn(value, CustomerFields)
+
+
+@disable_logging()
+class ApointmentConfigAdminTests(MockConfigMixin, WebTest):
```
sergei-maertens, line 39: I think similar remarks apply here as with JCC

src/openforms/appointments/tests/test_appointment_config.py
sergei-maertens, line 1: This test file can be deleted in favour of the plugin-specific tests

src/openforms/registrations/contrib/zgw_apis/tests/test_zgw_api_config.py
```diff
@@ -544,3 +550,55 @@ def test_get_zgw_config(self, m):
             api_group = plugin.get_zgw_config({})
 
         self.assertEqual(api_group, self.zgw_group1)
+
+
+@disable_2fa
+@temp_private_root()
+class ZGWApiGroupConfigAdminTests(WebTest):
```
sergei-maertens, line 557: please move this in a separate test file

otherwise this will conflict hard with #3509

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return super().formfield_for_dbfield(db_field, request, **kwargs)
+
+
+class DbFieldsExceptionHandlerTests(SimpleTestCase):
+    def test_none_crashing_field(self):
+        model_admin = TestAdmin(User, AdminSite(name="test"))
+
+        formfield = model_admin.formfield_for_dbfield(
+            User._meta.get_field("username"), ...
```
sergei-maertens, line 28: please use the django test client request factory to pass a dummy request: https://docs.djangoproject.com/en/4.2/topics/testing/advanced/#the-request-factory

src/openforms/appointments/contrib/jcc/tests/test_admin.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(form["plugin"].value, "jcc")
+        self.assertEqual(form["limit_to_location"].value, "test")
+
+        error_node = response.pyquery(
+            ".field-limit_to_location .openforms-error-widget .openforms-error-widget__column small"
+        )
+        self.assertIn(
+            "Could not load data - enable and check the request logs for more details",
+            error_node.text(),
+        )
```
sergei-maertens, line 108: this assertion will crash when we define the translations - use the `gettext("Could not load data - ...")` call in tests instead.

src/openforms/utils/tests/test_decorators.py
```diff
@@ -0,0 +1,62 @@
...
+
+        formfield = model_admin.formfield_for_dbfield(
+            User._meta.get_field("username"), self.request
+        )
+
+        self.assertIsInstance(formfield, forms.CharField)
+        self.assertNotIn(
+            "Could not load data - enable and check the request logs for more details",
+            formfield.widget.render("username", None),
+        )
```
sergei-maertens, line 37: this assertion will pass when we add translations and the Dutch translation would be present.

Can't you rewrite this assertion as something like:

```py
self.assertHTMLEqual(
    rendered_widget,
    "<html here>",
)
```

sergei-maertens (changes requested):

src/openforms/registrations/contrib/zgw_apis/tests/test_admin.py
```diff
@@ -0,0 +1,66 @@
...
+                _(
+                    "Could not load data - enable and check the request logs for more details."
+                ),
+            )
+
+            form.submit()
+
+            self.assertEqual(
+                form["zaaktype"].value, "https://catalogi-1.nl/api/v1/zaaktypen/1"
+            )
```
sergei-maertens, line 66: this is not asserting the result of the form submission, it should be:

```py
zgw_group.refresh_from_db()
self.assertEqual(zgw_group.zaaktype, "https://catalogi-1.nl/api/v1/zaaktypen/1")
```

PR #3479, opened 2023-09-11
[#3443] Date validation

### LINKED ISSUE #3443: Date component - no error for min/max date input.

2.3.0

DH: 577
Date components with validation for minimum and maximum dates do not inform the user about this requirement. 
It is clear when clicking the component, because out of range dates will not be available, but if you type in a date that is outside the available range the component will clear itself without informing the user why.

SilviaAmAm:
Fixes #3443 

Frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/548


src/openforms/forms/tests/test_migrations.py
```diff
@@ -1102,3 +1102,61 @@ def test_the_custom_error_is_added(self):
...
+        self.form_def = FormDefinition.objects.create(
+            name="Date",
+            slug="date",
+            configuration={
+                "components": [
+                    {
+                        "key": "dateComponent",
+                        "type": "date",
+                        "label": "Date component",
+                        "customOptions": {},
```
sergei-maertens, line 1126: let's add a component in this test that doesn't have this key at all to verify my comment above :)

src/openforms/js/components/form/date.js
```diff
@@ -54,6 +55,12 @@ class DateField extends DateTimeField {
             },
           },
         },
+        customOptions: {
+          // Issue #3443 - Flatpickr is preventing the value from being set when entering it manually if it is outside the set minDate/maxDate range.
+          // With allowInvalidPreload: true, we tell it to set the value also if it is invalid, so that then validators can kick in.
+          // https://github.com/flatpickr/flatpickr/blob/master/src/index.ts#L2498
```
sergei-maertens, line 61: I think this is better for accessibility in general too.

sergei-maertens (changes requested):

sergei-maertens (approved):
I pushed a fix for the reverse test suite failure to master, so if you rebase that failure should go away!

PR #3480, opened 2023-09-12
[#3470] Handle large form names which contain special characters

### LINKED ISSUE #3470: "/" in form title causes submission to fail: SuspiciousFileOperation in generate_submission_report

2.2.x - 2.3.0

When a form title contains a "/" this title is used as is to store sumbission pdf's which it cannot do. 
Subsequently the registration is unable to find the pdf

form title: foo/bar
dir_name = submission-reports/2023/09/07/foo
file_ext = '.pdf'
file_name = 'bar.pdf'
name = 'foo/bar.pdf'

SuspiciousFileOperation: Storage can not find an available filename for "submission-reports/2023/09/07/foo/bar_wb1x7PV.pdf". Please make sure that the corresponding file field allows sufficient "max_length".

sentry: 332615

vaszig:
Fixes #3470

### DIFF
--- a/src/openforms/submissions/models/submission_report.py
+++ b/src/openforms/submissions/models/submission_report.py
@@ -2,6 +2,7 @@ from typing import Optional
 
 from django.core.files.base import ContentFile
 from django.db import models
+from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _, override
 
 from celery.result import AsyncResult
@@ -74,7 +75,7 @@ class SubmissionReport(models.Model):
             )
             self.content = ContentFile(
                 content=pdf_report,
-                name=f"{form.name}.pdf",  # Takes care of replacing spaces with underscores
+                name=f"{slugify(form.name)}.pdf",
             )
         self.save()
         return html_report
--- a/src/openforms/submissions/models/submission_report.py
+++ b/src/openforms/submissions/models/submission_report.py
@@ -2,7 +2,6 @@ from typing import Optional
 
 from django.core.files.base import ContentFile
 from django.db import models
-from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _, override
 
 from celery.result import AsyncResult
@@ -75,7 +74,7 @@ class SubmissionReport(models.Model):
             )
             self.content = ContentFile(
                 content=pdf_report,
-                name=f"{slugify(form.name)}.pdf",
+                name=f"{form.slug}.pdf",
             )
         self.save()
         return html_report
--- a/src/openforms/submissions/tests/test_models.py
+++ b/src/openforms/submissions/tests/test_models.py
@@ -630,12 +630,7 @@ class PDFSubmissionReportTests(HypothesisTestCase):
             alphabet=st.characters(blacklist_characters="\x00", codec="utf-8"),
         )
     )
-    # see ticket #3470 - explicit value for form name (the final file name will be > max_length)
-    # Django's get_available_name truncates original name if required, so with a large
-    # name which has a '/' as well it confuses file_root (which is truncated)
-    @example(
-        "A large,really large name for the form in order to catch the above exception foo/bar"
-    )
+    # see ticket #3470
     @settings(deadline=500)
     def test_names_donot_break_pdf_generation(self, form_name):
         report = SubmissionReportFactory.create(submission__form__name=form_name)
--- a/src/openforms/submissions/tests/test_models.py
+++ b/src/openforms/submissions/tests/test_models.py
@@ -3,7 +3,7 @@ import os
 from unittest.mock import patch
 
 from django.core.exceptions import ValidationError
-from django.test import TestCase, override_settings
+from django.test import TestCase, override_settings, tag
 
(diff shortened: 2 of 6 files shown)


src/openforms/submissions/tests/test_models.py
```diff
@@ -617,3 +620,25 @@ def test_clear_execution_state_without_execution_state(self):
...
+class PDFSubmissionReportTests(HypothesisTestCase):
+    @given(
+        st.text(
+            min_size=1,
+            alphabet=st.characters(blacklist_characters="\x00", codec="utf-8"),
+        )
+    )
+    # see ticket #3470 - explicit value for form name (the final file name will be > max_length)
+    # Django's get_available_name truncates original name if required, so with a large
+    # name which has a '/' as well it confuses file_root (which is truncated)
```
SilviaAmAm, line 635: I don't really understand what this means :thinking: What I understand:
- When creating the `ContentFile`, we now slugify the form name (so the `/` problem is solved). I guess the length of the slugified form name is <= 150 (the max length of the form name).
- On save, the storage class possibly does some additional changes to the filename: https://github.com/django/django/blob/3.2.21/django/core/files/storage.py#L53
But I don't understand what causes the problem, since by now the slashes should have been removed?

vaszig: I added this comment in order to give some content for the specific error. I can remove it if you think it's redundant.

And yes, you are correct, with the slugified version we won't have a problem.

SilviaAmAm: No no! It's more that I didn't really understand what the comment meant! :sweat_smile: 
If with the slugified version this is no longer a problem, can we remove this test altogether?

SilviaAmAm: Ah no I see! I would then maybe just change the comment to not include the mention to the `/` since that is no longer a problem?

src/openforms/submissions/tests/test_models.py
```diff
@@ -617,3 +620,20 @@ def test_clear_execution_state_without_execution_state(self):
...
+
+@temp_private_root()
+class PDFSubmissionReportTests(HypothesisTestCase):
+    @given(
+        st.text(
+            min_size=1,
+            alphabet=st.characters(blacklist_characters="\x00", codec="utf-8"),
+        )
+    )
+    # see ticket #3470
```
sergei-maertens, line 633: you can tag it appropriately as well: `@tag("gh-3470")`

src/openforms/submissions/tests/test_models.py
```diff
@@ -617,3 +620,20 @@ def test_clear_execution_state_without_execution_state(self):
...
+    @settings(deadline=500)
+    def test_names_donot_break_pdf_generation(self, form_name):
```
sergei-maertens, line 635:
```suggestion
    def test_names_do_not_break_pdf_saving_to_disk(self, form_name):
```
slightly better name to address the problem - the PDF generation itself works, but saving it to disk fails

src/openforms/submissions/tests/test_models.py
```diff
@@ -617,3 +620,20 @@ def test_clear_execution_state_without_execution_state(self):
...
+        report = SubmissionReportFactory.create(submission__form__name=form_name)
+        report.generate_submission_report_pdf()
+
+        self.assertTrue(report.content)
```
sergei-maertens, line 639: might even make it more obvious that the file exists on disk:

```py
self.assertTrue(report.content.storage.exists(report.content.name))
```
`report.content` (if you stringify it) only returns the path where it should be, but that doesn't guarantee yet that it exists on disk.

src/openforms/submissions/models/submission_report.py
```diff
@@ -74,7 +75,7 @@ def generate_submission_report_pdf(self) -> str:
             )
             self.content = ContentFile(
                 content=pdf_report,
-                name=f"{form.name}.pdf",  # Takes care of replacing spaces with underscores
+                name=f"{slugify(form.name)}.pdf",
```
sergei-maertens, line 78: is there a reason you use `slugify` rather than using `form.slug` for this?

vaszig: I thought it would be better in case the form name has been updated. Should this be consistent only with the `form.slug`?

sergei-maertens: I think it makes little difference - the form slug is visible in the browser URL anyway. The filename of the PDF is just some courtesy, either approach works for me!

sergei-maertens (approved):

sergei-maertens:
Backports:

* `stable/2.3.x`: c434e4b1e9bbf58c6a04fb2fd9e04621f26c1cf5
* `stable/2.2.x`: c3d5ac52b32a4948be655b6bbe06725800d5eec1
* `stable/2.1.x`: 6f3195783bf36c53ac3fbf8659f82d3e881dc1ba
* `stable/2.0.x`: 1872770f3001be24d327ec2e303af946b05aae95

PR #3482, opened 2023-09-13
[#3413] Allow form link in cosign emails

### LINKED ISSUE #3413: Allow link in co-sign e-mails
### Thema / Theme

Admin

internal: Venr 71, rid 142

Allow OF users to decide if they want to include links in co-sign request e-mails.
As the feature currently does not support links in these e-mails the co-sign flow seems very basic and not user friendly. 
OF users should be able to make the choice whether they want to adhere to not include links in messages (like digid/mijnoverheid) or not. 

Added value / Toegevoegde waarde

UX!

vaszig:
Fixes #3413


src/openforms/emails/templates/emails/cosign.html
```diff
@@ -0,0 +1,25 @@
+{% load static i18n %}
+
+<p>
+    {% blocktranslate %}
+        This is a request to co-sign form "{{ form_name }}".<br />
```
SilviaAmAm, line 5: Does this use the right language of the form name?

src/openforms/emails/templates/emails/cosign.html
```diff
@@ -0,0 +1,25 @@
...
+    {% endblocktranslate %}
+    
+    {% if show_form_link %}
+        {% blocktranslate %}
+            Please click the link below to co-sign:<br />
+            <a class="utrecht-link utrecht-link--openforms" href="{{ form_url }}">Co-sign</a><br />
```
SilviaAmAm, line 11: Maybe the text of the message could be the same as for when the link is not enabled, except with an extra:
```
You can go to the form page also by clicking on <a class="utrecht-link utrecht-link--openforms" href="{{ form_url }}">this limk</a>
```
or something... :stuck_out_tongue: Maybe @LaurensBurger has a better suggestion

src/openforms/submissions/tasks/co_sign.py
```diff
@@ -30,17 +32,14 @@ def send_email_cosigner(submission_id: int) -> None:
...
-            "After authentication, fill in the following code to retrieve the form submission:\n\n{{ code }}\n\n."
-        ),
-        context={
+    template = get_template("emails/cosign.html")
+    content = template.render(
+        {
             "code": submission.public_registration_reference,
             "form_name": submission.form.name,
-        },
+            "form_url": submission.form.get_absolute_url(),
```
SilviaAmAm, line 40:
```suggestion
            "form_url": submission.form_url,
```
This needs to be the link to the form on the frontend

src/openforms/submissions/tests/test_task_co_sign.py
```diff
@@ -97,6 +98,48 @@ def test_happy_flow(self):
...
+                {
+                    "key": "cosign",
+                    "type": "cosign",
+                    "label": "Cosign component",
+                },
+            ],
+            submitted_data={"cosign": "test@test.nl"},
+        )
+
+        send_email_cosigner(submission.id)
```
SilviaAmAm, line 113: Try to patch the call to get the configuration, I think this is the problem for the reverse test suite.
(and do the same for the test below)

SilviaAmAm (changes requested):

src/openforms/emails/templates/emails/cosign.html
vaszig: I tested this by overriding `LANGUAGE_CODE` and it works. I hope this is what you mean. Maybe I should another test for this as well?

src/openforms/emails/templates/emails/cosign.html
```diff
@@ -0,0 +1,25 @@
...
+<p>
+    {% blocktranslate %}
+        This is a request to co-sign form "{{ form_name }}".<br />
+    {% endblocktranslate %}
+    
+    {% if show_form_link %}
+        {% blocktranslate %}
+            You can go to the form page also by clicking on:<br />
+            <a class="utrecht-link utrecht-link--openforms" href="{{ form_url }}">this link</a><br />
+        {% endblocktranslate %}
```
sergei-maertens, line 12:
```suggestion
        {% blocktranslate trimmed %}
            Please visit the form page by navigating to the following link: {{ form_url|urlize }}.
        {% endblocktranslate %}
```
`urlize` ensures that it's a clickable link. The CSS class names won't have an effect in e-mail templates, so there's not much point in using these.

Keep in mind that plain text versions of HTML emails are also automatically created if not explicitly provided, and those strip out the HTML tags, so this variant still has the URL in the body for users to copy and paste in their browser if they use the plain text email.

src/openforms/submissions/tasks/co_sign.py
```diff
@@ -30,17 +32,14 @@ def send_email_cosigner(submission_id: int) -> None:
...
-    # Send Co-sign email # TODO add configurable templates
-    content = render_email_template(
-        ugettext(
-            'This is a request to co-sign form "{{ form_name }}". Please go to the webpage of this form on our '
-            "website and click on the 'co-sign' button. You will then be redirected to authenticate yourself."
-            "After authentication, fill in the following code to retrieve the form submission:\n\n{{ code }}\n\n."
-        ),
-        context={
+    template = get_template("emails/cosign.html")
+    content = template.render(
```
sergei-maertens, line 36: as remarked by Silvia, I don't think this will actually use the submission language for the context since no translation is activated here. See `openforms.emails.confirmation_emails.get_confirmation_email_templates` for a reference.

sergei-maertens (changes requested):

src/openforms/submissions/tasks/co_sign.py
```diff
@@ -23,39 +25,39 @@
...
-            text_message=content,
-        )
-    except Exception:
-        logevent.cosigner_email_queuing_failure(submission)
-        raise
 
-    logevent.cosigner_email_queuing_success(submission)
+        try:
+            send_mail_html(
+                subject=f"Co-sign request for {submission.form.name}",
```
sergei-maertens, line 50: Shouldn't this also be translatable? :sweat_smile:

sergei-maertens (approved):

PR #3488, opened 2023-09-18
Provide generic APIClient implementation taking care of mTLS

### LINKED ISSUE #3328: Qmatic plugin does not use mtls

sergei-maertens:
... and other common configurations.

Closes #3328 

Will create a separate PR to refactor all other clients and document the preferred approach.


src/api_client/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+- specify a base URL and check that absolute URL requests fit within the base URL
+- manage resources by closing the session even for one-off requests
+
+There are two ways to instantiate a client.
+
+- from a factory (preferred)
+- manually
+
+The factory approach is preffered as it provides the most robust way to honour authentication
```
sergei-maertens, line 20:
```suggestion
The factory approach is preferred as it provides the most robust way to honour authentication
```

src/api_client/tests/test_client_api.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @given(http_methods)
+    def test_encouraged_usage_with_context_do_not_close_prematurely(self, method):
+        client = APIClient("https://example.com")
+
+        with patch.object(client, "close", wraps=client.close) as mock_close:
+            with requests_mock.Mocker() as m, client:
+                m.register_uri(requests_mock.ANY, requests_mock.ANY)
+
+                client.request(method, "foo")
```
CharString, line 195: Either go full nesting (for backwards compatibility) or use the modern way; on first pass my eye glanced over `client` being used as a context in this mix.
```suggestion
        with (
            patch.object(client, "close", wraps=client.close) as mock_close,
            requests_mock.Mocker() as m,
            client,
        ):
                m.register_uri(requests_mock.ANY, requests_mock.ANY)

                client.request(method, "foo")
```

src/openforms/appointments/contrib/qmatic/client.py
```diff
@@ -63,6 +66,19 @@ class BranchDetailDict(TypedDict):
 # API CLIENT IMPLEMENTATIONS, per major version of the API
 
 
+def QmaticClient() -> "Client":
```
CharString, line 69: Feels weird that a class that used to be a `Session` subclass is now a camelcase function. Should this be the `__new__`?

CharString (changes requested):
Looks ok. I've been moving some things from StuF client to the SOAP client. #3490

sergei-maertens:
> Looks ok. I've been moving some things from StuF client to the SOAP client. #3490

For completeness sake - we're coordinating on Slack what to move where, I will set up a separate PR for that so that we are both unblocked by that reorganization.

src/api_client/tests/test_client_api.py
```diff
@@ -0,0 +1,202 @@
...
+
+http_methods = st.one_of(
+    st.just("GET"),
+    st.just("OPTIONS"),
+    st.just("HEAD"),
+    st.just("POST"),
+    st.just("PUT"),
+    st.just("PATCH"),
+    st.just("DELETE"),
+)
```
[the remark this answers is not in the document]
sergei-maertens: ah, I was looking for something like that! afaik there's no enums defined in requests for that (and it also doesn't handle possible extensions like WebDAV which adds new HTTP methods)

src/api_client/client.py
```diff
@@ -0,0 +1,110 @@
...
+def has_same_base(url: furl, reference: furl) -> bool:
+    if not all(
+        (
+            getattr(url, attr) == getattr(reference, attr)
+            for attr in ("scheme", "username", "password", "host", "port")
+        )
+    ):
+        return False
+    # finally, if all that is the same, the path base should match too
+    return str(url).startswith(str(reference))
```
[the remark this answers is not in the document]
sergei-maertens: I do not find that more readable tbh.

And thinking about this a bit more... the `startswith` check itself is probably sufficient, if we establish that `reference` has a scheme and netloc, no?

src/openforms/appointments/contrib/qmatic/client.py
sergei-maertens: It's a factory that pretends to be a class to not break the public API. I think `__new__` is appropriate, I'll give it a try.

CharString (approved):

PR #3492, opened 2023-09-19
Update django.po

LaurensBurger:
help text still suggested using {{ submission_reference }} instead of {{ public_reference }}


sergei-maertens:
Can you please follow the correct procedure to update these translations?

1. Update the source string, in this case it's defined in `openforms/registrations/contrib/email/config.py:45`
2. Re-generate the message catalog using `./bin/makemessages.sh`
3. Commit the relevant changes

I think it'd be best to just do step 1 and we'll sort out the translations when preparing the release, as `makemessages` will pick up many other new/changed translations.

PR #3493, opened 2023-09-19
[#3481] Escape "special" characters when saving data in the Objects API

### LINKED ISSUE #3481: Escape "special" characters when saving data in the Objects API
*Disclaimer: Conceptually, I'm against - what I find - unnecessary escaping, just to make sure OTHER applications do not make mistakes but Den Haag wants this, so we add optional support for it.*

Currently, when you have a form with a textfield, let's say "name", and you fill in `<script>alert();</script>`, this value is eventually sent as-is to the Objects API. For example, the Objects API will have data like this:

```json
{
  "data": {
    "name": "<script>alert();</script>"
  }
}
```
Although in Open Forms, there is no problem whatsoever, an external application reading this data *might* interpret this as HTML and show an actual alert-box in the browser.

To prevent this, ~the external application should be fixed~ we need to store the data in the Objects API, with HTML-characters escaped, like this:

```json
{
  "data": {
    "name": "&lt;script&gt;alert();&lt;/script&gt;"
  }
}
```

So, we'll introduce an environment variable `ESCAPE_REGISTRATION_OUTPUT` that will perform HTML escaping on all data sent to the Objects API. By default the value should be `False` (ie. no escaping).

**Tasks**

* [ ] Add `ESCAPE_REGISTRATION_OUTPUT` environment var/setting, default `False`
* [ ] Use `html.escape` to escape all **form** values, passed to the Objects API.
* [ ] BONUS points: Add this in such a way, it works for ALL registration output.

vaszig:
Fixes #3481

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
+++ b/src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
@@ -1,4 +1,4 @@
-from django.test import TransactionTestCase
+from django.test import TransactionTestCase, override_settings
 
 from openforms.forms.tests.factories import FormFactory
 from openforms.registrations.constants import RegistrationAttribute
@@ -10,13 +10,15 @@ from openforms.template import openforms_backend, render_from_string
 
 
 class JsonSummeryTests(TransactionTestCase):
-    def test_render_json_summary(self):
-        form = FormFactory.create()
-        submission = SubmissionFactory.create(
-            form=form, completed=True, form_url="http://maykinmedia.nl/myform/"
+    def setUp(self):
+        self.form = FormFactory.create()
+        self.submission = SubmissionFactory.create(
+            form=self.form, completed=True, form_url="http://maykinmedia.nl/myform/"
         )
+
+    def test_render_json_summary(self):
         SubmissionStepFactory.create(
-            form_step__form=form,
+            form_step__form=self.form,
             form_step__form_definition__slug="json-summery-step",
             form_step__form_definition__configuration={
                 "display": "form",
@@ -44,7 +46,7 @@ class JsonSummeryTests(TransactionTestCase):
                     },
                 ],
             },
-            submission=submission,
+            submission=self.submission,
             data={
                 "voornaam": "Foo",
                 "achternaam": "Bar",
@@ -54,12 +56,80 @@ class JsonSummeryTests(TransactionTestCase):
 
         rendered = render_from_string(
             "{% json_summary %}",
-            context={"_submission": submission},
+            context={"_submission": self.submission},
             backend=openforms_backend,
             disable_autoescape=True,
         )
 
         expected = '{"json-summery-step": {"voornaam": "Foo", "achternaam": "Bar", "tussenvoegsel": "de"}}'
+
+        self.assertEqual(rendered, expected)
+
+    @override_settings(ESCAPE_REGISTRATION_OUTPUT=False)
+    def test_render_json_summary_doesnt_escape_html_when_disabled(self):
+        SubmissionStepFactory.create(
+            form_step__form=self.form,
+            form_step__form_definition__slug="json-summery-step",
+            form_step__form_definition__configuration={
+                "display": "form",
+                "components": [
+                    {
+                        "key": "voornaam",
+                        "type": "textfield",
+                        "registration": {
+                            "attribute": RegistrationAttribute.initiator_voornamen,
+                        },
+                    }
+                ],
+            },
+            submission=self.submission,
+            data={
+                "voornaam": "<script>alert();</script>",
+            },
+        )
+
+        rendered = render_from_string(
+            "{% json_summary %}",
+            context={"_submission": self.submission},
+            backend=openforms_backend,
+            disable_autoescape=True,
+        )
+
+        expected = '{"json-summery-step": {"voornaam": "<script>alert();</script>"}}'
+
+        self.assertEqual(rendered, expected)
+
+    @override_settings(ESCAPE_REGISTRATION_OUTPUT=True)
+    def test_render_json_summary_escapes_html_when_enabled(self):
+        SubmissionStepFactory.create(
+            form_step__form=self.form,
+            form_step__form_definition__slug="json-summery-step",
+            form_step__form_definition__configuration={
+                "display": "form",
+                "components": [
+                    {
+                        "key": "voornaam",
+                        "type": "textfield",
+                        "registration": {
+                            "attribute": RegistrationAttribute.initiator_voornamen,
+                        },
+                    }
+                ],
+            },
+            submission=self.submission,
+            data={
+                "voornaam": "<script>alert();</script>",
+            },
+        )
+
+        rendered = render_from_string(
+            "{% json_summary %}",
+            context={"_submission": self.submission},
+            backend=openforms_backend,
+            disable_autoescape=True,
+        )
+
+        expected = '{"json-summery-step": {"voornaam": "&lt;script&gt;alert();&lt;/script&gt;"}}'
 
         self.assertEqual(rendered, expected)
 
--- a/src/openforms/registrations/contrib/objects_api/tests/test_utils.py
+++ b/src/openforms/registrations/contrib/objects_api/tests/test_utils.py
@@ -1,0 +1,34 @@
+from django.test import TestCase
+
+from ..utils import escape_html_manually
+
+
+class EscapeHTMLTests(TestCase):
+    def test_given_dict_values_are_escaped(self):
+        sample = {
+            "test1": "normal value",
+            "test2": "<script>alert();</script>",
+            "nested1": {
+                "n_test": {"escape": "<script>alert();</script>"},
+            },
+        }
+        expected = {
+            "test1": "normal value",
+            "test2": "&lt;script&gt;alert();&lt;/script&gt;",
+            "nested1": {
+                "n_test": {"escape": "&lt;script&gt;alert();&lt;/script&gt;"},
+            },
+        }
+
+        result = escape_html_manually(sample)
+
+        self.assertEqual(expected, result)
+
+    def test_not_type_of_dict_returns_empty_dict(self):
+        samples = [["test"], "test", 6]
+
+        for sample in samples:
+            with self.subTest(sample=sample):
+                result = escape_html_manually(sample)
+
+                self.assertEqual(result, {})
--- a/src/openforms/registrations/contrib/objects_api/utils.py
+++ b/src/openforms/registrations/contrib/objects_api/utils.py
@@ -1,0 +1,19 @@
+from django.utils.html import escape
+
+from openforms.typing import JSONValue
+
+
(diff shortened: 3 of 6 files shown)


src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -1873,3 +1871,146 @@ def match_attachment(request):
...
+                        f"{step_slug}": {
+                            "voornaam": "&lt;script&gt;alert();&lt;/script&gt;",
+                        }
+                    },
+                    "type": objects_form_options["productaanvraag_type"],
+                    "submission_id": str(submission.uuid),
+                    "language_code": "en",
+                    "attachments": [],
+                    "pdf_url": "https://documenten.nl/api/v1/enkelvoudiginformatieobjecten/1",
+                },
```
sergei-maertens, line 1926: this was confusing for me when reading the text, because the shape of data does not match with what is configured in the template! I'd align this with what is actually sent to the objects API.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -1873,3 +1871,146 @@ def match_attachment(request):
             document_create_attachment_body["vertrouwelijkheidaanduiding"], "geheim"
         )
         self.assertEqual(document_create_attachment_body["titel"], "A Custom Title")
+
+    @override_settings(ESCAPE_REGISTRATION_OUTPUT=True)
+    def test_submission_with_objects_api_escapes_html(self, m):
```
sergei-maertens, line 1876: I think I would simplify this test a bit with an explicit content JSON template. This also gives the opportunity to test that manual context usage is escaped (right now the test mostly asserts that the `{% json_summary %}` output is escaped).

To do that, I would use a custom `content_json` template, which you can pass via `objects_form_options`:

```py
content_template = """
{
  "summary": {% json_summary %},
  "manual_variable": "{{ voornaam }}"
}
"""
objects_form_options = {
    ...,
    "content_json": content_template,
}
```

You can then make the assertion on the request body for the part that's relevant:

```py
posted_record_data = object_create_body["record"]["data"]
expected_record_data = {
    "summary": {
      step_slug: {
        "voornaam": "&lt;script&gt;alert();&lt;/script&gt;",
      },
    },
    "manual_variable": "&lt;script&gt;alert();&lt;/script&gt;"
}
```

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -10,13 +10,15 @@
 
 
 class JsonSummeryTests(TransactionTestCase):
```
sergei-maertens, line 12:
```suggestion
class JsonSummaryTests(TestCase):
```

I ran this locally and don't see why it would need a `TransactionTestCase` - those are much slower (2.505s vs. 0.082s on my system!)

And you can also fix the spelling mistake :)

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -10,13 +10,15 @@
...
-    def test_render_json_summary(self):
-        form = FormFactory.create()
-        submission = SubmissionFactory.create(
-            form=form, completed=True, form_url="http://maykinmedia.nl/myform/"
+    def setUp(self):
+        self.form = FormFactory.create()
+        self.submission = SubmissionFactory.create(
+            form=self.form, completed=True, form_url="http://maykinmedia.nl/myform/"
         )
```
sergei-maertens, line 17: Generally I'm not a fan of breaking this out into `setUp` methods, because it makes it harder to follow what is actually being set up if you look at the test code by itself.

I think this is also some poor factory usage in the original test, I'd think that this is better:

```py
form = FormFactory.create()
        step = SubmissionStepFactory.create(
            submission__form=form,
            submission__form_url="https://example.com/some-form",
            submission__completed=True,
            form_step__form=form,
            form_step__form_definition__slug="json-summery-step",
            form_step__form_definition__configuration={
                "display": "form",
                "components": [
                    {
                        "key": "voornaam",
                        "type": "textfield",
                        "registration": {
                            "attribute": RegistrationAttribute.initiator_voornamen,
                        },
                    },
                    {
                        "key": "achternaam",
                        "type": "textfield",
                        "registration": {
                            "attribute": RegistrationAttribute.initiator_geslachtsnaam,
                        },
                    },
                    {
                        "key": "tussenvoegsel",
                        "type": "textfield",
                        "registration": {
                            "attribute": RegistrationAttribute.initiator_tussenvoegsel,
                        },
                    },
                ],
            },
            data={
                "voornaam": "Foo",
                "achternaam": "Bar",
                "tussenvoegsel": "de",
            },
        )
```
and the lines below can use `step.submission` rather than `submission` or `self.submission`.

You can then do a similar thing in your own test, so you set up only the data that is actually needed and reading/debugging a test requires you to only look at the test itself.

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -63,6 +65,74 @@ def test_render_json_summary(self):
...
+            form_step__form_definition__slug="json-summery-step",
+            form_step__form_definition__configuration={
+                "display": "form",
+                "components": [
+                    {
+                        "key": "voornaam",
+                        "type": "textfield",
+                        "registration": {
+                            "attribute": RegistrationAttribute.initiator_voornamen,
+                        },
```
sergei-maertens, line 81: I think this configuration doesn't actually do anything, so feel free to leave it out.

There actually is a big shortcut to create submissions with particular data:

```py
    components_list=[{"key": "voornaam", "type": "textfield"}],
    submitted_data={"voornaam": "<script>alert();</script>"},
    form_definition_kwargs={"slug": "formstep-slug"},
    completed=True,
)
```
this will create the form, form step, submission, submission step and set the provided data.

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -63,6 +65,74 @@ def test_render_json_summary(self):
...
+        )
+
+        rendered = render_from_string(
+            "{% json_summary %}",
+            context={"_submission": self.submission},
+            backend=openforms_backend,
+            disable_autoescape=True,
+        )
+
+        expected = '{"json-summery-step": {"voornaam": "<script>alert();</script>"}}'
```
sergei-maertens, line 98: with the above code suggestion, the slug of the form is changed a bit:

```suggestion
        expected = '{"formstep-slug": {"voornaam": "<script>alert();</script>"}}'
```

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -63,6 +65,74 @@ def test_render_json_summary(self):
...
+                        "type": "textfield",
+                        "registration": {
+                            "attribute": RegistrationAttribute.initiator_voornamen,
+                        },
+                    }
+                ],
+            },
+            submission=self.submission,
+            data={
+                "voornaam": "<script>alert();</script>",
```
sergei-maertens, line 121: I'd also like to see a double code injected in there somewhere, as the HTML escaping and JSON escaping of double quotes is... interesting. Please also add it to the test with disabled escaping.

src/openforms/registrations/contrib/objects_api/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
+from django.test import TestCase
+
+from ..utils import escape_html_manually
+
+
+class EscapeHTMLTests(TestCase):
+    def test_given_dict_values_are_escaped(self):
+        sample = {
+            "test1": "normal value",
+            "test2": "<script>alert();</script>",
```
sergei-maertens, line 10: maybe the test input can be a bit more aimed at the html entities involved:

```suggestion
            "test2": """<>'"&""",
```
so that covers the characters <kbd><</kbd>, <kbd>></kbd>, <kbd>&</kbd>, <kbd>'</kbd>, <kbd>"</kbd>, which are HTML-unsafe

src/openforms/registrations/contrib/objects_api/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
...
+            "nested1": {
+                "n_test": {"n_test1": {"escape": "<script>alert();</script>"}},
+            },
+        }
```
sergei-maertens, line 14: I'm a missing a case with lists:

```py
"nested_list": [
  "escape < me",
  {"leave > me": "escape & me"},
],
```
with expected output:

```py
"nested_list": [
  "escape &lt; me",
  {"leave > me": "escape &amp; me"},
],
```

src/openforms/registrations/contrib/objects_api/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
+from django.test import TestCase
+
+from ..utils import escape_html_manually
+
+
+class EscapeHTMLTests(TestCase):
```
sergei-maertens, line 6: I'd also like to see some tests with escapable keys that assert the keys are not actually escaped!

src/openforms/registrations/contrib/objects_api/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
...
+                },
+            },
+        }
+
+        result = escape_html_manually(sample)
+
+        self.assertEqual(expected, result)
+
+    def test_not_type_of_dict_returns_empty_dict(self):
+        samples = [["test"], "test", 6]
```
sergei-maertens, line 30: I don't think the list behaviour is correct here - list items can be strings and those should also be escaped. Additionally, a list item may itself be a nested dict with values that can be escaped, so please add those samples.

src/openforms/registrations/contrib/objects_api/utils.py
```diff
@@ -0,0 +1,19 @@
+from django.utils.html import escape
+
+from openforms.typing import JSONValue
+
+
+def escape_html_manually(input: dict) -> dict[str, JSONValue]:
```
sergei-maertens, line 6: I don't like that this function has side effects (mutates the dict) *and* also returns values, that makes things confusing. IMO this should be written differently - and it can be really concise:

```py
from typing import TypeVar
from django.utils.html import escape
from openforms.typing import JSONValue

T = TypeVar("T", bound=JSONValue)


def html_escape_json(value: T) -> T:
    """
    Recursively apply HTML escaping to string value nodes.
    """
    match value:
        case list():
            return [html_escape_json(item) for item in value]
        case dict():
            return {key: html_escape_json(value) for key, value in value.items()}
        case str():
            return escape(value)
        case _:
            # nothing to do, return unmodified
            return value
```

The generic type hint declares that whatever input type you give it, you will get back the same return type - and the types are limited by valid JSON-serializable types.

This approach is a pure function without side-effects/mutations :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3494, opened 2023-09-19
Move certificate and zgw_consumers.Service factory

sergei-maertens:
As discussed on Slack with Chris.


CharString (changes requested):
Looks like you missed a spot in the e2e tests. Other than that :+1:

PR #3504, opened 2023-09-24
Dynamic cookie consent bar

sergei-maertens:
Related to #3499

Testing out the django-cookie-consent#100 functionality, seems to be working well locally!

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -151,7 +151,7 @@ django-capture-tag==1.0
     # via -r requirements/base.in
 django-colorfield==0.9.0
     # via -r requirements/base.in
-django-cookie-consent==0.4.0
+django-cookie-consent==0.5.0b0
     # via -r requirements/base.in
 django-cors-headers==3.11.0
     # via -r requirements/base.in
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -243,7 +243,7 @@ django-colorfield==0.9.0
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
-django-cookie-consent==0.4.0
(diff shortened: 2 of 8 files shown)


src/openforms/analytics_tools/tests/test_analytics_rendering.py
```diff
@@ -1,13 +1,15 @@
...
 from cookie_consent.models import CookieGroup
 from django_webtest import WebTest
 
 from openforms.analytics_tools.models import AnalyticsToolsConfiguration
 from openforms.forms.tests.factories import FormFactory
-from openforms.utils.tests.cache import clear_caches
+from openforms.tests.utils import NOOP_CACHES
 
 
+@override_settings(CACHES=NOOP_CACHES)
```
vaszig, line 12: Do we use this for CI? I see that with or without it the tests are passing locally.

requirements/base.txt
```diff
@@ -149,7 +149,7 @@ django-capture-tag==1.0
     # via -r requirements/base.in
 django-colorfield==0.9.0
     # via -r requirements/base.in
-django-cookie-consent==0.4.0
+django-cookie-consent==0.5.0b0
```
vaszig, line 152: Is it a safe approach to use beta versions of a library? I saw that you know-have implemented this but just asking..

vaszig (approved):

src/openforms/analytics_tools/tests/test_analytics_rendering.py
sergei-maertens: without it, there's a risk that other tests are affected because the cache is shared between tests. So it's to prevent future issues and ensure test isolation.

PR #3507, opened 2023-09-26
API client refactors/clean ups

### LINKED ISSUE #3489: Refactor API client implementations to be based on new API client
Follow up of #3488

The new client should be the pattern of all clients, including those from `SOAPServer` and/or `StUFService`.

Additionally, check if we use requests-oauth2client (indirectly) and ensure that the correct `session` instance is provided.

**Checklist**

- [x] Create adapter for SOAPService
- [x] Create adapter for StUFService
- [x] Check requests-oauth2client usage
    - [x] `openforms.contrib.microsoft`
- [x] Check any other client definitions (zeep transport supports session kwarg)
    - [x] `openforms.contrib.bag` (move this into kadaster!)
    - [x] `openforms.contrib.brp`
    - [x] `openforms.contrib.camunda`
    - [x] `openforms.contrib.kadaster`
    - [x] `openforms.contrib.kvk`
    - [x] Haal Centraal / BRP (pre-request/token exchange thing is relevant here!)
    - [x] zgw
- [x] Figure out how to handle exception suppressing while allowing config checks to "crash"
- [ ] Update docs
    - [ ]  Developer docs for api clients
    - [ ] BAG configuration?
    - [ ] Kadaster configuration -> update model field labels
- [x] Check extensions
    - [x] Token exchange
    - [x] [Haal Centraal HR Prefill](https://github.com/open-formulieren/open-forms-ext-haalcentraal-hr)
- [x] Ensure zgw-consumers NLX integration still works (look into `Session.send`, `Session.prepare_request` and response hooks)
    - [x] updating base/root URL
    - [x] rewriting response data / params / headers

ZGW/ZDS-client is beast by itself and probably best left for a separate PR 😬

sergei-maertens:
This is wrapping up the non-zgw/zgw-consumers clients. A separate PR will follow for those.

Partially closes #3489


src/openforms/contrib/kadaster/api/views.py
```diff
@@ -13,21 +15,88 @@
...
+    )
+    def get(self, request, *args, **kwargs):
+        serializer = GetStreetNameAndCityViewInputSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        data = serializer.validated_data
+        postcode, number = data["postcode"], data["house_number"]
+
+        # check the cache so we avoid hitting the remote API too often (and risk
+        # of being throttled, see #1832)
+        address_data = caches["default"].get_or_set(
```
SilviaAmAm, line 82: Could use `from django.core.cache import cache` instead of `caches["default"]`?

src/openforms/pre_requests/clients.py
```diff
@@ -1,4 +1,6 @@
-from typing import TYPE_CHECKING, Any, Optional, TypedDict
+from __future__ import annotations
```
SilviaAmAm, line 1: What is this needed for?

SilviaAmAm (approved):

src/openforms/pre_requests/clients.py
sergei-maertens: it ensures that annotations are evaluated lazily! so you can do a conditional import but reference the type in annotations without having to quote it as a string

in this case it makes it possible to do `Submission | None` even with circular imports, instead of the more verbose `Optional["Submission"]`, since `"Submission" | None` is interpreted as an expression rather than a type, causing it to raise a `TypeError`

It's essentially saying "use the type annotation behaviour of future Python versions even though I'm still on an older version"

SilviaAmAm: Ohh!

PR #3509, opened 2023-09-27
Refactor zgw and objects-api clients

### LINKED ISSUE #3489: Refactor API client implementations to be based on new API client

sergei-maertens:
Partially closes #3489
Depends on #3508


src/openforms/submissions/tests/form_logic/test_fetching_form_variable_values_from_services.py
```diff
@@ -2,7 +2,7 @@
 from pathlib import Path
 from typing import Any
 from unittest import skip
-from urllib.parse import unquote_plus
+from urllib.parse import unquote
```
sergei-maertens, line 5: unquote_plus in the documentation has remarks about it being for form data, not intended for URL paths. The behaviour is also slightly changed with us using `furl` for the final URL building instead of `urljoin`, and furl already encodes a couple of things

src/openforms/variables/models.py
```diff
@@ -145,7 +145,7 @@ def request_arguments(self, context: DataMapping) -> dict:
         # this catches faults requests doesn't: https://github.com/psf/requests/issues/6359
         HeaderValidator()(headers)
 
-        escaped_for_path = {k: quote_plus(str(v)) for k, v in context.items()}
+        escaped_for_path = {k: quote(str(v), safe="") for k, v in context.items()}
```
sergei-maertens, line 148: removed the default `safe="/"` argument - in our case it is definitely unsafe. the plus does not need to be quoted, as that's only relevant in form data apparently.

vaszig (commented):
This seems to be nice and neat - I only have a couple of remarks-questions. Tested objects-API as well locally and it works as expected.

We could also remove some remaining `oas` code (like `zrc_service__oas`) just not to be confusing during maintenance, if you agree.

src/openforms/contrib/zgw/clients/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class PaginatedResponseData(TypedDict):
+    count: int
+    next: str
+    previous: str
+    results: list
+
+
+def pagination_helper(client: APIClient, paginated_data: PaginatedResponseData):
```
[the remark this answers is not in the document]
sergei-maertens: I considered it but left it to inference - we don't know for sure what the type is of the data inside the `results` list, so that would require generics in the type hints, and ideally even a generic for the type of `paginated_data`. I'll probably move this helper into the future PyPI package and then go all out with the typing.

sergei-maertens:
> We could also remove some remaining oas code (like zrc_service__oas) just not to be confusing during maintenance, if you agree.

it's actually required because of the zgw-consumers admin mixin which still uses its own client. I plan to publish our api client stuff on PyPI and then slowly refactor zgw-consumer itself to use it, but that's going to be for 2.5.0 at the earliest :/

PR #3512, opened 2023-09-28
[#3402] Add support for logging status changes in yubin emails

### LINKED ISSUE #3402: Improve visibility/tracing of email delivery
Related to #3326

Currently emails are delivered asynchronously - this applies to all transactional emails (registration backend, confirmation email...) using django-yubin. This causes the logs for a submission to report a success status because the email was queued successfully.

However, the actual sending/delivery can fail for a multitude of reasons (not limited to just attachment size) and we cannot correlate that back to the submission now.

Ideally, we can attach metadata to the email to be able to correlate it to the record/instance that caused it to be sent so that we can update the relevant logs/audit trails.

E.g. using custom headers for the e-mail:

* `X-OF-Content-Type: submissions.submission`
* `X-OF-Content-UUID: c2e1c102-34c5-4f67-b207-02484454b92b`
* `X-OF-event: registration-email`

We can then periodically (using Celery?) process the mail queue and create `TimelineLog` records for the statuses/events like succesful delivery or failed delivery. This then also allows us to create digest e-mails to send to the people that have to deal with these issues.

**Technical specification**

- [x] For outgoing submission registration e-mails, add custom headers:
    - [x] `X-OF-Content-Type: submission.submission`
    - [x] `X-OF-Content-UUID: <uuid of the submission>`
    - [x] `X-OF-Event: registration`
- [x] Ensure that a timeline log (audit) event is created whenever an e-mail sending attempt is performed. Ensure that the success status is logged of this. The log entry must be related to the submission (look this up from the e-mail headers). Some pointers:
    - https://github.com/APSL/django-yubin/blob/master/django_yubin/models.py#L136
    - https://github.com/APSL/django-yubin/blob/master/django_yubin/models.py#L99
    - https://github.com/APSL/django-yubin/blob/master/django_yubin/tasks.py
    - possibly we need to hook into this model's `post_save` to record *every* (re)-send attempt
- [x] Set up a daily digest e-mail
    - [x] Ensure you can configure (a list of?) e-mail addresses that should receive the daily digest e-mail in the global configuration
    - [x] Set up a Celery task to collect (relevant) timeline log records that require attention (in this case: failed registration e-mail deliveries)
        - Each record should clearly state what is wrong (e.g. "Email could not be sent") and to which submission it applies (use the public reference so people can search in the admin)
        - Entries should be de-duplicated - django-yubin will probably automatically try re-sending so you will likely have multiple identical errors about mail delivery not succeeding
        - The e-mail should be sent once every day to all configured recipients, configure this using celery beat. Doing this at 11pm seems appropriate, so that it's in people's TODO list at the start of the next work day. Only consider log entries from that day (e.g. do not include errors from yesterday/the day before yesterday)
        - Do not send any mail if there's nothing to report.

Please break up the tasks into separate, small, easy to review PRs!

SilviaAmAm:
Fixes #3402 (partly)

### DIFF
--- a/src/openforms/emails/signals.py
+++ b/src/openforms/emails/signals.py
@@ -1,0 +1,29 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from django_yubin.models import Message
+
+from openforms.logging import logevent
+from openforms.submissions.models import Submission
+
+from .constants import X_OF_CONTENT_UUID_HEADER, X_OF_EVENT_HEADER
+
+
+@receiver(post_save, sender=Message)
+def yubin_messages_status_change_handler(signal, sender, instance, created, **kwargs):
+    if created:
+        return
+
+    message = instance.get_email_message()
+    submission_uuid = message.extra_headers.get(X_OF_CONTENT_UUID_HEADER)
+    event = message.extra_headers.get(X_OF_EVENT_HEADER)
+    if not submission_uuid:
+        return
+
+    status_label = instance.get_status_display()
+    submission = Submission.objects.filter(uuid=submission_uuid).first()
+    if not submission:
+        # Maybe add logging in case a submission is deleted before the email is sent?
+        return
+
+    logevent.email_status_change(submission, event, status_label)
--- a/src/openforms/emails/signals.py
+++ b/src/openforms/emails/signals.py
@@ -12,7 +12,7 @@ from .constants import (
     X_OF_CONTENT_TYPE_HEADER,
     X_OF_CONTENT_UUID_HEADER,
     X_OF_EVENT_HEADER,
-    EmaiContentTypeChoices,
+    EmailContentTypeChoices,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


src/openforms/emails/signals.py
```diff
@@ -0,0 +1,29 @@
...
+from .constants import X_OF_CONTENT_UUID_HEADER, X_OF_EVENT_HEADER
+
+
+@receiver(post_save, sender=Message)
+def yubin_messages_status_change_handler(signal, sender, instance, created, **kwargs):
+    if created:
+        return
+
+    message = instance.get_email_message()
+    submission_uuid = message.extra_headers.get(X_OF_CONTENT_UUID_HEADER)
```
sergei-maertens, line 18: wouldn't you need to check the content type header + value before knowing it's a submission-related email?

sergei-maertens (commented):
neat

src/openforms/emails/signals.py
SilviaAmAm: Oh yes! Good point :smile:

src/openforms/emails/signals.py
```diff
@@ -0,0 +1,50 @@
...
+        message.extra_headers.get(X_OF_CONTENT_TYPE_HEADER, "")
+        == EmaiContentTypeChoices.submission
+    )
+    if not has_submission:
+        return
+
+    submission_uuid = message.extra_headers.get(X_OF_CONTENT_UUID_HEADER)
+    event = message.extra_headers.get(X_OF_EVENT_HEADER)
+    if not submission_uuid:
+        return
```
sergei-maertens, line 37: there is missing coverage here, but rather than making up a false example where you would only include part of the e-mail messages, I think an `assert submission_uuid` makes more sense and should crash/error the tests when someone forgets to add the actual submission UUID header to an email.

sergei-maertens (approved):

PR #3514, opened 2023-09-29
[#3477] Add dynamic form-action csp directive

### LINKED ISSUE #3477: CSP header "form-action" should be set and include obvious values
Currently, Internet.nl gives us a warning for the CSP-headers regarding the missing form-action directive.

We should set this to: `form-action: 'self' https://digid.nl https://*.digid.nl https://eh.signicat.nl ` etc... Ogone as well!

Or, we should bite the bullet and implement the Redirect binding.

vaszig:
Fixes #3477


src/openforms/authentication/contrib/digid/migrations/0002_auto_20231003_0917.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.20 on 2023-10-02 06:22
+
+from django.db import migrations
+
+from ...utils import update_csp
+
+
+def update_csp(apps, _):
+    """Update csp form-action value if a digid configuration already exists"""
+    DigidConfiguration = apps.get_model("digid_eherkenning", "DigidConfiguration")
```
vaszig, line 10: I don't get why this fails in local machine and not in the CI..Here is what I get as an error in my local machine:

`DigidConfiguration = apps.get_model("digid_eherkenning", "DigidConfiguration")
AttributeError: 'DigidConfiguration' object has no attribute 'get_model'`

SilviaAmAm: I think because you are importing a function with name `update_csp` and then this function also has name `update_csp`

vaszig: Yes you are right!Thanks, I am pushing an update now.

src/openforms/authentication/contrib/eherkenning/migrations/0002_auto_20231003_0917.py
```diff
@@ -0,0 +1,26 @@
...
+        if eherkenning_config.idp_service_entity_id:
+            update_csp(eherkenning_config, "eherkenning")
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("authentication_eherkenning", "0001_convert_eherkenning_setting_to_db"),
+    ]
+
+    operations = []
```
vaszig, line 26: Missing function call

src/openforms/authentication/contrib/digid/tests/test_csp_update.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def setUp(self):
+        super().setUp()
+
+        clear_caches()
+        self.addCleanup(clear_caches)
+
+    def test_csp_updates(self):
+        # assert no csp entry is created
+        self.assertTrue(CSPSetting.objects.none)
```
SilviaAmAm, line 56: Why when the config is created, no CSP directives should be added? :thinking:

SilviaAmAm (changes requested):

src/openforms/authentication/contrib/utils.py
```diff
@@ -0,0 +1,36 @@
...
+        "directive": CSPDirective.FORM_ACTION,
+        "value": f"{ADDITIONAL_VALUES[type]} {new_entity_id}",
+    }
+    CSPSetting.objects.create(**csps)
+
+
+def update_csp(
+    instance: DigidConfiguration | EherkenningConfiguration, type: str
+) -> None:
+    if not instance._state.adding:
```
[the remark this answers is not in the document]
vaszig: `pre_save` signal does not have `create` as an argument so this is the safest way I found to determine if it's a new instance or not.

src/openforms/authentication/contrib/digid/migrations/0002_auto_20231003_0917.py
```diff
@@ -0,0 +1,24 @@
...
+
+    if DigidConfiguration.objects.exists():
+        digid_config = DigidConfiguration.objects.get()
+
+        if digid_config.idp_service_entity_id:
+            update_csp(digid_config, "digid")
```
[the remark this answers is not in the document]
sergei-maertens: `DigidConfiguration.objects.first()` is a nice shortcut around that - if nothing exists, it returns `None`

src/openforms/authentication/contrib/utils.py
```diff
@@ -0,0 +1,36 @@
+from digid_eherkenning.models.digid import DigidConfiguration
+from digid_eherkenning.models.eherkenning import EherkenningConfiguration
+
+from openforms.config.constants import CSPDirective
+from openforms.config.models import CSPSetting
+
+ADDITIONAL_VALUES = {
+    "digid": "'self' https://digid.nl https://*.digid.nl",
+    "eherkenning": "'self'",
+}
```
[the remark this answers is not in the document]
sergei-maertens: the JSON files also requires some string-replacements and in the end they don't contain much information. I think I prefer doing it in Python.

Also - I think we should configure the django-csp settings to emit the `self`, and then each additional aspect adds to it. e.g. digid adds https://*.digid.nl and https://digid.nl, so you end up with a total combination of `'self' https://digid.nl https://*.digid.nl`.

src/openforms/authentication/contrib/utils.py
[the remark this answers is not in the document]
sergei-maertens: you can check `instance.pk` for being `None` - if it's not none, it's an update

`instance._state` is private, undocumented Django API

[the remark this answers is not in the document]
vaszig: Discussed this with @sergei-maertens as well on slack. 

We are going to use the `post_init` signal in order to update the instance with the old endpoint and then in `post_save` signal we can see if CSPSettings need an update or not based on the two endpoints (the new and the old one).

src/openforms/authentication/contrib/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def update_csp(
+    instance: DigidConfiguration | EherkenningConfiguration, type: str
+) -> None:
+    if not instance._state.adding:
+        original_object = instance.__class__.objects.get(pk=instance.pk)
+
+        if original_object.idp_service_entity_id != instance.idp_service_entity_id:
+            CSPSetting.objects.filter(
```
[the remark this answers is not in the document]
vaszig: Discussed with @sergei-maertens as well on slack. We are going to be consistent with the analytics csp update flow, delete the old setting and add the new one. In that way we avoid possible conflicts as well.

src/openforms/authentication/contrib/eherkenning/tests/test_csp_update.py
```diff
@@ -0,0 +1,104 @@
...
+            kwargs={"slug": form.slug, "plugin_id": "eherkenning"},
+        )
+        form_path = reverse("core:form-detail", kwargs={"slug": form.slug})
+        form_url = f"http://testserver{form_path}"
+
+        # redirect_to_eherkenning_login
+        response = self.client.get(login_url, {"next": form_url}, follow=True)
+
+        self.assertIn(
+            "form-action 'self' urn:etoegang:DV:00000001111111111000:entities:9000;",
```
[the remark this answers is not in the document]
vaszig: The problem here is that we do this check here [https://github.com/open-formulieren/open-forms/blob/464e155866cc816f11be1fbe13da6344d5da7e5e/docs/installation/issues/form_auth.rst#L52]() and I am not able to change this to a suitable url. 

`urn:etoegang:DV:00000001111111111000:entities:9000` is what we have in the `eherkenning-metadata.xml` file [https://github.com/open-formulieren/open-forms/blob/master/src/openforms/authentication/contrib/eherkenning/tests/data/eherkenning-metadata.xml#L1](url) , so this is what is checked. 

Should we leave this until we tackle the digid_eherkenning repo issue [https://github.com/maykinmedia/django-digid-eherkenning/issues/44](url) ?What is your opinion?

vaszig:
FYI: this PR has been refactored according to the new version of `django-digid-eherkenning` library and after discussion with Sergei we decided that it's better to connect the CSPSetting model with each instance that creates a csp setting (same way we do it with `TimelineLog` ).

src/openforms/authentication/contrib/digid/signals.py
sergei-maertens, line 1: I think I'd move all the digid/eherkenning related code into `openforms.contrib.digid_eherkenning`, so that includes the stuff that's now smeared out over `openforms.authentication.contrib.digid`, `openforms.authentication.contrib.eherkenning` and `openforms.config`.

I also think that there could be a single signal handler to register:

```py
@receiver(post_save, sender=DigidConfiguration)
@receiver(post_save, sender=EHerkenningConfigurationConfiguration)
def trigger_csp_update(sender, instance: DigidConfiguration | EHerkenningConfiguration, **kwargs):
    ... # update CSP
```

if the double stacking of decorators doesn't work, you can manually connect them:

```py
post_save.connect(...)
```

src/openforms/authentication/contrib/digid/tests/test_csp_update.py
```diff
@@ -0,0 +1,125 @@
...
+    @patch(
+        "onelogin.saml2.idp_metadata_parser.OneLogin_Saml2_IdPMetadataParser.get_metadata"
+    )
+    def test_csp_updates(self, get_metadata):
+        # assert no csp entries exist with initial solo model
+        self.assertTrue(CSPSetting.objects.none)
+
+        # assert new csp entry is added after adding metadata url
+        with METADATA_POST.open("rb") as md_file:
+            metadata_content = md_file.read().decode("utf-8")
```
sergei-maertens, line 60: pathlib has some nicer methods for this:

```py
metadata_content = METADATA_POST.read_text()
```
if I remember correctly. It uses a context manager internally, so less indentation to handle in your own code :)

src/openforms/authentication/contrib/digid/tests/test_csp_update.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertFalse(CSPSetting.objects.filter(id=csp_added.id))
+        self.assertEqual(csp_updated.content_object, self.config)
+        self.assertEqual(csp_updated.directive, CSPDirective.FORM_ACTION)
+        self.assertEqual(
+            csp_updated.value,
+            "https://digid.nl "
+            "https://*.digid.nl "
+            "https://test-digid.nl/saml/idp/request_authentication "
+            "https://test-digid.nl/saml/idp/request_logout",
+        )
```
sergei-maertens, line 92: perhaps a `get_metadata.assert_called_once_with(...)` is also appropriate?

src/openforms/authentication/contrib/digid/tests/test_csp_update.py
```diff
@@ -0,0 +1,125 @@
...
+            self.config.save()
+
+        login_url = reverse(
+            "authentication:start", kwargs={"slug": form.slug, "plugin_id": "digid"}
+        )
+        form_path = reverse("core:form-detail", kwargs={"slug": form.slug})
+        form_url = f"http://testserver{form_path}?_start=1"
+
+        # redirect_to_digid_login
+        response = self.client.get(f"{login_url}?next={form_url}", follow=True)
```
sergei-maertens, line 115: please don't build URLs with querystrings by hand, the django test client serializers these things already:

```py
self.client.get(login_url, {"next": form_url}, follow=True)
```

src/openforms/config/constants.py
```diff
@@ -1,6 +1,11 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
+ADDITIONAL_CSP_VALUES = {
+    "digid": "https://digid.nl https://*.digid.nl",
+    "eherkenning": "",
+}
```
sergei-maertens, line 7: please move this to `openforms.contrib.digid_eherkenning`

In general, the `config` app is set up in a generic way and consults the various plugin registries, so hardcoding plugin-specific things outside of the plugin are a code smell - it introduces (more) tight coupling than needed.

src/openforms/config/models.py
```diff
@@ -658,9 +660,23 @@ class CSPSetting(models.Model):
...
+        ContentType,
+        verbose_name=_("content type"),
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+    )
+    object_id = models.TextField(
+        verbose_name=_("object id"),
+        blank=True,
+        null=True,
```
sergei-maertens, line 676: can we drop the `null=True` here?

src/openforms/config/utils.py
```diff
@@ -24,3 +35,38 @@ def verify_clamav_connection(host: str, port: int, timeout: int) -> "ClamAVStatu
...
+    from .models import CSPSetting
+
+    CSPSetting.objects.filter(
+        content_type=get_content_type_for_model(obj), object_id=str(obj.id)
+    ).delete()
+
+    CSPSetting.objects.create(content_object=obj, directive=directive, value=value)
+
+
+def create_digid_eherkenning_csp_settings(
```
sergei-maertens, line 50: please move these specific helpers into their respective apps.

Additionally, since `CSPSetting` is our own model, I think that `_create_csp_settings` should be a model manager method:

```py
CSPSetting.objects.set_for(obj, [("form-action", "https://*.digid.nl ..."])
```

the current implementation only ever allows a single `CSPSetting` for a given configuration object since you allow only a single `directive` to be passed. Calling the helper again for another directive will delete the first once through the GFK.

src/openforms/config/utils.py
```diff
@@ -24,3 +35,38 @@ def verify_clamav_connection(host: str, port: int, timeout: int) -> "ClamAVStatu
...
+    config: Union["DigidConfiguration", "EherkenningConfiguration"], config_type: str
```
sergei-maertens, line 51:
```suggestion
    config: DigidConfiguration | EherkenningConfiguration, config_type: str
```
is nicer syntax :)

src/openforms/config/utils.py
```diff
@@ -24,3 +35,38 @@ def verify_clamav_connection(host: str, port: int, timeout: int) -> "ClamAVStatu
...
+        urls = (
+            f"{ADDITIONAL_CSP_VALUES[config_type]} {urls['sso_url']} {urls['slo_url']}"
+        )
+    else:
+        urls = f"{urls['sso_url']} {urls['slo_url']}"
+
+    _create_csp_settings(config, CSPDirective.FORM_ACTION, urls)
+
+
+def create_ogone_csp_settings(
```
sergei-maertens, line 69: this should be moved into `openforms.payments.contrib.ogone` app

sergei-maertens (changes requested):

src/openforms/config/models.py
```diff
@@ -649,6 +656,33 @@ def as_dict(self):
         return {k: list(v) for k, v in ret.items()}
 
 
+class CSPSettingManager(models.Manager.from_queryset(CSPSettingQuerySet)):
+    def set_for(self, obj: models.Model, settings: list[Tuple[str, str]]) -> None:
```
Viicos, line 660:
```suggestion
    def set_for(self, obj: models.Model, settings: list[tuple[str, str]]) -> None:
```

No need to import `from typing import Tuple` on Python 3.10

src/openforms/config/models.py
```diff
@@ -649,6 +655,33 @@ def as_dict(self):
...
+                logger.error(
+                    "Could not create csp setting for model '%s'. '%s' is not a valid directive.",
+                    obj,
+                    directive,
+                )
+                return
+
+            instances.append(
+                CSPSetting(content_object=obj, directive=directive, value=value)
+            )
```
sergei-maertens, line 676:
```suggestion
class CSPSettingManager(models.Manager.from_queryset(CSPSettingQuerySet)):
    def set_for(self, obj: models.Model, settings: list[tuple[CSPDirective, str]]) -> None:
        """
        Deletes all the connected csp settings and creates new ones based on the new provided data.
        """
        instances = [
            CSPSetting(content_object=obj, directive=directive, value=value)
            for directive, value in settings
        ]
```
A couple things here to remark:

1. If you update the type hint from str -> CSPDirective, you force callers to use the `CSPDirective.foo` constants (at a type checking level). Relying on the type checker then allows you to drop the defensive programming clause to reject "invalid" values.
2. you can immediately destructure the `setting` tuple in the for loop
3. dropping the entire guard clause turns this into a trivial list comprehension.

src/openforms/config/models.py
```diff
@@ -649,6 +655,33 @@ def as_dict(self):
         return {k: list(v) for k, v in ret.items()}
 
 
+class CSPSettingManager(models.Manager.from_queryset(CSPSettingQuerySet)):
+    def set_for(self, obj: models.Model, settings: list[tuple[str, str]]) -> None:
```
sergei-maertens, line 659: I think a `@transaction.atomic` decorator would be good to guarantee that we don't end up without csp directives if the insert crashes for some reason.

src/openforms/contrib/digid_eherkenning/apps.py
```diff
@@ -0,0 +1,12 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class DigidEherkenningApp(AppConfig):
+    name = "openforms.contrib.digid_eherkenning"
+    label = "contrib_digid_eherkenning"
+    verbose_name = _("DigiD Eherkenning Contrib")
```
sergei-maertens, line 8:
```suggestion
    verbose_name = _("DigiD/Eherkenning utilities")
```

src/openforms/contrib/digid_eherkenning/constants.py
```diff
@@ -0,0 +1,12 @@
+from digid_eherkenning.models.digid import DigidConfiguration
+from digid_eherkenning.models.eherkenning import EherkenningConfiguration
```
sergei-maertens, line 2:
```suggestion
from digid_eherkenning.models import DigidConfiguration, EherkenningConfiguration
```

src/openforms/contrib/digid_eherkenning/signals.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from digid_eherkenning.models.digid import DigidConfiguration
+from digid_eherkenning.models.eherkenning import EherkenningConfiguration
```
sergei-maertens, line 5: you can import both models from `models` module (this is a pattern that you should always follow)

src/openforms/contrib/digid_eherkenning/signals.py
```diff
@@ -0,0 +1,19 @@
...
+
+from openforms.contrib.digid_eherkenning.utils import (
```
sergei-maertens, line 7:
```suggestion
from .utils import (
```
within a django app, please use relative imports. Use absolute imports if you're importing stuff from another app (this helps reviewing if you're importing stuff you shouldn't be importing ;) )

src/openforms/contrib/digid_eherkenning/signals.py
```diff
@@ -0,0 +1,19 @@
...
+    create_digid_eherkenning_csp_settings,
+)
+
+from .constants import CONFIG_TYPES
```
sergei-maertens, line 11: why go through `CONFIG_TYPES`, can't you map the extra directives directly to the model?

in `create_digid_eherkenning_csp_settings`
```py
from .constants import ADDITIONAL_CSP_VALUES

additional_values = ADDITIONAL_CSP_VALUES.get(type(config))
...
```
and then you can drop these "magic strings" & `CONFIG_TYPES` mapping.

src/openforms/contrib/digid_eherkenning/utils.py
```diff
@@ -2,7 +2,14 @@
 
 from django.templatetags.static import static
 
+from digid_eherkenning.models.digid import DigidConfiguration
+from digid_eherkenning.models.eherkenning import EherkenningConfiguration
```
sergei-maertens, line 6: see above for the imports

src/openforms/contrib/digid_eherkenning/utils.py
```diff
@@ -19,3 +26,22 @@ def get_eherkenning_logo(request) -> Dict[str, str]:
...
+
+    # create the new directives based on the POST bindings of the metadata XML and
+    # the additional constants
+    urls, _ = config.process_metadata_from_xml_source()
+    if ADDITIONAL_CSP_VALUES[config_type]:
+        urls = (
+            f"{ADDITIONAL_CSP_VALUES[config_type]} {urls['sso_url']} {urls['slo_url']}"
+        )
+    else:
+        urls = f"{urls['sso_url']} {urls['slo_url']}"
```
sergei-maertens, line 45:
```suggestion
    csp_values = [urls["sso_url"], urls["slo_url"]]
    if (additional_csp_values := ADDITIONAL_CSP_VALUES.get(type(config), [])):
        csp_values += additional_csp_values
   form_action_urls = " ".join(csp_values)
   ...
```
you can reduce the branching and avoid the duplication with the f-string arguments - just build a list of relevant URLs and at the end join it with spaces & then pass that to the `set_for` method.

Additionally, you start out with the `urls` variable from the digid-herkenning library which is a dictionary `dict[str, str]`, but then this variable becomes a `str`. While this is valid python, my type checker will definitely complain about it :) It's better to use a different variable and stick to "static" types at runtime.

sergei-maertens (approved):

PR #3515, opened 2023-09-29
[#3402] Email digest

### LINKED ISSUE #3402: Improve visibility/tracing of email delivery

SilviaAmAm:
Fixes #3402


src/openforms/config/models.py
```diff
@@ -587,6 +587,16 @@ class GlobalConfiguration(SingletonModel):
         validators=[MaxValueValidator(60)],
     )
 
+    recipients_email_digest = ArrayField(
+        models.EmailField(),
+        verbose_name=_("recipients email digest"),
+        help_text=_(
+            "The email addresses that should receive a daily report of any email sending failures."
```
sergei-maertens, line 594:
```suggestion
            "The email addresses that should receive a daily report of items requiring attention."
```
while we're starting out with email sending failures, I reckon this mechanism can be used much broader and be used to inform an admin on any issues that need investigating.

src/openforms/emails/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+    if not (recipients := config.recipients_email_digest):
+        return
+
+    period_start = timezone.now() - timedelta(days=1)
+
+    logs = TimelineLogProxy.objects.filter(
+        template="logging/events/email_status_change.txt",
+        timestamp__gt=period_start,
+        extra_data__status=Message.STATUS_FAILED,
+    ).distinct("content_type", "extra_data__status", "extra_data__event")
```
sergei-maertens, line 29: maybe add a flag or tag to log records "includeInDailyDigest: true" and query based on that rather than hardcoding a template?

src/openforms/emails/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if not logs.count():
+        return
```
sergei-maertens, line 32:
```suggestion
    if not logs:
        return
```
if there are logs, it's being evaluated anyway in the template context, so you can save a query here

src/openforms/emails/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+
+    content = render_to_string(
+        "emails/digest.html",
```
sergei-maertens, line 35:
```suggestion
        "emails/admin_digest.html",
```
maybe make the intended audience a bit more clear?

src/openforms/emails/tests/test_tasks.py
```diff
@@ -0,0 +1,154 @@
...
+                extra_data={"status": Message.STATUS_FAILED, "event": "registration"},
+            )
+
+        with (
+            freeze_time("2023-01-03T01:00:00+01:00"),
+            patch(
+                "openforms.emails.tasks.GlobalConfiguration.get_solo",
+                return_value=GlobalConfiguration(recipients_email_digest=[]),
+            ),
+            patch("openforms.emails.tasks.send_mail_html") as patch_email,
```
sergei-maertens, line 150: I think we can avoid that and instead use `assert len(mail.outbox) == 0`?

sergei-maertens (approved):

PR #3520, opened 2023-10-03
Bump dependencies

sergei-maertens:
Following @dependabot alerts.

None of the vulnerabilities seem actually exploitable, so low severity assumed.


docker/ci/requirements.txt
```diff
@@ -16,5 +16,5 @@ markupsafe==2.1.1
     # via jinja2
 requests==2.31.0
     # via -r requirements.in
-urllib3==1.26.14
+urllib3==2.0.6
```
sergei-maertens, line 19: checked the changelog, requests 2.31 is compatible

PR #3532, opened 2023-10-10
Move the definition of time component validators

### LINKED ISSUE #3531: Update minTime/maxTime validators
The time component now puts `minTime` and `maxTime` in the component root, while it should be put inside the `validate` nested object (for consistency sake).

The types repository uses some generic patterns that enforce this. This requires:

- [x] Updating the form builder component
- [x] Updating the validators in the SDK
- [x] Adding a data migration for existing components
- [x] Adding a conversion shim for importing old forms

sergei-maertens:
Partly closes #3531 (SDK update needed)

- Added custom `ConvertComponentsOperation` django migration type
- Added time component migration
- Ensure that imports use conversion registry


src/openforms/formio/migration_converters.py
```diff
@@ -0,0 +1,51 @@
...
+This registry is used by the data migrations *and* form import. It guarantees that
+component definitions are rewritten to be compatible with the current code.
+"""
+from typing import Protocol
+
+from .typing import Component
+
+
+class Converter(Protocol):
+    def __call__(self, component: Component) -> bool:  # pragma: no cover
```
CharString, line 13: Why not
```suggestion
    def __call__(self, component: Component) -> Component:  # pragma: no cover
```
That will give it nice properties like composability.

And in place mutation closes the door to parallelization.

sergei-maertens: there is a has changed/has not changed flag required for the data migration so it doesn't perform an update on `FormDefinition` records that weren't actually updated.

> And in place mutation closes the door to parallelization.

does it really? in a data migration you can still parallellize the queryset itself and run the converter in parallel for every form definition in the queryset, at import time you can still do something similar for every FD in the import data.

If  you have a 1000 components in a single form definition and that would *require* parallel conversion, then I think you're building shitty user hostile forms that are way too long too.

CharString: > And in place mutation closes the door to parallelization.

I don't know when the GILectomy will land, but I heard they're aiming for the year of the Linux desktop.

CharString (changes requested):
Still not convinced of the API, but we've talked about that.
The naming confusion needs to be tackled.
But it seems to work.

src/openforms/forms/migration_operations.py
```diff
@@ -0,0 +1,57 @@
+from django.core.exceptions import ImproperlyConfigured
+from django.db import migrations
+
+from openforms.formio.migration_converters import CONVERTERS
+from openforms.formio.utils import iter_components
+
+
+class Converter:
```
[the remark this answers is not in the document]
sergei-maertens: yeah that's a bit the drawback of protocols, namespace pollution :(

I'll rename a bunch of things, one is actually an entirely form converter, the other is a component converter

src/openforms/formio/migration_converters.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Mutate a component in place.
+
+        The component is guaranteed to be of the expected type.
```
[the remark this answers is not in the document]
sergei-maertens: > I guess since all json that comes from the FormIO always has all keys. It's the same.

that's actually not true, especially not if people add a component in the builder and then edit the JSON. You can specify only the non-default keys and have it work, which is why tests are written with minimal component definitions too :)

CharString (approved):

PR #3538, opened 2023-10-16
Backport yubin patches

### LINKED ISSUE #3448: E-mail: "BadHeaderError: Header values can't contain newlines "

2.3.0

Long subject lines (70+ chars) get `\n` added them and django-yubin will not allow the e-mail to be sent. 
sentry: 336550 
This results in a succesful submission: Registratieplugin "E-mail registratie" (email) meldt: Registratie geslaagd.

sergei-maertens:
Closes #3448

This includes the relevant git commits as `.patch` files and applies them in the backed image build to the installed django-yubin package.

Backporting needed to Open Forms 2.3.x, older is not affected due to yubin v1 being used there.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,9 +30,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         shared-mime-info \
     && rm -rf /var/lib/apt/lists/*
 
-        # build-essential python3-dev python3-pip python3-setuptools python3-wheel python3-cffi
-
-
 WORKDIR /app
 
 # Ensure we use the latest version of pip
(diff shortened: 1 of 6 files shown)


CharString (approved):
I guess it works, but isn't good 'ol OG patch a much smaller, cheaper dependency?

You suggested me to save just shy of 8 bytes, recently... This is 35M vs 254k :upside_down_face:

sergei-maertens:
> I guess it works, but isn't good 'ol OG patch a much smaller, cheaper dependency?

Might be, but when I googled "how to apply patch" the top hits were examples of `git apply` and I'm more familiar with git than deliberately searching for `patch` itself.

> You suggested me to save just shy of 8 bytes, recently... This is 35M vs 254k 🙃

I don't think I mentioned size anywhere? I just struck me as odd that a different configuration was picked/set from what is in `src/openforms/conf/base.py`:

```py
DATABASES = {
    "default": {
        "ENGINE": config("DB_ENGINE", "django.db.backends.postgresql"),
        "NAME": config("DB_NAME", "openforms"),
        "USER": config("DB_USER", "openforms"),
        "PASSWORD": config("DB_PASSWORD", "openforms"),
        "HOST": config("DB_HOST", "localhost"),
        "PORT": config("DB_PORT", 5432),
    }
}

# keep the current schema for now and deal with migrating to BigAutoField later, see
# https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-DEFAULT_AUTO_FIELD
DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
and I couldn't see a reason why considering it's a singleton model.

In terms of the size of the dependency here - `git` does not end up in the final distributed image, it's only added as a build dependency in the build stage, like a couple of other tools that are needed to be able to compile the pip dependencies.

sergei-maertens:
Backport commit: d43858f190a5a210284a0b4cdbb68d80c6921c64

Only 2.3.x is relevant

PR #3539, opened 2023-10-16
⬆️ Bump pip-tools for latest pip compatibility

Viicos:


sergei-maertens (approved):

Viicos:
Regarding last commit:
> DEPRECATION: o365 2.0.16 has a non-standard dependency specifier tzlocal<3.*,>=1.5.0. pip 24.0 will enforce this behaviour change.

Issue #3491, opened 2023-09-19
Slowdown due to reusable form defintions lookup when loading/starting a form.


LaurensBurger:
### Product versie / Product version

2.3.0

### Omschrijf het probleem / Describe the bug

Internal: LV 107

When a user creates a new form, or opens a existing form: Open forms always looks up all "reusable" form definitions. 

The calls to this endpoint: /api/v2/form-definitions?is_reusable=true&page=1&used_in= are paginated with 24 items per page. 

The paginated results contain all the data of the form definition, including translation. 

When the total amount of reusable form definition grows, the loading time automatically grows along with it. 
Example: 
environment with 416 defintions:
![image](https://github.com/open-formulieren/open-forms/assets/112621537/60131def-c48e-4668-bec3-31010e1eaaee)
Each request takes approx: 600ms x 17 = 10 seconds of loading only form definitions each time a form is opened or a new form is added.

This seems rather long and can only increase with the amount of forms but also extra data (translations) being included in these requests.

joeribekker:
Refinement: We can create a separate endpoint to only return the name/uuid to prevent the heaps of data returned for the pulldown. We do need to refactor the client to retrieve the form definition detail when you actually want to use it. Pagination can be dropped for this endpoint. Sergei is not convinced so i'll let this sink in a  bit.

Issue #3283, opened 2023-07-21
Deprecations to remove in 3.0


sergei-maertens:
A list of breaking changes to make when we get to version 3.0

<kbd>CTRL</kbd> + <kbd>F</kbd> on `DeprecationWarning` in the backend to find more.

- ~~INVESTIGATE: can we keep existing endpoints on /api/v2/ and new ones to /api/v3/ so we don't have to move everything at once?~~ Resolving by marking most of the API private API

- Appointments #4872 
    - [x] Remove old-style (form.io based) appointment flow
        - [x] most of `appointments.utils`
        - [x] extracting the appointment data from a submission
        - [x] Frontend code to manage appointment (tab)
        - [x] Remove support for fixed customer fields
    - [x] Remove "invalid input serializer -> return empty result list" in favour of raising a validation error exception
- Formio
    - [x] Remove `fd_translations_converter` module (this will affect exports < 3.0 that have legacy translations) #4828
    - [x] #3718
    - [x] Drop password component #4842
- Form design
    - [x] Remove 'formStep' in location action in favour of `formStepUuid` #4892 
    - [x] Make `FormDefinitionSerializer.name` `readOnly` (name is set via translations) @stevenbal 
    - [x] Delete `_data` model field https://github.com/open-formulieren/open-forms/blob/8f7b003e285893a158036dfc54ffd58f478bd12d/src/openforms/submissions/models/submission_step.py#L102
- Styling/NL DS
    -  [x] Remove compatibility shims from `of-page-header-*` and `of-page-footer-*` in favour of `utrecht-page-header-*` and `utrecht-page-footer-*` tokens @robinmolen 
        - `components/_page-header.scss`
        - `components/_page-footer.scss`
- Registration:
   - [x] Objects API/ZGW groups shouldn't have fields nullable
   - [x] Remove objecttype URL -> UUID import converter hook @robinmolen 
   - ~~Remove document type conversion hook (if it exists)~~
   - ~~Remove documenttype URLs - auto migrate where possible (management command to be added in 2.8.x?)~~
   - [x] Document migration procedures
   - [x] Remove automatic StUF-ZDS Utrecht extension conversion @robinmolen
   - [x] Remove (export/import) support/converter for single-registration backend configuration @robinmolen 

- Authentication
    - [x] Remove legacy machtigen context @sergei-maertens 
    - ~~#4890~~

- Misc
[...]

PR #3542, opened 2023-10-18
⚡️ Disable pagination when requesting form definitions in the admin

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/3491

I'm a bit frustrated by this issue, as what I wanted to do initially was:
- Be able to specify specific fields to include via query parameters, e.g. `path?include=url,name,internal_name` (as only these three fields are required by the `NewStepFormDefinitionPicker` component)
- Disable pagination in some way

First one would be doable with https://github.com/wimglenn/djangorestframework-queryfields but it would add an extra dependency not updated in 7 years.

Or another solution suggested by @joeribekker:
- Define a custom endpoint (e.g. via `@action`) to only fetch the required fields, but it would not inherit from the currently defined `filterset_class`.

I ended up using a custom pagination class, that would understand `page_size=0` to mean _no pagination_. We will have to see if this is viable with a lot of form definitions in the db


src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -187,7 +187,7 @@ function reducer(draft, action) {
      * Form-level actions
      */
     case 'BACKEND_DATA_LOADED': {
-      const {formData, supportingData} = action.payload;
```
Viicos, line 190: To be in line with:

https://github.com/open-formulieren/open-forms/blob/cc8fe709fcb40280c8d32bdab45f55b57304a7a0/src/openforms/js/components/admin/form_design/form-creation-form.js#L190

src/openforms/forms/tests/test_api_formdefinition.py
```diff
@@ -39,6 +39,29 @@ def test_list(self):
...
+        self.client.force_authenticate(user=user)
+
+        FormDefinitionFactory.create_batch(100)
+
+        url = reverse("api:formdefinition-list")
+
+        with self.subTest("no pagination"):
+            response = self.client.get(url, {"page_size": "0"}, format="json")
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            self.assertEqual(len(response.json()["results"]), 100)
```
Viicos, line 53: It appears the JSON response data is still formatted as a paginated response. Could it return the list of form definitions directly instead?

Viicos:
I think we can get rid of this?
https://github.com/open-formulieren/open-forms/blob/cc8fe709fcb40280c8d32bdab45f55b57304a7a0/src/openforms/forms/api/filters.py#L34-L42

sergei-maertens:
> I think we can get rid of this?
> 
> https://github.com/open-formulieren/open-forms/blob/cc8fe709fcb40280c8d32bdab45f55b57304a7a0/src/openforms/forms/api/filters.py#L34-L42

we can deprecate it, but removing it is a breaking change and requires a new major version of the API

Viicos:
Couldn't get e2e tests to work. My best guess (and I was able to reproduce only once locally but unfortunately the timeout closed the test browser) is that this happens when form definitions are loaded, and the e2e test clicks on the select element on a precise timing where the select options HTML elements don't have time to load, which kind of breaks React's reactivity.

src/openforms/forms/api/viewsets.py
```diff
@@ -183,7 +183,7 @@ def get_serializer_class(self):
 
     @extend_schema(
         summary=_("Retrieve form definition JSON schema"),
-        tags=["forms"],
+        tags=["forms", "form-definitions"],
```
sergei-maertens, line 186: doesn't this cause the same endpoint to be listed in multiple places? we don't want that

src/openforms/js/components/admin/form_design/form-creation-form.js
sergei-maertens: I don't understand what this fixes

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -980,6 +997,16 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl}) => {
     });
   }, []);
 
+  useAsync(async () => {
+    const reusableFormDefinitions = await loadFromBackend([
+      {endpoint: FORM_DEFINITIONS_ENDPOINT, query: {is_reusable: true}},
+    ]);
+    dispatch({
+      type: 'REUSABLE_FORM_DEFINITIONS_LOADED',
+      payload: reusableFormDefinitions[0],
```
sergei-maertens, line 1006: what is this magic `[0]` index doing?

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -246,6 +247,16 @@ function reducer(draft, action) {
       );
       break;
     }
+    case 'REUSABLE_FORM_DEFINITIONS_LOADED': {
+      const reusableFormDefinitions = action.payload;
+      const formDefinitionsIds = draft.formDefinitions.map(fd => fd.uuid);
+      draft.formDefinitions = [
+        ...draft.formDefinitions,
+        ...reusableFormDefinitions.filter(fd => !formDefinitionsIds.includes(fd.uuid)),
+      ];
```
sergei-maertens, line 256: I am wondering if there's a possible state update race condition? if both requests (used in form, reusable FDs) complete at the same time because they have parallel requests, we have two actions writing to the same state key, maybe one overwrites the other?

src/openforms/forms/api/viewsets.py
Viicos: This is already the case for a bunch of other endpoints, actually all the `FormDefinitionViewSet` endpoints

src/openforms/js/components/admin/form_design/form-creation-form.js
Viicos: This is to use the same order as defined [here](https://github.com/open-formulieren/open-forms/blob/cc8fe709fcb40280c8d32bdab45f55b57304a7a0/src/openforms/js/components/admin/form_design/form-creation-form.js#L979), less confusing imo

src/openforms/js/components/admin/form_design/form-creation-form.js
Viicos: `loadFromBackend` returns an array containing the results of each request. In our case, we are only making one request

src/openforms/js/components/admin/form_design/form-creation-form.js
Viicos: That might be the reason

src/openforms/js/components/admin/form_design/form-creation-form.js
sergei-maertens: right - my first hunch was to just use `get` directly, but that doesn't handle pagination - in this case I'd destructure the return value to increase readability:

```js
const response = await loadFromBackend(...);
const [reusableFormDefinitions] = responses;
```

src/openforms/js/components/admin/form_design/form-creation-form.js
sergei-maertens: I'd like to take some time today to dive into this myself and try to figure out what's going on

Viicos: Well turns out it might be a race condition issue:

`test_all_reuseable_form_definition_options_are_available_when_not_part_of_the_form`:
https://github.com/open-formulieren/open-forms/blob/632c23d78bec16f16ce4542b0bc276f6ae4b0920/src/openforms/js/components/admin/form_design/NewStepFormDefinitionPicker.js#L23
https://github.com/open-formulieren/open-forms/blob/632c23d78bec16f16ce4542b0bc276f6ae4b0920/src/openforms/forms/tests/e2e_tests/test_form_designer.py#L1528


test output:

`[]`

Locally:

`[Object, Object]`

Viicos: Fixed by last two commits

src/openforms/js/components/admin/form_design/form-creation-form.js
sergei-maertens: okay - next time please try to avoid refactoring code that's not directly related to the issue as it clutters the diff, instead make a separate PR to clean that up. Small cleanup PRs like that can be made without issue :)

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -5,7 +5,7 @@ import set from 'lodash/set';
 import sortBy from 'lodash/sortBy';
 import zip from 'lodash/zip';
 import PropTypes from 'prop-types';
-import React, {useContext} from 'react';
+import React, {useContext, useEffect} from 'react';
```
sergei-maertens, line 8: unused `useEffect` import

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -980,6 +997,20 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl}) => {
...
+    // Waiting for the last dispatch to be done to avoid state race conditions.
+    if (!loading) {
+      const responses = await loadFromBackend([
+        {endpoint: FORM_DEFINITIONS_ENDPOINT, query: {is_reusable: true}},
+      ]);
+      const [reusableFormDefinitions] = responses;
+      dispatch({
+        type: 'REUSABLE_FORM_DEFINITIONS_LOADED',
+        payload: reusableFormDefinitions[0],
+      });
```
sergei-maertens, line 1010: this is still using the `[0]` index? I thought the destructuring would've taken care of that :thinking:

sergei-maertens (commented):
Can you also deprecate that `filter_queryset` thing doing the `OR` filter? A `warnings.warn(..., DeprecationWarning)` is fine & add it to the list of things to remove in #3283

src/openforms/js/components/admin/form_design/form-creation-form.js
Viicos: woops

Viicos:
> Can you also deprecate that `filter_queryset` thing doing the `OR` filter? A `warnings.warn(..., DeprecationWarning)` is fine & add it to the list of things to remove in #3283

Shouldn't we deprecate the OpenAPI query parameters instead? Or you only want the OR filter to have a `warnings.warn` to not forgot about it in https://github.com/open-formulieren/open-forms/issues/3283?

sergei-maertens:
> > Can you also deprecate that `filter_queryset` thing doing the `OR` filter? A `warnings.warn(..., DeprecationWarning)` is fine & add it to the list of things to remove in #3283
> 
> Shouldn't we deprecate the OpenAPI query parameters instead? Or you only want the OR filter to have a `warnings.warn` to not forgot removing it when #3283 gets tackled?

Only the OR filter behaviour, the query params themselves are obviously still used, just not at the same time (anymore).

sergei-maertens (approved):
Please rebase your PR in a couple of atomic commits :) then it's good to merge!

sergei-maertens (approved):

PR #3543, opened 2023-10-18
⬆️ Upgrade django-simplecertmanager to 1.4.1

Viicos:


requirements/base.txt
```diff
@@ -195,7 +197,7 @@ django-sendfile2==0.7.0
     # via django-privates
 django-sessionprofile==2.0.0
     # via django-digid-eherkenning
-django-simple-certmanager==1.3.0
+django-simple-certmanager==1.4.1
```
sergei-maertens, line 200: shouldn't the `base.in` be modified to properly specify the `[test]` extra?

Now this works by coincidence because we have a direct dependency on factory boy.

Viicos: The thing is the `[testutils]` extra is only required for tests to run, so I could add the dependency in `base.in`, and in `test-tools.in` as well with the extra, but I don't know how it will play with pipt-tools

sergei-maertens (approved):

PR #3545, opened 2023-10-19
🩹 Raise a custom exception instead of `RuntimeError`

### LINKED ISSUE #3540: Replace raw RuntimeError with semantic exception when a service is not configured
From the PR review discussion in #3490

Places where raw `RuntimeError` is used rather than a custom `NoServiceConfigured` exception class:

* `openforms.appointments.contrib.qmatic.client.QmaticClient`

Seems to be the only place. Other raw `RuntimeError` is semantically correct.

Viicos:
Fixes #3540.

### DIFF
--- a/src/openforms/appointments/contrib/qmatic/client.py
+++ b/src/openforms/appointments/contrib/qmatic/client.py
@@ -63,6 +63,10 @@ class BranchDetailDict(TypedDict):
     updated: int
 
 
+class NoServiceConfigured(RuntimeError):
+    pass
+
+
 # API CLIENT IMPLEMENTATIONS, per major version of the API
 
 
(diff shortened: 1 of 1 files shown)


src/openforms/appointments/contrib/qmatic/client.py
```diff
@@ -73,7 +77,7 @@ def QmaticClient() -> "Client":
     config = QmaticConfig.get_solo()
     assert isinstance(config, QmaticConfig)
```
Viicos, line 78: Should `cast` be used instead? Adds an extra import though, I guess this is just a personal preference

sergei-maertens: no, @CharString has strong feelings about that :)

sergei-maertens (approved):

src/openforms/appointments/contrib/qmatic/client.py
Viicos: I'm curious to know why :)

sergei-maertens: `cast` is a lie and the runtime behaviour can differ from the declared type, while `assert` enforces runtime correctness too. If assert is causing too much overhead, we can also strip it out and run with `python -O`

basically `cast` is a last resort if there are really no better options (and I agree)

Issue #2952, opened 2023-03-31
Allow steps to become applicable (reverse n/a)


LaurensBurger:
### Thema / Theme

Form designer

### Omschrijving / Description
extension of: #2444 
As a form designer i do not want to show all/any the steps a form might contain. As the forms i design contain a lot of steps that most of the visitors do not need.
I would like the possibility to hide all steps, especially at the "start" of a form and show them only when they become applicable, the complete opposite of how we handle the n/a steps at this time.

Example: form step 1 the user provides input which make makes step 3 4 5 "applicable", but step 2 6 7 remain hidden. 
nowhere in the process will the user see these steps which will create a "cleaner" experience especially with forms that have 10+ steps.

### Aanvullende opmerkingen / Additional context

Current situation:
![image](https://user-images.githubusercontent.com/112621537/229086656-cfef633c-3f56-42c8-b59a-225c9fd28a5f.png)
![image](https://user-images.githubusercontent.com/112621537/229087737-57d99a86-555a-42c0-b282-81c12403628b.png)

With the possibility to only show steps when they become applicable:
![image](https://user-images.githubusercontent.com/112621537/229088540-006a8466-16c3-4a81-bf07-abef0143d9eb.png)
![image](https://user-images.githubusercontent.com/112621537/229088728-914545a3-446c-4a48-b223-564aa23b7272.png)

joeribekker:
Refinement: Als nice to have kunnen we een formulier initieel n/a maken (a la hidden voor velden). De use case om met regel 1 iets n/a te maken, en met regel 2 weer applicable, is niet van toepassing (ik weet ook niet of dit kan met volgorderlijkheid van regels).

Bijvoorbeeld: Als de leeftijd groter is dan 18 jaar, dan moeten er additionele gegevens worden ingevuld die in een stap staan.

sergei-maertens:
yes, let's apply the same logic

CharString:
When I set the that global to hide steps, this form already does the trick:
[dubstep.zip](https://github.com/open-formulieren/open-forms/files/13054809/dubstep.zip)

Almost.. because we don't evaluate the logic on the "start form" page, all steps show. But once you start the form they disappear.
So the smallest change that enables the use case is to not show the progress bar on the start form page.

Viicos:
[...]

Issue #3239, opened 2023-07-10
Remove logic rule logging


joeribekker:
Investigate if this feature is used (look at access.log).

As opposed to #2998

sergei-maertens:
Same check on test environments (limited access logs to `/admin/` URLs):

5 / 15533  are logic log views, which is `0.03%`. I think we can remove this.

joeribekker:
Refinement: @sergei-maertens will bring this up in refinement at DH

sergei-maertens:
We can remove it *before* building an alternative.

PR #3551, opened 2023-10-23
✨ Add the ability to set a form as non applicable by default

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/2952.

Todo:

- [x]  Validate that first step is applicable (if not, the submission cannot start). ~~I think a tooltip (using the [`title`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) HTML attribute) could be interesting to use here (requires an additional prop to our `Field` component), besides making it disabled:~~ -> reverted
image

- [x] Should we check if steps are applicable on the Start Page? Currently:
image
After accessing first step (if _Hide non-applicable form steps_ is set, n/a steps wouldn't even appear):
image
https://github.com/open-formulieren/open-forms-sdk/pull/575


SilviaAmAm (commented):
I only made it to 12/16 files, I will continue reviewing tomorrow!

src/openforms/forms/models/form_step.py
```diff
@@ -102,3 +108,11 @@ def delete(self, *args, **kwargs):
 
     def iter_components(self, recursive=True, **kwargs):
         yield from self.form_definition.iter_components(recursive=recursive, **kwargs)
+
+    def clean(self) -> None:
+        if not self.is_applicable and self.order == 0:
+            raise ValidationError(
+                {"is_applicable": _("First form step must be applicable.")},
+                code="invalid",
+            )
```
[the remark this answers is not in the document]
Viicos: I find it really sad to have this discussion still open: https://github.com/encode/django-rest-framework/discussions/7850

I've added a custom validator

src/openforms/submissions/logic/actions.py
```diff
@@ -191,6 +191,42 @@ def get_action_log_data(
...
+        all_variables: Dict[str, FormVariable],
+        initial_data: dict,
+        log_data: dict[str, JSONValue],
+    ) -> JSONObject:
+        return {
+            "type_display": LogicActionTypes.get_label(
+                LogicActionTypes.step_applicable
+            ),
+            "step_name": self.form_step_identifier,
+        }
```
SilviaAmAm, line 227: I see from codecov that this is not covered by tests?

src/openforms/submissions/models/submission_step.py
```diff
@@ -196,7 +196,15 @@ def can_submit(self) -> bool:
...
     def is_applicable(self) -> bool:
-        return self._is_applicable
+        if self._is_applicable is not None:
+            return self._is_applicable
+        return self.form_step.is_applicable
+
+    @is_applicable.setter
+    def is_applicable(self, value: bool) -> None:
+        if not isinstance(value, bool):
+            raise ValueError(f"'is_applicable' expects a boolean value, got {value}")
```
SilviaAmAm, line 206: In which case is this not a bool?

src/openforms/forms/models/form_step.py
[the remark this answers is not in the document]
Viicos: Yes I think the best thing in this case is to centralize validation logic in a single place

src/openforms/forms/api/validators.py
```diff
@@ -190,6 +190,19 @@ def __call__(self, attrs: dict, serializer: serializers.Serializer):
             )
 
 
+class FormStepIsApplicableIfFirstValidator:
+    def __call__(self, attrs: dict):
+        if not attrs.get("is_applicable", True) and attrs.get("order") == 0:
```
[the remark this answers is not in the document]
Viicos: We have to negate one of the condition anyhow. With your suggestion, if `is_applicable` is `True` and `order` is `0`, it will raise a validation error even though it is still correct

> Since `is_applicable` has a default

It has one on the model but I think it is not present in the data dict if not changed by the user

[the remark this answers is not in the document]
SilviaAmAm: I am an idiot, I don't know what I was thinking... nvm :P

src/openforms/forms/tests/test_api_formsteps.py
```diff
@@ -1092,3 +1092,42 @@ def test_update_with_translations_validate_literals(self, _mock):
...
+
+
+class FormStepsAPIApplicabilityTests(APITestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.user = UserFactory.create()
+        self.form = FormFactory.create()
+        self.form_definition = FormDefinitionFactory.create()
+        self.client.force_authenticate(user=self.user)
```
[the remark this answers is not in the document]
Viicos: Done, and added the happyflow test

src/openforms/submissions/models/submission_step.py
Viicos: Currently none (hopefully), but it might be better to have this `ValueError` raised here with a nice message instead of an unhandled exception somewhere else?

src/openforms/submissions/logic/actions.py
Viicos: Saw that too, but iirc the other `get_action_log_data` calls aren't either

src/openforms/submissions/models/submission_step.py
sergei-maertens: it's still an unhandled exception, no? The zen of python states "it's better to ask for forgiveness than permission". I prefer not having this sort of defensive "what if" programming unless there's a good concrete reason.

Viicos: > Additionally - this is really the domain of the type checker to prevent mistakes/bugs like that.

True but in this case `is_applicable` could be set dynamically when applying some kind of logic. I'll make the change

src/openforms/submissions/logic/actions.py
sergei-maertens: the logic logging is up for rework, but I'll only have a meeting about that with DH in a couple of weeks. We do know that the current state of logging is untrustworthy, so okay to skip tests here too.

Related ticket: #3239 and that probably links to a bunch of other tickets.

sergei-maertens (approved):

sergei-maertens:
Please rebase on master to confirm that the ci failures are resolved, after that you can merge this!

PR #3552, opened 2023-10-23
Confirm ability to use dynamic options in selectboxes

### LINKED ISSUE #2629: Add logic/conditions to items in a group (selectboxes/radios/dropdown)
### Thema / Theme

Form designer

request: Utrecht 27

As a form designer i want to be able to show and hide items from a group of answers/options. Just like i can show and hide a group or field.
This feature will reduce the amount of fields i need to create to build a form as each deviation from the "common" answers is now also contained in the same group instead of creating a separate object for them.

Current situation:
image

New situation:
image

image

_No response_

_No response_

sergei-maertens:
Closes #2629 

Discussed with Joeri - we are not backporting this to older versions.

Added documentation for extending the options of a dropdown/selectboxes/radio field based on particular logic rules. There was a "bug" or oversight in the variables editor that prevents you from setting this up in older versions, but you can use logic rules as a workaround for that to initialize the variable.

* Added stories for some components used in the form designer (this was easier to debug/implement the UX)
* Added documentation with importable example
* The array input widget now has a toggle between list-of-strings nice UI & raw JSON input.
* If the items in the array are anything other than string literals, then the UI is fixed to raw JSON input to avoid data corruption.


docs/manual/forms/examples/dynamic_options_3.rst
```diff
@@ -0,0 +1,125 @@
...
+               "var": "checkboxOptions"
+             },
+             [
+               [
+                 "option3",
+                 "Option 3"
+               ]
+             ]
+           ]
+         }
```
CharString, line 112: What happens if the submitter repeatedly toggles "Toon optie 3"?

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
CharString, line 1: What's the story?
Changing the Prefill-bron or Datatype selections, behaves weird.
If this is supposed to just show the wrapped array thing in a context, I think the story title should reflect that.

src/openforms/js/components/admin/form_design/logic/LiteralValueInput.stories.js
CharString, line 1: Toggling the "Use raw JSON input" doesn't keep the values I type in. They're tied to the value set in the Contols

CharString (changes requested):

src/openforms/js/components/admin/form_design/logic/LiteralValueInput.js
```diff
@@ -28,7 +31,18 @@ CheckboxChoices.propTypes = {
 };
 
 const WrapperArrayInput = ({name, value, onChange}) => {
-  return (
+  const [useRawJSON, setUseRawJSON] = useState(false);
+
+  // if the value is not an array of strings, the text inputs break the existing data
+  // structures. In that case, pin the input to use raw JSON.
```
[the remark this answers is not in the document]
sergei-maertens: yep, I missed that. I hope we can migrate the backend JS to TS too in 2024 so that the compiler can at least catch this...

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
sergei-maertens: I tried debugging/developing it through this component which I thought was responsible, while implementing it it turned out that I could isolate `LiteralValueInput` further.

The story is "this component exists and that's about what it looks like" like the majority of the other stories.

> Changing the Prefill-bron or Datatype selections, behaves weird.

It shows the triggered "Actions" in the panel for me, which shows that the `onChange` event handler (prop of the component) was called.

image

I understand that you'd expect the UI to update, but there is no parent state being managed here - the story shows the props and how they get used.

I don't know what a better name is - the storybook convention seems to be to just go with `Default` for a story without interaction tests or special things to just show the component in isolation.

src/openforms/js/components/admin/form_design/logic/LiteralValueInput.stories.js
sergei-maertens: intended - the `onChange` handler is called and storybook reflects that:

image

CharString (approved):

PR #3555, opened 2023-10-24
:arrow_up: Bump to form builder 0.8.0

sergei-maertens:
Added support for the component types:

* date
* datetime
* time
* phoneNumber
* postcode


package.json
```diff
@@ -66,6 +66,10 @@
     ],
     "modulePaths": [
       "<rootDir>/src/openforms/js"
+    ],
+    "transformIgnorePatterns": [
```
vaszig, line 70: Is this added for performance reasons?Do we ignore them because we don't test them?

vaszig (approved):

package.json
sergei-maertens: by default everything in `node_modules` is ignored because it's assumed it's already compiled to CommonJS.

However, some libraries ship the modern ESM (EcmaScript Modules) format, and not all tools understand that syntax (the `import` and `export` statements in particular).

This rule opts in to compiling packages that you know are not CommonJS yet, so it's kind of the opposite of performance.

PR #3556, opened 2023-10-26
:bug: Don't re-nest item, but prop spread it

### LINKED ISSUE #3557: Picking an informatieobjecttype from zgw registration backend for file upload component crashes

latest / 2.4.x

The file component allows (optionally) specifying a document type from the ZGW API's to use for registration.

However, when trying to view the available options in the dropdown, there is a javascript crash and no options are displayed.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #3557 

Stumbled upon this when trying to reverse engineer The Thing for the new form builder.

Fix confirmed with a real Open Zaak connection.

### DIFF
--- a/src/openforms/js/components/form/file.js
+++ b/src/openforms/js/components/form/file.js
@@ -56,7 +56,7 @@ const REGISTRATION = {
             // no need for label if we have one backend
             if (backends.length == 1 && labeledData.length) labeledData[0].backendLabel = '';
             const labeledItems = labeledData
-              .map(({backendLabel, data}) => data.map(item => ({backendLabel, item})))
+              .map(({backendLabel, data}) => data.map(item => ({backendLabel, ...item})))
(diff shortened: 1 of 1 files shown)


CharString (approved):
Can it become a problem if this clobbers an eventual `backendLabel` property we get from the (our) api response?

I don't know enough about the call sites of `InformatieObjectTypeSelectField.setItems` a FormIO/choices api?.

sergei-maertens:
I don't really think so - and the formio select & underlying library kinda suck anyway. The setItems api goes away in the new builder which at least has typing information about what options/groups are supposed to look like.

sergei-maertens:
Backport: b0a7bdf5c256b5cf63f86e966bc9fb5e5ef7add7

Issue #3553, opened 2023-10-24
ZGW Registration backend validation error without message


Bartvaderkin:
### Product versie / Product version

latest @ open-forms.test.maykin.opengem.nl

### Omschrijf het probleem / Describe the bug

ZGW Registration backend configuration screen shows validation error indicator on the tab but no message is shown and data is not saved:

![image](https://github.com/open-formulieren/open-forms/assets/54805673/9d3c6c96-2bb2-40fd-be94-57476a127e6e)

(via https://open-forms.test.maykin.opengem.nl/admin/forms/form/245/change/)

### Stappen om te reproduceren / Steps to reproduce

1. Create or edit a form
2. Added ZGW Apis to Registration backends
3. Enter the values as on the screenshot (copy-paste below)
````
Zaaktype: https://test.openzaak.nl/catalogi/api/v1/zaaktypen/2c1feba6-3163-4e15-9afa-fa4f01dcb4f9
IO type: https://test.openzaak.nl/catalogi/api/v1/informatieobjecttypen/1db73aba-cef4-42ec-bdf0-cd4c7c4f2514
Rol type: https://test.openzaak.nl/catalogi/api/v1/roltypen/335ba3f6-c55e-42d0-9b66-108edcce2e18
RSIN: 900222086
````
4. Save the form
5. Note the error indicator but no message.
6. On full page reload no data was saved

### Verwacht gedrag / Expected behavior

I'd expect validation error messages to be visible so we can fix the problem.

### Screen resolution

bigger

### Device

Desktop

### OS

Windows

### Browser

Firefox

Viicos:
Don't know if it is related, but I also have an error when I have to ZGW API set configured, because:

https://github.com/open-formulieren/open-forms/blob/66d9beea7aaa6eaa1737aec30a40ac1d2416e04f/src/openforms/registrations/contrib/zgw_apis/plugin.py#L72-L90

`RoltypeOmschrijvingValidator` expects `zgw_api_group` to exist, and is run before `validate`

PR #3560, opened 2023-10-27
🐛 Fix validation in `ZaakOptionsSerializer`

Viicos:
Part of https://github.com/open-formulieren/open-forms/issues/3553

For some context: https://github.com/open-formulieren/open-forms/pull/3146#discussion_r1228246353

When no ZGW API set is configured, we would get a `KeyError` in `RoltypeOmschrijvingValidator`, because it was running before `validate`:

https://github.com/open-formulieren/open-forms/blob/66d9beea7aaa6eaa1737aec30a40ac1d2416e04f/src/openforms/registrations/contrib/zgw_apis/plugin.py#L72-L90

By providing a default value, it will be used by `RoltypeOmschrijvingValidator` if not provided by the user.

---

~~Looking at the admin frontend, a json schema rendered form is used (dynamically), and error handling doesn't seem to work, hence the https://github.com/open-formulieren/open-forms/issues/3553 issue.~~ Seems to work:
image


sergei-maertens (commented):
this behaviour is not compatible, is it? because once it obtains the default, it's configured at the form level which ZGW API group to use, but if you then change the global configuration default, existing forms are still pointing to the old record.

Viicos:
> if you then change the global configuration default, existing forms are still pointing to the old record.

Ok now I understand the wanted behaviour (this also explains why `validate` would only check if a default api group was available, but it would not alter `attrs` before returning it)

Viicos:
The usage of `get_zgw_config` might feel a bit hacky, but couldn't find any better solution

Viicos:
Now regarding the original issue, we can still get 500 errors if requesting the catalogi client fails (due to `response.raise_for_status()`).

Should we catch them and throw a generic `ValidationError` (_something went wrong when checking catalogi blabla_), or give more info about the catalogi error in the validation error description?

sergei-maertens:
> Now regarding the original issue, we can still get 500 errors if requesting the catalogi client fails (due to `response.raise_for_status()`).
> 
> Should we catch them and throw a generic `ValidationError` (_something went wrong when checking catalogi blabla_), or give more info about the catalogi error in the validation error description?

An HTTP 4xx status code doesn't make sense here, because it's not a client error (the client cannot change their call to make it pass), so leaving it a HTTP 500 is fine.

What we can do instead is a similar approach to the SDK - have a generic error boundary that gets information about the status code and show a (more) appropriate message. However, that's too big of a scope for this ticket, it requires quite a bit of restructuring and there is a redesign planned, so I think we should keep that thought as a requirement for the new design.

Viicos:
> An HTTP 4xx status code doesn't make sense here, because it's not a client error (the client cannot change their call to make it pass), so leaving it a HTTP 500 is fine.

Yes apart from a 404 error, this makes sense (and I don't think a 404 could ever happen considering we are making a list request, so if query params are wrong it will simply return a 200 with an empty list).

So if this is out of scope I think this PR is ready for review

sergei-maertens (approved):
Rebase on master to check the CI failures resolution, then we can merge this

PR #3561, opened 2023-10-27
[#3524] Ignore family members without BSN

### LINKED ISSUE #3524: Deal with the case where Haal Centraal returns a partner without BSN
One of the test BSN of Haal Centraal v2 gives this back:

```json
{"type":"RaadpleegMetBurgerservicenummer","personen":[{"kinderen":[],"partners":[{"naam":{"voornamen":"Jean Marie","geslachtsnaam":"Beaudelaire","voorletters":"J.M."}}]}]}
```

However, we always expect a BSN for the family members component.

In Stuff BG, we ignore any family member without BSN, so maybe we should do the same thing.

SilviaAmAm:
Fixes #3524

### DIFF
--- a/docs/manual/forms/form_fields.rst
+++ b/docs/manual/forms/form_fields.rst
@@ -234,7 +234,7 @@ Gezinsleden
 ===========
 
 De *Gezinsledenscomponent* toont de namen van alle gezinsleden (partners en/of kinderen) als selectievakjes.
-Gezinsleden **zonder** BSN worden niet getoond.
+Gezinsleden zonder BSN **worden niet getoond**.
 De waarden die naar de registratie backend worden gestuurd zijn de BSNs.
 
 De gegevens voor de gezinsleden worden opgehaald van Haal Centraal BRP Personen Bevragen of StUF-BG. Dit kan in de Admin
--- a/docs/manual/forms/form_fields.rst
+++ b/docs/manual/forms/form_fields.rst
@@ -234,6 +234,7 @@ Gezinsleden
 ===========
 
 De *Gezinsledenscomponent* toont de namen van alle gezinsleden (partners en/of kinderen) als selectievakjes.
+Gezinsleden **zonder** BSN worden niet getoond.
 De waarden die naar de registratie backend worden gestuurd zijn de BSNs.
 
 De gegevens voor de gezinsleden worden opgehaald van Haal Centraal BRP Personen Bevragen of StUF-BG. Dit kan in de Admin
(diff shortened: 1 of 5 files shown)


docs/manual/forms/form_fields.rst
```diff
@@ -234,6 +234,7 @@ Gezinsleden
 ===========
 
 De *Gezinsledenscomponent* toont de namen van alle gezinsleden (partners en/of kinderen) als selectievakjes.
+Gezinsleden **zonder** BSN worden niet getoond.
```
CharString, line 237:
```suggestion
Gezinsleden zonder BSN **worden niet getoond**.
```
? I'm not sure.

src/openforms/js/components/form/np-family-members.js
```diff
@@ -55,6 +55,11 @@ class NpFamilyMembers extends SelectBoxes {
       key: 'basic',
       label: 'Basic',
       components: [
+        {
+          type: 'htmlelement',
+          tag: 'div',
+          content: 'Note that any family member without a BSN will not be displayed.',
+        },
```
CharString, line 62: Nice.

CharString (approved):
It's fine, so I approve, just a remark and a question:

- The v2 mocks return a different response for essentially the same call to the same endpoint. If I understand correctly, the real endpoint response would always contain both children and partners. I'd try to avoid these kinds of mocks that assume a certain implementation. In this case it never tries to reference the partners key if they aren't requested. For instance: a refactor of the implementation that always processes everything from the api response and only filters just what's being asked in the answer, would work against the real api, but trigger these tests. (contrived example, but in essence, I like be able to refactor without changing the test suite and have the tests as a canary in the coal mine to detect when I broke something)

- Is just mentioning the omission of people without BSN is enough information for a form designer to be able to cater for these cases? Do we need to give examples or suggestions, like a repeating group for "missing familiy members"? I guess they already have to deal with submitters that have no BSN themselves and when the BRP services are unreachable for some reason.

sergei-maertens:
> * Is just mentioning the omission of people without BSN is enough information for a form designer to be able to cater for these cases? Do we need to give examples or suggestions, like a repeating group for "missing familiy members"? I guess they already have to deal with submitters that have no BSN themselves and when the BRP services are unreachable for some reason.

They definitely can't cater for these cases, but to be able to do that they're going to have to tell us how they expect to handle that. I think we have too little domain knowledge and attempts to get that knowledge from actual teams building forms have been

sergei-maertens:
My sausage fingers accidentally closed the PR, sorry!

sergei-maertens (approved):
Please rebase on master to sort out the CI failures

PR #3569, opened 2023-10-31
🐛 Fix prefill handling for migrated forms

### LINKED ISSUE #3467: Copy form: null value in column "prefill_plugin" violates not-null constraint

2.3.0

Migrated form components contain incorrect "prefill" json:
```json
  "prefill": {
    "plugin": null,
    "attribute": null,
    "identifierRole": "main"
  },
```
This causes crashes on import and copy of forms.

Updating the component changes null to "" and solves the copy issue
```json
  "prefill": {
    "plugin": "",
    "attribute": "",
    "identifierRole": "main"
  },
```

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

Viicos:
Fixes #3467.

After a lot of time trying to find where it was coming from and how to fix it, I came across:

https://github.com/open-formulieren/open-forms/blob/f221ef67053c276e8209d58febfe48e2c0168c7d/src/openforms/forms/utils.py#L349-L364

which applies migrations on FD configuration components. The issue is it seems to be specifically designed for migrations related to specific component types, and in my case I need to apply a migration to any field type. ~~Should I refactor this utility?~~ discussed with Sergei, explicit is better even if it is repetitive


src/openforms/formio/migration_converters.py
```diff
@@ -45,10 +45,50 @@ def move_time_validators(component: Component) -> bool:
...
+    # Special components
+    "bsn": {
+        "alter_prefill_default_values": alter_prefill_default_values,
+    },
 }
+"""A mapping of the component types to their converter functions.
+
+Keys are ``component["type"]`` values, and values are dictionaries keyed by a
+unique converter identifier and the function to do the actual conversion.
+"""
```
Viicos, line 94: For better IDE experience, so that we can see what it is on hover

src/openforms/formio/migration_converters.py
```diff
@@ -45,10 +45,50 @@ def move_time_validators(component: Component) -> bool:
...
+        component["prefill"]["plugin"] = ""
+        altered = True
+    if prefill_attribute is None:
+        component["prefill"]["attribute"] = ""
+        altered = True
+
+    return altered
+
+
 CONVERTERS: dict[str, dict[str, ComponentConverter]] = {
```
Viicos, line 68: I was thinking we could add a `Literal` type alias somewhere referencing all the available types? e.g. `ComponentTypes: TypeAlias = Literal["textfield", ...]`.

It could be used as the key type here, but could also be useful for people to have reference

src/openforms/formio/migration_converters.py
```diff
@@ -45,10 +45,50 @@ def move_time_validators(component: Component) -> bool:
...
-# identifier and the function to do the actual conversion.
+def alter_prefill_default_values(component: Component) -> bool:
+    """A converter that replaces ``prefill`` dict values from ``None`` to an empty string."""
+    if "prefill" not in component:
+        return False
+
+    altered = False
+
+    prefill_plugin = component["prefill"]["plugin"]
+    prefill_attribute = component["prefill"]["attribute"]
```
sergei-maertens, line 56: there's a chance that `prefill` is present, but doesn't have the keys `plugin` or `attribute` (or is even `None`), so I'd be a bit more defensive here:

```py
unset = object()

if not (prefill := component.get("prefill") or {}):
    return False

prefill_plugin = prefill.get("plugin", unset)
if prefill_plugin is None:
    ... etc.
```

src/openforms/formio/migration_converters.py
sergei-maertens: I don't think the tradeoff of effort vs. benefit is worth it at this time.

The typescript type definitions will serve as The Reference when they're completed and should be the source of truth.

The python TypedDict approach has other drawbacks, preferably I'd go through a `dict -> dataclasses/pydantic -> dict -> JSON` flow so that we can benefit from python typing without suffering the typeddict limitations

sergei-maertens (changes requested):

src/openforms/formio/migration_converters.py
Viicos: I see, but without having to deal with `TypedDict`, the use case here could be:

`CONVERTERS: dict[ComponentType, dict[str, ComponentConverter]] = {...}`

src/openforms/formio/migration_converters.py
Viicos: Added tests for this as well

src/openforms/formio/migration_converters.py
sergei-maertens: And our own backend definitions, it would be great to have a similar type alias in the backend:

```py
AnyComponent = TextFieldComponent | CheckboxComponent | ...
```

and if possibly then infer `AnyComponent.type` from that, but I'm not even sure if that's technically possibly in python.

Perhaps we need to do it the other way around and define a literal for possibly types and constrains the base `Component` protocol/class using that. So all in all - too much uncertainty at the moment that needs to be cleared up first instead of haphazardly sprinkling types around over the entire codebase.

Viicos: > but I'm not even sure if that's technically possibly in python.

I'm afraid it isn't.

> too much uncertainty at the moment that needs to be cleared up first instead of haphazardly sprinkling types around over the entire codebase.

Yes, and dataclasses or dataclasses like libraries like pydantic could be used instead in the end (overwriting typedict fields is not allowed yet)

sergei-maertens (approved):

PR #3574, opened 2023-11-02
🐛 Handle hash based frontend routing

### LINKED ISSUE #3362: Server side redirects to the frontend break when the SDK uses hash based routing

latest, 2.2.x, 2.1.x

The backend uses deep-links at times (submission resume, redirect back after authentication, change appointment and possible other cases) to the form on the frontend. It does this by taking `submission.form_url` and modifiying the URL with additional path segments and/or query string parameters. `submission.form_url` is recorded when a submission is started and is the only way to know where the public URL of the form is.

However, when you enable hash-based routing, the backend does not know this. Redirects are broken in those situations, as hash-based routing is typically used when embedding forms in a CMS where defining catch-all routes for a particular form is not (reasonably) achievable.

E.g. a form can be available on:

* example.com/formulieren/foo
* example.com/formulieren/foo#/

Possible solutions:

* use some other form of input (query param, possibly HTTP header within the redirect response) to indicate which "route" or action is desired (`/stap/bar`, `/afspraak-annuleren`, `submission-start` etc.). This would decouple the backend from the specific routing solution AND any client-side URLs
* record the routing type during submission creation so that the backend can generate the appropriate URLs

_No response_

Server side redirects work in both cases.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

**Tasks**

- [x] Implement building redirect URL in backend
- [x] Update SDK to interpret backend actions & action args
- [ ] Add data migration to clean `submission.form_url`
- [ ] Deprecate "cleaned_form_url" wrapper.

Viicos:
Fixes #3362


sergei-maertens (commented):
I think I'd prefer a query param `_action=foo` rather than `redirect_path` so that we can decouple the actions from the actual client-side URLs.

We only need to synchronize the list of literal action names then and our backend and frontend become more agnostic of each other.

Viicos:
One thought that I had: The number of `action_args` arguments passed in `get_frontend_redirect_url` could be validated depending on the action used, but this is only beneficial for the developer. overloads could be useful but a bit overkill and cumbersome

src/openforms/appointments/tests/test_views.py
```diff
@@ -50,9 +50,10 @@ def test_good_token_and_submission_redirect_and_add_submission_to_session(self):
             response = self.client.get(endpoint)
 
         expected_redirect_url = (
-            furl("http://maykinmedia.nl/myform/afspraak-annuleren")
+            furl("http://maykinmedia.nl/myform")
             .add(
                 {
+                    "_action": "afspraak-annuleren",
```
sergei-maertens, line 56: so right now with hash based routing, this would result in a URL like `example.com/myform?time=...&submission_uuid=...#/afspraak-annuleren` while it should be:

`example.com/myform#/afspraak-annuleren?time=...&submission_uuid=...` - i.e. the time & submission_uuid really should be `_action_params` as they're directly related to the action.

src/openforms/appointments/tests/test_views.py
```diff
@@ -398,8 +399,8 @@ def test_good_token_and_submission_redirect_and_add_submission_to_session(self):
         # after initiating change, we expect the bsn to be stored in plain text (again)
         self.assertEqual(new_submission.auth_info.value, "000000000")
         expected_redirect_url = (
-            f"http://maykinmedia.nl/myform/stap/{form_definition.slug}"
-            f"?submission_uuid={new_submission.uuid}"
+            f"http://maykinmedia.nl/myform/?_of_action=stap&_of_action_args%3D{form_definition.slug}"
+            f"&submission_uuid={new_submission.uuid}"
         )
```
sergei-maertens, line 404: perhaps we can clean up this URL building to use `furl` if we're touching the code anyway? That probably makes the "double URLencoding" of the action_args more clear.

src/openforms/appointments/tests/test_views.py
```diff
@@ -398,8 +399,8 @@ def test_good_token_and_submission_redirect_and_add_submission_to_session(self):
         # after initiating change, we expect the bsn to be stored in plain text (again)
         self.assertEqual(new_submission.auth_info.value, "000000000")
         expected_redirect_url = (
-            f"http://maykinmedia.nl/myform/stap/{form_definition.slug}"
-            f"?submission_uuid={new_submission.uuid}"
+            f"http://maykinmedia.nl/myform/?_of_action=stap&_of_action_args%3D{form_definition.slug}"
```
sergei-maertens, line 402: please make this `_of_action=resume` with a `stepSlug` action param

src/openforms/appointments/tests/test_views.py
```diff
@@ -659,9 +660,15 @@ def test_after_successful_auth_redirects_to_form(self):
...
-        expected_redirect_url.args["submission_uuid"] = str(new_submission.uuid)
+        expected_redirect_url.add(
+            {
+                "_of_action": "stap",
+                "_of_action_args": {
+                    "next_step": "test-step",
+                    "submission_uuid": new_submission.uuid,
+                },
+            }
+        )
```
sergei-maertens, line 671: this does not produce the expected result:

```py
>>> url = furl("https://example.com")
>>> url.add({"_of_action_args": {"next_step": "test-step", "id": "foo"}})
furl('https://example.com?_of_action_args=next_step&_of_action_args=id')
>>> str(url)
'https://example.com?_of_action_args=next_step&_of_action_args=id'
```
expected is:

```py
>>> url = furl("https://example.com")
>>> furl().add({"next_step": "test-step", "id": "foo"}).query
Query('next_step=test-step&id=foo')
>>> url.add({"_of_action_args": str(furl().add({"next_step": "test-step", "id": "foo"}).query)})
furl('https://example.com?_of_action_args=next_step%3Dtest-step%26id%3Dfoo')
>>> str(url)
'https://example.com?_of_action_args=next_step%3Dtest-step%26id%3Dfoo'
```

sergei-maertens (changes requested):

Viicos:
There are still some failing tests, mainly because:

https://github.com/open-formulieren/open-forms/blob/feb37d3bab9dcddd5a2edc0fcdee57abec665c44/src/openforms/frontend.py#L23-L27

This performs a first round of URL encoding, e.g. `query={"_of_action_params": "time=00%3A00"}` (with `action_params={"time": "00:00"}`)

Then furl does a second round of URL encoding, resulting in `?_of_action_params=time%3D00%253A00`:
- `=` is correctly encoded as `%3D`
- `%` was incorrectly encoded as `%25`

Another option would be to have `query=urllib.parse.urlencode(json.dumps(action_params))`

src/openforms/submissions/tests/test_resume_form_view.py
```diff
@@ -50,11 +50,13 @@ def test_good_token_and_submission_redirect_and_add_submission_to_session(self):
...
-        # furl adds paths with the /= operator
-        f /= "stap"
-        f /= submission.get_last_completed_step().form_step.slug
-        # Add the submission uuid to the query param
-        f.add({"submission_uuid": submission.uuid})
+        # Add the redirect path and submission uuid to the query param
+        f.add(
+            {
+                "_of_action": "stap",
+                "_of_action_params": f"step_slug={submission.get_last_completed_step().form_step.slug}&submission_uuid={submission.uuid}",
```
sergei-maertens, line 57: I strongly dislike these long lines with complicated f-strings that also require you to manually do the URL encoding.

At this point, all I care about is that:

* the right parameters are specified
* they have been serialized (urlencoded or JSON, it's not relevant)

The exact form and order of parameters is not relevant and a source of hard-to-debug tests, so I encourage writing a small helper for this.

src/openforms/submissions/tests/test_views.py
```diff
@@ -54,9 +55,17 @@ def test_successfully_submit_form(self):
         form["code"] = submission.public_registration_reference
         submission_response = form.submit()
 
+        expected_redirect_url = furl("http://url-to-form.nl/")
+        expected_redirect_url.add(
+            {
+                "_of_action": "cosign",
+                "_of_action_params": f"submission_uuid={submission.uuid}",
+            }
+        )
```
sergei-maertens, line 64: Maybe a helper assertion with this signature would be nice:

```py
def assertRedirectsToFrontend(
    self,
    response,
    frontend_base_url: str,  # *may* contain query string parameters
    action: Literal[...],  # defined in openforms.frontend
    action_params: dict[str, str],
    **kwargs,
) -> None:
```

the assertion helper then performs the encoding itself (and you can write unit tests for the assertion itself too) and building the final URL as a string, and finally calls the underlying `self.assertRedirects(...)`

src/openforms/tests/test_frontend_utils.py
```diff
@@ -0,0 +1,52 @@
...
+            {"action_1": "arg_1", "action_2": "arg_2"},
+        )
+        self.assertEqual(
+            redirect_url,
+            "https://example.com/basepath?_of_action=resume&_of_action_params=action_1%3Darg_1%26action_2%3Darg_2",
+        )
+
+    def test_frontend_redirect_hash(self):
+        submission = SubmissionFactory.create(
+            form_url="https://example.com/basepath#after_hash",
```
sergei-maertens, line 25: good to have it (and keep it!), but let's also do the data migration where we clean/sanitize the URLs for existing submissions.

src/openforms/tests/test_frontend_utils.py
```diff
@@ -0,0 +1,52 @@
+from django.test import TestCase
+
+from openforms.frontend import get_frontend_redirect_url
+from openforms.submissions.tests.factories import SubmissionFactory
+
+
+class FrontendRedirectTests(TestCase):
```
sergei-maertens, line 7: Nice tests! Can you add a case with a querystring parameter in `Submission.form_url`?

src/openforms/tests/test_frontend_utils.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+
+        redirect_url = get_frontend_redirect_url(
+            submission,
+            "resume",
+            {"action&=": "arg_1"},
+        )
+        self.assertEqual(
+            redirect_url,
+            "https://example.com/basepath?_of_action=resume&_of_action_params=action_1%3Darg_1",
```
sergei-maertens, line 51: what's happening here? Is the `&` even a legal query param key name character?

src/openforms/tests/test_frontend_utils.py
Viicos: Added the migrations, not sure if it covers all cases? Let me know if you think some other cases should be taken into account into this migration

docs/developers/sdk/embedding.rst
```diff
@@ -57,7 +57,19 @@ Available options
...
+
+    .. note::
+        ``basePath`` only applies when using the default browser routing. If hash based routing
+        is used (see ``useHashRouting`` below), the option will be silently ignored.
+
+``useHashRouting``:
+    Whether hash based routing should be used. Defaults to ``false``. This option is useful when embedding
+    Open Forms with a CMS. If the SDK is hosted at ``https://example.com/basepath?cms_query=1``, the resulting URL
+    would be ``https://example.com/basepath?cms_query=1#/startpagina`` (SDK specific query parameters would come
+    at the end of the URL).
```
sergei-maertens, line 72: I'd add something like:

```rst
.. note::
   This is a last resort solution - preferably the backend where you embed the form would set up "wildcard" routes to
   ensure that refreshing the page works, e.g. `/some-path/<form-id>/*` should just load the CMS page for a specific form.
```

src/openforms/appointments/tests/test_views.py
```diff
@@ -20,13 +20,14 @@
     SubmissionFactory,
     SubmissionStepFactory,
 )
+from openforms.tests.mixins import FrontendRedirectMixin
```
sergei-maertens, line 23: Test helpers go in `openforms.utils.tests`, see the documentation: https://open-forms.readthedocs.io/en/stable/developers/backend/core/testing-tools.html

In this particular case I'd rather put them in `openforms.frontend.tests` even and import from there & this documentation page needs to be updated to mention the test helper.

src/openforms/appointments/tests/test_views.py
```diff
@@ -20,13 +20,14 @@
...
 
 from ..tokens import submission_appointment_token_generator
 from .factories import AppointmentFactory, AppointmentInfoFactory
 
 
 @freeze_time("2021-07-15T21:15:00Z")
-class VerifyCancelAppointmentLinkViewTests(TestCase):
+class VerifyCancelAppointmentLinkViewTests(TestCase, FrontendRedirectMixin):
```
sergei-maertens, line 30:
```suggestion
class VerifyCancelAppointmentLinkViewTests(FrontendRedirectMixin, TestCase):
```
mixins come before the base class for consistent MRO

src/openforms/submissions/migrations/0079_cleanup_urls.py
```diff
@@ -0,0 +1,29 @@
...
+def cleanup_submission_urls(
+    apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+
+    Submission = apps.get_model("submissions", "Submission")
+
+    for submission in Submission.objects.all():
+        f = furl(submission.form_url)
+        f.remove(fragment=True)
+        submission.form_url = f.url
```
sergei-maertens, line 18: this should also strip the `/startpagina` bit if present

src/openforms/submissions/migrations/0079_cleanup_urls.py
```diff
@@ -0,0 +1,29 @@
...
+from furl import furl
+
+
+def cleanup_submission_urls(
+    apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+
+    Submission = apps.get_model("submissions", "Submission")
+
+    for submission in Submission.objects.all():
```
sergei-maertens, line 15:
```suggestion
    for submission in Submission.objects.iterator():
```
there may be many records and you want to avoid loading them all in memory

src/openforms/submissions/migrations/0079_cleanup_urls.py
```diff
@@ -0,0 +1,29 @@
...
+        submission.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("submissions", "0078_submission_finalised_registration_backend_key"),
+    ]
+
+    operations = [migrations.RunPython(cleanup_submission_urls)]
```
sergei-maertens, line 29: please add the `migrations.RunPython.noop` reverse action so we can still reverse apply migrations.

src/openforms/tests/test_frontend_utils.py
```diff
@@ -0,0 +1,115 @@
...
+            submission,
+            "resume",
+            action_params,
+        )
+
+        excpected_redirect_url = furl("https://example.com/basepath")
+        excpected_redirect_url.add(
+            {
+                "_of_action": "resume",
+                "_of_action_params": (json.dumps(action_params)),
```
sergei-maertens, line 29:
```suggestion
                "_of_action_params": json.dumps(action_params),
```

sergei-maertens (changes requested):

src/openforms/tests/test_frontend_utils.py
Viicos: I was thinking black would pick it up when I left it :(

sergei-maertens (approved):

PR #3581, opened 2023-11-07
Patchup 18n errors

CharString:


src/openforms/emails/tests/test_tasks.py
```diff
@@ -15,6 +15,7 @@
 from ..tasks import send_email_digest
 
 
+@override_settings(LANGUAGE_CODE="en")
```
sergei-maertens, line 18: :(

I explicitly mentioned this on Slack as not being a sustainable approach. I'm only merging this now to not further block the release.

CharString:
I agree, it's how I bumped into that documentation fix. I was looking for a better way.

-------- Original Message --------

PR #3590, opened 2023-11-09
Set up automatic CSP form-action directive generation

sergei-maertens:
This is so the configuration is run automatically on existing instances, otherwise the lack of form-action directive values combined with our new setting in 2.4 would prevent DigiD/eHerkenning login or payment, unless the admins manually re-save the configuration the admin.

This is not manageable if you have many instances - previously we thought not doing this would not cause breaking changes and therefore no automatic migration was set up, but after discussion yesterday we realized things *do* break if you don't manage this.


src/openforms/config/tests/test_csp_update.py
```diff
@@ -27,3 +41,87 @@ def test_middleware_applies_cspsetting_models(self):
...
+        CSPSetting.objects.all().delete()  # delete creates from test data setup
+
+        call_command(
+            "create_csp_form_action_directives_from_config",
+            stdout=StringIO(),
+            stderr=StringIO(),
+        )
+
+        csp_settings = CSPSetting.objects.all()
+        self.assertEqual(len(csp_settings), 2)
```
vaszig, line 121: Is it better to use `len` instead of `CSPSetting.objects.count()`

src/openforms/config/tests/test_csp_update.py
```diff
@@ -27,3 +41,87 @@ def test_middleware_applies_cspsetting_models(self):
...
+
+        call_command(
+            "create_csp_form_action_directives_from_config",
+            stdout=StringIO(),
+            stderr=StringIO(),
+        )
+
+        self.assertFalse(CSPSetting.objects.exists())
+
+    def test_config_records_exists_but_are_incomplete(self):
```
vaszig, line 65: If there is something else to change let's fix this typo as well (exist)

src/openforms/config/tests/test_csp_update.py
```diff
@@ -27,3 +41,87 @@ def test_middleware_applies_cspsetting_models(self):
...
+
+
+class CreateCSPFormActionFromConfigTests(TestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.addCleanup(clear_caches)
+
+    def test_no_ogone_nor_digid_eherkenning_config(self):
+        assert not OgoneMerchant.objects.exists()
```
vaszig, line 53: Why don't we use `self.assertFalse` instead of `assert not`. This is a general question because I see it quite often in the tests.

vaszig (approved):

vaszig:
Now I saw the failing test concerning documentation..Please have a look at this as well

src/openforms/config/tests/test_csp_update.py
sergei-maertens: it depends - if you know you're going to evaluate the queryset anyway, then `len` is more efficient because a separate `.count()` performs an additional query

this version always does a single query, using `.count()` would do two queries.

src/openforms/config/tests/test_csp_update.py
sergei-maertens: so the `self.assertFOO` is to really do the test assertion - it makes it clear that this is the part that you are testing

the `assert` at the beginning of the test is more of a sanity check - did I set up my test data correctly? it makes certain implicit behaviours explicit. Here the assumption is that no records exists at all at the beginning of the test, such that calling the management command on empty data does not produce anything.

I actually had these assertions fail locally because of broken test isolation with other tests, and it helps in tracking down test failures that you can otherwise not explain easily.

So to summarize:

* assert at the beginning: sanity check for test data, not part of the test itself
* self.assertFOO: the actual meaningful part of the test

PR #3591, opened 2023-11-09
🐛 Add the ability to set Graph service to `NULL` from admin form

### LINKED ISSUE #3548: Unable to remove Graph service/registration if there is just 1 service

2.2.0 - 2.4.0

The Microsoft Graph registration, once set, cannot be unset which results in a loop where you cannot remove the service which is selected. Meaning there will always be a Graph service stuck in your configuration.
image

Viicos:
Fixes #3548.

As soon as a Microsoft Graph Service is set in the Microsoft Graph registration singleton model, it cannot be unset because `blank=False` currently.

To be able to delete Microsoft Graph Services, people would now need to unset it from the registration singleton model if linked, and then delete it.

---

This is also the case for a bunch of singleton models with a `ForeignKey`/`OneToOneField` to services, should we apply the same logic?


sergei-maertens (changes requested):
Please fix the CI errors

PR #3594, opened 2023-11-10
✨ [#3573] Add a new template tag `as_json`

### LINKED ISSUE #3573: Als gemeente wil ik geo-gegevens doorgeven aan ProductAanvraag...
..zodat deze in de Zaak kan eindigen.

Opnemen in template:
```
"geometry": {% submission.geo_json %}
```

Met als resultaat:

```
"geometry": {
  "type": "Point",
  "coordinates": [52.36673378967122, 4.893164274470299]
}
```

_Originally posted by @joeribekker in https://github.com/open-formulieren/open-forms/issues/2855#issuecomment-1595047926_

Viicos:
(Kind of) fixes #3573

### DIFF
--- a/docs/manual/forms/examples/form_with_geometry.rst
+++ b/docs/manual/forms/examples/form_with_geometry.rst
@@ -1,0 +1,95 @@
+.. _example_form_with_geometry:
+
+==============================================
+Form with geometry and registration attributes
+==============================================
+
+When adding a component to a form, it is possible to have the value filled in by
+the user set as a specific field on the selected registration backend.
+
+In this example, we will use the `Objects API`_ as a registration backend.
+
+.. note::
+
+    Depending on the selected backend, the supported registration attributes might differ.
+    In the case of `Objects API`_, only the *Location > Coordinate* attribute is supported.
+
+
+Creating the form
+=================
+
+1. Create an empty form and add a first step.
+2. Add a new map component: *Speciale velden > Kaart*.
+3. Under the **Registratie** tab, select the value *Location > Coordinate* in the **Registratie-attribuut** dropdown.
+4. Save the component, and go under the **Registration** tab of the current form.
+5. Add one of your configured Objects API backends (see :ref:`configuration_registration_objects` for more details on how to configure the backend).
+
+Registration process
+====================
+
+When a new submission for this form is created, it will be registered in the selected backend (in our case, Objects API).
+
+Internally, Open Forms will look into each component to see if a registration attribute is set. If this is the case,
+the submitted value for this component will be set to a fixed JSON path (this path differs depending on the registration backend).
+
+In the case of `Objects API`_, the service supports a geometry attribute set to ``record.geometry``. In our case,
+the JSON data sent to the Objects API will look like the following:
+
+.. code-block:: json
+
+    {
+        "type": "https://objecttype-example.nl/api/v2/objecttype/123",
+        "record": {
+            "typeVersion": "1",
+            "data": {},
+            "startAt": "2023-01-01",
+            "geometry": {"type": "Point", "coordinates": [1.0, 2.0]}
+        }
+    }
+
+Dealing with multiple geometry components
+=========================================
+
+When configuring registration attributes, having multiple components with the same attribute set will result in
+the last one overridding the others. With the Objects API registration, you can circumvent this issue by tweaking
+the JSON content template.
+
+.. seealso::
+
+    For more details on how to work with the JSON content template, see :ref:`objecten_api_registratie`.
+
+If we have two map components with the following Eigenschapnaam: ``map1`` and ``map2``; and ``map1`` is using the location registration attribute,
+we can use the following JSON content template to set ``map2`` to the custom data attribute ``map2_result``:
+
+.. code-block:: django
+
+    {
+        "data": {% json_summary %},
+        "type": "{{ productaanvraag_type }}",
+        "bsn": "{{ variables.auth_bsn }}",
+        "pdf_url": "{{ submission.pdf_url }}",
+         attachments": {% uploaded_attachment_urls %},
+        "submission_id": "{{ submission.kenmerk }}",
+        "language_code": "{{ submission.language_code }}",
+        "map2_result": : {% as_geo_json variables.map2 %}
+    }
+
+The resulting JSON data sent to the Objects API will look like:
+
+.. code-block:: json
+
+    {
+        "type": "<configured_objecttype>",
+        "record": {
+            "typeVersion": "<configured_version>",
+            "data": {
+                "data": "...",
+                "...": "...",
+                "map2_result": {"type": "Point", "coordinates": [3.0, 4.0]}
+            },
+            "startAt": "2023-01-01",
+            "geometry": {"type": "Point", "coordinates": [1.0, 2.0]}
+        }
+    }
+
+.. _`Objects API`: https://objects-and-objecttypes-api.readthedocs.io/
--- a/docs/manual/forms/examples/form_with_geometry.rst
+++ b/docs/manual/forms/examples/form_with_geometry.rst
@@ -1,39 +1,91 @@
 .. _example_form_with_geometry:
 
-==============================================
-Form with geometry and registration attributes
-==============================================
+================================================
+Kaartmateriaal en registratie in de Objecten API
+================================================
 
-When adding a component to a form, it is possible to have the value filled in by
-the user set as a specific field on the selected registration backend.
+In dit formulier maken we gebruik van de "Kaart"-component om coördinaten van een
+locatie uit te vragen en te registreren. Het formulier wordt gekoppeld met de
+`Objecten API`_ voor het registreren van de inzendingsgegevens.
 
-In this example, we will use the `Objects API`_ as a registration backend.
+We gaan er van uit dat de koppeling met de Objecten API
+:ref:`geconfigureerd <configuration_registration_objects>` is, en gebruiken als
+objecttype-URL ``https://objecttype-example.nl/api/v2/objecttype/123``.
 
-.. note::
+Formulier maken
+===============
 
-    Depending on the selected backend, the supported registration attributes might differ.
-    In the case of `Objects API`_, only the *Location > Coordinate* attribute is supported.
+We bouwen een formulier met twee kaarten - één kaart wordt gebruikt voor "dé" geometrie
+van het hele Object in de Objecten API en de andere kaart wordt als extra/vrije-vorm
+attribuut meegestuurd.
+
+#. Maak een formulier aan met de volgende gegevens:
+
+   * **Naam**: Kaartgegevens registreren in de Objecten API
+
+#. Klik op het tabblad **Stappen en velden**.
+#. Klik aan de linkerkant op **Stap toevoegen** en selecteer **Maak een nieuwe
+   formulierdefinitie**.
+#. Onder de sectie **(Herbruikbare) stapgegevens** vul het volgende in:
+
+   * **Naam**: Kaarten
+
+#. Scrol naar beneden en klik de sectie **Speciale velden** aan.
+#. Sleep een **Kaart** component op het witte vlak en vul de volgende
+   gegevens in:
+
+   * **Label**: Hoofdgeometrie
+
+#. Klik vervolgens de **Registratie** tab aan en vul de volgende gegevens in:
+
+   * **Registratie-attribuut**: Locatie > Coördinaten
+
+#. Druk daarna op **Opslaan**.
+
+#. Sleep een tweede **Kaart** component op het witte vlak, vul de volgende
+   gegevens in en druk daarna op **Opslaan**:
+
+   * **Label**: Extra geometrie
+
+#. Klik op het tabblad **Registratie**.
+#. Klik op **Registratiepluginoptie toevoegen** en vul de volgende gegevens in:
+
+    * **Naam**: Objecten API
+    * **Registratiemethode**: Objecten API registratie
+    * **Objecttype**: ``https://objecttype-example.nl/api/v2/objecttype/123``
+    * **JSON-inhoud sjabloon**:
+
+      .. code-block:: django
+
+          {
+              "extraGeometrie": {% as_geo_json variables.extraGeometrie %}
+          }
+
+#. Druk daarna op **Opslaan en opnieuw bewerken**. Het formulier kan nu ingevuld en
+   ingestuurd worden.
 
 
-Creating the form
-=================
+.. note:: Het inrichten van het Objecttype en de Objecten API is geen onderdeel van
+   dit voorbeeld, maar wel essentieel om zonder fouten formulierinzendingen te kunnen
+   verwerken.
 
-1. Create an empty form and add a first step.
-2. Add a new map component: *Speciale velden > Kaart*.
-3. Under the **Registratie** tab, select the value *Location > Coordinate* in the **Registratie-attribuut** dropdown.
-4. Save the component, and go under the **Registration** tab of the current form.
-5. Add one of your configured Objects API backends (see :ref:`configuration_registration_objects` for more details on how to configure the backend).
 
-Registration process
-====================
+Toelichting
+===========
 
-When a new submission for this form is created, it will be registered in the selected backend (in our case, Objects API).
+De eerste kaart is ingesteld via de "Registratie" tab op het component. Dit zorgt ervoor
+dat de Objecten API-registratieplugin dit veld gebruikt voor "de" geometrie van het object
+in de Objecten API. Deze kent hiervoor namelijk één vast attribuut.
 
-Internally, Open Forms will look into each component to see if a registration attribute is set. If this is the case,
-the submitted value for this component will be set to a fixed JSON path (this path differs depending on the registration backend).
+Echter, we kunnen extra geometrieën alsnog meesturen (als het objecttype dit modelleert),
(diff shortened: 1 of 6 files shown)


src/openforms/forms/models/form.py
```diff
@@ -64,7 +64,10 @@ class Form(models.Model):
         _("slug"), max_length=100, populate_from="name", editable=True, unique=True
     )
     product = models.ForeignKey(
-        "products.Product", null=True, blank=True, on_delete=models.CASCADE
+        "products.Product",
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
     )
```
Viicos, line 71: Black behaving weirdly

sergei-maertens: I'm going to guess that you have a different version than the one pinned in the requirements?

docs/manual/templates.rst
```diff
@@ -488,7 +488,8 @@ Variabele                              Beschrijving
 ``{{ submission.pdf_url }}``           De URL van het inzendingsrapport (in PDF formaat) in de documenten API.
 ``{{ submission.csv_url }}``           De URL van het inzendingsrapport (in CSV formaat) in de documenten API. Dit document is mogelijk niet aangemaakt
 ``{% json_summary %}``                 JSON met ``"<variabele-eigenschapsnaam>": "<waarde>"`` van alle formuliervelden.
-``{% uploaded_attachment_urls %}``     Een lijst met de URLs van documenten toegevoegd door de inzender. De URLs verwijzen naar het geregistreerde document in de Documenten API.
+``{% as_json "path.to.data.0" %}``     TODO in Dutch
```
sergei-maertens, line 491: this is not what we discussed, the argument to `as_json` should not be a string. It should just be a reference to the template variable:

```suggestion
``{% as_json path.to.data.0 %}``     Sluit de gerefereerde variabele (`path.to.data.0`) in als JSON.
```

docs/manual/templates.rst
```diff
@@ -506,7 +507,7 @@ Voorbeeld
            "type": "{{ productaanvraag_type }}",
            "bsn": "{{ variables.auth_bsn }}",
            "pdf_url": "{{ submission.pdf_url }}",
-           "attachments": {% uploaded_attachment_urls %},
+           "attachments": {% as_json" submission.uploaded_attachment_urls" %},
```
sergei-maertens, line 510:
```suggestion
           "attachments": {% as_json submission.uploaded_attachment_urls %},
```

src/openforms/registrations/contrib/objects_api/templates/registrations/contrib/objects_api/content_json.txt
```diff
@@ -8,7 +8,7 @@
   "type": "{{ tt_openvariable }} productaanvraag_type {{ tt_closevariable }}",
   "bsn": "{{ tt_openvariable }} variables.auth_bsn {{ tt_closevariable }}",
   "pdf_url": "{{ tt_openvariable }} submission.pdf_url {{ tt_closevariable }}",
-  "attachments": {{ tt_openblock }} uploaded_attachment_urls {{ tt_closeblock }},
+  "attachments": {{ tt_openblock }} as_json "submission.uploaded_attachment_urls" {{ tt_closeblock }},
```
sergei-maertens, line 11:
```suggestion
  "attachments": {{ tt_openblock }} as_json submission.uploaded_attachment_urls {{ tt_closeblock }},
```

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -1,27 +1,33 @@
...
+    warnings.warn(
+        "'uploaded_attachment_urls' is deprecated, use 'as_json' with a path argument instead.",
+        DeprecationWarning,
+    )
+    return as_json(context, "submission.uploaded_attachment_urls")
 
 
 @register.simple_tag(takes_context=True)
-def json_summary(context):
+def json_summary(context: template.Context) -> str:
```
sergei-maertens, line 30: type hint is wrong, it returns a `SafeString`, not just a bare `str`. And the if-branch on line 32 result in a dict being returned.

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -32,3 +38,22 @@ def json_summary(context):
...
+
+    Example usage:
+        {% as_json "variables.kaart" default=... %}  # Will output [52.3766727, 4.8877068]
+    """
+
+    # glom uses `getattr` by default which doesn't play well with `Context`:
+    glommer = Glommer()
+    glommer.register(template.Context, get=lambda t, name: t.get(name))
+
+    data = glommer.glom(context, glom_path, default=default)
```
sergei-maertens, line 57: there's no reason to use `glom` here, instead you can directly use Django's template mechanisms and it'll resolve the template variables during rendering.

sergei-maertens (changes requested):

src/openforms/forms/models/form.py
Viicos: Weird, still have the correct one (both styles are valid with black anyway).

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -11,17 +13,20 @@
...
-def uploaded_attachment_urls(context):
+def uploaded_attachment_urls(context: template.Context) -> SafeString:
     """
     Output a sequence of attachment URLs as a JSON-serialized list.
     """
-    # attachments is a list of URLs
+    warnings.warn(
+        "'uploaded_attachment_urls' is deprecated, use 'as_json' with a path argument instead.",
+        DeprecationWarning,
+    )
```
Viicos, line 23: Before this PR, `uploaded_attachmen_urls` tried to get attachment URLs in a safe way (`context.get("submission", {}).get("uploaded_attachment_urls", [])`).

This is not the case now with `as_json`? `as_json submission.uploaded_attachment_urls` will fail if `submission` isn't a `dict`.

sergei-maertens: Template tags are allowed to crash, template filters aren't, so accepted risk. We deliberately avoid exposing model instances in the user-controlled template evaluation since we don't want them to access the DB/call methods like that.

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -32,3 +37,10 @@ def json_summary(context):
         json_data = html_escape_json(json_data)
 
     return SafeString(json.dumps(json_data))
+
+
+@register.simple_tag
+def as_json(data: Any) -> SafeString:
```
sergei-maertens, line 43: Just for the record - this template tag is off-limits for our own templates/developers, it only exists to accommodate user-supplied templates. It's a footgun.

sergei-maertens (approved):

sergei-maertens (changes requested):
@Viicos this requires extensive updates to the documentation. Please draft it in English, and then a native Dutch speaker will translate it.

1. Document an example form on how to register the (primary) map component via the registration tab
2. Document in the same example how you could include additional map geometry when registering to the objects API.

Location in the docs: https://open-forms.readthedocs.io/en/stable/manual/forms/examples/index.html

Viicos:
Noticed while writing the example that `as_json` would not work exactly the same as the registration attribute, as we are using the following transform for the location reg. attr.:

```python
object_mapping = {
    "record.geometry": FieldConf(
        RegistrationAttribute.locatie_coordinaat, transform=_point_coordinate
    ),
}
```

`_point_coordinate` transforms `[lat, lng]` into a proper GeoJSON field

docs/manual/forms/examples/registration_attributes.rst
```diff
@@ -0,0 +1,95 @@
...
+
+    {
+        "type": "<configured_objecttype>",
+        "record": {
+            "typeVersion": "<configured_version>",
+            "data": "<rendered_configured_json_template>",
+            "startAt": "YYYY-MM-DD",
+            "geometry": "<The submitted value of the map component, in GeoJSON>"
+        }
+    }
```
sergei-maertens, line 48: Examples need to be concrete, don't use placeholders that the reader needs to fill out mentally. The target users for this documentation are not developers, but people building forms. "JSON" as a concept may already be too much.

docs/manual/forms/examples/registration_attributes.rst
```diff
@@ -0,0 +1,95 @@
+.. _example_registration_attributes:
+
+=================================
+Form with registration attributes
+=================================
```
sergei-maertens, line 5: Please do not make a generic "registration attributes" example, but a specific form-with-geometry/map example.

docs/manual/forms/examples/form_with_geometry.rst
```diff
@@ -0,0 +1,95 @@
...
+we can use the following JSON content template to set ``map2`` to the custom data attribute ``map2_result``:
+
+.. code-block:: django
+
+    {
+        "data": {% json_summary %},
+        "type": "{{ productaanvraag_type }}",
+        "bsn": "{{ variables.auth_bsn }}",
+        "pdf_url": "{{ submission.pdf_url }}",
+        attachments": {{ tt_openblock }} uploaded_attachment_urls {{ tt_closeblock }},
```
sergei-maertens, line 71:
```suggestion
        attachments": {% uploaded_attachment_urls %},
```

docs/manual/forms/examples/form_with_geometry.rst
```diff
@@ -0,0 +1,95 @@
+.. _example_form_with_geometry:
+
+==============================================
```
sergei-maertens, line 3: self-reminder: translate & clean up examples/instructions

docs/manual/templates.rst
```diff
@@ -488,6 +492,7 @@ Variabele                              Beschrijving
 ``{{ submission.pdf_url }}``           De URL van het inzendingsrapport (in PDF formaat) in de documenten API.
 ``{{ submission.csv_url }}``           De URL van het inzendingsrapport (in CSV formaat) in de documenten API. Dit document is mogelijk niet aangemaakt
 ``{% json_summary %}``                 JSON met ``"<variabele-eigenschapsnaam>": "<waarde>"`` van alle formuliervelden.
+``{% as_geo_json variables.map %}``    Sluit de gerefereerde variabele (`variables.map`) in als JSON.
```
sergei-maertens, line 495: can put this one as last element in the table? attachments is far more common than map components/geo data

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -32,3 +32,14 @@ def json_summary(context):
         json_data = html_escape_json(json_data)
 
     return SafeString(json.dumps(json_data))
+
+
+@register.simple_tag
+def as_geo_json(value: list[float]) -> SafeString:
+    """Output the ``value`` as a safe GeoJSON dumped string.
+
+    As of today, this only supports coordinates.
```
sergei-maertens, line 41: Please cross reference the `_transform_coordinates` helper - they go hand in hand.

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -1,4 +1,5 @@
 import json
+from typing import Any
```
sergei-maertens, line 2: this is an unused import, right?

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -119,3 +119,17 @@ def test_render_json_summary_no_submission(self):
         )
 
         self.assertEqual(rendered, "{}")
+
+
+class AsJsonTests(TestCase):
+    def test_render_as_geo_json(self):
```
sergei-maertens, line 125: and what happens when the map component is optional and left blank? I think the default/empty value is an empty string. That should not crash the registration backend.

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/tests/test_templatetags.py
```diff
@@ -119,3 +119,29 @@ def test_render_json_summary_no_submission(self):
...
+        )
+
+        expected = '{"type": "Point", "coordinates": [1.0, 2.0]}'
+
+        self.assertEqual(rendered, expected)
+
+    def test_render_as_geo_json_no_value(self):
+        rendered = render_from_string(
+            "{% as_geo_json variables.0.value %}",
+            context={"variables": [{"value": ""}]},
```
sergei-maertens, line 140: I've checked on our test environment and no value currently indeed presents itself as an empty string.

Which is weird, `null` would make more sense but that's the scope for a different ticket... :thinking:

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -32,3 +31,17 @@ def json_summary(context):
...
+
+@register.simple_tag
+def as_geo_json(value: list[float] | str) -> SafeString | dict:
+    """Output the ``value`` as a safe GeoJSON dumped string.
+
+    As of today, this only supports coordinates. This essentially does the same thing as
+    :func:`_transform_coordinates`, but for any map component.
+    """
+    if not value:
+        return {}
```
sergei-maertens, line 44: shouldn't this be `SafeString(json.dumps({}))`?

Now the string representation of an empty dict "happens" to be the same as an empty object in JSON, but that's a coincidence, and that would make the tag return a consistent type instead of `SafeString | dict`

Viicos: Was wondering the same, and went the same way as `json_summary` in the end. I'll switch for both

sergei-maertens:
@CharString I've assigned you to look at the documentation changes :)

docs/manual/forms/examples/form_with_geometry.rst
```diff
@@ -0,0 +1,115 @@
...
+
+    * **Naam**: Objecten API
+    * **Registratiemethode**: Objecten API registratie
+    * **Objecttype**: ``https://objecttype-example.nl/api/v2/objecttype/123``
+    * **JSON-inhoud sjabloon**:
+
+      .. code-block:: django
+
+          {
+              "extraGeometrie": : {% as_geo_json variables.extraGeometrie %}
```
CharString, line 61: Is the second `:` correct?
Also: why a template tag and not a filter? To support other projections?

docs/manual/forms/examples/form_with_geometry.rst
```diff
@@ -0,0 +1,115 @@
...
+                }
+            },
+            "startAt": "2023-01-01",
+            "geometry": {
+                "type": "Point",
+                "coordinates": [51.9225, 4.47917]
+            }
+        }
+    }
+
```
CharString, line 109: Ah nvm the filter vs tag remark.

docs/manual/forms/examples/form_with_geometry.rst
```diff
@@ -0,0 +1,115 @@
...
+.. code-block:: json
+
+    {
+        "type": "https://objecttype-example.nl/api/v2/objecttype/123",
+        "record": {
+            "typeVersion": "1",
+            "data": {
+                "extraGeometrie": {
+                    "type": "Point",
+                    "coordinates": [52.37403, 52.37403 4.88969]
```
CharString, line 99:
```suggestion
                    "coordinates": [52.37403, 4.88969]
```
?

CharString (changes requested):
@sergei-maertens

docs/manual/forms/examples/form_with_geometry.rst
sergei-maertens: it is not

there was some back and forth with a generic `{% as_json %}` tag which turned out to be not desired after all, because we want to apply geojson wrapping to the coordinates array.

I don't think that it was a conscious decision to not use a filter, but the pattern of custom tags is followed and I think I like that consistency. Filters are also a bit harder to understand/build if you're not familiar with the django template language, so for simplicity's sake this is probably better.

I do think that tags have a compile/validate advantage over filters which are expected to crash silently, while tags are expected to blow up spectacularly.

docs/manual/forms/examples/form_with_geometry.rst
sergei-maertens: all this typing for nothing :(

But yeah, right now we only support `Point`, it will change in the future where any GeoJSON can be specified and then we'll also have to change our frontend to send GeoJSON directly into the API instead of just naive coordinates now.

docs/manual/forms/examples/form_with_geometry.rst
sergei-maertens: I swear I did proof read this :black_cat:

sergei-maertens (approved):

PR #3603, opened 2023-11-13
Refactor celery tasks orchestration

### LINKED ISSUE #3005: As form designer I want to require payment before the form is submitted/registered
### Thema / Theme

Frontend

Although the original use case for payments was that payments are not required before submitting (ie. payment can also be done later or at the service desk), some organisations want to require payment before the submission is registered, to prevent an extra workflow/check to see if payment was made - which takes time to process.

For this, we need to restructure the post-submission flow by doing the payment beforehand. Basically saying a succesful payment means the form is successfully submitted. Keep in mind, the original flow should still be possible.

Before splitting this in tasks, I see overlap between this and co-signing (#1530) where we wait until the submission is co-signed before pushing it towards registration. The same can be done here. You can complete the submission (in open forms) but its not pushed to registration until payment is received.

In both the new co sign and new payment flow, there is a "final confirmation page". This page can show the zaaknumber and the final PDF (where it says paid and/or co-signed). Maybe we should offer a concept PDF before co-sign/payment is done to keep the flow clean but not sure about this.

_No response_

_No response_

SilviaAmAm:
Fixes #3005 (partially)


src/openforms/emails/confirmation_emails.py
```diff
@@ -20,9 +18,6 @@
 
 
 def get_confirmation_email_templates(submission: "Submission") -> tuple[str, str]:
-    if not submission.form.send_confirmation_email:
-        raise SkipConfirmationEmail("Confirmation e-mail sending is disabled.")
-
```
SilviaAmAm, line 25: Moved this check to the task that schedules the emails

src/openforms/emails/templates/emails/templatetags/cosign_information.html
```diff
@@ -3,7 +3,7 @@
 {% if cosign_complete or waiting_on_cosign %}<h2 style="margin: 1em 0; font-size: 1.5em;">{% trans "Co-sign information" %}</h2>{% endif %}
 {% if cosign_complete %}
     <p>
-        {% blocktranslate trimmed %}This email is a confirmation that this form has been co-signed by {{ cosigner_email }} and can now be processed.{% endblocktranslate %}
```
SilviaAmAm, line 6: The email could be triggered by the payment being completed, so it is not necessarily a confirmation of co-sign

src/openforms/payments/services.py
```diff
@@ -11,17 +10,6 @@
...
-    should_skip = any(
-        (
-            submission.registration_status != RegistrationStatuses.success,
-            not submission.payment_required,
-            submission.payment_registered,
-        )
-    )
-    if should_skip:
-        logevent.registration_payment_update_skip(submission)
-        return
```
SilviaAmAm, line 23: Moved this check to the task

src/openforms/submissions/tasks/__init__.py
```diff
@@ -22,128 +21,59 @@
...
 
-    retry_chain = chain(
-        pre_register_submission_task,
-        register_submission_task,
-        update_payments_task,
-        finalize_completion_retry_task,
+    # NOTE - this is "risky" since we're running outside of the transaction (this code
+    # should run in transaction.on_commit)!
+    PostCompletionTasksIDs.objects.create(
+        tasks_ids=task_ids, submission=Submission.objects.get(id=submission_id)
```
SilviaAmAm, line 74: We can't do the same as we were doing before:
```python
Submission.objects.filter(id=submission_id).update(on_completion_task_ids=task_ids)
```
Since multiple events can trigger this chain of tasks. This would lead to race conditions where the tasks IDs are overwritten

src/openforms/submissions/tasks/emails.py
```diff
@@ -77,9 +93,6 @@ def send_confirmation_email(submission_id: int) -> None:
         )
         return
 
-    if submission.confirmation_email_sent:
```
SilviaAmAm, line 80: Check moved to the schedule_emails task

src/openforms/submissions/tests/test_on_completion_chain.py
```diff
@@ -1,173 +0,0 @@
...
-        )  # 6 tasks in the chain
-        # registration result reference
-        self.assertTrue(submission.public_registration_reference.startswith("OF-"))
-        self.assertTrue(SubmissionReport.objects.filter(submission=submission).exists())
-        self.assertFalse(TemporaryFileUpload.objects.exists())
-        self.assertEqual(
-            len(mail.outbox), 2
-        )  # registration backend + confirmation email
-
-    def test_submission_form_with_incomplete_appointment(self):
```
SilviaAmAm, line 65: Moved to "test_post_submission_event"

src/openforms/submissions/tests/test_on_completion_retry_chain.py
```diff
@@ -235,6 +232,11 @@ def test_backend_registration_succeeds(self, mock_update_payment):
             form__registration_backend_options={"zgw_api_group": zgw_group.pk},
             # registration failed, so an internal reference was created
             public_registration_reference="OF-1234",
+            form__product__price=10,
+            form__payment_backend="demo",
+        )
+        SubmissionPaymentFactory.create(
+            submission=submission, amount=10, status=PaymentStatus.completed
```
SilviaAmAm, line 239: This needed to be added because on line 263 we check if the update payment task is called (but now this is skipped if no payment is required)

src/openforms/submissions/tests/test_post_submission_event.py
SilviaAmAm, line 1: This is testing that the expected tasks are skipped/called on the events (with the various combinations of payment/cosign required/not required):
- Submission completion
- Cosign
- Payment complete
- Retry flow
- Payment status update retry flow

It also tests that the right emails are sent and that the confirmation email contains the expected information about cosign/payment.

The tests in src/openforms/submissions/tests/test_on_completion_retry_chain.py test that the `needs_on_completion_retry` is being set properly.

src/openforms/submissions/tests/test_task_co_sign.py
```diff
@@ -145,34 +145,3 @@ def test_form_link_not_allowed_in_email(self, mock_get_solo):
...
-        )
-
-        on_cosign(submission.id)
-
-        self.assertEqual(len(mail.outbox), 1)
-
-        email = mail.outbox[0]
-
-        self.assertEqual(email.recipients(), ["main@test.nl", "cosign@test.nl"])
-        self.assertEqual(email.cc, ["cosign@test.nl"])
```
SilviaAmAm, line 178: Tested in src/openforms/submissions/tests/test_post_submission_event.py

src/openforms/submissions/utils.py
```diff
@@ -96,21 +95,10 @@ def remove_submission_uploads_from_session(
...
-        )
-    except SkipConfirmationEmail:
-        logger.debug(
-            "Form %d is configured to not send a confirmation email for submission %d, "
-            "skipping the confirmation e-mail.",
-            submission.form.id,
-            submission.id,
-        )
-        logevent.confirmation_email_skip(submission)
-        return
```
SilviaAmAm, line 113: Moved to the `schedule_email` task

src/openforms/submissions/tests/test_on_completion_chain.py
```diff
@@ -1,173 +0,0 @@
...
-        )
-
-        with patch(
-            "openforms.registrations.contrib.email.plugin.EmailRegistration.register_submission"
-        ) as mock_registration:
-            on_completion(submission.id)
-
-        mock_registration.assert_not_called()
-
-    def test_cosign_not_required_and_not_filled_in_proceeds_with_registration(self):
```
SilviaAmAm, line 104: Moved to "test_post_submission_event"

src/openforms/submissions/tests/test_on_completion_chain.py
```diff
@@ -1,173 +0,0 @@
...
-        )
-
-        with patch(
-            "openforms.registrations.contrib.email.plugin.EmailRegistration.register_submission"
-        ) as mock_registration:
-            on_completion(submission.id)
-
-        mock_registration.assert_called_once()
-
-    def test_cosign_not_required_but_filled_in_does_not_proceed_with_registration(self):
```
SilviaAmAm, line 151: Moved to "test_post_submission_event"

src/openforms/payments/views.py
```diff
@@ -235,9 +235,7 @@ def _handle_return(self, request, uuid: str):
                 )
                 raise ParseError(detail="redirect not allowed")
 
-        transaction.on_commit(
-            lambda: update_submission_payment_status.delay(payment.submission.id)
-        )
+        transaction.on_commit(lambda: on_post_submission_event(payment.submission.pk))
```
Viicos, line 238: Small (and not that important remark): it is recommended to use `partial` over lambdas to avoid late binding. But I think here this is fine (`payment.submission.pk` isn't going to change)

src/openforms/submissions/tasks/__init__.py
Viicos: Was this the reason `on_completion_task_ids` was moved to a separate model?

SilviaAmAm:
@Viicos 

> Was this the reason on_completion_task_ids was moved to a separate model?

Yes! Before the sequence of tasks in `on_completion` was only called when the submission was completed (and so we could save the tasks ids on the submission field). But now, the `on_post_submission_event` is called in multiple places, so if it triggered twice at the same time, then you risk that one call overwrites the task ids set by the other call. This is why I decided to go for a separate model.

SilviaAmAm:
I tried using the [select_for_update](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#select-for-update) to lock the row. However, it needs to run in a transaction and the `on_post_submission_event` runs when the transaction is committed so it is no longer in the transaction :S

src/openforms/submissions/models/submission.py
```diff
@@ -782,3 +788,13 @@ def registration_backend(self) -> FormRegistrationBackend | None:
...
             else None
         )
+
+    @property
+    def on_completion_task_ids(self) -> list[str]:
+        """Get the IDs of the tasks triggered when the submission was completed
+
+        Other tasks are triggered later once the submission is cosigned, a payment is made or a retry flow is triggered.
+        But to give feedback to the user we only need the first set of triggered tasks.
+        """
```
sergei-maertens, line 798: I'm not sure I like the `.first()` in here based on PK ordering - is there a chance we can let the chain scheduling function return the appropriate instance that contains the task IDs and use that in the `_complete` endpoint instead? That would make it deterministic since you only ever consider the record that was created in the same (HTTP) transaction.

*... some time passes ...*

I've reviewed the rest of the code and this is not viable because the record is only completed after the transaction commits, so there's no way to pass this record back and include it in the `SubmissionStatus` thing.

Perhaps we can pass a flag `from_complete_endpoint` to the `SubmissionTaskMetadata` instance, and add a unique together condition based on this flag so that we can uniquely identify the tasks ID set to use for the `/api/v2/submissions/:uuid/status` endpoint?

src/openforms/submissions/status.py
```diff
@@ -23,11 +23,31 @@ class SubmissionProcessingStatus:
...
+        """
+        if not hasattr(self, "_all_async_results"):
+            task_ids_per_event = (
+                self.submission.postcompletiontasksids_set.all().values_list(
+                    "tasks_ids", flat=True
+                )
+            )
+            task_ids = []
+            for tasks_ids_list in task_ids_per_event:
+                task_ids += tasks_ids_list
```
sergei-maertens, line 46: I wonder if you could do an aggregation query at the DB level instead? Something like ` self.submission.postcompletionmetadata_set.aggregate(models.Sum("tasks_ids"))` (untested of course)

https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/aggregates/#django.contrib.postgres.aggregates.ArrayAgg might be viable

src/openforms/submissions/tasks/__init__.py
```diff
@@ -22,127 +21,50 @@
 logger = logging.getLogger(__name__)
 
 
-def on_completion(submission_id: int) -> None:
-    """
-    Celery chain of tasks to execute on a submission completion event.
+def on_post_submission_event(submission_id: int) -> None:
+    # this can run any time because they have been claimed earlier
```
sergei-maertens, line 25: Can you restore the docstring as much as possible?

The part about `transaction.on_commit` is quite relevant still.

src/openforms/submissions/tasks/__init__.py
```diff
@@ -22,127 +21,50 @@
...
+    cleanup_temporary_files_for.delay(submission_id)
 
-    This SHOULD be invoked as a transaction.on_commit(...) handler, therefore it should
-    not execute any extra queries in the process this function is running in.
-    """
-    # use immutable signatures so that the result of previous tasks is not passed
-    # in as an argument to chained tasks
+    # If the form involves appointments and no appointment has been scheduled yet, schedule it.
+    # Todo: deprecated => Not needed with the new appointment flow
     register_appointment_task = maybe_register_appointment.si(submission_id)
```
sergei-maertens, line 30: You can add a `warnings.warn("...", PendingDeprecationWarning)` in the `maybe_register_appointment` task too if you want.

src/openforms/submissions/tasks/__init__.py
```diff
@@ -22,127 +21,50 @@
...
-    # representation of the task orchestration and dependencies (TODO, @joeri)
-    # The linked task (= next task) is only executed if the previous task returns
-    # successfully, so error handling needs to happen inside each task.
-    on_completion_chain = chain(
-        # The appointment must be registered before a confirmation PDF is generated and
-        # any backend registration happens, as on-failure, the user should get feedback
-        # about the failure.
+
+    # Payment status update
+    # Todo: deprecated => Not needed with the new payment flow
```
sergei-maertens, line 43: I was under the impression that both flows are equally viable - payment prevents registration or payment status update afterwards, so should we actually deprecate this? If there's a "mode select" in the global configuration or at the form level, deprecating it might be a bit too eager.

src/openforms/submissions/tasks/__init__.py
sergei-maertens: Yep

src/openforms/submissions/tasks/cleanup.py
```diff
@@ -46,12 +46,18 @@ def cleanup_on_completion_results():
...
 
-    submissions = Submission.objects.filter(
-        # only consider submissions that are completed and old enough
-        completed_on__lte=cutoff,
-        suspended_on__isnull=True,
-        # only clean up submissions that have task_ids stored (prevent re-processing)
-        on_completion_task_ids__len__gt=0,
+    submissions = (
+        Submission.objects.annotate(
+            number_tasks_ids=F("postcompletiontasksids__tasks_ids__len")
```
sergei-maertens, line 51: how does this `__len` work if there are multiple records for the same submission? Can `__len` result in `None` for some records?

src/openforms/submissions/tasks/cleanup.py
```diff
@@ -46,12 +46,18 @@ def cleanup_on_completion_results():
...
+        )
+        .filter(
+            # only consider submissions that are completed and old enough
+            completed_on__lte=cutoff,
+            suspended_on__isnull=True,
+            # only clean up submissions that have task_ids stored (prevent re-processing)
+            number_tasks_ids__gt=0,
```
sergei-maertens, line 58: I think this related query could be done directly inside the filter without annotate, but I'm not sure.

```suggestion
            # only clean up submissions that have task_ids stored (prevent re-processing)
            postcompletiontasksids__tasks_ids__len__gt=0,
```

src/openforms/submissions/tasks/cleanup.py
```diff
@@ -60,24 +66,7 @@ def cleanup_on_completion_results():
...
-    This task must be called last and mark the submission as not requiring any retries
-    anymore. Tasks being retried that fail again should result in a celery failure
-    to prevent this finalizer from running.
-
-    All other internal state will have been updated by the specific (sub)tasks.
-    """
-    submission = Submission.objects.get(id=submission_id)
-    submission.needs_on_completion_retry = False
-    submission.save(update_fields=["needs_on_completion_retry"])
+            submission.postcompletiontasksids_set.all().delete()
```
sergei-maertens, line 69: I was concerned that we lose debug information, but apparently we already do in the current versions anyway. Since this considers only stuff that went right, it's probably okay to delete this data.

src/openforms/submissions/tasks/emails.py
```diff
@@ -77,14 +63,109 @@ def send_confirmation_email(submission_id: int) -> None:
...
+            submission.id,
+        )
+        logevent.confirmation_email_skip(submission)
+        return
+
+    execution_options = {}
+    if not submission.confirmation_email_sent:
+        # The confirmation email is sent after the user has submitted a form. It contains the generated PDF report
+        # with information about whether payment/cosign is (still) needed or not. This is scheduled with a 15 countdown
+        # if a payment is needed
```
sergei-maertens, line 150:
```suggestion
        # with information about whether payment/cosign is (still) needed or not. This is scheduled with a (15 min) countdown
        # if a payment is needed
```

src/openforms/submissions/tasks/emails.py
```diff
@@ -77,14 +63,109 @@ def send_confirmation_email(submission_id: int) -> None:
...
+    # completed, we are in a registration-retry flow or we are in a payment-status-update-retry flow. The last 2
+    # cases don't need to send a confirmation email.
+    # Edge case: cosign and payment happen at the same time
+    if (
+        not submission.cosign_confirmation_email_sent and submission.cosign_complete
+    ) or (
+        not submission.payment_complete_confirmation_email_sent
+        and submission.payment_user_has_paid
+    ):
+        send_confirmation_email.delay(submission_id)
```
sergei-maertens, line 171: does this make sense? You can only end up in this block with the condition `if submission.confirmation_email_sent` (the opposite of L147).

This would lead to the confirmation email being sent more than once, no?

src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -0,0 +1,930 @@
...
+
+        with patch(
+            "openforms.registrations.contrib.email.plugin.EmailRegistration.register_submission"
+        ) as mock_registration, patch(
+            "openforms.submissions.public_references.get_reference_for_submission",
+            return_value="OF-TEST!",
+        ), patch(
+            "openforms.registrations.contrib.email.plugin.EmailRegistration.update_payment_status"
+        ) as mock_payment_status_update:
+            on_post_submission_event(submission.id)
```
sergei-maertens, line 62: Trick from Chris to force formatting of context managers:

```py
with (
  cm1() as cv1,
  cm2() as cv2,
  ...,
):
```

leads to more readable code

src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -0,0 +1,930 @@
...
+            confirmation_email_sent=True,
+            form__registration_backend="email",
+            form__registration_backend_options={"to_emails": ["test@registration.nl"]},
+            form__name="Pretty Form",
+            auth_info__attribute=AuthAttribute.bsn,
+            auth_info__value="111222333",
+            form__product__price=10,
+            form__payment_backend="demo",
+            needs_on_completion_retry=True,
+            registration_status=RegistrationStatuses.failed,
```
sergei-maertens, line 711: is there a reason you don't use the traits like `registration_failed=True`? They generally help enforce business logic state, like ensuring that `completed=True` and a `last_register_date` is set.

src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -0,0 +1,930 @@
...
+        )
+        SubmissionPaymentFactory.create(
+            submission=submission, amount=10, status=PaymentStatus.completed
+        )
```
sergei-maertens, line 715: this could be `with_completed_payment` trait, which ensures that `amount` and `submission__form__product__price` are in sync.

src/openforms/submissions/tests/test_submission_completion.py
```diff
@@ -81,7 +81,7 @@ def test_all_required_steps_validated(self):
             ],
         )
 
-    @patch("openforms.submissions.api.mixins.on_completion")
+    @patch("openforms.submissions.api.mixins.on_post_submission_event")
     @freeze_time("2020-12-11T10:53:19+01:00")
     def test_complete_submission(self, mock_on_completion):
```
sergei-maertens, line 86: rename `mock_on_completion`?

src/openforms/submissions/tests/test_submission_status.py
```diff
@@ -21,13 +21,18 @@
...
+    SubmissionReportFactory,
+)
 
 
 class SubmissionStatusPermissionTests(APITestCase):
     def test_valid_token(self):
         # Use empty task ID to not need a real broker
-        submission = SubmissionFactory.create(completed=True, on_completion_task_ids=[])
+        submission = SubmissionFactory.create(completed=True)
+        PostCompletionTasksIDsFactory.create(submission=submission, tasks_ids=[])
```
sergei-maertens, line 35: I'd consider putting the related model creation inside the `completed` trait too.

src/openforms/appointments/tests/test_confirmation_emails.py
```diff
@@ -127,7 +128,7 @@ def test_send_confirmation_mail_disabled(self):
             contact_details={"lastName": "Powers", "email": "austin@powers.net"},
         )
 
-        send_confirmation_email(appointment.submission)
+        schedule_emails(appointment.submission.id)
```
sergei-maertens, line 131: the behaviour has changed I believe, the previous test call would actually result in mails being scheduled if the logic was wrong, but now even if the logic is wrong I believe the test/assertion will pass because we don't send mail synchronously inside `schedule_emails`. For test robustiness/correctness, I think this needs wrapping in `CELERY_TASK_ALWAYS_EAGER=True` to for synchronous processing *or* you need to assert that none of the mail sending functions are called/scheduled.

sergei-maertens (changes requested):
Chonky PR gets a chonky review :grimacing:

src/openforms/payments/tasks.py
```diff
@@ -34,6 +48,7 @@ def update_submission_payment_status(submission_id: int):
         if is_retrying:
             raise
         return
-
-    # if this properly executed, we can schedule "instant" confirmation e-mail delivery
-    send_confirmation_email.delay(submission_id)
+    else:
+        if is_retrying:
+            submission.needs_on_completion_retry = False
+            submission.save(update_fields=["needs_on_completion_retry"])
```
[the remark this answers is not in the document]
SilviaAmAm: Initially I thought I had done it because I couldn't coordinate the registration task and the payment status update task to set it right :thinking: I have tried again to move it in the finalisation task and it seems to work :thinking:

src/openforms/submissions/migrations/0004_auto_20231117_1430.py
```diff
@@ -0,0 +1,86 @@
...
+
+    dependencies = [
+        ("submissions", "0003_cleanup_urls"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="submission",
+            name="on_completion_task_ids",
+        ),
```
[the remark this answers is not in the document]
SilviaAmAm: Ahh yes, I didn't think about the case where a submission is being processed during the update! I will deprecate the field.

src/openforms/submissions/models/submission.py
SilviaAmAm: I added a flag saying which event triggered the chain. The submission is completed only once, so we can filter on the on_completion event

src/openforms/submissions/status.py
SilviaAmAm: I had tried and miserably failed. I will try again :smile:

src/openforms/submissions/tasks/__init__.py
SilviaAmAm: Ahh I had understood that we eventually wanted to remove it. I will remove the comment.

src/openforms/submissions/tasks/cleanup.py
SilviaAmAm: There is this test in src/openforms/submissions/tests/test_submission_status.py which tests the scenario of multiple records:
```python
def test_cleanup_for_submission_with_multiple_post_completion_events(
        self, mock_forget
    ):
            completed=True,
            completed_on=timezone.now() - timedelta(days=2, seconds=10),
            suspended_on=None,
        )
        PostCompletionMetadataFactory.create(
            submission=submission, tasks_ids=["some-id-1", "some-id-2"]
        )
        PostCompletionMetadataFactory.create(
            submission=submission, tasks_ids=["some-id-3"]
        )
        PostCompletionMetadataFactory.create(submission=submission, tasks_ids=[])

        cleanup_on_completion_results()

        self.assertEqual(3, mock_forget.call_count)
```
I will add a test for the case without record!

src/openforms/submissions/tasks/cleanup.py
SilviaAmAm: Yes, I added it because of the line `submission.on_completion_task_ids = []` which was present before the refactor

src/openforms/submissions/tasks/emails.py
SilviaAmAm: My reasoning was that on first completion, the confirmation email is scheduled (with or without countdown).
Then, on payment or on cosign event, a new confirmation email is sent (with or without the cosigner in CC depending on the event).

src/openforms/registrations/tests/test_registration_hook.py
```diff
@@ -34,6 +35,7 @@ class ResultSerializer(serializers.Serializer):
     external_id = serializers.UUIDField()
 
 
+@override_settings(CELERY_TASK_ALWAYS_EAGER=True)
```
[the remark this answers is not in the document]
SilviaAmAm: I went for this because of the "retry flag" check. Since this happens at the end of the chain, if I only call the `register_submission` task then it doesn't get set :thinking: but I am a bit unhappy with how the flag is being set now, so maybe I need to rethink that :S

src/openforms/registrations/tasks.py
```diff
@@ -162,31 +192,40 @@ def register_submission(submission_id: int) -> None:
 
     if not plugin.is_enabled:
         logger.debug("Plugin '%s' is not enabled", backend)
-        e = RegistrationFailed("Registration plugin is not enabled")
+        exc = RegistrationFailed("Registration plugin is not enabled")
         submission.save_registration_status(
-            RegistrationStatuses.failed, {"traceback": str(e)}
+            RegistrationStatuses.failed, {"traceback": str(exc)}
```
sergei-maertens, line 197: probably should also be refactored to `traceback.format_exc()`? Though you'd have to use `traceback.format_exception(exc)` to be able to pass in the exception instance (and then you have to `"".join(...)`) it

src/openforms/registrations/tasks.py
```diff
@@ -162,31 +192,40 @@ def register_submission(submission_id: int) -> None:
...
         )
-        logevent.registration_failure(submission, e)
```
sergei-maertens, line 169: I think this was dropped unintentionally?

src/openforms/registrations/tasks.py
```diff
@@ -162,31 +192,40 @@ def register_submission(submission_id: int) -> None:
...
     try:
         options_serializer.is_valid(raise_exception=True)
-    except Exception as e:
+    except ValidationError as exc:
+        logevent.registration_failure(submission, exc, plugin)
+        logger.warning(
+            "Registration using plugin '%r' for submission '%s' failed",
+            plugin,
+            submission,
+        )
```
sergei-maertens, line 214:
```suggestion
        logger.warning(
            "Registration using plugin '%r' for submission '%s' failed",
            plugin,
            submission,
            exc_info=exc,
        )
```
when an exception is available, pass it along so we can see more details in Sentry :)

src/openforms/registrations/tests/test_pre_registration.py
```diff
@@ -2,16 +2,30 @@
...
+    def test_pre_registration_with_submission_not_completed(self):
+        submission = SubmissionFactory.create()
+
+        with self.assertLogs() as logs:
+            pre_registration(submission.id, PostSubmissionEvents.on_completion)
+
+        self.assertEqual(
+            logs.records[0].msg,
+            "Trying to pre-register submission '%s' which is not completed.",
+        )
```
sergei-maertens, line 27: just for my understanding - is it deliberate that the placeholders are not interpolated?

sergei-maertens (approved):

PR #3606, opened 2023-11-14
🚑️ [#3604] Add required headers for Objects API POST request

### LINKED ISSUE #3604: Objects API registration backend no longer works with OF 2.4.0, raises 412 HTTP

2.4.0

Sending in via the Objecten API no longer works with HTTP 412. Raised via Utrecht Taiga 146 and reproduced on the test environment:

https://open-forms.test.maykin.opengem.nl/admin/submissions/submission/2630/change/

{"traceback": "Traceback (most recent call last):\n  File \"/app/src/openforms/registrations/tasks.py\", line 185, in register_submission\n    result = plugin.register_submission(\n  File \"/app/src/openforms/registrations/contrib/objects_api/plugin.py\", line 200, in register_submission\n    response.raise_for_status()\n  File \"/usr/local/lib/python3.10/site-packages/requests/models.py\", line 1021, in raise_for_status\n    raise HTTPError(http_error_msg, response=self)\nrequests.exceptions.HTTPError: 412 Client Error: Precondition Failed for url: https://objecten.test.opengem.nl/api/v2/objects\n"}

This is most likely due to missing the Content-Crs header in the API client refactor, which is a required field when POSTing to the Objects API

" Value: "EPSG:4326"
The 'Coordinate Reference System' (CRS) of the request data. According to the GeoJSON spec, WGS84 is the default (EPSG: 4326 is the same as WGS84)."

https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/maykinmedia/objects-api/master/src/objects/api/v2/openapi.yaml#tag/objects/operation/object_create

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

Viicos:
Fixes #3604.

This is the easy solution, but isn't ideal as this could break again somewhere else if someones tries to do a post request and forget about the header.

Other solution:
- Create a new client class specific to the Objects API, that will add the header when required/supported (looking at the spec only unsafe requests require the header).


sergei-maertens (changes requested):
And the ZGW API's? The POST `/api/v1/zaken` endpoint also uses this header.

src/openforms/contrib/zgw/clients/zaken.py
```diff
@@ -56,7 +58,7 @@ def set_payment_status(self, zaak: dict, partial: bool = False):
             "betalingsindicatie": "gedeeltelijk" if partial else "geheel",
             "laatsteBetaaldatum": timezone.now().isoformat(),
         }
-        response = self.patch(url=zaak["url"], json=data)
+        response = self.patch(url=zaak["url"], json=data, headers=CRS_HEADERS)
```
Viicos, line 61: Not sure about this one, I don't know where `zaak["url"]` points to. If it matches `/zaken/{uuid}` then this is correct

sergei-maertens: yep, it does!

Issue #3593, opened 2023-11-10
Figure out a proper solution to consume NL DS design tokens in e-mail templates


sergei-maertens:
Spin off from #3575

Right now, we only support design token values set directly in our own custom JSON, but organizations should control their styling through their own JSON design tokens *somewhere*.

These typically get published as an NPM package, so it would be great if we could figure out which JSON file with token values to load, merge the tokens in our JSON field into that and then look up the values.

This removes duplication and makes it easier to refactor in these components.

sergei-maertens:
I did some poking around in the NL DS community. The NL DS "organization" encourages organizations to bootstrap their own theme/design token infrastructure with their template repository, which includes a style-dictionary configuration that outputs the resulting tokens in `.json` format. There are three forms for this:

* The Design Token (draft) spec format, so `{"foo": {"value": "3px"}}` with arbitrary levels of nesting (this could also be using the `$value` key rather than `value`). The value itself can be an unresolved reference, e.g. `"{other.token}"`
* The resolved tokens, which keeps the nesting of tokens, but leaf nodes are the values and there's no `$value` or `value` key present anymore. We *could* work with this if needed
* A flattened format which is a 1 dimensions key-value map, where the key is a camelCased joined string of the intermediate nodes. This format is too unpredictable to be usable

On the OF side, we have a JSON field where tokens can be specified too, and those use the `value` key - i.e. the're the first option with unresolved tokens.

The planned approach is thus to allow admins to configure (via URL or file upload) their design tokens source, and we feed this file + our own custom design tokens JSON field to style-dictionary. This results in consumable tokens (in JSON format) that we can process in Python to extract all the resolved values. The idea is that our JSON field values override possible duplicates and treat them as being "more specific".

If there's a notable performance impact, this result can be cached.

As long as the tokens adhere to the design token (draft) spec, we can process them with a fixed configuration.

sergei-maertens:
Discussed internally - we might break this open entirely and remove our local design_token_values `JSONField` in favour of doing NL DS patterns only - but that requires us to set up a centralized tool to build these themes for organizations without resources to manage NL DS repositories and vision. This is a massive task.

We're going to explore this potential through a hackaton, preferably in collaboration with NL DS themselves. Until that time, I'm parking/blocking this issue.

To prevent the breaking change where our old, custom tokens are removed as part of the data migration, I will disable the removal of remapped tokens in the data migration.

PR #3609, opened 2023-11-14
:adhesive_bandage: [#3593] Un-break mail styling with local design to…

sergei-maertens:
…ken overrides

Do not remove the refactored header/footer design tokens yet, as the mail styling makes use of these.

See the Github issue for more context on why we're not structurally fixing this yet, as we'd ideally like to use the dist/tokens.json of organizations in mail rendering too.


CharString (approved):
Maybe consider changing `OBSOLETE` to `DEPRECATED`, if they still have a use, they aren't obsolete. Semantics, but as is, this fixes what's broken.

sergei-maertens:
> Maybe consider changing `OBSOLETE` to `DEPRECATED`, if they still have a use, they aren't obsolete. Semantics, but as is, this fixes what's broken.

the intend is very much to make them obsolete, so if we revert this patch, obsolete is the correct term, but before we can do that, we need to figure out how to solve the email template design tokens consumption challenges.

PR #3610, opened 2023-11-14
Add typechecking to itemsExpression

### LINKED ISSUE #3596: Add typechecking to keuzeopties expressie
Add InfernoLogic typecheck to variable options, opties-expressie in

 - [ ] radio
 - [ ] selectievakjes
 - [ ] keuzelijst

As a start, expect Array of Array of String.

This catches malformed expression like the one in #2941  `{"map": [{"var": "iban"}]}` with `TypeError: expr is undefined`

Technically it would close #2941...

CharString:
Closes #3596

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@fortawesome/fontawesome-free": "^6.1.1",
     "@open-formulieren/design-tokens": "^0.48.0",
     "@open-formulieren/formio-builder": "^0.8.0",
+    "@open-formulieren/infernologic": "^0.1.0",
     "@rjsf/core": "^4.2.1",
     "@tinymce/tinymce-react": "^3.12.6",
     "@trivago/prettier-plugin-sort-imports": "^4.0.0",
--- a/src/openforms/forms/tests/e2e_tests/test_form_designer.py
+++ b/src/openforms/forms/tests/e2e_tests/test_form_designer.py
@@ -14,6 +14,7 @@ from openforms.tests.e2e.base import E2ETestCase, browser_page, create_superuser
 from openforms.utils.tests.cache import clear_caches
 from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
 
+from ...models import Form
 from ..factories import (
     FormDefinitionFactory,
     FormFactory,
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -69,7 +70,7 @@
     ],
     "transformIgnorePatterns": [
       "/node_modules/(?!@open-formulieren/formio-builder)/",
-      "\\.pnp\\.[^\\\/]+$"
+      "\\.pnp\\.[^\\/]+$"
```
sergei-maertens, line 73: uh this changes it from the pattern that's configured by default: https://jestjs.io/docs/configuration#transformignorepatterns-arraystring - I've no idea if this breaks something or not though.

CharString: npm :point_left: did it.

CharString: Had to turn off my formatter (jq) too, as it corrects it as well; "\/" is not a legal escape code in a string per se. Only inside a `/repattern/` literal it makes sense. So if I reason about it `/\.pnp\.[^\/]+$/` is probably the regex they're after, but this
```node
> JSON.stringify(/\.pnp\.[^\/]+$/.toString())
'"/\\\\.pnp\\\\.[^\\\\/]+$/"'
```
just made me happy about r"" strings in Python and revert the change.

Some people, when confronted with a problem think: "I know, I'll use regular expressions... and JavaScript." Now they have NaN problems.

No seriously:
```node
> /./ * 2
NaN
```

src/openforms/js/components/formio_builder/builder.js
```diff
@@ -17,6 +18,21 @@ const getBuilderOptions = () => {
...
+   *
+   * @param {object | array | number | string} logic - JsonLogic expression
+   * @param {object | array | number | string} expected - example expected result value [["label", "value"]]
+   * @returns {boolean | string} - OK or an error message
+   */
+  const validateLogic = (logic, expected) => {
+    // We don't evaluate logic, we just look at types.
+    // "===" forces the logic expression align with the expectancy
+    const result = infer({'===': [logic, expected]}, {});
+    return result.startsWith('result type:') || result;
```
CharString, line 33: `infer` API must change. This is ugly. Want to have more info from other tickets before making a decision and releasing a new npm package.

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -2076,3 +2077,71 @@ def setUpTestData():
...
+            editor = component.locator("css=.ace_content")
+            await editor.click()
+            await editor.press("Control+a")
+            await editor.press("Backspace")
+            await editor.type('{"map": [{"var": "somevar"}]}')
+            await page.get_by_role("button", name="Opslaan").click()
+            await expect(page.locator("css=.formio-dialog-content")).to_be_visible()
+            await expect(
+                page.locator('css=[data-component-key="openForms.itemsExpression"]')
+            ).to_contain_text("Error")
```
CharString, line 2126: Contents of "Error" will change. These assertions just assert that type check has occurred and accepts correct expressions.
Testing for the better error messages belongs in a unittest, E2E is *way* to slow and expensive for that.

package.json
sergei-maertens: I may also be wrong in how I copied it from the docs where a JS regex is expected, but all these levels of configuration make me :dizzy:

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -2076,3 +2077,71 @@ def setUpTestData():
...
+            await editor.click()
+            await editor.press("Control+a")
+            await editor.press("Backspace")
+            await editor.type(
+                '{"map": [{"var": "somevar"}, [{"var": ""}, {"cat": ["Optie ", {"var": ""}]}]]}'
+            )
```
SilviaAmAm, line 2132: How does it know that `{"var": ""}` in the context of `{"var": "somevar"}` (is this checked that it is of type array?) is a string?

CharString: I don't know the exact order, but it follows from 

 - `{"cat": ["Optie ", {"var": ""}]}`
 - the list of options is of `string[][]` / `list[list[str]]`, therefore the result of the `"map"` is also of that type.

It's why the `validateLogic` passed the expression into
```ts
const result = infer({'===': [logic, expected]}, {});
```
with `expected` = `[["", ""]]` and `logic` = (in this case) the `"map"`

It uses the fact that `"==="` is a `Callable[[T, T], boolean]`, so the parse of the result from the `infer` call is a lot easier. `infer({"===": [{"var": "a"}, {"var": "b"}]})` does *not* imply that `a` and `b` have the same value, just the same type.

`infer` shouldn't return a string, of course, but before changing it, I like to have more use cases.

sergei-maertens:
Converted it back to draft so we don't get distracted by this PR - if I understood it correctly from the daily, the work is now happening in https://github.com/open-formulieren/formio-builder

PR #3614, opened 2023-11-15
Add image scanning

joeribekker:
Closes open-formulieren/security-issues#10

Second attempt to add non-intrusive image scanning.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -329,6 +329,41 @@ jobs:
       target_env: ${{ matrix.target.env }}
       extensions: ${{ matrix.target.extensions }}
 
+  image_scan:
+    runs-on: ubuntu-latest
+    name: Scan docker image
+    needs:
+      - docker_build
+
+    steps:
+      - name: Derive version
+        id: vars
+        run: |
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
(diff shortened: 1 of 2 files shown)


sergei-maertens (commented):
From the log output I'm not yet convinced that it's using the locally downloaded/loaded image :thinking: 

I see that trivy itself is starting a container and mounting some volumes, so this should be double checked.

joeribekker:
@sergei-maertens maybe we should just do the tarball scan:

```
    - name: Run Trivy vulnerability scanner in tarball mode
      uses: aquasecurity/trivy-action@master
      with:
        input: /github/workspace/vuln-image.tar
        severity: 'CRITICAL,HIGH'
```

See: https://github.com/aquasecurity/trivy-action#scanning-a-tarball

sergei-maertens:
> @sergei-maertens maybe we should just do the tarball scan:
> 
> ```
>     - name: Run Trivy vulnerability scanner in tarball mode
>       uses: aquasecurity/trivy-action@master
>       with:
>         input: /github/workspace/vuln-image.tar
>         severity: 'CRITICAL,HIGH'
> ```
> 
> See: https://github.com/aquasecurity/trivy-action#scanning-a-tarball

yes, that's the way to go! I'll update the PR

I looked into the CodeQL scanning warning too, it's borking on some python syntax so I'm not sure what the proper way to resolve that is, but it's not urgent.

sergei-maertens (approved):

PR #3619, opened 2023-11-17
[#3558] Add an extra `identifier` field to account for analytics

### LINKED ISSUE #3558: Update analytics tools according to new CSPSetting model
We have already updated digid-eherkenning and Ogone (see PR #3514 ) in order to be connected with the CSPSetting model (with a generic foreign key).

We should do the same for the analytics tools for consistency.

Viicos:
Fixes #3558.


src/openforms/config/models.py
```diff
@@ -661,14 +662,25 @@ def as_dict(self):
...
         instances = [
-            CSPSetting(content_object=obj, directive=directive, value=value)
-            for directive, value in settings
+            CSPSetting(
+                content_object=obj,
+                directive=directive,
+                value=value,
+                identifier=identifier or "",
+            )
+            for directive, value, identifier in zip(*zip_longest(*settings))
```
Viicos, line 683: Impersonated @CharString's style :face_in_clouds:

src/openforms/config/models.py
```diff
@@ -661,14 +662,25 @@ def as_dict(self):
 class CSPSettingManager(models.Manager.from_queryset(CSPSettingQuerySet)):
     @transaction.atomic
     def set_for(
-        self, obj: models.Model, settings: list[tuple[CSPDirective, str]]
+        self,
+        obj: models.Model,
+        *settings: tuple[CSPDirective, str] | tuple[CSPDirective, str, str],
```
Viicos, line 667: Makes the annotation a bit simpler, but can be confusing to use. So I can revert if prefered.

SilviaAmAm (commented):
If I understand correctly, this new identifier field is mostly there for debugging purposes to know from which analytics tool the various CSP settings came from?

Should there be a test asserting that the identifier is set as expected when updating analytics tools?

Viicos:
> If I understand correctly, this new identifier field is mostly there for debugging purposes to know from which analytics tool the various CSP settings came from?

Yes, but primarily because currently `CSPSetting` has a generic foreign key ($\Leftrightarrow$ a foreign key to any kind of model) to the configuration model that "initiated" this CSP directive. For analytics, there's a single model holding logic for multiple analytics provider, so we need an extra field to identify the provider within the `AnalyticsToolsConfiguration` model (and probably other ones in the future).

> Should there be a test asserting that the identifier is set as expected when updating analytics tools?

True forgot to update this

sergei-maertens:
@SilviaAmAm @Viicos 

> this new identifier field is mostly there for debugging purposes

no, it is mostly there to be able to reliably "find and replace" the relevant CSP configuration entries when some configuration (aspect) changes.

If an Ogone Merchant is configured, some URLs can be configured per merchant which needs to be added to the CSP configuration. Consequently, when the merchant is deleted or modified, those CSP directives need to be removed again. Similarly with DigiD/EHerkenning - when that configuration changes, the new version of the configuration prescribes the required CSP configuration state, so we delete old records and re-insert the new desired state.

The GFK makes this possible - we can just delete the associated records without having to look up old values and deal with requirements that change over time, we can just delete any directive that is linked to a particular configuration object.

Now, when dealing with analytics, there are multiple providers and they are all configured in the same model instance (a solo model), so we need extra context to prevent us from deleting CSP Directives from analytics provider X if we're changing analytics provider Y configuration.

Viicos:
> The implementation details (that an identifier field is used) are not interesting here, what is relevant is the behaviour described above and that should be tested rather than asserting that a particular identifier was set.

You're right, the actual implementation did not take this into account. I've added a test.

src/openforms/analytics_tools/tests/test_cspsettings_identifier.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@override_settings(SOLO_CACHE=None)
+class CSPIdentifierTests(AnalyticsMixin, TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+        cls.gtm_code = "GTM-XXXX"
+        cls.ga_code = "UA-XXXXX-Y"
+        cls.matomo_url = "https://example.com"
```
sergei-maertens, line 17: can you remove this indirection and set the values directly in the test itself? This is an unnecessary context switch when debugging/reading the test.

src/openforms/analytics_tools/tests/test_cspsettings_identifier.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def disabling_analytics_does_not_delete_unrelated_csp_settings(self):
+        self.config.gtm_code = self.ga_code
+        self.config.ga_code = self.gtm_code
```
sergei-maertens, line 21: you've swapped the attributes :grimacing:

src/openforms/analytics_tools/tests/test_cspsettings_identifier.py
```diff
@@ -0,0 +1,43 @@
...
+
+        assert csp_settings.count() == 2
+
+        self.config.enable_google_analytics = False
+        self.config.save()
+
+        try:
+            CSPSetting.objects.get(identifier=AnalyticsTools.matomo)
+        except CSPSetting.DoesNotExist:
+            self.fail("CSPSetting instance should exist")
```
sergei-maertens, line 43: I'd also assert that the records for Google are gone:

```py
other_settings = CSPSetting.objects.filter(content_object=self.config)
self.assertFalse(other_settings.exists())
```

src/openforms/analytics_tools/tests/test_cspsettings_identifier.py
```diff
@@ -0,0 +1,43 @@
...
+        self.config.matomo_url = self.matomo_url
+        self.config.matomo_site_id = "1234"
+        self.config.enable_matomo_site_analytics = True
+        self.config.clean()
+        self.config.save()
+
+        csp_settings = CSPSetting.objects.filter(
+            content_type=get_content_type_for_model(self.config),
+            object_id=str(self.config.pk),
+        )
```
sergei-maertens, line 33: You can use the readable GFK queries instead:

```suggestion
        csp_settings = CSPSetting.objects.filter(content_object=self.config)
```

src/openforms/analytics_tools/utils.py
```diff
@@ -130,11 +125,13 @@ def update_analytical_cookies(
         Cookie.objects.filter(name__in=[cookie["name"] for cookie in cookies]).delete()
 
 
-def update_csp(csps: list[CSPDict], create: bool):
+def update_csp(
+    config_model: "AnalyticsToolsConfiguration",
```
sergei-maertens, line 129: it's an instance, not a model, so the variable name can be improved upon

"a django model" refers the model class and associated table :)

src/openforms/analytics_tools/utils.py
```diff
@@ -130,11 +125,13 @@ def update_analytical_cookies(
...
+):
     if create:
-        instances = [
-            CSPSetting(directive=csp["directive"], value=csp["value"]) for csp in csps
-        ]
-        CSPSetting.objects.bulk_create(instances)
+        CSPSetting.objects.set_for(config_model, settings=csps, identifier=identifier)
     else:
-        CSPSetting.objects.filter(value__in=[csp["value"] for csp in csps]).delete()
+        CSPSetting.objects.delete_for(config_model, identifier=identifier)
```
sergei-maertens, line 137: I'd say that the caller needs to properly pass teh `csps` list:

```py
csps = [] if not create else [...]
update_csp(config, csps, identifier)
```
and at this point, this `update_csp` util should probably be skipped entirely since it really just wraps `CSPSetting.objects.set_for(...)`

Setting an empty list of settings is equivalent to `CSPSetting.objects.delete_for`, which means that `delete_for` should not exist.

src/openforms/config/models.py
```diff
@@ -661,35 +661,59 @@ def as_dict(self):
...
         ]
 
-        CSPSetting.objects.filter(
-            content_type=get_content_type_for_model(obj), object_id=str(obj.id)
-        ).delete()
-
+        self.delete_for(obj, identifier)
         self.bulk_create(instances)
 
+    def delete_for(self, obj: models.Model, identifier: str | None = None) -> None:
```
sergei-maertens, line 686: as remarked above, I don't see why this should exist if `set_for` exists.

src/openforms/config/models.py
```diff
@@ -661,35 +661,59 @@ def as_dict(self):
...
+        CSPSetting.objects.filter(**filter).delete()
+
 
 class CSPSetting(models.Model):
     directive = models.CharField(
         _("directive"),
         max_length=64,
-        help_text=_("CSP header directive"),
         choices=CSPDirective.choices,
+        help_text=_("CSP header directive"),
```
sergei-maertens, line 701:
```suggestion
        help_text=_("CSP header directive."),
```

sergei-maertens (changes requested):
This also requires a data migration to set the identifier values for existing records, based on the old 'value' comparison/logic :)

src/openforms/analytics_tools/utils.py
```diff
@@ -9,9 +9,11 @@
 from cookie_consent.models import Cookie, CookieGroup
 
 from openforms.config.models import CSPSetting
+from openforms.config.utils import CSPEntry
 from openforms.typing import JSONObject
 
 if TYPE_CHECKING:  # pragma: no cover
+    from .constants import AnalyticsTools
```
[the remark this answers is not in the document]
Viicos: None actually, just that it was already there and this new import is only used in type hints so figured I might add it in the `if TYPE_CHECKING` block.

src/openforms/config/models.py
```diff
@@ -661,14 +661,25 @@ def as_dict(self):
...
+        obj: models.Model,
+        settings: list[CSPEntry],
     ) -> None:
         """
-        Deletes all the connected csp settings and creates new ones based on the new provided data.
+        Deletes all the connected CSP settings and creates new ones based on the new provided data.
+
+        :param obj: The configuration model providing this CSP entry.
+        :param `*settings`: Either a two-tuple or a three-tuple containing values used to create
+          the underlying ``CSPSetting`` model.
```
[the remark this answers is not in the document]
Viicos: So I first went from a list of tuples to a dto dataclass because I've added this extra `identifier` attribute. But I've then moved `identifier` to the signature of `set_for`.

I'm not a fan of DTO objects in general, so we can either revert to the old way (list of tuples), or use unsaved instances instead. But having unsaved instances feels weird as well:

```python
CSPSetting.objects.set_for(settings=[CSPSetting(...), CSPSetting(...)])
```

What do you think?

src/openforms/analytics_tools/tests/test_cspsettings_identifier.py
Viicos: I was probably thinking about adding multiple tests at first

src/openforms/analytics_tools/utils.py
Viicos: Simplifies things a lot indeed.

src/openforms/config/models.py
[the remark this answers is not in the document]
sergei-maertens: List of tuples is probably best then as it avoids potentially dealing with circular imports of the model.

src/openforms/config/models.py
```diff
@@ -661,19 +661,35 @@ def as_dict(self):
 class CSPSettingManager(models.Manager.from_queryset(CSPSettingQuerySet)):
     @transaction.atomic
     def set_for(
-        self, obj: models.Model, settings: list[tuple[CSPDirective, str]]
+        self,
+        obj: models.Model,
+        settings: list[tuple[CSPDirective, str]],
+        identifier: str | None = None,
```
sergei-maertens, line 667:
```suggestion
        identifier: str = "",
```
seems to make more sense to me so that you have a consistent type

sergei-maertens (approved):

src/openforms/config/models.py
Viicos: True, this is a leftover from the previous solution where I wanted to have the ability to _not_ filter against `identifier` (by leaving it as `None`). This now doesn't make sense to have it.

Viicos: My bad this is still relevant:

```python
predicate = {
    "content_type": get_content_type_for_model(obj),
    "object_id": str(obj.id),
}
if identifier is not None:
    predicate["identifier"] = identifier
```

The idea was that if you don't specify an `identifier`, all `CSPSettings` (matching the content object) would be deleted. You'd have to explicitly specify `identifier=""` if you only want to delete the ones with an empty/unset identifier (but not the ones _with_ an identifier!)

Is this desirable?

src/openforms/config/models.py
```diff
@@ -661,19 +661,35 @@ def as_dict(self):
...
 
-        CSPSetting.objects.filter(
-            content_type=get_content_type_for_model(obj), object_id=str(obj.id)
-        ).delete()
+        predicate = {
+            "content_type": get_content_type_for_model(obj),
+            "object_id": str(obj.id),
+        }
+        if identifier is not None:
+            predicate["identifier"] = identifier
```
sergei-maertens, line 691: this condition can/should be changed now into

```suggestion
        if identifier:
            predicate["identifier"] = identifier
```
or just drop it alltogether, IMO I think you can just pass in all three filter params directly to `.filter(...)`, the result is the same

src/openforms/config/models.py
sergei-maertens: in practice that behaviour is not used, right? if we don't have a need for identifier, records will never be set/created with an identifier, so filtering on empty string will effectively delete them. for analytics we need it, and we don't call that one without identifier provided.

sergei-maertens (approved):

PR #3622, opened 2023-11-20
:lock: Add CSP object-src directive to settings

sergei-maertens:
This configuration was missing (but as a workaround it can be added through the admin interface). Now we include it, which restricts some functionality (which was not intended in the first place).

The envvar lookup/fallback allows an escape hatch for those who are making use of object and embed tags.

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -1121,6 +1121,10 @@ CSP_IMG_SRC = (
     + config("CSP_EXTRA_IMG_SRC", default=[], split=True)
 )
 
+# affects <object> and <embed> tags, block everything by default but allow deploy-time
+# overrides.
+CSP_OBJECT_SRC = config("CSP_OBJECT_SRC", default=["'none'"], split=True)
+
 # we must include this explicitly, otherwise the style-src only includes the nonce because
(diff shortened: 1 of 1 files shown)


src/openforms/conf/base.py
```diff
@@ -1121,6 +1121,10 @@
     + config("CSP_EXTRA_IMG_SRC", default=[], split=True)
 )
 
+# affects <object> and <embed> tags, block everything by default but allow deploy-time
+# overrides.
+CSP_OBJECT_SRC = config("CSP_OBJECT_SRC", default=["'none'"], split=True)
```
CharString, line 1126: Funny... passing `split=False` would do the same.

CharString (approved):
Does this need mentioning in the docs or just a release note?

sergei-maertens:
> Does this need mentioning in the docs or just a release note?

Just a release note, I don't foresee anyone actually using `<object>` or `<embed>` but you never know :|

src/openforms/conf/base.py
sergei-maertens: Created an internal taiga issue for this one :grimacing:

PR #3628, opened 2023-11-22
[#3626] Test resume flow

### LINKED ISSUE #3626: Cannot pause and resume forms

19f34cb8ce79879069b3e3710027d6640ee06811 

When resuming the form, this happens:
Screenshot from 2023-11-21 15-29-53

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #3626

Add an E2E test for the resume flow, so that we can catch problems earlier in the future.

### DIFF
--- a/src/openforms/tests/e2e/test_resume_form.py
+++ b/src/openforms/tests/e2e/test_resume_form.py
@@ -1,0 +1,77 @@
+from unittest.mock import patch
+
+from django.urls import reverse
+
+from asgiref.sync import sync_to_async
+from furl import furl
+from playwright.async_api import expect
+from rest_framework.test import APIRequestFactory
+
+from openforms.forms.tests.factories import FormFactory
+from openforms.submissions.api.serializers import SubmissionSuspensionSerializer
+from openforms.submissions.models import Submission
+from openforms.tests.e2e.base import E2ETestCase, browser_page
+
+factory = APIRequestFactory()
+
+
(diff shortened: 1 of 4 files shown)


src/openforms/tests/e2e/test_resume_form.py
```diff
@@ -0,0 +1,69 @@
...
+
+                await page.get_by_role("button", name="Tussentijds opslaan").click()
+
+                await page.get_by_label("Uw e-mailadres").fill("test@test.nl")
+
+                await page.get_by_role("button", name="Later verdergaan").click()
+
+                resume_url = await get_resume_url(form, request, self.live_server_url)
+
+                time.sleep(1)
```
SilviaAmAm, line 63: For some reason, if you don't wait a bit, it doesn't want to go to the new page :thinking:

sergei-maertens: can you do an assertion that waits for the page to have a particular URL? it's because the button click makes an async request to the backend that needs to complete first, but the `page.goto` doesn't know that that is still pending

You can do stuff like:

```py
await page.wait_for_url(f"{BASE}#/overzicht")
```

SilviaAmAm:
Todo: need to fix so that the SDK build is present

sergei-maertens (approved):

PR #3630, opened 2023-11-23
🏷️ Fix generic type hinting of registry

Viicos:
A small improvement regarding type checking of the `BaseRegistry` class:

```python
# Before
a = register("coSign")
reveal_type(a)  # Revealed type is "def (type[Any]) -> type[Any]"
b = a(...)
reveal_type(b)  # Revealed type is "Any"

# After
a = register("coSign")
reveal_type(a)  # Revealed type is "def a(type[BaseValidator]) -> type[BaseValidator]"
b = a(...)
reveal_type(b)  # Revealed type is "type[BaseValidator]"
```

I think it doesn't make sense to have a type alias to `type[PluginType_co]` outside any definition, because `PluginType_co` isn't bound to any generic class/function yet.

Only downside here is that in `Callable[[type[PluginType_co]], type[PluginType_co]]`, `type[PluginType_co]` should also be a `TypeVar` bound to `type[PluginType_co]`, but typevars bound to other typevars are not allowed.

### DIFF
--- a/src/openforms/plugins/registry.py
+++ b/src/openforms/plugins/registry.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Callable, Generic, TypeAlias, TypeVar
+from typing import TYPE_CHECKING, Callable, Generic, TypeVar
 
 from django.db import OperationalError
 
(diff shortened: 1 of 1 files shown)


Viicos:
> shame that it makes the type annotations considerably less readable, but if this is the only way to get it to work properly in editors then so be it

Yeah maybe that could be improved on the Python side

Issue #3605, opened 2023-11-14
Leave number fields intact, don't add dot-formatting to registration backends


hlmr-erik:
### Thema / Theme

Plugins

### Omschrijving / Description

Please leave the formatting of the contents of number fields to BE systems. Dot-formatting is only useful when a human would be processing the form data (as maybe when the form is submitted as an e-mail). Having dots added to fields made to contain a year, KVK number etc. is undesirable. As is allowing users to fill a field with text as well, or setting a custom RegEx filter each and every time.

![image](https://github.com/open-formulieren/open-forms/assets/111565013/1512857e-e022-418f-8f93-54c7411a78d5)

sergei-maertens:
I suspect this is a template localization issue and not directly related to the frontend data input.

joeribekker:
Refinement: This can probably be solved by adding localize off-tag in the template. Preferably make this a global option on sandbox rendering backend (see: `openforms.template` module)

PR #3632, opened 2023-11-24
[#3605] Remove localization from numbers in STUF-ZDS registration

vaszig:


src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -2806,3 +2806,99 @@ def test_user_defined_variables(self, m, mock_task):
...
+                    "registration": {
+                        "attribute": RegistrationAttribute.initiator_voornamen,
+                    },
+                },
+                {
+                    "key": "extra",
+                },
+                {
+                    "key": "language_code",
+                },
```
sergei-maertens, line 2826: are these relevant for localization?

src/stuf/stuf_bg/templates/stuf_bg/StufBgRequest.xml
```diff
@@ -1,5 +1,6 @@
-{% extends "stuf/soap_envelope.xml" %}{% load stuf %}
+{% extends "stuf/soap_envelope.xml" %}{% load stuf l10n %}
 {% block body %}
+{% localize off %}
```
sergei-maertens, line 3: is it possible to put this outside the `block body`? That way child templates overriding the block body don't need to remember overriding the `localize` tag.

Viicos:
Just a general question regarding the way these SOAP envelopes are generated, are we using Django templates because it's easier than building the XML in Python directly, or is there something I am missing?

sergei-maertens:
> > Just a general question regarding the way these SOAP envelopes are generated, are we using Django templates because it's easier than building the XML in Python directly, or is there something I am missing?
> 
> I haven't asked this specifically to @sergei-maertens when we discussed this issue, but I assume it is one of the approaches and maybe an easier one.

@Viicos different people have different opinions about this and there's a tradeoff for ease of comparing the templates against request/response examples in SoapUI and the problems that using the DTL creates in an XML context.

vaszig:
Disabling localization outside the block doesn't work.
https://docs.djangoproject.com/en/4.2/topics/i18n/formatting/#localize

sergei-maertens (approved):
:+1:

PR #3650, opened 2023-11-30
[#3559] Add support for Piwik Pro Tag Manager

### LINKED ISSUE #3559: As municipality I want to add Piwik Pro tagmanager to Open Formulieren
### Thema / Theme

Other

... to measure statistics more dynamically.

Taiga DH: 598

Piwik Pro tagmanager requires a different script than Piwik Pro analytics.

_No response_

_No response_

Viicos:
Fixes #3559.

Script was taken from: https://developers.piwik.pro/en/latest/tag_manager/content_security_policy.html#adding-nonce-to-container-code

Tag manager "injects" tags (<=> javascript code) in two ways: asynchronously and synchronously.

For asynchronous tags, nothing is required as Piwik pro will inject the nonce. For sync tags however, we need to follow the following: https://developers.piwik.pro/en/latest/tag_manager/content_security_policy.html#adjusting-tags-to-work-with-content-security-policy. Not sure how we want to handle that.

Also, I've assumed tag manager would run on the same url/site id, so no need to add these fields in the admin


sergei-maertens (commented):
It now looks like you can't use tag manager without also using piwik pro analytics directly, but I'd expect that piwik pro tag manager is actually responsible for hooking up the analytics. Can you clarify that?

src/openforms/analytics_tools/templates/analytics_tools/piwik_pro_tag_manager.html
```diff
@@ -0,0 +1,33 @@
...
+</script>
+
+<script id="piwik-pro-tag-manager-sync" type="text/javascript" nonce="{{ request.csp_nonce }}">
+  (function(window, document, dataLayerName, id) {
+  function stgCreateCookie(a,b,c){var d="";if(c){var e=new Date;e.setTime(e.getTime()+24*c*60*60*1e3),d=";expires="+e.toUTCString()}document.cookie=a+"="+b+d+"; path=/"}
+  var isStgDebug=(window.location.href.match("stg_debug")||document.cookie.match("stg_debug"))&&!window.location.href.match("stg_disable_debug");
+  stgCreateCookie("stg_debug",isStgDebug?1:"",isStgDebug?14:-1);
+  var qP=[];dataLayerName!=="dataLayer"&&qP.push("data_layer_name="+dataLayerName),isStgDebug&&qP.push("stg_debug");
+  var qPString=qP.length>0?("?"+qP.join("&")):"";
+  document.write('<script src="{{ analytics_tools_config.piwik_pro_url|escapejs }}/'+id+'.sync.js' + qPString + '" nonce="{{ request.csp_nonce }}"></' + 'script>');
```
sergei-maertens, line 29:
```suggestion
  document.write('<script src="{{ analytics_tools_config.piwik_pro_url|escapejs }}/'+id+'.sync.js' + qPString + '" nonce="{{ request.csp_nonce|escapejs }}"></' + 'script>');
```
?

src/openforms/analytics_tools/templates/analytics_tools/piwik_pro_tag_manager.html
```diff
@@ -0,0 +1,33 @@
...
+<!-- Start Piwik PRO tag manager head tag -->
+{% comment %} See https://developers.piwik.pro/en/latest/tag_manager/content_security_policy.html#adding-nonce-to-container-code {% endcomment %}
+<script id="piwik-pro-tag-manager-async" type="text/javascript" nonce="{{ request.csp_nonce }}">
+  (function(window, document, dataLayerName, id) {
+  window[dataLayerName]=window[dataLayerName]||[],window[dataLayerName].push({start:(new Date).getTime(),event:"stg.start"});
+  var scripts=document.getElementsByTagName('script')[0],tags=document.createElement('script');
+  function stgCreateCookie(a,b,c){var d="";if(c){var e=new Date;e.setTime(e.getTime()+24*c*60*60*1e3),d=";expires="+e.toUTCString()}document.cookie=a+"="+b+d+"; path=/"}
+  var isStgDebug=(window.location.href.match("stg_debug")||document.cookie.match("stg_debug"))&&!window.location.href.match("stg_disable_debug");
+  stgCreateCookie("stg_debug",isStgDebug?1:"",isStgDebug?14:-1);
+  var qP=[];dataLayerName!=="dataLayer"&&qP.push("data_layer_name="+dataLayerName),tags.nonce="{{ request.csp_nonce }}",isStgDebug&&qP.push("stg_debug");
```
sergei-maertens, line 13:
```suggestion
  var qP=[];dataLayerName!=="dataLayer"&&qP.push("data_layer_name="+dataLayerName),tags.nonce="{{ request.csp_nonce|escapejs }}",isStgDebug&&qP.push("stg_debug");
```
?

src/openforms/analytics_tools/templates/analytics_tools/piwik_pro_tag_manager.html
Viicos: Other script templates don't have it :shrug:

Viicos:
> It now looks like you can't use tag manager without also using piwik pro analytics directly, but I'd expect that piwik pro tag manager is actually responsible for hooking up the analytics. Can you clarify that?

I'm not sure about that. Both products seem to be standalone packages, but I would assume clients having Tag manager already have Analytics.

src/openforms/analytics_tools/templates/analytics_tools/piwik_pro_tag_manager.html
Viicos: `csp_nonce` is not an untrusted value, and docs give the same example: https://django-csp.readthedocs.io/en/latest/nonce.html#middleware

sergei-maertens: it's a very different thing if the variable goes into an HTML tag attribute, for which Django applies regular HTML escaping, versus if you are using a variable inside of Javascript code which requires a different escape context.

Which other script templates don't have it, so that I can take a look at whether they should have it or not?

The thing is that we don't really know what character set is in the CSP nonce - now I happen to know it should be base64 chars so it is *unlikely* that it will cause problems, but the escaping should still be applied if only to serve as a not-bad example of how to use template variables inside of dynamic JS code. Someone else looking at this code could deduce that this is the way to use context variables in JS code inside django templates.

But, if the csp nonce value were to have a <kbd>`</kbd> character, you'll be in trouble because now your Javascript code is broken since a literal `'` is rendered.

sergei-maertens:
> > It now looks like you can't use tag manager without also using piwik pro analytics directly, but I'd expect that piwik pro tag manager is actually responsible for hooking up the analytics. Can you clarify that?
> 
> I'm not sure about that. Both products seem to be standalone packages, but I would assume clients having Tag manager already have Analytics.

Usually the reason to use a tag manager (especially the case with google) is that you don't need to micro-manage your analytics snippets anymore, so intuitively I'd say that one does not imply the other here either, and you should be able to enable tag manager without having to enable analytics.

src/openforms/analytics_tools/templates/analytics_tools/piwik_pro_tag_manager.html
Viicos: Yes sorry I thought you commented on the csp nonce tag here:
https://github.com/open-formulieren/open-forms/blob/7c89d2746199e8a4b6c466fbc6bec37e0a6fb315/src/openforms/analytics_tools/templates/analytics_tools/piwik_pro_tag_manager.html#L6

So in this case yes it makes sense!

sergei-maertens (changes requested):

src/openforms/analytics_tools/tests/test_analytics_rendering.py
```diff
@@ -84,6 +84,40 @@ def test_piwik_pro_rendering(self):
...
+        self.config.piwik_pro_url = self.url
+        # Enable Tag manager
+        self.config.enable_piwik_pro_tag_manager = True
+        self.config.save()
+
+        # Accept cookies
+        form_page = self.app.get(self.url)
+
+        tag_manager_async = form_page.pyquery("#piwik-pro-tag-manager-async")
+        tag_manager_sync = form_page.pyquery("#piwik-pro-tag-manager-sync")
```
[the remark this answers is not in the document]
sergei-maertens: https://help.piwik.pro/support/tag-manager/synchronous-asynchronous-tag/ let's mention this link as a comment in the template and/or model code :)

sergei-maertens (approved):

PR #3658, opened 2023-12-04
Payment status update for the Object API registration backend

### LINKED ISSUE #713: Implement payment status update for ObjectsAPI registration backend
The Objects API registration backend plugin doesn't have a update_payment_status() implementation.

(it wasn't part of the initial tickets and could be out of sight)

SilviaAmAm:
Fixes #713


src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -229,6 +228,67 @@ def get_custom_templatetags_libraries(self) -> list[str]:
...
         ]
 
+    def update_payment_status(self, submission: "Submission", options: dict) -> None:
+        config = ObjectsAPIConfig.get_solo()
+        assert isinstance(config, ObjectsAPIConfig)
+        config.apply_defaults_to(options)
+
+        if not options["payment_status_update_json"]:
+            logger.warning(
+                "Skipping payment status update because not template was configured."
```
Viicos, line 238:
```suggestion
                "Skipping payment status update because no template was configured."
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -229,6 +228,67 @@ def get_custom_templatetags_libraries(self) -> list[str]:
...
+            return
+
+        context = {
+            "variables": get_variables_for_context(submission),
+            "payment": {
+                "completed": submission.payment_user_has_paid,
+                "amount": submission.payments.sum_amount(),
+                "public_order_ids": ",".join(
+                    submission.payments.get_completed_public_order_ids()
+                ),
```
Viicos, line 249: This is repeated in the `register_submission` method, I don't know if it's worth moving it in a single method?

Viicos (approved):

src/openforms/registrations/contrib/objects_api/plugin.py
sergei-maertens: probably just using a utility function at the module level `def get_payment_context_data(submission: Submission)` would be sufficient :)

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -229,6 +228,67 @@ def get_custom_templatetags_libraries(self) -> list[str]:
...
+        updated_record_data = render_from_string(
+            options["payment_status_update_json"],
+            context=context,
+            disable_autoescape=True,
+            backend=openforms_backend,
+        )
+        self._check_size_rendered_json(updated_record_data)
+
+        with catch_json_decode_errors():
+            updated_record_data = json.loads(updated_record_data)
```
sergei-maertens, line 263: I think it'd be good to have an anti-corruption layer that will allow us to refactor in the future, so wrapping this (here) and in the `register` hook into something with the following signature would be useful:

```py
def render_to_json(template: str, context: dict[str, Any]) -> JSONValue:
    """Render the provided template and context into a JSON serializable object"""
    ...
```

we can then in the future change this implementation with a better solution than Django templates.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -229,6 +228,67 @@ def get_custom_templatetags_libraries(self) -> list[str]:
...
+    def _check_size_rendered_json(self, record_data: str) -> None:
+        if (
+            data_size := sys.getsizeof(record_data)
+        ) > settings.MAX_UNTRUSTED_JSON_PARSE_SIZE:
+            formatted_size = filesizeformat(data_size)
+            max_size = filesizeformat(settings.MAX_UNTRUSTED_JSON_PARSE_SIZE)
+            raise SuspiciousOperation(
+                f"Templated out content JSON exceeds the maximum size {max_size} ("
+                f"it is {formatted_size})."
+            )
```
sergei-maertens, line 290: `self` is unused here and nothing about this implementation is really tied to the objects API (it's just currently the only place where we use it), so I'd opt to put this in a simple module-level utility function.

We can then bundle/move it together with the `render_to_json` (and make it part of that function maybe) when we get to it.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -144,6 +148,13 @@ def register_submission(
         context = {
             "_submission": submission,
             "productaanvraag_type": options["productaanvraag_type"],
+            "payment": {
+                "completed": submission.payment_user_has_paid,
+                "amount": submission.payments.sum_amount(),
```
sergei-maertens, line 153: I think we need to properly stringify this before passing to the template engine, since the localization in the templates will cause problems with large numbers. Please add a test case for that situation too (try using number `1_000` but also `3.141592` and see how many decimals are outputted and what the rounding behaviour is).

Because floats are inherently limited, I think the resulting JSON type should be a string where a `.` is used as decimal separator and no grouping occurs, so expected is:

* `1_000` -> `"1000"`
* `3.1415926535` -> `"3.1415926535"`

Unfortunately I don't see a [`format`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#page-13) for decimal, so maybe a string isn't the correct solution as in theory the object type schema can also specify the value to have a particular precision by using:

```
type: number
multipleOf: 0.01
```

So perhaps accepting the limitations of floats is okay here, since we process the payment already anyway :thinking:

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -144,6 +148,13 @@ def register_submission(
...
+                "public_order_ids": ",".join(
```
sergei-maertens, line 154: as remarked above, shouldn't this be an array of IDs instead of making the consumer comma-split it? Let's use the flexibility of JSON we have available.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -1191,3 +1208,60 @@ def test_submission_with_objects_api_escapes_html(self, m):
...
+        self.assertEqual(len(m.request_history), 2)
+
+        object_create = m.last_request
+        body = object_create.json()
+
+        self.assertEqual(
+            body["record"]["data"]["payment"],
+            {
+                "completed": False,
+                "amount": "10,00",
```
sergei-maertens, line 1264: This is not reliably machine-parseable, it should probably be:

```suggestion
                "amount": 10.00,
```

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/plugin.py
SilviaAmAm: Discussed:
The object type schema should dictate the precision of this field (`type: number, multipleOf: 0.01`). We then send a number in the json and not a string.

docs/manual/templates.rst
```diff
@@ -476,6 +476,17 @@ alle gegevens uit het formulier en de waarden ingevuld door de gebruiker.
 
    See :ref:`example_form_with_geometry` for a more detailed example.
 
+Voor formulieren die een betaling vereisen, is het ook mogelijk om informatie over de betaling toe te voegen.
+Als de gebruiker betaalt, kan de status van de betaling in de Object API bijgewerkt worden. De structuur van het
+veld ``payment`` binnen het ``record`` veld is ook per formulier instelbaar met een sjabloon.
+In dit sjabloon kunnen alleen de inzending variabelen (``variables.<naam van variabele>``) en de ``payment`` variabele
```
sergei-maertens, line 482:
```suggestion
In dit sjabloon kunnen alleen de inzendingsvariabelen (``variables.<naam van variabele>``) en de ``payment`` variabele
```

docs/manual/templates.rst
```diff
@@ -494,6 +505,9 @@ Variabele                              Beschrijving
 ``{% json_summary %}``                 JSON met ``"<variabele-eigenschapsnaam>": "<waarde>"`` van alle formuliervelden.
 ``{% uploaded_attachment_urls %}``     Een lijst met de URLs van documenten toegevoegd door de inzender. De URLs verwijzen naar het geregistreerde document in de Documenten API.
 ``{% as_geo_json variables.map %}``    Sluit de gerefereerde variabele (`variables.map`) in als JSON.
+``{{ payment.completed }}``            Indicatie of de betaling voltooid is.
+``{{ payment.amount }}``               Bedrag die betaald moet worden.
```
sergei-maertens, line 509:
```suggestion
``{{ payment.amount }}``               Bedrag dat betaald moet worden.
```

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -229,6 +205,41 @@ def get_custom_templatetags_libraries(self) -> list[str]:
...
+            "variables": get_variables_for_context(submission),
+            "payment": get_payment_context_data(submission),
+        }
+
+        updated_record_data = render_to_json(
+            options["payment_status_update_json"], context
+        )
+        updated_object_data = {
+            "record": {
+                "data": {"payment": updated_record_data},
```
sergei-maertens, line 229: isn't it more logical to leave the form builders to specify the `payment` key rather than hardcoding it? IMO this should be part of the template.

src/openforms/registrations/contrib/objects_api/tests/test_utils.py
```diff
@@ -63,3 +69,46 @@ def test_different_input(self):
...
+        submission = SubmissionFactory.create()
+        payment1, payment2 = SubmissionPaymentFactory.create_batch(
+            2, submission=submission, status=PaymentStatus.completed, amount=10
+        )
+
+        template = textwrap.dedent(
+            """{
+            "completed": {% if payment.completed %}true{% else %}false{% endif %},
+            "amount": {{ payment.amount }},
+            "public_order_ids": {{ payment.public_order_ids }}
```
sergei-maertens, line 101: I don't understand how this variable results in valid JSON other than pure coincidence. In fact, it may not even be valid JSON due to python strings typically being printend with single quotes rather than double quotes.

src/openforms/registrations/contrib/objects_api/utils.py
```diff
@@ -21,3 +29,43 @@ def html_escape_json(value: T) -> T:
...
+    return json_data
+
+
+def get_payment_context_data(submission) -> dict[str, Any]:
+    return {
+        "completed": submission.payment_user_has_paid,
+        "amount": str(submission.payments.sum_amount()),
+        "public_order_ids": json.dumps(
+            submission.payments.get_completed_public_order_ids()
+        ),
```
sergei-maertens, line 70: ah, that explains...

To be honest, I think this is very confusing for the people making templates. Sometimes a variable already outputs, JSON, other times they need to use a special template tag... It's also hard for us to track what is already serialized as JSON and what isn't.

I'd replace this in the default template with a simple for loop:

```django
[{% for order_id in payment.public_order_ids%}"{{ order_id|escapejs }}"{% if not forloop.last %},{% endif %}{% endfor %}
```

sergei-maertens (changes requested):

docs/manual/templates.rst
SilviaAmAm: One day I will get this merging of words right... :stuck_out_tongue:

SilviaAmAm:
Note: should not be merged before the Object API is fixed

sergei-maertens (approved):
Still waiting for confirmation on the Objects API patch before we merge this.

PR #3663, opened 2023-12-05
[#3601] Register submission data to both Objects and ZGW APIs

### LINKED ISSUE #3601: As a municipality I want to register part of the submission data as a Zaak in ZGW and an object in the Objects API
... so that I can relate the object to the created Zaak.

Ref: Utr 138

vaszig:
Fixes #3601 

The task describes our decision to make this explicitly possible only for ZGW and Objects API. 
There is an image as well, which helps a bit concerning the flow after a submission is completed.


src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -444,3 +490,61 @@ def get_config_actions(self) -> list[Action]:
                 ),
             ),
         ]
+
+    def register_submission_to_objects_api(
```
vaszig, line 494: This could be a function in the Objects API itself (in `utils.py` or in `client.py)` and used by both ZGW and Objects API. I thought that it is more clear to have it here duplicated and show exactly what we are doing but feel free to tell me if this should be in another file and used by both.

Viicos: Taking a look at the `ObjectsAPIRegistration.register_submission`, most of it seems to be duplicated, but at the same time the context doesn't seem to be the same. So trying to be DRY here might end up being more confusing. Perhaps having a single method handling things starting at `record_data = ...`?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -401,6 +437,16 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
 
+        # Register submission to Objects API if configured
+        if (
+            options.get("objecttype")
+            and options.get("objecttype_version")
+            and options.get("content_json")
+        ):
+            result = self.register_submission_to_objects_api(
+                submission, options, result
+            )
```
Viicos, line 448: I find the API here a bit unusual, as `register_submission_to_objects_api` mutates `result` and returns it. I think it would be better to return the object API result and set `result["objects_api"] = res` after the call to the method. Might be related to my other comment

Viicos (changes requested):

src/openforms/registrations/contrib/zgw_apis/plugin.py
vaszig: Yes, you are right!It 's not clear enough the way I did it, I will fix that thanks.

src/openforms/registrations/contrib/objects_api/helpers.py
Viicos, line 1: Silvia's PR also has some refactoring here. In particular, I think we could make use of some [utils](https://github.com/open-formulieren/open-forms/pull/3658/files#diff-bebf5c215247eab992b166baedd0e3dc1f419b760ed6b8321d2b3797ba1aeb83). This could be done when rebasing.

vaszig:
After discussing this with @sergei-maertens via slack we decided-found the following:
- from functional perspective, we need to also connect the zaak with the object (`ZAAKOBJECT`)
- only `content_json` is needed in the `ZGWConfig` singleton
- a new issue should be created for improving the validation concerning `objecttype` and `objecttype_version` - the last one is required in the ObjectsAPI so if the `objecttype` is provided, then we should check for the `objecttype_version` as well

vaszig:
Apart from what has been discussed with Sergei and mentioned above I have combined Silvia's PR (https://github.com/open-formulieren/open-forms/pull/3658/files#diff-bebf5c215247eab992b166baedd0e3dc1f419b760ed6b8321d2b3797ba1aeb83) with my `helpers.py` as well, according to Victorien's suggestion.

src/openforms/contrib/zgw/clients/zaken.py
```diff
@@ -130,3 +130,22 @@ def create_rol(
         response.raise_for_status()
 
         return response.json()
+
+    def create_zaakobject(
+        self, zaak: dict, object: str, objecttype_version: str
+    ) -> dict | None:
```
sergei-maertens, line 136:
```suggestion
    ) -> dict:
```
I don't see a code path that can return `None`

src/openforms/registrations/contrib/objects_api/helpers.py
sergei-maertens, line 1: A refactor has landed in `master`, you can now use `openforms.contrib.objects_api` utilities instead of these modules. Please rebase!

src/openforms/registrations/contrib/objects_api/plugin.py
sergei-maertens, line 1: This refactor was done separately, so can probably be rebased away.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -401,6 +438,31 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
         if submission.has_registrator:
             result["medewerker_rol"] = medewerker_rol
 
+        # Register submission to Objects API if configured
+        if (
+            options.get("objecttype")
+            and options.get("objecttype_version")
+            and options.get("content_json")
+        ):
```
sergei-maertens, line 446:
```suggestion
        if (
            (object_type := options.get("objecttype"))
            and (object_type_version := options.get("objecttype_version"))
            and (content_json := options.get("content_json"))
        ):
```
and now you can use those new variables inside the `if`-block, so you no longer need to repeat the `options.get(...)` calls :)

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -401,6 +438,31 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
+            result["objects_api"] = self.register_submission_to_objects_api(
```
sergei-maertens, line 447:
```suggestion
            result["objects_api_object"] = self.register_submission_to_objects_api(
```
I'd use the same nomenclature as the Objects API itself so that looking at the data in isolation makes a bit more sense :)

sergei-maertens, line 447: I'd wrap this entire call into `execute_unless_result_exists` so you can avoid doing a bunch of work like templating out the JSON again if it's already registered.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -401,6 +438,31 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
+                / str(options.get("objecttype_version"))
+            )
+
+            _create_zaakobject = zaken_client.create_zaakobject(
+                zaak,
+                result["objects_api"]["url"],
+                objecttype_version.url,
+            )
+
+            result["zaakobject"] = _create_zaakobject
```
sergei-maertens, line 464: this also needs wrapping in `execute_unless_result_exists`

src/openforms/registrations/contrib/objects_api/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def send_data_to_api(object_data: dict) -> dict:
+    with get_objects_client() as objects_client:
+        response = objects_client.post(
+            "objects", json=object_data, headers={"Content-Crs": "EPSG:4326"}
+        )
+        response.raise_for_status()
+
+    return response.json()
```
sergei-maertens, line 39: I think it's neater to define a client subclass in `openforms.contrib.objects_api.client` which is returned by `get_objects_client()` so that you can then call:

```py
client.create_object(object_data)
```
& possibly even break out some parameters into their own arguments so you can enforce that all the necessary information is provided, like the object type URL + version. The peculiarities of the record/object data/shape are then encapsulated on the client, and the objects API/ZGW backend don't need to bother with that, they can just pass along the record data and not worry about other things.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -444,3 +506,37 @@ def get_config_actions(self) -> list[Action]:
...
+                "public_reference": submission.public_registration_reference,
+                "kenmerk": str(submission.uuid),
+                "language_code": submission.language_code,
+            },
+        }
+
+        object_data = prepare_data_for_registration(
+            submission, context, options, object_mapping
+        )
+        response = send_data_to_api(object_data)
```
sergei-maertens, line 534: see earlier remarks about this `send_data_to_api` helper - it's nicer to be consistent with the rest of the client implementation here and follow an OOP approach. That also makes it obvious where the Objects API client is coming from.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3666, opened 2023-12-06
[#3662] Remove deprecation marker on `geo/address-autocomplete`

### LINKED ISSUE #3662: `geo/address-autocomplete` shouldn't be a deprecated API route

latest

`/api/v2/location/get-street-name-and-city/` is deprecated, indicating that we should use [`/api/v2/geo/address-autocomplete`](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/open-formulieren/open-forms/master/src/openapi.yaml#/geo/geo_address_autocomplete_retrieve), which is also deprecated

image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

Viicos:
Fixes #3662

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -2906,10 +2906,8 @@ paths:
   /api/v2/geo/address-autocomplete:
     get:
       operationId: geo_address_autocomplete_retrieve
-      description: |-
-        Get the street name and city for a given postal code and house number.
-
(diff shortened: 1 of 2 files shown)


sergei-maertens:
This cannot be removed yet because multiple URLs currently point to the same view.

Viicos:
Ah, this could be fixed by trying to use a separate view in a smart way but not worth spending time on this

PR #3667, opened 2023-12-06
Add typecheck to "action sets the value"

### LINKED ISSUE #3598: Add typecheck to "action sets the value" JsonLogic
Using a context from #3597 this will partially close #2707.

Even without a context it would give immediate feedback about malformed expressions *before* saving the form. Possibly on change the moment it contains a JSON.parse-able string. Otherwise on blur.

CharString:
Depends on release containing open-formulieren/formio-builder#69

Closes #3598


sergei-maertens:
Cleaning up clutter from open PRs, sorry!

PR #3670, opened 2023-12-07
[#3647] Prevent invalid time from crashing form

### LINKED ISSUE #3647: Form crash when invalid time is entered by user

2.4.1

DH 624

ValueError - Could not parse time 'Invalid date'
image

if validation is set (min max) the form does not crash:
image

Add a time component to a form
enter: 25:00
form crashes

SilviaAmAm:
Fixes #3647

### DIFF
--- a/src/openforms/submissions/models/submission_value_variable.py
+++ b/src/openforms/submissions/models/submission_value_variable.py
@@ -359,7 +359,7 @@ class SubmissionValueVariable(models.Model):
 
             maybe_naive_datetime = parse_datetime(self.value)
             if maybe_naive_datetime is None:
-                return ""
+                return
 
             if timezone.is_aware(maybe_naive_datetime):
                 return maybe_naive_datetime.date()
@@ -368,14 +368,13 @@ class SubmissionValueVariable(models.Model):
         if self.value and data_type == FormVariableDataTypes.datetime:
             maybe_naive_datetime = parse_datetime(self.value)
             if maybe_naive_datetime is None:
-                return ""
+                return
 
             if timezone.is_aware(maybe_naive_datetime):
                 return maybe_naive_datetime
             return timezone.make_aware(maybe_naive_datetime)
 
         if self.value and data_type == FormVariableDataTypes.time:
-            value = parse_time(self.value)
-            return value or ""
+            return parse_time(self.value)
 
         return self.value
--- a/src/openforms/submissions/tests/form_logic/test_endpoint.py
+++ b/src/openforms/submissions/tests/form_logic/test_endpoint.py
@@ -3,6 +3,7 @@ from django.test import tag
 from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APIRequestFactory, APITestCase
+from testfixtures import LogCapture
 
 from openforms.accounts.tests.factories import SuperUserFactory
 from openforms.forms.tests.factories import (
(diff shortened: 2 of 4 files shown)


src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -363,7 +363,7 @@ def to_python(self) -> Any:
                     return maybe_naive_datetime.date()
                 return timezone.make_aware(maybe_naive_datetime).date()
 
-            raise ValueError(f"Could not parse date '{self.value}'")  # pragma: nocover
+            return ""
```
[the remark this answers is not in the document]
SilviaAmAm: Currently the parsing method is doing the logging (level info) but I can move it out if you prefer (and change it to warning?)?

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -358,28 +358,24 @@ def to_python(self) -> Any:
...
-                if timezone.is_aware(maybe_naive_datetime):
-                    return maybe_naive_datetime.date()
-                return timezone.make_aware(maybe_naive_datetime).date()
+            if maybe_naive_datetime is None:
+                return ""
 
-            raise ValueError(f"Could not parse date '{self.value}'")  # pragma: nocover
+            if timezone.is_aware(maybe_naive_datetime):
+                return maybe_naive_datetime.date()
+            return timezone.make_aware(maybe_naive_datetime).date()
```
SilviaAmAm, line 366: I just moved this line, but there was no test for it. Although I think this line may be unreacheable, because if a field of type "date" gets a datetime string in isoformat without timezone, then the `parse_date` on L355 raises an error.

src/openforms/utils/date.py
```diff
@@ -31,3 +35,19 @@ def parse_date(value: str) -> date:
     dt = datetime.fromisoformat(value)
     assert dt.tzinfo is not None, "Expected the input variable to be timezone aware!"
     return timezone.localtime(value=dt).date()
+
+
+def parse_datetime(value: str) -> None | datetime:
+    try:
+        return _parse_datetime(value)
+    except ValueError:
+        logger.info("Can't parse datetime '%s', using empty string.", value)
```
sergei-maertens, line 44: the "using empty string" part is confusing, because you actually return `None`, and this function itself does not know how the caller will handle this `None` value.

```suggestion
        logger.info("Can't parse datetime '%s', falling back to 'None' instead.", value)
```

src/openforms/submissions/tests/form_logic/test_endpoint.py
```diff
@@ -195,3 +197,45 @@ def test_updating_data_marks_step_as_applicable_again(self):
...
+            "api:submission-steps-logic-check",
+            kwargs={
+                "submission_uuid": submission.uuid,
+                "step_uuid": submission.submissionstep_set.first().form_step.uuid,
+            },
+        )
+        self._add_submission_to_session(submission)
+
+        with (self.subTest("Invalid time"), self.assertLogs(level="INFO") as logs):
+            self.client.post(endpoint, data={"data": {"time": "25:00"}})
```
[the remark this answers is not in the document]
SilviaAmAm: With the frontend fix, it shouldn't arrive anymore, but I can add a test for it anyway.

src/openforms/submissions/models/submission_value_variable.py
sergei-maertens: theory holds up with the coverage on master branch: https://app.codecov.io/gh/open-formulieren/open-forms/blob/master/src%2Fopenforms%2Fsubmissions%2Fmodels%2Fsubmission_value_variable.py#L364

src/openforms/submissions/tests/form_logic/test_endpoint.py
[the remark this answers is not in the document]
sergei-maertens: yeah I can't tell (can we check Sentry?) whether the 25:00 caused the server side `ValueError` or whether that string `Invalid date` got sent to the server.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -358,28 +358,23 @@ def to_python(self) -> Any:
                 return aware_date.date()
 
             maybe_naive_datetime = parse_datetime(self.value)
-            if maybe_naive_datetime is not None:
-                if timezone.is_aware(maybe_naive_datetime):
-                    return maybe_naive_datetime.date()
-                return timezone.make_aware(maybe_naive_datetime).date()
+            if maybe_naive_datetime is None:
+                return
```
sergei-maertens, line 362: (don't change this, but take it into account for the future)

generally, if we're deliberately returning `None` I prefer to make that explicit: `return None`, while a bare `return` is more intended as "stop with whatever we're doing and we don't particularly care about the exact return value".

The end result/return value is identical, but there's some different meaning/implication in the exact statement used.

sergei-maertens (approved):

sergei-maertens:
Backports:

* 4405d69047016c9c12b953952a27f584d5c9b67d

I do not find this critical enough to also backport to older versions. The backport is on specific request of DH for this particular issue.

PR #3673, opened 2023-12-08
[#3649] Add bestandsomvang field

### LINKED ISSUE #3649: Battletest the ZGW API-connection between Rxmission and Open Formulieren
### Thema / Theme

Other

There are few unexpected issues when using Rxmission as ZGW API registration backend.

Where we typically read the OAS to adjust our calls to the backend, this concept came unexpected to Visma Roxit. The OAS was not readily available for retrieval. After manually adding the OAS, some other issues came to light, like passing an API-version header attribute `bestandsomvang` that didn't match the upload size (for documents).

These issues needs to be tracked, analysed and fixed. Since the ZGW API in Rxmission is also subject to changes, we might ask them to fix things or workaround issues for now.

Harden the ZGW API registration backend to be used in more infra setups.

No claims are made who is correct or not, we just want things to work :)

### Tasks

* [ ] Add version selection option to `Service`. This value can be empty. If filled, it should have an effect on the `API-version` header that is sent.
* [ ] Figure out why the error doesn't show up in request-logging

By setting API-version to 1.0 for Documenten API, this should work on Rxmission.

SilviaAmAm:
Fixes #3649 

For compatibility with the RXMission API (for which this field is required and `null` values are not accepted).

### DIFF
--- a/docs/configuration/registration/zgw.rst
+++ b/docs/configuration/registration/zgw.rst
@@ -35,7 +35,7 @@ OAS                           URL to the OAS.
 Client ID                     Client ID for the JWT-token.
 Secret                        Secret for the JWT-token.
 **Documenten API**
-API root URL                  Root URL for the Documenten API that Open Forms can access.
+API root URL                  Root URL for the Documenten API (version 1.1+) that Open Forms can access.
 OAS                           URL to the OAS.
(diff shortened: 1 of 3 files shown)


sergei-maertens (changes requested):
Please add a test :)

PR #3686, opened 2023-12-13
Upgrade to NodeJS LTS 20

sergei-maertens:
Node 16 has been end-of-life since a couple of months (sept. 2023), so it's time.


vaszig (approved):
Is there something we should do or worry about concerning the other libraries and the new version-upgrade of node?

sergei-maertens:
> Is there something we should do or worry about concerning the other libraries and the new version-upgrade of node?

I don't think so, otherwise there's something very wrong with our CI pipeline passing :P

You may need to `rm -rf node_modules` if you run into any issues, but since the `package-lock.json` was not changed I don't expect issues.

PR #3687, opened 2023-12-13
[#3656] OIDC-DigiD return proper message after login cancellation

### LINKED ISSUE #3656: Digid-oidc: Return after cancel not according to DigiD requirements

2.4.1

Digid-oidc (and eHerkenning) has no route for cancelled logins. The user is always shown a "generic" error screen which in the case of DigiD does not meet the requirement:

_Als de gebruiker het authenticatieproces annuleert (SAML-statuscode "AuthnFailed" of CGI-statuscode 0040), komt de gebruiker terug op de inlogpagina vanwaar getracht is de authenticatie te starten. Dit gebeurt in hetzelfde browserscherm. Er dient een melding getoond te worden met de mededeling dat het inloggen is geannuleerd._

digid-oidc
image
Digid
image

Digid-oidc does not return to the login page, and does not mention the cancellation

vaszig:
Fixes #3656 

- After investigation we found out (@SilviaAmAm helped with debugging) that OIDC returns an error (required) and an error_description (optional) as query parameters `error=access_denied&error_description=The user cancelled` . 
That's why we explicitly return a message according to DigiD and Logius specs instead of the generic one.

References:

- https://openid.net/specs/openid-connect-core-1_0.html#AuthError
- https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1.2.1

In order to test this you will have to set up DigiD-OIDC in the admin so let me know and I will provide you with the needed settings.

### DIFF
--- a/src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
+++ b/src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
@@ -22,6 +22,7 @@ from openforms.authentication.contrib.digid.views import (
     LOGIN_CANCELLED,
 )
 from openforms.authentication.contrib.eherkenning.views import MESSAGE_PARAMETER
+
 from ...views import BACKEND_OUTAGE_RESPONSE_PARAMETER
 from .backends import (
     OIDCAuthenticationDigiDBackend,
--- a/src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
+++ b/src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
@@ -17,7 +17,11 @@ from digid_eherkenning_oidc_generics.views import (
     OIDCAuthenticationCallbackView as _OIDCAuthenticationCallbackView,
     OIDCAuthenticationRequestView as _OIDCAuthenticationRequestView,
 )
-
+from openforms.authentication.contrib.digid.views import (
+    DIGID_MESSAGE_PARAMETER,
+    LOGIN_CANCELLED,
+)
+from openforms.authentication.contrib.eherkenning.views import MESSAGE_PARAMETER
 from ...views import BACKEND_OUTAGE_RESPONSE_PARAMETER
 from .backends import (
     OIDCAuthenticationDigiDBackend,
@@ -61,6 +65,17 @@ class OIDCAuthenticationRequestView(_OIDCAuthenticationRequestView):
 
 
 class OIDCAuthenticationCallbackView(_OIDCAuthenticationCallbackView):
+    def get_error_message_parameters(
+        self, parameter: str, problem_code: str
+    ) -> tuple[str, str]:
+        """
+        Return a tuple of the parameter type and the problem code.
+        """
+        return (
+            parameter or BACKEND_OUTAGE_RESPONSE_PARAMETER,
+            problem_code or self.plugin_identifier,
+        )
+
     @property
     def failure_url(self):
         """
@@ -68,7 +83,13 @@ class OIDCAuthenticationCallbackView(_OIDCAuthenticationCallbackView):
         """
         f = furl(self.success_url)
         f = furl(f.args["next"])
-        f.args[BACKEND_OUTAGE_RESPONSE_PARAMETER] = self.plugin_identifier
+
+        parameter, problem_code = self.get_error_message_parameters(
+            self.request.GET.get("error", ""),
+            self.request.GET.get("error_description", ""),
+        )
+        f.args[parameter] = problem_code
+
         return f.url
 
 
@@ -84,6 +105,14 @@ class DigiDOIDCAuthenticationCallbackView(
     plugin_identifier = "digid_oidc"
     auth_backend_class = OIDCAuthenticationDigiDBackend
 
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if error == "access_denied" and error_description == "The user cancelled":
+            return (DIGID_MESSAGE_PARAMETER, LOGIN_CANCELLED)
+
+        return (BACKEND_OUTAGE_RESPONSE_PARAMETER, self.plugin_identifier)
+
 
 class eHerkenningOIDCAuthenticationRequestView(
     SoloConfigEHerkenningMixin, OIDCAuthenticationRequestView
@@ -96,6 +125,17 @@ class eHerkenningOIDCAuthenticationCallbackView(
 ):
     plugin_identifier = "eherkenning_oidc"
     auth_backend_class = OIDCAuthenticationEHerkenningBackend
+
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if error == "access_denied" and error_description == "The user cancelled":
+            return (
+                MESSAGE_PARAMETER % {"plugin_id": self.plugin_identifier.split("_")[0]},
+                LOGIN_CANCELLED,
+            )
+
+        return (BACKEND_OUTAGE_RESPONSE_PARAMETER, self.plugin_identifier)
 
 
 class DigiDMachtigenOIDCAuthenticationRequestView(
@@ -110,6 +150,14 @@ class DigiDMachtigenOIDCAuthenticationCallbackView(
     plugin_identifier = "digid_machtigen_oidc"
     auth_backend_class = OIDCAuthenticationDigiDMachtigenBackend
 
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if error == "access_denied" and error_description == "The user cancelled":
+            return (DIGID_MESSAGE_PARAMETER, LOGIN_CANCELLED)
+
+        return (BACKEND_OUTAGE_RESPONSE_PARAMETER, self.plugin_identifier)
+
 
 class EHerkenningBewindvoeringOIDCAuthenticationRequestView(
     SoloConfigEHerkenningBewindvoeringMixin, OIDCAuthenticationRequestView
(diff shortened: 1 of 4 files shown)


src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -68,7 +72,19 @@ def failure_url(self):
...
+
+        # When the user cancels login we want an explicit error for DigiD authentication
+        # according to the spec.
+        if (
+            oidc_error := self.request.GET.get("error")
+        ) and oidc_error == "access_denied":
+            if (
+                oidc_error_description := self.request.GET.get("error_description")
+            ) and oidc_error_description == "The user cancelled":
+                f.args[DIGID_MESSAGE_PARAMETER] = LOGIN_CANCELLED
```
sergei-maertens, line 84:
```suggestion
        if self.request.GET.get("error") == "access_denied" and self.request.GET.get("error_description") == "The user cancelled":
            f.args[DIGID_MESSAGE_PARAMETER] = LOGIN_CANCELLED
```

Though it also seems wrong to set the digid message when the user is logging in with eHerkenning, which shares this base class.

sergei-maertens (changes requested):

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -68,7 +72,18 @@ def failure_url(self):
...
-        f.args[BACKEND_OUTAGE_RESPONSE_PARAMETER] = self.plugin_identifier
+
+        # When the user cancels login we want an explicit error for DigiD authentication
+        # according to the spec.
+        if (
+            self.request.GET.get("error") == "access_denied"
+            and self.request.GET.get("error_description") == "The user cancelled"
+            and self.plugin_identifier in ["digid_oidc", "digid_machtigen_oidc"]
+        ):
+            f.args[DIGID_MESSAGE_PARAMETER] = LOGIN_CANCELLED
```
sergei-maertens, line 83: I think it would be cleaner separation of concerns to implement a `def get_error_message_parameters(self, error: str, error_description: str):` method and implement the DigiD check in the subclasses.

It doesn't sit right with me that subclass-specific behaviour is implemented in the base class, this means you can never alter the behaviour from a third-party plugin because you need to modify the base class behaviour.

Moving this to a custom/helper method improves loose coupling instead of the current tight coupling implementation.

sergei-maertens (changes requested):

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -61,14 +62,28 @@ def get(self, request):
...
         """
         On failure, redirect to the form with an appropriate error message
         """
         f = furl(self.success_url)
         f = furl(f.args["next"])
-        f.args[BACKEND_OUTAGE_RESPONSE_PARAMETER] = self.plugin_identifier
+
+        parameter, problem_code = self.get_error_message_parameters(
+            self.request.GET.get("error"), self.request.GET.get("error_description")
+        )
```
sergei-maertens, line 84:
```suggestion
        parameter, problem_code = self.get_error_message_parameters(
            self.request.GET.get("error", ""), self.request.GET.get("error_description", "")
        )
```
for type safety :) a bare `.get` returns `None` if the params are not present.

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -61,14 +62,28 @@ def get(self, request):
...
 
 class OIDCAuthenticationCallbackView(_OIDCAuthenticationCallbackView):
+    def get_error_message_parameters(
+        self, parameter: str, problem_code: str
+    ) -> tuple[str, str]:
+        """
+        Return a tuple of the parameter type and the problem code.
+        """
+
+        return (parameter, problem_code)
```
sergei-maertens, line 72:
```suggestion
        return (parameter or BACKEND_OUTAGE_RESPONSE_PARAMETER, problem_code or self.plugin_identifier)
```
probably should have this as fallback behaviour in case the error/description are empty?

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -84,6 +99,18 @@ class DigiDOIDCAuthenticationCallbackView(
...
     auth_backend_class = OIDCAuthenticationDigiDBackend
 
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if (
+            error == "access_denied"
+            and error_description
+            and error_description == "The user cancelled"
+        ):
```
sergei-maertens, line 109:
```suggestion
        if (
            error == "access_denied"
            and error_description == "The user cancelled"
        ):
```

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -84,6 +99,18 @@ class DigiDOIDCAuthenticationCallbackView(
...
+            return ("_digid-message", LOGIN_CANCELLED)
```
sergei-maertens, line 110:
```suggestion
            return (DIGID_MESSAGE_PARAMETER, LOGIN_CANCELLED)
```
re-use the constant?

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -97,6 +124,18 @@ class eHerkenningOIDCAuthenticationCallbackView(
...
     auth_backend_class = OIDCAuthenticationEHerkenningBackend
 
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if (
+            error == "access_denied"
+            and error_description
+            and error_description == "The user cancelled"
+        ):
```
sergei-maertens, line 134:
```suggestion
        if (
            error == "access_denied"
            and error_description == "The user cancelled"
        ):
```

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -97,6 +124,18 @@ class eHerkenningOIDCAuthenticationCallbackView(
...
+            return ("_eherkenning-message", LOGIN_CANCELLED)
```
sergei-maertens, line 135: this could also be eidas instead of eherkenning, see `MESSAGE_PARAMETER` in the eherkenning contrib views :thinking: 

I don't know if those get special treatment in the frontend though.

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -110,6 +149,18 @@ class DigiDMachtigenOIDCAuthenticationCallbackView(
...
     auth_backend_class = OIDCAuthenticationDigiDMachtigenBackend
 
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if (
+            error == "access_denied"
+            and error_description
+            and error_description == "The user cancelled"
+        ):
```
sergei-maertens, line 159:
```suggestion
        if (
            error == "access_denied"
            and error_description == "The user cancelled"
        ):
```

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -122,3 +173,15 @@ class EHerkenningBewindvoeringOIDCAuthenticationCallbackView(
...
     auth_backend_class = OIDCAuthenticationEHerkenningBewindvoeringBackend
+
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if (
+            error == "access_denied"
+            and error_description
+            and error_description == "The user cancelled"
+        ):
```
sergei-maertens, line 184:
```suggestion
        if (
            error == "access_denied"
            and error_description == "The user cancelled"
        ):
```

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -110,6 +149,18 @@ class DigiDMachtigenOIDCAuthenticationCallbackView(
...
 
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if (
+            error == "access_denied"
+            and error_description
+            and error_description == "The user cancelled"
+        ):
+            return ("_digid-message", LOGIN_CANCELLED)
```
sergei-maertens, line 160: see earlier

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -122,3 +173,15 @@ class EHerkenningBewindvoeringOIDCAuthenticationCallbackView(
...
+
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if (
+            error == "access_denied"
+            and error_description
+            and error_description == "The user cancelled"
+        ):
+            return ("_eherkenning-message", LOGIN_CANCELLED)
```
sergei-maertens, line 185: see earlier

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -122,3 +171,14 @@ class EHerkenningBewindvoeringOIDCAuthenticationCallbackView(
 ):
     plugin_identifier = "eherkenning_bewindvoering_oidc"
     auth_backend_class = OIDCAuthenticationEHerkenningBewindvoeringBackend
+
+    def get_error_message_parameters(
+        self, error: str, error_description: str
+    ) -> tuple[str, str]:
+        if error == "access_denied" and error_description == "The user cancelled":
+            return (
+                MESSAGE_PARAMETER % {"plugin_id": self.plugin_identifier.split("_")[0]},
```
Viicos, line 180: Maybe later in another PR would be nice to switch to `.format`

sergei-maertens (approved):

sergei-maertens:
Backports: 03a79fdf8008b10576cacc2a086edd95a4b15d07, 0443f03fe5fae605161a59f9392da05c075923ef, 12bdad3bc285fe4590acf0c8b88232cbe9218751 and 84af07e78eb6bfd834e41913cece853faa626f13

PR #3696, opened 2023-12-15
[#3692] Fix next url when authenticated user cancels login

### LINKED ISSUE #3692: DigiD-OIDC does not redirect to form if the user is authenticated and cancels the login

latest

When a user is authenticated, for example an admin user is logged in and clicks on a form which uses DigiD-OIDC authentication, if he cancels the login flow then the site crashes (`KeyError` - there is no `next-url`).

Sentry: 336625

1. Create a form which uses DigiD-OIDC authentication
2. Click on _show form_ and then on the startpage of the form, try to log in
3. Being on the selection page (DigiD, Eherkenning etc) click cancel

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Fixes #3692  

Based on #3656 so it needs first to be merged.

### DIFF
--- a/src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
+++ b/src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
@@ -43,6 +43,10 @@ class OIDCAuthenticationRequestView(_OIDCAuthenticationRequestView):
         if not next_url:
             raise DisallowedRedirect
 
+        # We add our own key to keep track of the redirect URL. In the case of authentication failure (or canceled logins),
+        # the session is cleared, so in OIDCAuthenticationCallbackView we store this URL so that we know where to.
+        self.request.session["of_redirect_next"] = next_url
+
         try:
             # Verify that the identity provider endpoint can be reached
             response = requests.get(self.OIDC_OP_AUTH_ENDPOINT)
@@ -66,6 +70,10 @@ class OIDCAuthenticationRequestView(_OIDCAuthenticationRequestView):
 
 
 class OIDCAuthenticationCallbackView(_OIDCAuthenticationCallbackView):
+    def get(self, request):
+        self._redirect_next = request.session.get("of_redirect_next")
+        return super().get(request)
+
     def get_error_message_parameters(
         self, parameter: str, problem_code: str
     ) -> tuple[str, str]:
@@ -82,7 +90,7 @@ class OIDCAuthenticationCallbackView(_OIDCAuthenticationCallbackView):
         """
         On failure, redirect to the form with an appropriate error message
         """
-        f = furl(self.success_url)
+        f = furl(self._redirect_next or self.get_settings("LOGIN_REDIRECT_URL", "/"))
         f = furl(f.args["next"])
 
         parameter, problem_code = self.get_error_message_parameters(
--- a/src/openforms/authentication/contrib/digid_eherkenning_oidc/tests/digid/test_auth_procedure.py
+++ b/src/openforms/authentication/contrib/digid_eherkenning_oidc/tests/digid/test_auth_procedure.py
@@ -8,6 +8,7 @@ from furl import furl
 from rest_framework import status
 
 from digid_eherkenning_oidc_generics.models import OpenIDConnectPublicConfig
+from openforms.accounts.tests.factories import StaffUserFactory
 from openforms.authentication.views import BACKEND_OUTAGE_RESPONSE_PARAMETER
 from openforms.forms.tests.factories import FormFactory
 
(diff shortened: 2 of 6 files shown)


src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -43,6 +43,11 @@ def get(self, request):
         if not next_url:
             raise DisallowedRedirect
 
+        # This is needed for the next steps in order to redirect the user to the right place again.
+        # When a user is authenticated and cancels the login flow, the next url (oidc_login_next)
+        # has been removed from the session (the user is logged out-session flushed).
+        self.request.session["of_redirect_next"] = next_url
```
Viicos, line 49: I'm not a 100% certain, but: shouldn't session be saved?

vaszig: Do you mean generally?If so, django does this when the session is modified (afaik). Let me know if you mean something like that [https://docs.djangoproject.com/en/5.0/topics/http/sessions/#when-sessions-are-saved](url) or if you mean something else.

vaszig: Yes indeed. In tests I have to save it as you mention. Maybe @sergei-maertens  can provide us with a more clear answer.

sergei-maertens: the remarks above are correct :)

SilviaAmAm, line 49: Is it necessary to have this extra param in the session? The overwritten `get` of the `OIDCAuthenticationCallbackView` would still have `oidc_login_next` added by the `OIDCAuthenticationRequestView` (https://github.com/mozilla/mozilla-django-oidc/blob/1.2.4/mozilla_django_oidc/views.py#L179) when creating the `self._redirect_next`?

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -93,6 +102,17 @@ def failure_url(self):
...
         return f.url
 
+    @property
+    def success_url(self):
+        # Pull the next url from the session or settings--we don't need to
+        # sanitize here because it should already have been sanitized.
+        next_url = self.request.session.get("oidc_login_next", None)
+        return (
+            next_url
+            or self._redirect_next
```
SilviaAmAm, line 112: This should be needed only in the case of a failed login right? So maybe it makes more sense to keep it in the `failure_url` property?

Also, this issue in the Mozilla package seems relevant: https://github.com/mozilla/mozilla-django-oidc/issues/458, maybe we could mention it in a comment?

sergei-maertens (commented):
Had some pending comments!

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -43,6 +43,11 @@ def get(self, request):
         if not next_url:
             raise DisallowedRedirect
 
+        # This is needed for the next steps in order to redirect the user to the right place again.
+        # When a user is authenticated and cancels the login flow, the next url (oidc_login_next)
```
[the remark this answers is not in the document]
sergei-maertens: IMO being logged out of the admin is not an issue - I also don't see it reported in the ticket description as being an issue, only the `KeyError` started showing up in Sentry because of information being lost.

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
sergei-maertens: yes, I specifically requested this :laughing: 

The reasoning is that *we* should not overload the `oidc_login_next` parameter meaning, it's actually wrong in the first place (IMO) that the failure URL makes use of the `self.success_url`. This is particularly dangerous if the upstream library decides to use this session key in other security sensitive areas (hypothetical example - checking if someone is considered "logged in" by checking the presence of this key).

Either way, this particular session key is private API of the mozilla-django-oidc library and we should *not* (ab)use that, the safe approach is adding our own semantics that are similar, but isolated.

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
sergei-maertens: I think something was lost in communication here - IMO the `success_url` should not be touched, instead the `failure_url` property should be updated to make use of `self._redirect_next` instead of relying on `self.success_url`.

It feels wrong and dangerous to mix success/failure behaviours.

sergei-maertens (approved):
I'll rebase the commits so that it's easier to cherry pick for backports!

sergei-maertens:
Backports:

* stable/2.4.x: a700a2aa2fa94f3a7f68072979ee43b3683f0727
* stable/2.3.x: 5ee9426cff14c4c0626803b7c8d7dab7d4fabdc6
* stable/2.2.x: 05d382319ba4ccdbca2149daae27d11a19b4fcad
* stable/2.1.x: c94654a6c3eead284123b1c38a78c229eebeb9e1

PR #3698, opened 2023-12-15
Add `typings/` as an entry to `.gitignore`

Viicos:
I have a local copy of django-stubs that is used by pyright


sergei-maertens (commented):
For these very specific cases, perhaps it's better to maintain a local git ignore configuration? https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

Viicos:
Didn't know it was possible, indeed easier this way

PR #3699, opened 2023-12-16
Add missing `FormsExport` import

Viicos:
I don't really know how Django was still able to take this model into account, but according to docs:

https://docs.djangoproject.com/en/5.0/topics/db/models/#organizing-models-in-a-package

> You must import the models in the `__init__.py` file.


sergei-maertens (approved):

sergei-maertens:
It worked because the `form.py` file gets imported anyway and then it's present in django's app registry.

PR #3710, opened 2023-12-22
Migrate formio component translations

### LINKED ISSUE #2958: All translation disapear after adding/modifying components in repeating group & field group

2.1.x
Utrecht, Venray

Translating some text of a selectboxes component:
image
- Save translations
- Save the forms
- reload form
Translations are saved correctly

In the same step:

- Add repeating group, or add any field to a repeating group in the same step:
image
- Save form
- Open "selctboxes" with translation:

image
All translations are gone

### LINKED ISSUE #3466: translations: Cannot save translation in different "layers" (groups)

2.3.0

utr: 116
hadm: 70

Translating a form where certain components are part of a group, saving translations for these components seem to remove all other translations in the same form step. 
This seems to only work 1 way > Adding translation to the components within groups. 
When you add translations to the non nested components - all translations survive.

Create form with translations
create a step
Add textfield1 to step
add fieldgroup1 to step
add textfield2 to fieldgroup1

Translate textfield1 + textfield2 save = texdfield1 lost translation

sergei-maertens:
Closes #2958 - blocked until we fully enable the new formio builder (!)
Closes #3466 
Depends on #3720 

This sets up a data migration and converter utility to map the translations from the old format to the new.

- [x] Implement converter
- [x] Use converter in data migration
- [x] Use converter during form import

Issues to fix

- [x] Licenseplace with missing `validate` key crashes in preview -> add data migration + conversion to ensure it's there
- [x] Crashes because of missing `dataSrc`
- [x] Localization in tests


src/openforms/forms/fd_translations_converter.py
```diff
@@ -0,0 +1,176 @@
+"""
+Implements the translation converter for form definitions.
+
+Github issue: https://github.com/open-formulieren/open-forms/issues/2958
+
+Before, translations for formio components were saved in the JSONField
+``FormDefinition.component_translations``, which is a key-value mapping of language
+code to translations mapping, which itself is a key-value mapping of source string to
+translation string. This had the drawback that if different components happen to use the
+same literal, they also shared the same translation.
```
Viicos, line 10: So this means that if two components were defined from a reusable form definition, they couldn't have custom translations?

src/openforms/forms/fd_translations_converter.py
```diff
@@ -0,0 +1,176 @@
...
+    translations_from_store: dict[str, str],
+    locale: str,
+    fields: list[str],
+) -> None:
+    translations = {
+        key: translation
+        for key in fields
+        if (literal := obj.get(key))
+        and (translation := translations_from_store.get(literal))
+    }
```
Viicos, line 45: code is read more often than it is written, I find it hard to know what the resulting `translations` dict is. But fine to keep it as is

src/openforms/forms/fd_translations_converter.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _move_translations(component: Component, locale: str, translations: dict[str, str]):
+    """
+    Given a component and translation store, mutate the component to bake in translations.
+
+    This mutates the ``component`` parameter!
+    """
+    assert "type" in component
+
+    match component:
```
Viicos, line 59: Any way we could have the translatable fields set somewhere so that they can be retrieved for later use? Maybe not that relevant if this is the only place we will make use of them

src/openforms/forms/fd_translations_converter.py
sergei-maertens: yeah it's a side-effect of bad data structures that we're resolving :(

But it's basically making sure that the translation is grabbed from the `[literal]: [translation]` mapping if the literal is not empty-ish, and assigns it to the field/property where the literal was found (`label`, `description` etc...).

The frontend code for this is horrible :sweat_smile:

src/openforms/forms/fd_translations_converter.py
sergei-maertens: Couldn't and can't - as discussed on Slack, a FD is reusable as a whole or not at all. Translations are baked in and not contextualized in a form - which is as intended.

src/openforms/forms/fd_translations_converter.py
sergei-maertens: This is the only place - the frontend types basically define what is and isn't translatable and that should be sufficient.

The data structure of translations is crucial, since it is processed in a generic way in the backend, see: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/registry.py#L170

For components that have additional translations support, there is a hook, e.g.: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/components/vanilla.py#L178

src/openforms/formio/migration_converters.py
```diff
@@ -76,6 +78,48 @@ def set_openforms_datasrc(component: Component) -> bool:
...
+
+        if size_ok and size_mobile_ok:
+            continue
+
+        changed = True
+        if not size_ok:
+            try:
+                column["size"] = int(size)
+            except (TypeError, ValueError):
+                column["size"] = 6
```
Viicos, line 100: Would be nice to see what was set initially in this case. I'm wondering what could have happened to have an non integer value set. I guess the bin script in the other PR will reveal

src/openforms/forms/api/serializers/form_definition.py
```diff
@@ -131,6 +134,21 @@ def validate(self, attrs):
...
+
+        # during import, process legacy format component translations
+        if self.context.get("is_import", False) and (
+            translations_store := attrs.get("component_translations")
+        ):
+            warnings.warn(
+                "Form-definition component translations are deprecated, the "
+                "compatibility layer wil be removed in Open Forms 3.0.",
+                DeprecationWarning,
+            )
```
Viicos, line 146: Maybe set a relevant `stacklevel` so that we can actually see where the serializer has been used (not sure if this is possible, this `validate` method is probably nested under a lot of calls)

src/openforms/forms/tests/test_translations_converter.py
```diff
@@ -0,0 +1,263 @@
...
+    def test_any_component_is_properly_processed(self, component):
+        do_processing(component)
+
+        self.assertLabelsTranslated(component)
+
+    @given(textfield_component())
+    def test_textfield_component(self, component: Component):
+        do_processing(component)
+
+        self.assertLabelsTranslated(component)
```
Viicos, line 89: Couldn't this be done in a reusable way, maybe using `subTest`?

Viicos (approved):

src/openforms/formio/migration_converters.py
sergei-maertens: I considered logging, but nobody is actually going to look at that and do anything with it. The detection script can we wired up as an upgrade check so that upgrading is blocked until those issues are resolved, but in this case it would cause run-time issues anyway so I opted to silently fix it instead.

I don't see a way without excessive effort to nicely package and format the information and get it delivered to the people actually building/maintaining the forms themselves.

src/openforms/forms/api/serializers/form_definition.py
sergei-maertens: I don't think this has any added value, since the people importing legacy forms will not see this warning anyway. It's mostly a reminder for ourselves to remove this shim in Open Forms 3.0.

src/openforms/forms/tests/test_translations_converter.py
sergei-maertens: I wanted:

* a guarantee that every supported component type goes through CI
* possible combinations of optional/required fields for each component type does not cause problems

The second bullet is not exhaustive yet, but using hypothesis we can at least get closer to that goal. I think I tried using hypothesis with subtest and it didn't work, I'm seeing some more info [here](https://github.com/HypothesisWorks/hypothesis/issues/1071#issuecomment-357687573) but it's not conclusive whether subtest is supported or not.

That being said, debugging individual failures with 30 subtests is a pain, while I can now run `src/manage.py test src -k test_textfield_component` to only run that test. I find the DX of this approach much better and easier to follow.

src/openforms/formio/migration_converters.py
```diff
@@ -76,6 +78,68 @@ def set_openforms_datasrc(component: Component) -> bool:
     return True
 
 
+def fix_column_sizes(component: Component) -> bool:
+    component = cast(ColumnsComponent, component)
```
sergei-maertens, line 82: can't `isinstance` on a typed dict - normally I'd `# type: ignore` this but there's quite a few more linter warnings :/

PR #3713, opened 2023-12-27
:recycle: Refactor utilities for dealing with JSON templates

SilviaAmAm:
Needed by https://github.com/open-formulieren/open-forms/pull/3658 and https://github.com/open-formulieren/open-forms/pull/3663

### DIFF
--- a/src/openforms/contrib/objects_api/rendering.py
+++ b/src/openforms/contrib/objects_api/rendering.py
@@ -1,0 +1,49 @@
+import json
+import sys
+from contextlib import contextmanager
+from typing import Any
+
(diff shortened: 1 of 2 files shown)


sergei-maertens (commented):
I think I'd move the utilities themselves into `openforms.contrib.objects_api` for a more general-purpose shared utility. The directory/package still needs to be created.

I think it will also benefit in creating a particular client with methods like `Client.create_object(...)` in the future so that the objects_api/zgw registration plugins have to be aware of less objects API implementation details.

sergei-maertens (approved):

PR #3720, opened 2023-12-28
:hammer: [#2958] Define formio-component specific search strategies

sergei-maertens:
Split off from #3710

### DIFF
--- a/src/openforms/formio/tests/search_strategies.py
+++ b/src/openforms/formio/tests/search_strategies.py
@@ -1,0 +1,504 @@
+"""
+Expose hypothesis derived strategies to work with Formio.js data structures.
+
+All the formio component definitions must be JSON-serializable. JSON in itself can
+handle NULL bytes inside strings (they turn into \u0000), but JSONB as used by
+postgresql and Django's model.JSONField - where we persist these component definitions -
+cannot handle NULL bytes. That's why we opt for jsonb_text rather than the plain st.text
+strategy.
+
+The component definitions are by no means complete and it is possible hypothesis
+generates some combinations with optional fields that semantically make little sense
+(like setting both ``deriveStreetName`` and ``deriveCity`` to ``true`` in the textfield
+component). This shall have to be iterated on.
+"""
+from string import ascii_letters, digits
+
+from hypothesis import strategies as st
+
+from openforms.tests.search_strategies import jsonb_text
+
+
+def formio_key():
+    """
+    A search strategy that produces valid Formio.js key values.
+
+    Formio.js keys must start and end with an alphanumeric character. Dashes and dots
+    as separators are allowed. A value like ``foo..bar`` is valid, in the resulting
+    data structure empty strings are used as keys:
+    ``{"foo": {"": {"": {"bar": $value}}}}``
+
+    See :func:`openforms.forms.models.form_variable.variable_key_validator` for the
+    validator implementation.
+    """
+    only_alphanumeric = st.text(alphabet=ascii_letters + digits, min_size=1, max_size=1)
+    middle = st.text(max_size=100, alphabet=".-" + ascii_letters + digits)
+    # Combine them using st.tuples and then concatenate the results
+    return st.tuples(only_alphanumeric, middle, only_alphanumeric).map(
+        lambda x: "".join(x)
+    )
+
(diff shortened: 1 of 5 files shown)


src/openforms/formio/tests/search_strategies.py
```diff
@@ -0,0 +1,504 @@
...
+
+    See :func:`openforms.forms.models.form_variable.variable_key_validator` for the
+    validator implementation.
+    """
+    only_alphanumeric = st.text(alphabet=ascii_letters + digits, min_size=1, max_size=1)
+    middle = st.text(max_size=100, alphabet=".-" + ascii_letters + digits)
+    # Combine them using st.tuples and then concatenate the results
+    return st.tuples(only_alphanumeric, middle, only_alphanumeric).map(
+        lambda x: "".join(x)
+    )
```
Viicos, line 39: Maybe [`st.from_regex`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.from_regex) could be more readable? something like `st.from_regex("^[\d\w][\d\w\.-]{,100}[\d\w]$", alphabet=ascii_letters + digits + '.-')`

src/openforms/formio/tests/search_strategies.py
```diff
@@ -0,0 +1,504 @@
...
+            columns_component(),
+            fieldset_component(),
+        ]
+    deprecated = [
+        postcode_component(),
+        password_component(),
+        cosign_v1_component(),
+    ]
+    all_types = inputs + special + layout + deprecated
+    return st.one_of(*all_types)
```
Viicos, line 504:
```suggestion
    return st.one_of(all_types)
```

Can also be used

Viicos (commented):
Just a few comments.

I wonder if it would be possible to avoid defining the strategies manually if we had proper `TypedDict` definitions for the components, and then use `st.from_type`

sergei-maertens:
> I wonder if it would be possible to avoid defining the strategies manually if we had proper TypedDict definitions for the components, and then use st.from_type

I don't think so, we can't express the peculiarities of `key` in a type :(

src/openforms/formio/tests/search_strategies.py
```diff
@@ -0,0 +1,504 @@
+"""
+Expose hypothesis derived strategies to work with Formio.js data structures.
+
+All the formio component definitions must be JSON-serializable. JSON in itself can
+handle NULL bytes inside strings (they turn into \u0000), but JSONB as used by
```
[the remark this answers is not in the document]
sergei-maertens: The `\u0000` serialization? No idea, I just checked what it looks like with `json.dumps(b"\xx")` or something like that, and `\u` escape sequences are used in JSON commonly afaik.

Viicos:
> I don't think so, we can't express the peculiarities of `key` in a type :(

Would be a great use case of https://github.com/HypothesisWorks/hypothesis/pull/3780 :) (but I don't think regex is supported in annotated-types)

src/openforms/formio/tests/search_strategies.py
sergei-maertens: I don't know how that performs - one of the issues we run into that cause flakiness with timings is that filtering out bad values via `assume` or `.filter` is *after* generation, while this approach ensures that only valid values can be generated.

So, if that strategy properly generates values rather than filtering them out, that looks like a decent improvement :)

Viicos: Yeah you're right I don't really know how regex is handled on hypothesis, your solution has the benefit of being "stable"

sergei-maertens: would love to hear the perspective of @CharString here :)

sergei-maertens: I am getting 4 hits on `.filter` in https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/regex.py#L311 but checking the docs on filtering (https://hypothesis.readthedocs.io/en/latest/data.html#filtering)

I think that the combination of regex + alphabet is sufficiently simple to satisfy that recommendation. If it causes issues, we can revert back to this less-readable version.

PR #3734, opened 2024-01-04
:pencil: Improve documentation for release process

sergei-maertens:
Following some PR feedback about VCR.

### DIFF
--- a/docs/developers/releases.rst
+++ b/docs/developers/releases.rst
@@ -46,6 +46,35 @@ missing JS translations.
 Note that the translations in ``src/openforms/js/lang/formio/[locale].json`` cannot be
 automatically extracted and requires tedious manual checking and adding translations.
 
+**Re-recording the VCR cassettes**
+
+(Stable) minor and major releases must re-record the VCR cassettes to ensure that any
+feature that interacts with external APIs still works properly. Re-recording the
+cassettes forces the tests to talk to the real services again instead of replaying the
+recorded mock data, which can give a false sense of security.
+
+You may have to run subsets of the test suite separately, due to different integrations
+all configured on the same obfuscated URL (e.g. ``http://localhost:8080``), for example:
+
+.. code-block:: bash
+
+    python src/manage.py test openforms.authentication.contrib.digid
+
+.. note::
+
+    One example of this is in appointments - ``mitmproxy`` is used and exposes the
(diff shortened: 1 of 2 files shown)


docs/developers/releases.rst
```diff
@@ -46,6 +46,24 @@ missing JS translations.
 Note that the translations in ``src/openforms/js/lang/formio/[locale].json`` cannot be
 automatically extracted and requires tedious manual checking and adding translations.
 
+**Re-recording the VCR cassettes**
+
+(Stable) minor and major releases must re-record the VCR cassettes to ensure that any
+refactoring still works properly. Re-recording the cassettes forces the tests to
```
SilviaAmAm, line 52:
```suggestion
(Stable) minor and major releases must re-record the VCR cassettes to ensure that any
feature that interacts with external APIs still works properly. Re-recording the cassettes forces the tests to
```
Maybe?

docs/developers/releases.rst
```diff
@@ -46,6 +46,24 @@ missing JS translations.
...
+talk to the real services again instead of replaying the recorded mock data, which can
+give a false sense of security.
+
+You may have to run the test suite in partially, due to different integrations all configured on the same obfuscated URL (e.g. ``http://localhost:8080``), for example:
```
SilviaAmAm, line 56:
```suggestion
You may have to run subsets of the test suite separately, due to different integrations all configured on the same obfuscated URL (e.g. ``http://localhost:8080``), for example:
```
I'm not sure I understand what this means: "different integrations all configured on the same obfuscated URL"

docs/developers/releases.rst
sergei-maertens: better yeah!

docs/developers/releases.rst
sergei-maertens: the docker compose for camunda exposes `localhost:8080`, but the man-in-the-middle proxy setup for appointments and suwinet (?) also provides those services on `localhost:8080`

So running the full test suite in one go with `mitm` running will also try to run the camunda tests (because the test suite sees localhost 8080 is available), and that will fail because it's not actually camunda on that port :)

SilviaAmAm: Ahh I understand now! Thank you :slightly_smiling_face:

Viicos (approved):

PR #3736, opened 2024-01-04
[#3731] Add required headers for BRP clients

### LINKED ISSUE #3731: Add BRP headers to fulfill undocumented Haal Centraal BRPv2 requirements
When using the Haal Centraal BRP v2, we need to pass 4 additional headers.

- `x-origin-oin` -- We should already have this value from the ~~DigiD settings.~~ general settings (to be added)
- `x-doelbinding` -- Let's create a general setting, that can be overruled per form. Admin wise, add it to a new fieldset "Verwerkingen" in both general settings and in the base tab of the form.
- `x-verwerking` -- Same as above. Max length = 242 and validate that it can have 0 or 1 "@".
- `x-gebruiker` -- Always fill with the value "BurgerZelf". Only configurable with a kwarg when instantiating the client. This should also be overridden with the employee id, in case the registrator is filling the form

This is needed to fulfill the so-called "protocollering" that is required for accessing the BRPv2 which can also only be called via a gateway that can do this protocollering (of which I cannot find any technical documentation). It's also unclear which headers are needed (if any) for the protocollering and which are needed for PinkRoccade gateway or logging solution but it is what it is.

Viicos:
Fixes #3731.

Before adding tests, I'd like to know if we keep the implementation as is


src/openforms/forms/models/form.py
```diff
@@ -327,6 +327,27 @@ class Form(models.Model):
...
+        max_length=242,
+        blank=True,
+        validators=[
+            RegexValidator(
+                regex="^[^@]*@?[^@]*$",
+                message=_("The header must have 0 or 1 '@' character."),
+            )
+        ],
+        help_text=_("The x-verwerking header to be used when requesting the BRP API."),
+    )
```
Viicos, line 349: Duplicated from the `HaalCentraalConfig` model, any better way of doing this?

src/openforms/pre_requests/base.py
```diff
@@ -10,10 +9,7 @@
     from .clients import PreRequestClientContext
 
 
-@dataclass
 class PreRequestHookBase(ABC, AbstractBasePlugin):
-    identifier: str
-
```
Viicos, line 16: `identifier` is already part of the `AbstractBasePlugin`'s `__init__` method

src/openforms/pre_requests/clients.py
```diff
@@ -38,6 +38,9 @@ class PreRequestMixin:
     than just some aspects of it.
     """
 
+    enabled_plugins: ClassVar[list[str]] = []
+    """A list of plugin identifiers that will be called before preparing the request."""
+
```
Viicos, line 43: When for an empty default: I feel that having all the plugins applied just by subclassing this mixin is a bit "dangerous". Instead, subclasses explicitly set the plugins they want to be applied.

src/openforms/submissions/models/submission.py
```diff
@@ -412,7 +414,7 @@ def is_authenticated(self):
         return hasattr(self, "auth_info")
 
     @property
-    def registrator(self):
+    def registrator(self) -> "AuthInfo | RegistratorInfo | None":
```
Viicos, line 417: or `-> BaseAuthInfo | None`, but slightly incorrect as it either one or the other, never the base common class.

sergei-maertens: can you use `from __future__ import annotations` so you can do this instead:

```suggestion
    def registrator(self) -> AuthInfo | RegistratorInfo | None:
```

sergei-maertens (changes requested):

src/openforms/config/models/config.py
```diff
@@ -536,6 +536,15 @@ class GlobalConfiguration(SingletonModel):
...
     )
 
+    # Processing (BRP)
+
+    x_origin_oin_header = models.CharField(
+        _("x-origin-oin header"),
+        max_length=255,
+        blank=True,
+        help_text=_("The x-origin-oin header to be used when requesting the BRP API."),
+    )
```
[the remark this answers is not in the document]
Viicos: @joeribekker wants it in the global config

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
Viicos: @joeribekker

src/openforms/contrib/haal_centraal/clients/pre_request_hook.py
```diff
@@ -0,0 +1,32 @@
...
+            if submission.form.brp_x_doelbinding_header:
+                headers["x-doelbinding"] = submission.form.brp_x_doelbinding_header
+            if submission.form.brp_x_verwerking_header:
+                headers["x-verwerking"] = submission.form.brp_x_verwerking_header
+
+            if (
+                submission.registrator
+                and submission.registrator.attribute is AuthAttribute.employee_id
+            ):
+                headers["x-gebruiker"] = submission.registrator.value
```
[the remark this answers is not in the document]
joeribekker: I think this looks alright. I can't determine if this will be the hashed value or the normal value but the idea is correct.

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
joeribekker: Hmm, then how is this done with the DigiD-level of assurance?

src/openforms/config/models/config.py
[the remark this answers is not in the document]
sergei-maertens: I can see how it's useful for other things, but then we should definitely name it something else, `organization_oin` in line with `organization_name`?

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
sergei-maertens: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/models/form.py#L99

using a `JSONField` storing the `authentication_backend_options`. Assurance levels is a concept that applies to all plugins, just DigiD and eHerkenning actually make use of it :)

src/openforms/contrib/haal_centraal/clients/pre_request_hook.py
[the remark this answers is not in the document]
sergei-maertens: as long as a submission is not completed or suspended, it's a value in plain text indeed :)

src/openforms/pre_requests/clients.py
Viicos: I added more info about this in the registry docstring

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
joeribekker: @Viicos so, what @sergei-maertens says :)

src/openforms/contrib/haal_centraal/clients/__init__.py
```diff
@@ -21,5 +52,8 @@ def get_brp_client(**kwargs) -> BRPClient:
             f"No suitable client class configured for API version {version}"
         )
 
-    service_client_factory = ServiceClientFactory(service)
+    service_client_factory = BRPServiceClientFactory(
+        haal_centraal_config=config,
+        x_origin_oin_header=global_config.x_origin_oin_header,
+    )
     return ClientCls.configure_from(service_client_factory, **kwargs)
```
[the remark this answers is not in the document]
Viicos: Isn't it what the client factory is for? If we pass everything in `kwargs` so that it get picked up in the `__init__` then I feel we are loosing the purpose of the factory + `configure_from`. On the other side, having everything in the `__init__` makes it more explicit, as this factory pattern makes me think about Java 🤢

src/openforms/contrib/haal_centraal/clients/pre_request_hook.py
```diff
@@ -0,0 +1,32 @@
+from openforms.authentication.constants import AuthAttribute
+from openforms.pre_requests.base import PreRequestHookBase
+from openforms.pre_requests.clients import PreRequestClientContext
+from openforms.pre_requests.registry import register
+
+
+@register("brp-headers")
+class BRPHeadersPreRequestHook(PreRequestHookBase):
```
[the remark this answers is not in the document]
Viicos: Indeed being aware of this plugin isn't trivial

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
Viicos: Maybe this is unlikely to happen, but what if the BRP client is used in a context where a form is available, but not related to prefill? Should we still fetch the headers from this `PrefillOptions` model?

src/openforms/contrib/haal_centraal/clients/__init__.py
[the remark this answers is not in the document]
sergei-maertens: I'll take another look after lunch - maybe it's the combination of this + the pre-request hook plugin that caused it to be overwhelming.

I'm thinking a little bit from the perspective of unit tests if we use the client class/instance directly or exclusively go through `get_brp_client`. If it's the former, making it an `__init__` parameter makes more sense from the perspective of loose coupling, if it's the latter, it's more integrated/self-contained. Let's discuss after lunch!

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
sergei-maertens: good point - there are two such instances:

* `npFamilyMembers` which runs at component-time rather than prefill (submission start) time
* fetching cosign (v1) data for display purposes

Let's evaluate options after lunch!

src/openforms/contrib/haal_centraal/clients/__init__.py
```diff
@@ -1,25 +1,67 @@
...
+        x_headers["x-origin-oin"] = global_config.organization_oin
+
+    if (
+        submission.registrator
+        and submission.registrator.attribute is AuthAttribute.employee_id
+    ):
+        x_headers["x-gebruiker"] = submission.registrator.value
+    else:
+        x_headers["x-gebruiker"] = DEFAULT_X_GEBRUIKER_HEADER
+
```
Viicos, line 54: Not sure where all of this should leave. Either in this helper function or in the factory.

sergei-maertens: I think this is fine, I think you can remove some branches by creating an empty/default instance of the request options if no submission/configuration is present and then simplify lookups like:

```
purpose_limitation = request_options.purpose_limitation or config.default_purpose_limitation
```

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,147 @@
...
+                "attribute": "burgerservicenummer",
+            },
+            "multiple": False,
+        },
+        {
+            "key": "address",
+            "type": "textfield",
+            "label": "Address",
+            "prefill": {
+                "plugin": "haalcentraal",
```
sergei-maertens, line 36: Can you please use the [constant](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/prefill/contrib/haalcentraal_brp/plugin.py#L27) instead? We need to rename the plugin identifier at some point, this is one of the last places where the incorrect assumption is made that `haalcentraal == brp personen`

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,147 @@
...
+        },
+    ],
+}
+
+
+class BRPIntegrationTest(OFVCRMixin, TestCase):
+    """An integration test against a test instance, to make sure headers are set correctly.
+    The prefill feature is being used here, although any other feature making use of the Haal Centraal API
+    could have been used.
+    """
```
sergei-maertens, line 49: Please format your docstrings with the [PEP-257](https://peps.python.org/pep-0257/) convention

Most notably, a single long line that is wrapping is too much, break it up into a short summary line, a blank line and then more elaborate descriptions:

```py
"""
Run full integration tests against a real API instance.

These tests ensure the correct behaviour of our implementation using a real API as reference. HTTP 
traffic is recorded using VCR... etc.
"""
```

The remark at the class level about headers being tested is also in the wrong location - you can add other tests/assertions to this case that test things *other* than the headers, so that part of the docstring belongs with the test itself.

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,147 @@
...
+            "openforms.contrib.haal_centraal.clients.GlobalConfiguration.get_solo",
+            return_value=global_config,
+        )
+        global_config_patcher.start()
+        self.addCleanup(global_config_patcher.stop)
+
+        form_step = FormStepFactory.create(form_definition__configuration=CONFIGURATION)
+        self.submission_step = SubmissionStepFactory.create(
+            submission__form=form_step.form, form_step=form_step
+        )
```
sergei-maertens, line 68: please look for examples of `SubmissionFactory.create(form__generate_minimal_example=True, form__formstep__form_definition__configuration=...)`

These ensure that all data is correctly wired up and let you reason about the submission directly instead of taking a detour via `submission_step`.

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,147 @@
...
+            m.return_value.brp_personen_service.client_certificate
+            and m.return_value.brp_personen_service.server_certificate
+        ):
+            m.return_value.brp_personen_service.client_certificate.save()
+            m.return_value.brp_personen_service.server_certificate.save()
+
+        prefill_variables(submission=self.submission_step.submission)
+
+        state = self.submission_step.submission.load_submission_value_variables_state()
+        self.validate_prefill_variables(state.variables["address"].value)
```
sergei-maertens, line 119: when you're writing custom assertions, call them as such: `self.assertPrefillVariableValues`

In a test it must be visually clear that you're doing:

```
setup phase

do work phase

assertion phase
```
so please also add those blank lines in your test code to make these blocks visually clear

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,147 @@
...
+            x_verwerking_header="Financiële administratie@Correspondentie factuur",
+        )
+
+        prefill_variables(submission=self.submission_step.submission)
+
+        state = self.submission_step.submission.load_submission_value_variables_state()
+        self.validate_prefill_variables(state.variables["address"].value)
+
+
+print("", end="\n\n")
```
sergei-maertens, line 147:
```suggestion
```

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,147 @@
...
+            m.return_value.brp_personen_service.server_certificate.save()
+
+        prefill_variables(submission=self.submission_step.submission)
+
+        state = self.submission_step.submission.load_submission_value_variables_state()
+        self.validate_prefill_variables(state.variables["address"].value)
+
+    @patch(
+        "openforms.contrib.haal_centraal.clients.HaalCentraalConfig.get_solo",
+        return_value=HaalCentraalConfigFactory.build(),
```
sergei-maertens, line 123: please make explicit that no values are set in this case, now you're relying on implicit behaviour of the factory

src/openforms/formio/components/custom.py
```diff
@@ -106,7 +106,11 @@ def normalizer(component: Component, value: str) -> str:
 
 class FamilyMembersHandler(Protocol):
     def __call__(
-        self, bsn: str, include_children: bool, include_partner: bool
+        self,
+        bsn: str,
+        include_children: bool,
+        include_partners: bool,
```
sergei-maertens, line 112: please only make the minimal fix, I see the actual arg being used in the call on line 184 is different from this signature, but refrain from refactoring.

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -123,413 +123,472 @@ const FormConfigurationFields = ({
...
+              />
+            }
+            helpText={
+              <FormattedMessage
+                defaultMessage="The x-doelbinding header to be used when requesting the BRP API."
+                description="Form 'BRP x-doelbinding header' field help text"
+              />
+            }
+          >
+            <TextInput value={internalName} onChange={onChange} maxLength="255" />
```
sergei-maertens, line 568: `internalName` as value for purpose limitation field?

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -123,413 +123,472 @@ const FormConfigurationFields = ({
...
+              />
+            }
+            helpText={
+              <FormattedMessage
+                defaultMessage="The x-verwerking header to be used when requesting the BRP API."
+                description="Form 'BRP x-verwerking header' field help text"
+              />
+            }
+          >
+            <TextInput value={internalName} onChange={onChange} maxLength="242" />
```
sergei-maertens, line 587: `internalName`?

src/openforms/prefill/base.py
```diff
@@ -18,7 +18,7 @@ class BasePlugin(AbstractBasePlugin):
     requires_auth: AuthAttribute | None = None
     for_components: Container[str] = AllComponentTypes()
 
-    def get_available_attributes(self) -> Iterable[Tuple[str, str]]:
+    def get_available_attributes(self) -> Iterable[tuple[str, str]]:
```
sergei-maertens, line 21: this clutters the diff, only make those changes/cleanups if the line is already affected anyway (due to changes in method signature for example)

sergei-maertens (changes requested):
I don't think your frontend code to configure the headers at the form level is functional now, these fields are not exposed in the API yet.

src/openforms/contrib/haal_centraal/clients/__init__.py
```diff
@@ -1,25 +1,67 @@
...
+    origin_oin = global_config.organization_oin
+    doelbinding = config.default_x_doelbinding_header
+    verwerking = config.default_x_verwerking_header
+    gebruiker = DEFAULT_X_GEBRUIKER_HEADER
+
+    if submission is not None:
+        request_options = cast(
+            HaalCentraalRequestOptions | None,
+            getattr(submission.form, "haal_centraal_request_options", None),
+        )
```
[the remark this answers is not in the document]
Viicos: `assert isinstance(object, "tuple_or_union_type")` is not a construct understood by type checkers to this day, so using it would defeat the purpose of the current type casting

src/openforms/contrib/haal_centraal/clients/brp.py
```diff
@@ -41,13 +41,32 @@ class Person:
...
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        if origin_oin:
+            self.headers["x-origin-oin"] = origin_oin
+        if doelbinding:
+            self.headers["x-doelbinding"] = doelbinding
+        if verwerking:
+            self.headers["x-verwerking"] = verwerking
+        if gebruiker:
+            self.headers["x-gebruiker"] = gebruiker
```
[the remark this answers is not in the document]
Viicos: I don't know how the requested APIs will behave if they receive a header (even if empty) they don't know. I think it doesn't hurt to have these checks, doesn't cost anything on our side

src/openforms/contrib/haal_centraal/migrations/0002_copy_config_from_prefill.py
```diff
@@ -21,7 +21,6 @@ def copy_config(apps, _):
 
         old_config, _ = OldConfig.objects.get_or_create(pk=CONFIG_PK)
         new_config, _ = NewConfig.objects.get_or_create(pk=CONFIG_PK)
-
```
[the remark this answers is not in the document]
Viicos: wasn't on purpose

src/openforms/contrib/haal_centraal/models.py
```diff
@@ -40,8 +41,62 @@ class HaalCentraalConfig(SingletonModel):
...
+    default_x_verwerking_header = models.CharField(
+        _("x-verwerking header"),
+        max_length=242,
+        blank=True,
+        validators=[validate_verwerking_header],
+        help_text=_(
+            "The default 'x-verwerking' header to be used when requesting the BRP API, "
+            "in case it is not available on the current form."
+        ),
+    )
```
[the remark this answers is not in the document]
Viicos: > I don't expect other API's in the future to use the same headers

How was I suppose to know this? Anyway I think this relates to what was discussed a while ago on the fact that we are missing a lot of context on how all these APIs/services/movement/initiatives relate together

src/openforms/contrib/haal_centraal/tests/files/pub.crt
```diff
@@ -0,0 +1,48 @@
+Bag Attributes
+    friendlyName: labtenant1.api.mijniconnect.nl
+    localKeyID: 54 69 6D 65 20 31 36 36 37 34 38 36 35 34 30 38 35 32 
+subject=C = NL, O = PinkRoccade Local Government B.V., serialNumber = 00000003273229750000, CN = labtenant1.api.mijniconnect.nl
```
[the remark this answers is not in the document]
Viicos: Yes: `src/suwinet/tests/data/pub.cert`

src/openforms/contrib/haal_centraal/models.py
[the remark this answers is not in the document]
Viicos: > that's the only API where they are currently used, and I don't expect other API's in the future to use the same headers, looking at how things have been designed traditionally...

I hope this will be the case as if a new haal centraal service needs these headers we will need to change the field names. For reference, this is how `open-inwoner` [handles it](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/haalcentraal/models.py); but I'm not sure if they are doing it correctly as well

src/openforms/contrib/haal_centraal/clients/__init__.py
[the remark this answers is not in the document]
sergei-maertens: cool - didn't know about the type checker limitations, so if it pops up again, add a comment describing why `cast` is used instead of `assert` :+1:

src/openforms/contrib/haal_centraal/clients/brp.py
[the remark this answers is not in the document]
sergei-maertens: As far as I understood, currently nothing actually works in a real setup because of these missing headers. Can you check what the behaviour is of the test API you have available? That is the only representative environment we currently have.

Standard behaviour is that unknown headers are ignored, reverse proxies add a bunch of headers too, WAF-etc. add headers too, so we can assume that an implementation that doesn't know these headers will just ignore them.

src/openforms/contrib/haal_centraal/models.py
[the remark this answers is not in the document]
sergei-maertens: > How was I suppose to know this? Anyway I think this relates to what was discussed a while ago on the fact that we are missing a lot of context on how all these APIs/services/movement/initiatives relate together

You weren't - precisely because of the reasons you mention! After seeing this, I started putting more information on Taiga about this and started a discussion on Slack to try and provide more education. Feel free to extend this Taiga documentation with additional information, I've only done the bare minimum to get it started so far.

> I hope this will be the case as if a new haal centraal service needs these headers we will need to change the field names. For reference, this is how open-inwoner [handles it](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/haalcentraal/models.py); but I'm not sure if they are doing it correctly as well

If this becomes the case, then we change the field names, but we can at least trace it back to a request for change. It's easier to make things more generic from specific compared to the other way around.

I can't speak about Open Inwoner - it's not a project I'm directly involved with but I know that I'm one of the strictest people for reviewing etc. so I'm sure I'll have _opinions_ if I were to look into it in more detail :stuck_out_tongue:

src/openforms/contrib/haal_centraal/tests/files/pub.crt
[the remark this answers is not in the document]
sergei-maertens: Okay, it'll be fine then probably :grimacing:

src/openforms/contrib/haal_centraal/models.py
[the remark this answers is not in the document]
Viicos: > You weren't - precisely because of the reasons you mention! After seeing this, I started putting more information on Taiga about this and started a discussion on Slack to try and provide more education. Feel free to extend this Taiga documentation with additional information, I've only done the bare minimum to get it started so far.

I'll see in the future if I can make some kind of diagram or image that explains the relation between these, and also how it relates with the open forms solo models; I think that could be valuable

src/openforms/contrib/haal_centraal/clients/__init__.py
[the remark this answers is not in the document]
Viicos: Yes I was actually surprised by it too

src/openforms/config/migrations/0068_globalconfiguration_organization_oin.py
sergei-maertens, line 1: before merging, please re-generate these migrations into a single one

src/openforms/config/models/config.py
```diff
@@ -285,6 +285,13 @@ class GlobalConfiguration(SingletonModel):
         ),
     )
 
+    organization_oin = models.CharField(
+        _("organization OIN"),
+        max_length=20,
```
sergei-maertens, line 290: can you add a comment pointing to the documentation about OIN I linked?

src/openforms/contrib/haal_centraal/admin.py
```diff
@@ -3,15 +3,25 @@
...
+                    "default_brp_personen_purpose_limitation_header_value",
+                    "default_brp_personen_processing_header_value",
+                )
+            },
         ),
     )
     autocomplete_fields = ("brp_personen_service",)
+
+
+admin.site.register(BRPPersonenRequestOptions)
```
sergei-maertens, line 27: please spend some effort on nice admin integration, `admin.site.register(model)` is almost never good enough.

In particular, ensure we have proper:

* `list_display`
* `list_select_related`
* `list_filter`
* `search_fields`
* `raw_id_fields` (the form FK will render a dropdown with an excessive amount of options)

src/openforms/contrib/haal_centraal/tests/factories.py
```diff
@@ -0,0 +1,56 @@
...
+from factory.django import DjangoModelFactory, FileField
+from simple_certmanager.constants import CertificateTypes
+from zgw_consumers.constants import AuthTypes
+
+from zgw_consumers_ext.tests.factories import ServiceFactory
+
+from ..constants import BRPVersions
+from ..models import HaalCentraalConfig
+
+X_API_KEY = os.getenv("HC_BRP_PERSONEN_API_KEY", "placeholder_key")
```
sergei-maertens, line 14:
```suggestion
HC_BRP_PERSONEN_API_KEY = os.getenv("HC_BRP_PERSONEN_API_KEY", "placeholder_key")
```
with the factories being "shared" I think namespacing needs to happen at this level

src/openforms/contrib/haal_centraal/tests/test_brp_clients.py
```diff
@@ -549,6 +561,160 @@ def test_invalid_version_raises_error(self):
...
+
+    @patch(
+        "openforms.contrib.haal_centraal.clients.GlobalConfiguration.get_solo",
+        return_value=GlobalConfiguration(),
+    )
+    def test_uses_default_headers(self, m):
+        self.requests_mock.post(
+            "https://personen/api/irrelevant",
+            status_code=200,
+        )
```
sergei-maertens, line 605: You can also just blanket mock any request (and then maybe generate a URL using hypothesis to show this applies to any URL/method)

```suggestion
        self.requests_mock.register_uri(requests_mock.ANY, requests_mock.ANY, status_code=200)
```
doesn't necessarily need to be changed, just FYI

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,152 @@
...
+        },
+    ],
+}
+
+
+class BRPIntegrationTest(OFVCRMixin, TestCase):
+    """Run full integration tests against a real API instance.
+
+    These tests ensure the correct behaviour of our implementation using a real API as reference. HTTP
+    traffic is recorded using VCR... etc.
```
sergei-maertens, line 49: the `... etc.` was meant for you to complete :stuck_out_tongue:

src/openforms/js/components/admin/form_design/FormAdvancedConfiguration.js
```diff
@@ -0,0 +1,105 @@
...
+import {TextInput} from 'components/admin/forms/Inputs';
+import {COSIGN_OLD_TYPE_KEY} from 'components/form/coSignOld';
+
+/**
+ * Component to render the metadata admin form for an Open Forms form.
+ */
+const FormAdvancedConfiguration = ({form, formSteps, onChange}) => {
+  const {brpPersonenRequestOptions} = form;
+
+  const components = formSteps.flatMap(fs => fs.configuration.components);
```
SilviaAmAm, line 17: I think that this may miss nested components? For example components inside columns?
Maybe this util could be useful: https://github.com/open-formulieren/open-forms/blob/2bbf2a6f52cb6bb9a2d51441e5679aca83f9cd68/src/openforms/js/components/admin/form_design/utils.js#L19

src/openforms/submissions/models/submission.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
SilviaAmAm, line 2: Why is this needed?

Viicos: To avoid having to use forward references in type hints if imported in an `if TYPE_CHECKING` block:

`def registrator(self) -> "AuthInfo | RegistratorInfo | None":` -> `def registrator(self) -> AuthInfo | RegistratorInfo | None:`

(`from __future__ import annotations` will make all type annotations as forward string references by default)

src/openforms/contrib/haal_centraal/tests/test_integration.py
```diff
@@ -0,0 +1,148 @@
...
+            default_brp_personen_processing_header_value="Financiële administratie@Correspondentie factuur",
+        ),
+    )
+    def test_brp_integration_prefill_default_headers(self, m: MagicMock) -> None:
+        # certificate instances need to be saved so that they get a correct path
+        # (see https://forum.djangoproject.com/t/7533/)
+        if (
+            m.return_value.brp_personen_service.client_certificate
+            and m.return_value.brp_personen_service.server_certificate
+        ):
```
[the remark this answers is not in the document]
Viicos: When you run tests without the correct env variables pointing to the private key, no certificate is set (this doesn't matter for VCR as it doesn't record any certificate info). You can see how the `HaalCentraalConfigFactory` is defined to see how the `brp_personen_service` attr. is instantiated.

src/openforms/js/components/admin/form_design/FormAdvancedConfiguration.js
Viicos: This will avoid my dirty `flatMap` as well :)

src/openforms/js/components/admin/form_design/FormAdvancedConfiguration.js
SilviaAmAm, line 1: I noticed: 
On an existing form where no prefill were configured, I added a prefill configuration to an existing text field, then I went to the Advanced configuration tab and the form crashed (because `brpPersonenRequestOptions` was null)

src/openforms/js/components/admin/form_design/FormAdvancedConfiguration.js
```diff
@@ -0,0 +1,108 @@
...
+              <TextInput
+                value={brpPersonenRequestOptions.brpPersonenProcessingHeaderValue}
+                onChange={onChange}
+                maxLength="242"
+              />
+            </Field>
+          </FormRow>
+        </Fieldset>
+      ) : (
+        <p style={{padding: '1em 0 0', color: 'var(--body-quiet-color)'}}>
```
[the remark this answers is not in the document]
sergei-maertens: I'm not sure that we're even loading the SDK stylesheet here - so if it's too much hassle, I can pretend not having seen this :wink:

src/openforms/js/components/admin/form_design/FormAdvancedConfiguration.js
```diff
@@ -0,0 +1,108 @@
...
+  );
+};
+
+FormAdvancedConfiguration.propTypes = {
+  form: PropTypes.shape({
+    brpPersonenRequestOptions: PropTypes.shape({
+      brpPersonenPurposeLimitationHeaderValue: PropTypes.string,
+      brpPersonenProcessingHeaderValue: PropTypes.string,
+    }),
+  }).isRequired,
```
sergei-maertens, line 103: I hope we can get started on TypeScript this year so we can consolidate these prop type definitions

src/openforms/js/components/admin/form_design/FormAdvancedConfiguration.js
```diff
@@ -0,0 +1,108 @@
...
+  formSteps: PropTypes.array.isRequired,
```
sergei-maertens, line 104:
```suggestion
  formSteps: PropTypes.arrayOf(TYPES.FormStep).isRequired,
```
with https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/form_design/types/index.js#L9, I think?

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
sergei-maertens, line 1: are the changes here still relevant? I see a huge diff but it seems to be mostly because everything is now wrapped in a single fragment, but since the configuration is moved to its own tab, I don't think the fragment is no longer needed?

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1442,6 +1451,14 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl}) => {
               />
             </TabPanel>
           )}
+
+          <TabPanel>
+            <FormAdvancedConfiguration
+              form={state.form}
+              formSteps={state.formSteps}
+              onChange={onFieldChange}
+            ></FormAdvancedConfiguration>
```
sergei-maertens, line 1460:
```suggestion
            <FormAdvancedConfiguration
              form={state.form}
              formSteps={state.formSteps}
              onChange={onFieldChange}
            />
```

src/openforms/js/components/form/coSignOld.js
```diff
@@ -9,6 +9,8 @@ import {localiseSchema} from './i18n';
 
 const FieldComponent = Formio.Components.components.field;
 
+export const COSIGN_OLD_TYPE_KEY = 'coSign';
```
sergei-maertens, line 12:
```suggestion
export const COSIGN_V1_TYPE = 'coSign';
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3737, opened 2024-01-04
Replace django-better-admin-arrayfield

sergei-maertens:
Part of #3049

Some more styling tweaks to do, but it's pretty much a drop-in replacement :tada:


Viicos (approved):
I find it weird to have a custom db field being used just for form related stuff. Looking at the source, the `ArrayField` mainly overrides the `formfield` method.

But I guess this is coming from Django

sergei-maertens:
> I find it weird to have a custom db field being used just for form related stuff. Looking at the source, the `ArrayField` mainly overrides the `formfield` method.
> 
> But I guess this is coming from Django

Correct, the `def formfield` is a hook to allow you to specify the default form field class/widget to use without needing a bunch of mixins. You don't *need* to do it, but then you have to override the admin form for every relevant model and specify the overridden form fields at that level, which is a lot (not-interesting) more boiler plate code.

PR #3738, opened 2024-01-05
Minor cleanups

sergei-maertens:


Viicos (approved):
Nice, would be great at some point to remove the startup logging of weasyprint, as it appears also on every management command.

I think on the weasyprint side they should avoid having their stylesheets constant computed at import, as not only it adds logging in our case, it slows down on import:

```shell
$ time python -c 'import weasyprint'

real	0m0.591s
user	0m0.502s
sys	0m0.092s
```

But it would involve having a module level `__getattr__` and I don't have time for that

sergei-maertens:
> Nice, would be great at some point to remove the startup logging of weasyprint, as it appears also on every management command.
> 
> I think on the weasyprint side they should avoid having their stylesheets constant computed at import, as not only it adds logging in our case, it slows down on import:
> 
> ```shell
> $ time python -c 'import weasyprint'
> 
> real	0m0.591s
> user	0m0.502s
> sys	0m0.092s
> ```
> 
> But it wouls involve having a module level `__getattr__` and I don't have time for that

yeah that's what prompted me to look into it, since it also seems to happen on every reload of the django dev server. All this import-time work is not very elegant imo :(

PR #3739, opened 2024-01-05
[#3726] Payment views rework

### LINKED ISSUE #3726: Update submission confirmation page (to clarify payment)
Currently, when a user submits a form which requires payment, they see this page:
Screenshot from 2023-12-29 10-25-44
Once they pay, then they see this screen:
Screenshot from 2023-12-29 10-37-06

This can be misleading, because from the confirmation page one may overlook the fact that a payment is still required.

**New design idea**

If the form requires payment, there will be a page with heading "Betalen" and an extra 'payment' step in the progress indicator.
The body will have the submission reference and the button to pay. Maybe the link to download the submission report could be added under the payment button.

After the user has paid (or if no payment is needed), then comes the "bevestiging" pagina. This says (if relevant) that the payment has been received, the submission reference and it shows the link to download the submission report. It also should have the 'Terug naar de website' button. 
Here there is no need to show the progress indicator.

SilviaAmAm:
Fixes #3726 (partially)

Frontend PR: https://github.com/open-formulieren/open-forms-sdk/pull/628

### DIFF
--- a/src/openforms/payments/contrib/demo/plugin.py
+++ b/src/openforms/payments/contrib/demo/plugin.py
@@ -1,6 +1,10 @@
 from django.http import HttpResponseRedirect
 from django.utils.translation import gettext_lazy as _
 
+from rest_framework.reverse import reverse
+
+from openforms.submissions.tokens import submission_status_token_generator
+
 from ...base import BasePlugin, PaymentInfo
 from ...constants import PaymentStatus, UserAction
 from ...registry import register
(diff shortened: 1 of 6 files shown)


src/openforms/payments/contrib/demo/plugin.py
```diff
@@ -19,12 +23,21 @@ def handle_return(self, request, payment):
...
 
+        token = submission_status_token_generator.make_token(payment.submission)
+        status_url = request.build_absolute_uri(
+            reverse(
+                "api:submission-status",
+                kwargs={"uuid": payment.submission.uuid, "token": token},
+            )
+        )
+
         form_url = payment.submission.cleaned_form_url
```
sergei-maertens, line 34: is there any more work needed to make this work properly with hash-based routing too? Using https://github.com/open-formulieren/open-forms/blob/master/src/openforms/frontend/frontend.py#L9

SilviaAmAm: I am not sure, I will look into it!

SilviaAmAm: I added the support for it!

SilviaAmAm:
The E2E test won't succeed until the SDK PR is merged

sergei-maertens (approved):

PR #3740, opened 2024-01-05
[#3600] Map form fields to eigenschappen in zgw registration

### LINKED ISSUE #3600: As municipality I want to map form fields/vars to Zaak.eigenschappen in the registration backend
... so we can make use of the eigenschappen at all.

ref: Utr 139

Some discussion about making this generic is in order.

**Tasks**

* [x] Create a custom React form (like the Camunda form) for the ZGW API Registration details (in form > registration tab)
* [x] Add an array of (eigenschap, variable) to the form
* [x] You should be able to type the eigenschap-naam (no smart selecting here), and you should select a variable from the list of all variables (this is a re-usable React component)
* [x] You should be able to add and remove items from the array (plus-icon, trashcan)

vaszig:
Fixes #3600

### DIFF
--- a/docs/configuration/registration/zgw.rst
+++ b/docs/configuration/registration/zgw.rst
@@ -79,11 +79,11 @@ Configuration
       * **Type**: Select the type, one of: ``ZRC``, ``DRC`` or ``ZTC``
       * **API root url**: The root of this API, *For example* ``https://example.com/zaken/api/v1/``
 
-      * **Client ID**: Fill the value provided by the backend service *For example:* ``open-zaak`` (*NOTE* this could be different for each component)
+      * **Client ID**: Fill the value provided by the backend service. *For example:* ``open-zaak`` (*NOTE* this could be different for each component)
       * **Secret**: Fill the value provided by the backend service (*NOTE* this could be different for each component)
       * **Authorization type**: Select the option: ``ZGW client_id + secret``
-      * **OAS**: URL that points to the OAS, same URL as used for **API root url** with ``/schema/openapi.yaml`` added to it
-        *for example:* ``https://example.com/zaken/api/v1/schema/openapi.yaml``
+      * **OAS**: URL that points to the OAS, same URL as used for **API root url** with ``/schema/openapi.yaml`` added to it.
+        *For example:* ``https://example.com/zaken/api/v1/schema/openapi.yaml``
 
       * **NLX**: Support for NLX can be selected here if enabled in the installation
       * **User ID**: Audit trail user ID, usually same as the Client ID
@@ -109,11 +109,11 @@ Configuration
 
 #. You can combine ZGW API with Objects API. The submission data will be sent to both if the following have been provided:
 
-   * **Objects API - Objecttype**: Fill in the default URL of the ProductAanvraag OBJECTTYPE in the Objecttypes API *For example* ``https://example.com/api/v1/objecttypes``
-   * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API *For example:* ``1``
+   * **Objects API - Objecttype**: Fill in the default URL of the ProductAanvraag OBJECTTYPE in the Objecttypes API. *For example* ``https://example.com/api/v1/objecttypes``
+   * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API. *For example:* ``1``
 
-#. You can map a form variable with a ZAAK property (eigenshap) by clicking the related button. A modal will open where you have to choose the variable along with providing a valid property name.
-   Both fields should be provided in order to create a connection.
+#. You can map a form variable with a Zaak property (eigenshap) by clicking the related button (*Connect variables with ZAAK properties*). 
+   A modal will open where you have to choose the variable along with providing a valid property name. Both fields should be provided in order to create a connection.
 
 #. Click **Opslaan**
 
--- a/docs/configuration/registration/zgw.rst
+++ b/docs/configuration/registration/zgw.rst
@@ -113,7 +113,7 @@ Configuration
    * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API. *For example:* ``1``
 
 #. You can map a form variable with a Zaak property (eigenschap) by clicking the related button (*Map variables to case properties*). 
-   A modal will open where you have to choose the variable along with providing a valid property name. Both fields should be provided in order to create a connection.
+   A modal will open where you have to choose the variable along with providing a valid property name. Both fields must be provided in order to create a mapping.
 
 #. Click **Opslaan**
 
--- a/docs/configuration/registration/zgw.rst
+++ b/docs/configuration/registration/zgw.rst
@@ -112,7 +112,7 @@ Configuration
    * **Objects API - Objecttype**: Fill in the default URL of the ProductAanvraag OBJECTTYPE in the Objecttypes API. *For example* ``https://example.com/api/v1/objecttypes``
    * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API. *For example:* ``1``
 
-#. You can map a form variable with a Zaak property (eigenshap) by clicking the related button (*Connect variables with ZAAK properties*). 
+#. You can map a form variable with a Zaak property (eigenschap) by clicking the related button (*Map variables to case properties*). 
    A modal will open where you have to choose the variable along with providing a valid property name. Both fields should be provided in order to create a connection.
 
 #. Click **Opslaan**
--- a/docs/configuration/registration/zgw.rst
+++ b/docs/configuration/registration/zgw.rst
@@ -112,6 +112,9 @@ Configuration
    * **Objects API - Objecttype**: Fill in the default URL of the ProductAanvraag OBJECTTYPE in the Objecttypes API *For example* ``https://example.com/api/v1/objecttypes``
    * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API *For example:* ``1``
 
+#. You can map a form variable with a ZAAK property (eigenshap) by clicking the related button. A modal will open where you have to choose the variable along with providing a valid property name.
+   Both fields should be provided in order to create a connection.
+
 #. Click **Opslaan**
 
 If you have added services to multiple ZGW APIs, you can create multiple ZGW API sets. This will enable you to specify
--- a/src/openforms/js/components/admin/form_design/data/complete-form.js
+++ b/src/openforms/js/components/admin/form_design/data/complete-form.js
@@ -96,6 +96,24 @@ const handleAppointmentForm = draft => {
 };
 
 /**
+ * Options for ZGW registration backend can be empty strings but the serializer does not allow them.
+ * This is the way the regular forms treat options, they don't send data for the field when empty str.
+ */
+const handleZgwRegistrationOptions = draft => {
+  if (draft.form?.registrationBackends) {
+    draft.form.registrationBackends.forEach(backend => {
+      if (backend?.backend === 'zgw-create-zaak') {
+        for (const key in backend?.options) {
+          if (backend.options[key] === '') {
+            delete backend.options[key];
+          }
+        }
+      }
+    });
+  }
+};
+
+/**
  * Save the form itself without any related objects.
  */
 const saveForm = async (state, csrftoken) => {
@@ -112,6 +130,7 @@ const saveForm = async (state, csrftoken) => {
     normalizeLimit(draft, 'allSubmissionsRemovalLimit');
     normalizeEmptyStrField(draft, 'activateOn');
     normalizeEmptyStrField(draft, 'deactivateOn');
+    handleZgwRegistrationOptions(draft);
     handleAppointmentForm(draft);
   });
 
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsForm.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsForm.js
@@ -40,7 +40,7 @@ ZGWOptionsForm.propTypes = {
     ),
     zaakVertrouwelijkheidaanduiding: PropTypes.string,
     zaaktype: PropTypes.string,
-    zgwApiGroup: PropTypes.number,
+    zgwApiGroup: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   }),
   formData: PropTypes.shape({
     contentJson: PropTypes.string,
@@ -57,7 +57,7 @@ ZGWOptionsForm.propTypes = {
     ).isRequired,
     zaakVertrouwelijkheidaanduiding: PropTypes.string,
     zaaktype: PropTypes.string,
-    zgwApiGroup: PropTypes.number,
+    zgwApiGroup: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   }),
   onChange: PropTypes.func.isRequired,
 };
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsForm.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsForm.js
@@ -1,0 +1,63 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import Field from 'components/admin/forms/Field';
+
+import ZGWFormFields from './OptionsFormFields';
+
+const ZGWOptionsForm = ({index, name, label, schema, formData, onChange}) => {
+  return (
+    <Field name={name} label={label} errors={[]}>
+      <ZGWFormFields
+        index={index}
+        name={name}
+        schema={schema}
+        formData={formData}
+        onChange={formData => onChange({formData})}
+      />
+    </Field>
+  );
+};
+
+export default ZGWOptionsForm;
+
+ZGWOptionsForm.propTypes = {
+  index: PropTypes.number,
+  name: PropTypes.string,
+  label: PropTypes.node,
+  schema: PropTypes.shape({
+    contentJson: PropTypes.string,
+    informatieobjecttype: PropTypes.string,
+    medewerkerRoltype: PropTypes.string,
+    objecttype: PropTypes.string,
+    objecttypeVersion: PropTypes.number,
+    organisatieRsin: PropTypes.string,
+    variablesProperties: PropTypes.arrayOf(
+      PropTypes.shape({
+        componentKey: PropTypes.string,
+        eigenshap: PropTypes.string,
+      })
+    ),
+    zaakVertrouwelijkheidaanduiding: PropTypes.string,
+    zaaktype: PropTypes.string,
+    zgwApiGroup: PropTypes.number,
+  }),
+  formData: PropTypes.shape({
+    contentJson: PropTypes.string,
+    informatieobjecttype: PropTypes.string,
+    medewerkerRoltype: PropTypes.string,
+    objecttype: PropTypes.string,
+    objecttypeVersion: PropTypes.string,
+    organisatieRsin: PropTypes.string,
+    variablesProperties: PropTypes.arrayOf(
+      PropTypes.shape({
+        componentKey: PropTypes.string,
+        eigenshap: PropTypes.string,
+      })
+    ).isRequired,
+    zaakVertrouwelijkheidaanduiding: PropTypes.string,
+    zaaktype: PropTypes.string,
+    zgwApiGroup: PropTypes.number,
+  }),
+  onChange: PropTypes.func.isRequired,
+};
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
@@ -46,6 +46,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
   return (
     <Wrapper>
       <CustomFieldTemplate
+        id="root_zgwApiGroup"
         label={intl.formatMessage({
           defaultMessage: 'ZGW API group',
           description: 'ZGW API group',
@@ -72,6 +73,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_zaaktype"
         label={intl.formatMessage({
           defaultMessage: 'Zaaktype',
           description: 'URL of the ZAAKTYPE in the Catalogi API',
@@ -92,6 +94,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_informatieobjecttype"
         label={intl.formatMessage({
           defaultMessage: 'Informatieobjecttype',
           description: 'URL of the INFORMATIEOBJECTTYPE in the Catalogi API',
@@ -117,6 +120,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_organisatieRsin"
         label={intl.formatMessage({
           defaultMessage: 'Organisatie rsin',
           description: 'RSIN of organization, which creates the ZAAK',
@@ -142,6 +146,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_zaakVertrouwelijkheidaanduiding"
         label={intl.formatMessage({
           defaultMessage: 'Confidentiality',
           description:
@@ -170,6 +175,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_medewerkerRoltype"
         label={intl.formatMessage({
           defaultMessage: 'Medewerker roltype',
           description:
@@ -197,6 +203,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_objecttype"
         label={intl.formatMessage({
           defaultMessage: 'Objects API - objecttype',
           description:
@@ -224,6 +231,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_objecttypeVersion"
         label={intl.formatMessage({
           defaultMessage: 'Objects API - objecttype version',
           description: 'Version of the object type in the Object Types API',
@@ -249,6 +257,7 @@ const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
       </CustomFieldTemplate>
 
       <CustomFieldTemplate
+        id="root_contentJson"
         label={intl.formatMessage({
           defaultMessage: 'Objects API - JSON content field',
           description: 'JSON template for the content of the request sent to the Objects API',
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
@@ -1,0 +1,340 @@
+import {produce} from 'immer';
+import PropTypes from 'prop-types';
+import React, {useContext} from 'react';
+import {useIntl} from 'react-intl';
+
+import {CustomFieldTemplate} from 'components/admin/RJSFWrapper';
+import {TextArea, TextInput} from 'components/admin/forms/Inputs';
+import Select from 'components/admin/forms/Select';
+import {ValidationErrorContext} from 'components/admin/forms/ValidationErrors';
+
+import {VariablePropertyModal} from './OptionsVariablesProperties';
+import {getChoicesFromSchema, getErrorMarkup} from './utils';
+
+const Wrapper = ({children}) => (
+  <form className="rjsf" name="form.registrationBackendOptions">
+    <CustomFieldTemplate displayLabel={false} errors={null}>
+      <fieldset id="root">{children}</fieldset>
+    </CustomFieldTemplate>
+  </form>
+);
+
+const ZGWFormFields = ({index, name, schema, formData, onChange}) => {
+  const intl = useIntl();
+  const validationErrors = useContext(ValidationErrorContext);
+
+  const {
+    zgwApiGroup = '',
+    zaakVertrouwelijkheidaanduiding = '',
+    zaaktype = '',
+    informatieobjecttype = '',
+    organisatieRsin = '',
+    medewerkerRoltype = '',
+    objecttype = '',
+    objecttypeVersion = '',
+    contentJson = '',
+  } = formData;
+
+  const onFieldChange = event => {
+    const {name, value} = event.target;
+    const updatedFormData = produce(formData, draft => {
+      draft[name] = value;
+    });
+    onChange(updatedFormData);
+  };
+
+  const getFieldErrors = (index, errors, field) => {
+    const errorMessages = [];
+
+    for (const [errorName, errorReason] of errors) {
+      if (errorName.startsWith(name)) {
+        const errorNameBits = errorName.split('.');
+        if (
+          errorNameBits[2] === String(index) &&
+          errorNameBits[errorNameBits.length - 1] === field
+        ) {
+          errorMessages.push(errorReason);
+        }
+      }
+    }
+
+    return errorMessages.length > 0 ? errorMessages : null;
+  };
+
+  return (
+    <Wrapper>
+      <CustomFieldTemplate
+        label={intl.formatMessage({
+          defaultMessage: 'ZGW API set',
+          description: 'ZGW API set',
+        })}
+        rawDescription={intl.formatMessage({
+          description: 'ZGW API set selection',
+          defaultMessage: 'Which ZGW API set to use.',
+        })}
+        rawErrors={null}
+        errors={null}
+        displayLabel
+      >
+        <Select
+          id="zgwApiGroup"
+          name="zgwApiGroup"
+          choices={getChoicesFromSchema(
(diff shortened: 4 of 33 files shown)


vaszig:
FYI interactive tests for the new form's components will be added to storybook after the first review (when we conclude that these are ok for our purpose) - that's what was discussed during last week's standup (Sergei mentioned this).

docs/configuration/registration/zgw.rst
```diff
@@ -112,6 +112,9 @@ Configuration
    * **Objects API - Objecttype**: Fill in the default URL of the ProductAanvraag OBJECTTYPE in the Objecttypes API *For example* ``https://example.com/api/v1/objecttypes``
```
Viicos, line 112:
```suggestion
   * **Objects API - Objecttype**: Fill in the default URL of the ProductAanvraag OBJECTTYPE in the Objecttypes API. *For example* ``https://example.com/api/v1/objecttypes``
```

docs/configuration/registration/zgw.rst
```diff
@@ -112,6 +112,9 @@ Configuration
...
    * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API *For example:* ``1``
```
Viicos, line 113:
```suggestion
   * **Objects API - Objecttype version**: Fill in the default version of the OBJECTTYPE in the Objecttypes API. *For example:* ``1``
```

docs/configuration/registration/zgw.rst
```diff
@@ -112,6 +112,9 @@ Configuration
...
 
+#. You can map a form variable with a ZAAK property (eigenshap) by clicking the related button. A modal will open where you have to choose the variable along with providing a valid property name.
```
Viicos, line 115:
```suggestion
#. You can map a form variable with a Zaak property (eigenshap) by clicking the related button. A modal will open where you have to choose the variable along with providing a valid property name.
```

By "related button", do you mean this one?

image

If this is the case, might be worth explicitly referencing the button label

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -117,6 +170,22 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+
+                if (component_key and not property) or (property and not component_key):
+                    raise serializers.ValidationError(
+                        {
+                            "variables_properties": _(
+                                "Both variable selection and property name are required."
+                            )
+                        },
+                        code="invalid",
+                    )
```
Viicos, line 187: Can't we make `component_key` and `eigenshap` both required in `MappedVariablePropertySerializer`? Or is there a use case when both can be empty?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -464,6 +533,43 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
                 "intermediate.objects_api_zaakobject",
             )
 
+        # Connect variables with zaak eigenshappen (if a connection has been in ZGW registration)
```
Viicos, line 536:
```suggestion
        # Connect variables with zaak eigenshappen (if a connection has been used in ZGW registration)
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -38,6 +38,52 @@
 logger = logging.getLogger(__name__)
 
 
+def get_variables_properties_from_submission(
+    submission: Submission, options: Dict[str, Any]
+) -> Dict[str, Any]:
```
Viicos, line 43:
```suggestion
def get_variables_properties_from_submission(
    submission: Submission, options: dict[str, Any]
) -> dict[str, Any]:
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -38,6 +38,52 @@
...
+    variables_properties = {}
+
+    connections = options.get("variables_properties", [])
```
Viicos, line 46:
```suggestion
    connections = options["variables_properties"]
```

Line 537, we already have this check: `if options.get("variables_properties"):`

Viicos (changes requested):
Backend review

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsVariablesProperties.js
```diff
@@ -0,0 +1,265 @@
...
+const HeadColumns = () => {
+  const intl = useIntl();
+
+  const componentText = (
+    <FormattedMessage description="Manage component column title" defaultMessage="Variable" />
+  );
+  const componentHelp = intl.formatMessage({
+    description: 'Manage component help text',
+    defaultMessage: 'The value of the selected field will be the process variable value.',
+  });
```
Viicos, line 33: Is there a reason to use `FormattedMessage` and `intl.formatMessage` and not only one of them?

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
Viicos, line 1: I'm not sure I understand why defining the whole form explicitly is needed if we are making use of `RJSF`? Is it because of the translations that needs to be set for each field label/help text?

Other than that, the frontend looks fine (other than imo it is utterly verbose but it is what it is :smiling_face_with_tear:) but I can't really say in details as there are some things I'm not sure myself

vaszig: I had discussed a bit this with @sergei-maertens and he had told me that the form should be defined explicitly and as the task mentions like the camunda one, so that's why it's implemented in the same way. I don't know if I misunderstood something, maybe @SilviaAmAm could tell us her opinion too.

Viicos: My understanding is that `RJSF` uses a JSON Schema to automatically create the form, so that you don't have to define the inputs yourself, hence my question

src/openforms/registrations/contrib/zgw_apis/plugin.py
SilviaAmAm: I agree that it would make sense to make both fields required in the `MappedVariablePropertySerializer`. In addition, I think we should add a check like the one of the roltype below:
If the form designer wants to relate a variable to an eigenschap, we need to make sure that the eigenschap has an eigenschaptype compatible (related) to the zaaktype specified.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -117,6 +170,22 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
                     code="invalid",
                 )
 
+        # Make sure we have a variable-eigenshap mapping
+        if variables_properties := attrs.get("variables_properties"):
```
SilviaAmAm, line 174: The value of the zaakeigenschap can only be a string, so maybe we also need to check that the variables selected don't have complex types.

SilviaAmAm (commented):
Also an initial backend review! I will do later the frontend review.

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
SilviaAmAm: As discussed yesterday with Vasileios, I think the custom form (which defines explicitly all the fields) was added so that we can use a custom field with the modal that opens up and shows a table where you can select the variables (instead of having to type the component key manually).

Maybe a potential alternative would be to use a custom widget for the `variablesProperties` so that you don't have to re-specify the whole form :thinking:

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -95,6 +95,24 @@ const handleAppointmentForm = draft => {
...
+        for (const key in backend?.options) {
+          if (backend.options[key] === '') {
+            delete backend.options[key];
+          }
+        }
+      }
+    });
+  }
+};
+
```
SilviaAmAm, line 115: What about the case where you have a value specified in one of the fields (form specific configuration), but you want to revert back to using the global configuration? Does it let you empty the field?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -464,6 +533,43 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
+                submission, options
+            )
+
+            eigenshappen = execute_unless_result_exists(
+                partial(catalogi_client.list_eigenshappen, options["zaaktype"]),
+                submission,
+                "intermediate.eigenshap",
+            )
+
+            retrieved_eigenshappen = {
```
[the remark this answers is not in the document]
vaszig: I think this should be safe enough according to [https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/components/catalogi/api/serializers/eigenschap.py#L100](url) and the fact that we create connections for a specific `zaaktype`

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsForm.js
```diff
@@ -0,0 +1,63 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import Field from 'components/admin/forms/Field';
+
+import ZGWFormFields from './OptionsFormFields';
+
+const ZGWOptionsForm = ({index, name, label, schema, formData, onChange}) => {
+  return (
+    <Field name={name} label={label} errors={[]}>
```
SilviaAmAm, line 10: I suspect that here we pass an empty array to the errors because the errors are then handled inside the `ZGWFormFields` component explicitly? If yes, maybe worth adding a comment.

I also wonder: is this component needed? Could maybe this component contain directly all the implementation of `ZGWFormFields`?

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsForm.js
SilviaAmAm, line 1: It is generally helpful to have the name of the file match the name of the component inside (so ZGWOptionsForm.js instead of OptionsForm.js)

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
```diff
@@ -0,0 +1,331 @@
...
+    contentJson = '',
+  } = formData;
+
+  const onFieldChange = event => {
+    const {name, value} = event.target;
+    const updatedFormData = produce(formData, draft => {
+      draft[name] = value;
+    });
+    onChange(updatedFormData);
+  };
```
SilviaAmAm, line 44: Is this function needed? Or could the onChange be used directly?

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
```diff
@@ -0,0 +1,331 @@
...
+
+  return (
+    <Wrapper>
+      <CustomFieldTemplate
+        id="root_zgwApiGroup"
```
SilviaAmAm, line 49: What is the reason for these ids (instead of for example `registrationBackendOptions.zgwApiGroup`)?

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
```diff
@@ -0,0 +1,331 @@
...
+        })}
+        rawDescription={intl.formatMessage({
+          description: 'Informatieobjecttype URL',
+          defaultMessage: 'URL of the INFORMATIEOBJECTTYPE in the Catalogi API.',
+        })}
+        rawErrors={getFieldErrors(name, index, validationErrors, 'informatieobjecttype')}
+        errors={
+          getFieldErrors(name, index, validationErrors, 'informatieobjecttype')
+            ? getErrorMarkup(getFieldErrors(name, index, validationErrors, 'informatieobjecttype'))
+            : null
```
SilviaAmAm, line 110: This pattern is repeated a lot, maybe can be extracted?

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.stories.js
```diff
@@ -0,0 +1,8 @@
+import ZGWOptionsForm from './OptionsForm';
+
+export default {
+  title: 'Form design/Registration/ZGW',
+  component: ZGWOptionsForm,
+};
+
+export const Default = {};
```
SilviaAmAm, line 8: As mentioned in the other comments: this will be added after the first reviews

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsVariablesProperties.js
```diff
@@ -0,0 +1,265 @@
...
+      <div className="actions actions--horizontal actions--roomy">
+        <DeleteIcon onConfirm={onDelete} message={confirmDeleteMessage} />
+      </div>
+
+      <ComponentSelection
+        name="componentKey"
+        value={componentKey}
+        onChange={onChange}
+        filter={componentFilterFunc.bind(null, componentKey)}
+      />
```
SilviaAmAm, line 78: Since the eigenschappen can only have a string as a value, maybe this should filter out the components with more complex values

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsVariablesProperties.js
```diff
@@ -0,0 +1,265 @@
...
+
+const SelectVariablesProperties = ({variablesProperties = [], onChange, onAdd, onDelete}) => {
+  const formContext = useContext(FormContext);
+  const allComponents = formContext.components;
+
+  const usedComponents = variablesProperties
+    .filter(connection => connection.componentKey !== '')
+    .map(connection => connection.componentKey);
+
+  const filterFunc = (componentKey, {key}) => componentKey === key || !usedComponents.includes(key);
```
SilviaAmAm, line 106: Is this present so that you can't map the same variable to multiple eigenschappen? Is it needed?

src/openforms/js/components/admin/form_design/registrations/zgw/utils.js
```diff
@@ -0,0 +1,43 @@
...
+    <div className="rjsf-field__errors">
+      <div>
+        <ul className="error-detail bs-callout bs-callout-info">
+          {errorMessages.map((msg, index) => (
+            <li key={index} className="text-danger">
+              {msg}
+            </li>
+          ))}
+        </ul>
+      </div>
```
SilviaAmAm, line 38: Why the extra div without class name? Also, would it make sense to add this to the `CustomFieldTemplate` instead of calling it per custom field?

src/openforms/js/components/admin/form_design/data/complete-form.js
vaszig: Yes, basically this is the exact reason I added this here. Without this I get an error from the backend that the field cannot be empty (empty string).

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
vaszig: As discussed in slack this function receives the `formData` so it's necessary here.

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
vaszig: This id here is used in `CustomFieldTemplate` component for the `htmlFor` property of the label. Otherwise it's not accessible. If I use `registrationBackendOptions.zgwApiGroup` I have to change it in the `Select` component below, correct?

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsVariablesProperties.js
vaszig: Yes, we said for now we will assume one variable can be connected to one property.

src/openforms/js/components/admin/form_design/registrations/zgw/utils.js
vaszig: Yes, I removed this div. The way I implemented this has to do with what the `CustomFieldTemplate` component receives for the errors, which is a `ReactElement`. So I thought it would be better to do this explicitly.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -38,6 +39,46 @@
...
+        variables_properties[eigenshap] = form_value
+
+    return variables_properties
+
+
+class MappedVariablePropertySerializer(serializers.Serializer):
+    component_key = serializers.CharField(
+        label=_("Component key"),
+        help_text=_("Key of the Formio.js component to take the value from."),
+    )
```
sergei-maertens, line 73: this should (re) use the `openforms.forms.models.form_variable.variable_key_validator` validator.

To make sure that code is located in the right place, this existing validator should be moved into `openforms.formio.validators` too :)

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -38,6 +39,46 @@
...
+    eigenshap = serializers.CharField(
```
sergei-maertens, line 74:
```suggestion
    eigenschap = serializers.CharField(
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -38,6 +39,46 @@
...
+        label=_("Property"),
+        help_text=_(
+            "This is the name of the property with which the variable will be connected."
+        ),
+    )
```
sergei-maertens, line 79: You should probably also add validators in here to apply the same validations as `naam` in the Eigenschap resource from https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/VNG-Realisatie/catalogi-api/1.3.1/src/openapi.yaml#tag/eigenschappen/operation/eigenschap_retrieve:

* max length
* possible character set? Check the implementation of Open Zaak to be sure

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -102,6 +143,13 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
         required=False,
     )
 
+    # Eigenshappen
+    variables_properties = MappedVariablePropertySerializer(
+        many=True,
+        label=_("Mapped properties variables"),
+        required=False,
+    )
```
sergei-maertens, line 151:
```suggestion
    # Eigenschappen
    property_mappings = MappedVariablePropertySerializer(
        many=True,
        label=_("Mapped properties variables"),
        required=False,
    )
```
perhaps?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -117,12 +165,33 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
 
-        if not ("medewerker_roltype" in attrs and "zaaktype" in attrs):
-            return attrs
-
         # We know it exists thanks to the previous check
         group_config = ZGWRegistration.get_zgw_config(attrs)
 
+        # Make sure the property (eigenshap) related to the zaaktype exists
+        if connections := attrs.get("variables_properties"):
+            with get_catalogi_client(group_config) as client:
```
sergei-maertens, line 173: there's a bunch more spelling mistakes to fix here :stuck_out_tongue:

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -117,12 +165,33 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+                for conn in connections:
+                    if conn["eigenshap"] not in retrieved_eigenshappen:
+                        raise serializers.ValidationError(
+                            {
+                                "variables_properties": _(
+                                    "Could not find a property with the name '{property_name}' related to the zaaktype."
+                                ).format(property_name=conn["eigenshap"])
+                            },
+                            code="invalid",
+                        )
```
sergei-maertens, line 190: Can we collect all the errors and then only raise an error once? If a user makes two typo's, ideally you get that reported back once instead of fixing the first, and then getting another validation error for the second.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -464,6 +533,47 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
+        # Connect variables with zaak eigenshappen (if a connection has been used in ZGW registration)
+        if variables_properties := options.get("variables_properties"):
+            mapped_variables_properties = get_variables_properties_from_submission(
+                submission, variables_properties
+            )
+
+            eigenshappen = execute_unless_result_exists(
+                partial(catalogi_client.list_eigenshappen, options["zaaktype"]),
+                submission,
+                "intermediate.eigenshap",
```
sergei-maertens, line 545:
```suggestion
                "intermediate.zaaktype_eigenschappen",
```
probably better to namespace this a bit to improve readability

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -464,6 +533,47 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
+                    eigenschap_url = furl(retrieved_eigenshappen[key])
+                    eigenschap_uuid = eigenschap_url.path.segments[-1]
+
+                    created_zaakeigenshap = execute_unless_result_exists(
+                        partial(
+                            zaken_client.create_zaakeigenschap,
+                            {
+                                "zaak": zaak["url"],
+                                "eigenschap": retrieved_eigenshappen[key],
+                                "waarde": value,
```
sergei-maertens, line 563: I think we need to do some processing on the `value` based on the `eigenschap.specificatie.formaat`

Looking at the issues in VNG standard, I find [this](https://github.com/VNG-Realisatie/gemma-zaken/issues/1751#issuecomment-1198034861)

Which says that format "datum" is written as `YYYYMMDD`, and for "datumtijd" this becomes `YYYYMMDDHHmmSS`

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -38,6 +39,46 @@
 logger = logging.getLogger(__name__)
 
 
+def get_variables_properties_from_submission(
+    submission: Submission, connections: dict[str, Any]
```
sergei-maertens, line 43: looking at L51, `connections` does not seem to be a dict, but a list of dicts instead. Perhaps you could define a `TypedDict` subclass to describe the shape?

sergei-maertens (changes requested):
Can you create a ticket in the backlog to set up VCR tests against Open Zaak?

All these mocks for ZGW API's are unwieldly, and we can quite easily spin up a real Open Zaak instance using docker-compose.

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsVariablesProperties.js
```diff
@@ -0,0 +1,317 @@
...
+
+      {usedComponents.length < getSimpleComponentsLength() ? (
+        <ButtonContainer onClick={onAdd}>
+          <FormattedMessage
+            description="Add process variable button"
+            defaultMessage="Add variable"
+          />
+        </ButtonContainer>
+      ) : (
+        'All simple variables are mapped to a case property.'
```
sergei-maertens, line 151: please use `<FormattedMessage>` so that this is translated correctly

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsVariablesProperties.js
```diff
@@ -0,0 +1,317 @@
...
+            onDelete={onDelete.bind(null, index)}
+            {...mappedVariable}
+          />
+        ))}
+      </ChangelistTableWrapper>
+
+      {usedComponents.length < getSimpleComponentsLength() ? (
+        <ButtonContainer onClick={onAdd}>
+          <FormattedMessage
+            description="Add process variable button"
```
sergei-maertens, line 146:
```suggestion
            description="Add zaakeigenschap (mapping) button"
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -34,10 +36,60 @@
...
 logger = logging.getLogger(__name__)
 
 
+class VariablesProperties(TypedDict):
+    component_key: str
+    eigenschap: str
+
+
+def get_property_mappings_from_submission(
+    submission: Submission, connections: list[VariablesProperties]
```
sergei-maertens, line 50:
```suggestion
    submission: Submission, mappings: list[VariablesProperties]
```
please consistently use the "mapping" variable name(s)

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -34,10 +36,60 @@
...
+    }
+
+    merged_data = dict(
+        submission.submissionvaluevariable_set.filter(
+            key__in=[conn["component_key"] for conn in connections]
+        ).values_list("key", "value")
+    )
+
+    for key, form_value in merged_data.items():
+        if key not in simple_mappings:
```
sergei-maertens, line 69: this can never be `True` because the the keys of `simple_mapping` are used as a query `filter` clause, so the check is not needed. There's also no test coverage for that code and it's not possible to get coverage, I believe.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -34,10 +36,60 @@
...
+            continue
+
+        eigenschap = simple_mappings.get(key)
```
sergei-maertens, line 72: due to the above query, `key` is guaranteed to be present in `simple_mappings`, so you can use `[key]` notation instead of `.get`, this can not raise `KeyError`

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -117,12 +176,37 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
                 )
 
-        if not ("medewerker_roltype" in attrs and "zaaktype" in attrs):
-            return attrs
-
         # We know it exists thanks to the previous check
         group_config = ZGWRegistration.get_zgw_config(attrs)
 
+        # Make sure the property (eigenschap) related to the zaaktype exists
+        if connections := attrs.get("property_mappings"):
```
sergei-maertens, line 183: s/connections/mappings

Connection makes me think of HTTP/TCP connections and especially with the client interaction further this may be confusing.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -464,6 +548,55 @@ def register_submission(self, submission: Submission, options: dict) -> dict | N
...
+                    "url": eigenschap["url"],
+                    "specificatie": eigenschap["specificatie"],
+                }
+                for eigenschap in eigenschappen
+            }
+
+            for key, value in property_mappings.items():
+                if key in retrieved_eigenschappen:
+                    processed_value = process_according_to_eigenschap_format(
+                        retrieved_eigenschappen[key].get("specificatie"), value
```
sergei-maertens, line 574:
```suggestion
                        retrieved_eigenschappen[key]["specificatie"], value
```
your dict comprehension above already guarantees that the key exists

src/openforms/registrations/contrib/zgw_apis/tests/test_utils.py
```diff
@@ -0,0 +1,75 @@
+from django.test import TestCase
```
sergei-maertens, line 1: this doesn't seem to do DB queries, so you can use `SimpleTestCase` instead which is faster :)

src/openforms/registrations/contrib/zgw_apis/tests/test_validators.py
```diff
@@ -103,3 +103,105 @@ def test_no_zgw_api_group_and_no_default(self):
...
+        }
+        serializer = ZaakOptionsSerializer(data=data)
+        is_valid = serializer.is_valid()
+
+        self.assertFalse(is_valid)
+        self.assertIn("property_mappings", serializer.errors)
+        self.assertEqual(
+            "Could not find a property with the name 'wrong variable' related to the zaaktype.",
+            serializer.errors["property_mappings"][0],
+        )
```
sergei-maertens, line 207: I would expect an error `serializer.errors["property_mappings"][0]["eigenschap"][0]`, specific to the eigenschap field.

Let's make an improvement ticket for that for the new UI/design and leave this as is now.

src/openforms/registrations/contrib/zgw_apis/utils.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    format = specificatie.get("formaat")
+    if format not in ["datum", "datum_tijd"]:
+        return value
+
+    if format == "datum":
+        try:
+            valid_date = date.fromisoformat(value)
+        except ValueError as exc:
+            raise exc
```
sergei-maertens, line 19: since you just re-raise, this could instead be:

```suggestion
        valid_date = date.fromisoformat(value)
```

src/openforms/registrations/contrib/zgw_apis/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+        processed_value = valid_date.strftime("%Y%m%d")
+
+    elif format == "datum_tijd":
+        try:
+            valid_datetime = datetime.fromisoformat(value)
+        except ValueError as exc:
+            raise exc
+
+        amsterdam_timezone = pytz.timezone("Europe/Amsterdam")
```
sergei-maertens, line 29: you can re-use `openforms.contrib.clients.utils.TIMEZONE_AMS` instead :thinking: 

It would probably be useful to have a helper function in that module:

```py

def datetime_in_amsterdam(value: datetime) -> datetime:
    return timezone.make_naive(value, timezone=TIMEZONE_AMS)

def get_today() -> str:
    now = datetime_in_amsterdam(timezone.now())
    return now.date().isoformat()
```

src/openforms/registrations/contrib/zgw_apis/tests/test_utils.py
```diff
@@ -0,0 +1,75 @@
...
+            "groep": "",
+            "formaat": "datum_tijd",
+            "lengte": "14",
+            "kardinaliteit": "1",
+            "waardenverzameling": [],
+        }
+        sample = "2024-01-31T12:07:00+01:00"
+        processed_value = process_according_to_eigenschap_format(specificatie, sample)
+
+        self.assertEqual(processed_value, "20240131130700")
```
sergei-maertens, line 31: I don't think this is correct:

* `12:07:00` in UTC + 1 is `11:07:00` in UTC
* currently (in winter time), AMS is UTC+1, so the output would be (time part): `120700`

For reference, the following code should be correct:

```py
from openforms.contrib.zgw.clients.utils import TIMEZONE_AMS
from datetime import datetime

>>> sample = datetime.fromisoformat("2024-01-31T16:24:00+00:00")
>>> sample.astimezone(TIMEZONE_AMS).strftime("%Y%m%d%H%M%S")
'20240131172400'
```

but instead, it's easier to use Django's `django.utils.timezone.make_naive` function.

In summer, it's UTC+2, so the output would be one hour later for the same sample.

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/registrations/zgw/OptionsFormFields.js
sergei-maertens: we want to make other improvements with dependent fields (zaaktype -> show available relevant documenttypen etc.) anyway at some point, so this rework would need to be done sooner or later. Moving this into our own react components makes adopting it to a UI library easier too when we implement the new design.

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
```diff
@@ -0,0 +1,63 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import Field from 'components/admin/forms/Field';
+
+import ZGWFormFields from './ZGWOptionsFormFields';
+
+const ZGWOptionsForm = ({index, name, label, schema, formData, onChange}) => {
```
Viicos, line 8: Is the `index` prop provided anywhere? The form is called here:
https://github.com/open-formulieren/open-forms/blob/beb9c7c336868abbd43b487ffa22e1d99bf3b8f5/src/openforms/js/components/admin/form_design/RegistrationFields.js#L36-L55

But I don't see any `index` argument

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.js
```diff
@@ -0,0 +1,308 @@
...
+  };
+
+  const buildErrorsComponent = field => {
+    const rawErrors = getFieldErrors(name, index, validationErrors, field);
+    return rawErrors ? getErrorMarkup(rawErrors) : null;
+  };
+
+  return (
+    <Wrapper>
+      <CustomFieldTemplate
```
Viicos, line 53: `CustomFieldTemplate` has a required `schema` prop type which is not passed. Looking at the rjsf implementation it does not actually lead to an exception but this could change at any moment

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
vaszig: Yes, this have been added as a prop by me in this PR (line 43 `index={index}`), it's not present in master.

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.js
sergei-maertens: Let's leave it at this - our dependencies are pinned and we are in no hurry to ugrade to a newer version. Most likely, we will even replace the dependency entirely with hand-rolled forms to provide a better user experience.

Let's add a comment warning/documenting this though - you can backlink to this comment with a short summary that we choose to ignore the `schema` prop as it doesn't seem to do anything.

Having some end-to-end tests to verify that the zgw forms can be configured would be good to have some protection against regressions.

sergei-maertens (approved):
Alright, let's get this in and see which feedback we'll get :)

PR #3701, opened 2023-12-19
Refactor validation plugins

Viicos:
Part of #3607
Depends on #3646

PR #3741, opened 2024-01-05
Refactor validation plugins

Viicos:
Cherry picked from https://github.com/open-formulieren/open-forms/pull/3701

### DIFF
--- a/src/openforms/contrib/kvk/validators.py
+++ b/src/openforms/contrib/kvk/validators.py
@@ -7,7 +7,8 @@ from django.utils.translation import gettext_lazy as _
 from requests import RequestException
 
 from openforms.utils.validators import validate_digits, validate_rsin
-from openforms.validations.registry import StringValueSerializer, register
+from openforms.validations.base import BasePlugin
+from openforms.validations.registry import register
 
 from .client import NoServiceConfigured, SearchParams, get_client
 
@@ -20,7 +21,7 @@ class NumericBaseValidator:
         "too_short": _("%(type)s should have %(size)i characters."),
     }
 
-    def __call__(self, value):
+    def __call__(self, value: str):
         validate_digits(value)
 
         if len(value) != self.value_size:
@@ -47,18 +48,16 @@ validate_kvk = KVKNumberValidator()
 validate_branchNumber = KVKBranchNumberValidator()
 
 
-class KVKRemoteBaseValidator:
+class KVKRemoteValidatorMixin:
     query_param: Literal["kvkNummer", "rsin", "vestigingsnummer"]
     value_label: str
-
-    value_serializer = StringValueSerializer
 
     error_messages = {
         "not_found": _("%(type)s does not exist."),
         "too_short": _("%(type)s should have %(size)i characters."),
     }
 
-    def __call__(self, value: str) -> bool:
+    def validate(self, value: str) -> bool:
         assert self.query_param
         query = cast(
             SearchParams, {self.query_param: value}
@@ -84,38 +83,43 @@ class KVKRemoteBaseValidator:
         return True
 
 
-@register("kvk-kvkNumber", verbose_name=_("KvK number"), for_components=("textfield",))
+@register("kvk-kvkNumber")
 @deconstructible
-class KVKNumberRemoteValidator(KVKRemoteBaseValidator):
+class KVKNumberRemoteValidator(KVKRemoteValidatorMixin, BasePlugin[str]):
     query_param = "kvkNummer"
     value_label = _("KvK number")
 
-    def __call__(self, value, submission):
+    verbose_name = _("KvK number")
+    for_components = ("textfield",)
+
+    def __call__(self, value: str, submission):
         validate_kvk(value)
-        super().__call__(value)
+        self.validate(value)
 
 
-@register("kvk-rsin", verbose_name=_("KvK RSIN"), for_components=("textfield",))
+@register("kvk-rsin")
 @deconstructible
-class KVKRSINRemoteValidator(KVKRemoteBaseValidator):
+class KVKRSINRemoteValidator(KVKRemoteValidatorMixin, BasePlugin[str]):
     query_param = "rsin"
     value_label = _("RSIN")
 
-    def __call__(self, value, submission):
+    verbose_name = _("KvK RSIN")
+    for_components = ("textfield",)
+
+    def __call__(self, value: str, submission):
         validate_rsin(value)
-        super().__call__(value)
+        self.validate(value)
 
 
-@register(
-    "kvk-branchNumber",
-    verbose_name=_("KvK branch number"),
-    for_components=("textfield",),
-)
+@register("kvk-branchNumber")
 @deconstructible
-class KVKBranchNumberRemoteValidator(KVKRemoteBaseValidator):
+class KVKBranchNumberRemoteValidator(KVKRemoteValidatorMixin, BasePlugin[str]):
     query_param = "vestigingsnummer"
     value_label = _("Branch number")
 
-    def __call__(self, value, submission):
+    verbose_name = _("KvK branch number")
+    for_components = ("textfield",)
+
+    def __call__(self, value: str, submission):
         validate_branchNumber(value)
-        super().__call__(value)
+        self.validate(value)
--- a/src/openforms/validations/registry.py
+++ b/src/openforms/validations/registry.py
@@ -1,6 +1,6 @@
 import dataclasses
 import logging
-from typing import Callable, Iterable, List, Type, Union
+from typing import Iterable
 
 from django.core.exceptions import ValidationError as DJ_ValidationError
 from django.utils.translation import gettext_lazy as _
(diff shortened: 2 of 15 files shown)


src/openforms/validations/registry.py
```diff
@@ -58,7 +40,7 @@ def _flat(it):
     return list(_flat(iterables))
 
 
-class Registry(BaseRegistry[RegisteredValidator]):
+class Registry(BaseRegistry[BasePlugin[T]]):
```
Viicos, line 43: Not sure having this nested type variable makes sense. I'm currently using it a few lines below:

```python
    def validate(
        self, plugin_id: str, value: T, submission: Submission
    ) -> ValidationResult:
```
But it will never be bound anyway. I guess it makes more sense to switch to:

```suggestion
class Registry(BaseRegistry[BasePlugin]):
```

src/openforms/validations/registry.py
```diff
@@ -1,50 +1,32 @@
...
-from typing import Callable, Iterable, List, Type, Union
+from typing import Iterable
 
 from django.core.exceptions import ValidationError as DJ_ValidationError
 from django.utils.translation import gettext_lazy as _
 
 import elasticapm
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError as DRF_ValidationError
+from typing_extensions import TypeVar
```
sergei-maertens, line 11: can this not be imported from `typing`?

src/openforms/validations/registry.py
sergei-maertens: yeah we typically don't instantiate `Registry` at run-time, it's just a sentinel where any type can be provided, so I don't see added value in providing the typevar.

Plugin is now a base class, but effectively it's more of a protocol with specializations for possible value types.

src/openforms/validations/registry.py
```diff
@@ -70,51 +52,14 @@ class Registry(BaseRegistry[RegisteredValidator]):
...
     @elasticapm.capture_span("app.validations.validate")
     def validate(
-        self, plugin_id: str, value: JSONValue, submission: Submission
+        self, plugin_id: str, value: T, submission: Submission
     ) -> ValidationResult:
         try:
             validator = self._registry[plugin_id]
         except KeyError:
-            logger.warning(f"called unregistered plugin_id '{plugin_id}'")
+            logger.warning("called unregistered plugin_id %s", plugin_id)
```
sergei-maertens, line 62: :pray:

src/openforms/validations/registry.py
```diff
@@ -124,15 +69,15 @@ def validate(
                 ],
             )
 
-        if not getattr(validator.callable, "is_enabled", True):
+        if not getattr(validator, "is_enabled", True):
```
sergei-maertens, line 72: this attribute should now always exist, otherwise the type checking would fail, right? And it's defined on the base class if I saw it right:

```suggestion
        if not validator.is_enabled:
```
makes grepping through the code a bit easier :)

src/openforms/validations/registry.py
```diff
@@ -151,6 +96,12 @@ def validate(
         else:
             return ValidationResult(True)
 
+    def check_plugin(self, plugin: BasePlugin) -> None:
+        if not hasattr(plugin, "value_serializer"):
+            raise ValueError(
+                f"Validator '{plugin.identifier}' must have a 'value_serializer' attribute."
+            )
```
sergei-maertens, line 103: is there a way we could enforce this via `@abc` perhaps instead of making a runtime check?

I'm also okay with changing this to an `assert ...` instead.

src/openforms/validations/validators/formats.py
```diff
@@ -1,30 +1,29 @@
...
+class PhoneNumberBaseValidator(ABC):
+    country: str | None
     country_name: str = ""
     error_message = _("Not a valid %(country)s phone number")
-    value_serializer = StringValueSerializer
-    _parse_phonenumber: ParsePhoneNumber
 
-    def __call__(self, value, submission):
+    @abstractmethod
+    def _parse_phonenumber(self, value: str) -> "PhoneNumber":
```
sergei-maertens, line 23:
```suggestion
    def _parse_phonenumber(self, value: str) -> PhoneNumber:
```
combined with `from __future__ import annotations`?

src/openforms/validations/registry.py
Viicos: Was experimenting with type var defaults. The base plugin now looks like:

```python
T = TypeVar("T", bound=JSONValue, default=str)

class BasePlugin(ABC, AbstractBasePlugin, Generic[T]):
    value_serializer: ClassVar[type[serializers.BaseSerializer]] = StringValueSerializer
```

As I'm giving a default serializer that works with strings, it makes sense to use `str` as a default to the type variable.

Let me know if you want plugins to be explicit with respect to their `value_serializer`

src/openforms/validations/registry.py
Viicos: This check can actually be removed if the behavior described in https://github.com/open-formulieren/open-forms/pull/3741#discussion_r1444796139 is fine. Otherwise, we could enforce it as an abstract property maybe

src/openforms/contrib/kvk/validators.py
```diff
@@ -84,38 +83,43 @@ def __call__(self, value: str) -> bool:
...
     value_label = _("RSIN")
 
-    def __call__(self, value, submission):
+    verbose_name = _("KvK RSIN")
+    for_components = ("textfield",)
+
+    def __call__(self, value: str, submission):
         validate_rsin(value)
-        super().__call__(value)
+        KVKRemoteBaseValidator.__call__(self, value)
```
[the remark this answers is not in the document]
Viicos: Having `KVKRemoteBaseValidator` as a subclass of `BasePlugin` would require restricting `for_components` to this base class. I think using another method name instead of `__call__` might be better?

src/openforms/validations/registry.py
sergei-maertens: I can't follow, what exactly is needed to remove this check?

src/openforms/validations/registry.py
Viicos: The `default` argument was added, and is not available in `typing.TypeVar`

src/openforms/validations/registry.py
Viicos: If we keep the default value for `value_serizlier`:

```python
value_serializer: ClassVar[type[serializers.BaseSerializer]] = StringValueSerializer
```

then all subclasses are guaranteed to have the attribute

src/openforms/contrib/kvk/validators.py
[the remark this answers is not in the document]
sergei-maertens: I don't follow - the `for_components` is semantically related for plugins/components, and technically comes down to the expected type of `value`.

However, I think changing that to `KVKRemoteValidatorMixin` with a different method name than `__call__` would make sense, I do not expect this base class to be called directly/by itself.

src/openforms/contrib/kvk/validators.py
```diff
@@ -84,38 +83,43 @@ def __call__(self, value: str) -> bool:
...
     value_label = _("KvK number")
 
-    def __call__(self, value, submission):
+    verbose_name = _("KvK number")
+    for_components = ("textfield",)
+
+    def __call__(self, value: str, submission):
         validate_kvk(value)
-        super().__call__(value)
+        super().validate(value)
```
sergei-maertens, line 97: Because of MRO reasons, mixins go to the left, base classes to the right.

It's also weird to use `super()` while you know that you're putting a mixin in there that provides a `validate` method.

```py
@register("kvk-kvkNumber")
@deconstructible
class KVKNumberRemoteValidator(KVKRemoteValidatorMixin, BasePlugin[str]):
    query_param = "kvkNummer"
    value_label = _("KvK number")

    verbose_name = _("KvK number")
    for_components = ("textfield",)

    def __call__(self, value: str, submission):
        validate_kvk(value)
        self.validate(value)
```

should be the correct implementation.

src/openforms/validations/registry.py
sergei-maertens: I thought we'd only use features from `typing_extensions` that are available in newer versions of Python - this is exactly what I was worried about when I was pushing back against using `typing_extensions`.

The fact that it's being used by people is a non-argument - it's either putting the language itself in a hostage situation where they MUST adopt something because people started using it, or leads to non-standard behaviour, which in turn can interact negatively with other tooling (for example, the github code analysis recently had issues on some edge cases of pattern matching syntax, and that was on a core language feature, so you can expect non-standard things to lead to similar/worse issues).

I really like type annotations, but I'm also really concerned about the amount of effort and workarounds/coping strategies goes into applying them and making it work all in a complex project like Open Forms. Let's not lose focus of the actual features we're building.

src/openforms/validations/registry.py
sergei-maertens: great, keep the default value.

src/openforms/validations/validators/formats.py
```diff
@@ -1,30 +1,31 @@
...
+    country: str | None
     country_name: str = ""
     error_message = _("Not a valid %(country)s phone number")
-    value_serializer = StringValueSerializer
-    _parse_phonenumber: ParsePhoneNumber
 
-    def __call__(self, value, submission):
+    @abstractmethod
+    def _parse_phonenumber(self, value: str) -> PhoneNumber:
+        pass
```
sergei-maertens, line 26: Please add the `# pragma: no cover` for the time being.

src/openforms/validations/validators/formats.py
```diff
@@ -48,19 +49,18 @@ def __call__(self, value, submission):
...
 
 
-@register(
-    "phonenumber-international",
-    verbose_name=_("International phone number"),
-    for_components=("phoneNumber",),
-)
-class InternationalPhoneNumberValidator(PhoneNumberBaseValidator):
+@register("phonenumber-international")
+class InternationalPhoneNumberValidator(PhoneNumberBaseValidator, BasePlugin[str]):
```
sergei-maertens, line 53: Probably this should also be renamed to `PhoneNumberValidatorMixin` similar to the `KVKRemoveValidatorMixin`?

The ABC base class conflicts a bit with the idea of `BasePlugin[str]` being a base too - which one dictates/enforces what? Using a `*Mixins, Base` inheritance order leans more towards composition than weird MRO challenges with multiple "true" bases.

sergei-maertens (changes requested):

src/openforms/validations/registry.py
Viicos: That's reasonable; however if we don't provide a default to the type var then we shouldn't provide a default value to the `value_serializer` imo, as it would lead to an inconsistent state between runtime and the type representation (even though nothing enforces a correct `value_serializer` to be set with respect to the solved type variable anyway)

sergei-maertens: Please let's not get hung up on these typing things and focus on getting the code functional in the first place. I find it absurd that typing challenges (of which there are many because it's an evolving thing in the language) are limiting our DRY-ness of the actual runtime code.

By all means, apply them when it's low cost, but screw the typing information if it's not. There is too much energy going in these discussions that is better spent elsewhere.

sergei-maertens (approved):

PR #3742, opened 2024-01-08
Simplify `isort` config, add global coverage excludes

Viicos:


setup.cfg
```diff
@@ -1,5 +1,3 @@
-[pycodestyle]
-[pep8]
```
sergei-maertens, line 2: Can you check that nobody actually uses these tools in their editors? I don't know about PyCharm users etc. which sections are supported and which aren't.

setup.cfg
```diff
@@ -61,3 +53,8 @@ omit =
 
 [coverage:report]
 skip_covered = True
+exclude_also =
```
sergei-maertens, line 56: can we put this in the `coverage:run` section? that might be the issue why codecov is not picking up this option - this only affects the output of `coverage html` and the like :)

sergei-maertens, line 56: maybe include `raise NotImplementedError`?

sergei-maertens:
Once this is good to go, please also update our default-project template if you haven't done so yet!

setup.cfg
Viicos: I believe `pycodestyle` needs to be installed for it to be picked up by IDEs. Anyway, we don't enforce it at the CI level

Viicos: Ah wait maybe I'll then need to edit some of the `ignore` rules that are not relevant anymore.

setup.cfg
Viicos: `exclude_also` seems to be only available in `report`, but I actually need to upgrade `coveragepy` as it is a new config value

setup.cfg
sergei-maertens: > Anyway, we don't enforce it at the CI level

We don't care which tools people use locally, but CI needs to pass with flake8. Flake8 includes a bunch of other ones as you found out, and the configuration currently doesn't break anything, but removing it may break some of people's environments. This is all hypothetical, which is why I'm asking to check wit your peers ;)

CharString (approved):

PR #3751, opened 2024-01-09
Add script to detect component issues

sergei-maertens:
Probably good to have this script at-hand so we can run it in existing environments.

### DIFF
--- a/bin/report_component_problems.py
+++ b/bin/report_component_problems.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+import sys
+from pathlib import Path
+
+import django
+
+from tabulate import tabulate
+
+SRC_DIR = Path(__file__).parent.parent / "src"
+sys.path.insert(0, str(SRC_DIR.resolve()))
+
+
+def report_problems() -> bool:
+    from openforms.forms.models import FormDefinition
+
+    problems = []
+
+    fds = FormDefinition.objects.iterator()
+    for form_definition in fds:
+        for component in form_definition.iter_components():
+            assert "key" in component
+            assert "type" in component
+
+            problem_base = [
+                form_definition.admin_name,
+                component.get("label") or component["key"],
+                component["type"],
+            ]
+
+            match component:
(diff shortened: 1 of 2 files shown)


bin/report_component_problems.py
sergei-maertens, line 1: We might make this part of the upgrade check? Some of the data migrations silently fix issues in #3710

Viicos: > Some of the data migrations silently fix issues in #3710

What kind of issues?

Viicos, line 1: Is there a reason to have this as a separate `bin` script, and not "built-in" into django as a management command? (would remove the need to setup openforms in `main`)

bin/report_component_problems.py
```diff
@@ -0,0 +1,91 @@
...
+                                ]
+                            )
+                            break
+
+    if not problems:
+        print("No problems found.")
+        return True
+
+    print("Found problems in form definition components.")
+    print("")
```
Viicos, line 69:
```suggestion
    print("Found problems in form definition components.", end="\n\n")
```
:shrug:

Viicos (approved):

bin/report_component_problems.py
sergei-maertens: if someone sets `sizeMobile: "xxs"` for example, the formio columns themselves use size designators for the bootstrap classes, which doesn't translate to our system, so there may be confusion.

We silently correct `size` and `sizeMobile` to their default value if we can't parse them.

sergei-maertens: yes, makes it easier for Sjoerd to grab it and apply it on older instances that don't have it, while not needing an entry in `INSTALLED_APPS`

bin/report_component_problems.py
Viicos: Yeah I think both are fine!

bin/report_component_problems.py
Viicos: Ah yes, then this is related to my comment on #3710

PR #3780, opened 2024-01-15
[#3758] Add docs regarding the BRK validation plugin

### LINKED ISSUE #3758: BRK/Zzakelijk gerechtigde documentation is missing
* https://open-forms.readthedocs.io/en/latest/configuration/validation/index.html only mentions the KVK validator plugins
* searching the entire documentation on the term "BRK" does not produce any results

Documentation is part of a feature and should not be ignored. It's essential for ops/administrators to correctly configure Open Forms.

Viicos:
Closes #3758

### DIFF
--- a/docs/configuration/validation/brk.rst
+++ b/docs/configuration/validation/brk.rst
@@ -1,0 +1,43 @@
+.. _configuration_validation_brk:
+
+=========================
+BRK - Zakelijk gerechtigd
+=========================
+
+The special *AddressNL* component has a specific validator available, that can
+check if the currently logged in user is the business owner of the filled-in property.
+
+Configuration
+=============
+
+1. Obtain a BRK API key.
+2. In Open Forms, navigate to: **Miscellaneous** > **KvK configuration**.
+3. Click **Add** next to the **KvK API Basisprofiel** field and fill in the following
+   details:
+
+   * **Label**: *Fill in a human readable label*, for example: ``My BRK API``
+   * **Type**: ORC (Overige).
+   * **API root URL**: *The API URL provided by KvK but typically:* ``https://api.brk.kadaster.nl/esd-eto-apikey/bevragen/v2/``.
+   * **Authorization type**: API key.
+   * **Header key**: ``X-Api-Key``.
+   * **Header value**: *The KvK API key obtained in step 1*.
+   * **OAS file**: ``https://raw.githubusercontent.com/VNG-Realisatie/Haal-Centraal-BRK-bevragen/master/specificatie/genereervariant/openapi.yaml``.
+
+4. Click **Save** in the popup to save and close it.
+5. When building a form, add a *AddressNL* component from the *Special* dropdown.
+6. In the **Validation** tab, select *BRK - Zakelijk gerechtigd*.
+
+.. warning::
+
+   This validation plugin is only available if an :ref:`authentication plugin <configuration_authentication_index>`
+   that provides a BSN is being used.
+
+
+Technical
+=========
+
+=======  ==================
+API      Supported versions
+=======  ==================
+BRK API  2.0
+=======  ==================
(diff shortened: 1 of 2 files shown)


vaszig (approved):
I think this is ok. Speaking for myself I have no idea concerning BRK but this seems enough to start with.

Viicos:
Maybe @joeribekker has some thoughts about what needs to be added to the docs?

docs/configuration/validation/brk.rst
```diff
@@ -0,0 +1,43 @@
...
+=========================
+
+The special *AddressNL* component has a specific validator available, that can
+check if the currently logged in user is the business owner of the filled-in property.
+
+Configuration
+=============
+
+1. Obtain a BRK API key.
+2. In Open Forms, navigate to: **Miscellaneous** > **KvK configuration**.
```
sergei-maertens, line 14: we tend to point people to the configuration overview page (the Misc menu entry should not be used directly) and have them click the respective configure option there. I see that the [KVK basisprofiel](https://open-forms.readthedocs.io/en/latest/configuration/prefill/kvk.html#configuration-prefill-kvk) documentation is also wrong in that regard.

A good example is: https://open-forms.readthedocs.io/en/latest/configuration/prefill/bag.html

docs/configuration/validation/brk.rst
```diff
@@ -0,0 +1,43 @@
...
+3. Click **Add** next to the **KvK API Basisprofiel** field and fill in the following
+   details:
```
sergei-maertens, line 16: I think we preferably have people add the service via **Configuration > Services** and then make them select the option in the dropdown rather than adding it through the popup/window.

docs/configuration/validation/brk.rst
```diff
@@ -0,0 +1,43 @@
...
+
+   * **Label**: *Fill in a human readable label*, for example: ``My BRK API``
```
sergei-maertens, line 18:
```suggestion
   * **Label**: *Fill in a human readable label*, for example: ``Basisregistratie Kadaster (BRK)``
```
it's a central service - see https://www.kadaster.nl/zakelijk/registraties/basisregistraties/brk

What *can* happen is that the organisation itself requires access to go through their own gateway, so a different host can be configured.

docs/configuration/validation/brk.rst
```diff
@@ -0,0 +1,43 @@
...
+   * **Type**: ORC (Overige).
+   * **API root URL**: *The API URL provided by KvK but typically:* ``https://api.brk.kadaster.nl/esd-eto-apikey/bevragen/v2/``.
+   * **Authorization type**: API key.
+   * **Header key**: ``X-Api-Key``.
+   * **Header value**: *The KvK API key obtained in step 1*.
+   * **OAS file**: ``https://raw.githubusercontent.com/VNG-Realisatie/Haal-Centraal-BRK-bevragen/master/specificatie/genereervariant/openapi.yaml``.
```
sergei-maertens, line 24: this is only required for legacy reasons right, we don't actually make use of this file anymore?

docs/configuration/validation/brk.rst
```diff
@@ -0,0 +1,43 @@
...
+
+4. Click **Save** in the popup to save and close it.
+5. When building a form, add a *AddressNL* component from the *Special* dropdown.
```
sergei-maertens, line 27:
```suggestion
5. When building a form, add an *AddressNL* component from the *Special* group.
```

docs/configuration/validation/brk.rst
```diff
@@ -0,0 +1,43 @@
...
+6. In the **Validation** tab, select *BRK - Zakelijk gerechtigd*.
+
+.. warning::
```
sergei-maertens, line 30: maybe a note instead of a warning?

sergei-maertens (changes requested):
Some fine tuning.

docs/configuration/validation/brk.rst
Viicos: I thought at first a required field, glad to remove it as it did not play well with the RTD theme

sergei-maertens (approved):

PR #3802, opened 2024-01-23
🏷️ Use new style typing syntax

Viicos:
How it was done:
- Using ruff, I've set the following rules: [`UP006`](https://docs.astral.sh/ruff/rules/non-pep585-annotation/) and [`UP007`](https://docs.astral.sh/ruff/rules/non-pep604-annotation/)
- Ran `ruff check --fix src/`: only `UP006` is considered safe
- Used [`F401`](https://docs.astral.sh/ruff/rules/unused-import/) for unused imports, which has a safe fix
- Ran black and isort

`Union` and `Optional` still remains, as the fix for `UP007` is unsafe. I believe the only place where it can lead to actual errors is when using forward references:

`JSONValue = Union[JSONPrimitive, "JSONObject", list["JSONValue"]]`

to

`JSONValue = JSONPrimitive | "JSONObject" | list["JSONValue"]`

which can be rewritten as `"JSONPrimitive | JSONObject | list[JSONValue]"`.

I would prefer having only this merged for now (after 2.5), and I can investigate further for the remaining `Union/Optional`

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,6 @@
 # adapted prettier for JS and SCSS
 a39b15a21d28ee995e861cad489f0648fc971f7e
 6115015c0cadfdff238e2a0f0871d4c7a5ac3bf7
+
+# New style typing
+f185fedaa603a1ef0285dd30e47f74218677c2c6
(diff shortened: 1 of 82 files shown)


sergei-maertens:
Nice to clean this up, we were doing this gradually before when touching the involved lines of code anyway, but having a git-ignored single fix can speed up this process.

Please rebase on master and fix the merge conflicts and then we can get this merged.

Issue #3616, opened 2023-11-17
Release 2.5 testing notes


sergei-maertens:
I'm adding some notes for release-testing (by hand) that are relevant and need to be double/triple checked.

- [x] Hash based routing (or not) @SilviaAmAm 
    - [ ] ~Validate against ODRU (see Taiga)~
    - [x] Embed a form in a "third party CMS" with hash based routing, on a subpath with the `?form=<form-id>` query string
    - [x] Ensure that a form can be started from this embed page
    - [x] Ensure that a form can be paused and resumed from this embed page
    - [x] Ensure that the co-sign flow still works (form with co-sign and have another person do the cosign check/confirmation)
    - [x] Appointment with demo plugin
    - [x] Check the appointment cancellation flow 
- [x] Test the  above flows in our own backend without hash based routing:
    - [x] Pause and resume
    - [x] Cosign
    - [x] Appointment cancellation

Test task orchestration refactor @vaszig :

- [x] Test a form with payment and cosign:
           - Ensure that upon submitting the form, no confirmation email is sent but a cosign request is sent.
           - Upon payment, a confirmation email is sent
           - Upon cosign a confirmation email is sent with the cosigner in the CC
           - Check that registration works
- [x] Test a form without payment:
            - Upon completion the confirmation email is sent
- [x] Test retrying to register submissions
- [x] Test a form with appointments

Styling refactors:  @SilviaAmAm 

- [x] Styling tests when submitting a form as registrator on behalf of a customer

PDF generation: @SilviaAmAm 

- [x] Check styling of confirmation/summary PDF (following WeasyPrint upgrade)

React 18 upgrade (#3697) @Viicos 

- [x] Check form designer (tabbed interface, adding/modifying/removing components...)
- [x] Check form definition admin (add / edit)
- [x] Design token editor for style/theme page
- [x] Plugin configuration in global configuration page
- [x] Form versions, viewing + restoring
- [x] Session timeout widget (bottom right in admin, counter + popup)

SilviaAmAm:
Styling problems noticed:
 - [ ] Dropdown (missing the arrow and a bit too short): 
![Screenshot from 2024-01-23 11-53-25](https://github.com/open-formulieren/open-forms/assets/19154114/72ded450-b006-4a58-b4cb-f437b82b3588)
[...]

PR #3804, opened 2024-01-24
[#3616] Missing margin in form definition configuration

vaszig:
Part of #3616 

This has to do with the form definition page. In a form it's working properly due to different css classes.


vaszig:
This was an existing error and will be fixed with the new design.

PR #3812, opened 2024-01-25
[#3616] Fix PDF template of submission appointment

Viicos:
Part of #3616

### DIFF
--- a/src/openforms/appointments/templates/appointments/appointment__pdf.html
+++ b/src/openforms/appointments/templates/appointments/appointment__pdf.html
@@ -8,8 +8,10 @@
       {% trans "Products" %}
     </div>
     <div class="submission-step-row__value">
-      {% for product in appointment.products %} {{ product.name }}{% if product.amount > 1 %} (x{{
-      product.amount }}){% endif %}{% if not forloop.last %}<br />{% endif %} {% endfor %}
+      {% for product in appointment.products %}
+        {{ product.name }}{% if product.amount > 1 %} (x{{product.amount }}){% endif %}
+        {% if not forloop.last %}<br />{% endif %}
+      {% endfor %}
     </div>
   </div>
   <div class="submission-step-row">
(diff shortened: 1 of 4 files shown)


src/openforms/appointments/templates/appointments/appointment__pdf.html
```diff
@@ -1,3 +1,4 @@
+{% autoescape off %}
```
Viicos, line 1: I assume this is fine, as it is only being rendered in a PDF

vaszig (approved):
Tested locally and works as expected now.

src/openforms/appointments/templates/appointments/appointment__pdf.html
Viicos: As discussed during standup: this could potentially be unsafe even in PDF files. Perhaps it makes sense to escape (the default with the DTL) the rendered variables after all. I don't expect the values coming from the appointment plugins to be HTML anyway. If necessary (e.g. for the QR code), `format_html` is called

Viicos: And I don't know if any user input are rendered in the appointment PDF, but if it is the case then it makes sense to actually escape it

sergei-maertens: The product names, location names etc. that we retrieve from external services are *also* untrusted user input, so they need to be escaped indeed.

src/openforms/appointments/tests/test_pdf.py
```diff
@@ -31,8 +32,11 @@ def test_appointment_info_is_included(self):
         self.assertIn("Afspraakinformatie", html)
         self.assertIn("Test product 1", html)
         self.assertIn("Test location", html)
+        self.assertIn("Test address", html)
         self.assertIn("Datum en tijd", html)
 
+        self.assertNotIn(escape("<h1>"), html)
```
sergei-maertens, line 38: yeah so we want to see the escaped data here, plugin developers need to explicitly `mark_safe` or `format_html` that stuff IMO.

That also requires updating the comment for `openforms.appointments.base.AppointmentDetails.other`:

> \# These are typically key/values-pairs where both the key and value are
> \# considered *not* to be HTML-safe. Plugins need to explicitly mark trusted
> \# content as safe.

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Backports:

* stable/2.5.x: 9db8fa58bc4f502389c0472fbae87a1755bb088e
* stable/2.4.x: 038fc9983002f61ef8a246be1701d0e4270bdcfd

Older versions were not affected, this template was introduced in 2.4

PR #3813, opened 2024-01-25
[#3809] Fix change view of submission

### LINKED ISSUE #3809: Unhandled exception when accessing an non existing submission detail page through the admin
Sentry 341140

https://github.com/open-formulieren/open-forms/blob/c12c51b94c55895490ba8e8c671a060d080e1586/src/openforms/submissions/admin.py#L269-L275

`get_object` returns an instance of the model or `None`. Probably just a matter of raising a 404.

(This can happen when accessing the (deleted) submission from the submission report detail page.

Viicos:
Fixes #3809

### DIFF
--- a/src/openforms/submissions/tests/test_admin.py
+++ b/src/openforms/submissions/tests/test_admin.py
@@ -293,6 +293,12 @@ class TestSubmissionAdmin(WebTest):
 
         self.assertEqual(200, response.status_code)
 
+    def test_change_view(self):
+        change_url = reverse(
+            "admin:submissions_submission_change", kwargs={"object_id": "0"}
+        )
+        self.app.get(change_url, user=self.user, status=404)
+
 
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/tests/test_admin.py
```diff
@@ -293,6 +293,13 @@ def test_search(self):
 
         self.assertEqual(200, response.status_code)
 
+    def test_change_view(self):
+        breakpoint()
```
vaszig, line 297: Let's remove this

vaszig (approved):
As discussed in the office as well, the report is deleted along with the submission deletion so it's probably an edge case, but anyway this fix makes sense generally.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3815, opened 2024-01-25
:construction_worker: Update CI action version to latest Node

sergei-maertens:


Viicos (approved):
Let's maybe raise `setup-python` to `v5` as well?

sergei-maertens:
> Let's maybe raise `setup-python` to `v5` as well?

Yep! I wanted to see which ones I'd missed :)

PR #3818, opened 2024-01-26
Use new style typing for unions

Viicos:
This is the last, I promise :pray: 

How it was done:
- the `typing.py` file was treated manually for extra safety
- Ruff rule `UP007` was applied.
- Remaining usage of `Union` and `Optional` was treated manually. This happened when forward references where used. This affected around 10 files, and `from __future__ import annotations` was used in these cases. This doesn't mean that every forward reference in the codebase was "unstrigified" with the future import, only the file where `Union` and `Optional` was used.
- `F401` applied to remove unused imports
- black and isort
- Documentation updated in one place

### DIFF
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import logging
 from dataclasses import asdict
 from functools import partial
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING
 
 from django.db.models import Model
 from django.utils import timezone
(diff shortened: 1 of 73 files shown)


src/openforms/logging/logevent.py
```diff
@@ -18,12 +20,9 @@
 from .tasks import log_logic_evaluation
 
 if TYPE_CHECKING:
+    from openforms.payments.models import SubmissionPayment
```
Viicos, line 23: `SubmissionPayment` was imported from the wrong module

sergei-maertens (approved):

sergei-maertens:
Nice :pray:

PR #3819, opened 2024-01-26
Remove logic logging

### LINKED ISSUE #3239: Remove logic rule logging

Viicos:
Fixes #3239


src/openforms/logging/templates/logging/events/submission_logic_evaluated.txt
sergei-maertens, line 1: I think we should keep this template because there are existing log records in the database. The generic timeline log list view will crash if this template is absent.

Do add a comment saying that this functionality was removed and the template is only here for legacy reasons.

src/openforms/submissions/admin.py
```diff
@@ -125,7 +122,6 @@ class SubmissionLogInline(GenericTabularInline):
     def get_queryset(self, request):
         qs = super().get_queryset(request)
         qs = qs.exclude_tag(TimelineLogTags.AVG)
-        qs = qs.filter(~Q(template="logging/events/submission_logic_evaluated.txt"))
```
sergei-maertens, line 128: we need to keep this line to ignore those log rules in the inline logs view.

I think it'd be good to create a separate issue to add a `prune_logs` management command/celery task (preferably in django-timeline-logger library itself) so we can remove logs. It's a DB table that grows over time (append only now) so we should give administrators the controls to delete logs older than X number of days or logs related to a submission that has since been deleted. So, eventually, over time, all these log records will be deleted and then we'll be able to remove these templates/log records too for full removal.

src/openforms/submissions/templates/admin/submissions/submission/logs.html
sergei-maertens, line 1: There was some Javascript code related to this file too: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/admin/submissions/filter.js

sergei-maertens (changes requested):

src/openforms/submissions/admin.py
Viicos: Issue created: https://github.com/maykinmedia/django-timeline-logger/issues/26

sergei-maertens (approved):

PR #3820, opened 2024-01-26
Feature/3680 remove auth plugin from cosign

### LINKED ISSUE #3680: Cosign V2 component configuration should not require authPlugin to be specified
### Thema / Theme

Form designer

Currently the cosign v2 component edit form requires you to specify the auth plugin to use for cosign, with the help text that it should be the same as used on the form itself for the primary person to authenticate.

image

This is obsolete configuration - if it really has to be the same method (digid vs. eherkenning etc.), then this information can be derived from the `Submission` instance itself since the method of authentication is recorded there. The form builder then does not need to specify it.

If it's a *different* method, the description is wrong, but Joeri & I both don't see a realistic use case for this.

Less configuration options -> less confusion & no more risk of conflicting configuration and the associated runtime errors.

Noticed this will implementing the component in the new builder, which should also be updated once this issue is picked up.

I'm currently implementing this field still in the new builder, on the assumption that the backend is actually using the component configuration currently and removing it would introduce a breaking change.

SilviaAmAm:
Fixes #3680 

This has a breaking change, because the `cosign_login_info` is no longer returned by the form endpoint. If we want to avoid the breaking change, we could let it return the first of the authentication methods or something else :thinking: 

SDK PR: https://github.com/open-formulieren/open-forms-sdk/pull/639
builder PR: https://github.com/open-formulieren/formio-builder/pull/118 + https://github.com/open-formulieren/formio-builder/pull/119
types PR: https://github.com/open-formulieren/types/pull/41

### DIFF
--- a/src/openforms/authentication/api/serializers.py
+++ b/src/openforms/authentication/api/serializers.py
@@ -5,7 +5,6 @@ from rest_framework import serializers
 from openforms.plugins.api.serializers import PluginBaseSerializer
 
 from ..constants import LogoAppearance
-from ..utils import get_cosign_login_info
 
 
 class TextChoiceSerializer(serializers.Serializer):
@@ -79,8 +78,3 @@ class LoginOptionSerializer(serializers.Serializer):
         ),
         read_only=True,
     )
-
-
-class CosignLoginInfoSerializer(LoginOptionSerializer):
-    def get_attribute(self, form):
-        return get_cosign_login_info(self.context["request"], form)
--- a/src/openforms/authentication/api/serializers.py
+++ b/src/openforms/authentication/api/serializers.py
@@ -1,7 +1,11 @@
 from django.utils.translation import gettext_lazy as _
 
+from furl import furl
 from rest_framework import serializers
+from rest_framework.reverse import reverse
 
+from openforms.authentication.base import LoginInfo
+from openforms.authentication.registry import register as auth_register
 from openforms.plugins.api.serializers import PluginBaseSerializer
 
 from ..constants import LogoAppearance
@@ -78,3 +82,36 @@ class LoginOptionSerializer(serializers.Serializer):
         ),
         read_only=True,
     )
+
+
+class CosignLoginInfoSerializer(LoginOptionSerializer):
+    def get_attribute(self, form):
+        if not form.cosign_component:
+            return None
+
+        auth_plugin_id = form.authentication_backends[0]
+        auth_url = reverse(
+            "authentication:start",
+            kwargs={
+                "slug": form.slug,
+                "plugin_id": auth_plugin_id,
+            },
+            request=self.context["request"],
+        )
+        next_url = reverse(
+            "submissions:find-submission-for-cosign",
+            kwargs={"form_slug": form.slug},
+            request=self.context["request"],
+        )
+        auth_page = furl(auth_url)
+        auth_page.args.set("next", next_url)
+
+        auth_plugin = auth_register[auth_plugin_id]
+
+        return LoginInfo(
+            auth_plugin.identifier,
+            auth_plugin.get_label(),
+            url=auth_page.url,
+            logo=auth_plugin.get_logo(self.context["request"]),
+            is_for_gemachtigde=auth_plugin.is_for_gemachtigde,
+        )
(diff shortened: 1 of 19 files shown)


src/openforms/js/components/admin/form_design/Warnings/MissingAuthCosignWarning.js
```diff
@@ -1,39 +1,27 @@
...
+  if (selectedAuthPlugins.length > 0) return null;
 
   const warning = {
     level: 'warning',
     message: (
       <FormattedMessage
         description="MissingAuthCosignWarning message"
-        defaultMessage="The the co-sign component requires the {plugin} authentication plugin, but this plugin is not enabled on the form."
-        values={{plugin: getAuthPluginLabel(cosignComponent.authPlugin)}}
+        defaultMessage="The the co-sign component requires at least one authentication plugin to be enabled."
```
vaszig, line 17: Remove `the` and update translations?

vaszig (changes requested):
I don't see something weird except for the small remarks and the update to the newer version of the form-builder.

Maybe this should be reviewed by Sergei as well, in case I am missing something.

src/openforms/js/components/admin/form_design/Warnings/MissingAuthCosignWarning.js
```diff
@@ -1,39 +1,27 @@
...
-  if (selectedAuthPlugins.includes(cosignComponent.authPlugin)) return null;
-
-  const getAuthPluginLabel = pluginId => {
-    for (const plugin of plugins.availableAuthPlugins) {
-      if (plugin.id === pluginId) return plugin.label;
-    }
-
-    return pluginId;
-  };
+  if (selectedAuthPlugins.length > 0) return null;
```
sergei-maertens, line 10: how does this affect the cosign v1 component?

sergei-maertens (commented):
Does this now mean that the cosigner can select any of the enabled authentication backends on the form? Or does it mean that they have to use the same backend as the one who originally submitted the form?

src/openforms/authentication/api/serializers.py
```diff
@@ -79,8 +78,3 @@ class LoginOptionSerializer(serializers.Serializer):
         ),
         read_only=True,
     )
-
-
-class CosignLoginInfoSerializer(LoginOptionSerializer):
-    def get_attribute(self, form):
-        return get_cosign_login_info(self.context["request"], form)
```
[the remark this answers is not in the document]
SilviaAmAm: I don't think it would be too tricky, I will add it back.

src/openforms/forms/models/form.py
```diff
@@ -415,11 +415,19 @@ def get_authentication_backends_display(self):
...
+    def get_cosign_component(self) -> Component | None:
         for component in self.iter_components():
             if component["type"] == "cosign":
                 return component
 
+    @property
+    def cosign_component(self) -> Component | None:
+        if not self._cosign_component:
+            self._cosign_component = self.get_cosign_component()
+        return self._cosign_component
```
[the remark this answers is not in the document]
SilviaAmAm: Ohh im a silly :( I will add the sentinel

src/openforms/js/components/admin/form_design/Warnings/MissingAuthCosignWarning.js
SilviaAmAm: Urgh, no warning anymore :S I shall add this back

SilviaAmAm: Oh! But checking the code, this never showed a warning for the old cosign. I tried it on the test env now, and no warning is displayed for the cosign V1 if the auth plugin is not configured :thinking:

SilviaAmAm:
> Does this now mean that the cosigner can select any of the enabled authentication backends on the form? Or does it mean that they have to use the same backend as the one who originally submitted the form?

The cosigner can select any of the enabled authentication backends (the user has to login before the submission is retrieved, so we don't know with which plugin the main person logged in).

vaszig (approved):

PR #3827, opened 2024-01-30
[#3793] Change payment reference number

### LINKED ISSUE #3793: As customer I want to see the form reference on the payment transaction
Ref: Taiga UTR 159

The referencenumber (`OF-<xxxxxx>` or `<zaaknummer>`) should be set as payment reference. Currently, OF uses a separately generated number.

If we have payment retries we could do: `{{ zaaknummer }}_{{ poging }}` if needed

Since the transaction numbers need to be unique, we need to make sure the OF-numbers are "made unique" maybe by adding a prefix or whatever.

**criteria**

* [ ] The payment reference needs to be globally unique (the OF public reference number gets reused over time).
* [ ] The OF-reference needs to be in the payment reference
* [ ] Add VCR tests
* [ ] Determine if an attempt postfix is needed (for example when you cancel the first payment, and make another attempt).

Viicos:
Fixes #3793


src/openforms/payments/contrib/ogone/tests/test_client.py
```diff
@@ -1,14 +1,62 @@
...
+        info = client.get_payment_info(
+            "xyz2024_OF-123456_987654321",
+            1000,
+            "http://foo.bar/return?bazz=buzz",
+            RETURN_ACTION_PARAM,
+        )
+
+        payment_request = requests.post(info.url, data=info.data)
+
+        self.assertNotIn("Er is een fout opgetreden", payment_request.text)
```
Viicos, line 38: Not a pretty assertion but the HTTP status code is always 200

Viicos:
> Please add the steps to the internal taiga as we did with the other VCR recordings.

Good catch

src/openforms/payments/tests/factories.py
```diff
@@ -8,12 +8,15 @@
...
+        SubmissionFactory, with_public_registration_reference=True
+    )
     plugin_id = "demo"
     plugin_options = {"foo": 123}
-    order_id = factory.Sequence(lambda n: int(f"2020{n:05}"))
     amount = Decimal("10.00")
-    public_order_id = factory.LazyAttribute(lambda obj: f"ref{obj.order_id}")
+    public_order_id = factory.LazyAttributeSequence(
+        lambda obj, n: f"{obj.submission.public_registration_reference}_{n}"
+    )
```
sergei-maertens, line 19: can't we invoke a method/function instead that implements the real behaviour? There are things like `PostGenerationMethodCall` available.

src/openforms/payments/tests/test_models.py
```diff
@@ -12,86 +13,69 @@
...
-        config = GlobalConfiguration.get_solo()
-        config.payment_order_id_prefix = ""
-        config.save()
 
         submission = SubmissionFactory.create(
             form__payment_backend="plugin1",
             form__payment_backend_options=options,
+            public_registration_reference="OF-123456",
         )
         # create some existent records
```
sergei-maertens, line 36:
```suggestion
        # create some pre-existing records
```

sergei-maertens (changes requested):
Some fine-tuning, looks good otherwise!

src/openforms/payments/contrib/ogone/tests/test_client.py
```diff
@@ -1,14 +1,62 @@
...
+
+from openforms.utils.tests.vcr import OFVCRMixin
+
 from ..client import OgoneClient
 from ..plugin import RETURN_ACTION_PARAM
 from .factories import OgoneMerchantFactory
 
 
-class OgoneClientTest(TestCase):
+class OgoneClientTest(OFVCRMixin, TestCase):
```
sergei-maertens, line 15: In addition to the Taiga documentation, can you also update the [release issue template](https://github.com/open-formulieren/open-forms/blob/master/.github/ISSUE_TEMPLATE/prepare-release.md)?

src/openforms/payments/contrib/ogone/tests/test_client.py
```diff
@@ -1,14 +1,62 @@
...
+    VCR_TEST_FILES = Path(__file__).parent / "files"
+
+    def setUp(self):
+        super().setUp()
+        self.merchant = OgoneMerchantFactory(
+            pspid=os.getenv("OGONE_PSPID", "maykinmedia"),
+            sha_in_passphrase=os.getenv("OGONE_SHA_IN", "placeholder_sha_in"),
+            sha_out_passphrase=os.getenv("OGONE_SHA_OUT", "placeholder_sha_out"),
+        )
```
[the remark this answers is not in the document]
vaszig: aaah that makes sense!I was wondering why we use create since this can be considered as obvious!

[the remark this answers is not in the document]
Viicos: Got misled again here: wanted to do a `build`, but got used to the Django model instantiation (which does the same). It would have been great if factory boy used the `build` strategy on instantiation instead of `create`

src/openforms/payments/models.py
```diff
@@ -36,37 +35,34 @@ def create_for(
...
         config = GlobalConfiguration.get_solo()
-        prefix = config.payment_order_id_prefix
-        prefix = prefix.replace("{year}", str(payment.created.year))
-        order_id = str(payment.order_id).rjust(ORDER_ID_PAD_LENGTH, "0")
-        return prefix + order_id
+        prefix = config.payment_order_id_prefix.format(year=payment.created.year)
+        if prefix:
+            prefix = f"{prefix}_"
+        return (
+            f"{prefix}{payment.submission.public_registration_reference}_{payment.pk}"
```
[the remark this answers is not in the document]
Viicos: :+1: Ogone allows it

src/openforms/payments/tests/factories.py
Viicos: `PostGenerationMethodCall` only allows method on the model itself. The thing is the real behavior will call `GlobalConfiguration.get_solo()`, so I'm not sure this is a desirable side effect

sergei-maertens: I'm okay with wrapping it in a `unittest.mock.patch` CM. Via LazyAttribute this should be possible :)

src/openforms/submissions/tests/factories.py
```diff
@@ -56,7 +69,7 @@ class Params:
                 lambda s: s.completed_on - timedelta(hours=4)
             ),
             pre_registration_completed=True,
-            price=factory.PostGenerationMethodCall("calculate_price"),
```
Viicos, line 59: `PostGenerationMethodCall` is somewhat counterintuitive to use. Calling:

`SubmissionFactory.create(completed=True, price=Decimal("1"))` will not result in `submission.price == Decimal("1")`. Instead, `Decimal("1")` is passed to the `calculate_price` method as a positional argument. For the record, here is the signature:

```python
def calculate_price(self, save=False): ...  # save would take the value of `Decimal("1")` with the example above.
```

By switching to `factory.PostGenerationMethodCall("calculate_price", save=False)`, it would lead to a test failure where `SubmissionFactory.create(completed=True, price=Decimal("1"))` was used. The test in question was accidentally still passing :o.

Went for `PostGeneration` instead which is more verbose but more intuitive.

Viicos:
Tests passed without https://github.com/open-formulieren/open-forms/pull/3827/commits/7c5de5ccc4fe372b7514ee878575f0f44361e8f0, isn't there any tests/checks to have admin fields actually present on the model?

sergei-maertens:
> Tests passed without [7c5de5c](https://github.com/open-formulieren/open-forms/commit/7c5de5ccc4fe372b7514ee878575f0f44361e8f0), isn't there any tests/checks to have admin fields actually present on the model?

unfortunately not, since they can also be lazy/callables to a number of places (model methods, admin methods and perhaps even plain python callables isntead of string references)

src/openforms/payments/models.py
```diff
@@ -36,37 +35,45 @@ def create_for(
...
-        prefix = prefix.replace("{year}", str(payment.created.year))
-        order_id = str(payment.order_id).rjust(ORDER_ID_PAD_LENGTH, "0")
-        return prefix + order_id
+        prefix = config.payment_order_id_prefix.format(year=payment.created.year)
+        if prefix:
+            prefix = f"{prefix}/"
+
+        reference = payment.submission.public_registration_reference
+        if not reference and config.wait_for_payment_to_register:
+            from openforms.submissions.public_references import (  # circ. ref
```
sergei-maertens, line 61: this codepath appears to not be covered by tests, but I'd reckon that it's pretty important for what will likely be the popular configuration (register after payment)

sergei-maertens (changes requested):

src/openforms/payments/tests/test_models.py
```diff
@@ -12,86 +13,110 @@
...
+        "openforms.submissions.public_references.get_random_reference",
+        return_value="OF-123456",
+    )
+    def test_create_for_no_registration(self, m1: MagicMock, m2: MagicMock):
+        amount = Decimal("11.25")
+        options = {
+            "foo": 123,
+        }
+
+        submission = SubmissionFactory.create(completed_not_preregistered=True)
```
sergei-maertens, line 105: @SilviaAmAm is this right? a deferred registration because it's not paid yet, will it also not pre-register?

Viicos: The special code in payments was added because I had a test failure where this view:
https://github.com/open-formulieren/open-forms/blob/9223c82e4ac480ac60c1f54c905ef9c76c5a6e1a/src/openforms/payments/views.py#L99-L126

would call `create_for` for a submission without a reference. This was coming from a test related to this view I think, but maybe that test wasn't realistic and should be updated?

sergei-maertens: yes, take a look at that test (and check with @SilviaAmAm ) if that is a representative setup - I just fixed another test in Vasileios' PR that had a wrong setup and caused test failure.

src/openforms/submissions/tests/factories.py
```diff
@@ -126,6 +139,10 @@ class Params:
                 status=PaymentStatus.registered,
             ),
         )
+        with_public_registration_reference = factory.Trait(
+            registration_success=True,
+            public_registration_reference=factory.LazyFunction(get_random_reference),
+        )
```
sergei-maertens, line 145: I think this implies `completed=True` too since a public registration reference can only created from the on-post-completion event.

sergei-maertens (changes requested):

src/openforms/tests/e2e/database_sync_to_async.py
```diff
@@ -0,0 +1,53 @@
...
+
+from django.db import close_old_connections
+
+from asgiref.sync import SyncToAsync
+
+
+class DatabaseSyncToAsync(SyncToAsync):
+    """
+    SyncToAsync version that cleans up old database connections when it exits.
+    """
```
sergei-maertens, line 42: looks familiar :) https://github.com/maykinmedia/zgw-consumers/blob/main/zgw_consumers/concurrent.py#L27

sergei-maertens (approved):
Let's hope this is the one :crossed_fingers:

PR #3831, opened 2024-01-30
[#3730] Part 1 - DMN logic action backend

### LINKED ISSUE #3730: Integrate DMN evaluation into form logic

**Functional UX**

> 
>

    - employment status (yes/no)
    - age (number)
    - yearly income (currency)
3. Navigate to logic tab
4. Add complex logic rule
    - map the input parameters:
        - `age`: `age`
        - `income`: `yearly income`
    - map the output parameters:
    - Click the "save" button
6. Add a second logic rule

**Technical tasks**

Must have

    - must be added to storybook

    - must be added to storybook

- [x] Documentation updates

Should have

Nice to have

**Resources**

SilviaAmAm:
Fixes #3730 (partially)

This implements the "evaluate_dmn" logic action in the backend.


src/openforms/submissions/logic/actions.py
```diff
@@ -308,6 +309,57 @@ def get_action_log_data(
...
+class EvaluateDMNAction(ActionOperation):
+    input_mapping: dict[str, str]
+    output_mapping: dict[str, str]
+    definition_id: str
+    plugin_id: str
+    definition_version: str = ""
+
+    @classmethod
+    def from_action(cls, action: ActionDict) -> "ActionOperation":
+        dmn_config: DMNConfig = action["action"]["value"]
```
sergei-maertens, line 330: does it make sense to use a `value` key here? I'd rather expect something like `"config"` as key or even no nesting at all. Perhaps this is something for the future since it deals with the polymorphism of logic actions :thinking:

src/openforms/submissions/logic/actions.py
```diff
@@ -308,6 +309,57 @@ def get_action_log_data(
...
+
+        return cls(**dmn_config)
+
+    def eval(
+        self,
+        context: DataMapping,
+        log: Callable[[JSONValue], None],
```
sergei-maertens, line 337: I think this is now gone-ish with the removal of logic rule logging? @Viicos can you tell more?

src/openforms/submissions/logic/actions.py
```diff
@@ -308,6 +309,57 @@ def get_action_log_data(
...
+        submission: Submission,
+    ) -> DataMapping | None:
+        # Mapping from form variables to DMN inputs
+        dmn_inputs = {
+            key_dmn: glom(context, key_form)
```
sergei-maertens, line 342: this will probably cause problems with optional fields where deep nested keys don't exist? Any chance this looks like or can be brought into the shape of the `FormioData` data structure? It allows you to do `data[formio_key]` without having to worry about the `glom` implementation details.

src/openforms/submissions/logic/actions.py
```diff
@@ -308,6 +309,57 @@ def get_action_log_data(
...
+            plugin_id=self.plugin_id,
+        )
+
+        log({"dmn_inputs": dmn_inputs, "dmn_outputs": dmn_outputs})
+
+        # Map DMN output to form variables
+        return {
+            key_form: dmn_outputs[key_dmn]
+            for key_form, key_dmn in self.output_mapping.items()
+        }
```
sergei-maertens, line 360: does this return value work correctly with deep keys like `foo.bar`?

sergei-maertens (commented):
Some clarification questions mostly

src/openforms/submissions/logic/actions.py
Viicos: After taking a look at the `submissions.logic` module, I think it can safely be removed. I'll do it in a separate PR once this is merged

src/openforms/submissions/logic/actions.py
Viicos: As I will probably need something similar for the Objects API ticket, wouldn't it make sense to have the `FormioData` user dict made "generic"? (By "generic" I just mean changing the name to `GlomableDict` or similar)

src/openforms/submissions/logic/actions.py
SilviaAmAm: Initially I had done it with a 'config' key, but then saw that the `ServiceFetchAction` had the config in the value, so I followed that in order to be consistent.

src/openforms/submissions/logic/actions.py
sergei-maertens: careful - when I added that datastructure I tried updating the code to make use of it in more places and the test suite broke spectacularly :grimacing: it's probably a lot more work than you expect.

First I'd like to see this datastructure properly adapted everywhere and the raw `glom` usage replaced, then we can consider making it more generic. And about that... I'd like to hide the implementation details that `glom` is used and so not include `Glom` in the name. The purpose of the datastructure is mirroring how formio submission data is processed if there are <kbd>.</kbd> characters in their component/variable names, and glom is just a means to an end for that.

We're bound to the formio data model, and all the rest follows from that.

src/openforms/submissions/logic/actions.py
SilviaAmAm: The second test added in src/openforms/submissions/tests/form_logic/test_modify_variables.py (line 503) tests the nested variable!

src/openforms/submissions/logic/actions.py
sergei-maertens: `ServiceFetchAction` is weird because it persists the configuration on the variable itself (which makes sense!) and only stores a reference to the service fetch configuration. I missed during review that it was stored in `value` too which I don't think is the best fit, there a `config` key would also be a better name.

`value` was really intended for the expression (JsonLogic) to produce the value based on other inputs - or a fixed value. It's being abused now when we added more features :grimacing:

SilviaAmAm: Ok! Then I will move it to a config key!

src/openforms/submissions/logic/actions.py
Viicos: > careful - when I added that datastructure I tried updating the code to make use of it in more places and the test suite broke spectacularly 😬 it's probably a lot more work than you expect.

yes we would have to be careful and look at the `lodash.get` implementation. In fact, `FormioData` is currently inconsistent with lodash as glom does not support `[0]` array indexing

sergei-maertens (approved):

PR #3832, opened 2024-01-30
[#3730] Part 2 - DMN config panel

### LINKED ISSUE #3730: Integrate DMN evaluation into form logic

SilviaAmAm:
Fixes #3730 (partially)

Screenshot from storybook:
Screenshot from 2024-02-01 16-53-46

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
+++ b/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
@@ -1,0 +1,191 @@
+import {Field, Form, Formik, useFormikContext} from 'formik';
+import PropTypes from 'prop-types';
+import React, {useContext, useState} from 'react';
+import {FormattedMessage} from 'react-intl';
+import {useAsync} from 'react-use';
+
+import {FormContext} from 'components/admin/form_design/Context';
+import {
+  DMN_DECISION_DEFINITIONS_LIST,
+  DMN_DECISION_DEFINITIONS_VERSIONS_LIST,
+} from 'components/admin/form_design/constants';
+import {get} from 'utils/fetch';
+
+import VariableMapping from './VariableMapping';
+import {EMPTY_OPTION} from './constants';
+import {inputValuesType} from './types';
+
+const DecisionDefinitionIdField = () => {
+  const {values, setFieldValue} = useFormikContext();
+  const [decisionDefinitions, setDecisionDefinitions] = useState([]);
+
+  useAsync(async () => {
+    if (!values.pluginId) {
+      setFieldValue('decisionDefinitionId', '');
+      setDecisionDefinitions([]);
+      return;
+    }
+
+    const response = await get(DMN_DECISION_DEFINITIONS_LIST, {
+      engine: values.pluginId,
+    });
+
+    const definitionChoices = response.data.map(item => [item.id, item.label]);
+    setDecisionDefinitions(definitionChoices);
+
+    // If pluginId has changed, reset the value of the definition
+    if (
+      values.decisionDefinitionId !== '' &&
+      !definitionChoices.find(item => item[0] === values.decisionDefinitionId)
+    ) {
+      setFieldValue('decisionDefinitionId', '');
+    }
+  }, [values.pluginId, setFieldValue]);
+
+  return (
+    <>
+      <label htmlFor="decisionDefinitionId">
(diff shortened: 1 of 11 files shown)


src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -0,0 +1,232 @@
...
+
+import VariableMapping from './VariableMapping';
+import {inputValuesType} from './types';
+
+const ERRORS = {
+  required: defineMessage({
+    description: 'Required error message',
+    defaultMessage: 'This field is required.',
+  }),
+};
```
sergei-maertens, line 24: in the future we should use Zod for client-side validation.

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -0,0 +1,232 @@
...
+    </Field>
+  );
+};
+
+const DecisionDefinitionVersionField = () => {
+  const {
+    values: {pluginId, decisionDefinitionId},
+    setFieldValue,
+    getFieldProps,
+  } = useFormikContext();
```
sergei-maertens, line 85: with the low-level usage, perhaps using `useField(...)` is more appropriate?

You'll also need `useFormikContext` to get access to the other field values, but `useField` has the advantage that is also registers the field with the formik state.

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -0,0 +1,232 @@
...
+
+const ERRORS = {
+  required: defineMessage({
+    description: 'Required error message',
+    defaultMessage: 'This field is required.',
+  }),
+};
+
+const DecisionDefinitionIdField = () => {
+  const {values, setFieldValue, getFieldProps, errors, touched} = useFormikContext();
```
sergei-maertens, line 27: Please consider using `useField` from formik, see remark below for the other field.

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -0,0 +1,232 @@
...
+                  }
+                >
+                  <Select
+                    id="pluginId"
+                    name="pluginId"
+                    allowBlank={true}
+                    choices={plugins.availableDMNPlugins.map(choice => [choice.id, choice.label])}
+                    {...formik.getFieldProps('pluginId')}
+                  />
+                </Field>
```
sergei-maertens, line 178: can we make this a component that automatically selects the only option if there's only one? :) You can then also make the field readonly

src/openforms/js/components/admin/form_design/logic/actions/dmn/VariableMapping.js
```diff
@@ -0,0 +1,105 @@
...
+                    <Field
+                      name={`${mappingName}.${index}.formVar`}
+                      htmlFor={`${mappingName}.${index}.formVar`}
+                    >
+                      <Select
+                        id={`${mappingName}.${index}.formVar`}
+                        name={`${mappingName}.${index}.formVar`}
+                        allowBlank={true}
+                        choices={formVariables.map(variable => [variable.key, variable.label])}
+                        {...getFieldProps(`${mappingName}.${index}.formVar`)}
```
sergei-maertens, line 60: I think this already includes the `name` prop so you can probably remove it from line 57?

sergei-maertens (approved):
Some small remarks

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
SilviaAmAm: Added this to the 'nice to have'

Issue #3688, opened 2023-12-13
Objects API registration backend - Improve JSON content field / mapping editor


alextreme:
### Thema / Theme

Plugins

### Omschrijving / Description

Taiga DH ZGW 539

Currently the JSON Schema of the 'productaanvraag' isn't very strict. This allows the Form Designer in Open Formulieren to easily make changes to the form, however the consequence is that the data in the 'productaanvraag' changes. Fieldname changes for instance cause the data to be placed at a different location or with a different type, leading to problems in GZAC when processing the 'productaanvraag'

As such, the GZAC team would like to adhere to a stricter JSON schema, this to ensure that the fieldnames stay the same. The JSON schema is to be a more rigid contract between Open Formulieren and GZAC.

In order to assist with this Open Formulieren has a 'JSON content editor' field in the Objects API registration backend. This is documented in:

https://open-forms.readthedocs.io/en/latest/manual/templates.html?highlight=Objects%20api#objecten-api-registratie

The problem is that this 'JSON content editor' is rather basic and misses the ability to validate the form against the 'productaanvraag', an editor doesn't know if changes made in the form or in the editor will lead to a valid 'productaanvraag', the only way to currently check this is by manually sending in a form submission. Adding a validator will ensure that changes in the form and editor can be properly verified.

## Analysis

The object type definition and the associated json schema is leading. In this user story, we'll update the UI to configure the data sent to the objects API to remove human errors from the process as much as possible and run validation at form-design time rather than see errors during runtime.

This will prepare us in the direction of the new UI design - while staying backwards compatible.

Some topics have no clear answers yet and we will get to them in due time:

* what does the UI/process look like when updating to a new version of the object type?
* how to deal with the array of uploaded attachment URLs in the documenten API?
* how to (automatically migrate from the legacy system to the new approach?
* what to do with composite data, e.g. field1 concatenated with field2, and registering that somewhere? (user defined variable is now an option)
* include common composite operations -> let's first tackle the basis
* datatype validation - we don't have information about the type inside an object/array -> yolo

### Tasks

This rework touches many aspects - please break them up in separate PRs that are easy to review and merge without breaking the current situation.

**Phase 1: User interface 1**

The changes here are useful for the legacy version too.

- [x] Add a field `objecttypes_service` to the `ObjectsAPIConfig` model. We will use this to retrieve a list of available object types and present them to the user.
[...]

PR #3836, opened 2024-01-31
Add a new `objecttypes_service`

Viicos:
Part of #3688, tackes the three first checkboxes of phase 1

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/client.py
+++ b/src/openforms/registrations/contrib/objects_api/client.py
@@ -33,7 +33,7 @@ def get_objecttypes_client() -> ObjecttypesClient:
     config = ObjectsAPIConfig.get_solo()
     assert isinstance(config, ObjectsAPIConfig)
     if not (service := config.objecttypes_service):
-        raise NoServiceConfigured("No Objects API service configured!")
+        raise NoServiceConfigured("No Objecttypes API service configured!")
     return build_client(service, client_factory=ObjecttypesClient)
 
 
--- a/src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
+++ b/src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
@@ -39,47 +39,43 @@ def create_objecttypes_api_service(
 
     # If no default objecttypes URL was configured, get it from form registration backends:
     if not objecttypes_url:
-        form_registration_urls: set[str] = set()
+        form_registration_bases_urls: set[str] = set()
 
         FormRegistrationBackend = apps.get_model("forms", "FormRegistrationBackend")
         for form_registration_backend in FormRegistrationBackend.objects.filter(
             backend=OBJECTS_API_PLUGIN_IDENTIFIER
         ).iterator():
             if url := form_registration_backend.options.get("objecttype"):
-                form_registration_urls.add(url)
+                index = url.find("/api/")
+                if index != -1:
+                    objecttypes_url = url
+                    form_registration_bases_urls.add(url[:index])
 
-        if len(form_registration_urls) >= 2:
+        if len(form_registration_bases_urls) >= 2:
             logger.warning(
                 "Found %d different Objecttypes URLs in form registration backend options",
-                len(form_registration_urls),
+                len(form_registration_bases_urls),
             )
-
-        if form_registration_urls:
-            objecttypes_url = form_registration_urls.pop()
 
     if objecttypes_url:
         Service = apps.get_model("zgw_consumers", "Service")
 
-        root_url = furl(objecttypes_url).remove(fragment=True, query=True)
-        path_segments = root_url.path.segments
+        try:
+            base, _ = re.split(r"api/v\d", objecttypes_url, 1)
+        except ValueError:
+            logger.warning(
+                "URL %s is not compliant, skipping migration", objecttypes_url
+            )
+            return
 
-        # Keep the relevant part of the path,
-        # e.g. /path/v2/other/path/api/v2/objecttypes/uuid -> /path/v2/other/path/api/v2/
-        for i, path in enumerate(reversed(path_segments)):
-            if re.match(r"v\d", path):
-                root_url.set(path=path_segments[: len(path_segments) - i])
-                break
+        api_version = re.search(r"/api/v(\d)", objecttypes_url).group(1)
+        base += f"api/v{api_version}/"
 
-        # This "path normalization" is normally done in the `Service.save` method,
-        # but we are dealing with "fake" models during migrations.
-        if not root_url.path.isdir:
-            root_url.path.segments.append("")  # Empty string means leading '/'
-
-        objects_api_config.objecttypes_service = Service.objects.create(
-            api_type=APITypes.orc,
-            api_root=str(root_url),
-            oas=str(root_url / "schema/openapi.yaml"),
+        service, _ = Service.objects.get_or_create(
+            api_root=base,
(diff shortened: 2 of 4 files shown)


src/openforms/registrations/contrib/objects_api/checks.py
```diff
@@ -4,13 +4,23 @@
...
+    get_objects_client,
+    get_objecttypes_client,
+)
 from .models import ObjectsAPIConfig
 
 
 def check_objects_service():
     with get_objects_client() as client:
-        client.get("objects")
+        client.get("objects", params={"pageSize": 1})
```
Viicos, line 18: What about doing a head request instead?

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -2,7 +2,7 @@
 
 from zgw_consumers.nlx import NLXClient
 
-from .utils import pagination_helper
+from openforms.utils.api_clients import pagination_helper
```
sergei-maertens, line 5: FYI, this will be included in the next version of zgw-consumers!

src/openforms/registrations/contrib/objects_api/checks.py
sergei-maertens: `HEAD` is not listed in the API spec: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/maykinmedia/objects-api/2.2.0/src/objects/api/v2/openapi.yaml#tag/objects

src/openforms/registrations/contrib/objects_api/client.py
```diff
@@ -22,14 +21,22 @@ class NoServiceConfigured(RuntimeError):
...
     if not (service := config.objects_service):
         raise NoServiceConfigured("No Objects API service configured!")
     return build_client(service, client_factory=ObjectsClient)
 
 
+def get_objecttypes_client() -> ObjecttypesClient:
+    config = ObjectsAPIConfig.get_solo()
+    assert isinstance(config, ObjectsAPIConfig)
+    if not (service := config.objecttypes_service):
+        raise NoServiceConfigured("No Objects API service configured!")
```
sergei-maertens, line 36: missing coverage, so there's some check tests missing :)

src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
```diff
@@ -0,0 +1,95 @@
...
+        # Extra safety, even though it shouldn't happen
+        return
+
+    objecttypes_url = objects_api_config.objecttype
+
+    # If no default objecttypes URL was configured, get it from form registration backends:
+    if not objecttypes_url:
+        form_registration_urls: set[str] = set()
+
+        FormRegistrationBackend = apps.get_model("forms", "FormRegistrationBackend")
```
sergei-maertens, line 44: please explicitly add the forms app to the migration dependencies :)

src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
```diff
@@ -0,0 +1,95 @@
...
+            backend=OBJECTS_API_PLUGIN_IDENTIFIER
+        ).iterator():
+            if url := form_registration_backend.options.get("objecttype"):
+                form_registration_urls.add(url)
+
+        if len(form_registration_urls) >= 2:
+            logger.warning(
+                "Found %d different Objecttypes URLs in form registration backend options",
+                len(form_registration_urls),
+            )
```
sergei-maertens, line 55: it's likely this will happen, and it's not a problem as long as they have a common base (e.g. `https://example.com/objecttypes/api/v2/...`)

I'd only emit a warning if there are different bases found, and you can be very pragmatic here and scan for the first occurence of `/api/`

src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
```diff
@@ -0,0 +1,95 @@
...
+
+        root_url = furl(objecttypes_url).remove(fragment=True, query=True)
+        path_segments = root_url.path.segments
+
+        # Keep the relevant part of the path,
+        # e.g. /path/v2/other/path/api/v2/objecttypes/uuid -> /path/v2/other/path/api/v2/
+        for i, path in enumerate(reversed(path_segments)):
+            if re.match(r"v\d", path):
+                root_url.set(path=path_segments[: len(path_segments) - i])
+                break
```
sergei-maertens, line 71: isn't it simpler to do `base, rest = path.split("/api/v2", 1)`?

```py
>>> url = "https://objecttypes.example.com/subpath/api/v2/objecttypes/123"
>>> base, rest = url.split("/api/v2/", 1)
>>> f"{base}/api/v2/"
'https://objecttypes.example.com/subpath/api/v2/'
```

furl is a bit overkill here IMO, we know what objecttypes API urls look like and if garbage was provided, then we output garbage again. It's not our job to fix misconfiguration errors.

src/openforms/contrib/zgw/clients/catalogi.py
Viicos: Ah nice to hear, I've had to use some `typing_extensions` features in our current implementation (not in draft this time :stuck_out_tongue:) but this is not a big deal without it

src/openforms/registrations/contrib/objects_api/checks.py
Viicos: Let's keep it this way then

src/openforms/utils/api_clients.py
```diff
@@ -0,0 +1,35 @@
+from typing import Generic, Iterator, TypeVar
+
+from ape_pie import APIClient
+from typing_extensions import TypedDict  # py3.11 import from typing
```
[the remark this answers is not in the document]
Viicos: Correct, and documentation says generic TDs are supported on >= 3.11 (tested locally on 3.10 and it indeed fails)

src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
```diff
@@ -0,0 +1,92 @@
...
+        try:
+            base, _ = re.split(r"api/v\d", objecttypes_url, 1)
+        except ValueError:
+            logger.warning(
+                "URL %s is not compliant, skipping migration", objecttypes_url
+            )
+            return
+
+        api_version = re.search(r"/api/v(\d)", objecttypes_url).group(1)
+        base += f"api/v{api_version}/"
```
sergei-maertens, line 72: I still think this is overly complicated and doesn't require regex since we only need to account for the current situation of available versions (as the mgiration only applies to existing instances), but sure, let's go with this.

Also, note that `\d` in python regex also captures non-arabic numerals and can lead to interesting surprises :) 

```py
>>> malayalam = "൦"
>>> import re
>>> re.search(r"\d", malayalam)
<re.Match object; span=(0, 1), match='൦'>
```

sergei-maertens (approved):

src/openforms/registrations/contrib/objects_api/migrations/0011_create_objecttypesypes_service_from_url.py
Viicos: Yes but I wasn't sure if the v1 API was still used or not. If both v1 and v2 can exist, how should this be handled then? Should the endpoint be forced to use v2?

PR #3837, opened 2024-02-01
🏷️ Improve generic type hinting of registry

Viicos:
This is a simple follow up of https://github.com/open-formulieren/open-forms/pull/3630/.

This isn't a perfect solution either, but got hit by this when dealing with such code:

```python
plugin = ObjectsAPIRegistration("...")  # ObjectsAPIRegistration is a subclass of the base plugin class

reveal_type(plugin)  # "BasePlugin", should be "ObjectsAPIRegistration"
```

so I was missing the specific attributes of `ObjectsAPIRegistration`.

The only downside of this change is that a registry could register a plugin that does not inherit from the correct base, although this isn't a big concern I think

### DIFF
--- a/src/openforms/plugins/registry.py
+++ b/src/openforms/plugins/registry.py
@@ -9,10 +9,13 @@ if TYPE_CHECKING:
     from .plugin import AbstractBasePlugin  # noqa: F401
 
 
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
I'll believe it :P

PR #3839, opened 2024-02-01
Fix docstring of `ValidatorsListView`

Viicos:


sergei-maertens (approved):
Re-generate the API schema please :wink:

Viicos:
That's a weird test failure, never had this one before

PR #3841, opened 2024-02-02
[#2617] Disable payment if price is zero

### LINKED ISSUE #2617: Allow payment to be disabled for certain users
### Thema / Theme

Form designer

internal: utr 39 + lv 70

Utr: As a form designer i want to create forms that (for instance) need to be paid directly when user is auth_bsn, but when the user is auth_kvk we do not require this payment and instead send a invoice. The former is an example, I would like to be able to use logic in order to determine if payment is necessary or not 

LV: We would like to skip/disable payment under certain conditions - this condition could be when the form is completed by someone else (like a employee) or when the user has no possibility to pay online. 

_No response_

This could be achieved by creating a condition where if the price = 0, payment gets disabled. At the moment 0 will start a payment request for 0.

vaszig:
Fixes #2617 
Depends on open-formulieren/open-forms-sdk#641


src/openforms/products/admin/product.py
```diff
@@ -1,8 +1,24 @@
...
+    class Meta:
+        model = Product
+        fields = "__all__"
+
+    def clean_price(self):
+        value = self.cleaned_data["price"]
+        if value <= 0:
+            raise ValidationError(_("The value must be greater than 0."))
+
+        return value
```
sergei-maertens, line 19: I think it'd be nicer to create a validator based on `MinValueValidator` from `django.core.validators` and apply that on the model field through `validators=[NonZeroValidator()]`, what do you think?

src/openforms/products/tests/test_admin.py
```diff
@@ -0,0 +1,40 @@
...
+        form["name"] = "example product"
+        form["price"] = Decimal("10.00")
+        response = form.submit()
+
+        product = Product.objects.get()
+
+        self.assertRedirects(response, reverse("admin:products_product_changelist"))
+        self.assertEqual(product.name, "example product")
+
+    def test_product_with_invalid_price_is_saved(self):
```
sergei-maertens, line 28:
```suggestion
    def test_product_with_invalid_price_is_not_saved(self):
```

src/openforms/products/tests/test_admin.py
```diff
@@ -0,0 +1,40 @@
...
+        create_page = self.app.get(admin_url, user=superuser)
+        form = create_page.form
+        form["name"] = "example product"
+        form["price"] = Decimal("0.00")
+        response = form.submit()
+
+        expected_error = [[_("The value must be greater than 0.")]]
+
+        self.assertEqual(response.context["errors"], expected_error)
+        self.assertFalse(Product.objects.exists())
```
sergei-maertens, line 40: Instead, what about `self.assertFormError(...)`? See https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.SimpleTestCase.assertFormError

src/openforms/submissions/models/submission.py
```diff
@@ -708,7 +709,39 @@ def calculate_price(self, save=True) -> None:
 
     @property
     def payment_required(self) -> bool:
-        return self.form.payment_required
+        if not self.form.payment_required:
+            return False
+
+        # test the rules one by one, if relevant
+        price_rules = self.form.formpricelogic_set.all()
+        for rule in price_rules:
```
sergei-maertens, line 717: I'm not a fan of this code duplication with `calculate_price` - can we not use `calculate_price` instead and if that returns a non-zero value, payment is required, otherwise it's not?

I think we can remove the `payment_required` check from that function, since there's a circular dependency otherwise.

sergei-maertens (changes requested):

src/openforms/products/tests/test_admin.py
vaszig: Thanks for the hint, I wasn't aware of this..I will remove though the admin tests since I added the django's `MinValueValidator` now to the model field.

src/openforms/submissions/models/submission.py
vaszig: Yes, I agree I didn't want to add duplicate code but I was facing some problems..
I will look again if removing the `payment_required` from the function does not affect any other functionality.

src/openforms/submissions/models/submission.py
```diff
@@ -708,7 +708,13 @@ def calculate_price(self, save=True) -> None:
...
     @property
     def payment_required(self) -> bool:
-        return self.form.payment_required
+        if not self.form.payment_required:
+            return False
+
+        self.calculate_price()
+        if self.price:
+            return True
+        return False
```
Viicos, line 717:
```suggestion
        self.calculate_price(save=False)
        return bool(self.price)
```

Last line is just a personal preference, however I think we shouldn't save on a property access.

src/openforms/products/models/product.py
```diff
@@ -20,7 +22,12 @@ class Product(models.Model):
         unique=True,
     )
     name = models.CharField(_("name"), max_length=50)
-    price = models.DecimalField(_("price"), max_digits=10, decimal_places=2)
+    price = models.DecimalField(
+        _("price"),
+        max_digits=10,
+        decimal_places=2,
+        validators=[MinValueValidator(Decimal("0.01"))],
+    )
```
Viicos, line 30: I think a validator that does `value > Decimal("0.0")` would be better. It's a shame that Django does not include such a validator (it only does _greater than or equal_), you can define it like so:

```python
@deconstructible
class StrictMinValueValidator(MinValueValidator):
    message = _('Ensure this value is greater than %(limit_value)s.')
    code = 'strict_min_value'

    def compare(self, a, b):
        return a <= b
```
However this requires translations for the message etc. and I don't think we'll see a case where people would want prices between 0 and 0.01 so it's also fine as is.

vaszig:
FYI I have modified the methods a bit since there were problems with a form which does not require payment. Basically, I saw that we do need the `submission.payment_required` check because it is used when a submission is completed, so that's why in my new commit I have added this again. 

If you agree with this approach I can then try to find out why this test is failing.

src/openforms/submissions/models/submission.py
```diff
@@ -702,13 +702,24 @@ def calculate_price(self, save=True) -> None:
         """
         Calculate and save the price of this particular submission.
         """
+        logger.debug("Calculating submission %s price", self.uuid)
```
Viicos, line 705: Maybe move this logging call after `if not self.payment_required`?

src/openforms/submissions/models/submission.py
```diff
@@ -702,13 +702,24 @@ def calculate_price(self, save=True) -> None:
...
         if save:
             self.save(update_fields=["price"])
 
     @property
     def payment_required(self) -> bool:
-        return self.form.payment_required
+        if not self.form.payment_required:
+            return False
+
+        return bool(get_submission_price(self))
```
Viicos, line 722: `get_submission_price` calls `submission.payment_required`, so this probably end up in an infinite loop?

vaszig: No, `calculate_price` is calling `submission.payment_required` according to the last commit. The `get_submission_price` is now only evaluating the rules or the product price

src/openforms/submissions/models/submission.py
sergei-maertens: IMO it makes sense to see the following log lines:

```
DEBUG Calculating submission xyz price
DEBUG Submission xyz does not require payment, skipping price calculation
```
or if payment is required:

```
DEBUG Calculating submission xyz price
```
I guess it provides a consistent "anchor" point to grep for in the logs, and possibly there is extra information?

src/openforms/submissions/pricing.py
```diff
@@ -21,16 +21,8 @@ def get_submission_price(submission: "Submission") -> Decimal | None:
 
     :param submission: the :class:`openforms.submissions.models.Submission: instance
       to calculate the price for.
-    :return: the calculated price, or ``None`` if no payment is required.
+    :return: the calculated price.
```
sergei-maertens, line 24: below `assert submission.form.product`, I'd add another:

```py
assert submission.form.product.price, "'get_submission_price' may only be called for forms that require payment"
```
That way, we can make the intent clear after removing the guard clause.

src/openforms/submissions/models/submission.py
sergei-maertens: I prefer this implementation using the `get_submission_price` call so that property access does not perform side effects anymore now.

I have some performance concerns, but instead of prematurely optimizing, let's keep the code as-is in this PR and keep an eye on APM for possible performance issues.

Viicos: Ah yes missed the diff on that part

sergei-maertens (approved):

sergei-maertens:
@Viicos apologies for the merge conflicts this causes in your PR :grimacing:

PR #3843, opened 2024-02-02
[#3730] Part 3 - Logic action tab

### LINKED ISSUE #3730: Integrate DMN evaluation into form logic

SilviaAmAm:
Fixes #3730 (partially)

### DIFF
--- a/src/openforms/forms/api/serializers/logic/action_serializers.py
+++ b/src/openforms/forms/api/serializers/logic/action_serializers.py
@@ -11,6 +11,7 @@ from json_logic.typing import Primitive
 from rest_framework import serializers
 
 from openforms.api.serializers import DummySerializer
+from openforms.formio.validators import variable_key_validator
 from openforms.utils.json_logic.api.validators import JsonLogicValidator
 from openforms.variables.constants import FormVariableDataTypes
 
@@ -72,6 +73,17 @@ class LogicFetchActionSerializer(serializers.Serializer):
     )
 
 
+class VariableMappingSerializer(serializers.Serializer):
+    form_variable = serializers.CharField(
+        required=True,
+        validators=[variable_key_validator],
+        label=_("Key of the form variable."),
+    )
+    dmn_variable = serializers.CharField(
+        required=True, label=_("DMN input parameter name.")
+    )
+
+
 class DMNEvaluateActionConfigSerializer(serializers.Serializer):
     plugin_id = serializers.CharField(
         label=_("Plugin ID"), required=True, allow_blank=False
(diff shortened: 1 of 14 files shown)


src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -225,6 +226,102 @@ const ActionSetRegistrationBackend = ({action, errors, onChange}) => {
...
+  };
+
+  const convertMappingBackwards = mapping => {
+    if (!mapping) return {};
+
+    let convertedMapping = {};
+    mapping.map(item => {
+      convertedMapping[item.formVar] = item.dmnVar;
+    });
+    return convertedMapping;
```
sergei-maertens, line 249: FYI - there also exists `Object.fromEntries` which is the reverse of `Object.entries` :)

sergei-maertens (changes requested):

src/openforms/scss/components/admin/_index.scss
```diff
@@ -27,6 +27,7 @@
 @import './logic-action';
 @import './logic-type-selection';
 @import './servicefetchconfiguration-picker';
+@import './mappings';
```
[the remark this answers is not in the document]
SilviaAmAm: I added them to get spacing between the title/the tables, so I don't think that they are relevant!

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -72,6 +73,17 @@ class LogicFetchActionSerializer(serializers.Serializer):
...
 
 
+class VariableMappingSerializer(serializers.Serializer):
+    form_variable = serializers.CharField(
+        required=True,
+        validators=[variable_key_validator],
+        help_text=_("Key of the form variable."),
+    )
+    dmn_variable = serializers.CharField(
+        required=True, help_text=_("DMN input expression.")
```
sergei-maertens, line 83: I think "expression" is confusing here, since inside the DMN tables you can use expressions too.

Maybe:
```suggestion
        required=True, label=_("DMN input parameter name.")
```

src/openforms/forms/api/serializers/logic/action_serializers.py
```diff
@@ -72,6 +73,17 @@ class LogicFetchActionSerializer(serializers.Serializer):
     )
 
 
+class VariableMappingSerializer(serializers.Serializer):
+    form_variable = serializers.CharField(
+        required=True,
+        validators=[variable_key_validator],
+        help_text=_("Key of the form variable."),
```
sergei-maertens, line 80:
```suggestion
        label=_("Key of the form variable."),
```
help text is more intended for longer descriptive texts. There should always be a label, but a help text without label looks off :)

sergei-maertens (approved):
Minor final tweaks :)

PR #3847, opened 2024-02-05
[#3688] Implement API endpoints and the Objects API regisration form

Viicos:
Part of #3688


src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormField.js
```diff
@@ -0,0 +1,360 @@
...
+        label={intl.formatMessage({
+          defaultMessage: 'Objecttype',
+          description: 'Objects API registration options "Objecttype" label',
+        })}
+        rawDescription={intl.formatMessage({
+          defaultMessage:
+            'URL that points to the ProductAanvraag objecttype in the Objecttypes API. The objecttype should have the following three attributes: "submission_id", "type" (the type of productaanvraag) and "data" (the submitted form data)',
+          description: 'Objects API registration options "Objecttype" description',
+        })}
+        rawErrors={getFieldErrors(name, index, validationErrors, 'objecttype')}
```
Viicos, line 106: This `rawErrors` prop is weird: it is described as an array of string, but is only used in a "truthy" context (`if rawErrors`/`rawErrors && ...`)

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+const getFieldErrors = (name, index, errors, field) => {
+  const errorMessages = [];
+
+  for (const [errorName, errorReason] of errors) {
+    if (errorName.startsWith(name)) {
+      const errorNameBits = errorName.split('.');
+      if (errorNameBits[2] === String(index) && errorNameBits[errorNameBits.length - 1] === field) {
```
Viicos, line 9:
```suggestion
      if (errorNameBits[2] === String(index) && errorNameBits.at(-1) === field) {
```
Can we safely make use of recent features? (i.e. is there any polyfills in the build)

src/openforms/registrations/contrib/objects_api/api.py
```diff
@@ -0,0 +1,79 @@
...
+        "object-types",
+        ObjecttypesListView.as_view(),
+        name="object-types",
+    ),
+    path(
+        "object-types/<uuid:submission_uuid>/versions",
+        ObjecttypeVersionsListView.as_view(),
+        name="object-type-versions",
+    ),
+]
```
SilviaAmAm, line 79: Maybe split this file in `views.py`, `serializers.py` and `urls.py` so that it's more intuitive to find things?

src/openforms/registrations/contrib/objects_api/api.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    uuid = serializers.UUIDField(label=_("Unique identifier (UUID4)."))
+    name = serializers.CharField(label=_("Name of the object type."))
+    namePlural = serializers.CharField(label=_("Plural name of the object type."))
+
+
+class ObjecttypeVersionSerializer(serializers.Serializer):
+    version = serializers.IntegerField(
+        label=_("Integer version of the Objecttype."),
+    )
```
SilviaAmAm, line 29: This currently would return a list with objects `{"version": "X"}`. I wonder if it would be nice for the endpoint to return a list of versions directly (without the objects) :thinking:

src/openforms/registrations/contrib/objects_api/api.py
```diff
@@ -0,0 +1,79 @@
...
+from drf_spectacular.utils import extend_schema, extend_schema_view
+from rest_framework import authentication, permissions, serializers, views
+
+from openforms.api.views import ListMixin
+
+from .client import get_objecttypes_client
+
+
+class ObjecttypeSerializer(serializers.Serializer):
+    # Keys are defined in camel case as this is what we get from the Objecttype API
```
SilviaAmAm, line 15: The JSON parser on the view does not convert it to snake case?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,360 @@
...
+import {CustomFieldTemplate} from 'components/admin/RJSFWrapper';
+import {Checkbox, NumberInput, TextArea, TextInput} from 'components/admin/forms/Inputs';
+import Select from 'components/admin/forms/Select';
+import {ValidationErrorContext} from 'components/admin/forms/ValidationErrors';
+import {get} from 'utils/fetch';
+
+import {
+  REGISTRATION_OBJECTTYPES_ENDPOINT,
+  REGISTRATION_OBJECTTYPE_VERSIONS_ENDPOINT,
+} from '../../constants';
```
SilviaAmAm, line 16:
```suggestion
import {
  REGISTRATION_OBJECTTYPES_ENDPOINT,
  REGISTRATION_OBJECTTYPE_VERSIONS_ENDPOINT,
} from 'components/admin/form_design/constants';
```

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,360 @@
...
+        rawErrors={getFieldErrors(name, index, validationErrors, 'uploadSubmissionCsv')}
+        errors={buildErrorsComponent('uploadSubmissionCsv')}
+        displayLabel
+      >
+        <Checkbox
+          id="root_uploadSubmissionCsv"
+          name="uploadSubmissionCsv"
+          value={uploadSubmissionCsv}
+          onChange={onFieldChange}
+        />
```
SilviaAmAm, line 236: Checkbox also has a required prop `label`

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,360 @@
...
+          defaultMessage: 'Version of the objecttype in the Objecttypes API',
+          description: 'Objects API registration options "Objecttype version" description',
+        })}
+        rawErrors={getFieldErrors(name, index, validationErrors, 'objecttypeVersion')}
+        errors={buildErrorsComponent('objecttypeVersion')}
+        displayLabel
+      >
+        {objecttypeVersionLoading ? (
+          <Loader />
+        ) : unknownObjecttype ? (
```
SilviaAmAm, line 147: For some reason, when I change objecttype the version field disappears for a moment but I don't see the loading spinner :thinking:

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,360 @@
...
+    const rawErrors = getFieldErrors(name, index, validationErrors, field);
+    return rawErrors ? getErrorMarkup(rawErrors) : null;
+  };
+
+  // API fetches:
+  const {loading: objecttypesLoading} = useAsync(async () => {
+    const response = await get(REGISTRATION_OBJECTTYPES_ENDPOINT);
+    if (!response.ok) {
+      throw new Error('Error when loading objecttypes');
+    }
```
SilviaAmAm, line 72: I think that these errors are being swallowed somewhere :thinking:

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
SilviaAmAm, line 1: This looks very similar to the code in `src/openforms/js/components/admin/form_design/registrations/zgw/utils.js`.

Maybe it can be reused?

src/openforms/registrations/contrib/objects_api/api.py
Viicos: I've mirrored the `camunda` contrib module, but I can make the change to both modules if required

src/openforms/registrations/contrib/objects_api/api.py
Viicos: Yes probably better. I'm also wondering if it would make sense to have a single endpoint. I don't have any preference, in the frontend it's mostly a matter of doing a request vs `.filter` coupled with `useEffect`

src/openforms/registrations/contrib/objects_api/api.py
Viicos: We are using a custom `ListMixin` view in this case, which does everything "by hand": instantiating the serializer and returning a `Response` with the serializer data. In other words, the JSON parser/renderer is not involved :/

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
Viicos: I'm a bit stuck here, moving the label from the `CustomFieldTemplate` to the `Checkbox` leads to the following:

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
Viicos: It's too fast to be able to start animating (the loader animation has a small delay before appearing).
Maybe the select components should always display, and if it is loading then it's just a matter of having the choices updated by the API call

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
Viicos: Can confirm by raising in the backend. Probably related to how `useAsync` or hooks work? I don't really know how these errors should be handled. Another example in the codebase, where this time it is ignored:

https://github.com/open-formulieren/open-forms/blob/537cddd1f83953409608ec91bcca2745d4125424/src/openforms/js/components/admin/form_design/variables/VariablesTable.js#L130-L139

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
Viicos: I kept it as is as I think this code won't be here for too long anyway (after the redesign)

src/openforms/registrations/contrib/objects_api/api.py
Viicos: Turns out a single API endpoint might not be a good idea: I would have to make `1 + n` requests for `n` objecttypes

src/openforms/contrib/objects_api/clients/objecttypes.py
Viicos, line 1: I don't know if this client needs tests.. VCR ones? requests_mock? (the later is cumbersome to do if I want every code path covered)

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
sergei-maertens: we use `browserslist` which informs babel about the required transforms, so I think it's okay and will be transpiled away

browserslist config is:

```
  "browserslist": [
    ">0.5%",
    "not dead",
    "not op_mini all"
  ],
```
in `package.json`

src/openforms/registrations/contrib/objects_api/api.py
sergei-maertens: keep in mind the scale of data - it's reasonable to assume tens/hundreds of objecttypes and versions will exist at some point in time (say 100 objecttypes with 20 versions each -> 2000 records returned).

Holding all that data in memory on the client-side is not optimal.

sergei-maertens: w/r to the `version` key - having a list of objects allows for future changes in the API to add additional metadata. returning a list of strings will require you to fight DRF and offers less flexibility, so I'd stick with the explicit serializer.

src/openforms/registrations/contrib/objects_api/api.py
sergei-maertens: we also don't want the renderer (not parser) to touch the output since we want to replicate the objecttypes API spec exactly for consistency reasons.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
sergei-maertens: select with `options={[['...', 'loading...']]}` could be a nice compromise?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
sergei-maertens: `useAsync` returns `{loading, value, error}` so exception should show up in the `error` key.

you may also need explicit `try ... catch` and re-throw the error because of how `async/await` work in JS :unamused:

src/openforms/contrib/objects_api/clients/objecttypes.py
sergei-maertens: at least some smoke tests would be nice, using VCR and a docker-compose set up for objects API (in `docker/ci`). The point is that any developer can bring up a live, functional objecttypes API that is pre-configured so that the VCR tests can be re-recorded. That'll probably involve dumping/loading some fixtures for the docker-compose setup.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
Viicos: So we keep it as is? (with `CustomFieldTemplate`)

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
Viicos: I grabbed the error key to also fallback to textinputs, as if no objecttype client is configured, the available objecttypes can't be retrieved and we need to provide the legacy way

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
sergei-maertens: yes

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormField.js
sergei-maertens: yep, we're stuck with it until we migrate to Formik unfortunately

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,372 @@
...
+            id="root_objecttypeVersion"
+            name="objecttypeVersion"
+            choices={
+              objecttypeVersionLoading
+                ? LOADING_OPTION
+                : availableObjecttypeVersions.map(otv => [otv.version, otv.version])
+            }
+            value={objecttypeVersion}
+            onChange={onFieldChange}
+            allowBlank={true}
```
sergei-maertens, line 174: this should not be optional, a version must be selected. And let's make sure to order the versions from newest -> oldest so that the newest version is selected by default.

That will probably require an explicit `useEffect` with `onChange` call to select a default/initial version

src/openforms/js/components/admin/form_design/registrations/objectsapi/mocks.js
```diff
@@ -0,0 +1,16 @@
...
+export const BASE_URL = process.env.SB_BASE_URL || '';
+
+export const mockObjecttypesGet = objecttypes =>
+  rest.get(`${BASE_URL}/api/v2/registration/plugins/objects-api/object-types`, (req, res, ctx) => {
+    return res(ctx.json(objecttypes));
+  });
+
+export const mockObjecttypeVersionsGet = (objecttypeUuid, versions) =>
+  rest.get(
+    `${BASE_URL}/api/v2/registration/plugins/objects-api/object-types/${objecttypeUuid}/versions`,
```
sergei-maertens, line 12: you can use actual params syntax here as used by express.js (if I understand it correctly)

```suggestion
    `${BASE_URL}/api/v2/registration/plugins/objects-api/object-types/:uuid/versions`,
```
so you then don't need to pass in the object type UUID in your mock setup in the stories.

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
sergei-maertens: add a todo comment that the duplication is deliberate and needs to be cleaned up when we get to refactor our backend UI code

src/openforms/registrations/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+@extend_schema_view(
+    get=extend_schema(
+        tags=["registration"],
+    ),
+)
+class ObjecttypesListView(ListMixin, views.APIView):
+    """
+    List the available Objecttypes.
+    """
```
sergei-maertens, line 20:
```suggestion
    """
    List the available Objecttypes.
    
    Note that the response data is essentially proxied from the configured Objecttypes API.
    """
```

src/openforms/registrations/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
```
sergei-maertens, line 23: This threw me off but apparently it's applied consistently through the codebase. Perhaps we need a `AdminUIAccessMixin` to properly centralize this. I had expected that there were additional permissions required to check that you can actually edit forms instead of this being accessible to any admin user :thinking:

src/openforms/registrations/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,47 @@
...
+    serializer_class = ObjecttypeSerializer
+
+    def get_objects(self) -> list[dict[str, Any]]:
+        with get_objecttypes_client() as client:
+            return client.list_objecttypes()
```
sergei-maertens, line 28: it is concerning that this is missing test coverage :)

src/openforms/registrations/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,47 @@
...
+    List the available versions for an Objecttype.
+    """
+
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+    serializer_class = ObjecttypeVersionSerializer
+
+    def get_objects(self) -> list[dict[str, Any]]:
+        with get_objecttypes_client() as client:
+            return client.list_objecttype_versions(self.kwargs["submission_uuid"])
```
sergei-maertens, line 47: I'm interested to see how this handles 404s for bogus UUIDs

src/openforms/registrations/contrib/objects_api/tests/test_objecttypes_client.py
```diff
@@ -0,0 +1,67 @@
...
+from ..client import get_objecttypes_client
+from ..models import ObjectsAPIConfig
+
+OBJECTTYPES_API_BASE_URL = os.getenv(
+    "OBJECTTYPES_API_BASE_URL", "http://localhost:8001/api/v2/"
+)
+# The default is the API key loaded in the docker service:
+OBJECTTYPES_API_KEY = os.getenv(
+    "OBJECTTYPES_API_KEY", "171be5abaf41e7856b423ad513df1ef8f867ff48"
+)
```
sergei-maertens, line 21: with the docker-compose setup we have an env 100% under our control, so there is no reason to make this configurable through the environment. Just a constant is sufficient.

src/openforms/registrations/contrib/objects_api/tests/test_objecttypes_client.py
```diff
@@ -0,0 +1,67 @@
...
+from zgw_consumers.models import Service
+
+from openforms.utils.tests.vcr import OFVCRMixin
+
+from ..client import get_objecttypes_client
+from ..models import ObjectsAPIConfig
+
+OBJECTTYPES_API_BASE_URL = os.getenv(
+    "OBJECTTYPES_API_BASE_URL", "http://localhost:8001/api/v2/"
+)
```
sergei-maertens, line 17: Managed through docker-compose with a fixed URL, so doesn't need to be configurable through the env.

sergei-maertens (changes requested):
I'd like to see this PR broken up in smaller PRs, there's too much going on all at once now :grimacing: 

The UI code could've been a separate PR limited to storybook, the additional endpoints could've been a separate PR.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,372 @@
...
+
+  const {loading: objecttypeVersionLoading, error: objecttypeVersionLoadingError} =
+    useAsync(async () => {
+      if (!objecttype) return;
+
+      const response = await get(
+        REGISTRATION_OBJECTTYPE_VERSIONS_ENDPOINT.replace(
+          '{objecttypeUuid}',
+          availableObjecttypes.find(o => o.url === objecttype).uuid
+        )
```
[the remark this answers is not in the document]
Viicos: Well technically `objecttype.uuid` is trusted input, as `objecttype` is determined like this:
`availableObjecttypes.find(o => o.url === objecttype)`

and `availableObjecttypes` comes from the API

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,372 @@
...
+    if (!response.ok) {
+      throw new Error('Error when loading objecttypes');
+    }
+
+    setAvailableObjecttypes(response.data);
+  });
+
+  const {loading: objecttypeVersionLoading, error: objecttypeVersionLoadingError} =
+    useAsync(async () => {
+      if (!objecttype) return;
```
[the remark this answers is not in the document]
Viicos: Did you mean _clear the available objecttype versions_?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -0,0 +1,372 @@
...
+            allowBlank={true}
+          />
+        ) : (
+          <Select
+            id="root_objecttype"
+            name="objecttype"
+            choices={
+              objecttypesLoading
+                ? LOADING_OPTION
+                : availableObjecttypes.map(ot => [ot.url, `${ot.name} (${ot.uuid})`])
```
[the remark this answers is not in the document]
Viicos: Having the uuid as a value would require (backwards incompatible) changes in the backend serializer, no?

src/openforms/registrations/contrib/objects_api/api/views.py
Viicos: Objecttypes API raises a `ValidationError` but for some reason a 500 is returned. Should we validate on our side that it is a correct UUID?

docker/docker-compose.objecttypes.yml
```diff
@@ -0,0 +1,20 @@
+version: '3'
+
+services:
+  db:
+    image: postgres
+    environment:
+      - POSTGRES_USER=${DB_USER:-objecttypes}
+      - POSTGRES_PASSWORD=${DB_PASSWORD:-objecttypes}
```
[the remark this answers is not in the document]
sergei-maertens: I'd take the same structure/set up like in our root `docker-init-db.sql` - the reason for this is that you'll be created a database for objects and objecttypes each, but in the same DB cluster. You really don't want both django apps to look for the same tables in the same DB, that will cause problems.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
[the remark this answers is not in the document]
sergei-maertens: response data from an API endpoint can never be trusted. At this point in time, yes, it's a UUID that's returned which is URL-safe, but changes in the future (intended or not) may lead to different behaviour. 

Of course, this is far fetched, so it's more about the fact that a parameter from a source outside of our own control is being interpolated, which can lead to problems. It's best practice to always escape your inputs :)

> and availableObjecttypes comes from the API

just to drive this point home - if my zaaktype has HTML in the description, and I retrieve that through our own API, are you going to blindly interpolate that `description` field into some `<div>`? :wink:

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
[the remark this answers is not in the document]
sergei-maertens: correct, the available versions need to be reset

src/openforms/registrations/contrib/objects_api/api/views.py
sergei-maertens: no, a 500 is never an intended/deliberate outcome. Create an issue in the objecttypes API for this case, and I think a 404 is more appropriate.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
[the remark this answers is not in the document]
Viicos: So if not discriminator should I also make the objecttype URL and version required for v1 as well?

docker/docker-compose.objecttypes.yml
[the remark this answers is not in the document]
Viicos: I created the init script but still got some weird errors (`FATAL: role "user" does not exist`)

I'm also unsure what the `open-forms-dev` network does and how it relates to the one in our docker compose.

I'm leaving this for now it takes too much of my time and I don't have all the knowledge so I'll come back to this later

src/openforms/registrations/contrib/objects_api/api/views.py
Viicos: Opened: https://github.com/maykinmedia/objects-api/issues/361

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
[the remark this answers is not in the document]
sergei-maertens: yes, let's bite the bullet on that. that will require a data migration to add those values from the global configuration if they're missing on the form-specific configuration.

docker/docker-compose.objecttypes.yml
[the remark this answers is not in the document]
sergei-maertens: okay - some information:

`FATAL: role "user" does not exist` - in docker-compose, I basically tend to enable TRUST mode, it's not for prod anyway. So we let all the SQL be executed by the postgres superuser. That init script then ensures a DB user exists and creates the necessary database(s) owned by that user.

So in compose you typically want to have:
```
    environment:
      - POSTGRES_HOST_AUTH_METHOD=trust
```

The `open-forms-dev` network is so that we can spin up multiple compose files each from their own shell but all the services are on the same network. That way, we can make the root `docker-compose.yml` talk to the camunda from `docker/docker-compose.camunda.yml` etc. That essentially allows you to spin up *all the services* and have them be connected to use all of OF's features at once.

PR #3852, opened 2024-02-07
Upgrade OIDC libraries

sergei-maertens:
Partly closes [#3623](https://github.com/open-formulieren/open-forms/issues/3623)

Tested against the keycloak instance via docker-compose and appears to work as expected still.


sergei-maertens:
Looks like something broke in the views after all

Viicos (approved):

PR #3860, opened 2024-02-09
[#3688] Phase 1 - Add Objecttypes API endpoints

Viicos:
Part of #3688

To be resolved:
- https://github.com/open-formulieren/open-forms/pull/3847#discussion_r1483237637

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
+++ b/src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
@@ -1,0 +1,176 @@
+from unittest.mock import patch
+
+import requests_mock
+from rest_framework import status
+from rest_framework.reverse import reverse_lazy
+from rest_framework.test import APITestCase
+from zgw_consumers.models import Service
+
+from openforms.accounts.tests.factories import StaffUserFactory, UserFactory
+
+from ..models import ObjectsAPIConfig
+
+
+@requests_mock.Mocker()
+class ObjecttypesAPIEndpointsTests(APITestCase):
+
+    def setUp(self) -> None:
+        super().setUp()
+
+        self.endpoint = reverse_lazy("api:objects_api:object-types")
+
+        patcher = patch(
+            "openforms.registrations.contrib.objects_api.client.ObjectsAPIConfig.get_solo",
+            return_value=ObjectsAPIConfig(
+                objecttypes_service=Service(api_root="https://objecttypen.nl/api/v2/")
+            ),
+        )
+
+        self.config_mock = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def test_auth_required(self, m):
+        response = self.client.get(self.endpoint)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_staff_user_required(self, m: requests_mock.Mocker):
+        m.get(
+            "https://objecttypen.nl/api/v2/objecttypes",
+            json={"count": 0, "next": None, "previous": None, "results": []},
+        )
+
+        user = UserFactory.create()
+        staff_user = StaffUserFactory.create()
+
+        with self.subTest(staff=False):
+            self.client.force_authenticate(user=user)
+
+            response = self.client.get(self.endpoint)
+
+            self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+        with self.subTest(staff=True):
+            self.client.force_authenticate(user=staff_user)
+
+            response = self.client.get(self.endpoint)
+
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_list_objecttypes(self, m: requests_mock.Mocker):
+        m.get(
+            "https://objecttypen.nl/api/v2/objecttypes",
+            json={
+                "count": 1,
+                "next": None,
+                "previous": None,
+                "results": [
+                    {
+                        "url": "https://objecttypen.nl/api/v2/objecttypes/2c77babf-a967-4057-9969-0200320d23f1",
+                        "uuid": "2c77babf-a967-4057-9969-0200320d23f1",
+                        "name": "Tree",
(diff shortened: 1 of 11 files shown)


src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
```diff
@@ -0,0 +1,176 @@
...
+        super().setUp()
+
+        self.endpoint = reverse_lazy("api:objects_api:object-types")
+
+        patcher = patch(
+            "openforms.registrations.contrib.objects_api.client.ObjectsAPIConfig.get_solo",
+            return_value=ObjectsAPIConfig(
+                objecttypes_service=Service(api_root="https://objecttypen.nl/api/v2/")
+            ),
+        )
```
Viicos, line 27: I don't like this. I'm touching implementation details of the client but this is not what I'm testing. A better solution would have been to create a mock client class, provide an mock implementation for the `get_objecttypes`/`get_objecttype_versions` methods, but I generally don't like going this way (considering the many footguns in the `unittest.mock` module. Having to rely on `mock` is also a sign that we are missing proper DI, however I don't know how this could be properly implemented in the drf ecosystem

src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
```diff
@@ -0,0 +1,178 @@
...
+from rest_framework.reverse import reverse_lazy
+from rest_framework.test import APITestCase
+from zgw_consumers.models import Service
+
+from openforms.accounts.tests.factories import StaffUserFactory, UserFactory
+
+from ..models import ObjectsAPIConfig
+
+
+@requests_mock.Mocker()
```
sergei-maertens, line 14: please use VCR instead

Viicos: See https://github.com/open-formulieren/open-forms/pull/3860#discussion_r1484368602, in my opinion VCR/requests_mock isn't the solution

sergei-maertens: I don't see how that addresses the actual data going over the wire.

The behaviour that you are testing is:

1. when calling our own endpoint, actually the data from the objecttypes API is returned
2. this data itself is a stripped down set of properties from the upstream API
3. our endpoint does not allow extracting this data if the service is not properly configured

A mock client class does not do any of that.

I have near-zero interest in unit-tests for this functionality, I don't particularly care about the client implementation either. What I require is some integration-test functionality: from my frontend, I can retrieve the available object types and their versions from my configured objecttypes API. All the rest are implementation details.

src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
sergei-maertens: discussion moved to https://github.com/open-formulieren/open-forms/pull/3860#discussion_r1486262781

src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
Viicos: The objecttypes client itself is tested [here](https://github.com/open-formulieren/open-forms/pull/3847/files#diff-02ce813114e188adba916eec0086bcb7d784a9ae84462385bb346352ac71f8d1), hence my original comment on duplication.

This test was here to test the API endpoint, where the client used is an implementation detail, and I feel in this case we want to test that the serializier is doing its job, the permissions are correct, etc.

This test file is almost a duplicate of this [test file](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/camunda/tests/test_api_endpoints.py), where the client data is mocked, so that's why I went the same way.

If this should be an integration test, then I should remove  [here](https://github.com/open-formulieren/open-forms/pull/3847/files#diff-02ce813114e188adba916eec0086bcb7d784a9ae84462385bb346352ac71f8d1)

sergei-maertens: The Camunda integration is very old from before when we even knew VCR exists - that's no longer a good reference.

I would indeed rather see the existing client tests be "lifted up" to the functionality that we're actually interested in - i.e. our own API endpoints. The access control tests of course also need to happen, but those are easy to shortcut & you can also deliberately include a mis-configuration to see how our own API endpoints react (a missing/bad API key for example).

I don't think I agree on the duplication part. yes, some test code is duplicated to set up the service/client, but they serve to distinct goals:

* do our own API endpoints return what is needed?
* is the client implementation solid enough?

The client tests are a lower-level aspect to achieve the first item, but they have the advantage that they're easier (less-integration-y) to debug breakage with. If both the client + api endpoint tests, it should be easier to dive into the client tests to resolve the problem

I've seen too many issues with stale mocks, incorrect mocks or mocks that correctly implement the API spec but the service itself doesn't actually behave according to its documentation. That's why I strongly prefer using VCR for any external HTTP traffic instead of coming up with my own edge cases/things that can go wrong. E.g. with a mock client, you wouldn't think of returning an HTTP 500 for the `/api/v2/objecttypes/:bad-uuid/versions` endpoint, right?

src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class ObjecttypesAPIEndpointsTests(OFVCRMixin, APITestCase):
+
+    VCR_TEST_FILES = Path(__file__).parent / "files"
+
+    def setUp(self) -> None:
+        super().setUp()
+
+        self.endpoint = reverse_lazy("api:objects_api:object-types")
```
sergei-maertens, line 21: FYI - you can just use `reverse_lazy` with a class attribute for these kind of situations

sergei-maertens (approved):

PR #3867, opened 2024-02-12
[#3864] Fix handling of StufBG responses with one partner

### LINKED ISSUE #3864: Processing the family member choices leads to errors since it contains non-arrays

2.5.2

Sentry: 344594

Iterating over the list of family member options includes `@entiteittype` and also converts arrays of 1 item to a dict...

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

Viicos:
Fixes #3864

I think it would be nice to get a more robust solution for xml parsing. OIP uses https://github.com/tefra/xsdata which looks great. https://github.com/dapper91/pydantic-xml looks interesting as well

### DIFF
--- a/src/openforms/formio/components/np_family_members/stuf_bg.py
+++ b/src/openforms/formio/components/np_family_members/stuf_bg.py
@@ -8,15 +8,15 @@ def get_np_family_members_stuf_bg(
     include_partners: bool,
     submission: Submission | None = None,
 ) -> list[tuple[str, str]]:
-    values = []
+    attributes: list[str] = []
     if include_children:
-        values.append("inp.heeftAlsKinderen")
+        attributes.append("inp.heeftAlsKinderen")
 
     if include_partners:
-        values.append("inp.heeftAlsEchtgenootPartner")
+        attributes.append("inp.heeftAlsEchtgenootPartner")
 
     with get_client() as client:
-        data = client.get_values(bsn, values)
+        data = client.get_values(bsn, attributes)
 
     # Kids
     family_members = []
--- a/src/openforms/formio/components/np_family_members/tests/test_family_members.py
+++ b/src/openforms/formio/components/np_family_members/tests/test_family_members.py
@@ -2,7 +2,7 @@ import json
 from pathlib import Path
 from unittest.mock import patch
 
-from django.test import TestCase
+from django.test import TestCase, tag
 from django.utils.html import format_html
 from django.utils.translation import gettext as _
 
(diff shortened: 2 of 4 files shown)


src/openforms/formio/components/np_family_members/stuf_bg.py
```diff
@@ -8,15 +8,15 @@ def get_np_family_members_stuf_bg(
...
-        values.append("inp.heeftAlsKinderen")
+        attributes.append("inp.heeftAlsKinderen")
 
     if include_partners:
-        values.append("inp.heeftAlsEchtgenootPartner")
+        attributes.append("inp.heeftAlsEchtgenootPartner")
 
     with get_client() as client:
-        data = client.get_values(bsn, values)
+        data = client.get_values(bsn, attributes)
```
sergei-maertens, line 19: nothing changed here except for the type annotation and the variable name, right?

Can you break that up into two separate commits so that I can cherry pick only the actual bugfix for the backports?

src/openforms/formio/components/np_family_members/tests/responses/stuf_bg_one_family_member.xml
sergei-maertens, line 1: the file has 'one family member' but I'm seeing two children and one partner? That's three family members. Is the name or the mock data wrong?

src/openforms/formio/components/np_family_members/tests/test_family_members.py
```diff
@@ -150,6 +150,31 @@ def test_get_partners_stuf_bg(self, mock_stufbg_config_get_solo):
...
+        )
+
+        with requests_mock.Mocker() as m:
+            m.post(
+                stuf_bg_service.get_endpoint(type=EndpointType.vrije_berichten),
+                content=response_content,
+            )
+
+            partners_choices = get_np_family_members_stuf_bg(
+                bsn="111222333", include_children=False, include_partners=True
```
sergei-maertens, line 172: `include_children=False` should result in no children being returned in the response data, no? Since we don't request it...

src/stuf/stuf_bg/client.py
```diff
@@ -69,6 +69,7 @@ def get_values(self, bsn: str, attributes: list[str]) -> dict:
                 response_data,
                 process_namespaces=True,
                 namespaces=NAMESPACE_REPLACEMENTS,
+                force_list=["inp.heeftAlsEchtgenootPartner", "inp.heeftAlsKinderen"],
```
sergei-maertens, line 72: :cry: why is this standard like this

sergei-maertens (changes requested):

sergei-maertens:
> I think it would be nice to get a more robust solution for xml parsing. OIP uses https://github.com/tefra/xsdata which looks great. https://github.com/dapper91/pydantic-xml looks interesting as well

I really want all of this stuff *out* of OF/OIP and properly organized in a centralized library used by the final projects. It's crazy how many times we implement the same things over and over but in slightly different ways.

src/openforms/formio/components/np_family_members/tests/responses/stuf_bg_one_family_member.xml
Viicos: Renamed to `stuf_bg_one_partner.xml`, and removed children (the test doesn't include them anyway)

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.5.x: 5734b936d9a322d1a172620c20a856a8f08565f4
* 2.4.x: fb5dd8e6e1de31cce1553097b0d4a2ea504da2f9

Older versions didn't support this functionality yet.

PR #3870, opened 2024-02-12
[#3688] Add migration to set required fields from global configuration

Viicos:
Part of #3688

For misconfigured backends (the ones without a URL and version and when no global default is configured), should we add an upgrade check?


sergei-maertens:
> For misconfigured backends (the ones without a URL and version and when no global default is configured), should we add an upgrade check?

I don't think this makes a lot of sense - if in $current version registration configuration is broken, you don't really win anything by preventing an upgrade. It'll just stay broken, and they'll be forced to pick an object type next time they edit the form.

sergei-maertens (approved):

PR #3882, opened 2024-02-14
[#3853] Update kvk zoeken API to v2.0

### LINKED ISSUE #3853: Update KvK Zoeken API from v1.0to  v2.0
As far as I'm concerned, we don't need to keep version 1 since it's a nationally available service. This means everyone that is using version 1 also has access to version 2 (if gateways are not messing things up).

https://developers.kvk.nl/nl/documentation/migration-zoeken-v2

vaszig:
Fixes #3853


src/openforms/contrib/kvk/client.py
```diff
@@ -29,16 +29,18 @@ class SearchParams(TypedDict, total=False):
...
+    huisletter: str
+    postbusnummer: int
     huisnummerToevoeging: str
-    type: str
-    InclusiefInactieveRegistraties: Literal["true", "false"]
+    type: list[str]
+    inclusiefInactieveRegistraties: Literal["true", "false"]
     pagina: int  # [1, 1000]
-    aantal: int  # [1, 100]
+    resultatenPerPagina: int  # [1, 100] - default is 10
```
Viicos, line 43: I guess if tests are passing it means we weren't making use of these fields?

vaszig: Yes, I understood that we declare here all the possible query params, but in fact we only validate against `kvkNummer`, `rsin` and `vestigingsnummer`

SilviaAmAm (approved):

src/openforms/contrib/kvk/tests/test_validators.py
sergei-maertens, line 1: Can we update these tests to also make use of VCR instead of the mocks data from JSON files?

docs/configuration/prefill/kvk.rst
```diff
@@ -52,7 +52,7 @@ Configuration
 
    * **Label**: *Fill in a human readable label*, for example: ``My KvK Zoeken API``
    * **Type**: ORC (Overige)
-   * **API root URL**: *The API URL provided by KvK but typically:* ``https://api.kvk.nl/api/v1/zoeken/``
+   * **API root URL**: *The API URL provided by KvK but typically:* ``https://api.kvk.nl/api/v2/zoeken/``
```
sergei-maertens, line 55: I think this is wrong since add the `v2/zoeken` part ourselves in the API client

```suggestion
   * **API root URL**: *The API URL provided by KvK but typically:* ``https://api.kvk.nl/api/``
```

which is quite nice since we don't need a data migration now to handle this change

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3886, opened 2024-02-14
[#3855] Retrieve variables from decision definitions

### LINKED ISSUE #3855: Allow for a better DMN user experience for both editor and end-user
Followup on #3730

**Tasks**

SilviaAmAm:
Fixes #3855 (partially)

Backend part to get input/output "variables" from DMN decision definitions

### DIFF
--- a/src/openforms/dmn/api/urls.py
+++ b/src/openforms/dmn/api/urls.py
@@ -1,6 +1,7 @@
 from django.urls import path
 
 from .views import (
+    DecisionDefinitionInputOutputView,
     DecisionDefinitionListView,
     DecisionDefinitionVersionListView,
     DecisionDefinitionXMLView,
@@ -23,5 +24,10 @@ urlpatterns = [
         "decision-definitions/xml",
         DecisionDefinitionXMLView.as_view(),
         name="dmn-definition-xml",
+    ),
+    path(
+        "decision-definitions/inputs-outputs",
+        DecisionDefinitionInputOutputView.as_view(),
+        name="dmn-definition-inputs-outputs",
     ),
 ]
--- a/src/openforms/dmn/base.py
+++ b/src/openforms/dmn/base.py
@@ -2,6 +2,8 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Any
 
+from django_camunda.dmn.datastructures import DMNIntrospectionResult
+
 from openforms.plugins.plugin import AbstractBasePlugin
 
 
@@ -65,3 +67,16 @@ class BasePlugin(ABC, AbstractBasePlugin):
         If this is not available, return an empty string.
         """
         return ""
+
+    def get_decision_definition_parameters(
+        self, definition_id: str, version: str = ""
+    ) -> DMNIntrospectionResult | None:
+        """
+        Return the input/output clauses for a given decision definition (version).
+
+        An input clause defines the id, label, expression and type of a decision table input. It is represented by an
+        input element inside a decisionTable XML element.
+        An output clause defines the id, label, name and type of a decision table output. It is represented by an
+        output element inside a decisionTable XML element.
+        """
+        return None
--- a/src/openforms/dmn/base.py
+++ b/src/openforms/dmn/base.py
@@ -2,9 +2,9 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Any
 
-from django_camunda.dmn.datastructures import DMNIntrospectionResult
-
 from openforms.plugins.plugin import AbstractBasePlugin
+
+from .types import DMNInputOutput
 
 
 @dataclass
@@ -68,9 +68,10 @@ class BasePlugin(ABC, AbstractBasePlugin):
         """
         return ""
 
+    @abstractmethod
     def get_decision_definition_parameters(
         self, definition_id: str, version: str = ""
-    ) -> DMNIntrospectionResult | None:
+    ) -> DMNInputOutput:
         """
         Return the input/output clauses for a given decision definition (version).
 
@@ -79,4 +80,4 @@ class BasePlugin(ABC, AbstractBasePlugin):
         An output clause defines the id, label, name and type of a decision table output. It is represented by an
         output element inside a decisionTable XML element.
         """
-        return None
+        ...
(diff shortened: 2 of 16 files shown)


sergei-maertens (commented):
Before fully diving in, how does this work with a decision definition that has multiple tables?

E.g.:

* table1, inputs A, B, output X
* table 2, input C, outputs Y, Z
* table 3: inputs X, Y, Z, outputs V, W

table 1 and 2 both feed into table 3 which returns the actual interesting outputs, so I then expect:

* inputs: A, B, C
* outputs: V, W

Possibly if there are any other outputs from table 1 / table 2 that are *not* used in table 3, those should also be included in the final outputs

SilviaAmAm:
If there is a DRD with multiple decision tables, the tables and their relationship are all described in the same XML.

When you evaluate a decision table with Camunda, if the table has another table as a dependency it will be evaluated automatically. Only the outputs of the table being run (the 'final' table) will be returned. So if the 'dependency' tables have intermediate outputs, these are not returned in the final result.

sergei-maertens:
> If there is a DRD with multiple decision tables, the tables and their relationship are all described in the same XML.
> 
> When you evaluate a decision table with Camunda, if the table has another table as a dependency it will be evaluated automatically. Only the outputs of the table being run (the 'final' table) will be returned. So if the 'dependency' tables have intermediate outputs, these are not returned in the final result.

That's at least intuitive, nice. Let's document this in our internal dev documentation (either in a docstring somewhere or some explicit .rst file documentation)

src/openforms/dmn/api/urls.py
```diff
@@ -24,4 +26,14 @@
...
+    path(
+        "decision-definitions/inputs",
+        DecisionDefinitionInputView.as_view(),
+        name="dmn-definition-inputs",
+    ),
+    path(
+        "decision-definitions/outputs",
+        DecisionDefinitionOutputView.as_view(),
+        name="dmn-definition-outputs",
+    ),
```
sergei-maertens, line 38: does it make sense to make these two separate endpoints? that implies parsing the XML twice, where-as an endpoint to get both the inputs and outputs in a single response sounds more practical to me (also from a UI perspective -> you only have to make the single API call)

src/openforms/dmn/contrib/camunda/plugin.py
```diff
@@ -122,3 +124,85 @@ def get_definition_xml(definition_id: str, version: str = "") -> str:
 
     def check_config(self):
         check_config(self)
+
+    def _get_inputs(self, xml, definition_id: str) -> list[InputElement]:
+        decision_table = xml.find(
```
sergei-maertens, line 129: I would move the actual implementation of this into its own module, e.g. `parser.py` and add a bit of an abstraction:

Even better would be to implement that in our django-camunda library :)

```py
class Parser:
    def __init__(self, xml: str):
        self.xml = xml

    def extract_inputs(self) -> DMNInputParameter:
        ...

    def extract_outputs(self) -> DMNOutputParameter:
        ...
```

You can then leverage the LXML tree as a cached attribute in the parser instance while keeping the code to process/extract information well organized.

sergei-maertens (changes requested):

src/openforms/dmn/contrib/camunda/plugin.py
SilviaAmAm: https://github.com/maykinmedia/django-camunda/pull/12

src/openforms/dmn/api/urls.py
SilviaAmAm: I went for 2 endpoints because I thought that I would end up doing 2 calls in the frontend anyway (since with Formik the 2 variables tables will end up being 2 'inter related' fields with the plugin/decision id/decision version). But maybe I can try to figure out another way of dealing with the frontend.

src/openforms/dmn/base.py
```diff
@@ -2,6 +2,8 @@
 from dataclasses import dataclass
 from typing import Any
 
+from django_camunda.dmn.datastructures import DMNIntrospectionResult
```
sergei-maertens, line 5: I don't think we should use camunda-specific data shapes for the generic implementation.

There should be a decoupled base class/protocol describing what data should be returned by a plugin, since you have an endpoint that is plugin-agnostic.

So you'll either need to define a base class or a protocol to prescribe what needs to be returned by a plugin.

src/openforms/dmn/contrib/camunda/plugin.py
```diff
@@ -122,3 +124,12 @@ def get_definition_xml(definition_id: str, version: str = "") -> str:
...
         check_config(self)
+
+    def get_decision_definition_parameters(
+        self, definition_id: str, version: str = ""
+    ) -> DMNIntrospectionResult:
+        xml_doc = self.get_definition_xml(definition_id, version)
+
+        parser = Parser(xml=xml_doc.encode("utf-8"))
+
+        return parser.extract_parameters(definition_id)
```
sergei-maertens, line 135: You should now be able to replace this with:

```py
from django_camunda.dmn import get_dmn_parser

with get_client() as client:
    camunda_id = _get_decision_definition_id(client, definition_id, version)
    parser = get_dmn_parser(dmn_id=camunda_id, client=client)
return parser.extract_parameters(definition_id)
```
right?

The retrieval of the DMN XML is already done in `get_dmn_parser`, no need for this code to still exist here.

src/openforms/dmn/tests/test_api.py
```diff
@@ -1,5 +1,7 @@
 from unittest.mock import patch
 
+from django_camunda.dmn.datastructures import DMNIntrospectionResult
+from django_camunda.dmn.types import DMNInputParameter, DMNOutputParameter
```
sergei-maertens, line 4: importing the camunda library in a module/plugin that has nothing to do with Camunda is a sign there's something missing in the abstraction layer :smile:

sergei-maertens (changes requested):

sergei-maertens (approved):

src/openforms/dmn/tests/test_management_commands.py
```diff
@@ -21,6 +22,12 @@ def get_available_decision_definitions():
     def evaluate(*args, **kwargs):
         pass
 
+    @staticmethod
+    def get_decision_definition_parameters(
+        definition_id: str, version: str = ""
+    ) -> DMNInputOutput:
+        pass
```
sergei-maertens, line 29: this doesn't type check :grimacing:

PR #3894, opened 2024-02-15
[#3688] Add a new registration column in variables with summary

Viicos:
Part of #3688

Styling TBD

### DIFF
--- a/src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
+++ b/src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
@@ -1,10 +1,18 @@
 import React from 'react';
-import {useIntl} from 'react-intl';
+import {FormattedMessage, useIntl} from 'react-intl';
 
 import FAIcon from 'components/admin/FAIcon';
 import {BACKEND_OPTIONS_FORMS} from 'components/admin/form_design/registrations';
+import ButtonContainer from 'components/admin/forms/ButtonContainer';
 
-const RegistrationSummary = ({name, registrationPath}) => {
+/**
+ *
+ * @param {Object} p
+ * @param {string} p.name - The name of the current registration backend
+ * @param {JSX.Element} p.registrationSummary - The rendered summary of the registration backend for a specific variable
+ * @returns {JSX.Element} - The rendered summary, containing the backend name, summary and edit icon
+ */
+const RegistrationSummary = ({name, registrationSummary}) => {
   const intl = useIntl();
 
   return (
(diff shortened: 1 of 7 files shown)


src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -0,0 +1,55 @@
...
+    <>
+      <span>{name}</span>
+      <span>
+        <FAIcon
+          icon="edit"
+          title={intl.formatMessage({
+            defaultMessage: 'Edit variable registration',
+            description: "'Edit variable registration' icon label",
+          })}
+          extraClassname="fa-lg actions__action"
```
sergei-maertens, line 20: the `actions__action` class probably doesn't have any effect because there's no `.actions` parent.

```suggestion
          extraClassname="fa-lg"
```

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -0,0 +1,55 @@
...
+
+const RegistrationsSummary = ({variableKey, registrationBackends}) => {
+  const summaries = [];
+
+  for (const backend of registrationBackends) {
+    const SummaryHandler = BACKEND_OPTIONS_FORMS[backend.key].summaryHandler;
+    if (SummaryHandler) {
+      summaries.push({
+        name: backend.name,
+        registrationPath: (
```
sergei-maertens, line 37: `registrationPath` is a term that is too tied to structured data, which doesn't apply for the email registration for example (there I expect just a line like "included / not included".

Maybe rename this `registrationPath` property to `representation`?

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import {useIntl} from 'react-intl';
+
+import FAIcon from 'components/admin/FAIcon';
+import {BACKEND_OPTIONS_FORMS} from 'components/admin/form_design/registrations';
+
+const RegistrationSummary = ({name, registrationPath}) => {
```
sergei-maertens, line 7:
```suggestion
const RegistrationSummary = ({name, representation}) => {
```
reasoning for why this rename is below :)

sergei-maertens (changes requested):
Probably for a separate PR, but we will also need to add this mapping to the static variables tab :grimacing:

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
Viicos: I went for `registrationSummary` :)

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
Viicos: I had to use it or it wouldn't to anything on hover

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -0,0 +1,101 @@
...
+      continue;
+
+    // ... if it is the case, check if the variable is already configured:
+    const formVariableConfigured = backendInfo.formVariableConfigured(variableKey, backend.options);
+    if (formVariableConfigured) {
+      // a summary handler is guaranteed to exist if `configurableFromVariables` is true:
+      const SummaryHandler = BACKEND_OPTIONS_FORMS[backend.key].summaryHandler;
+      const registrationSummary = (
+        <SummaryHandler variableKey={variableKey} backendOptions={backend.options} />
+      );
```
Viicos, line 70: I initially wanted to this `formVariableConfigured` logic inside the component directly:

If the variable wasn't registered, it would return `null`.

However, doing:

```jsx
const registrationSummary =
  <SummaryHandler variableKey={variableKey} backendOptions={backend.options} />
);
```

`registrationSummary` cannot be checked against `null`, as it will be a JSX component no matter the return value

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
sergei-maertens: it violates the BEM principle, an element cannot exist without being contained inside its parent block.

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -0,0 +1,105 @@
...
+          <span>{name}</span>
+        </div>
+        <span>
+          <FAIcon
+            icon="edit"
+            title={intl.formatMessage({
+              defaultMessage: 'Edit variable registration',
+              description: "'Edit variable registration' icon label",
+            })}
+            extraClassname="fa-lg actions__action"
```
sergei-maertens, line 31: allowing this for now until the styles are sorted out, but ideally the icons would follow the rest of the UI and be to the left:

```jsx
<div style={display: 'flex'}>
  <div className="actions actions--vertical">
    <FAIcon extraClassName="actions__action" />
  </div>
  <div>
    ...
  </div>
<div>
```

sergei-maertens (approved):

PR #3902, opened 2024-02-16
[#3855] Select DMN variables in config (frontent)

### LINKED ISSUE #3855: Allow for a better DMN user experience for both editor and end-user

SilviaAmAm:
Fixes #3855 (partially)

depends on https://github.com/open-formulieren/open-forms/pull/3886

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNParametersForm.js
+++ b/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNParametersForm.js
@@ -9,18 +9,17 @@ import {get} from 'utils/fetch';
 
 import VariableMapping from './VariableMapping';
 
+const EMPTY_DMN_PARAMS = {inputs: [], outputs: []};
+
 const DMNParametersForm = () => {
   const {
     values: {pluginId, decisionDefinitionId, decisionDefinitionVersion},
   } = useFormikContext();
   const {formVariables} = useContext(FormContext);
 
-  const [dmnParams, setDmnParams] = useState({inputs: [], outputs: []});
-
-  const {loading} = useAsync(async () => {
+  const {loading, value: dmnParams = EMPTY_DMN_PARAMS} = useAsync(async () => {
     if (!pluginId || !decisionDefinitionId) {
-      setDmnParams({inputs: [], outputs: []});
-      return;
+      return EMPTY_DMN_PARAMS;
     }
 
     const queryParams = {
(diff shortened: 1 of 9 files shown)


src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNParametersForm.js
```diff
@@ -0,0 +1,73 @@
...
+      queryParams.version = decisionDefinitionVersion;
+    }
+
+    const response = await get(DMN_DECISION_DEFINITIONS_PARAMS_LIST, queryParams);
+
+    setDmnParams({
+      inputs: response.data.inputs.map(inputParam => [inputParam.expression, inputParam.label]),
+      outputs: response.data.outputs.map(outputParam => [outputParam.name, outputParam.label]),
+    });
+  }, [pluginId, decisionDefinitionId, decisionDefinitionVersion]);
```
sergei-maertens, line 41: so this usage of `useAsync` is what I also told Victorien - it's not really how it's intended.

`useAsync` itself already maintains the state for the result, in this case, `dmnParams`. The return value of the callback in `useAsync` is stored in the state. So, a more correct/intended usage is:

```suggestion
  const EMPTY_DMN_PARAMS = {inputs: [], outputs: []);

  const {loading, value: dmnParams = EMPTY_DMN_PARAMS} = useAsync(async () => {
    if (!pluginId || !decisionDefinitionId) {
      return EMPTY_DMN_PARAMS;
    }

    const queryParams = {
      engine: pluginId,
      definition: decisionDefinitionId,
    };

    if (decisionDefinitionVersion) {
      queryParams.version = decisionDefinitionVersion;
    }

    const response = await get(DMN_DECISION_DEFINITIONS_PARAMS_LIST, queryParams);
    const {inputs, outputs} = response.data;
    return {
      inputs: inputs.map(inputParam => [inputParam.expression, inputParam.label]),
      outputs: outputs.map(outputParam => [outputParam.name, outputParam.label]),
    };
  }, [pluginId, decisionDefinitionId, decisionDefinitionVersion]);
```

src/openforms/js/components/admin/form_design/logic/actions/dmn/VariableMapping.js
```diff
@@ -65,8 +61,11 @@ const VariableMapping = ({mappingName, values, formVariables}) => {
                       htmlFor={`${mappingName}.${index}.dmnVariable`}
                       name={`${mappingName}.${index}.dmnVariable`}
                     >
-                      <TextInput
+                      <Select
                         id={`${mappingName}.${index}.dmnVariable`}
+                        allowBlank={true}
```
sergei-maertens, line 66:
```suggestion
                        allowBlank
```
JSX makes this `true` :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3903, opened 2024-02-16
[#3688] Add `get_objecttype_version` method to client

Viicos:
Part of #3688.

This endpoint will be required later to retrieve the objecttype JSON Schema


src/openforms/contrib/objects_api/clients/objecttypes.py
```diff
@@ -52,3 +52,12 @@ def list_objecttype_versions(
         return self._get_paginated(
             f"objecttypes/{objecttype_uuid}/versions", page=page, page_size=page_size
         )
+
+    def get_objecttype_version(
+        self,
+        objecttype_uuid: str | UUID,
```
vaszig, line 58: Shouldn't this always be a `str`?Otherwise it should be handled before passed to the url constructor right?I see  that this is already used for the above calls in the main branch,but just wanted to ask.

vaszig (approved):

src/openforms/contrib/objects_api/clients/objecttypes.py
Viicos: I allowed both as type hints as I'm using it in an fstring, which will call the `__str__` method of `UUID`

PR #3905, opened 2024-02-19
Address broken test isolation

### LINKED ISSUE #3900: New flakiness in tests to be solved w/r to session locking
```
ERROR: test_raising_loa_requirements_fails_in_flight_authentications (openforms.authentication.contrib.eherkenning.tests.test_signicat_integration.SignicatEHerkenningIntegrationTests)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/unittest/mock.py", line 1379, in patched
    return func(*newargs, **newkeywargs)
  File "/home/runner/work/open-forms/open-forms/src/openforms/authentication/contrib/eherkenning/tests/test_signicat_integration.py", line 525, in test_raising_loa_requirements_fails_in_flight_authentications
    acs_response_2 = self.client.get(acs_url_2.url, follow=True)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 744, in get
    response = self._handle_redirects(response, data=data, **extra)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 850, in _handle_redirects
    response = request_method(path, data=data, content_type=content_type, follow=False, **extra)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 742, in get
    response = super().get(path, data=data, secure=secure, **extra)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 396, in get
    return self.generic('GET', path, secure=secure, **{
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 473, in generic
    return self.request(**r)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 719, in request
    self.check_exception(response)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/test/client.py", line 580, in check_exception
    raise exc_value
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/runner/work/open-forms/open-forms/src/openforms/submissions/views.py", line 46, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/django/views/generic/base.py", line 189, in get
    url = self.get_redirect_url(*args, **kwargs)
  File "/home/runner/work/open-forms/open-forms/src/openforms/submissions/views.py", line 167, in get_redirect_url
    add_submmission_to_session(submission, self.request.session)
  File "/home/runner/work/open-forms/open-forms/src/openforms/submissions/utils.py", line 151, in add_submmission_to_session
    append_to_session_list(session, SUBMISSIONS_SESSION_KEY, str(submission.uuid))
  File "/home/runner/work/open-forms/open-forms/src/openforms/submissions/utils.py", line 130, in append_to_session_list
    with _session_lock(session, session_key):
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/home/runner/work/open-forms/open-forms/src/openforms/submissions/utils.py", line 81, in _session_lock
    with redis_cache.lock(
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/redis/lock.py", line 168, in __exit__
    self.release()
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/redis/lock.py", line 253, in release
    self.do_release(expected_token)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/redis/lock.py", line 259, in do_release
    raise LockNotOwnedError("Cannot release a lock that's no longer owned")
redis.exceptions.LockNotOwnedError: Cannot release a lock that's no longer owned
```

### LINKED ISSUE #3904: Session locking leads to CI failures
https://github.com/open-formulieren/open-forms/blob/0070e8bc3f9d31dd2b783ab8342b6fe4e2310429/src/openforms/submissions/utils.py#L85

Failure: https://github.com/open-formulieren/open-forms/actions/runs/7949563803/job/21700885697

~~What's happening here is that the timeout and blocking timeout are both the same value, meaning the client (open forms) will wait that duration to acquire a lock. However, the timeout value specifies after how much time the server should evict the key/lock itself, making it no longer owned by the client.~~

~~Likely what happens here then is:~~

~~1. Some sort of slow test or race condition is leading to up to 2s of trying to acquire a lock~~
~~2. The server evicts it after 2s~~
~~3. The original holder of the lock continues/exits, but can't release the lock (anymore) because it was removed by the server.~~

~~So it's important to figure out what is causing acquiring of the lock in < 2s to happen, and second, the server timeout should be higher than the wait/blocking timeout.~~

edit: probably due to the `--parallel 2` option with caches being cleared in-between, this is affecting the portalocker/redis lock.

sergei-maertens:
Closes #3904
Closes #3900

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -116,7 +116,7 @@ CACHES = {
     # redis locks?
     "portalocker": {
         "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"redis://{config('CACHE_PORTALOCKER', 'localhost:6379/0')}",
+        "LOCATION": f"redis://{config('CACHE_PORTALOCKER', 'localhost:6379/4')}",
         "OPTIONS": {
             "CLIENT_CLASS": "django_redis.client.DefaultClient",
             "IGNORE_EXCEPTIONS": False,
--- a/src/openforms/utils/tests/cache.py
+++ b/src/openforms/utils/tests/cache.py
@@ -2,5 +2,10 @@ from django.core.cache import caches
 
 
 def clear_caches():
-    for cache in caches.all():
+    for alias in caches:
+        # clearing the distributed session locks causes issues when tests are run
+        # in parallel (--parallel X)
+        if alias == "portalocker":
+            continue
+        cache = caches[alias]
         cache.clear()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
           --concurrency=multiprocessing \
           --parallel-mode \
           src/manage.py test src \
-            --parallel 2 \
+            --parallel 4 \
             --exclude-tag=e2e \
             --verbosity 2
           coverage combine
@@ -167,7 +167,7 @@ jobs:
           python src/manage.py compilemessages
           python src/manage.py collectstatic --noinput --link
           src/manage.py test src \
-            --parallel 2 \
+            --parallel 4 \
             --exclude-tag=e2e \
             --reverse
 


src/openforms/utils/tests/cache.py
```diff
@@ -2,5 +2,10 @@
 
 
 def clear_caches():
-    for cache in caches.all():
+    for alias in caches:
+        # clearing the distributed session locks causes issues when tests are run
+        # in parallel (--parallel X)
+        if alias == "portalocker":
+            continue
+        cache = caches[alias]
```
Viicos, line 10: Is there a way we could only clear the Redis cache values related to the current test? Not clearing _all_ cache values from Redis makes sense, but perhaps we could clear only the relevant keys? Not sure if this is possible.

My idea was to "record" all keys set during a test, and use `delete_many` on these keys as a cleanup. Probably too hard to implement and not worth the effort.

src/openforms/conf/base.py
```diff
@@ -116,7 +116,7 @@
     # redis locks?
     "portalocker": {
         "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"redis://{config('CACHE_PORTALOCKER', 'localhost:6379/0')}",
+        "LOCATION": f"redis://{config('CACHE_PORTALOCKER', 'localhost:6379/4')}",
```
Viicos, line 119: Was `4` chosen as a database number for a specific reason?

src/openforms/utils/tests/cache.py
sergei-maertens: no, it's near impossible to trace down which keys are all used and I really don't want this implementation detail to leak into the test setup since every time a library adds caching we need to be aware of the keys used. We also often need access to local-scoped variables for dynamic keys that we simply don't have.

The premise for tests is still that they should be fully isolated, meaning that nuking redis/cache entirely after a test-run is the semantically correct approach.

> My idea was to "record" all keys set during a test, and use delete_many on these keys as a cleanup. Probably too hard to implement and not worth the effort.

I don't think this is feasible. Databases are much nicer in that regard in the sense that a transaction can just be rolled back.

Personally I was thinking in getting the test process ID/index and using that to point to a particular redis DB so that each test process has its own redis database, but settings are loaded before the multiprocessing pool is started and tests are distributed to it. Django actually sort-of monkeypatches the DB settings after that. I do want to explore subclassing the test runner to apply the same behaviour to the portalocker cache `LOCATION` since that is more structurally sound, but for now I went with this bandaid fix, since the django test runner code is quite complicated and I don't know how easy it is to add in this behaviour.

src/openforms/conf/base.py
sergei-maertens: it was the key closest to my middle finger :D

src/openforms/utils/tests/cache.py
sergei-maertens: @Viicos I tried the above, but I couldn't get it to work reliably and I have to touch/monkepatch a lot of private API of django-redis which I'm not feeling comfortably with.

Additionally, this implicitly makes the redis DB 0 not available, and if DB 1 would be used for celery for example, the test process can only start using DB 2 or higher, which in turn again limits the available number of parallel processes since default redis clusters have 16 DBs. In my case with 16 CPUs on my system, I can't rely on automatic detection for the parallel count because of these arbitrary constraints. You can also configure redis to have more or less databases, so hardcoding these numbers is also annoying.

I'll stick with this approach of explicitly skipping portalocker + assigning an explicit DB index for the portalocker cache, this at least makes things as small as possible and keeps them predictable.

Viicos: I think if tests passes currently with the solution it should probably be fine

sergei-maertens: yeah I'm seeing failures on the flaky hypothesis test now, but in general this patch should fix a bunch of frustrations

sergei-maertens:
Backports: 056aeaf133de336986f798cb58cc69f4c0d29be6, ad05c4fb5d2763ff2b1f9a871f64512fef4c481f, 3398f80dc5675a9aeb3803ada5f9dab7e7a42294 and 3cc65fa09045c77fbad1a9c543b7543df71c0a1a

PR #3906, opened 2024-02-19
Lazily import `weasyprint`

Viicos:
My upstream PR got declined and lazy importing `weasyprint` was the recommended solution


vaszig (approved):

sergei-maertens (approved):
Can you update this in default-project too?

PR #3908, opened 2024-02-19
[#3672] Allow access to data as list, dict in service fetch

### LINKED ISSUE #3672: Service fetch: variables are always string, allow access to data as list, dict etc.
### Thema / Theme

Form designer

Currently all variables used in service fetch as query/header parameters are string, this does not allow a form builder to retrieve elements from list, dicts etc.

list = `["foo", "bar"]`
When using {{ list.0 }} i expect "foo" in my query, but instead i get "["

Workaround currently is to first store list.0 in a separate string variable to be used as "foo" in my request.

vaszig:
Fixes #3672

### DIFF
--- a/src/openforms/formio/utils.py
+++ b/src/openforms/formio/utils.py
@@ -1,11 +1,11 @@
 import logging
 from dataclasses import dataclass
-from typing import Any, Iterator, TypeAlias, TypeGuard
+from typing import Any, Callable, Iterator, TypeAlias, TypeGuard
 
 import elasticapm
 from glom import Coalesce, Path, glom
 
-from openforms.typing import DataMapping, JSONObject
+from openforms.typing import DataMapping, JSONObject, JSONValue
 from openforms.utils.glom import _glom_path_to_str
 from openforms.variables.constants import DEFAULT_INITIAL_VALUE, FormVariableDataTypes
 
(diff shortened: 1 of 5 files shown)


src/openforms/formio/utils.py
```diff
@@ -427,3 +427,45 @@ def iterate_data_with_components(
...
+                for key, nested_bit in input.items()
+            }
+
+        # other primitive or complex object - we can't template this out, so return it
+        # unmodified.
+        case _:
+            if transform_leaf:
+                return func(input)
+
+            return input
```
sergei-maertens, line 471:
```suggestion
        case _:
            # other primitive or complex object - we can't template this out, so return it
            # unmodified unless the transformation is explicitly requested
            return func(input, *args, **kwargs) if transform_leaf else input
```

src/openforms/variables/models.py
```diff
@@ -126,12 +127,13 @@ def request_arguments(self, context: DataMapping) -> dict:
         #
         # extra knowledge not in the RFC: latin1 is a different name for ISO-8859-1
 
+        ctx = recursive_apply(context, str, transform_leaf=True)
```
sergei-maertens, line 130:
```suggestion
        # Explicitly cast values to strings to avoid localization
        ctx = recursive_apply(context, str, transform_leaf=True)
```

sergei-maertens (changes requested):
Some polishing :)

sergei-maertens (approved):

PR #3910, opened 2024-02-19
[#3688] Add utility function to iter over JSON schema

Viicos:
Part of #3688.

The end goal with this will be to iterate over this `iter_json_schema` generator, and compare[^1] each sub JSON Schema to the one from the form variable. If it matches, we keep the path as a valid one to be displayed in the UI.

[^1]: This is yet to be defined: a `as_json_schema` property/method could be defined on the `FormVariable` model along with a function that gets this form variable schema and the sub one from the iter function and returns a boolean, or a `def is_compatible_with_json_schema(self, json_schema) -> bool` function that handles everything.

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/json_schema.py
+++ b/src/openforms/registrations/contrib/objects_api/json_schema.py
@@ -1,0 +1,70 @@
+from dataclasses import dataclass
+from typing import Iterator, Literal, overload
+
+from jsonschema_specifications import REGISTRY
+from referencing import Resource
+from referencing.exceptions import Unresolvable
+from referencing.jsonschema import ObjectSchema
+
+
+@dataclass
+class InvalidReference:
+    """A class representing an unknown/invalid reference."""
+
+    uri: str
+    """The URI of the unknown reference."""
+
+
+@overload
+def iter_json_schema_paths(
+    json_schema: ObjectSchema, fail_fast: Literal[False]
+) -> Iterator[tuple[list[str], ObjectSchema | InvalidReference]]: ...
+
+
+@overload
+def iter_json_schema_paths(
+    json_schema: ObjectSchema, fail_fast: Literal[True] = ...
+) -> Iterator[tuple[list[str], ObjectSchema]]: ...
+
+
+def iter_json_schema_paths(
+    json_schema: ObjectSchema, fail_fast: bool = True
+) -> Iterator[tuple[list[str], ObjectSchema | InvalidReference]]:
+    """Recursively iterate over the JSON Schema paths, resolving references if required.
+
+    Yields a two-tuple containing the current path (as a list of string segments) and the matching (sub) JSON Schema.
+
+    Known to be unsupported:
+    - Composition (https://json-schema.org/understanding-json-schema/reference/combining)
+    """
+    resource = Resource.from_contents(json_schema)
+    # Or referencing.jsonschema.EMPTY_REGISTRY?
+    resolver = REGISTRY.resolver_with_root(resource)
+
+    parent_json_path: list[str] = []
+
+    def _iter_json_schema(
+        json_schema: ObjectSchema, parent_json_path: list[str]
+    ) -> Iterator[tuple[list[str], ObjectSchema | InvalidReference]]:
+        assert json_schema.get("type") == "object"
+
+        yield parent_json_path, json_schema
+
+        for k, v in json_schema["properties"].items():
+            json_path = parent_json_path + [k]
+            match v:
+                case {"type": "object"}:
+                    yield from _iter_json_schema(v, json_path)
+                case {"$ref": str(uri)}:
+                    try:
+                        resolved = resolver.lookup(uri)
+                    except Unresolvable:
+                        if fail_fast:
+                            raise
+                        yield json_path, InvalidReference(uri)
+                    else:
+                        yield from _iter_json_schema(resolved.contents, json_path)
+                case {}:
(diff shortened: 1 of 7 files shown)


src/openforms/registrations/contrib/objects_api/tests/test_json_schema.py
```diff
@@ -0,0 +1,211 @@
...
+                        "type": "object",
+                    },
+                ),
+                (
+                    ["complainant", "residence"],
+                    {"properties": {"street": {"type": "string"}}, "type": "object"},
+                ),
+                (["complainant", "residence", "street"], {"type": "string"}),
+            ],
+        )
```
Viicos, line 176: This is still kind of a naive implementation as "deep" references are not resolved until encountered.

This test shows an example of this: the first yielded value has `properties.residence` still unresolved. However, I'm wondering if it should be fully resolved, as I don't think we'll be able to support variables with JSON schemas (yet to be defined, as discussed the `FormVariable` might benefit from a `as_json_schema` method) with deep nested properties

src/openforms/registrations/contrib/objects_api/tests/test_json_schema.py
```diff
@@ -0,0 +1,319 @@
...
+    simply be tested as ``([], CURRENT_SCHEMA_TESTED)``.
+    """
+
+    def test_iter_json_schema_no_refs(self):
+        paths_list = [
+            (path.segments, schema)
+            for path, schema in iter_json_schema_paths(JSON_SCHEMA_NO_REFS)
+        ]
+
+        self.assertListEqual(
```
sergei-maertens, line 135: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual

> These methods are used by default when comparing lists or tuples with [assertEqual()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual).

So in general you should just always be calling `self.assertEqual` and not the type-specific variants.

```suggestion
        self.assertEqual(
```

src/openforms/registrations/contrib/objects_api/tests/test_json_schema.py
```diff
@@ -0,0 +1,319 @@
...
+            ],
+        )
+
+    def test_iter_json_schema_unknown_ref(self):
+        self.assertRaises(
+            Unresolvable,
+            lambda: list(
+                iter_json_schema_paths(JSON_SCHEMA_UNKOWN_REF, fail_fast=True)
+            ),
+        )
```
sergei-maertens, line 219: we use the context manager form usually:

```py
with self.assertRaises(Unresolvable):
    list(iter_json_schema_paths(JSON_SCHEMA_UNKOWN_REF, fail_fast=True))
```

it's a bit more readable than the lambda variant.

src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -0,0 +1,143 @@
...
+
+@dataclass
+class InvalidReference:
+    """An unknown/invalid reference."""
+
+    uri: str
+    """The URI of the unknown reference."""
+
+    exc: Unresolvable
+    """The ``referencing`` catched exception."""
```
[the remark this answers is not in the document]
Viicos: Oops

src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -0,0 +1,143 @@
...
+    return [
+        r_path.segments
+        for r_path, _ in iter_json_schema_paths(json_schema)
+        if r_path.required
+        # If a child key is provided (e.g. "a.b"), any parent required key is dismissed (e.g. "a"):
+        if not any(JsonSchemaPath(path).startswith(r_path) for path in paths)
+        # If a parent key is provided (e.g. "a"), any child required key is dismissed (e.g. "a.b")
+        # (this one assumes the provided "a" value is valid with respect to the required children keys.
+        # We are dealing with path segments, so we can't really make any assumptions on the provided value):
+        if not any(r_path.startswith(path) for path in paths)
```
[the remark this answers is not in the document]
Viicos: It's part of the Python syntax for list comprehensions, I like it as it aligns every condition correctly :)

src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -0,0 +1,143 @@
...
+    ]
```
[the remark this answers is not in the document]
Viicos: Yes it's starting to be hard to understand

src/openforms/registrations/contrib/objects_api/tests/test_json_schema.py
Viicos: Aaah unittest..

src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -0,0 +1,149 @@
...
+            continue
+
+        # If a parent key is provided (e.g. "a"), any required child key is dismissed (e.g. "a.b").
+        # This one assumes the provided value for "a" is valid with respect to the required children keys.
+        # The JSON Schema could specify "a" as an object with some required keys, but we are dealing with
+        # path segments, so we can't really make any assumptions on the provided value.
+        if any(r_path.startswith(path) for path in paths):
+            continue
+
+        missing_paths.append(r_path.segments)
```
sergei-maertens, line 149: missing a `return` statement ;)

PR #3912, opened 2024-02-20
[#3855] Catch errors when DMN plugin is not working

### LINKED ISSUE #3855: Allow for a better DMN user experience for both editor and end-user

SilviaAmAm:
Fixes #3855 
Screenshot from 2024-02-20 11-48-07

Maybe the plugin ID field should be kept out of the boundary so that the user can select a different plugin in case of errors. Or they can just create a new action and delete the one with errors?

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
+++ b/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
@@ -1,6 +1,6 @@
 import {Form, Formik, useField, useFormikContext} from 'formik';
 import PropTypes from 'prop-types';
-import React, {useContext, useState} from 'react';
+import React, {useContext, useEffect} from 'react';
 import {FormattedMessage, defineMessage} from 'react-intl';
 import {useAsync} from 'react-use';
 
@@ -11,6 +11,7 @@ import {
 } from 'components/admin/form_design/constants';
 import Field from 'components/admin/forms/Field';
 import Select from 'components/admin/forms/Select';
+import ErrorBoundary from 'components/errors/ErrorBoundary';
 import {get} from 'utils/fetch';
 
 import DMNParametersForm from './DMNParametersForm';
@@ -24,32 +25,40 @@ const ERRORS = {
 };
 
 const DecisionDefinitionIdField = () => {
-  const {values, setFieldValue, getFieldProps, errors, touched} = useFormikContext();
+  const {
+    values: {pluginId},
+    setFieldValue,
+    getFieldProps,
+    errors,
+    touched,
+    setFieldTouched,
+    handleChange,
+  } = useFormikContext();
   const [field] = useField('decisionDefinitionId');
-  const [decisionDefinitions, setDecisionDefinitions] = useState([]);
 
-  useAsync(async () => {
-    if (!values.pluginId) {
+  const {error, value: decisionDefinitions = []} = useAsync(async () => {
+    if (!pluginId) return [];
+
+    const response = await get(DMN_DECISION_DEFINITIONS_LIST, {
+      engine: pluginId,
+    });
+
+    return response.data.map(item => [item.id, item.label]);
+  }, [pluginId]);
+
+  useEffect(() => {
+    if (!touched.pluginId) return;
+
+    if (!pluginId) {
       setFieldValue('decisionDefinitionId', '');
-      setDecisionDefinitions([]);
       return;
     }
 
-    const response = await get(DMN_DECISION_DEFINITIONS_LIST, {
-      engine: values.pluginId,
-    });
-
-    const definitionChoices = response.data.map(item => [item.id, item.label]);
-    setDecisionDefinitions(definitionChoices);
-
     // If pluginId has changed, reset the value of the definition
-    if (
-      values.decisionDefinitionId !== '' &&
-      !definitionChoices.find(item => item[0] === values.decisionDefinitionId)
-    ) {
-      setFieldValue('decisionDefinitionId', '');
-    }
-  }, [values.pluginId, setFieldValue]);
+    setFieldValue('decisionDefinitionId', '');
+  }, [pluginId, setFieldValue]);
+
+  if (error) throw error;
 
   return (
     <Field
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):
Yes, let's bring the plugin field outside of the error boundary :) that makes more sense

SilviaAmAm:
Screenshot from 2024-02-21 11-21-50

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -104,13 +107,15 @@ const DecisionDefinitionVersionField = () => {
     // If decisionDefinitionId has changed, reset the value of the definition version if the new decision definition
     // does not have that version
     if (
-      values.decisionDefinitionVersion !== '' &&
-      !versionChoices.find(item => item[0] === values.decisionDefinitionVersion)
+      decisionDefinitionVersion !== '' &&
```
sergei-maertens, line 110: shouldn't `decisionDefinitionVersion` be passed as dependency to `useAsync` too?

SilviaAmAm: I thought no, because we don't need to refetch/update things if the version changes :thinking:

sergei-maertens: but it's used inside the hook in a check? if the version changes, the callback still has the old value and behaviour is not as expected.

Looking at it again, I think the `setFieldValue` call based on the available version choices actually belongs in a separate `useEffect`. And you can also simplify the logic IMO - if the definition ID changes, then the version should always be reset to make users aware they explicitly need to select the version. Version X for definition A does not have the same semantic meaning as Version X for definition B.

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -24,32 +25,40 @@ const ERRORS = {
...
 
     const response = await get(DMN_DECISION_DEFINITIONS_LIST, {
-      engine: values.pluginId,
+      engine: pluginId,
     });
 
-    const definitionChoices = response.data.map(item => [item.id, item.label]);
-    setDecisionDefinitions(definitionChoices);
+    return response.data.map(item => [item.id, item.label]);
+  }, [pluginId, setFieldValue]);
```
sergei-maertens, line 47:
```suggestion
  }, [pluginId]);
```
this is not used in the hook, is it?

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -24,32 +25,40 @@ const ERRORS = {
...
+
+    if (!pluginId) {
+      await setFieldValue('decisionDefinitionId', '');
+      return;
     }
-  }, [values.pluginId, setFieldValue]);
+
+    // If pluginId has changed, reset the value of the definition
+    await setFieldValue('decisionDefinitionId', '');
+  }, [pluginId, setFieldValue]);
```
sergei-maertens, line 59: What's the reason for calling these with `async`/`await`? I'd expected this to be a normal, sync `useEffect` hook and not `useAsync` :thinking:

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -82,35 +96,37 @@ const DecisionDefinitionVersionField = () => {
...
 
     const response = await get(DMN_DECISION_DEFINITIONS_VERSIONS_LIST, {
       engine: pluginId,
       definition: decisionDefinitionId,
     });
 
-    const versionChoices = response.data.map(item => [item.id, item.label]);
-    setDecisionDefinitionVersions(versionChoices);
+    return response.data.map(item => [item.id, item.label]);
+  }, [pluginId, decisionDefinitionId, setFieldValue]);
```
sergei-maertens, line 114:
```suggestion
  }, [pluginId, decisionDefinitionId]);
```

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -82,35 +96,37 @@ const DecisionDefinitionVersionField = () => {
...
-    ) {
-      setFieldValue('decisionDefinitionVersion', '');
+    if (!pluginId || !decisionDefinitionId) {
+      await setFieldValue('decisionDefinitionVersion', '');
+      return;
     }
+
+    // If decisionDefinitionId has changed, reset the value of the definition version
+    await setFieldValue('decisionDefinitionVersion', '');
   }, [pluginId, decisionDefinitionId, setFieldValue]);
```
sergei-maertens, line 126: here too I'd expect `useEffect` rather than `useAsync`

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
SilviaAmAm: I dont know, the docs say that setFieldValue returns a promise, so I thought that the async would be needed?

sergei-maertens: ah, that only really impacts what you do with the return value:

```js
foo = setFieldValue(...);
```
-> `foo` will be a promise so if you want to do something with the return value, you need to:

```js
foo = await setFieldValue(...)
// do more stuff with the return value
```

but since we don't use the return value of the (promise of) setFieldValue, it doesn't really matter. The side-effect is going to run no matter what.

I don't even know what the return value of `setFieldValue` promises looks like :D

sergei-maertens: > If validateOnChange is set to true and there are errors, they will be resolved in the returned Promise.

right, so it returns validation errors. I don't see how that's useful for our needs, so you don't need to `await` it.

sergei-maertens (approved):

PR #3914, opened 2024-02-20
[#3164] Add custom timeout for services

### LINKED ISSUE #3164: Allow setting custom timeout values for services

any

**Bug:** 
nopolder 58
Registration task fails when submission includes multiple large attachments.

Tested on problematic service that fails when multiple large attachments were included. increasing the timeout from 10s to 30s solved the issue and registration task completed with no issues.

**improvement:**
Not all services are created equal, and some services might need some more time to proces incoming data (like dealing with attachments in zaken) then others. 

At the moment requests get a 10s timeout from [base.py](https://github.com/open-formulieren/open-forms/blob/5d0a3b8043ba7725721733e953ce492f42bfa52e/src/openforms/conf/base.py#LL579C1-L579C25)

Allowing services to have their own timeout value instead of the default 10s would prevent these services from timing out when a large request is made.

**Tasks**

* [ ] Add environment variable to set timeout for document requests; default=60s
* [ ] Change ZGW document requests to use above timeout
* [ ] Change StUF-document requests to use above timeout
* [ ] Change Objects API document requests (Document API) to use above timeout
* [ ] Change Sharepoint attachment requests to use above timeout

vaszig:
Fixes #3164

- [x] Needs https://github.com/maykinmedia/zgw-consumers/pull/85
- [x] Needs new version (0.30.0) of zgw-consumers
Services using zgw-consumers
  - BRK
  - HaalCentraal
  - Kadaster
  - KVK
  - Objects API
  - Qmatic
- [x] Add timeout field to Soap Service
Services using Soap
  - JCC
  - StUF
  - StUF-BG
  - StUF-ZDS
  - Suwinet
- [x] Add timeout field to Microsoft Graph


src/openforms/setup.py
sergei-maertens, line 1: discussed on slack - we want to keep the global defaults/fallbacks as a safeguard measure

there will always be libraries that use `requests` under the hood without a way for us to specify timeouts, the monkeypatch addresses those

sergei-maertens (approved):

PR #3916, opened 2024-02-20
[#3855] Cache DMN requests results

### LINKED ISSUE #3855: Allow for a better DMN user experience for both editor and end-user

SilviaAmAm:
Fixes #3855 (partially)

Do we want a customisable timeout?

### DIFF
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass
 from typing import Any, Mapping, TypedDict
+
+from django.core.cache import cache
+from django.core.serializers.json import DjangoJSONEncoder
 
 from glom import assign
 from json_logic import jsonLogic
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/logic/actions.py
```diff
@@ -245,13 +248,17 @@ def eval(
...
+        def _evaluate_dmn():
+            return evaluate_dmn(
+                definition_id=self.decision_definition_id,
+                version=self.decision_definition_version,
+                input_values=dmn_inputs,
+                plugin_id=self.plugin_id,
+            )
+
+        # Perform DMN call or retrieve result from cache
+        cache_key = hash(str(submission.uuid) + json.dumps(dmn_inputs, sort_keys=True))
```
sergei-maertens, line 260: How do we guarantee that the inputs are JSON serializable? What if there ends up a UUID or datetime in the inputs?

sergei-maertens: I'd also expect the definition identifiers to be part of the key - a different table with the same inputs does not produce the same result.

Submission ID then is not strictly necessary, though it's probably a good defensive approach to avoid mixing data of different users

SilviaAmAm: > How do we guarantee that the inputs are JSON serializable? What if there ends up a UUID or datetime in the inputs?

I added a similar thing to what the camunda package does, but I realised that this will not work if we have non-serializable types nested in lists/dicts :thinking: Possibly a good idea to update both so that some recursive serialisation is done?

src/openforms/submissions/logic/actions.py
```diff
@@ -245,13 +249,23 @@ def eval(
...
+            )
+
+        # Perform DMN call or retrieve result from cache
+        cache_key = hash(
+            str(submission.uuid)
+            + self.decision_definition_id
+            + self.decision_definition_version
+            + self.plugin_id
+            + json.dumps(serialize_variables(dmn_inputs), sort_keys=True)
         )
```
sergei-maertens, line 267: instead of `serialize_variables`, you can probably leverage `from django.core.serializers.json import DjangoJSONEncoder` :sweat_smile: 

I'd then also add some type hints that incorporate `openforms.typing.DjangoJSONEncodable` so we have type-checking taking care of which variable types are serializable (or at least have the intent documented).

sergei-maertens (approved):

PR #3917, opened 2024-02-20
[#3688] Add implementation of the Objects API mapping editor

Viicos:
Part of #3688

Relies on https://github.com/open-formulieren/open-forms/pull/3925 for the API request.

# PR Summary

This PR builds on top of a previous one, where the "registration summary" was displayed, but nothing would happen when you tried editing the configuration:

image

To do so, the `RegistrationSummary` component (responsible for displaying the summary shown above) was tweaked to display a `FormModal`. Inside this modal:
- the variable configuration editor is rendered in a generic way (the `RegistrationSummary` doesn't know anything about it). This will allow other registration backends to provide their own edit form in the future.
- a submit button that will update the options of the matching reg. backend.
- this modal makes use of a Formik form, and the actual configuration editor form can make use of the formik context to edit the options.

For the Objects API, the actual form is defined in the `ObjectsApiVariableConfigurationEditor` component:

image

- Unfortunately, I have to pass the `onFieldChange` callback all the way down to allow backend options to be edited.
- The logic is a bit complex, because we have to handle both edits to an existing mapping and configuring a new variable.
- The target dropdown is also tricky to handle: options values for the select component need to be strings, so the best way we can do this is using `JSON.stringify/parse` to be sure it translates to string and back to an actual value without any collisions.
- If the select is set to the empty option (`----`), it means the variable will remain unmapped. This logic can be found in `TargetPathSelect` component, where we make use of formik's array helpers.

---

This PR also fixes quite some issues not directly related:
- the `LOADING_OPTION` is now properly translated
- When switching from v1 to v2, we make sure an empty `variablesMapping` array is set to avoid errors
- In various components, the complete backend/variable data is passed to props, instead of specific attributes (e.g. `variableKey` -> `variable`)
- Simplified the summary logic: instead of displaying a _Add another_ button, every backend summary is displayed (and they can display a fallback message if it is not yet configured)
- Added the registration column for static variables
- Fixed a crash where the wrong attribute was used to determine the backend conf

---

What remains and will be tackled in another PR:
- Updating the variables mapping array when the name (and thus the key) of a variable changes. Currently, you will see it gets out of sync when you change the name of a user defined variable. Same goes for the component variables when they are edited in the form builder (for this, I will make use of the [`onStepEdit` ](https://github.com/open-formulieren/open-forms/pull/3917/files#diff-2b49dda99f17cba3438b72ce1b13790be227ea434e1307275b6976b9a77375a6R28), however I still need to figure out a way for user defined variables where no such logic is currently defined).
- ~~Error handling if the API request to fetch the compatible target paths fails~~ Done
- Mapping a component variable inside `record.geometry`: the current implementation only allows mapping variables to a target in the `record.data` attribute on the Objects API side. However, it is possible, with the `apply_data_mapping` functionality in the backend, to map a component variable to the geometry attribute. This will need special casing unfortunately.
- Only display targets not yet mapped?

### DIFF
--- a/src/openforms/js/components/admin/form_design/Context.js
+++ b/src/openforms/js/components/admin/form_design/Context.js
@@ -7,7 +7,7 @@ const FeatureFlagsContext = React.createContext({});
 FeatureFlagsContext.displayName = 'FeatureFlagsContext';
 
 const FormContext = React.createContext({
-  form: {url: ''},
+  form: {url: '', uuid: ''},
   components: {},
   formSteps: [],
   formDefinitions: [],
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
@@ -1,5 +1,6 @@
 import {FieldArray, useFormikContext} from 'formik';
 import isEqual from 'lodash/isEqual';
+import PropTypes from 'prop-types';
 import React, {useContext} from 'react';
 import {FormattedMessage} from 'react-intl';
 import {useAsync, useToggle} from 'react-use';
@@ -135,6 +136,12 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
   );
 };
 
+ObjectsApiVariableConfigurationEditor.propTypes = {
+  variable: PropTypes.shape({
+    key: PropTypes.string.isRequired,
+  }).isRequired,
+};
+
 const TargetPathSelect = ({name, index, choices}) => {
   const {getFieldProps, setFieldValue} = useFormikContext();
   const props = getFieldProps(name);
@@ -163,6 +170,12 @@ const TargetPathSelect = ({name, index, choices}) => {
   );
 };
 
+TargetPathSelect.propTypes = {
+  name: PropTypes.string.isRequired,
+  index: PropTypes.number.isRequired,
+  choices: PropTypes.array.isRequired,
+};
+
 const TargetPathDisplay = ({target}) => {
   const path = target.targetPath.join(' > ');
   return (
@@ -172,6 +185,13 @@ const TargetPathDisplay = ({target}) => {
       values={{path, required: target.isRequired}}
     />
   );
+};
+
+TargetPathDisplay.propTypes = {
+  target: PropTypes.shape({
+    targetPath: PropTypes.arrayOf(PropTypes.string).isRequired,
+  }).isRequired,
+  isRequired: PropTypes.bool.isRequired,
 };
 
 export default ObjectsApiVariableConfigurationEditor;
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
@@ -28,13 +28,13 @@ import {asJsonSchema} from './utils';
  *
  * @param {Object} p
  * @param {Object} p.variable - The current variable
- * @returns {JSX.Element} - The summary, represented as a the parts of the target path separated by '>'
+ * @returns {JSX.Element} - The configuration form for the Objects API
  */
 const ObjectsApiVariableConfigurationEditor = ({variable}) => {
   const {csrftoken} = useContext(APIContext);
 
   const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
-  const {values: backendOptions, getFieldProps, setFieldValue} = useFormikContext();
+  const {values: backendOptions, getFieldProps} = useFormikContext();
 
   /** @type {ObjectsAPIRegistrationBackendOptions} */
   const {objecttype, objecttypeVersion, variablesMapping, version} = backendOptions;
@@ -191,6 +191,7 @@ TargetPathSelect.propTypes = {
   name: PropTypes.string.isRequired,
   index: PropTypes.number.isRequired,
   choices: PropTypes.array.isRequired,
+  mappedVariable: PropTypes.object.isRequired,
 };
 
 const TargetPathDisplay = ({target}) => {
@@ -207,8 +208,8 @@ const TargetPathDisplay = ({target}) => {
 TargetPathDisplay.propTypes = {
   target: PropTypes.shape({
     targetPath: PropTypes.arrayOf(PropTypes.string).isRequired,
+    isRequired: PropTypes.bool.isRequired,
   }).isRequired,
-  isRequired: PropTypes.bool.isRequired,
 };
 
 export default ObjectsApiVariableConfigurationEditor;
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
@@ -1,49 +1,120 @@
-import React from 'react';
+import {Form, Formik} from 'formik';
+import React, {useContext} from 'react';
 import {FormattedMessage} from 'react-intl';
+import {useAsync} from 'react-use';
 
+import {FormContext} from 'components/admin/form_design/Context';
 import Select from 'components/admin/forms/Select';
+import {get} from 'utils/fetch';
+
+// TODO: properly translate this
+const LOADING_OPTION = [['...', 'loading...']];
 
 /**
  * Returns the Objects API Configuration editor modal for a specific variable. This only applies to V2 Options
  *
  * @typedef {{
- *   variablesMapping: {variableKey: string, targetPath: string[]}[]
- * }} ObjectsAPIV2Options
+ *   backend: string;
+ *   key: string;
+ *   name: string;
+ *   options: {
+ *     variablesMapping: {variableKey: string, targetPath: string[]}[]
+ *   }
+ * }} ObjectsAPIRegistrationBackend
  *
  * @param {Object} p
- * @param {string} p.variable - The current variable
- * @param {ObjectsAPIV2Options} p.backendOptions - The Objects API Options (guaranteed to be v2)
+ * @param {Object} p.variable - The current variable
+ * @param {ObjectsAPIRegistrationBackend} p.backend - The Objects API registration backend (options guaranteed to be v2)
  * @returns {JSX.Element} - The summary, represented as a the parts of the target path separated by '>'
  */
-const ObjectsApiVariableConfigurationEditor = ({variable, backendOptions}) => {
+const ObjectsApiVariableConfigurationEditor = ({variable, backend}) => {
+  const {
+    form: {uuid},
+  } = useContext(FormContext);
+
+  const {
+    loading,
+    value: targetPaths,
+    error,
+  } = useAsync(async () => {
+    const response = await get('/api/v2/registration/plugins/objects-api/target-paths', {
+      backendKey: backend.key,
+      formUuid: uuid,
+      variableKey: variable.key,
+    });
+
+    if (!response.ok) {
+      throw new Error('Loading available target paths failed');
+    }
+
+    return response.data;
+  }, []);
+
+  const choices =
+    loading || error
+      ? LOADING_OPTION
+      : targetPaths.map(t => [t.targetPath, t.targetPath.join(' > ')]);
+
   return (
-    <table>
-      <thead>
-        <tr>
-          <th>
-            <FormattedMessage defaultMessage="Variable name" description="'Variable name' label" />
-          </th>
-          <th>
-            <FormattedMessage defaultMessage="Variable key" description="'Variable key' label" />
-          </th>
-          <th>
-            <FormattedMessage
-              defaultMessage="JSON Schema target"
-              description="'JSON Schema target' label"
-            />
-          </th>
-        </tr>
-      </thead>
-      <tbody>
-        <tr>
-          <td>{variable.name}</td>
-          <td>{variable.key}</td>
-          <td>
-            <Select choices={[['test', 'test']]} />
-          </td>
-        </tr>
-      </tbody>
-    </table>
+    <Formik
+      initialValues={{
+        targetPath:
+          backend.options.variablesMapping.find(mapping => mapping.variableKey === variable.key)
+            .targetPath || '',
+      }}
+    >
+      {formik => (
+        <Form>
+          <table>
+            <thead>
+              <tr>
+                <th>
+                  <FormattedMessage
+                    defaultMessage="Variable name"
+                    description="'Variable name' label"
+                  />
+                </th>
+                <th>
+                  <FormattedMessage
+                    defaultMessage="Variable key"
+                    description="'Variable key' label"
+                  />
+                </th>
+                <th>
+                  <FormattedMessage
+                    defaultMessage="JSON Schema target"
+                    description="'JSON Schema target' label"
+                  />
+                </th>
+                <th>
+                  <FormattedMessage
+                    defaultMessage="JSON Schema info"
+                    description="'JSON Schema info' label"
+                  />
+                </th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>{variable.name}</td>
+                <td>{variable.key}</td>
+                <td>
+                  <Select
+                    id="targetPath"
+                    name="targetPath"
+                    choices={choices}
+                    {...formik.getFieldProps('targetPath')}
+                  />
+                </td>
+                <td>
+                  <pre>{JSON.stringify({}, null, 2)}</pre>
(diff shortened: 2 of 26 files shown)


src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -1,19 +1,53 @@
...
+  };
+
+  const saveRegistrationOptions = () => {
+    const newOptions = getOptions();
+    onRegistrationOptionsChange(
+      produce(backend, draft => {
+        draft.options = newOptions;
+      })
+    );
+  };
```
Viicos, line 50: I am under the impression that trying to have the save button outside the `variableConfigurationEditor` leads to really dirty code:
- It is impossible to reason about: `getOptions` is a callback set by the configuration editor, by using a `useState` hook in the parent `RegistrationsSummary` component:
https://github.com/open-formulieren/open-forms/blob/d8f15e192d03400aa0dd9d755e5140adc9be49e6/src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js#L128-L130
And this is set in the configuration editor component like so:
https://github.com/open-formulieren/open-forms/blob/d8f15e192d03400aa0dd9d755e5140adc9be49e6/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js#L35-L41
So we end up with logic split up in three different components
- In the configuration editor, I can't make use of the `onSubmit` prop of `Formik` as the save button isn't in this component. And I am now stuck because I can't fetch the formik values outside of the `<Formik>` component

So I think it's better to have the save button explicitly handled by each backend option editor

src/openforms/js/components/admin/form_design/variables/VariablesEditor.js
```diff
@@ -51,14 +51,15 @@ const VariablesEditor = ({variables, onAdd, onChange, onDelete}) => {
           </TabList>
 
           <TabPanel>
-            <VariablesTable variables={componentVariables} />
+            <VariablesTable variables={componentVariables} onFieldChange={onFieldChange} />
```
Viicos, line 54: I haven't added all the missing props as this panel is for component variables which are readonly. So only the `onFieldChange` prop is relevant here (used to edit the registration options, the already existing `onChange` dispatches `CHANGE_USER_DEFINED_VARIABLE` so I can't use it :/)

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,151 @@
...
+
+  const getTargetPath = pathSegment =>
+    targetPaths.find(t => isEqual(t.targetPath, JSON.parse(pathSegment)));
+
+  const choices =
+    loading || error
+      ? LOADING_OPTION
+      : targetPaths.map(t => [
+          JSON.stringify(t.targetPath),
+          `${t.targetPath.join(' > ')}${t.required ? ' (required)' : ''}`,
```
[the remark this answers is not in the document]
Viicos: Does this work?

```js
`${t.targetPath.join(' > ')}${t.required ? ` (${intl.formatMessage(...)})` : ''}`
```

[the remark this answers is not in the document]
sergei-maertens: technically yes, but you're breaking up a string outside of localization and depending on the language, words may require re-ordering.

The advice from ICU/react-intl is to put as much as possible inside the translation string so the translators have maximum flexibility.

You can make your life easier by defining your own helper component `<TargetPathDisplay target={t} />` with

```jsx
const TargetPathDisplay = ({target}) => {
    const intl = useIntl();
    const path = target.targetPath.join(' > ');
    return (
        <FormattedMessage
            description="Representation of a json-schema target path"
            defaultMessage="{required, select, true {{path} (required)}, other {{path}}}"
            values={{path, required: target.required}}
        />
    );
};
```

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,169 @@
...
+  );
+  let mappedVariable = variablesMapping[index];
+  if (index === -1) {
+    // if not found, grab the next available index to add it as a new record/entry
+    index = variablesMapping.length;
+    mappedVariable = {
+      variableKey: variable.key,
+      targetPath: undefined,
+    };
+    setFieldValue(`variablesMapping.${index}`, mappedVariable);
```
sergei-maertens, line 54: you really need to get used to perform these sort of updates in an effect, not during rendering :grimacing: 

these formik callbacks do more than just updating state too, they trigger validation runs, so you can get in a render-cascade. The `useState` callback can be called during render, but only inside the same component as React somehow gives that special treatment (?)

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,169 @@
...
+        )}
+      </div>
+    </Fieldset>
+  );
+};
+
+const TargetPathSelect = ({name, index, choices}) => {
+  const {getFieldProps, setFieldValue} = useFormikContext();
+  const props = getFieldProps(name);
+  console.log(props, choices);
```
sergei-maertens, line 144:
```suggestion
```

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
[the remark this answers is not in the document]
Viicos: I wish we all spoke the same language..

src/openforms/js/components/admin/form_design/story-decorators.js
```diff
@@ -1,6 +1,6 @@
 import {FeatureFlagsContext, FormContext} from 'components/admin/form_design/Context';
 
-import {FormLogicContext} from './Context';
+import {APIContext, FormLogicContext} from './Context';
```
sergei-maertens, line 3: unused import?

sergei-maertens (changes requested):
Please also include prop types for all the internal components like the rest of the code does.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
Viicos: My understanding is that this code path won't be hit more than once, as the variablesMapping will then contain the newly added entry?

Viicos: And I also don't understand why there isn't a better api with functional components to initialize once other than using `useEffect` with empty dependencies..

src/openforms/js/components/admin/form_design/Context.js
```diff
@@ -7,7 +7,7 @@ const FeatureFlagsContext = React.createContext({});
 FeatureFlagsContext.displayName = 'FeatureFlagsContext';
 
 const FormContext = React.createContext({
-  form: {url: ''},
+  form: {url: '', uuid: ''},
```
Viicos, line 10: This change was once required to make my tests working, but I think it is no longer necessary. Kept it as it could be useful in the future.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+ * @typedef {{
+ *   version: 1 | 2;
+ *   objecttype: string;
+ *   objecttypeVersion: number;
+ *   variablesMapping: {variableKey: string, targetPath: string[]}[];
+ * }} ObjectsAPIRegistrationBackendOptions
+ *
+ * @param {Object} p
+ * @param {Object} p.variable - The current variable
+ * @returns {JSX.Element} - The summary, represented as a the parts of the target path separated by '>'
```
SilviaAmAm, line 31:
```suggestion
 * @returns {JSX.Element} - The summary, represented as the parts of the target path separated by '>'
```

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+    loading,
+    value: targetPaths,
+    error,
+  } = useAsync(
+    async () => {
+      const response = await post(REGISTRATION_OBJECTS_TARGET_PATHS, csrftoken, {
+        objecttypeUrl: objecttype,
+        objecttypeVersion,
+        variableJsonSchema: asJsonSchema(variable),
+      });
```
SilviaAmAm, line 71: I think that the post already does the `throwForStatus` thing, so I am not sure if the if statement below is needed?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+    // Load only once:
+    []
+  );
+
+  const getTargetPath = pathSegment => targetPaths.find(t => isEqual(t.targetPath, pathSegment));
+
+  const choices =
+    loading || error
+      ? LOADING_OPTION
+      : targetPaths.map(t => [JSON.stringify(t.targetPath), <TargetPathDisplay target={t} />]);
```
SilviaAmAm, line 88: Does it make sense to say that it's loading in case of an error?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * @param {Object} p
+ * @param {Object} p.variable - The current variable
+ * @returns {JSX.Element} - The summary, represented as a the parts of the target path separated by '>'
+ */
+const ObjectsApiVariableConfigurationEditor = ({variable}) => {
+  const {csrftoken} = useContext(APIContext);
+
+  const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
+  const {values: backendOptions, getFieldProps, setFieldValue} = useFormikContext();
```
SilviaAmAm, line 37:
```suggestion
  const {values: backendOptions, getFieldProps} = useFormikContext();
```
I don't think it's used here?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+      )}
+    />
+  );
+};
+
+TargetPathSelect.propTypes = {
+  name: PropTypes.string.isRequired,
+  index: PropTypes.number.isRequired,
+  choices: PropTypes.array.isRequired,
+};
```
SilviaAmAm, line 194: `mappedVariable` is missing in the propTypes

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+    />
+  );
+};
+
+TargetPathDisplay.propTypes = {
+  target: PropTypes.shape({
+    targetPath: PropTypes.arrayOf(PropTypes.string).isRequired,
+  }).isRequired,
+  isRequired: PropTypes.bool.isRequired,
+};
```
SilviaAmAm, line 212:
```suggestion
TargetPathDisplay.propTypes = {
  target: PropTypes.shape({
    targetPath: PropTypes.arrayOf(PropTypes.string).isRequired,
    isRequired: PropTypes.bool.isRequired,
  }).isRequired,
};
```
Maybe it was meant to be inside?

src/openforms/js/components/admin/forms/Select.js
```diff
@@ -25,7 +31,7 @@ const Select = ({
 
   const prefix = useContext(PrefixContext);
   name = prefix ? `${prefix}-${name}` : name;
-  const options = allowBlank ? [BLANK_OPTION].concat(choices) : choices;
+  const options = allowBlank ? BLANK_OPTION.concat(choices) : choices;
```
SilviaAmAm, line 34: Does this not cause problems with the 'original' `BLANK_OPTION` being edited?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+ */
+const ObjectsApiVariableConfigurationEditor = ({variable}) => {
+  const {csrftoken} = useContext(APIContext);
+
+  const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
+  const {values: backendOptions, getFieldProps, setFieldValue} = useFormikContext();
+
+  /** @type {ObjectsAPIRegistrationBackendOptions} */
+  const {objecttype, objecttypeVersion, variablesMapping, version} = backendOptions;
+  if (version !== 2) throw new Error('Not supported, must be config version 2.');
```
vaszig, line 41: Don't we need a translatable error here (and basically for this kind of errors)?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
Viicos: I will test when I add proper error handling in a follow up PR

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
Viicos: Follow up PR will take care of error handling, so it will display a proper error message

src/openforms/js/components/admin/forms/Select.js
Viicos: It is only used in this file and not imported anywhere :+1:

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,214 @@
...
+      values={{path, required: target.isRequired}}
+    />
+  );
+};
+
+TargetPathDisplay.propTypes = {
+  target: PropTypes.shape({
+    targetPath: PropTypes.arrayOf(PropTypes.string).isRequired,
+  }).isRequired,
+  isRequired: PropTypes.bool.isRequired,
```
vaszig: I don't see that in the received arguments of `TargetPathDisplay`

vaszig (commented):
A couple of remarks-questions and minor remarks because I am not exactly sure if some of these will be part of the following PRs.

src/openforms/js/components/admin/form_design/variables/registration/RegistrationsSummary.js
```diff
@@ -29,31 +48,75 @@ const RegistrationSummary = ({name, registrationSummary}) => {
...
+      </FormModal>
     </>
   );
 };
 
+RegistrationSummary.propTypes = {
+  name: PropTypes.string.isRequired,
+  variable: PropTypes.object.isRequired,
+  backend: PropTypes.object.isRequired,
+  backendIndex: PropTypes.number,
```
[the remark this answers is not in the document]
Viicos: It is indeed

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
Viicos: This will be a bare error logged in the console, so no translations are required. In this case, this is just a sanity check, as we already made sure before that an objects api backend with `version==1` would not be configurable from the variables.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
Viicos: Wasn't necessary

SilviaAmAm (approved):

PR #3918, opened 2024-02-21
:alien: [#3049] Ensure the data-theme attribute is set

sergei-maertens:
Set to auto so that the browser preferences and storybooks default dark/light mode tool (hopefully) work out of the box without requiring an explicit theme switcher.


SilviaAmAm (approved):
Changing the browser preferences in firefox and setting the dark theme is picked up by storybook.

PR #3919, opened 2024-02-21
:sparkles: [#3885] add open/collapsed for form categories

### LINKED ISSUE #3885: Admin form list - remember open/collapsed state of categories
The forms in the list are displayed in categories, where the current implementation always opens the top-level categories and closes the nested categories. However, this can still lead to a cluttered UI when you have many forms listed (within a category).

I'd like to open/closed state of a category to be persisted so that navigating/refreshing the page doesn't undo my category management. It's okay to store this in the browser `localStorage` - storing this server-side is overkill.

nikkiysendoorn1:
Closes #3885

### DIFF
--- a/src/openforms/forms/templates/admin/forms/form/change_list.html
+++ b/src/openforms/forms/templates/admin/forms/form/change_list.html
@@ -55,7 +55,7 @@
             {% for category in categories %}
                 <tbody>
                     <tr
-                        class="form-category {% if not category.is_leaf or category.form_count %}form-category--has-children{% endif %} {% if category.is_leaf %}form-category--leaf{% endif %} {% if category.is_leaf and category.depth > 1 %}form-category--collapsed{% endif %}"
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):

sergei-maertens (commented):
Gaat nog niet helemaal lekker met geneste categoriëen :grimacing: 

image

sergei-maertens:
Hrm, testing this out locally still doesn't seem to behave as I'd expect :thinking: I've cleared my local storage before testing.

My categories:

image

Initial load of the form list:

image

Collapsing all top-level categories:

image

Then refresh page (all good):

image

Expand "Debugging":

image

After page refresh (all good):

image

Expand "Github issues" (all good):

image

After page refresh "Github issues" is collapsed instead of open:

image

Then I expand Github issues again, and afterwards collapse "Debugging":

image

After refreshing the page this looks as expected.

---

I saw some console errors at some point, but I can't reproduce that anymore now. So the one thing that doesn't behave as expected is that nested categories that are recorded as "open" do not stay open after refreshing. If it's too much work to sort that out, I can live with it :)

sergei-maertens (approved):

PR #3925, opened 2024-02-22
[#3688] Add API endpoint to return available target paths

Viicos:
Part of #3688.

This is the backend part of https://github.com/open-formulieren/open-forms/pull/3917.

# PR Summary

- The frontend needs to display the compatible JSON Schema targets [^1] for a specific variable. Because variables might not exist yet in the backend, the frontend instead does a bit logic to construct a JSON Schema for a specific variable (see [this code](https://github.com/open-formulieren/open-forms/blob/9a6fa7f86c0df8e6334335bf3384645f6d531207/src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js#L33-L62) in the frontend PR) and sends this schema to the backend.

This is not really ideal as usually we would want this kind of logic to live in the backend, but this is the only option for now.

- The backend fetches the JSON Schema corresponding to the objecttype and version provided by the frontend, iterates over the possible paths using the [`iter_json_schema_paths`](https://github.com/open-formulieren/open-forms/blob/b00d33bedf16b43adf5ef2446c31f06454685f28/src/openforms/registrations/contrib/objects_api/json_schema.py#L61-L63) utility function. For each path, if the corresponding (sub) JSON Schema at that path matches (see `json_schema_matches`) the one provided by the frontend, it is kept.

- The `test_list_target_paths` test shows all this logic. It makes use of VCR, and it uses the configured docker instance. The fixture used can be found [here](https://github.com/open-formulieren/open-forms/blob/b00d33bedf16b43adf5ef2446c31f06454685f28/docker/objects-apis/fixtures/objecttypes_api_fixtures.json#L95-L109) (so that you can make sense of the data used in the test).

[^1]: a "JSON Schema target" is really similar to a dotted formio path. For the JSON document `{"a": {"b.c": 1}}`, we represent the path to `1` as `["a", "b.c"]`. The frontend then displays it in a pretty way: `a > b.c`.

### DIFF
--- a/src/openforms/forms/models/form_variable.py
+++ b/src/openforms/forms/models/form_variable.py
@@ -15,7 +15,11 @@ from openforms.formio.utils import (
 )
 from openforms.formio.validators import variable_key_validator
 from openforms.prefill.constants import IdentifierRoles
-from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
+from openforms.variables.constants import (
+    COMPATIBLE_JSON_SCHEMA_TYPES,
+    FormVariableDataTypes,
+    FormVariableSources,
+)
 from openforms.variables.utils import check_initial_value
 
 from .form_definition import FormDefinition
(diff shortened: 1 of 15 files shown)


src/openforms/forms/models/form_variable.py
```diff
@@ -222,6 +222,9 @@ class Meta:
     def __str__(self):
         return _("Form variable '{key}'").format(key=self.key)
 
+    def matches_json_schema(self, json_schema: dict[str, Any]) -> bool:
+        pass
+
```
Viicos, line 227: I initially had something like `as_json_schema(self) -> dict[str, Any]` in mind. But as discussed to have the compatibility logic simple at first, producing JSON Schemas for any variable isn't possible in all cases. So this method will take care of implementing custom logic instead of comparing `as_json_schema() == sub_json_schema`

sergei-maertens (changes requested):
Feedback provided in person: https://github.com/open-formulieren/open-forms/pull/3917#discussion_r1507341692

src/openforms/registrations/contrib/objects_api/api/views.py
```diff
@@ -49,3 +60,50 @@ class ObjecttypeVersionsListView(ListMixin, views.APIView):
...
+            )
+            if not isinstance(json_schema, InvalidReference)
+            if json_schema_matches(
+                input_serializer.validated_data["variable_json_schema"], json_schema
+            )
+        ]
+
+        output_serializer = TargetPathsSerializer(many=True, data=return_data)
+        output_serializer.is_valid(raise_exception=True)
+        return Response(data=output_serializer.validated_data)
```
Viicos, line 109: Probably best to reraise a 500 instead here? A validation error on the output serializer means something is wrong on our end, not from the API consumer.

sergei-maertens: discussedin office, should use `instance=` kwarg instead of `data` and then you can output with `serializer.data` which doesn't require `is_valid` to be called (that doesn't make sense in output serializers)

src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -168,3 +168,39 @@ def get_missing_required_paths(
...
+        target_types = [target_types]
+
+    variable_types: str | list[str] = variable_schema["type"]
+    if not isinstance(variable_types, list):
+        variable_types = [variable_types]
+
+    if not set(variable_types).issubset(target_types):
+        return False
+
+    if "string" in target_types and (target_format := target_schema.get("format")):
```
SilviaAmAm, line 200: Would a case be possible where the `variable_format` is not `None` but the `target_format` is? In this case I think the `json_schema_matches` function would return `True`, but I am not sure if it is intended?

Viicos: I added a test covering this behavior. It is hard to think about all the possible cases :sweat_smile:

PR #3927, opened 2024-02-22
Squashing migrations

sergei-maertens:
* Require a minimum OF version of 2.5 before you can upgrade to 2.6
* Converted earlier squashed migrations into normal migrations (per https://docs.djangoproject.com/en/4.2/topics/migrations/#migration-squashing)
* Removed `RunPython` operations - they are guaranteed to have been executed because the starting state for upgrades is OF 2.5.0
* Deleted obsoleted tests
* Squashed old migrations (pre 2.5.0) again


vaszig (approved):
I don't understand which part of the code is causing the codecov to fail.

sergei-maertens:
> I don't understand which part of the code is causing the codecov to fail.

Deleting covered code will naturally result in that. Nothing you can do about it.

PR #3933, opened 2024-02-23
[#3359] Refactor pre registration 

### LINKED ISSUE #3359: Registration plugin flow could be more developer friendly
### Thema / Theme

Plugins

The registration flow was broken up recently into more steps:

1. pre-registration
2. generate PDF
3. main registration

but the base plugin classes have not been updated appropriately.

I've developed a plugin for a registration service that generates its own references. This requires me to use the hook `pre_register_submission` and store the reference + relevant data directly on the submission model:

```py
        submission.registration_result = {
            "id": data["id"],
            "reference": data["reference"],
            "qrcode_data": data["qrcode_data"],
        }
        submission.public_registration_reference = data["reference"]
        submission.save()
```

The previous API (`register_submission` and `get_reference_from_result`) did not rely on internals - instead I could reason 100% in terms of what my registration backend returns.

To make matters worse, `get_reference_from_result` and `register_submission` no longer have any purpose - all my registration is already done in the pre-registration hook, but the base plugin class enforces me to implement these two methods.

Looking at the test suite, there are many tests that call `get_reference_from_result` directly, but this method isn't really called anymore during the registration flow, the expectation is that a reference is set and saved during pre-registration (as the PDF generation tries to read it, which runs after pre-registration!)

1. Better developer experience for registration plugin developers
2. Backwards compatibility (my existing plugin was now broken after updating to latest open forms)

I think it'd make sense to expect pre_register_submission return a `RegistrationResult` dataclass, which has `result: dict` and `reference: str` fields (optional). The tasks should then extract that information and save it on the submission.

It would probably also still be nice to call the `get_reference_from_result` hook after running pre-registration, instead of expecting the plugin to set the field value directly on the model.

Finally, `register_submission` should probably receive an explicit `result` field, as the pre-registration task may need to store temporary data and do the expensive processing after PDF generation.

SilviaAmAm:
Fixes #3359


src/openforms/registrations/tasks.py
```diff
@@ -92,6 +92,16 @@ def pre_registration(submission_id: int, event: PostSubmissionEvents) -> None:
...
 
+    if result.reference is None:
+        set_submission_reference(submission)
+    else:
+        submission.public_registration_reference = result.reference
+
+    if result.data is not None:
+        if not submission.registration_result:
+            submission.registration_result = {}
+        submission.registration_result.update(result.data)
```
Viicos, line 103: What would be the use case where there is already a `registration_result` set when this `pre_registration` task is being called?

Unrelated: this makes me wonder if there would be a use case where the registration result isn't a `dict` :thinking:
I guess dict is the one that makes most sense.

Viicos (approved):

src/openforms/registrations/tasks.py
SilviaAmAm: I think so: if the `pre_registration` fails, then the traceback is saved in `registration_result`. So when the chain is retried, it would be already set.

SilviaAmAm: > this makes me wonder if there would be a use case where the registration result isn't a dict 🤔

Mhmm I guess that if a plugin sets it to something that is not an object it could be problematic :thinking:

SilviaAmAm: > Similarly, could Submission.save_registration_status be used?

I wasn't sure if I should use it, since the success of the `pre_registration` doesn't affect the 'status'. So I thought that it may be confusing. But it could be used like this:
```python
submission.save_registration_status(
            submission.registration_result, result.data
)
```
> shouldn't we remove the traceback key from the registration result if it succeeded?

I am also not sure if it needs to remain for possible debugging :thinking:

sergei-maertens (commented):
That's a bigger cleanup than I anticipated, but I'm very happy with it!

src/openforms/registrations/base.py
```diff
@@ -38,33 +45,18 @@ def register_submission(
...
+        no need to implement this method.
         """
-        pass
-
-    def obtain_submission_reference(
-        self, submission: "Submission", options: dict
-    ) -> None:
-        """Obtain the reference number generated by the registration backend and set it on the submission."""
-        pass
+        return PreRegistrationResult()
```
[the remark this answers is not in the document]
SilviaAmAm: I find it a bit redundant, since we need an `is_empty` for both the reference and the data. So i think that it's okay to just look if they are empty?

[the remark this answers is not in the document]
sergei-maertens: yeah the data is not empty as a whole, just parts, so okay to leave it as is

sergei-maertens (approved):

sergei-maertens:
@Viicos sorry, this will introduce merge conflicts in #3954 :grimacing:

Viicos:
> @Viicos sorry, this will introduce merge conflicts in #3954 😬

Yes saw that yesterday :smiling_face_with_tear:

sergei-maertens:
To make it better, there were already conflicts because of Vasileios cleanups w/r to documents API anyway :grimacing:

PR #3934, opened 2024-02-26
[#3890] Allow hidden fields to be sent to registration systems

### LINKED ISSUE #3890: Allow hidden fields to be sent to registration systems
### Thema / Theme

Admin

Hidden fields are currently not sent to the registration backend. Can we make this happen, optionally?

Request comes from Utrecht who wants this, using the Objects API where it's added to `{{ data }}` so it automatically ends up in the registration system. It was not enough that they could manually add it elsewhere in the JSON field.

_No response_

_No response_

vaszig:
Fixes #3890

### DIFF
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -143,7 +143,9 @@ class ComponentNode(Node):
                 self.component, artificial_repeating_group_data
             ):
                 return False
-        elif not is_visible_in_frontend(self.component, self.step_data):
+        elif self.mode != RenderModes.registration and not is_visible_in_frontend(
+            self.component, self.step_data
+        ):
             return False
 
         render_configuration = RENDER_CONFIGURATION[self.mode]
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -125,7 +125,8 @@ class ComponentNode(Node):
         from .default import EditGridGroupNode
 
         # everything is emitted in export mode to get consistent columns
-        if self.mode == RenderModes.export:
+        # the same happens with the registration in order to include hidden fields as well
+        if self.mode in {RenderModes.export, RenderModes.registration}:
             return True
 
         # explicitly hidden components never show up. Note that this property can be set
@@ -143,9 +144,7 @@ class ComponentNode(Node):
                 self.component, artificial_repeating_group_data
             ):
                 return False
-        elif self.mode != RenderModes.registration and not is_visible_in_frontend(
-            self.component, self.step_data
-        ):
+        elif not is_visible_in_frontend(self.component, self.step_data):
             return False
 
         render_configuration = RENDER_CONFIGURATION[self.mode]
--- a/src/openforms/formio/rendering/tests/test_component_node.py
+++ b/src/openforms/formio/rendering/tests/test_component_node.py
@@ -323,6 +323,46 @@ class FormNodeTests(TestCase):
             self.assertEqual(len(nodelist), 1)
             self.assertEqual(nodelist[0].label, "A container without visible children")
 
+    def test_explicitly_hidden_component_not_skipped_when_registration(self):
+        # we always need a renderer instance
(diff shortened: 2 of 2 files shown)


src/openforms/formio/rendering/tests/test_component_node.py
```diff
@@ -323,6 +323,46 @@ def test_explicitly_hidden_component_skipped(self):
...
+    def test_explicitly_hidden_component_not_skipped_when_registration(self):
+        # we always need a renderer instance
+        renderer = Renderer(
+            self.submission, mode=RenderModes.registration, as_html=False
+        )
+
+        with self.subTest("Simple hidden leaf component"):
+            component = self.step.form_step.form_definition.configuration["components"][
+                1
+            ]
```
Viicos, line 335: Unfortunate styling :cry:

src/openforms/formio/rendering/nodes.py
```diff
@@ -143,7 +143,9 @@ def is_visible(self) -> bool:
                 self.component, artificial_repeating_group_data
             ):
                 return False
-        elif not is_visible_in_frontend(self.component, self.step_data):
+        elif self.mode != RenderModes.registration and not is_visible_in_frontend(
```
Viicos, line 146: you can use

```suggestion
        elif self.mode is not RenderModes.registration and not is_visible_in_frontend(
```

(I see this `!=` / `==` is used as well in the function already, I think they can be changed as well)

Viicos (approved):
Didn't know the rendering module was used for registrations, it seems to only be used for email and objects api registration?

src/openforms/formio/rendering/nodes.py
sergei-maertens: I don't fully agree with this, it's down to an implementation detail of how enum is used, but I think in practical reasons you can compare this is a string comparison which is typically done with `==` and `!=`. Let's stick to the existing usage rather than changing the existing code.

sergei-maertens, line 146: I don't think this is right, because now it falls back to render configuration (either explicit or using a default) which depends on the component type.

Probably a more correct approach is to *always* consider a component visible in registration mode, like the export above, which should be changed to:

```py
        if self.mode in {RenderModes.export, RenderModes.registration}:
            return True
```
we want to bypass `is_visible_in_frontend` entirely during registration, however I'm not sure how this plays with the registration email backend :thinking:

sergei-maertens (changes requested):

src/openforms/formio/rendering/nodes.py
Viicos: Well according to docs equality comparison seems to be the implementation detail, not identity (`Enumeration members are compared by identity` / `Equality comparisons are defined though`).

But fine to keep it as is

sergei-maertens: > Equality comparisons are defined though:

That's explicit support for equality comparisons, which is different than an implementation detail. It means you can rely on it and it's guaranteed to work as expected, or will come with appropriate release notes if the behaviour changes :wink: 

The fact that enum is used under the hood however is the implementation detail, as that happens to be how Django implements its `models.*Choices` classes. Unfortunately Django docs don't seem to give a comparison example, they do give examples of comparing `member in {Foo.bar, Foo.baz}` though. Which in turn leads to the implementation detail of how `set.__contains__` is implemented: identity or equality. I suspect it's the latter.

Viicos: > The fact that enum is used under the hood however is the implementation detail

Ah ok I thought you meant the identity comparison is the implementation detail of the enum class, but when taking django choices into account it makes sense

vaszig: I have tested the `email registration`  and I don't see a problem. Do you mean something like showing the hidden field's value for example?

sergei-maertens: Yes, that's what I was a bit worried about! Thanks for testing :)

sergei-maertens (approved):

PR #3936, opened 2024-02-26
Add setting & support redirecting to OpenID Connect identity provider

### LINKED ISSUE #3695: Improve login screen and process
### Thema / Theme

Admin

The login process is suboptimal. 2FA is globally set but when using OIDC this should be decided by the OIDC-provider. If OIDC is used, you can still use local accounts. This should be discouraged visually.

**Tasks**
 
* [x] Update the two-factor library to use upstream version again
* [x] Differentiate between OIDC-login and regular logins whether to use 2FA or not (thus, make sure this is not some global setting but based on the session that holds the login-backend information). This probably require library code changes or some custom code in Open Formulieren.
* [x] Make the Django `LOGIN_URL` an environment variable, and document it. This allows us to set it to `/oidc/authenticate/` to automatically go to OIDC provider instead of showing the regular login form.

2FA upgrade is also required for #3049 

_No response_

_No response_

sergei-maertens:
Closes #3695

* Added & documented setting to automatically do OIDC authenticate endpoint redirect
* Broke up classic username/password login screen to use different URL
* `/admin/login/` now redirects to either classic login or OIDC provider depending on value of setting

### DIFF
--- a/docs/configuration/general/oidc.rst
+++ b/docs/configuration/general/oidc.rst
@@ -86,3 +86,6 @@ Klik tot slot rechtsonder op **Opslaan**.
 Je kan vervolgens het makkelijkst testen of alles werkt door in een incognitoscherm
 naar ``https://open-formulieren.gemeente.nl/admin/`` te navigeren en op 
 *Inloggen met organisatieaccount* te klikken.
+
+.. note:: We raden aan om Open Formulieren te deployen met de ``USE_OIDC_FOR_ADMIN_LOGIN=1``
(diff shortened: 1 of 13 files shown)


Viicos (approved):
Is this going to be included in the default project as well?

sergei-maertens:
> Is this going to be included in the default project as well?

I'm not sure - default project currently does not have the OIDC provider setup in it and I don't know if it should. We'll see what Joeri thinks :)

PR #3937, opened 2024-02-26
[#3688] Fix some edge cases when parsing JSON Schemas

Viicos:
Part of #3688

- `type` can be an array
- Exception when trying to iterate over a reference which points to a non-`object` sub schema

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/json_schema.py
+++ b/src/openforms/registrations/contrib/objects_api/json_schema.py
@@ -76,31 +76,35 @@ def iter_json_schema_paths(
     def _iter_json_schema(
         json_schema: ObjectSchema, parent_json_path: JsonSchemaPath
     ) -> Iterator[tuple[JsonSchemaPath, ObjectSchema | InvalidReference]]:
-        assert json_schema.get("type") == "object"
 
         yield parent_json_path, json_schema
 
-        required = json_schema.get("required", [])
+        type_list: str | list[str] = json_schema.get("type", [])
+        if not isinstance(type_list, list):
+            type_list = [type_list]
 
-        k: str
-        for k, v in json_schema["properties"].items():
-            json_path = parent_json_path / k
-            json_path.required = k in required
+        if "object" in type_list and "properties" in json_schema:
+            required = json_schema.get("required", [])
 
-            match v:
-                case {"type": "object"}:
-                    yield from _iter_json_schema(v, json_path)
-                case {"$ref": str(uri)}:
-                    try:
-                        resolved = resolver.lookup(uri)
-                    except Unresolvable as exc:
-                        if fail_fast:
-                            raise
-                        yield json_path, InvalidReference(uri, exc)
-                    else:
-                        yield from _iter_json_schema(resolved.contents, json_path)
-                case {}:
-                    yield json_path, v
+            k: str
+            for k, v in json_schema["properties"].items():
+                json_path = parent_json_path / k
+                json_path.required = k in required
+
+                match v:
+                    case {"type": "object"}:
+                        yield from _iter_json_schema(v, json_path)
+                    case {"$ref": str(uri)}:
+                        try:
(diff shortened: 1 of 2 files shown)


src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -76,31 +76,35 @@ def iter_json_schema_paths(
...
-                    else:
-                        yield from _iter_json_schema(resolved.contents, json_path)
-                case {}:
-                    yield json_path, v
+        type_list: str | list[str] = json_schema.get("type", [])
+        if not isinstance(type_list, list):
+            type_list = [type_list]
+
+        if "object" in type_list and "properties" in json_schema:
+            required = json_schema.get("required", [])
```
sergei-maertens, line 87: the `type` can be absent, this is valid:

```yaml
required:
  - foo
properties:
  foo:
    type: string
```

and is equivalent to:

```yaml
required:
  - foo
type: object  # or ['object']
properties:
  foo:
    type: string
```

src/openforms/registrations/contrib/objects_api/json_schema.py
```diff
@@ -76,31 +76,35 @@ def iter_json_schema_paths(
...
+
+            k: str
+            for k, v in json_schema["properties"].items():
+                json_path = parent_json_path / k
+                json_path.required = k in required
+
+                match v:
+                    case {"type": "object"}:
+                        yield from _iter_json_schema(v, json_path)
```
sergei-maertens, line 96:
```suggestion
                    case {"type": "object"} | {"properties": dict}:
                        yield from _iter_json_schema(v, json_path)
                    case {"type": [*types]} and "object" in types:
                        yield from _iter_json_schema(v, json_path)
```
I think?

This would also need the check that `object` is in the nested `type` array.

The pattern matching itself can probably be done in some clever way, or perhaps you would do the matching only on the `properties` key and add an `assert` inside the match:

```py
type_ = v.get("type", "object")
assert isinstance(type_, (str, list))
assert type_ == "object" or "object" in type_
```

src/openforms/registrations/contrib/objects_api/json_schema.py
Viicos: I'm not sure this is the case. If the `type` is absent, any data should validate. A thread is available [here](https://github.com/json-schema/json-schema/issues/172#issuecomment-114076650).

Unfortunately, this isn't explicitly stated in the spec (or I couldn't find it). However, the [introduction](https://json-schema.org/learn/getting-started-step-by-step#introduction-to-json-schema) seems to imply it:

> The most basic schema is a blank JSON object, which constrains nothing, allows anything, and describes nothing:
> `{}`
> By adding validation keywords to the schema, you can apply constraints to an instance. For example, you can use the `type` keyword to constrain an instance to an object, array, string, number, boolean, or null.

I'm surprised to see the Python `jsonschema` library does not respect this, nor does this [online validator](https://www.jsonschemavalidator.net/).

Viicos: Hum in fact apparently not: https://github.com/OAI/OpenAPI-Specification/issues/1388#issuecomment-339051239

sergei-maertens: ah yeah, I missed some nuance, TIL!

I was using that same JSON schema validator, so definitely didn't use the spec or a python library to validate my statement.

I think we should take a pragmatic approach here for our practical means. Maybe it's a good idea to create an issue in the object types API to show some warnings if such schemas get uploaded (missing a `type` key) because they may unintentionally be too loose.

Viicos: Yes I ended up not relying on the `type` value, and only assert it is correctly set to `object` if provided and `properties` is set

sergei-maertens (approved):

PR #3938, opened 2024-02-27
:lock: Obfuscate incorrect password in Axes logs

stevenbal:
From: https://github.com/maykinmedia/open-inwoner/pull/1053#discussion_r1503072563


sergei-maertens (approved):
Please also add this to default-project!

PR #3939, opened 2024-02-27
:lipstick: Display errors above input field

SilviaAmAm:
Before:
Screenshot from 2024-02-27 10-40-34

After:
Screenshot from 2024-02-27 10-39-12

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
+++ b/src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
@@ -185,7 +185,7 @@ const DMNActionConfig = ({initialValues, onSave}) => {
       >
         {formik => (
           <Form>
-            <fieldset className="aligned">
+            <fieldset className="admin-fieldset aligned">
               <div className="form-row form-row--no-bottom-line">
                 <Field
                   name="pluginId"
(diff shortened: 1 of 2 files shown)


src/openforms/scss/components/builder/_builder.scss
```diff
@@ -166,6 +166,10 @@ div.flatpickr-calendar.open {
     display: block;
     margin-left: 0;
     margin-right: 0;
+
+    ul.errorlist {
+      padding-left: 0px;
```
SilviaAmAm, line 171: Doesn't seem to break the error styling in the registration tab.

Viicos (approved):

src/openforms/js/components/admin/form_design/logic/actions/dmn/DMNActionConfig.js
```diff
@@ -185,7 +185,7 @@ const DMNActionConfig = ({initialValues, onSave}) => {
       >
         {formik => (
           <Form>
-            <fieldset className="aligned">
+            <fieldset className="admin-fieldset aligned">
```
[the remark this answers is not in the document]
sergei-maertens: I'm curious whether using these components fixes the issue or not. Otherwise, I would compare the markup and styling from a non-JS enhanced admin page with validation error and see what the HTML/class names look like.

Since this is styling inside of a modal, I think we should be able to replicate the appearance without CSS overrides, but I'm not sure, there may be some top-level classes missing because the admin CSS is... not pleasant.

src/openforms/scss/components/builder/_builder.scss
sergei-maertens: does it break the styling of validation errors in the form slug field, for example?

SilviaAmAm: No, those ones have the class `module aligned` instead of `admin-fieldset`

PR #3944, opened 2024-02-27
Add patch release notes

sergei-maertens:
Ready to be merged when 2.2.10 is on Docker Hub.


docker/ci/config.json
```diff
@@ -1,5 +1,6 @@
 {
-    "supportedTags": [
+    "supportedTags":
+    [
```
sergei-maertens, line 3: Prettier update? I think :shrug:

PR #3948, opened 2024-02-27
Public API endpoints

### LINKED ISSUE #3825: Retrieving the available forms via the Formulieren API (/api/v2/forms) is slow
### Thema / Theme

API

We noticed in OIP 1768 that the /api/v2/forms endpoint is slower than our 5-second timeout, it is now 10-12 seconds on average.

We'll increase our timeout duration as a workaround and the OF testenvironment has quite a few complicated forms, but would it be possible to speed up this endpoint?

The integration in OIP uses Joeri's OF client: https://github.com/open-formulieren/open-forms-client-django 

image

_No response_

_No response_

SilviaAmAm:
Fixes #3825


src/openapi.yaml
```diff
@@ -3731,6 +3731,37 @@ paths:
...
+    get:
+      operationId: public_forms_list
+      description: List the available forms with minimal information. Does not include
+        soft deleted forms.
+      summary: List forms
+      parameters:
+      - in: query
+        name: category__name
+        schema:
+          type: string
```
sergei-maertens, line 3744: I would use category UUID, I think you can have duplicate category names now.

src/openforms/forms/api/public_api/tests/__init__.py
sergei-maertens, line 1: please keep the tests at the top-level of the app in `src/openforms/forms/tests/`. You can create a `public_api` subdirectory there.

src/openforms/forms/api/public_api/filters.py
```diff
@@ -0,0 +1,10 @@
+from django_filters import rest_framework as filters
+
+from openforms.forms.models import Form
```
sergei-maertens, line 3:
```suggestion
from ...models import Form
```
consistent with the views :)

src/openforms/forms/api/public_api/views.py
```diff
@@ -0,0 +1,29 @@
...
+    get=extend_schema(
+        summary=_("List forms"),
+        description=_(
+            "List the available forms with minimal information. Does not include soft deleted forms."
+        ),
+    ),
+)
+class FormListView(ListAPIView):
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
```
sergei-maertens, line 24: I don't think we should require a staff user - having an API client user that is able to log in to the admin is maybe a bit too much exposure.

Instead, I would:

* require the `view_form` permission (builtin django permissions)
* create a new group in the default_groups fixture, named "Public API client", and add the necessary permissions (view forms, view categories)

sergei-maertens (changes requested):

src/openforms/forms/tests/public_api/test_endpoints.py
```diff
@@ -0,0 +1,78 @@
...
+            reverse(
+                "api:public:forms:forms-list",
+            )
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_cant_access_without_view_form_permission(self):
+        user = UserFactory.create()
+        token = TokenFactory(user=user)
```
sergei-maertens, line 42:
```suggestion
        token = TokenFactory()
```
it will create a user automatically

sergei-maertens (approved):

PR #3951, opened 2024-02-27
Fix prefill plugin/attribute possibly being `null`

### LINKED ISSUE #3920: Cannot remove (simple) logic from components

2.5.x (new form builder)

Once set, the option to remove the simple logic triggers does not appear. and the form builder can only select True/False
image
or other components. 

Can only be removed by editing the JSON "conditional"

### LINKED ISSUE #3922: IntegrityError null value in column "prefill_plugin" of relation "forms_formvariable" violates not-null constraint

2.5.x (new form builder)

when copying a form that contains a textfield a 500 occurs. 
the cause seems to be:
```json
  "prefill": {
    "plugin": null,
    "attribute": null,
    "identifierRole": "main"
  },
```
which is included in new textfields. removing this part from the json fixes the issue in the form builder

**Tasks**

- [x] Do not allow `null` in TS types https://github.com/open-formulieren/types/pull/44
- [x] Fix formio builder to set empty strings as default instead of `null` open-formulieren/formio-builder#127
- [x] Fix backend with data migration to nuke the `null`s and change them into empty string

sergei-maertens:
Closes #3922
Closes #3920 (new formio-builder version contains necessary patches)

TODO:

- [x] Upgrade formio-builder patch version
- [x] Add (data) migration (only to main branch, the robustness fix should prevent the crashes)
- [x] Tests


sergei-maertens:
Only f7f7fc8d0055144f4eacb1a64a3311262dabe01f should be backported.

vaszig (approved):
Could you please explain what we mean with "Add (data) migration (only to main branch)"? I am referring to the 2nd task in your comment.

sergei-maertens:
> Could you please explain what we mean with "Add (data) migration (only to main branch)"? I am referring to the 2nd task in your comment.

It's this one: https://github.com/open-formulieren/open-forms/pull/3951/files#diff-0d93ae86b1472842077313544fca9c5351761e2c95f992863b0d17fcea586f08

It fixes existing data, but we will not backport this data migration because that causes too much trouble with managing the migration history. It's not essential for the (backported) fix either :)

sergei-maertens:
Backported to 2.5.x through 575e2fc44ad2aec72843ab23c01e782128e59213 and 2accdd377c684a4d1d7a37cfadbdb677c465432e

PR #3954, opened 2024-02-28
[#3688] Refactor Objects API plugin to allow for v2

Viicos:
Part of #3688.

# PR summary

This PR refactors the Objects API registration backend to split the registration logic depending on whether the registration options are v1 (legacy way, with the JSON templates) or v2 (with variables mapped individually):
- Added a `typing` module to provide TypedDicts for the options.
- Moved all the business logic in the `register_submission` method to the `submission_registration` module. A `Protocol` is defined, and an implementation for the V1 logic is provided (V2 is for another PR). This protocol is responsible for providing the object data JSON payload to be sent to the API. The `ObjectsAPIV1Handler` implements this protocol for V1, and all the utility functions that where in the plugin module are now set as staticmethods for clarity (i.e. to do some "namespacing").
- The registration plugin then instantiates the correct handler depending on V1 or V2, and simply grabs the record data to be sent.
- Some cleanups (the `get_today` function was moved to `openforms.utils.date`)

---

This also raises the question of how the record data will be generated for v2:
- we currently have control over the mapping of the variables, but v1 also allows registering more things (`bsn`, `uploaded_attachments_urls`, etc).
For reference, the default template looks like:
```jinja
{
  "data": {% json_summary %},
  "type": "{{ productaanvraag_type }}",
  "bsn": "{{ variables.auth_bsn }}",
  "pdf_url": "{{ submission.pdf_url }}",
  "attachments": {% uploaded_attachment_urls %},
  "submission_id": "{{ submission.kenmerk }}",
  "language_code": "{{ submission.language_code }}"
}
```
The idea would be to have the `variable_mapping` configuration translated to a JSON object that would fit in `"data"`. But the user has then no control over the rest, as he can't provide a JSON template anymore. In a follow up PR, this will probably be implemented in a way where registration backends can provide custom submission variables. For instance with the Objects API backend, a variable `pdf_url` of type `string` will be provided, and the end user will be able to map it to a specific JSON Schema target, the same way it is done currently for existing variables (so most probably a new tab in the variables tab will appear). This is still TBD.

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/plugin.py
+++ b/src/openforms/registrations/contrib/objects_api/plugin.py
@@ -5,42 +5,27 @@ from typing import Any
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from openforms.contrib.objects_api.helpers import prepare_data_for_registration
-from openforms.contrib.objects_api.rendering import render_to_json
-from openforms.contrib.zgw.clients import DocumentenClient
-from openforms.contrib.zgw.clients.utils import get_today
-from openforms.contrib.zgw.service import (
-    create_attachment_document,
-    create_csv_document,
-    create_report_document,
-)
+from typing_extensions import override
+
 from openforms.registrations.utils import execute_unless_result_exists
-from openforms.submissions.exports import create_submission_export
-from openforms.submissions.mapping import SKIP, FieldConf
-from openforms.submissions.models import Submission, SubmissionReport
-from openforms.variables.utils import get_variables_for_context
+from openforms.submissions.models import Submission
+from openforms.utils.date import get_today
 
 from ...base import BasePlugin
-from ...constants import RegistrationAttribute
 from ...registry import register
 from .checks import check_config
-from .client import get_documents_client, get_objects_client
+from .client import get_objects_client
 from .config import ObjectsAPIOptionsSerializer
 from .models import ObjectsAPIConfig
-from .utils import get_payment_context_data
+from .submission_registration import HANDLER_MAPPING
+from .typing import RegistrationOptions
 
 PLUGIN_IDENTIFIER = "objects_api"
 
 logger = logging.getLogger(__name__)
 
 
-def _point_coordinate(value):
-    if not value or not isinstance(value, list) or len(value) != 2:
-        return SKIP
-    return {"type": "Point", "coordinates": [value[0], value[1]]}
-
-
-def build_options(plugin_options: dict, key_mapping: dict) -> dict:
+def build_options(plugin_options: RegistrationOptions, key_mapping: dict) -> dict:
     """
     Construct options from plugin options dict, allowing renaming of keys
     """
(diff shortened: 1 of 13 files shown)


src/openforms/registrations/contrib/objects_api/typing.py
```diff
@@ -0,0 +1,34 @@
+from typing import Literal, TypeAlias, TypedDict
+
+from typing_extensions import Required
+
+
+class _BaseRegistrationOptions(TypedDict, total=False):
```
Viicos, line 6: Theoretically this is `total=False`, but every time the options are used, `apply_defaults_to` is used, effectively giving values for every key. So I'm wondering if it should be total?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -0,0 +1,115 @@
...
+from openforms.contrib.zgw.clients.utils import get_today
+from openforms.registrations.constants import REGISTRATION_ATTRIBUTE
+from openforms.submissions.mapping import MappingConfig, apply_data_mapping
+from openforms.submissions.models import Submission
+from openforms.variables.utils import get_variables_for_context
+
+from .typing import RegistrationOptions, RegistrationOptionsV1
+
+
+class BaseObjectsAPIRegistrator(ABC):
```
Viicos, line 18: The class hierarchy is a bit useless, but I didn't want to have 4 functions for both v1 and v2 & `get_record_data` and `get_update_payment_status_data`.

Considering no `__init__` method is defined, I might use a protocol instead.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -66,17 +66,20 @@ class ObjectsAPIRegistration(BasePlugin):
         ),
     }
 
+    @override
```
Viicos, line 69: Thoughts on the `override` decorator? When dealing with plugins I often get confused on which method is actually an implementation of the base class and which one isn't.

It's a bit useless in this case, as every method is an override :confused:

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -205,17 +212,12 @@ def update_payment_status(self, submission: "Submission", options: dict) -> None
             )
             return
 
-        context = {
-            "variables": get_variables_for_context(submission),
-            "payment": get_payment_context_data(submission),
-        }
+        registrator = REGISTRATOR_MAPPING[options["version"]]
+        updated_object_data = registrator.get_update_payment_status_data(submission)
```
Viicos, line 216: This is still TBD for v2. Currently, the following check is defined but only makes sense for v1:
https://github.com/open-formulieren/open-forms/blob/c3592e20869570d684827cf31897da26283637a9/src/openforms/registrations/contrib/objects_api/plugin.py#L209-L213

src/openforms/registrations/contrib/objects_api/plugin.py
SilviaAmAm: In my editor I generally see a little icon on the left of the method which indicates if this method overrides a method of a parent and if it is overwritten by a child (or both). It is very handy

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -658,8 +659,13 @@ def register_submission_to_objects_api(
...
         }
 
+        record_data = render_to_json(options["content_json"], context)
         object_data = prepare_data_for_registration(
-            submission, context, options, object_mapping
+            submission=submission,
+            record_data=record_data,
+            objecttype=options["objecttype"],
+            objecttype_version=options["objecttype_version"],
+            object_mapping=object_mapping,
```
sergei-maertens, line 668: I seem to remember that this feature was a temporary solution for Utrecht. Does that mean we don't need to provide the new mechanism for object creation in the ZGW registration @joeribekker ?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -0,0 +1,115 @@
...
+
+    object_data = {
+        "type": objecttype,
+        "record": {
+            "typeVersion": objecttype_version,
+            "data": record_data,
+            "startAt": get_today(),
+        },
+    }
+    apply_data_mapping(submission, object_mapping, REGISTRATION_ATTRIBUTE, object_data)
```
sergei-maertens, line 113: this will only remain relevant for V1, in V2 we will have an explicit mapping.

We will have to figure out a migration path from the component-level "Registration" tab, but perhaps pointing to the "Variables" tab and configuration will be sufficient already.

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -0,0 +1,115 @@
...
+    ) -> dict[str, Any]:
+        context = {
+            "variables": get_variables_for_context(submission),
+            "payment": self.get_payment_context_data(submission),
+        }
+
+        return render_to_json(options["payment_status_update_json"], context)
+
+
+REGISTRATOR_MAPPING: dict[int, BaseObjectsAPIRegistrator] = {
```
sergei-maertens, line 91: could be more strictly typed:

```py
ConfigVersion: TypeAlias = Literal[1, 2]

# and then
dict[ConfigVersion, ...]
```

sergei-maertens (dismissed):

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -33,6 +33,7 @@
 from ...exceptions import RegistrationFailed
 from ...registry import register
 from ...utils import execute_unless_result_exists
+from ..objects_api.submission_registration import prepare_data_for_registration
```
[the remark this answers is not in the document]
Viicos: Hum, this would then to be applied to some zgw stuff imported in the objects api plugin module as well:
https://github.com/open-formulieren/open-forms/blob/7b31de3ff4bdb85aefc94d06280d1f69949b8ede/src/openforms/registrations/contrib/objects_api/plugin.py#L10-L16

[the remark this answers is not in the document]
sergei-maertens: no, because it's using the shared functionality in `openforms.contrib.zgw`, not using private API from `openforms.registrations.contrib.zgw_apis` :)

src/openforms/registrations/contrib/objects_api/typing.py
Viicos: Making typeddicts partial/required is a pain currently, I wasn't able to make it work without redefining all the fields in a separate typeddict. It's not a big deal so I'll leave it as is for now and we could revisit once partial/required eventually lands :)

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -0,0 +1,233 @@
...
+            client=documents_client,
+            name=f"{submission.form.admin_name} (csv)",
+            csv_data=submission_csv,
+            options=submission_csv_options,
+            language=submission.language_code,
+        )
+
+        return submission_csv_document["url"]
+
+    @override
```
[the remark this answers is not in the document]
Viicos: I added a comment [here](https://github.com/open-formulieren/open-forms/pull/3954#discussion_r1505994269) related to this. This decorator was first introduced for type checkers, with the canonical example being:

```python
class Base:
    def base_method(self): ...

class Sub(Base):
    def base_method(self): ...
```

In this scenario, you could change `Base.base_method` to be `Base.new_method`, and type checkers wouldn't be aware of such a change (they would see `Sub.base_method` as a new method not present on `Base`). By adding the `@override` decorator on `Sub.base_method`, it will require the method to exist in `Base`, essentially catching the issue I described.

Because we don't run type checking, it doesn't really change anything in our case, but I find it easier to see which method is tied to the plugin ABC, especially with our plugin architecture.

> Are we going to keep it and use it generally?

I guess it's personal preference

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -0,0 +1,233 @@
...
+    def get_object_data(
+        self,
+        submission: Submission,
+        options: RegistrationOptionsV1,
+    ) -> dict[str, Any]:
+        """Get the object data payload to be sent to the Objects API.
```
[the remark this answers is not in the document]
Viicos: I updated the docstring of `register_submission` :+1:

PR #3955, opened 2024-02-28
[#851] Clean up DocumentenClient

### LINKED ISSUE #851: Investigate & clean up code duplication in creating Documenten API documents
src/openforms/registrations/contrib/zgw_apis/service.py

functions:

* `create_document`
* `create_report_document`
* `create_attachment_document`
* `create_csv_document`

vaszig:
Fixes #851

### DIFF
--- a/src/openforms/contrib/zgw/service.py
+++ b/src/openforms/contrib/zgw/service.py
@@ -1,12 +1,14 @@
 import logging
 from io import BytesIO
+from typing import Literal, TypeAlias
 
 from openforms.submissions.models import SubmissionFileAttachment, SubmissionReport
-from openforms.translations.utils import to_iso639_2b
 
 from .clients.documenten import DocumentenClient
 
 logger = logging.getLogger(__name__)
+
+SupportedLanguage: TypeAlias = Literal["nl", "en"]
 
 
 def create_report_document(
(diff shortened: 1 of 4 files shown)


src/openforms/contrib/zgw/service.py
```diff
@@ -41,7 +40,7 @@ def create_csv_document(
     name: str,
     csv_data: str,
     options: dict,
-    language: str = "nld",
+    language: str,
```
sergei-maertens, line 43: this should use the same type alias (and so the type alias should probably be moved from `documenten.py` into `service.py`), otherwise we lose the benefits of strict types again.

sergei-maertens (changes requested):

src/openforms/contrib/zgw/service.py
vaszig: So we mean that each function (`create_report_document`, `create_csv_document`, `create_attachment_document`) should have an explicit argument `language` right?In order to define the type in the `service.py` file.

sergei-maertens: Yes, because otherwise you get type check errors - `str` cannot be assigned to the literal :)

sergei-maertens (approved):

PR #3956, opened 2024-02-28
[#3825] Public categories endpoint

### LINKED ISSUE #3825: Retrieving the available forms via the Formulieren API (/api/v2/forms) is slow

SilviaAmAm:
Fixes #3825

### DIFF
--- a/src/openforms/forms/api/viewsets.py
+++ b/src/openforms/forms/api/viewsets.py
@@ -12,7 +12,7 @@ from django.utils.translation import gettext_lazy as _
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
 from rest_framework import parsers, permissions, response, status, views, viewsets
-from rest_framework.authentication import TokenAuthentication
+from rest_framework.authentication import SessionAuthentication, TokenAuthentication
 from rest_framework.decorators import action
 from rest_framework.mixins import ListModelMixin
 from rest_framework.request import Request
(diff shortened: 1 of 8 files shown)


sergei-maertens:
> I feel that maybe this endpoint should also move to the new public folder 

Yep, let's do it if it hasn't been done yet

src/openforms/forms/api/viewsets.py
```diff
@@ -817,6 +817,7 @@ class CategoryViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Category.objects.all()
     serializer_class = CategorySerializer
     # TODO this permission isn't ideal with its mix of public and write access
+    authentication_classes = (TokenAuthentication, SessionAuthentication)
     permission_classes = [FormAPIPermissions]
```
sergei-maertens, line 821: I think a separate permission class for this that checks if you have explicit `view_category` or `FormAPIPermissions` seems appropriate

I think in DRF these days you can do `Permission1 | Permission2` to do an OR

SilviaAmAm: But wouldn't this be problematic for the POST actions?

sergei-maertens: it's a `ReadOnlyModelViewSet` :)

but it's a good idea to add some tests that you get an HTTP 405 on write actions through the API, yes.

SilviaAmAm: oh oups, silly me :see_no_evil:

sergei-maertens (approved):

PR #3960, opened 2024-02-29
[#3718] Drop FormDefinition.component_translations JSONField

### LINKED ISSUE #3718: Drop FormDefinition.component_translations JSONField
Depends on #2958 being completed and the new formio builder being **the** builder, using the new translations data structure.

The `FormDefinition` model field `component_translations` is the legacy format, and during release 2.5.x we're still keeping it around in case things go wrong so that we can recover from bad situations, but once we're sufficiently confident that the new translation format is stable, we can drop this column as it's unused.

There is more translation related code that can then be cleaned up, dealing with the old format.

**Tasks**

* [x] Drop model field
* [x] Remove tests and application code that handled the old translation format

vaszig:
Closes #3718


src/openforms/js/components/formio_builder/translation.js
```diff
@@ -52,25 +52,45 @@ export const extractComponentLiterals = component => {
...
     // Do not show warnings for date/datetime placeholders, they are not translatable:
     .filter(property => (isDateOrDatetimeComponent ? property !== 'placeholder' : property))
     .filter(property => !isRegex(property))
-    .map(property => get(component, property));
+    .map(property => {
+      return {
+        property: property,
+        literal: get(component, property),
+      };
+    });
```
Viicos, line 64:
```suggestion
    .map(property => ({property, literal: get(component, property)}));
```

Could be also applied for the two below?

src/openforms/js/components/formio_builder/translation.spec.js
```diff
@@ -21,8 +21,8 @@ describe('Checking date/datetime translation literals', () => {
       tooltip: '',
       format: 'dd-MM-yyyy HH:mm',
     };
-
-    expect(extractComponentLiterals(dateComponent)).toStrictEqual(['Date']);
-    expect(extractComponentLiterals(dateTimeComponent)).toStrictEqual(['Date / Time']);
+    console.log(extractComponentLiterals(dateComponent));
```
Viicos, line 24:
```suggestion
```

Or was it on purpose?

Viicos (approved):

src/openforms/js/components/formio_builder/translation.spec.js
vaszig: By mistake!It will be removed

src/openforms/js/components/admin/form_design/FormStepDefinition.js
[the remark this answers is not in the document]
vaszig: Probably I don't understand your point..I mean `component_translations` JSONField had the translations for all the components (literals) and it was part of the form definition model.

src/openforms/js/components/formio_builder/translation.js
vaszig: I didn't do that there since it is specifically for the label.

Viicos: My suggestion does not introduce any changes, it just makes the syntax shorter

vaszig: aaah I thought you meant using the same approach but you mean only removing the `return` etc.

src/openforms/js/components/formio_builder/translation.js
```diff
@@ -56,37 +56,22 @@ export const extractComponentLiterals = component => {
...
-      };
-    });
+    .map(property => ({property, literal: get(component, property)}));
 
   // See ADDITIONAL_PROPERTIES_BY_COMPONENT_TYPE for the reverse check
   switch (component.type) {
     case 'select': {
       const values = component?.data?.values || [];
       finalTranslatableProperties.push(
-        ...values.map(v => {
```
vaszig: I also changed the name from `label` to `values_label` because I found out that there were multiple keys with the same name and the one was overriding the other.

src/openforms/formio/rendering/default.py
```diff
@@ -311,7 +311,6 @@ def get_children(self) -> Iterator["ComponentNode"]:
 
     def __post_init__(self):
         self._base_label = self.component.get("groupLabel", self.default_label)
-        super().__post_init__()
```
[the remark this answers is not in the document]
vaszig: Yes,I removed the `__post_init__` from the `ComponentNode` so that's why it was removed here as well.
[https://github.com/open-formulieren/open-forms/pull/3960/files/24efb00ce3b30699dc4529c23e14eb34f5ed673e#diff-5fa3e1ab66d508ceaa4cb2a42ce27a9339cf2f4ad82e038fd6fb4272c52e318fL87](url)

src/openforms/formio/variables.py
```diff
@@ -89,19 +87,17 @@ def inject_variables(
...
                 case [str(), *_]:
-                    property_value = [
-                        translate(s) for s in property_value if isinstance(s, str)
-                    ]
+                    property_value = [s for s in property_value if isinstance(s, str)]
                 case [{"label": _}, *_]:
                     for item in property_value:
                         if "label" in item:
-                            item["label"] = translate(item["label"])
+                            item["label"] = item["label"]
```
sergei-maertens, line 96: this is also a noop and can/should be removed entirely, no?

src/openforms/formio/variables.py
```diff
@@ -89,19 +87,17 @@ def inject_variables(
...
                 case {"values": [*defined_values]}:
                     for item in defined_values:
                         if "label" in item:
-                            item["label"] = translate(item["label"])
+                            item["label"] = item["label"]
```
sergei-maertens, line 100: another noop that can be removed I think :thinking:

src/openforms/forms/tests/test_import_export.py
```diff
@@ -988,73 +988,3 @@ def test_rountrip_form_with_theme_override(self):
 
         imported_form = Form.objects.exclude(pk=form.pk).get()
         self.assertIsNone(imported_form.theme)
-
-    def test_import_form_with_legacy_formio_component_translations_format(self):
```
sergei-maertens, line 992: please revert the removal of this test

src/openforms/forms/tests/test_translations_converter.py
sergei-maertens, line 1: please revert

sergei-maertens (changes requested):

src/openforms/forms/api/serializers/form_definition.py
```diff
@@ -161,6 +161,14 @@ def validate(self, attrs):
 
         return attrs
 
+    def save(self, **kwargs):
+        # field 'component_translations' has been deprecated and it's not part of
+        # the 'form_definition' model anymore. We use it only in the serializer.
+        if "component_translations" in self.validated_data:
+            del self.validated_data["component_translations"]
+
+        return super().save(**kwargs)
```
sergei-maertens, line 170: as this probably only applies to the import process, I'd move this to `openforms.forms.utils.import_form_data` and drop the data right before `instance = deserialized.save()` is called. There already is some data pre-processing happening in there :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3963, opened 2024-03-01
:memo: [#3838] Document Public API

### LINKED ISSUE #3838: Properly document and communicate that (most of) our API endpoints are private API
- [x] Properly designate the few public endpoints (form list, categories)
- [x] Make stripped down versions?
- [x] Update documentation about release and versioning policy
- [x] Remove API spec links from docs (instead, put them in the RTD docs?)

SilviaAmAm:
Fixes #3838

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,52 +46,11 @@ Component
 
 This component includes the **Open Forms Admin UI** and the **Open Forms API**.
 
-API specification
------------------
+API Specifications
+------------------
 
-==============  ==============  =============================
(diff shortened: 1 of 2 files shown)


Viicos (approved):
I'm wondering if we could leave at least the latest redoc/swagger spec link in the readme, as sometimes I find it convenient to just go in the readme and open the spec from there.

Or maybe a simple link from the readme pointing to the section in the docs?

PR #3970, opened 2024-03-04
Fix/new logo e idas

nikkiysendoorn1:


nikkiysendoorn1:
Closes #3957

SilviaAmAm (approved):

Viicos (approved):
Was there any issue with the svg logo?

nikkiysendoorn1:
@Viicos Yes, the SVG only worked on the mobile device, because there was missing some styling. So I can also add the SVG and then add some styling? What do you prefer?

Viicos:
> @Viicos Yes, the SVG only worked on the mobile device, because there was missing some styling. So I can also add the SVG and then add some styling? What do you prefer?

Was just wondering what was the issue with the svg; I think it's fine to have a png for such a small icon (and svgs can be tricky with styling so I don't think it's worth spending time on it)

PR #3973, opened 2024-03-04
🗑️ Update deprecated unittest methods

Viicos:
Will fail in 3.12


vaszig:
Should we run again the CI?Does it have to do with flakiness or is it something else?Otherwise I guess it can be approved and merged.

Viicos:
> Should we run again the CI?Does it have to do with flakiness or is it something else?Otherwise I guess it can be approved and merged.

Yes this is flakyness, it seems to be happening quite a lot these days :/

SilviaAmAm (approved):

PR #3979, opened 2024-03-06
[#3975] Service fetch errors

### LINKED ISSUE #3052: Remove handling of old cases from ServiceFetchAction.eval
Previously in the 2.1.x release of the Service Fetch the admin UI of the ServiceFetchAction logic action, a ServiceFetchConfiguration was not tied to a User Defined Variable and the action allowed to change the values of both variables and components directly. This has changed with the configuration modal.

`ServiceFetchAction.eval()`  still contains a shim that handles the old case. This should be tackled with a datamigration and/or handled in the serializer/import step.

This change in behaviour should also be documented #2884 , both in the manual and in test cases.

### LINKED ISSUE #3975: Updating service fetch configuration in the form editor causes problems

2.4.2

Observed on Den Haag ACC environment.

The service fetch action had the following JSON:
```json
{
  "component": "",
  "variable": "result",
  "formStep": "",
  "formStepUuid": null,
  "action": {
    "type": "fetch-from-service",
    "value": "1"
  }
}
```
And updating the service in the modal changed it to:
```json
{
  "component": "",
  "variable": "result",
  "formStep": "",
  "formStepUuid": null,
  "action": {
    "type": "fetch-from-service",
    "value": "1"
  },
  "fetchConfiguration": {
    "existing": "9"
  }
}
```

But then upon saving the form, the initial JSON was not updated.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SilviaAmAm:
Fixes #3975
Fixes #3052

**Context**

**The symptoms**
In the DH environment, we saw that a form with service fetch would give 500 error on starting. The error was "ServiceFetchConfiguration does not exist".
The logic action trying to perform the service fetch had this JSON:
```json
{
  "component": "",
  "variable": "result",
  "formStep": "",
  "formStepUuid": null,
  "action": {
    "type": "fetch-from-service",
    "value": "1"
  }
}
```
There was no `ServiceFetchConfiguration` with `pk=1`, but trying to update the configuration through the logic tab in the admin was not removing the `"value": "1"`.

**The problem**
Before version 2.2, the configuration of service fetch was set in the action (i.e. this is what the `"value": "1"` was here). Since version 2.2, the service fetch configuration was specified on the variable (in this case, in variable `"result"`). So, when updating the service fetch configuration in the admin, the `"value": "1"` was no longer changed.

However, in the backend, the evaluation of the fetch action had this code:
```python
       if self.fetch_config:  # Taken from the value parameter in the action
            var = FormVariable(
                key=self.variable,
                service_fetch_configuration=ServiceFetchConfiguration.objects.get(
                    pk=self.fetch_config
                ),
            )
        else:  
            var = self.rule.form.formvariable_set.get(key=self.variable)
```
So, for old forms that still had this value, it always tried to use the old configuration.

**The solution**
There are multiple parts to this. 
1. Remove the backend code that still supported the 'old' behaviour, since the frontend doesn't support this behaviour anymore.
2. Add a data migration so that forms that still have this parameter get cleaned up
3. Add a step to clear this parameter for the import form functionality.

**Note**
I think that for backporting this, only point 1. of the solution is really needed? (since backporting migrations is a bit painful).
However, we currently support only 2.3 onwards, so 2.2.x versions will still have the broken behaviour (but I guess it doesn't matter since they are not supported?)

Also: there is a work around. Deleting the action in the rule and re-creating it.


src/openforms/forms/migrations/0094_convert_old_service_fetch_config.py
```diff
@@ -0,0 +1,86 @@
...
+                continue
+
+            variable.service_fetch_configuration = service_fetch_config
+            updated_variables.append(variable)
+            rule_updated = True
+
+        if rule_updated:
+            updated_rules.append(rule)
+
+    return UpdatedObjects(rules=updated_rules, variables=updated_variables)
```
Viicos, line 54: I think it's just personal preference, but a two-tuple `tuple[list[FormLogic], list[FormVariable]]` could avoid defining a typed dict.

The drawback is it's probably less readable. If we keep the typed dict however, the lists types can probably be parametrized with `list[FormLogic]` and `list[FormVariable]`

src/openforms/forms/migrations/0094_convert_old_service_fetch_config.py
```diff
@@ -0,0 +1,86 @@
...
+    FormVariable.objects.bulk_update(
+        updated_objects["variables"], fields=["service_fetch_configuration"]
+    )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("forms", "0093_fix_prefill_bis"),
+    ]
```
Viicos, line 79: I think it's best to depend on all latest migrations related to the models being used ([docs](https://docs.djangoproject.com/en/dev/topics/migrations/#accessing-models-from-other-apps)), as we are using `ServiceFetchConfiguration` from `variables` for instance

src/openforms/forms/tests/test_migrations.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        FormLogic.objects.create(
+            order=0,
+            form=form,
+            json_logic_trigger={"!!": [True]},
+            actions=[
+                {
+                    "action": {
+                        "type": "fetch-from-service",
+                        "value": str(config.pk),  # An existing service
```
Viicos, line 54: As per my previous comment, maybe a test with an integer value as well

docs/developers/backend/service-fetch.rst
```diff
@@ -0,0 +1,36 @@
...
+In the JSON definition of the action, there is a ``variable`` parameter which points to a user-defined variable.
+The corresponding variable contains an attribute ``service_fetch_configuration`` which contains a reference to
+a :class:`ServiceFetchConfiguration`. This configuration specifies:
+
+- the :class:`Service` that will be used to perform the request
+- the endpoint that will be queried
+- any query parameters
+- any expression to apply to the response
+
+While the form is being filled in by a user, the method :meth:`openforms.submissions.form_logic.evaluate_form_logic`
```
Viicos, line 23:
```suggestion
While the form is being filled in by a user, the :meth:`openforms.submissions.form_logic.evaluate_form_logic` method
```

:shrug:

docs/developers/backend/service-fetch.rst
```diff
@@ -0,0 +1,36 @@
...
+performs the request specified in the logic rule action if the trigger evaluates to ``True``.
+
+The data in the response from the external service is processed with any expression (jq or json-logic) specified in the
```
Viicos, line 26:
```suggestion
The data in the response from the external service is processed with any expression (``jq`` or JsonLogic) specified in the
```
fine to leave it as is as well

src/openforms/submissions/tests/form_logic/test_endpoint.py
```diff
@@ -355,3 +359,58 @@ def test_handling_none_values_in_logic(self):
...
+        )
+        FormLogicFactory.create(
+            form=form,
+            json_logic_trigger={"!!": [True]},
+            actions=[
+                {
+                    "action": {
+                        "type": "fetch-from-service",
+                        "value": "4000",
+                    },  # A non-existing service
```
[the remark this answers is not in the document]
Viicos: Nevermind, was on an older commit

src/openforms/forms/migrations/0094_convert_old_service_fetch_config.py
```diff
@@ -0,0 +1,86 @@
...
+) -> UpdatedObjects:
+    updated_rules = []
+    updated_variables = []
+    for rule in rules:
+        rule_updated = False
+        for action in rule.actions:
+            if action["action"]["type"] != LogicActionTypes.fetch_from_service:
+                continue
+
+            if (service_fetch_config_pk := action["action"]["value"]) == "":
```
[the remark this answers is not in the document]
SilviaAmAm: Mhmm but the `ServiceFetchAction` `fetch_config` is getting removed :see_no_evil:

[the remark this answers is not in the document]
Viicos: Yes what I meant is you are comparing `action["action"]["value"]` to `""`, and we can see in the `from_action` constructor that `action["action"]["value"]` is also used to instantiate the `fetch_config` attribute, typed as `int`. So I'm wondering if it's possible for the value to be an `int`, and in that case maybe something like:

`if (service_fetch_config_pk := str(action["action"]["value"])) == "":`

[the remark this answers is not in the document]
SilviaAmAm: Ahh I see what you mean now!

sergei-maertens:
Backports:

* 2.5.x: 46886981ade45ddb25a39ecbb99ac3a0729fb448
* 2.4.x: c2d414f1e47eb6b57cd9f0338863332f30ffb0fa
* 2.3.x: cda862b9cc547087a4bbc38f76b9581d52483409

PR #3982, opened 2024-03-06
[#3688] Add base logic to register submission with v2 options

Viicos:
Part of #3688.

frontend part:
- https://github.com/open-formulieren/open-forms/pull/3988

Reminder on the current flow:

When a submission gets registered in the Objects API plugin:
- `ObjectsAPIRegistration.register_submission` is called
  - Depending on the version (v1 or v2) of the options, a specific handler is instantiated
  - `handler.save_registration_data` is called (same method for v1 and v2). This creates the documents in the Documents API, and save the urls in a temporary `ObjectsAPIRegistrationData` model.
  - `handler.get_object_data` is called to get the payload sent to the Objects API. The v1 renders the JSON template, the v2 creates the object from the mapping (this is what this PR implements)
  - Payload is sent to the Objects API

### DIFF
--- a/docker/docker-compose.objects-apis.yml
+++ b/docker/docker-compose.objects-apis.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.8'
 
 name: objects-apis
 
--- a/docker/docker-compose.objects-apis.yml
+++ b/docker/docker-compose.objects-apis.yml
@@ -1,11 +1,15 @@
 version: '3'
 
+name: objects-apis
+
 services:
-  redis:
+  objects-redis:
     image: redis:7
     command: ["redis-server", "--appendonly", "yes"]
+    networks:
+      - open-forms-dev
 
-  db:
+  objects-objecttypes-db:
     image: postgis/postgis:${PG_VERSION:-14-master}
     environment:
       - POSTGRES_HOST_AUTH_METHOD=trust
@@ -21,10 +25,11 @@ services:
       - DJANGO_SETTINGS_MODULE=objecttypes.conf.docker
       - SECRET_KEY=${SECRET_KEY:-fgv=c0hz&tl*8*3m3893@m+1pstrvidc9e^5@fpspmg%cy$15d}
       - ALLOWED_HOSTS=*
+      - DB_HOST=objects-objecttypes-db
     ports:
       - 8001:8000
     depends_on:
-      - db
+      - objects-objecttypes-db
     networks:
       - open-forms-dev
 
@@ -34,22 +39,26 @@ services:
       - DJANGO_SETINGS_MODULE=objects.conf.docker
       - SECRET_KEY=${SECRET_KEY:-fgv=c0hz&tl*8*3m3893@m+1pstrvidc9e^5@fpspmg%cy$15d}
       - ALLOWED_HOSTS=*
+      - DB_HOST=objects-objecttypes-db
       - DISABLE_2FA=True
+      - CELERY_BROKER_URL=redis://objects-redis:6379/0
+      - CELERY_RESULT_BACKEND=redis://objects-redis:6379/0
+      - NOTIFICATIONS_DISABLED=True
     ports:
       - 8002:8000
     depends_on:
-      - db
-      - redis
+      - objects-objecttypes-db
+      - objects-redis
     networks:
       - open-forms-dev
 
-  celery:
+  objects-celery:
     image: maykinmedia/objects-api:${OBJECTS_VERSION:-2.3.0}
     environment: *objects_web_env
     command: /celery_worker.sh
     depends_on:
-      - db
-      - redis
+      - objects-objecttypes-db
+      - objects-redis
     networks:
       - open-forms-dev
 
--- a/src/openforms/registrations/contrib/objects_api/submission_registration.py
+++ b/src/openforms/registrations/contrib/objects_api/submission_registration.py
@@ -1,7 +1,10 @@
+import logging
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
+from datetime import date, datetime
 from typing import Any, Generic, Iterator, TypeVar, cast
 
+import glom
 from typing_extensions import override
 
 from openforms.contrib.objects_api.helpers import prepare_data_for_registration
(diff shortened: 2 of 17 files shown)


src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -228,6 +241,51 @@ def get_update_payment_status_data(
...
+        for mapping in variables_mapping:
+            variable_key = mapping["variable_key"]
+            target_path = mapping["target_path"]
+
+            # Type hint is wrong: currently some static variables are of type date/datetime
+            value = cast(Any, submission_variables[variable_key])
+
+            # Comply with JSON Schema "format" specs:
+            if isinstance(value, (datetime, date)):
+                value = value.isoformat()
```
Viicos, line 263: This is a bit surprising: some static variables return `date` or `datetime` objects, while all the type annotations related to this were saying it was JSON data (e.g. here `submission_variables` is typed as `JSONValue`). I think this relates again to proper type checking of variables/JSON Logic

sergei-maertens: That's why we created #2324 :(

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -228,6 +241,51 @@ def get_update_payment_status_data(
         return render_to_json(options["payment_status_update_json"], context)
 
 
+class ObjectsAPIV2Handler(ObjectsAPIRegistrationHandler[RegistrationOptionsV2]):
+
+    @override
+    def get_object_data(
+        self, submission: Submission, options: RegistrationOptionsV2
+    ) -> dict[str, Any]:
+        submission_variables = get_variables_for_context(submission)
```
sergei-maertens, line 250: as remarked in the office - this is intended to be used with the Django template engine which is explicitly what we don't want to do here.

A more correct, low-level and consistent approach is to use:

```py
submission.load_submission_value_variables_state()
```
see also `openforms.submissions.logic.datastructures.DataContainer` which works with this, in particular there's the `to_python` method on variables that may need to be used to normalize data into the python domain.

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -228,6 +241,51 @@ def get_update_payment_status_data(
...
+        variables_mapping = options["variables_mapping"]
+        record_data: dict[str, Any] = {}
```
sergei-maertens, line 252: this should effectively be `JSONObject`, no? Instead of the `Any` values.

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -1253,3 +1257,127 @@ def test_submission_with_payment(self, m):
...
+                },
+            ],
+            "geometry_variable_key": "myMapField",
+        }
+
+        m.post(
+            "https://objecten.nl/api/v1/objects",
+            status_code=201,
+            json=get_create_json,
+        )
```
sergei-maertens, line 1330: can we please use docker-compose + VCR instead of these manual mocks?

I'd also like to see the tests for v1 and v2 in separate modules, there's nearly 1400 lines in this file and I can't put them side by side to compare if the same scope of functionality is covered

sergei-maertens (changes requested):

docker/docker-compose.objects-apis.yml
```diff
@@ -1,11 +1,15 @@
 version: '3'
```
SilviaAmAm, line 1: I get the warning that `version` parameter is obsolete, but running with the old `docker-compose` complains about the `name`.

SilviaAmAm (approved):

docker/docker-compose.objects-apis.yml
sergei-maertens: I'd use `version: '3.8'`, 3 is very old and equivalent to 3.0

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -292,7 +307,14 @@ def get_object_data(
     @override
     def get_update_payment_status_data(
         self, submission: Submission, options: RegistrationOptionsV1
-    ) -> dict[str, Any]:
+    ) -> dict[str, Any] | None:
+
+        if not options["payment_status_update_json"]:
```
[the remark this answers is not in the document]
Viicos: This was discussed here: https://github.com/open-formulieren/open-forms/pull/3954#discussion_r1505989404.

Defaults are applied from the default configuration but type hints currently does not (and can't) reflect this

Related issue: https://github.com/open-formulieren/open-forms/issues/4024

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -301,6 +323,70 @@ def get_update_payment_status_data(
...
+        static_values = state.static_data()
+        static_values.update(
+            {
+                variable.key: variable.initial_value
+                for variable in get_static_variables(
+                    submission=submission,
+                    variables_registry=variables_registry,
+                )
+            }
+        )
```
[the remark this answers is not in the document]
Viicos: > the static data does not include the registration static variables

correct and they shouldn't (as described [here](https://github.com/open-formulieren/open-forms/pull/3984#issue-2174280226))

> I wonder if this could be done by a method of the state 🤔

I first updated the `sate.static_data` method to take a registry as an argument, so that it could return the appropriate variables, but the static data is saved to a private attribute whenever this method gets called.

This adds a bit of duplication, so maybe in the future if other registration backends needs to do the same it can be revisited.

docker/docker-compose.objects-apis.yml
Viicos: `docker-compose` is probably too old to support the `name` attribute, but I guess `docker compose` should be used anyway.

Regarding the `version` attribute, it seems to be there for [backwards compatibility](https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-optional). The compose format is now generic and not tied to docker anymore apparently. Maybe in the future we should update all our docker compose files to remove this version, but we would have to make sure everyone is on the latest docker engine version.

sergei-maertens: my experience is that with docker-compose versions you need to be very conservative. I remember using a version that was released more than 1+ year ago and had to revert to an older one because people did not have those runtimes on their machines yet and didn't have admin rights to update.

PR #3984, opened 2024-03-07
Add logic for the new registration (static) variables

Viicos:
After a lot of thinking about the technical implementation:
- Having all registration static variables registered under the same registry is not reliable, as I expect registration plugins to provide similar (if not identical) static variables, and duplicate identifiers would fail.
- To overcome this issue, each plugin can provide their own registry. 
- `get_static_variables` was adapted so that it can iterate over a provided registry, instead of the default one for vanilla static variables.
- A bit unrelated but can be reviewed in separate commits: an endpoint returning the available registration variables was added. It is designed mainly as a backend for frontend endpoint. 

This preserves backward compatibility, as `get_static_variables([, submission=...])` will still use the default registry. It also works well this way because `get_static_variables` is currently used in:
- jsonlogic trigger validation: ✅
https://github.com/open-formulieren/open-forms/blob/bed811838f8dabbf003a1c3c0827f9c14cfaf2ec/src/openforms/forms/api/validators.py#L126-L128
registration static variables shouldn't be referenced.
- Validate form variables are unique: :warning: 
https://github.com/open-formulieren/open-forms/blob/bed811838f8dabbf003a1c3c0827f9c14cfaf2ec/src/openforms/forms/api/serializers/form_variable.py#L54-L55
Ideally registration variables should be taken into account as well. Not sure how common a name collision between a form variable and a registration variable would be. If we want to validate this, it might get tricky as every registration backend should be taken into account. So perhaps a new `include_registration` boolean option to `get_static_variables` could be added? (Issue is it is conflicting with the `variables_registry` argument).
- `get_variables_for_context`: ✅
https://github.com/open-formulieren/open-forms/blob/bed811838f8dabbf003a1c3c0827f9c14cfaf2ec/src/openforms/variables/utils.py#L52-L56
`get_variables_for_context` is only used for the Objects API (where it makes sense to not include reg. variables, those are added manually in the context by accessing `ObjectsAPIRegistrationData`) and the email backend (not sure about this one, probably doesn't make sense to have registration variables for this plugin anyway?). 
- static variables endpoint: ✅
https://github.com/open-formulieren/open-forms/blob/bed811838f8dabbf003a1c3c0827f9c14cfaf2ec/src/openforms/variables/api/views.py#L24-L30
As expected, reg. variables are in a separate endpoint.

---

For tests, ~~I'm planning on testing~~ I tested the `ObjectsAPIRegistrationData` creation, in particular how it behaves when I get an API error when trying to create documents. Not sure where to put these, ideally I don't want to touch the big 1k lines `ObjectsAPIBackendTests` test class, although it might require some refactor?

Added registration variables will be tested in https://github.com/open-formulieren/open-forms/pull/3982, when the v2 registration will be tested.


src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -195,3 +196,38 @@ def apply_defaults_to(self, options):
...
+    )
+    attachment_urls = ArrayField(
+        models.URLField(
+            _("attachment url"), help_text=_("The attachment URL on the Documents API.")
+        ),
+        verbose_name=_("attachment urls"),
+        help_text=_("The list of attachment URLs on the Documents API."),
+        blank=True,
+        default=list,
+    )
```
sergei-maertens, line 233: I think this is too naive, you lose the information of which file/component the Document URL belongs too.

I'd still prefer to have the variable of a file upload explicitly mapped into the schema, e.g. something like:

```yaml
type: object
properties:
  proofOfEmployment:
    type: string
    format: uri
  bankStatements:
    type: array
    items:
        type: string
        format: uri
```
so that you can make the (single) file upload "proof of employment" to `/proofOfEmployment` and the multi-file upload "bank statements" to `/bankStatements`

Possibly we do this in a separate PR

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -0,0 +1,94 @@
...
+class PdfUrl(BaseStaticVariable):
+    name = _("PDF Url")
+    data_type = FormVariableDataTypes.string
+
+    def get_initial_value(self, submission: Submission | None = None):
+        return (
+            submission.objects_api_registration_data.pdf_url
+            if submission is not None
+            else None
+        )
```
sergei-maertens, line 37: I much prefer guard clauses:

```suggestion
    def get_initial_value(self, submission: Submission | None = None):
        if submission is None:
             return None
        return submission.objects_api_registration_data.pdf_url
```
it makes it a lot clearer that the submission may be missing.

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -9,34 +11,216 @@
...
+    """
+    options = {
+        new_key: plugin_options[key_in_opts]
+        for new_key, key_in_opts in key_mapping.items()
+        if key_in_opts in plugin_options
+    }
+    return options
+
+
+def registration_submission_pdf(
```
sergei-maertens, line 45: can you make the function name consistent? `registration_submission_pdf` vs. `register_submission_csv` - please use the same verb :)

src/openforms/variables/api/registration_serializer.py
```diff
@@ -0,0 +1,23 @@
...
+    )
+    plugin_verbose_name = serializers.CharField(
+        label=_("backend verbose name"),
+        help_text=_("The verbose name of the registration plugin."),
+    )
+    plugin_variables = serializers.ListSerializer(
+        child=FormVariableSerializer(),
+        label=_("variables"),
+        help_text=_("The list of corresponding registration variables."),
+    )
```
sergei-maertens, line 23:
```suggestion
    plugin_variables = FormVariableSerializer(many=True, label=..., help_text=...),
```
is the idiomatic way to do this.

src/openforms/variables/service.py
```diff
@@ -2,27 +2,62 @@
...
+      the default registry will be used.
 
     You should not rely on the order of returned variables, as they are registered in
     the order the Django apps are loaded - and this may change without notice.
     """
     return [
         registered_variable.get_static_variable(submission=submission)
-        for registered_variable in register
+        for registered_variable in variables_registry or static_variables_registry
+    ]
```
sergei-maertens, line 40: Can you not provide the `static_variables_registry` as default kwarg instead of `None`?

Otherwise, I prefer the explicit form:

```py
if variables_registry is None:
    variables_registry = static_variables_registry
```
the `or` in the list comprehension is less readable

src/openforms/variables/service.py
```diff
@@ -2,27 +2,62 @@
...
+
+    return [
+        {
+            "plugin_identifier": plugin.identifier,
+            "plugin_verbose_name": plugin.verbose_name,
+            "plugin_variables": get_static_variables(variables_registry=registry),
+        }
+        for plugin in registration_plugins_registry.iter_enabled_plugins()
+        if (registry := plugin.get_variables_registry()) is not None
     ]
```
sergei-maertens, line 63: The re-packaging of the plugin information into a dict is a bit weird, why not make your serializer so that you can simply pass the plugin instance?

```py
serializer = RegistrationPluginVariablesSerializer(instance=plugin)
serializer.data  # spits out the same data
```

This is pretty straight forward:

```py
class RegistrationPluginVariablesSerializer(serializers.Serializer):
    plugin_identifier = serializers.CharField(source="identifier")
    plugin_verbose_name = serializers.CharField(source="verbose_name")
    plugin_variables = VariableSerializer(source="get_variables")
```

and then implement the `get_variables` method on your plugin class.

You don't need to deal with the typed dict at all or the mappings/transformations, that's what the serializer is supposed to do.

sergei-maertens (changes requested):

src/openforms/variables/service.py
Viicos: > Can you not provide the `static_variables_registry` as default kwarg instead of `None`?

I went for this in the first place, but felt a bit weird. So I'll go with explicit is None check

src/openforms/variables/service.py
Viicos: Yeah `get_registration_plugins_variables` was supposed to be used elsewhere before, but now it makes sense as it is only being used by the endpoint

src/openforms/registrations/contrib/objects_api/models.py
Viicos: Considering this requires thinking and probably some work, it will probably be part of the next PR

sergei-maertens (approved):
Just need to fix CI :)

PR #3991, opened 2024-03-11
[#3881] Update variables in reusable form definitions

### LINKED ISSUE #3881: Two forms with a re-usable form definition "sometimes" leads to an incorrect form in one of the two
We found this issue when figuring out why voorletters was suddenly not working in the prefill when migrating from 2.4.3 to 2.4.5.

The client said it didn't work. Made the initials field writable (it was readonly) so it could be filled, and saved the form (this step was a re-usable form definition). When we tested it, it worked. Silvia suggested it could be related to a variable not being created and it was then created when the form was saved, making the form function.

This might also be the case why a re-usable form definition works in form A but not in form B, after the form definition is changed in form A.

vaszig:
Fixes #3881

### DIFF
--- a/src/openforms/forms/tasks.py
+++ b/src/openforms/forms/tasks.py
@@ -6,12 +6,33 @@ from django.db import DatabaseError, transaction
 from django.db.utils import IntegrityError
 from django.utils import timezone
 
+from celery import chain
 from djangorestframework_camel_case.util import camelize
+
+from openforms.variables.constants import FormVariableSources
 
 from ..celery import app
 from .models import Form
 
 logger = logging.getLogger(__name__)
+
+
+def on_variables_bulk_update_event(form_id: int) -> None:
+    """
+    Celery chain of tasks to execute on a bulk update of variables event.
+    """
+    repopulate_reusable_definition_variables_task = (
+        repopulate_reusable_definition_variables_to_form_variables.si(form_id)
+    )
+    recouple_submission_variables_task = (
+        recouple_submission_variables_to_form_variables.si(form_id)
+    )
+
+    actions_chain = chain(
+        repopulate_reusable_definition_variables_task,
+        recouple_submission_variables_task,
+    )
+    actions_chain.delay()
 
 
 @app.task(ignore_result=True)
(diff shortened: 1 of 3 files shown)


src/openforms/forms/tasks.py
```diff
@@ -14,6 +15,23 @@
 logger = logging.getLogger(__name__)
 
 
+def on_variables_bulk_update_event(form_id: int) -> None:
+    """
+    Celery chain of tasks to execute on a bulk update of variables completion event.
```
sergei-maertens, line 20:
```suggestion
    Celery chain of tasks to execute on a bulk update of variables event.
```

src/openforms/forms/tasks.py
```diff
@@ -14,6 +15,23 @@
...
+    """
+    recouple_reusable_definition_variables_task = (
+        recouple_reusable_definition_variables_to_form_variables.si(form_id)
+    )
```
sergei-maertens, line 24:
```suggestion
    repopulate_reusable_definition_variables_task = (
        recouple_reusable_definition_variables_to_form_variables.si(form_id)
    )
```
perhaps? it's not necessarily fixing broken relations, rather it's deriving the variables from the new FD again :thinking:

src/openforms/forms/tasks.py
```diff
@@ -87,6 +105,33 @@ def recouple_submission_variables_to_form_variables(form_id: int) -> None:
...
 
 
+@app.task(ignore_result=True)
+def recouple_reusable_definition_variables_to_form_variables(form_id: int) -> None:
+    """
+    When the FormVariable bulk create/update endpoint is called, all existing FormVariable related to the form are
+    deleted and new are created. If there are any form definitions which are reusable, we want to update all the forms
+    which are also using these FormDefinitions (concerning the FormVariables). This task updates the FormVariables
+    of each related form in the database, by deleteing the old ones and recreating them (the most safe way).
+    """
```
sergei-maertens, line 115:
```suggestion
    """
    Fix inconsistencies created by updating a re-usable form definition which is used in other forms.

    When the FormVariable bulk create/update endpoint is called, all existing FormVariable related to the form are
    deleted and new are created. If there are any form definitions which are reusable, we want to update all the forms
    which are also using these FormDefinitions (concerning the FormVariables). This task updates the FormVariables
    of each related form in the database, by replacing the old state with the newly derived set of form variables.
    """
```
in docstrings, you should always provide a one-line summary and the put the extra information below it. See PEP-257 :)

src/openforms/forms/tasks.py
```diff
@@ -87,6 +105,33 @@ def recouple_submission_variables_to_form_variables(form_id: int) -> None:
...
+    for fd in reusable_form_definitions:
+        form_ids_to_update = fd.formstep_set.exclude(form=form).values_list(
+            "form", flat=True
+        )
+
+        for form_id in form_ids_to_update:
+            form_to_update = Form.objects.get(id=form_id)
+            form_to_update.formvariable_set.all().delete()
+
+            FormVariable.objects.create_for_form(form_to_update)
```
sergei-maertens, line 132: I think you can simplify and optimize this:

```py
fds = FormDefinition.objects.filter(formstep__form=form_id, is_reusable=True)  # 0 queries
other_forms = Form.objects.filter(formstep__form_definition__in=fds).exclude(id=form_id)  # 0 queries
# delete the existing form variables, we will re-create them
FormVariable.objects.filter(form__in=other_forms, source=FormVariableSources.component).delete()  # 1 query
for form in other_forms:  # 2 queries
    FormVariable.objects.create_for_form(form)  # x number of queries for each form
```

We can avoid of nested loops (but you'll have to mark the task as `@transaction.atomic`!

The logic is:

1. Figure out which re-usable form definitions are involved
2. Get a collection of all the other forms that use any of these re-usable form definitions
3. Delete all their component form variables
4. Regenerate the component form variables for each form

src/openforms/forms/tasks.py
```diff
@@ -87,6 +105,33 @@ def recouple_submission_variables_to_form_variables(form_id: int) -> None:
...
+    ).distinct()
+
+    for fd in reusable_form_definitions:
+        form_ids_to_update = fd.formstep_set.exclude(form=form).values_list(
+            "form", flat=True
+        )
+
+        for form_id in form_ids_to_update:
+            form_to_update = Form.objects.get(id=form_id)
+            form_to_update.formvariable_set.all().delete()
```
sergei-maertens, line 130: this is also destroying the user defined variables, which is *not* okay. Please add a test case of a form with user defined variables too and assert that those are not broken.

sergei-maertens (changes requested):

src/openforms/forms/tasks.py
vaszig: Yes, this was done by accident. At the beginning I was filtering out only `FormVariableSources.component` sources but yes, I will add a test for this as well.

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.5.x: 0f5cdba087d549a82b0b00350d59d0abcf5ae19f
* 2.4.x: afafb954903c5247d262c4fb7d04be198af34041
* 2.3.x: ad3eabd3110f0343f7766f1f55c1b46e12c0facc

PR #3992, opened 2024-03-11
Type improvements for validation feature

### LINKED ISSUE #3977: Huisletter and huisletter-toevoeging staan meer tekens toe dan de BAG

TODO's:

sergei-maertens:
Partly closes #3977

Some type-hint preparations before committing the feature itself (to keep PRs smaller)


src/openforms/formio/typing/base.py
```diff
@@ -9,32 +9,36 @@
...
-    maxLength: int
-    min: int | float
-    max: int | float
+class Validate(TypedDict):
+    required: NotRequired[bool]
+    maxLength: NotRequired[int]
+    pattern: NotRequired[str]
+    min: NotRequired[int | float]
+    max: NotRequired[int | float]
+    plugins: NotRequired[list[str]]
```
Viicos, line 27: Why not `total=False`? This would avoid using `NotRequired` for all fields

src/openforms/formio/typing/base.py
```diff
@@ -53,7 +57,7 @@ class PrefillConfiguration(TypedDict):
     identifierRole: Literal["main", "authorised_person"]
 
 
-class Component(TypedDict, total=False):
+class Component(TypedDict):
```
Viicos, line 60: I think it would make more sense to make it `total=False` and use `Required` for the couple required attributes?

Viicos (approved):

src/openforms/formio/typing/base.py
sergei-maertens: discussed in office, leaving this one as it is :)

src/openforms/formio/registry.py
```diff
@@ -27,22 +27,21 @@
...
     from openforms.submissions.models import Submission
 
+T = TypeVar("T", bound=Component, contravariant=True)
 
-class FormatterProtocol(Protocol):
-    def __init__(self, as_html: bool): ...
-
-    def __call__(self, component: Component, value: Any) -> str: ...
 
+class FormatterProtocol(Generic[T], Protocol):
```
Viicos, line 33: You can use the shorthand `Protocol[T]` here, this is a special case for protocols

src/openforms/formio/registry.py
```diff
@@ -27,22 +27,21 @@
 if TYPE_CHECKING:
     from openforms.submissions.models import Submission
 
+T = TypeVar("T", bound=Component, contravariant=True)
```
Viicos, line 30: I like having the actual bound type specified in the name, makes it clearer in my opinion: `ComponentT`.

But this is just personal preference

src/openforms/formio/registry.py
```diff
@@ -52,18 +51,18 @@ class BasePlugin(Generic[T], AbstractBasePlugin):
...
-    formatter: type[FormatterProtocol]
+    formatter: type[FormatterProtocol[T]]
     """
     Specify the callable to use for formatting.
 
     Formatter (class) implementation, used by
     :meth:`openforms.formio.registry.ComponentRegistry.format`.
     """
-    normalizer: None | NormalizerProtocol = None
+    normalizer: None | NormalizerProtocol[T] = None
```
Viicos, line 61:
```suggestion
    normalizer: NormalizerProtocol[T] | None = None
```

Feels weird to have `None` as the first type of the union?

Viicos (approved):

src/openforms/formio/registry.py
sergei-maertens: TIL, didn't know that was already in 3.10

Viicos: Haven't thought about it until now, but it does look similar to the new 3.12 syntax

PR #3994, opened 2024-03-11
Implement formio form-to-serializer translation

### LINKED ISSUE #3977: Huisletter and huisletter-toevoeging staan meer tekens toe dan de BAG

sergei-maertens:
Partly closes #3977

This PR implements the conversion layer from a formio form definition (as a sequence of components) to a DRF serializer, where each component is translated into a serializer field (or nested serializer).

Each field is then capable/responsible of validating the user input, effectively running the formio validation server-side.

A future PR will be hook this up into our submissions/appointment serializers.

### DIFF
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -41,6 +41,7 @@ from ..formatters.formio import (
 from ..registry import BasePlugin, register
 from ..serializers import build_serializer
 from ..typing import (
+    Component,
     ContentComponent,
     FileComponent,
     RadioComponent,
@@ -104,6 +105,26 @@ class TextField(BasePlugin[TextFieldComponent]):
 @register("email")
 class Email(BasePlugin):
     formatter = EmailFormatter
+
+    def build_serializer_field(
+        self, component: Component
+    ) -> serializers.EmailField | serializers.ListField:
+        multiple = component.get("multiple", False)
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        if validate.get("plugins", []):
+            raise NotImplementedError("Plugin validators not supported yet.")
+
+        # dynamically add in more kwargs based on the component configuration
+        extra = {}
+        if (max_length := validate.get("maxLength")) is not None:
+            extra["max_length"] = max_length
+
+        base = serializers.EmailField(
+            required=required, allow_blank=not required, allow_null=False, **extra
+        )
+        return serializers.ListField(child=base) if multiple else base
 
 
 @register("time")
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -300,6 +300,26 @@ class Radio(BasePlugin[RadioComponent]):
             return
         translate_options(options, language_code, enabled)
 
+    def build_serializer_field(
+        self, component: RadioComponent
+    ) -> serializers.ChoiceField:
+        """
+        Convert a radio component to a serializer field.
+
+        A radio component allows only a single value to be selected, but selecting a
+        value may not be required. The available choices are taken from the ``values``
+        key, which may be set dynamically (see :meth:`mutate_config_dynamically`).
+        """
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+        choices = [(value["value"], value["label"]) for value in component["values"]]
+        return serializers.ChoiceField(
+            choices=choices,
+            required=required,
+            allow_blank=not required,
+            allow_null=not required,
+        )
+
 
 @register("signature")
 class Signature(BasePlugin):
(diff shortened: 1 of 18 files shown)


src/openforms/formio/registry.py
```diff
@@ -78,6 +80,26 @@ def mutate_config_dynamically(
...
 
+    def build_serializer_field(self, component: ComponentT) -> serializers.Field:
+        # the default implementation is a compatibility shim while we transition to
+        # the new backend validation mechanism.
+        warnings.warn(
+            "Relying on the default/implicit JSONField for component type "
+            f"{component['type']} is deprecated. Instead, define the "
+            "'build_serializer_field' method on the specific component plugin.",
+            DeprecationWarning,
+        )
```
Viicos, line 91:
```suggestion
    @typing_extensions.deprecated("msg")
    def build_serializer_field(self, component: ComponentT) -> serializers.Field:
        # the default implementation is a compatibility shim while we transition to
        # the new backend validation mechanism.
```

Identical runtime behavior + typing support :)

src/openforms/formio/service.py
```diff
@@ -105,3 +107,9 @@ def get_dynamic_configuration(
     # as checkboxes/dropdowns/radios/...
     inject_prefill(config_wrapper, submission)
     return config_wrapper
+
+
+def build_serializer(
+    components: list[Component], _register: ComponentRegistry | None = None, **kwargs
+):
+    return _build_serializer(components, register=_register or register, **kwargs)
```
Viicos, line 115: If this is supposed to be the public function, would be great to have the docstring here as well

src/openforms/formio/components/vanilla.py
```diff
@@ -62,6 +67,39 @@ class Default(BasePlugin):
...
 
+    def build_serializer_field(
+        self, component: TextFieldComponent
+    ) -> serializers.CharField | serializers.ListField:
+        multiple = component.get("multiple", False)
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        if validate.get("plugins", []):
+            raise NotImplementedError("Plugin validators not supported yet.")
```
Viicos, line 80: Are we certain that no components currently have plugins set? If not, maybe we should just ignore them.

If this is purely a formio thing, then it's probably fine

src/openforms/formio/components/vanilla.py
```diff
@@ -348,3 +349,28 @@ def rewrite_for_request(component: ContentComponent, request: Request):
...
+    def build_serializer_field(
+        self, component: EditGridComponent
+    ) -> serializers.ListField:
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+        nested = build_serializer(
+            components=component.get("components", []),
+            # XXX: check out type annotations here, there's some co/contra variance
+            # in play
+            register=self.registry,
```
Viicos, line 365: This can be fixed by adding `registry: "ComponentRegistry"` as a type annotation on `BasePlugin` (currently `self.registry` refers to `BaseRegistry`, and `build_serializer` expects `ComponentRegistry` -- inheritance is `BaseRegistry > ComponentRegistry`).

Viicos (approved):
Dynamic serializers is an interesting approach, I'm wondering if it could have been implemented differently

src/openforms/formio/registry.py
sergei-maertens: but breaks my <kbd>CTRL</kbd> + <kbd>F</kb> on `DeprecationWarning` to track down stuff to remove in future releases :/

src/openforms/formio/components/vanilla.py
sergei-maertens: no, this will break and I will resolve it in the next few days

this is a negative side-effect of ignoring coverage for `NotImplementedError` btw, could've ended up in runtime like this without being aware :/

sergei-maertens:
> I'm wondering if it could have been implemented differently

I had a different approach initially where I was constructing validation chains, but then I ended up having to parse/normalize input data and realized I was re-inventing serializer fields/serializers.

The original idea was to run formio.js in a separate container and programmatically invoke the validation that way, but with our efforts to move away from formio.js that's not viable anymore anyway. And formio.js validation implementation is not pleasant to work with anyway.

src/openforms/formio/registry.py
Viicos: > but breaks my CTRL + F on `DeprecationWarning` to track down stuff to remove in future releases :/

`@deprecated("", category=...)` could do it, but as you said not having access to `component` kills it anyway

src/openforms/formio/components/vanilla.py
Viicos: Ah yes saw it fixed the XXX comment, but created this new error

PR #3996, opened 2024-03-11
[#3961] Provide separate urls for KVK Zoeken and Basisprofiel

### LINKED ISSUE #3961: KvK Zoeken and Basisprofiel APIs don't necesserily share the same base URL when using gateways

2.4.x

Since OF 2.4 you can only provide 1 base URL for these two KvK API. But in cases where gateways are used, these APIs can have different base URL.

We should go back to having two services.

If possible, if upgrading from 2.3, skip the step where these two services were merged.

vaszig:
Closes #3961


src/openforms/contrib/kvk/migrations/0007_kvkconfig_profile_service_kvkconfig_search_service.py
```diff
@@ -0,0 +1,76 @@
...
+    be the same for both operations so in this case we use one Service configuration.
+    """
+    KVKConfig = apps.get_model("kvk", "KVKConfig")
+    config = KVKConfig.objects.first()
+    if config is None:
+        return
+
+    if service := config.service:
+        if not (root := service.api_root):
+            return
```
Viicos, line 24:
```suggestion
    if (service := config.service) and not (root := service.api_root):
        return
```

But I'll not that this code is unsafe (`root` should be marked as possibly unbound below)

At the end of this check, you either have `service` and `service.api_root` both not `None`, or `service` is `None`. In the later case, `root` was never assigned to a value.

src/openforms/contrib/kvk/migrations/0007_kvkconfig_profile_service_kvkconfig_search_service.py
```diff
@@ -0,0 +1,76 @@
...
+        config.search_service = service
+    elif "basisprofielen" in root:
+        config.profile_service = service
+    else:
+        # case where we have defined generic API root url (e.g. https://kvk.nl/test/api/)
+        config.search_service = service
+        config.profile_service = service
+
+    config.save()
+    caches[settings.SOLO_CACHE].clear()
```
Viicos, line 36: I'm not sure this is necessary. When working with models in migrations, we are dealing with "fake" models, and all the custom stuff added to model classes are not available. So I'm wondering if the solo cache exists at this point :thinking:

Viicos (changes requested):
Will continue after lunch!

src/openforms/contrib/kvk/client.py
```diff
@@ -81,7 +79,21 @@ def get_search_results(self, query_params: SearchParams):
...
+) -> KVKProfileClient | KVKSearchClient:
+    service_mappings = {
+        "profile": KVKProfileClient,
+        "search": KVKSearchClient,
+    }
+    config = KVKConfig.get_solo()
+    assert isinstance(config, KVKConfig)
+    if not (service := getattr(config, f"{client}_service")):
+        raise NoServiceConfigured(f"No KVK {client} service configured!")
+    return build_client(service, client_factory=service_mappings[client])
```
Viicos, line 99: I think it would be easier to split this helper function in two ones. Without the use of overloads, it's not possible to know the return type from the `client` argument. So perhaps something like `get_kvk_profile_client` and `get_kvk_search_client`

src/openforms/contrib/kvk/client.py
```diff
@@ -43,12 +35,14 @@ class SearchParams(TypedDict, total=False):
...
 
 
-class KVKClient(HALClient):
+class KVKProfileClient(HALClient):
     @elasticapm.capture_span("app.kvk")
     def get_profile(self, kvk_nummer: str) -> BasisProfiel:
         """
         Retrieve the profile of a single entity by chamber of commerce number.
 
+        :arg kvk_nummer: must be a Dutch Chamber of Commerce number and consists of 8 digits.
```
Viicos, line 44:
```suggestion
        :arg kvk_nummer: a Dutch Chamber of Commerce number consisting of 8 digits.
```

src/openforms/contrib/kvk/tests/test_migrations.py
Viicos, line 1: I don't think it's necessary to have a mixin if `KVKZoekenServicesMigrationTests` is the only test class making use of the super's `setUpBeforeMigration`. Or if we keep the mixin it should only define the `app` and `migrate_from/to` and had `setUpBeforeMigration` defined in all test classes

Viicos (changes requested):

src/openforms/contrib/kvk/client.py
vaszig: It was my first thought but then I realized that they were almost identical that's why I did that in one function.
In my opinion, what you suggest is more clear so I will fix that.

src/openforms/contrib/kvk/models.py
```diff
@@ -20,13 +23,24 @@ class KVKConfig(SingletonModel):
...
+    profile_service = models.OneToOneField(
         "zgw_consumers.Service",
-        verbose_name=_("KvK API"),
-        help_text=_("Service for API interaction with the KVK."),
+        verbose_name=_("KvK API Basisprofiel"),
+        help_text=_("API used to retrieve basis profielen."),
         on_delete=models.PROTECT,
         limit_choices_to={"api_type": APITypes.orc},
         related_name="+",
+        blank=True,
```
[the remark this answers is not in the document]
vaszig: A user can define one service (the one in which he/she is interested in) so I wanted to make this possible in the admin interface. Do you agree?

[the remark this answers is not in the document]
Viicos: Yes, I'm always confused with how `blank` behaves with foreign keys. So is it true that before this change it wasn't possible to _not_ have any service set on the model? (as it was `null=True`, but not `blank`)

src/openforms/contrib/kvk/migrations/0007_kvkconfig_profile_service_kvkconfig_search_service.py
Viicos: Anyway it doesn't hurt to have this clear call :)

src/openforms/contrib/kvk/models.py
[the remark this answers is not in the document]
sergei-maertens: both need to be specified explicitly, let's not do any magic inference. the data migration should copy the single service to both fields to make sure users of `api.kvk.nl` do not have their functionality broken.

src/openforms/contrib/kvk/migrations/0007_kvkconfig_profile_service_kvkconfig_search_service.py
```diff
@@ -0,0 +1,75 @@
...
+        return
+
+    if "zoeken" in root:
+        config.search_service = service
+    elif "basisprofielen" in root:
+        config.profile_service = service
+    else:
+        # case where we have defined generic API root url (e.g. https://kvk.nl/test/api/)
+        config.search_service = service
+        config.profile_service = service
```
sergei-maertens, line 32: having `zoeken` or `basisprofielen` in root would already break the API clients, since they use the `v2/zoeken` relative paths themselves, so that'd break the existing clients.

Just assigning the service to both fields is enough:

```suggestion
    config.search_service = service
    config.profile_service = service
```

src/openforms/contrib/kvk/migrations/0007_kvkconfig_profile_service_kvkconfig_search_service.py
sergei-maertens: if recently a request was made with KVK the value is populated in the cache, then the DB migration updates the config in the DB but the cache is still stale and will be re-used the next time the KVK feature is used.

So cache invalidation is required.

src/openforms/contrib/kvk/models.py
```diff
@@ -20,13 +23,26 @@ class KVKConfig(SingletonModel):
...
+    profile_service = models.OneToOneField(
         "zgw_consumers.Service",
-        verbose_name=_("KvK API"),
-        help_text=_("Service for API interaction with the KVK."),
+        verbose_name=_("KvK API Basisprofiel"),
+        help_text=_("Service for API used to retrieve basis profielen."),
         on_delete=models.PROTECT,
         limit_choices_to={"api_type": APITypes.orc},
         related_name="+",
+        blank=True,
```
sergei-maertens, line 33:
```suggestion
```

src/openforms/contrib/kvk/models.py
```diff
@@ -20,13 +23,26 @@ class KVKConfig(SingletonModel):
...
+    search_service = models.OneToOneField(
+        "zgw_consumers.Service",
+        verbose_name=_("KvK API Zoeken"),
+        help_text=_(
+            "Service for API used for validation of KvK, RSIN and vestigingsnummer's."
+        ),
+        on_delete=models.PROTECT,
+        limit_choices_to={"api_type": APITypes.orc},
+        related_name="+",
+        blank=True,
```
sergei-maertens, line 45:
```suggestion
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #3998, opened 2024-03-11
Apply formio component validations during submission complete

### LINKED ISSUE #3977: Huisletter and huisletter-toevoeging staan meer tekens toe dan de BAG

sergei-maertens:
Closes #3977
Depends on #3997


Viicos (approved):
Two questions/remarks:
- Will this be handled properly in the SDK?
- Maybe we should add a test for the specific "huisletter" and "huisletter-toevoeging"?

sergei-maertens:
The SDK already runs client validation. This covers people using the API directly, so it will not be handled in the SDK

I will add more (end to end) tests in another pr. There's already a unit test for house number regex!

Issue #3705, opened 2023-12-21
__str__ representation of a Submission uses GMT time instead of Europe/AMS time


alextreme:
### Thema / Theme

Admin

### Omschrijving / Description

Taiga Dimpact SSC 36, classified as a low-prio enhancement

In the admin the __str__ representation (shown in the breadcrumb, but also in sentry) shows the time as GMT, even though the timezone is set to Europe/Amsterdam. In the listview of submissions the time is shown using the correct timezone, so this isn't consistent

![image](https://github.com/open-formulieren/open-forms/assets/226839/96be2746-8ba9-42f9-9627-662e533b6502)

alextreme:
Taiga Dimpact 54

Issue doesn't seem properly resolved in OF 2.5, confirmed on openforms test instance using latest:


![image](https://github.com/open-formulieren/open-forms/assets/226839/e42c6c37-fd84-43ad-b43f-2431c94b3d4d)

alextreme:
@Viicos does this also fix the times further down in the section 'inzendingsstappen' ?

![image](https://github.com/open-formulieren/open-forms/assets/226839/b0d1f63f-350c-4abc-b12c-1fdd8095be3c)

alextreme:
This is also an issue when in the admin -> Payments and seeing the submission, there is a difference of 2 hours
Taiga LV 161

alextreme:
Taiga Dimpact 54 (updated july 10th)

This has been resolved for the Submissions, however not on the SubmissionSteps

sergei-maertens:
Added to the board, this should be picked up. Whoever picks this up, please talk to Alex so we get a full inventory of incorrectly reported timestamps.

sergei-maertens:
Not backporting this, it's a bit confusing but the detail information is available if you know where to look.

PR #4001, opened 2024-03-13
[#3705] Fix submission `__str__` again

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/3705, take two.


vaszig:
Does it have to do with 3705 or 3688?I see different tasks in the title, comment and commit message.

Viicos:
> Does it have to do with 3705 or 3688?I see different tasks in the title, comment and commit message.

I'm so used to issue 3688 now, that's why :sob:

vaszig (approved):

PR #4003, opened 2024-03-13
Fix config checks for Objects API plugin

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/4007

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/checks.py
+++ b/src/openforms/registrations/contrib/objects_api/checks.py
@@ -15,17 +15,24 @@ from .models import ObjectsAPIConfig
 
 def check_objects_service():
     with get_objects_client() as client:
-        client.get("objects", params={"pageSize": 1})
+        resp = client.get("objects", params={"pageSize": 1})
+        resp.raise_for_status()
+        if not 200 <= resp.status_code < 300:
+            raise InvalidPluginConfiguration(_("Missing API credentials"))
 
 
 def check_objecttypes_service():
     with get_objecttypes_client() as client:
-        client.get("objecttypes", params={"pageSize": 1})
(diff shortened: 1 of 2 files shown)


sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/checks.py
```diff
@@ -15,17 +15,24 @@
...
+            raise InvalidPluginConfiguration(_("Missing API credentials"))
 
 
 def check_objecttypes_service():
     with get_objecttypes_client() as client:
-        client.get("objecttypes", params={"pageSize": 1})
+        resp = client.get("objecttypes", params={"pageSize": 1})
+        resp.raise_for_status()
+        if not 200 <= resp.status_code < 300:
+            raise InvalidPluginConfiguration(_("Missing API credentials"))
```
sergei-maertens, line 29: there's a missing test :)

Viicos: :expressionless:

Viicos: Implementation is flawed in my opinion because tests rely on the order of checks, which makes it annoying to test

sergei-maertens: setting it up is also annoying, so testing/checking makes it annoying too

the reality is that the plugin as a whole is validated and multiple aspects require correct configuration for that. The tests should really be set up as all relevant permutations: one service is not properly configured, all the rest are -> the check should highlight something is wrong. If the "other services" are set up correctly, the order aspect becomes irrelevant.

sergei-maertens (approved):

PR #4011, opened 2024-03-14
End to end tests for input validation

### LINKED ISSUE #3977: Huisletter and huisletter-toevoeging staan meer tekens toe dan de BAG

sergei-maertens:
Closes #3977
Depends on #4010

Adds tests to lock in the parity between frontend and backend behaviour.

### DIFF
--- a/src/openforms/tests/e2e/input_validation_base.py
+++ b/src/openforms/tests/e2e/input_validation_base.py
@@ -1,0 +1,142 @@
+from typing import TypedDict
+
+from django.test import override_settings, tag
+from django.urls import reverse
+
+from asgiref.sync import async_to_sync
+from furl import furl
+from playwright.async_api import expect
+from rest_framework import status
+from typing_extensions import NotRequired, Unpack
+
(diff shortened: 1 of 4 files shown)


src/openforms/tests/e2e/input_validation_base.py
```diff
@@ -0,0 +1,145 @@
...
+# allow all origins, since we don't know exactly the generated live server port number
+@override_settings(CORS_ALLOW_ALL_ORIGINS=True)
+class ValidationsTestCase(SubmissionsMixin, E2ETestCase):
+
+    def assertValidationIsAligned(
+        self,
+        component: Component,
+        ui_input: str | int | float,
+        expected_ui_error: str,
+        **kwargs: Unpack[ValidationKwargs],
```
Viicos, line 61: Funny alternative to not use an Unset default value

sergei-maertens: I tried using a singleton but that's horrendous with type checking :(

PR #4013, opened 2024-03-14
[#3688] Add docker configuration for Objects API

Viicos:
Part of #3688.

Not tested yet, but do not hesitate to let me know if things need to be changed

### DIFF
--- a/docker/docker-compose.objects-apis.yml
+++ b/docker/docker-compose.objects-apis.yml
@@ -1,12 +1,16 @@
 version: '3'
 
 services:
+  redis:
+    image: redis:7
+    command: ["redis-server", "--appendonly", "yes"]
+
   db:
-    image: postgres:${PG_VERSION:-14}
+    image: postgis/postgis:${PG_VERSION:-14-master}
     environment:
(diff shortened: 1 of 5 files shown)


docker/docker-compose.objects-apis.yml
```diff
@@ -7,6 +7,7 @@ services:
       - POSTGRES_HOST_AUTH_METHOD=trust
     volumes:
       - ./objects-apis/docker-init-objecttypes-api-db.sql:/docker-entrypoint-initdb.d/init_db.sql
+      - ./objects-apis/docker-init-objects-api-db.sql:/docker-entrypoint-initdb.d/init_db.sql
```
sergei-maertens, line 10: this overwrites the previous entry. Either you mount the directory itself and put the SQL scripts in there like `001_objecttypes.sql` and `002_objects.sql`, or just add the extra SQL to the existing file. It's okay to treat objects and objecttypes API as one "unit" here.

sergei-maertens (changes requested):

sergei-maertens (approved):

Issue #72, opened 2020-09-16
As developer I want to implement frontend and backend validation of form fields


svenvandescheur:
The current setup moves to arbitrary data storage based on frontend validation. Form.io support validation rules on the form (at the time of writing not yet enabled) but we do need to enforce these on the backend as well for security reasons.

Components to still implement validation for:

- [x] textarea
- [x] datetime @sergei-maertens 
- [x] time
- [x] postcode
- [x] bsn (11-check)
- [x] file @Viicos 
- [x] selectboxes @sergei-maertens 
- [x] select
- [x] checkbox
- [x] currency
- [x] iban
- [x] licenseplate
- ~~npFamilyMembers~~ nothing to do - this gets transformed into a radio component
- [ ] addressNL
- ~~co-sign v1~~ no more efforts in legacy stuff
- [x] co-sign v2 
- [x] password
- [x] signature
- [x] map

Components that don't take values: `content`, `columns`, `fieldset`

sergei-maertens:
not done

PR #4014, opened 2024-03-14
:sparkles: [#3977] Add submission context to formio data serializer

sergei-maertens:
Part of #3977
Part of #72 

So that the plugin validators can run in the context of the provided submission and not crash.


vaszig (approved):
I think the failing test is a flaky one,correct?

sergei-maertens:
> I think the failing test is a flaky one,correct?

Unfortunately, yes!

PR #4029, opened 2024-03-20
[#72] Add backend validation of form rules

vaszig:
Partly closes #72

This PR handles the following components, the rest of them will be handled in an another PR:

- textarea
- time
- postcode
- bsn (11-check)
- select
- checkbox
- currency
- signature
- map

### DIFF
--- a/src/openforms/formio/api/validators.py
+++ b/src/openforms/formio/api/validators.py
@@ -158,9 +158,3 @@ class NoVirusValidator:
                 raise serializers.ValidationError(
                     _("The virus scan returned an unexpected status.")
                 )
-
-
-class TruthyBooleanValueValidator:
-    def __call__(self, value):
-        if not value:
-            raise serializers.ValidationError(_("This field must be true."))
--- a/src/openforms/formio/api/validators.py
+++ b/src/openforms/formio/api/validators.py
@@ -158,3 +158,9 @@ class NoVirusValidator:
                 raise serializers.ValidationError(
                     _("The virus scan returned an unexpected status.")
                 )
+
+
+class TruthyBooleanValueValidator:
+    def __call__(self, value):
+        if not value:
+            raise serializers.ValidationError(_("This field must be true."))
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -5,6 +5,7 @@ Custom component types (defined by us or third parties) need to be organized in 
 adjacent custom.py module.
 """
 
+import logging
 from datetime import time
 from typing import TYPE_CHECKING
 
@@ -57,6 +58,9 @@ from .utils import _normalize_pattern
 
 if TYPE_CHECKING:
     from openforms.submissions.models import Submission
+
+
+logger = logging.getLogger(__name__)
 
 
 @register("default")
@@ -147,23 +151,79 @@ class Email(BasePlugin):
         return serializers.ListField(child=base) if multiple else base
 
 
+class FormioTimeField(serializers.TimeField):
+    def validate_empty_values(self, data):
+        is_empty, data = super().validate_empty_values(data)
+        # base field only treats `None` as empty, but formio uses empty strings
+        if data == "":
+            if self.required:
+                self.fail("required")
+            return (True, "")
+        return is_empty, data
+
+
+class TimeBetweenValidator:
+
+    def __init__(self, min_time: time, max_time: time) -> None:
+        self.min_time = min_time
+        self.max_time = max_time
+
+    def __call__(self, value: time):
+        # same day - straight forward comparison
+        if self.min_time < self.max_time:
+            if value < self.min_time:
+                raise serializers.ValidationError(
+                    _("Value is before minimum time"),
+                    code="min_value",
+                )
+            if value > self.max_time:
+                raise serializers.ValidationError(
+                    _("Value is after maximum time"),
+                    code="max_value",
+                )
+
+        # min time is on the day before the max time applies (e.g. 20:00 -> 04:00)
+        else:
+            if value < self.min_time and value > self.max_time:
+                raise serializers.ValidationError(
+                    _("Value is not between mininum and maximum time."), code="invalid"
+                )
+
+
 @register("time")
 class Time(BasePlugin[Component]):
     formatter = TimeFormatter
 
     def build_serializer_field(
         self, component: Component
-    ) -> serializers.TimeField | serializers.ListField:
+    ) -> FormioTimeField | serializers.ListField:
         multiple = component.get("multiple", False)
         validate = component.get("validate", {})
         required = validate.get("required", False)
 
         validators = []
-        if min_time := validate.get("minTime"):
-            validators.append(MinValueValidator(time.fromisoformat(min_time)))
-        if max_time := validate.get("maxTime"):
-            validators.append(MaxValueValidator(time.fromisoformat(max_time)))
-        base = serializers.TimeField(
+
+        match (
+            min_time := validate.get("minTime"),
+            max_time := validate.get("maxTime"),
+        ):
+            case (None, None):
+                pass
+            case (str(), None):
+                validators.append(MinValueValidator(time.fromisoformat(min_time)))
+            case (None, str()):
+                validators.append(MaxValueValidator(time.fromisoformat(max_time)))
+            case (str(), str()):
+                validators.append(
+                    TimeBetweenValidator(
+                        time.fromisoformat(min_time),
+                        time.fromisoformat(max_time),
+                    )
+                )
+            case _:
+                logger.warning("Got unexpected min/max time in component %r", component)
+
+        base = FormioTimeField(
             required=required,
             allow_null=not required,
             validators=validators,
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -5,9 +5,10 @@ Custom component types (defined by us or third parties) need to be organized in 
 adjacent custom.py module.
 """
 
+from datetime import datetime
 from typing import TYPE_CHECKING
 
-from django.core.validators import RegexValidator
+from django.core.validators import MaxValueValidator, MinValueValidator, RegexValidator
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
@@ -42,30 +43,25 @@ from ..formatters.formio import (
 from ..registry import BasePlugin, register
 from ..serializers import build_serializer
 from ..typing import (
+    CheckboxComponent,
     Component,
     ContentComponent,
+    CurrencyComponent,
     EditGridComponent,
     FileComponent,
     RadioComponent,
     SelectBoxesComponent,
     SelectComponent,
+    SignatureComponent,
+    TextAreaComponent,
     TextFieldComponent,
+    TimeComponent,
 )
 from .translations import translate_options
+from .utils import _normalize_pattern
 
 if TYPE_CHECKING:
     from openforms.submissions.models import Submission
-
-
-def _normalize_pattern(pattern: str) -> str:
-    """
-    Normalize a regex pattern so that it matches from beginning to the end of the value.
-    """
-    if not pattern.startswith("^"):
-        pattern = f"^{pattern}"
-    if not pattern.endswith("$"):
-        pattern = f"{pattern}$"
-    return pattern
 
 
 @register("default")
@@ -157,8 +153,31 @@ class Email(BasePlugin):
 
 
 @register("time")
-class Time(BasePlugin):
+class Time(BasePlugin[TimeComponent]):
     formatter = TimeFormatter
+
+    def build_serializer_field(
+        self, component: TimeComponent
+    ) -> serializers.TimeField | serializers.ListField:
+        multiple = component.get("multiple", False)
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        validators = []
+        if min_time := validate.get("minTime"):
+            validators.append(
+                MinValueValidator(datetime.strptime(min_time, "%H:%M").time())
+            )
+        if max_time := validate.get("maxTime"):
+            validators.append(
+                MaxValueValidator(datetime.strptime(max_time, "%H:%M").time())
+            )
+        base = serializers.TimeField(
+            required=required,
+            allow_null=not required,
+            validators=validators,
+        )
+        return serializers.ListField(child=base) if multiple else base
 
 
 @register("phoneNumber")
@@ -234,8 +253,25 @@ class File(BasePlugin[FileComponent]):
 
 
 @register("textarea")
-class TextArea(BasePlugin):
+class TextArea(BasePlugin[TextAreaComponent]):
     formatter = TextAreaFormatter
+
+    def build_serializer_field(
+        self, component: TextAreaComponent
+    ) -> serializers.CharField | serializers.ListField:
+        multiple = component.get("multiple", False)
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        # dynamically add in more kwargs based on the component configuration
+        extra = {}
+        if (max_length := validate.get("maxLength")) is not None:
+            extra["max_length"] = max_length
+
+        base = serializers.CharField(
+            required=required, allow_blank=not required, allow_null=False, **extra
+        )
+        return serializers.ListField(child=base) if multiple else base
 
 
 @register("number")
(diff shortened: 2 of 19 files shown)


src/openforms/formio/components/vanilla.py
```diff
@@ -306,11 +349,49 @@ def localize(self, component: SelectComponent, language_code: str, enabled: bool
...
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+        choices = [
+            (value["value"], value["label"]) for value in component["data"]["values"]
+        ]
+        return serializers.MultipleChoiceField(
+            choices=choices,
+            required=required,
+            allow_blank=not required,
+            allow_null=not required,
```
SilviaAmAm, line 364: I think the docs for Multiple choice field recommend to use one of allow_blank/allow_null. https://www.django-rest-framework.org/api-guide/fields/#multiplechoicefield
Since here the choices are text, maybe we should only use allow_blank?

src/openforms/formio/components/vanilla.py
```diff
@@ -306,11 +349,49 @@ def localize(self, component: SelectComponent, language_code: str, enabled: bool
...
+        self, component: CurrencyComponent
+    ) -> serializers.FloatField:
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        extra = {}
+        if max_value := validate.get("max"):
+            extra["max_value"] = max_value
+        if min_value := validate.get("min"):
+            extra["min_value"] = min_value
```
SilviaAmAm, line 382: I think that this will not have the expected result if `max_value` or `min_value` are 0?

src/openforms/formio/components/vanilla.py
```diff
@@ -348,9 +429,16 @@ def build_serializer_field(
...
-class Signature(BasePlugin):
+class Signature(BasePlugin[SignatureComponent]):
     formatter = SignatureFormatter
 
+    def build_serializer_field(
+        self, component: SignatureComponent
+    ) -> serializers.CharField:
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+        return serializers.CharField(required=required, allow_null=not required)
```
SilviaAmAm, line 440: I think `allow_blank` is preferred? https://www.django-rest-framework.org/api-guide/fields/#charfield

src/openforms/formio/tests/validation/test_currency.py
```diff
@@ -0,0 +1,92 @@
...
+            "validate": {
+                "required": False,
+                "min": 10.7,
+                "max": 15,
+            },
+        }
+
+        invalid_values = [
+            ({"foo": 9}, "min_value"),
+            ({"foo": 17}, "max_value"),
```
SilviaAmAm, line 61: I would add a test for when these are zero

src/openforms/formio/typing/vanilla.py
```diff
@@ -10,6 +10,38 @@ class TextFieldComponent(Component):
     pass
 
 
+class TextAreaComponent(Component):
```
SilviaAmAm, line 13: What is the advantage of having more specific types than `Component` if they don't add any extra fields?

vaszig: I saw above that we did that for the `TextField` component and I thought I should go in the same direction. I agree with you and I cannot think of any advantages apart from seeing a more specific type in the class definition.

src/openforms/formio/components/vanilla.py
```diff
@@ -150,9 +141,32 @@ def build_serializer_field(
...
+        self, component: Component
+    ) -> serializers.TimeField | serializers.ListField:
+        multiple = component.get("multiple", False)
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        validators = []
+        if min_time := validate.get("minTime"):
+            validators.append(
+                MinValueValidator(datetime.strptime(min_time, "%H:%M").time())
```
sergei-maertens, line 157: can't we use `.fromisoformat` instead?

```py
>>> from datetime import time
>>> time.fromisoformat("10:00")
datetime.time(10, 0)
>>> time.fromisoformat("10:00:03")
datetime.time(10, 0, 3)
```

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -0,0 +1,30 @@
...
+            "key": "foo",
+            "label": "Test",
+            "validate": {"required": True},
+        }
+
+        invalid_values = [
+            ({}, "required"),
+            ({"foo": ""}, "not_a_list"),
+            ({"foo": None}, "null"),
+        ]
```
sergei-maertens, line 21: I'd additionally check that `[<number>]` and `[<number1>, <number2>, <number3>]` are not accepted - you can specify min/max number of items on the serializer list field if I remember correctly.

src/openforms/formio/tests/validation/test_time.py
```diff
@@ -0,0 +1,73 @@
...
+                self.assertEqual(error.code, error_code)
+
+    def test_min_max_time(self):
+        component: Component = {
+            "type": "time",
+            "key": "foo",
+            "label": "Foo",
+            "validate": {
+                "required": False,
+                "minTime": "10:00",
```
sergei-maertens, line 40: probably to have a case with seconds specified too, for legacy components (the new builder outputs hours and minutes only, but the old one did more I think)

sergei-maertens (changes requested):
I'm not seeing the matching e2e tests yet? These are important to ensure that the frontend and backend behaviour are aligned: 

https://github.com/open-formulieren/open-forms/blob/master/src/openforms/tests/e2e/test_input_validation.py

src/openforms/tests/e2e/input_validation_base.py
```diff
@@ -104,16 +108,24 @@ async def _assertFrontendValidation(
             await page.goto(url)
             await page.get_by_role("button", name="Formulier starten").click()
 
-            # fill in the test input
-            await self._locate_input(page, label).fill(ui_input)
+            # fill in the test input in case of text otherwise continue
+            if input_type == "text":
+                await self._locate_input(page, label).fill(ui_input)
```
sergei-maertens, line 113: I'm not a fan of this, I think it's better to wrap this in its own private method that can be overridden on a test case basis so that you can implement the UI interaction specifically for the component type being tested. E.g. for a select, you would need to do some aria-role comobox shenanigans and for checkboxes it would be a `.click` or maybe some control to check/uncheck it (not sure what Playwright offers here).

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -187,6 +187,24 @@ def test_required_field(self):
...
+        self.assertValidationIsAligned(
+            component,
+            ui_input="1234",
+            expected_ui_error="Ongeldig BSN",
+        )
+        self.assertValidationIsAligned(
+            component,
+            ui_input="123456781",
+            expected_ui_error="Ongeldig BSN",
+        )
```
sergei-maertens, line 206: there should be only one `assertValidationIsAligned` call per test case, since it already runs two test cases and sets up all the test data under the hood. Probably best to break this up into two separate tests.

sergei-maertens (changes requested):

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,209 @@ def test_max_date_fixed_value(self):
             api_value="2025-01-01",
             expected_ui_error="De opgegeven datum ligt te ver in de toekomst.",
         )
+
+
+class SingleCheckboxTests(ValidationsTestCase):
+
+    async def apply_ui_input(self, page: Page, label: str, ui_input: str | int | float):
+        await expect(
+            page.get_by_role("checkbox", name="Required checkbox")
```
sergei-maertens, line 326:
```suggestion
            page.get_by_role("checkbox", name=label)
```
this avoids hardcoding it to the label of a single test

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,209 @@ def test_max_date_fixed_value(self):
...
+            component,
+            ui_input=50,
+            expected_ui_error="De waarde moet 15 of kleiner zijn.",
+        )
+
+
+class SingleMapTests(ValidationsTestCase):
+    async def apply_ui_input(self, page: Page, label: str, ui_input: str | int | float):
+        page.wait_for_selector(
+            ".openforms-leaflet-map, [aria-label='Required map']", state="visible"
```
sergei-maertens, line 392:
```suggestion
            f".openforms-leaflet-map, [aria-label='{label}']", state="visible"
```

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,209 @@ def test_max_date_fixed_value(self):
...
+        self.assertValidationIsAligned(
+            component,
+            ui_input=50,
+            expected_ui_error="De waarde moet 15 of kleiner zijn.",
+        )
+
+
+class SingleMapTests(ValidationsTestCase):
+    async def apply_ui_input(self, page: Page, label: str, ui_input: str | int | float):
+        page.wait_for_selector(
```
sergei-maertens, line 391: shouldn't this be awaited?

```suggestion
        await page.wait_for_selector(
```

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,209 @@ def test_max_date_fixed_value(self):
...
+            component,
+            ui_input="",
+            expected_ui_error="Het verplichte veld Required postcode is niet ingevuld.",
+        )
+
+
+class SingleSignatureTests(ValidationsTestCase):
+
+    async def apply_ui_input(self, page: Page, label: str, ui_input: str | int | float):
+        page.wait_for_selector("[aria-label='Required signature']", state="visible")
```
sergei-maertens, line 429: same remark about hardcoed label here :)

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,209 @@ def test_max_date_fixed_value(self):
...
+            "key": "requiredTextarea",
+            "label": "Required textarea",
+            "validate": {"maxLength": 10},
+        }
+        invalid_sample = "word" * 4
+
+        self.assertValidationIsAligned(
+            component,
+            ui_input=invalid_sample,
+            api_value=invalid_sample,
```
sergei-maertens, line 473:
```suggestion
            ui_input=invalid_sample,
```
if they're the same, don't specify `api_value` :) the test is more robust by using `ui_input` for both inputs (UI + API)

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,209 @@ def test_max_date_fixed_value(self):
...
+
+    def test_max_value(self):
+        component: Component = {
+            "type": "time",
+            "key": "requiredTime",
+            "label": "Required time",
+            "validate": {
+                "minTime": "10:00",
+                "maxTime": "12:00",
+            },
```
sergei-maertens, line 518: what's the behaviour when you swap min and max? The frontend has some special logic.

E.g. min time is 20:00 and max is 4:00 -> 15:00 is an invalid value.

src/openforms/formio/api/validators.py
```diff
@@ -158,3 +158,9 @@ def __call__(self, uploaded_file: UploadedFile) -> None:
                 raise serializers.ValidationError(
                     _("The virus scan returned an unexpected status.")
                 )
+
+
+class TruthyBooleanValueValidator:
+    def __call__(self, value):
+        if not value:
+            raise serializers.ValidationError(_("This field must be true."))
```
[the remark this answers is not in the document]
vaszig: It is required in my opinion since we want to make sure we have `True` value when it should be checked (I was only passing `required=required` which doesn't have anything to do with the real value), correct?
And I had a mistake in the tests, I hadn't provided the right value in order to trigger the validator.

src/openforms/tests/e2e/test_input_validation.py
vaszig: This test was failing when I was using `await` so I assumed that it is not needed, that's why I removed it from above as well (the one above though is passing in both ways).

src/openforms/formio/api/validators.py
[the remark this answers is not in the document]
sergei-maertens: I think this is already handled through the `required` for a `serializers.BooleanField`, that would explain why this code is not covered.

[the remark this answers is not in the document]
vaszig: But the test is failing without this validator and by only having `serializers.BooleanField(required=required)`. Am I missing something else?I thought that the `required` is only for the field to be there but not the explicit boolean value.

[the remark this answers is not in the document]
sergei-maertens: I will have to look into it, because the missing coverage seems to indicate this validator isn't running at all :thinking:

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,223 @@ def test_max_date_fixed_value(self):
...
+            component,
+            ui_input="",
+            expected_ui_error="Het verplichte veld Required postcode is niet ingevuld.",
+        )
+
+
+class SingleSignatureTests(ValidationsTestCase):
+
+    async def apply_ui_input(self, page: Page, label: str, ui_input: str | int | float):
+        page.wait_for_selector(f"[aria-label='{label}']", state="visible")
```
sergei-maertens, line 427: seems to also need the await: https://playwright.dev/python/docs/api/class-elementhandle#element-handle-wait-for-selector

```suggestion
        await page.wait_for_selector(f"[aria-label='{label}']", state="visible")
```

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -292,3 +317,223 @@ def test_max_date_fixed_value(self):
...
+            component,
+            ui_input="",
+            expected_ui_error="Het verplichte veld Required time is niet ingevuld.",
+        )
+
+    def test_min_value(self):
+        component: Component = {
+            "type": "time",
+            "key": "requiredTime",
+            "label": "Required time",
```
sergei-maertens, line 494: it's nitpicking, but can you make the `key` and `label` have values that reflect back to the test itself? This component is not about the "required" validation, but about the min value. :)

src/openforms/formio/components/vanilla.py
```diff
@@ -150,9 +142,28 @@ def build_serializer_field(
...
+    ) -> serializers.TimeField | serializers.ListField:
+        multiple = component.get("multiple", False)
+        validate = component.get("validate", {})
+        required = validate.get("required", False)
+
+        validators = []
+        if min_time := validate.get("minTime"):
+            validators.append(MinValueValidator(time.fromisoformat(min_time)))
+        if max_time := validate.get("maxTime"):
+            validators.append(MaxValueValidator(time.fromisoformat(max_time)))
```
sergei-maertens, line 159: I don't really see how the `minTime: 20:00` and `maxTime: 4:00` are handled here - it doesn't look like it'll be valid for a value like `02:00` - can you add some unit tests too? Rather than only the end-to-end tests.

src/openforms/tests/e2e/test_input_validation.py
sergei-maertens: I checked out your branch locally and am running the tests and as expected, there is a warning :)

```
src/openforms/tests/e2e/test_input_validation.py:427: RuntimeWarning: coroutine 'Page.wait_for_selector' was never awaited
  page.wait_for_selector(f"[aria-label='{label}']", state="visible")
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

sergei-maertens:
I tested some more complex forms that revealed some validation issues before, and as far as I can tell this doesn't introduce new problems yet, but not all component types were used in there :grimacing:

src/openforms/formio/components/vanilla.py
sergei-maertens: I took care of it :)

sergei-maertens (approved):
@vaszig do you mind sanity checking my changes?

src/openforms/formio/tests/validation/test_time.py
```diff
@@ -73,3 +151,72 @@ def test_multiple(self):
...
+            "label": "Misconfigured min/max time",
+            "validate": {
+                "minTime": True,  # type:ignore
+                "maxTime": 42,  # type:ignore
+            },
+        }
+
+        is_valid, _ = validate_formio_data(component, {"time": "15:36"})
+
+        self.assertTrue(is_valid)
```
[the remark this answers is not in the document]
sergei-maertens: coverage reporting complained about missing branch coverage on the `match` statement, so this is what it addresses - it happens that invalid configuration ends up in the JSON and this ensures it is logged/covered by tests :)

Issue #4019, opened 2024-03-18
Objects API "contract" feature - release blockers


sergei-maertens:
Follow up from #3688, these aspects should land in the 2.6 release

- [x] Handling of file fields - a file component (single or multiple uploads) should be possible to explicitly map to a particular (nested) property in the object type definition. Currently all upload fields are processed and result in an array of Documenten API urls, and it's not possible to match which component/file upload matches to which URL. We have the means now to work with structured data, this would be an unfortunate omission.
- [x] Document how one would use a "productaanvraag" user-defined variable and map that into the structure. This facilitates migrating from v1 to v2
- [x] Check if we can replace the payment status PATCH call with a PUT call instead (payment related variables need to be mappable)

Viicos:
For the first checkbox:

- File upload components will be special cased during registration. This part of the code should be adapted:

https://github.com/open-formulieren/open-forms/blob/c963f22309102360ede0799b6473a8d5ac0d49db/src/openforms/registrations/contrib/objects_api/submission_registration.py#L333-L350

- Because we only offer compatible target paths to the user (e.g. when configuring the mapping for a `textfield` component, only the JSON Schema targets of `{"type": "string"}` will be offered in the UI), we have to special case this compatibility logic:
  - the `fileupload` component is of datatype `array`, but when registered, we will deal with the Documents URL (of datatype `string`). The admin frontend is the one providing a JSON Schema matching a variable, and requests the backend for a list of target paths compatible with such a JSON Schema. So this code will have to be adapted:
https://github.com/open-formulieren/open-forms/blob/1bbaeed3bfa5e239bf087ffc3641518c27a76ed4/src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js#L49-L60
  - Maybe the variables tab should display `datatype: string` instead of `datatype: array` in the `datatype` column? To make things less confusing

PR #4032, opened 2024-03-20
[#3688] Use Documents URLs as fileupload variables values

Viicos:
Part of #3688, #4019

### DIFF
--- a/src/openforms/submissions/tests/test_variables/test_num_queries.py
+++ b/src/openforms/submissions/tests/test_variables/test_num_queries.py
@@ -125,10 +125,7 @@ class SubmissionVariablesPerformanceTests(APITestCase):
         # 4.  Retrieve the submission variables to be deleted - deletion of data happens
         #     because the step is marked N/A
         # 5.  Retrieve the submission attachment files to be deleted
-        # 6.  SAVEPOINT
-        # 7.  Delete submission attachment files
-        # 8.  RELEASE SAVEPOINT
-        # 9.  Delete submission values
+        # 6.  Delete submission values
         with self.assertNumQueries(6):
             evaluate_form_logic(submission, submission_step2, data)
 


src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -221,12 +218,19 @@ def save_registration_data(
...
+                        ObjectsAPISubmissionAttachment.objects.get(
+                            submission_file_attachment=attachment
+                        )
+                    except ObjectsAPISubmissionAttachment.DoesNotExist:
+                        ObjectsAPISubmissionAttachment.objects.create(
+                            submission_file_attachment=attachment,
+                            documents_url=register_submission_attachment(
+                                submission, attachment, options, documents_client
+                            ),
+                        )
```
SilviaAmAm, line 233: If I understand correctly, this prevents that if the Objects API registration fails, the attachments are sent again to the Documents API during the retry flow?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -339,8 +349,23 @@ class ObjectsAPIV2Handler(ObjectsAPIRegistrationHandler[RegistrationOptionsV2]):
...
+            objects_api_attachments = ObjectsAPISubmissionAttachment.objects.filter(
+                submission_file_attachment__submission_variable__submission=submission,
+                submission_file_attachment__submission_variable__key=key,
+            )
+            if objects_api_attachments.exists():
+                urls = list(
+                    objects_api_attachments.values_list("documents_url", flat=True)
+                )
+                dynamic_values[key] = urls[0] if len(urls) == 1 else urls
+
```
SilviaAmAm, line 368: Would it be better to get all the attachments for the submission, annotate them with the variable key, and then make them into a dictionary of key: attachment. Then iterate over the attachment and with the key update the corresponding variables? I think it might save queries for the variables that are not files?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -221,12 +218,19 @@ def save_registration_data(
...
-                # just urls.
-                registration_data.attachment_urls = register_submission_attachments(
-                    submission, options, documents_client
-                )
+            if options["informatieobjecttype_attachment"]:
+                for attachment in submission.attachments:
+                    try:
+                        ObjectsAPISubmissionAttachment.objects.get(
+                            submission_file_attachment=attachment
+                        )
```
sergei-maertens, line 226: for performance, you probably want to build up a list of URLs:

```py
existing = {o.submission_file_attachment: o for o in ObjectsAPISubmissionAttachment.objects.filter(submission_file_attachment__submission=submission)}
for attachment in submission.attachments:
    if attachment in existing:
        continue
    ...
```

src/openforms/registrations/contrib/objects_api/submission_registration.py
sergei-maertens: let's optimize this indeed, a `defaultdict(list)` should work well for this. You also need to keep track of the `component["multiple"]` and not determine the type based on the amount of files, since you can have a multiple-allowed field with just the single upload, which will be mapped to an array in the json schema. The current code would write a string instead of an array.

This is also a test case that should be added :-)

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -222,12 +221,19 @@ class ObjectsAPIRegistrationData(models.Model):
...
+        "submissions.SubmissionFileAttachment",
+        on_delete=models.CASCADE,
+        verbose_name=_("submission file attachment"),
+        help_text=_("The submission file attachment."),
+    )
+
+    documents_url = models.URLField(
+        _("documents_url"),
+        help_text=_("The URL of the submission registered on the Documents API."),
     )
```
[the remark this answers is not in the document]
Viicos: By `documents` I meant the _Documents API_, but indeed this can be confusing

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -221,12 +221,28 @@ def save_registration_data(
...
+                for attachment in submission.attachments:
+                    if attachment not in existing:
+                        objs.append(
+                            ObjectsAPISubmissionAttachment(
+                                submission_file_attachment=attachment,
+                                document_url=register_submission_attachment(
+                                    submission, attachment, options, documents_client
+                                ),
+                            )
+                        )
```
sergei-maertens, line 243: (please create a follow up issue for this)

This can still cause (repeated) failures - if you have two documents and the first succeeds, while the second does not, the for-loop will crash out and it will never be persisted that the first document creation actually succeeded.

you'll probably want to `try:` around this to make sure that you can create the model instances for whatever's in `objs` in the `finally` block:

```py
try:
    for attachmenet in ...:
        # do the stuff
except:
    raise
finally:
    ObjectsAPISubmissionAttachment.objects.bulk_create(objs)
```

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -339,8 +361,34 @@ class ObjectsAPIV2Handler(ObjectsAPIRegistrationHandler[RegistrationOptionsV2]):
...
+            if key in urls_map:
+                variable = state.get_variable(key)
+                is_multiple = variable.form_variable.form_definition.configuration_wrapper.component_map[
+                    key
+                ].get(
+                    "multiple", False
+                )
+                dynamic_values[key] = (
+                    urls_map[key][0] if not is_multiple else urls_map[key]
+                )
```
sergei-maertens, line 390: I did some reasoning about the `form_variable` FK potentially being `None` and how this works if registration is run after a file-upload component is removed from the form, and I concluded it's not an issue, because:

* `state` is used, which ensures that we only consider currently available form variables (through the `collect_variables` method)
* `dynamic_values` is used, which relies on the above state considering only currently available form variables

So there may be extra `SubmissionFileAttachment` records for the submission, and those will exist in our DB but not sent along to the objects API. And this behaviour is as (currently) intended.

src/openforms/submissions/tests/test_variables/test_num_queries.py
```diff
@@ -129,7 +129,7 @@ def test_evaluate_form_logic_with_rules(self):
         # 7.  Delete submission attachment files
         # 8.  RELEASE SAVEPOINT
         # 9.  Delete submission values
-        with self.assertNumQueries(9):
+        with self.assertNumQueries(6):
```
sergei-maertens, line 132: where's the explanation update of the queries being executed? The description above describes 9 queries, the assertion says 6.

src/openforms/registrations/contrib/objects_api/submission_registration.py
sergei-maertens: Created #4041

sergei-maertens (approved):

PR #4036, opened 2024-03-21
[#4019] Mapping file uploads in Object API registration (frontend)

### LINKED ISSUE #4019: Objects API "contract" feature - release blockers

SilviaAmAm:
Fixes #4019  (partially)

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
@@ -48,9 +48,21 @@ const FORMAT_TYPE_MAP = {
 /**
  * Return a JSON Schema definition matching the provided variable.
  * @param {Object} variable - The current variable
+ * @param {Object} components - The components available in the form. The key is the component key, the value is the
+ * component definition.
  * @returns {Object} - The JSON Schema
  */
-const asJsonSchema = variable => {
+const asJsonSchema = (variable, components) => {
+  // Figure out if the component is a file component (special case)
+  const componentDefinition = components[variable.key];
+  if (componentDefinition && componentDefinition.type === 'file') {
+    // If it is, and it has multiple == True, then type is array
+    if (componentDefinition.multiple)
+      return {type: 'array', items: {type: 'string', format: 'uri'}};
+    // Otherwise it's string (URL of the document)
+    return {type: 'string', format: 'uri'};
+  }
+
   if (VARIABLE_TYPE_MAP.hasOwnProperty(variable.dataType))
     return {type: VARIABLE_TYPE_MAP[variable.dataType]};
   return {
--- a/src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
+++ b/src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
@@ -1,5 +1,6 @@
 import {expect} from '@storybook/jest';
-import {userEvent, within} from '@storybook/testing-library';
+import {userEvent, waitFor, within} from '@storybook/testing-library';
+import {screen} from '@storybook/testing-library';
 
 import {BACKEND_OPTIONS_FORMS} from 'components/admin/form_design/registrations';
 import {mockTargetPathsPost} from 'components/admin/form_design/registrations/objectsapi/mocks';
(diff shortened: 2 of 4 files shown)


src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
```diff
@@ -48,9 +48,20 @@ const FORMAT_TYPE_MAP = {
...
-const asJsonSchema = variable => {
+const asJsonSchema = (variable, components) => {
+  // Figure out if the component is a file component (special case)
+  const componentDefinition = components[variable.key];
+  if (componentDefinition && componentDefinition.type === 'file') {
+    // If it is, and it has multiple == True, then type is array
+    if (componentDefinition.multiple) return {type: 'array'};
+    // Otherwise it's string (URL of the document)
+    return {type: 'string', format: 'uri'};
+  }
```
SilviaAmAm, line 63: I left this as a 'special case'. I thought that if in the future we have more special cases we can refactor this.

Viicos: I'm wondering if this special case should be moved in the `ObjectsApiVariableConfigurationEditor` component instead :thinking:, as it is purely hardcoding a json schema.

Or maybe instead of passing the array of `components`, only pass the relevant `component` to this function?

Viicos (approved):

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
sergei-maertens: more accurate would also be:

```yaml
type: array
items:
  type: string
  format: uri
```

sergei-maertens: I think leaving it in here is more appropriate, it fits the idea of providing the json schema for a given variable. The fact that the schema is hard coded doesn't really matter, it's the same how your fallback is a hardcoded `type: string`, no?

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
```diff
@@ -181,18 +228,22 @@ export const WithObjectsAPIRegistrationBackends = {
...
-            isRequired: true,
-            jsonSchema: {type: 'string'},
-          },
-          {
-            targetPath: ['other', 'path'],
-            isRequired: false,
-            jsonSchema: {type: 'object', properties: {a: {type: 'string'}}, required: ['a']},
-          },
-        ]),
+        mockTargetPathsPost({
```
sergei-maertens, line 231: @Viicos I can't really judge if this still accurately reflects what the real API endpoint does, can you confirm this is okay?

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
```diff
@@ -211,6 +262,117 @@ export const WithObjectsAPIRegistrationBackends = {
...
+    const dropdown = await screen.findByRole('combobox');
+
+    await expect(dropdown).toBeInTheDocument();
+
+    await waitFor(async () => {
+      // Only the targets of type array should appear
+      await expect(
+        screen.getByRole('option', {name: 'path > to > array (required)'})
+      ).toBeVisible();
+    });
```
sergei-maertens, line 372: can't you do:

```suggestion
    // Only the targets of type array should appear
    expect(await screen.findByRole('option', {name: 'path > to > array (required)'})).toBeVisible();
```
`findBy*` is basically `await waitFor( getBy* )`

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
SilviaAmAm: It should, I also updated the `mockTargetPathsPost` so that it returns the schema for the requested type (but the structure of the returned data is unchanged).

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
SilviaAmAm: I will try again though, maybe I did something else wrong when I tried

SilviaAmAm: I had forgot the `await` inside the expect :facepalm: I'm a silly

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
sergei-maertens: the old code returned an array of two items, while the new variant grabs only a single item based on the request body. I think the new code is more correct in how the backend actually works, but I'm not well enough aware of the implementation details to be certain :grimacing:

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
sergei-maertens: yeah, the await for `find*` is needed, the `await` before `expect` isn't :)

async is hard!

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
Viicos: Yeah it is reflecting the backend better, as the available target paths depend on the JSON Schema type of the variable :)

PR #4042, opened 2024-03-22
[#3688] Fix payment updates: use PATCH, represent payment as float

Viicos:
Part of #4019, #3688


src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -65,7 +65,7 @@ class PaymentAmount(BaseStaticVariable):
     def get_initial_value(self, submission: Submission | None = None):
         if submission is None:
             return None
-        return str(submission.payments.sum_amount())
+        return float(submission.payments.sum_amount())
```
Viicos, line 68: This seems to be the [documented way](https://docs.python.org/3/library/decimal.html#quick-start-tutorial)
image
:

sergei-maertens (approved):

PR #4043, opened 2024-03-22
[#3688] Remove `productaanvraagType` for v2 options

Viicos:
Part of #3688

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
@@ -47,6 +47,7 @@ const ObjectsApiOptionsFormFields = ({index, name, schema, formData, onChange}) 
         } else {
           delete draft.variablesMapping;
         }
+        delete draft.productaanvraagType;
         delete draft.contentJson;
         delete draft.paymentStatusUpdateJson;
       })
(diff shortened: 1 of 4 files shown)


src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -47,6 +47,7 @@ const ObjectsApiOptionsFormFields = ({index, name, schema, formData, onChange})
         } else {
           delete draft.variablesMapping;
         }
+        delete draft.productaanvraagType;
```
SilviaAmAm, line 50: Maybe nice to add a check in the story that this field is not present?

SilviaAmAm (approved):

vaszig (commented):
Looks good. I think we missed only the removal of `productaanvraagType` in the definition (`V2ConfigFields.propTypes`)

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
sergei-maertens: Those kind of checks are a bit fragile, you can easily satisfy them by changing a translation to something slightly different. I trust the review process more on this :)

sergei-maertens (approved):
Processed the feedback (proptype removal)

PR #4046, opened 2024-03-22
Handle object types with no geometry allowed

Viicos:
I will probably close this PR at some point, as I think it's better to handle it as sergei said:

> instead of handling the mapping as everything inside `data` and special casing `geometry`, can't we just implicitly add the `data` path prefix to everything so then you get a list of targets:
> - `geometry`
> - `data.firstName`
> - `data.lastName`
> - `data.country`
> but the end user never needs to worry about the `data.` prefix
> (of course keeping the types in mind here)
> the reason I'm asking this is also because you can have a second map in the form which goes to `data.additionalGeometry`, and that thing also needs to be pre-processed to produce geo-json

With this solution in my mind, we would just have to remove `geometry` if `allowGeometry` is `False` in the objecttypes API

### DIFF
--- a/src/openforms/contrib/objects_api/clients/objecttypes.py
+++ b/src/openforms/contrib/objects_api/clients/objecttypes.py
@@ -43,6 +43,14 @@ class ObjecttypesClient(NLXClient):
             page_size=page_size,
         )
 
+    def get_objecttype(
+        self,
+        objecttype_uuid: str | UUID,
+    ) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


Viicos:
https://github.com/open-formulieren/open-forms/pull/4102 is an attempt to do so, but I'm closing it as it turns out the solution is to hard to implement

joeribekker:
Refinement: @Viicos please make a ticket for this with traige label

joeribekker:
Refinement: Closing since its a very edge case and no stakeholder.

Issue #4031, opened 2024-03-20
Objects API "contract" feature - improvements


Viicos:
Follow up of #3688:

- [ ] Backend validation of the variables mapping (Phase 4). Requires the refactor (see https://github.com/open-formulieren/open-forms/issues/3941) to land first.
- [ ] Display backend validation errors back to the frontend:
  - For specific variables, error should be displayed in the variables tab
  - An error should also be displayed if required targets are not mapped. The issue is we are constrained by the DRF error format (strings), and this probably requires a more complex structure.
- [ ] `target-paths/` endpoint: handle connection errors to the OT API. Currently returns 500.
- [ ] Display a message when no target path is available? Currently only displays the ------ empty option, which can be confusing.
- [ ] Handle object types where `allowGeometry` is false
- [ ] To figure out once we get more feedback: the order in which mappings are applied matters (e.g. you can map an object to `path > to`, and then a string to `path > to > otherpath`)
- [ ] Use React select for the target-path dropdown
- [ ] Add a popup when switching from v1 to v2, saying all the configured variable mappings will be deleted.

PR #4054, opened 2024-03-25
[#4031] Add a confirm popup when switching to v2 options

Viicos:
Part of #4031.

This was discussed at some point but forgot about it


src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsFormFields.js
```diff
@@ -33,6 +40,12 @@ const ObjectsApiOptionsFormFields = ({index, name, schema, formData, onChange})
 
   const changeVersion = v => {
     const realVersion = v + 1;
+
+    if (realVersion === 1) {
+      const confirmV1Switch = window.confirm(v1SwitchMessage);
+      if (!confirmV1Switch) return;
+    }
```
Viicos, line 47: I could also not show the popup if `variablesMapping` is empty, but probably not necessary

sergei-maertens (approved):

PR #4058, opened 2024-03-25
[#4044] Improve dev documentation regarding forms and submissions

### LINKED ISSUE #4044: Improve developer documentation of submission state and component configuration
While working on the Objects API registration, I sometimes found it hard to understand how the submission state should be loaded. Here's a couple things I noted:
- The `SubmissionValueVariablesState` (available through `submission.load_submission_value_variables_state`) has 11 public methods/attributes, making it hard to find out the right one to use. Naming is sometimes ambiguous (e.g. `get_data`). None of them have docstrings.
- There seem to be a distinction between "saved variables" (with a PK) and variables not persisted, but it is unclear what is the use case for this.
- The `SubmissionValueVariable` name is confusing: it should probably be `SubmissionVariableValue` instead. However not sure if it is possible to rename it.

---

I had the same trouble with components configuration:
- The recommended way to access the component configuration is through the `FormioConfigurationWrapper` class, but is also missing docstrings.
- While I understand a component is following the `{"type": "...", "key": "..."}` structure, it is still unclear to me how that plays with repeating groups and/or nested components. This relates to the `FormioData` structure.

For this, I would really like having some kind of introduction in the docs to the formio structure, how it is integrated in Open forms and how we manage to extend it.

---

I would also really like having some kind of "cheat sheet" in docs, stating the recommended ways of dealing with the submission state, the formio configuration wrapper, etc

Viicos:
Fixes #4044

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,8 +40,6 @@ extensions = [
     "sphinx.ext.todo",
     "sphinx.ext.intersphinx",
     "sphinx_tabs.tabs",
-    "recommonmark",
-    # "sphinx_markdown_tables",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,6 +40,7 @@ extensions = [
     "sphinx.ext.todo",
     "sphinx.ext.intersphinx",
     "sphinx_tabs.tabs",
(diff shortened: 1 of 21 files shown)


Viicos:
Replaced by https://github.com/open-formulieren/open-forms/pull/4091, without the submission state changes.

PR #4062, opened 2024-03-26
[#72] Remaining backend validation of form rules

vaszig:


vaszig:
Wrong base by accident, will not be merged.

PR #4078, opened 2024-03-28
Fix another validation mismatch

sergei-maertens:
Reported by Bob, thanks :pray:

### DIFF
--- a/src/openforms/formio/registry.py
+++ b/src/openforms/formio/registry.py
@@ -91,7 +91,10 @@ class BasePlugin(Generic[ComponentT], AbstractBasePlugin):
             DeprecationWarning,
         )
 
-        if is_layout_component(component):
+        # not considered a layout component (because it doesn't have children)
+        if component["type"] == "content":
+            required = False
+        elif is_layout_component(component):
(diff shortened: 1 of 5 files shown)


src/openforms/formio/registry.py
```diff
@@ -91,7 +91,10 @@ def build_serializer_field(self, component: ComponentT) -> serializers.Field:
             DeprecationWarning,
         )
 
-        if is_layout_component(component):
+        # not considered a layout component (because it doesn't have children)
+        if component["type"] == "content":
+            required = False
```
Viicos, line 96: Probably at some point if we manage to define components as dataclass-likes structure, this will be easier to handle (via a method/property)

Viicos (approved):

src/openforms/formio/registry.py
sergei-maertens: yeah, this patch definitely had me longing for such a thing again :)

sergei-maertens:
Backported with d466b69504997f7347646292ea9bac99766f2384 and create a new tag `2.6.2-beta.1`

PR #4085, opened 2024-04-02
:mute: Mute logging during test runs in CI

sergei-maertens:
CI is showing irrelevant logs, making it harder to debug actual test failures.


vaszig (approved):
That sounds really nice! I don't know though if there are any tests that it could be useful to have the logs as well in CI. I guess we can run them locally in order to debug.

PR #4086, opened 2024-04-02
[#3889] Improve timeline logs admin

### LINKED ISSUE #3889: As Privacy/Security Officer I want to see and export log entries in a certain period
### Thema / Theme

Admin

* [ ] Make AVG-timelinelogger admin view faster (it now takes forever to load)
* [x] Allow filter on period (i think there's an admin filter lib for this)
* [x] Allow filter on user
* [x] Log views of the request-log entries
* [x] Allow an export of the AVG-timelinelogger entries

As part of NIS2 (security norm), this becomes a requirement and thus more used feature of Open Formulieren.

_No response_

Viicos:
Fixes #3889

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -168,7 +168,7 @@ django-formtools==2.5.1
     # via django-two-factor-auth
 django-hijack==3.4.1
     # via -r requirements/base.in
-django-import-export==2.9.0
+django-import-export==3.3.7
     # via -r requirements/base.in
 django-ipware==5.0.0
     # via django-axes
@@ -478,7 +478,7 @@ sqlparse==0.4.4
     # via django
 stringcase==1.2.0
     # via o365
-tablib[html,ods,xls,xlsx,yaml]==3.0.0
+tablib[html,ods,xls,xlsx,yaml]==3.5.0
     # via
     #   -r requirements/base.in
     #   django-import-export
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -88,7 +88,11 @@ cryptography==42.0.4
 cssselect2==0.7.0
     # via weasyprint
 defusedxml==0.7.1
-    # via -r requirements/base.in
+    # via
+    #   -r requirements/base.in
+    #   odfpy
+diff-match-patch==20230430
+    # via django-import-export
 django==4.2.11
     # via
     #   -r requirements/base.in
@@ -104,6 +108,7 @@ django==4.2.11
     #   django-filter
     #   django-formtools
     #   django-hijack
+    #   django-import-export
     #   django-jsonform
     #   django-log-outgoing-requests
     #   django-modeltranslation
@@ -162,6 +167,8 @@ django-filter==23.2
 django-formtools==2.5.1
     # via django-two-factor-auth
 django-hijack==3.4.1
+    # via -r requirements/base.in
+django-import-export==2.9.0
     # via -r requirements/base.in
 django-ipware==5.0.0
     # via django-axes
@@ -310,6 +317,8 @@ mail-cleaner==1.2.0
     # via -r requirements/base.in
 mail-parser==3.15.0
     # via django-yubin
+markuppy==1.14
+    # via tablib
 maykin-2fa==1.0.0
     # via -r requirements/base.in
 maykin-json-logic-py==0.13.0
@@ -324,6 +333,8 @@ o365==2.0.31
     # via -r requirements/base.in
 oauthlib==3.2.2
     # via requests-oauthlib
+odfpy==1.4.1
+    # via tablib
 openapi-schema-validator==0.6.2
     # via openapi-spec-validator
 openapi-spec-validator==0.6.0
@@ -395,6 +406,7 @@ pyyaml==6.0.1
     # via
     #   drf-spectacular
     #   jsonschema-spec
+    #   tablib
 qrcode==7.4.2
     # via django-two-factor-auth
 redis==4.5.4
@@ -466,8 +478,10 @@ sqlparse==0.4.4
     # via django
 stringcase==1.2.0
     # via o365
-tablib[xlsx]==3.0.0
-    # via -r requirements/base.in
+tablib[html,ods,xls,xlsx,yaml]==3.0.0
+    # via
+    #   -r requirements/base.in
+    #   django-import-export
 tabulate==0.8.9
     # via -r requirements/base.in
 tinycss2==1.1.0
@@ -516,6 +530,10 @@ webencodings==0.5.1
     #   tinycss2
 wrapt==1.15.0
     # via elastic-apm
+xlrd==2.0.1
+    # via tablib
+xlwt==1.3.0
+    # via tablib
 xmlsec==1.3.10
     # via maykin-python3-saml
 xmltodict==0.12.0
--- a/src/openforms/logging/admin.py
+++ b/src/openforms/logging/admin.py
@@ -12,6 +12,7 @@ class TimelineLogProxyResource(resources.ModelResource):
     user = Field(attribute="user")
     related_object = Field(attribute="content_object")
     message = Field(attribute="message")
+    event = Field(attribute="extra_data")
 
     class Meta:
         model = TimelineLogProxy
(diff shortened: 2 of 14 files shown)


requirements/base.txt
```diff
@@ -163,6 +168,8 @@ django-formtools==2.5.1
     # via django-two-factor-auth
 django-hijack==3.4.1
     # via -r requirements/base.in
+django-import-export==2.9.0
```
sergei-maertens, line 171: why 2.9.0 and not 3.3.x?

src/openforms/logging/admin.py
```diff
@@ -30,6 +75,19 @@ def has_change_permission(self, request, obj=None):
...
+        request: HttpRequest,
+        object_id: str,
+        form_url: str = "",
+        extra_context: dict[str, Any] | None = None,
+    ) -> HttpResponse:
+        timelinelog = self.get_object(request, object_id)
+        if timelinelog is None:
+            raise Http404(f"No {self.model._meta.object_name} matches the given query.")
+        timelinelog_details_view_admin(timelinelog, request.user)
+        return super().change_view(request, object_id, form_url, extra_context)
```
sergei-maertens, line 89: that's not what the ticket specified - the ticket specified viewing request logs from django-log-outgoing-requests

Viicos: I'm not sure this is the case. Viewing these outgoing requests logs is already available at `http://localhost:8000/admin/log_outgoing_requests/outgoingrequestslog/`.

Joeri clarified this this morning

requirements/base.txt
Viicos: I guess I need to pass more dependencies to `-P`

src/openforms/logging/models.py
```diff
@@ -126,7 +126,7 @@ def fmt_url(self) -> str:
     def content_admin_url(self) -> str:
         if not (self.object_id and self.content_type_id):
             return ""
-
+        breakpoint()
```
sergei-maertens, line 129: please remove this and ensure there is sufficient test coverage :)

src/openforms/logging/templates/logging/events/outgoing_request_log_details_view_admin.txt
```diff
@@ -0,0 +1,4 @@
+{% load i18n %}
+{% blocktrans trimmed with lead=log.fmt_lead user=log.fmt_user timelinelog=log.pk %}
+   {{ lead }}: User {{ user }} viewed outgoing request log {{ timelinelog }} in the admin
```
sergei-maertens, line 3: seeing a primary key of a log record that is periodically pruned is not very useful :)

Please include:

* host
* path
* http method

of the request as the representation of the log record.

src/openforms/logging/tests/test_admin.py
```diff
@@ -50,7 +54,14 @@ def test_superuser_cant_delete_logs_with_bulk_action(self):
...
         # It is not possible to select the "delete_selected" option
-        self.assertIsNone(html_form.fields.get("action"))
+        self.assertEqual(
+            [
+                option
+                for option in html_form.fields.get("action")[0].options
+                if option[0] == "delete_selected"
+            ],
+            [],
+        )
```
sergei-maertens, line 64: can't we just assert what *is* present?

```py
options = {value for value, *_ in html_form.fields.get("action")[0].options}
self.assertEqual(options, {"_export"})  # or whatever the relevant value is
```

src/openforms/logging/tests/test_admin.py
```diff
@@ -85,3 +96,46 @@ def test_deleted_submission_doesnt_crash_logs(self):
...
+        user = StaffUserFactory.create()
+        bare_log = TimelineLogProxyFactory.create(user=user)
+
+        dataset = TimelineLogProxyResource().export()
+
+        self.assertEqual(
+            json.loads(dataset.json),
+            [
+                {
+                    "message": bare_log.message().strip(),
```
sergei-maertens, line 112: it's weird to assert the exact same code as the implementation. I much prefer setting up the actual test data in the factory and asserting the produced content. Tests need to cover the outcome, not the implementation.

src/openforms/utils/admin.py
```diff
@@ -49,3 +58,22 @@ def json_as_html(self, instance):
...
+    def change_view(
+        self,
+        request: HttpRequest,
+        object_id: str,
+        form_url: str = "",
+        extra_context: dict[str, Any] | None = None,
+    ) -> HttpResponse:
+        outgoing_request_log = self.get_object(request, object_id)
+        if outgoing_request_log is None:
+            raise Http404(f"No {self.model._meta.object_name} matches the given query.")
```
sergei-maertens, line 73: missing coverage

sergei-maertens (changes requested):

src/openforms/logging/tests/test_admin.py
Viicos: This test is originally to assert that you can't delete records. With your assertion suggestion the test would break whenever a new action gets added? Probably not going happen anyway :thinking:

sergei-maertens: yeah and if the value of the action option in django changes, your version will break and be harder to debug, while this proposed change is a lot more readable and obvious

the previous test assertion worked because it checked that the action dropdown as a whole is absent :)

src/openforms/utils/admin.py
sergei-maertens: that doesn't absolve from testing bad ID lookups :) a decorator is an implementation detail

src/openforms/logging/templates/logging/events/outgoing_request_log_details_view_admin.txt
```diff
@@ -0,0 +1,4 @@
+{% load i18n %}
+{% blocktrans trimmed with lead=log.fmt_lead user=log.fmt_user requestlog=log.content_object %}
+   {{ lead }}: User {{ user }} viewed outgoing request log {{ requestlog.method }} {{ requestlog.url }} ({{ requestlog.status_code }}) in the admin
```
sergei-maertens, line 3: status code is not needed

sergei-maertens (changes requested):

src/openforms/logging/templates/logging/events/outgoing_request_log_details_view_admin.txt
```diff
@@ -0,0 +1,4 @@
+{% load i18n %}
+{% blocktrans trimmed with lead=log.fmt_lead user=log.fmt_user requestlog=log.content_object %}
+   {{ lead }}: User {{ user }} viewed outgoing request log {{ requestlog.method }} {{ requestlog.url }} in the admin
```
sergei-maertens, line 3: this syntax is broken - try it out in your local environment

you can't use lookups inside `blocktrans`

src/openforms/logging/tests/test_admin.py
sergei-maertens: why is the `bare_log.message().strip()` still here instead of a literal string?

src/openforms/logging/tests/test_admin.py
```diff
@@ -85,3 +95,59 @@ def test_deleted_submission_doesnt_crash_logs(self):
...
+        log.timestamp = datetime(2024, 1, 1, tzinfo=dt_timezone.utc)
+        log.save()
+
+        dataset = TimelineLogProxyResource().export()
+
+        self.assertEqual(
+            json.loads(dataset.json),
+            [
+                {
+                    "message": log.message().strip(),
```
sergei-maertens, line 139: please change to a string literal.

src/openforms/utils/tests/test_admin.py
```diff
@@ -0,0 +1,45 @@
...
+            ),
+            user=user,
+        )
+
+        self.assertEqual(
+            TimelineLogProxy.objects.filter(
+                template="logging/events/outgoing_request_log_details_view_admin.txt"
+            ).count(),
+            1,
+        )
```
sergei-maertens, line 45: I am missing an assertion on the content of the log message

sergei-maertens (changes requested):

PR #4093, opened 2024-04-02
[#3725] Add problems to the email digest

### LINKED ISSUE #3725: Detect more problems and add them to the email digest

    - logs

vaszig:
Partly closes #3725 

This PR handles only the first checkbox of the task, regarding the failed submissions.

### DIFF
--- a/src/openforms/emails/tasks.py
+++ b/src/openforms/emails/tasks.py
@@ -18,10 +18,10 @@ def send_email_digest() -> None:
     if not (recipients := config.recipients_email_digest):
         return
 
-    desired_period = timezone.now() - timedelta(days=1)
+    yesterday = timezone.now() - timedelta(days=1)
 
-    failed_emails = collect_failed_emails(desired_period)
-    failed_registrations = collect_registrations_failures(desired_period)
+    failed_emails = collect_failed_emails(yesterday)
+    failed_registrations = collect_registrations_failures(yesterday)
 
     if not (failed_emails or failed_registrations):
         return
--- a/src/openforms/emails/tasks.py
+++ b/src/openforms/emails/tasks.py
@@ -5,13 +5,11 @@ from django.template.loader import render_to_string
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
-from django_yubin.models import Message
-
 from openforms.celery import app
 from openforms.config.models import GlobalConfiguration
-from openforms.logging.models import TimelineLogProxy
+from openforms.registrations.utils import collect_registrations_failures
 
-from .utils import send_mail_html
+from .utils import collect_failed_emails, send_mail_html
 
 
 @app.task
@@ -20,32 +18,24 @@ def send_email_digest() -> None:
     if not (recipients := config.recipients_email_digest):
         return
 
-    period_start = timezone.now() - timedelta(days=1)
+    desired_period = timezone.now() - timedelta(days=1)
 
-    logs = TimelineLogProxy.objects.filter(
-        timestamp__gt=period_start,
-        extra_data__status=Message.STATUS_FAILED,
-        extra_data__include_in_daily_digest=True,
-    ).distinct("content_type", "extra_data__status", "extra_data__event")
+    failed_emails = collect_failed_emails(desired_period)
+    failed_registrations = collect_registrations_failures(desired_period)
 
-    if not logs:
+    if not (failed_emails or failed_registrations):
         return
 
     content = render_to_string(
         "emails/admin_digest.html",
         {
-            "logs": [
-                {
-                    "submission_uuid": log.content_object.uuid,
-                    "event": log.extra_data["event"],
-                }
-                for log in logs
-            ],
+            "failed_emails": failed_emails,
+            "failed_registrations": failed_registrations,
         },
     )
 
     send_mail_html(
-        _("[Open Forms] Daily summary of failed emails"),
+        _("[Open Forms] Daily summary of failed procedures"),
         content,
         settings.DEFAULT_FROM_EMAIL,
         recipients,
--- a/src/openforms/emails/tasks.py
+++ b/src/openforms/emails/tasks.py
@@ -1,4 +1,5 @@
-from datetime import timedelta
+from datetime import datetime, timedelta
+from typing import Any
 
 from django.conf import settings
 from django.template.loader import render_to_string
@@ -7,9 +8,35 @@ from django.utils.translation import gettext_lazy as _
 
 from openforms.celery import app
 from openforms.config.models import GlobalConfiguration
-from openforms.registrations.utils import collect_registrations_failures
+from openforms.logging.service import (
+    collect_failed_emails,
+    collect_failed_registrations,
+)
 
-from .utils import collect_failed_emails, send_mail_html
+from .utils import send_mail_html
+
+
+class Digest:
+    def __init__(self, since: datetime) -> None:
+        self.since = since
+
+    def get_context_data(self) -> dict[str, Any]:
+        failed_emails = collect_failed_emails(self.since)
+        failed_registrations = collect_failed_registrations(self.since)
+
+        if not (failed_emails or failed_registrations):
+            return {}
+
+        return {
+            "failed_emails": failed_emails,
+            "failed_registrations": failed_registrations,
+        }
+
+    def render(self) -> str:
+        if not (context := self.get_context_data()):
+            return ""
+
+        return render_to_string("emails/admin_digest.html", context)
 
 
 @app.task
@@ -19,23 +46,14 @@ def send_email_digest() -> None:
         return
 
     yesterday = timezone.now() - timedelta(days=1)
+    digest = Digest(since=yesterday)
+    content = digest.render()
 
-    failed_emails = collect_failed_emails(yesterday)
-    failed_registrations = collect_registrations_failures(yesterday)
-
-    if not (failed_emails or failed_registrations):
+    if not content:
         return
 
-    content = render_to_string(
-        "emails/admin_digest.html",
-        {
-            "failed_emails": failed_emails,
-            "failed_registrations": failed_registrations,
-        },
-    )
-
     send_mail_html(
-        _("[Open Forms] Daily summary of failed procedures"),
+        _("[Open Forms] Daily summary of detected problems"),
         content,
         settings.DEFAULT_FROM_EMAIL,
         recipients,
--- a/src/openforms/emails/templates/emails/admin_digest.html
+++ b/src/openforms/emails/templates/emails/admin_digest.html
@@ -9,7 +9,7 @@
     <ul>
         {% for email in failed_emails %}
             <li>
-                {% blocktranslate with event=email.event submission_uuid=email.submission_uuid %}
+                {% blocktranslate with event=email.event submission_uuid=email.submission_uuid trimmed %}
                     Email for the event "{{ event }}" for submission {{ submission_uuid }}.
                 {% endblocktranslate %}
             </li>
@@ -22,15 +22,16 @@
     <ul>
         {% for registration in failed_registrations %}
             <li>
-                {% blocktranslate with form_name=registration.form_name counter=registration.failed_submissions_counter first_failure_at=registration.initial_failure_at|time:"H:i" last_failure_at=registration.last_failure_at|time:"H:i" %}
-                    Form '{{ form_name }}' failed {{ counter }} times between {{ first_failure_at }} and {{ last_failure_at }}.
+                {% blocktranslate with form_name=registration.form_name counter=registration.failed_submissions_counter first_failure_at=registration.initial_failure_at|time:"H:i" last_failure_at=registration.last_failure_at|time:"H:i" trimmed %}
+                    Form '{{ form_name }}' failed {{ counter }} times between {{ first_failure_at }} and {{ last_failure_at }}.</br>
                 {% endblocktranslate %}
+            
+                <a href="{{ registration.admin_link }}">
+                    {% blocktranslate with form_name=registration.form_name %}
+                        View failed '{{ form_name }}' submissions
+                    {% endblocktranslate %}
+                </a>
             </li>
-            <a href="{{ registration.admin_link }}">
-                {% blocktranslate with form_name=registration.form_name %}
-                    View failed '{{ form_name }}' submissions
-                {% endblocktranslate %}
-            </a>
         {% endfor %}
     </ul>
 {% endif %}
--- a/src/openforms/emails/templates/emails/admin_digest.html
+++ b/src/openforms/emails/templates/emails/admin_digest.html
@@ -1,10 +1,24 @@
 {% load static i18n %}
 
 <p>
-    {% blocktranslate %}Here is a summary of the emails that failed to send yesterday:{% endblocktranslate %}
+    {% blocktranslate %}Here is a summary of the failed procedures in the past 24 hours:{% endblocktranslate %}
 </p>
-<ul>
-    {% for log in logs %}<li>- Email for the event "{{ log.event }}" for submission {{ log.submission_uuid }}.</li>
-    {% endfor %}
-</ul>
 
+{% if failed_emails %}
+    <h5>{% trans "Emails" %}</h5>
+    <ul>
+        {% for email in failed_emails %}
+            <li>- Email for the event "{{ email.event }}" for submission {{ email.submission_uuid }}.</li>
+        {% endfor %}
+    </ul>
+{% endif %}
+
+{% if failed_registrations %}
+    <h5>{% trans "Registrations" %}</h5>
+    <ul>
+        {% for registration in failed_registrations %}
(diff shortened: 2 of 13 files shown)


src/openforms/emails/tasks.py
```diff
@@ -20,32 +18,24 @@ def send_email_digest() -> None:
     if not (recipients := config.recipients_email_digest):
         return
 
-    period_start = timezone.now() - timedelta(days=1)
+    desired_period = timezone.now() - timedelta(days=1)
```
Viicos, line 21: I find it a bit weird to have _period_ used to describe an actual date (`timezone.now() - timedelta(days=1)` is a `datetime` object), perhaps `from_date(time)`? (same for the `collect_failed_emails` function)

sergei-maertens: I'd opt for something like:

```py
yesterday = timezone.now() - timedelta(days=1)
failed_emails = collect_failed_emails(since=yesterday)
failed_registrations = collect_failed_registrations(since=yesterday)
```

src/openforms/emails/utils.py
```diff
@@ -100,3 +103,23 @@ def render_email_template(
         backend=openforms_backend,
         disable_autoescape=disable_autoescape,
     )
+
+
+def collect_failed_emails(since: datetime) -> list[dict[str, str] | None]:
```
Viicos, line 108:
```suggestion
def collect_failed_emails(since: datetime) -> list[dict[str, str]] | None:
```

src/openforms/emails/utils.py
```diff
@@ -100,3 +103,23 @@ def render_email_template(
...
+    logs = TimelineLogProxy.objects.filter(
+        timestamp__gt=since,
+        extra_data__status=Message.STATUS_FAILED,
+        extra_data__include_in_daily_digest=True,
+    ).distinct("content_type", "extra_data__status", "extra_data__event")
```
Viicos, line 113: The docs [say](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.distinct):

> When you specify field names, you must provide an `order_by()` in the `QuerySet`, and the fields in `order_by()` must start with the fields in `distinct()`, in the same order.

I see the code was moved, so it was probably already working. Is there a need to add an `order_by` call here?

src/openforms/logging/logevent.py
```diff
@@ -208,12 +208,17 @@ def registration_success(submission: Submission, plugin):
     )
 
 
-def registration_failure(submission: Submission, error: Exception, plugin=None):
+def registration_failure(
+    submission: Submission, error: Exception, include_in_daily_digest: bool, plugin=None
+):
```
Viicos, line 213:
```suggestion
def registration_failure(
    submission: Submission, error: Exception, *, plugin: ... = None, include_in_daily_digest: bool
):
```
To avoid [boolean traps](https://adamj.eu/tech/2021/07/10/python-type-hints-how-to-avoid-the-boolean-trap/). Although this makes it a little weird with `plugin` having a default value

src/openforms/emails/utils.py
Viicos: As it is done for `collect_registrations_failures`, maybe we should also filter on `extra_data__log_event="email_status_change"`? It might be that other logs are using `4` (the value of `Message.STATUS_FAILED`) as a value for `extra_data["status"]`.

src/openforms/registrations/utils.py
```diff
@@ -30,3 +33,34 @@ def execute_unless_result_exists(
     assign(submission.registration_result, spec, result, missing=dict)
     submission.save(update_fields=["registration_result"])
     return callback_result
+
+
+def collect_registrations_failures(since: datetime) -> list[dict[str, str] | None]:
```
Viicos, line 38:
```suggestion
def collect_registrations_failures(since: datetime) -> list[dict[str, str]] | None:
```

src/openforms/registrations/utils.py
```diff
@@ -30,3 +33,34 @@ def execute_unless_result_exists(
...
+            connected_forms[form.uuid]["failed_submissions_counter"] += 1
+            connected_forms[form.uuid]["last_failure_at"] = log.timestamp
+        else:
+            connected_forms[form.uuid] = {
+                "form_name": form.name,
+                "failed_submissions_counter": 1,
+                "initial_failure_at": log.timestamp,
+                "last_failure_at": log.timestamp,
+                "admin_link": get_filtered_submission_admin_url(form.id, 1, "24hAgo"),
+            }
```
Viicos, line 62: Maybe a `defaultdict` is better suited here

src/openforms/registrations/utils.py
```diff
@@ -30,3 +33,34 @@ def execute_unless_result_exists(
...
+
+    failed_registrations = [form_detail for form_detail in connected_forms.values()]
+
+    return failed_registrations
```
Viicos, line 66:
```suggestion
    return list(failed_registrations.values())
```

src/openforms/emails/utils.py
```diff
@@ -100,3 +103,23 @@ def render_email_template(
...
+
+    if not logs:
+        return
```
Viicos, line 116: Probably you can return an empty list anyway, makes the return type a bit simpler?

src/openforms/emails/tasks.py
```diff
@@ -20,32 +18,24 @@ def send_email_digest() -> None:
...
-        timestamp__gt=period_start,
-        extra_data__status=Message.STATUS_FAILED,
-        extra_data__include_in_daily_digest=True,
-    ).distinct("content_type", "extra_data__status", "extra_data__event")
+    failed_emails = collect_failed_emails(yesterday)
+    failed_registrations = collect_registrations_failures(yesterday)
 
-    if not logs:
+    if not (failed_emails or failed_registrations):
         return
```
Viicos, line 27: Might be worth sending the email even if they are no failures? In that case, this early return should be removed, and I think my comment in the `src/openforms/emails/templates/emails/admin_digest.html` file would make more sense.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -1,10 +1,24 @@
...
-</ul>
 
+{% if failed_emails %}
+    <h5>{% trans "Emails" %}</h5>
+    <ul>
+        {% for email in failed_emails %}
+            <li>- Email for the event "{{ email.event }}" for submission {{ email.submission_uuid }}.</li>
+        {% endfor %}
+    </ul>
+{% endif %}
```
Viicos, line 14: Maybe using [`for ... empty`](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#for-empty), we could show a default message?

sergei-maertens: if there's nothing to report, don't include necessary bloat in the email. If you want people to read the emails, keep them as short as possible.

src/openforms/emails/utils.py
Viicos: Ah yes I meant the template field indeed

src/openforms/emails/tasks.py
```diff
@@ -5,13 +5,11 @@
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
-from django_yubin.models import Message
-
 from openforms.celery import app
 from openforms.config.models import GlobalConfiguration
-from openforms.logging.models import TimelineLogProxy
+from openforms.registrations.utils import collect_registrations_failures
```
sergei-maertens, line 10: there's a remark further down suggesting to move this to `openforms.logging.service`

src/openforms/emails/tasks.py
```diff
@@ -5,13 +5,11 @@
...
 
-from .utils import send_mail_html
+from .utils import collect_failed_emails, send_mail_html
```
sergei-maertens, line 12: can we keep the naming consistent please?

`collection_registration_failures` vs. `collect_failed_emails` -> `collect_failed_registrations` & `collect_failed_emails`

src/openforms/emails/tasks.py
```diff
@@ -20,32 +18,24 @@ def send_email_digest() -> None:
...
-                for log in logs
-            ],
+            "failed_emails": failed_emails,
+            "failed_registrations": failed_registrations,
         },
     )
 
     send_mail_html(
-        _("[Open Forms] Daily summary of failed emails"),
+        _("[Open Forms] Daily summary of failed procedures"),
```
sergei-maertens, line 38:
```suggestion
        _("[Open Forms] Daily summary of detected problems"),
```

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -1,10 +1,24 @@
 {% load static i18n %}
 
 <p>
-    {% blocktranslate %}Here is a summary of the emails that failed to send yesterday:{% endblocktranslate %}
+    {% blocktranslate %}Here is a summary of the failed procedures in the past 24 hours:{% endblocktranslate %}
```
sergei-maertens, line 4:
```suggestion
    {% blocktranslate %}Here is a summary of the problems detected in the past 24 hours:{% endblocktranslate %}
```

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -1,10 +1,24 @@
...
 </p>
-<ul>
-    {% for log in logs %}<li>- Email for the event "{{ log.event }}" for submission {{ log.submission_uuid }}.</li>
-    {% endfor %}
-</ul>
 
+{% if failed_emails %}
+    <h5>{% trans "Emails" %}</h5>
```
sergei-maertens, line 8:
```suggestion
    <h5>{% trans "Emails that failed to send" %}</h5>
```

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -1,10 +1,24 @@
...
+    <ul>
+        {% for email in failed_emails %}
+            <li>- Email for the event "{{ email.event }}" for submission {{ email.submission_uuid }}.</li>
+        {% endfor %}
```
sergei-maertens, line 12: should be properly translated with `{% blocktranslate %}`

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -1,10 +1,24 @@
...
+    </ul>
+{% endif %}
+
+{% if failed_registrations %}
+    <h5>{% trans "Registrations" %}</h5>
+    <ul>
+        {% for registration in failed_registrations %}
+            <li>- {{ registration.form_name }} failed {{ registration.failed_submissions_counter }} times between {{ registration.initial_failure_at|date:"H:i" }} and {{ registration.last_failure_at|date:"H:i" }}.</li>
```
sergei-maertens, line 20: make sure this is translatable please :)

sergei-maertens, line 20: since we're displaying times, the `|time` filter is more appropriate.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -1,10 +1,24 @@
...
+            <a href="{{ registration.admin_link }}">Click here to see all the failed submissions for form {{ registration.form_name }} in the admin</a>
```
sergei-maertens, line 21: "click here" is a bit of an anti-pattern, better would be:

```django
<a href="{{ ... }}">View failed '{{ form_name }}' submissions</a>
```

src/openforms/emails/tests/test_tasks.py
```diff
@@ -177,3 +181,92 @@ def test_no_recipients(self):
...
+                    Click here to see all the failed submissions for form {form_1.name} in the admin
+                    </a>
+                    <li>- {form_2.name} failed 1 times between 12:30 and 12:30.</li>
+                    <a href="/admin/submissions/submission/?form__id__exact={form_2.id}&needs_on_completion_retry__exact=1&from_time=24hAgo">
+                    Click here to see all the failed submissions for form {form_2.name} in the admin
+                    </a>
+                </ul>
+
+            """
+            ).strip()
```
sergei-maertens, line 267: why do we need to mock the email sending? Django itself already sets up an in-memory mail backend, so you can introspect sent emails via `django.core.mail.outbox`.

src/openforms/emails/utils.py
sergei-maertens, line 108: I'd prefer a more specific type hint than just a dict. I'd even set this up as a dataclass:

```py
@dataclass
class FailedEmail:
    submission_uuid: uuid.UUID
    event: str  # or Literal?


def collect_failed_emails(since: datetime) -> Iterable[FailedEmail]:
    ...
```

src/openforms/emails/utils.py
sergei-maertens: w/r to the order_by, it's most likely implicit on the model itself. Postgres complains/crashes hard if it's misaligned, so the existing test coverage helps enforce this.

src/openforms/emails/utils.py
sergei-maertens: +1 from me on this

src/openforms/registrations/utils.py
sergei-maertens, line 38: this would also benefit from using a dataclass for the processed log data instead of `dict[str, str]` (which is also a lie because some values are `int` and `datetime` rather than `str`)

src/openforms/registrations/utils.py
```diff
@@ -30,3 +33,34 @@ def execute_unless_result_exists(
     assign(submission.registration_result, spec, result, missing=dict)
     submission.save(update_fields=["registration_result"])
     return callback_result
+
+
+def collect_registrations_failures(since: datetime) -> list[dict[str, str] | None]:
+    logs = TimelineLogProxy.objects.filter(
+        timestamp__gt=since,
+        extra_data__log_event="registration_failure",
+        extra_data__include_in_daily_digest=True,
```
sergei-maertens, line 42: where are records being created that should not be included?

src/openforms/registrations/utils.py
```diff
@@ -30,3 +33,34 @@ def execute_unless_result_exists(
...
+    ).order_by("timestamp")
```
sergei-maertens, line 43: I think you can do some clever `.annotate` behaviour to get all the necessary information in one go rather than having to post-process in Python:

```py
qs = TimelineLogProxy.objects\
    .filter(...)\
    .values("object_id")\
    .annotate(
        num_failed=Count("object_id"), 
        first_at=Min("timestamp"), 
        last_at=Max("timestamp"),
    )
```
which gives you the raw data, and then you have the form IDs so you can fetch the UUIDs and names:

```py
class FailedSubmission:
    form: Form
    num_failed: int
    first_failure_at: datetime
    last_failure_at: datetime

    @property
    def admin_link(self) -> str:
       ...


forms = Form.objects.in_bulk(id_list=[item["object_id"] for item in qs], field_name="id")
aggregated_logs = [
    FailedSubmission(form=forms[ int(item["object_id"]) ], num_failed=item["num_failed"], ...)
    for item in qs
]
```

sergei-maertens (changes requested):

src/openforms/registrations/utils.py
vaszig: This was discussed with @sergei-maertens in the office and we decided that we will go for the pythonic way since adding django's `GenericRelation` for this is a pain.

src/openforms/registrations/utils.py
vaszig: Discussed in the office, `groupby` from itertools is the best choice here.

src/openforms/submissions/utils.py
```diff
@@ -335,3 +336,13 @@ def get_report_download_url(request: Request, report: SubmissionReport) -> str:
...
+
+
+def get_filtered_submission_admin_url(form_id: int, state: int, from_time: str) -> str:
+    query_params = {
+        "form__id__exact": form_id,
+        "needs_on_completion_retry__exact": state,
+        "from_time": from_time,
+    }
+    submissions_admin_url = furl(reverse("admin:submissions_submission_changelist"))
+    return submissions_admin_url.add(query_params).url
```
sergei-maertens, line 348: the call-sites with positional arguments and "magic numbers" are not very readable, so let's enforce you have to use kwargs:

```suggestion
def get_filtered_submission_admin_url(form_id: int, *, filter_retry: bool, from_time: str) -> str:
    query_params = {
        "form__id__exact": form_id,
        "needs_on_completion_retry__exact": 1 if filter_retry else 0,
        "from_time": from_time,
    }
    submissions_admin_url = furl(reverse("admin:submissions_submission_changelist"))
    return submissions_admin_url.add(query_params).url
```

src/openforms/logging/service.py
```diff
@@ -0,0 +1,78 @@
...
+    if not logs:
+        return []
+
+    failed_emails = [
+        {
+            "submission_uuid": log.content_object.uuid,
+            "event": log.extra_data["event"],
+        }
+        for log in logs
+    ]
```
sergei-maertens, line 45: the type signature is not respected here - you are returning dicts rather than data class instances

```suggestion
    failed_emails = [
        FailedEmail(submission_uuid=log.content_object.uuid, event=log.extra_data["event"])
        for log in logs
    ]
```

sergei-maertens (changes requested):
Some more details :)

PR #4095, opened 2024-04-03
[#4005] Filter submission reports

### LINKED ISSUE #4005: Add more columns to admin/submissionreport
### Thema / Theme

Admin

As a admin when looking for a specific submission report i have a hard time finding the right one.
The admin view currently only shows: "form title: Document"

I would like to at least be able to search for submission id and public reference. 

Alternatively perhaps it would be nice to be able to download these documents directly from submissions/submission as is already possible for attachments. 
image

SilviaAmAm:
Fixes #4005


vaszig (approved):
I guess a rebase will solve the problem with the docker in the CI.

sergei-maertens (approved):

PR #4108, opened 2024-04-03
Use the plugin appointment ID instead of our internal DB pk

### LINKED ISSUE #4103: JCC appointment: Creating an appointment uses the wrong data, creating an appointment for a date in the past and an incorrect product

2.5.4 

Taiga HadM #93

Creating an appointment via JCC uses the wrong data, this leads to an appointment for a date in the past and an incorrect product.

image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #4103

Appointment details were being retrieved using our internal database primary key, but that looks up records from the appointment system that are unlikely to match with the actual appointment ID recorded for a given appointment.

### DIFF
--- a/src/openforms/appointments/renderer.py
+++ b/src/openforms/appointments/renderer.py
@@ -45,8 +45,9 @@ class AppointmentRenderer(Renderer):
         )
         plugin = get_plugin(plugin=appointment.plugin)
 
+        identifier: str = self.submission.appointment_info.appointment_id
         ctx = {
-            "appointment": plugin.get_appointment_details(appointment.pk),
+            "appointment": plugin.get_appointment_details(identifier),
             "contact_details": self.get_children(),  # todo: a bit of a wart
         }
         return render_to_string(template_name, ctx)
--- a/src/openforms/appointments/tests/test_pdf.py
+++ b/src/openforms/appointments/tests/test_pdf.py
@@ -1,10 +1,13 @@
-from django.test import RequestFactory, TestCase, override_settings
+from unittest.mock import patch
+
+from django.test import RequestFactory, TestCase, override_settings, tag
 from django.utils.html import escape
 
 from openforms.accounts.tests.factories import SuperUserFactory
 from openforms.submissions.dev_views import SubmissionPDFTestView
 from openforms.submissions.tests.factories import SubmissionFactory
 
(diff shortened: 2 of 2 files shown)


Viicos (approved):

src/openforms/appointments/renderer.py
```diff
@@ -45,8 +45,9 @@ def __str__(self) -> SafeString:
         )
         plugin = get_plugin(plugin=appointment.plugin)
 
+        identifier: str = self.submission.appointment_info.appointment_id
```
SilviaAmAm, line 48: I see in the `AppointmentInfo` model that the `appointment_id` can also be empty. Are we sure that it always has a value here?

src/openforms/appointments/tests/test_pdf.py
```diff
@@ -59,3 +63,17 @@ def test_submission_without_appointment_shows_no_appointment_info(self):
...
+
+    @tag("gh-4103")
+    def test_uses_remote_appoinment_id(self):
+        plugin = DemoAppointment("demo")
+
+        with (
+            patch("openforms.appointments.renderer.get_plugin", return_value=plugin),
+            patch.object(
+                plugin, "get_appointment_details", wraps=plugin.get_appointment_details
+            ) as m_get_details,
```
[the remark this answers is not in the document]
sergei-maertens: I tried and it causes crashes in the template engine, even when using `wraps`

src/openforms/appointments/renderer.py
sergei-maertens: yes, for appointments that were succesfully created in JCC/Qmatic/demo, this value is set. It is the return value of `plugin.create_appointment`: https://github.com/open-formulieren/open-forms/blob/3ef4ea6bd4a1d720f67073da2bec37269fd62bc7/src/openforms/appointments/core.py#L55

sergei-maertens:
Backports:

* stable/2.6.x: cba3d8cbd63e4aa02ec0581c64da30a2e33da0fa
* stable/2.5.x: b37ebf1cc49204120eb4d18f8926ea21c65b2913
* stable/2.4.x: 2eed48c434a5e0d9e5edc52c4c0ff53884b321c4
* stable/2.3.x: n/a

PR #4109, opened 2024-04-03
Run e2e tests in parallel

sergei-maertens:


vaszig (approved):
Why do we say "if we can run them in parallel"?Did you expect something specific to fail?

sergei-maertens:
> Why do we say "if we can run them in parallel"?Did you expect something specific to fail?

the cache is not isolated between parallel processes, so it can cause race conditions :)

PR #4114, opened 2024-04-03
Fix validation issue with (dynamically) optional field

sergei-maertens:
Taiga DH 671

~~If the test suite passes first try I'm eating my shoe.~~

Needs backport to stable/2.6.x

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -12,14 +12,9 @@ from openforms.formio.service import (
 from openforms.formio.utils import get_component_empty_value
 from openforms.typing import DataMapping
 
-from .logic.actions import PropertyAction
+from .logic.actions import ActionOperation
 from .logic.datastructures import DataContainer
-from .logic.rules import (
-    EvaluatedRule,
-    get_current_step,
-    get_rules_to_evaluate,
-    iter_evaluate_rules,
-)
+from .logic.rules import EvaluatedRule, get_rules_to_evaluate, iter_evaluate_rules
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/form_logic.py
```diff
@@ -214,15 +208,19 @@ def check_submission_logic(
         submission_variables_state.set_values(unsaved_data)
     data_container = DataContainer(state=submission_variables_state)
 
-    mutation_operations = []
+    mutation_operations: list[ActionOperation] = []
     for operation in iter_evaluate_rules(rules, data_container, submission):
```
Viicos, line 212: Probably this can be built using a list comprehension now

Viicos (approved):
So if I understand correctly this was a bug before backend validation, but only appears now with this validation?

sergei-maertens:
> So if I understand correctly this was a bug before backend validation, but only appears now with this validation?

No, because the logic evaluation for the entire submission runs after all form steps have been submitted. It was deliberate to skip the property actions, since they affect the UI/client side validation, but that was not presented anymore because the user is on the "confirmation" screen.

src/openforms/submissions/form_logic.py
sergei-maertens: I think a list comprehension makes the code look pure, but it isn't, data mutations are applied as the logic rules are evaluated by looping through them.

sergei-maertens:
Backported to stable/2.6.x in b0dc21dca80b6bbefaefdfd956a793e85ee721dd

PR #4116, opened 2024-04-04
[#4096] Fix validation of hidden radio component 

### LINKED ISSUE #4096: Hidden required choicefields that do not have a value, don't validate
Ref: Taiga Utrecht 195

```
rest_framework.exceptions.ValidationError: {'steps': [{}, {'data': {'bentUEenPersoonOfEenBedrijf': [ErrorDetail(string='"" is een ongeldige keuze.', code='invalid_choice')]}}, {}]}
```

> In dit formulier wordt deze vraag verborgen d.m.v. de logica. Als dit veld wél wordt getoond is het een verplichte vraag. Kan het zijn dat doordat de verborgen velden sinds 2.6.1 worden meegestuurd de backend deze waardes ook valideert en daarover struikelt?

Viicos:
Fixes #4096

### DIFF
--- a/src/openforms/formio/serializers.py
+++ b/src/openforms/formio/serializers.py
@@ -87,6 +87,9 @@ class StepDataSerializer(serializers.Serializer):
                 field.min_length = None
                 field.max_length = None
 
+            case serializers.ChoiceField():
+                field.allow_blank = True
+
     def _get_required(self) -> bool:
         return any(field.required for field in self.fields.values())
 
--- a/src/openforms/formio/tests/validation/test_radio.py
+++ b/src/openforms/formio/tests/validation/test_radio.py
@@ -1,0 +1,31 @@
+from django.test import SimpleTestCase, tag
+
+from openforms.typing import JSONObject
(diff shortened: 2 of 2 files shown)


src/openforms/formio/serializers.py
```diff
@@ -87,6 +87,9 @@ def _remove_validations_from_field(self, field: serializers.Field) -> None:
                 field.min_length = None
                 field.max_length = None
 
+            case serializers.ChoiceField():
+                field.allow_blank = True
+
```
Viicos, line 92: Can't we just have a way to disable validation completely if the field is hidden? This feels like a hacky way of making validation pass

sergei-maertens: no, because the datatype still needs to match and the JSON input needs to be converted to the appropriate python type.

sergei-maertens (approved):

sergei-maertens:
Backport: a0cf097afd93e8e1993f311ee562908d615ab1b6

PR #4118, opened 2024-04-04
:white_check_mark: Add missing validation tests for radio input

sergei-maertens:
These were missed during initial implementation :grimacing:


Viicos (approved):
The empty file was already there but I thought it was on purpose as some kind of placeholder :sweat_smile:

PR #4120, opened 2024-04-04
Deal with flaky test

sergei-maertens:
* Add retry behaviour
* Log that test if flaky
* Ensure that flaky tests get reported in github annotations

How it looks in Github:
image

### DIFF
--- a/.github/actions/report-flaky/report_flaky.js
+++ b/.github/actions/report-flaky/report_flaky.js
@@ -1,0 +1,26 @@
+// Taken from: https://nodejs.org/api/readline.html#readline_example_read_file_stream_line_by_line
+const fs = require('node:fs');
+const readline = require('node:readline');
+
+const core = require('@actions/core');
+
+const logFilePath = core.getInput('logFile');
+
+async function emitAnnotations() {
+  const fileStream = fs.createReadStream(logFilePath);
+
+  const rl = readline.createInterface({
+    input: fileStream,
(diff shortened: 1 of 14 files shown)


src/openforms/formio/tests/test_api_fileupload.py
```diff
@@ -395,4 +399,19 @@ def do_upload() -> str:
         }
 
         session_uuids = set(self.client.session[UPLOADS_SESSION_KEY])
+
+        log_flaky()
```
sergei-maertens, line 403: TODO: move to inside the if-branch when the experiment succeeds :)

.github/actions/report-flaky/report_flaky.js
```diff
@@ -0,0 +1,26 @@
...
+
+async function emitAnnotations() {
+  const fileStream = fs.createReadStream(logFilePath);
+
+  const rl = readline.createInterface({
+    input: fileStream,
+    crlfDelay: Infinity,
+  });
+  // Note: we use the crlfDelay option to recognize all instances of CR LF
+  // ('\r\n') in input.txt as a single line break.
```
CharString, line 17: LOL. And pray you never encounter MacOS < X files... Those use only \r.

sergei-maertens: copy and pasted from the nodejs docs and we run github actions on ubuntu-latest :feelsDabMan:

PR #4121, opened 2024-04-04
[#4084] Add migration to fix multiple select default value

### LINKED ISSUE #4084: Select with multiple: true has wrong default value
Currently the form builder produces a default value of `[""]` (i.e. a list with one element - the empty string), which should just be an empty list entirely (`[]`) instead.

This causes validation issues, because the empty string is typically not a valid choice.

**Tasks**

- [x] Patch form builder
- [x] Add data migration (+ form import fixer) to convert `defaultValue: [""]` into `defaultValue: []`
- [x] Check/patch similar components:
    - [x] Radio
    - [x] Selectboxes

Viicos:
Partly fixes #4084,
requires:
- https://github.com/open-formulieren/formio-builder/pull/135
- https://github.com/open-formulieren/formio-builder/pull/138


sergei-maertens:
I've put it back in draft, I'd like to see the formio-builder package upgrade in this PR too before we merge it.

SilviaAmAm (approved):

PR #4123, opened 2024-04-05
Fix Objects API JSON data for missing values

### LINKED ISSUE #3890: Allow hidden fields to be sent to registration systems

sergei-maertens:
Closes #3890

Follow up of the patch for #3890

- [x] Ensure that the component-specific empty value is used for missing fields/variables
- [x] Ensure that fields inside hidden fieldsets are sent to the objects API

### DIFF
--- a/src/openforms/formio/rendering/default.py
+++ b/src/openforms/formio/rendering/default.py
@@ -33,7 +33,9 @@ class ContainerMixin:
     def is_visible(self) -> bool:
         # fieldset/editgrid components do not support the showInFoo properties, so we don't use the super
         # class.
-        if self.mode == RenderModes.export:
+        # In registration mode, we need to treat layout/container nodes as visible so
+        # that their children are emitted too.
+        if self.mode in {RenderModes.export, RenderModes.registration}:
             return True
 
         # We only pass the step data, since frontend logic only has access to the current step data.
--- a/src/openforms/formio/rendering/tests/test_component_node.py
+++ b/src/openforms/formio/rendering/tests/test_component_node.py
@@ -1,6 +1,6 @@
 from unittest.mock import patch
 
-from django.test import TestCase
+from django.test import TestCase, tag
 
 from openforms.submissions.rendering import Renderer, RenderModes
 from openforms.submissions.tests.factories import SubmissionFactory
@@ -8,6 +8,7 @@ from openforms.submissions.tests.factories import SubmissionFactory
 from ..constants import RenderConfigurationOptions
 from ..nodes import ComponentNode
 from ..registry import Registry
+from ..structured import render_json
 
 
 class FormNodeTests(TestCase):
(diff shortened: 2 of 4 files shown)


src/openforms/formio/rendering/tests/test_component_node.py
```diff
@@ -363,6 +364,40 @@ def test_explicitly_hidden_component_not_skipped_when_registration(self):
...
 
+    @tag("dh-673")
+    def test_visible_component_inside_hidden_fieldset_not_skipped(self):
+        submission = SubmissionFactory.from_components(
+            [
+                {
+                    "type": "fieldset",
+                    "key": "fieldset",
+                    "label": "Hidden fieldset",
+                    "hidden": False,
```
Viicos, line 375: Shouldn't it be `hidden: True`?

src/openforms/registrations/contrib/objects_api/tests/test_template.py
```diff
@@ -305,3 +309,109 @@ def test_submission_with_objects_api_content_json_not_valid_json(self):
...
+                f"{prefix}.submission_registration.get_documents_client"
+            ) as mock_documents_client,
+            patch(f"{prefix}.plugin.get_objects_client") as mock_objects_client,
+        ):
+            _documents_client = (
+                mock_documents_client.return_value.__enter__.return_value
+            )
+            _documents_client.create_document.return_value = {
+                "url": "https://example.com/dummy"
+            }
```
Viicos, line 386: Can't we skip the mocking of the documents client, by setting `informatieobjecttype_submission_report` to `None` in the registration options?

src/openforms/registrations/contrib/objects_api/tests/test_template.py
```diff
@@ -305,3 +309,109 @@ def test_submission_with_objects_api_content_json_not_valid_json(self):
...
+            with_report=True,
+            submitted_data={"radio": "2"},
+            form_definition_kwargs={"slug": "stepwithnulls"},
+        )
+        self._add_submission_to_session(submission)
+        # complete the submission
+        endpoint = reverse("api:submission-complete", kwargs={"uuid": submission.uuid})
+        with self.captureOnCommitCallbacks(execute=True):
+            response = self.client.post(endpoint, {"privacy_policy_accepted": True})
+        assert response.status_code == status.HTTP_200_OK
```
Viicos, line 362: Was this necessary to have the empty values for hidden components (as described in the comment later in this test)?

src/openforms/formio/rendering/default.py
```diff
@@ -33,7 +33,9 @@ class ContainerMixin:
     def is_visible(self) -> bool:
         # fieldset/editgrid components do not support the showInFoo properties, so we don't use the super
         # class.
-        if self.mode == RenderModes.export:
+        # In registration mode, we need to treat layout/container nodes as visible so
+        # that their children are emitted too.
+        if self.mode in {RenderModes.export, RenderModes.registration}:
```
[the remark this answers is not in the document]
sergei-maertens: it does, but I checked how it behaves when the components are visible and that's also the current behaviour, because it threw me off.

src/openforms/formio/rendering/tests/test_component_node.py
sergei-maertens: oops, that was my sanity check indeed.

src/openforms/registrations/contrib/objects_api/tests/test_template.py
sergei-maertens: hrm, no, in hindsight it turns out that the completion of the submission did not mutate the data in any way, which I suspected at first :thinking:

src/openforms/registrations/contrib/objects_api/tests/test_template.py
sergei-maertens: succeeded!

PR #4125, opened 2024-04-05
Apply client-side show/hide logic in nested components during editgrid validation

### LINKED ISSUE #4126: Fix validation of conditionally visible components inside repeating groups
Taiga DH 674

Too complicated (and edge case) for 2.6.2, but needs to be in 2.6.3.

If you have a component inside a repeating group that is `{validate: {required: true}}` and conditionally visible, the frontend skips the validation but the backend is not able to see that the validation should be skipped.

This is complicated because every row in the list behaves independent from its siblings.

sergei-maertens:
Closes #4126

Simple conditionals which run as client-side logic are supported (not deliberatey, but here we are...) for edit grids (aka repeating groups). This works fine in the client side since the Formio.js SDK takes care of it (in some questionable ways, but it works...). Components with validation requirements don't trigger errors, because when they are hidden, all validations are ignored.

However, the backend created a simple `serializers.ListField(...)` for edit grids, with a child/nested serializer derived from the "blueprint" of nested components, which cannot take into account the show/hide-and-drop-validations behaviour. This is because the `child` instance is re-used to run the validation for each item in the data, and the nested serializer field configuration depends on the values of each item in the data.

So, to patch this, a similar implementation as `serializers.ListField` is used which dynamically creates the `child` serializer for each item in the data (instead of creating that upfront). This allows tweaking/modification depending on the visibility state in the "frontend".

### DIFF
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -651,7 +651,7 @@ class EditGridField(serializers.Field):
             # configuration and apply the dynamic hide/visible logic to it.
             # Note that we add the editgrid key as container so that the
             # conditional.when values resolve, as these look like `editgridparent.child`.
-            data = FormioData({self.field_name: item}).data
+            data = FormioData({**self.root.initial_data, self.field_name: item}).data
             nested_serializer = self._build_child(data=data)
             try:
                 result.append(nested_serializer.run_validation(item))
--- a/src/openforms/formio/components/vanilla.py
+++ b/src/openforms/formio/components/vanilla.py
@@ -6,15 +6,23 @@ adjacent custom.py module.
 """
 
 import logging
+from collections.abc import Mapping
 from datetime import time
 from typing import TYPE_CHECKING, Any
 
-from django.core.validators import MaxValueValidator, MinValueValidator, RegexValidator
+from django.core.validators import (
+    MaxLengthValidator,
+    MaxValueValidator,
+    MinLengthValidator,
+    MinValueValidator,
+    RegexValidator,
+)
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
 from rest_framework.request import Request
 from rest_framework.reverse import reverse
+from rest_framework.utils.formatting import lazy_format
 
 from csp_post_processor import post_process_html
 from openforms.config.constants import UploadFileType
@@ -23,6 +31,7 @@ from openforms.typing import DataMapping
 from openforms.utils.urls import build_absolute_uri
 from openforms.validations.service import PluginValidator
 
+from ..datastructures import FormioData
 from ..dynamic_config.dynamic_options import add_options_to_config
 from ..formatters.formio import (
     CheckboxFormatter,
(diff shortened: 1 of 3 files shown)


src/openforms/formio/components/vanilla.py
```diff
@@ -563,24 +573,112 @@ def rewrite_for_request(component: ContentComponent, request: Request):
...
+
+    def to_representation(self, value: list):
+        """
+        List of object instances -> List of dicts of primitive datatypes.
+        """
+        child = self._build_child()
+        return [
+            child.to_representation(item) if item is not None else None
+            for item in value
+        ]
```
sergei-maertens, line 638: These probably are missing coverage

src/openforms/formio/components/vanilla.py
```diff
@@ -563,24 +573,112 @@ def rewrite_for_request(component: ContentComponent, request: Request):
...
+
+    def _build_child(self, **kwargs):
+        return build_serializer(
+            components=copy.deepcopy(self.components),
```
sergei-maertens, line 642: Come to think of it, these are probably not mutated at all during the logic check, so likely we can drop the `deepcopy` :thinking:

src/openforms/formio/components/vanilla.py
```diff
@@ -563,24 +573,112 @@ def rewrite_for_request(component: ContentComponent, request: Request):
...
+
+    def run_child_validation(self, data):
+        result = []
+        errors = {}
+
+        for idx, item in enumerate(data):
+            # given the local scope of data, build a nested serializer for the component
+            # configuration and apply the dynamic hide/visible logic to it.
+            # Note that we add the editgrid key as container so that the
+            # conditional.when values resolve, as these look like `editgridparent.child`.
```
sergei-maertens, line 657: this is similar to what formio does, however they do it the other way around - the find the nearest data container component for a given component and then calculate the key prefixes and do some `startsWith` checks that all look very fragile.

src/openforms/formio/serializers.py
```diff
@@ -68,6 +68,12 @@ def apply_hidden_state(
             self._remove_validations_from_field(serializer_field)
 
     def _remove_validations_from_field(self, field: serializers.Field) -> None:
+        from .components.vanilla import EditGridField  # circular import
```
sergei-maertens, line 71: I don't like this, but I'm not sure how I can leverage pattern matching here with kwargs that don't exist on `serializers.Field` :thinking:

SilviaAmAm (approved):
Does it also work if the component  specified in the `when` is outside of the edit grid?

sergei-maertens:
> Does it also work if the component specified in the `when` is outside of the edit grid?

oh dear lord... :grimacing:

sergei-maertens:
So in Formio's builder you can also refer to components inside nested editgrids:

```
parentEditgrid
  
  someTextField
  
  nestedEditgrid
  
    someNestedTextField
```
which gives the impression that you can make `someTextField` conditional visibility based on someNestedTextField. However, that doesn't seem to work on their playground (https://formio.github.io/formio.js/app/builder.html) so I'm leaving that out of scope for the backend too.

sergei-maertens:
Backported to 2.6.x in c4ebb17c32c795b4808f1423a7dcd90fcd75f16a

PR #4132, opened 2024-04-08
[#4101] Disable "Show form" button if form not active

### LINKED ISSUE #4101: If form is not active do not show "show form" button in list/detail view

Viicos:
Fixes #4101

Using `pointer-events: none;` prevents `title` from showing on hoover, and using a top level `span` element with a `title` didn't help either :/

### DIFF
--- a/src/openforms/forms/admin/form.py
+++ b/src/openforms/forms/admin/form.py
@@ -14,6 +14,7 @@ from openforms.api.utils import underscore_to_camel
 from openforms.emails.models import ConfirmationEmailTemplate
 from openforms.payments.admin import PaymentBackendChoiceFieldMixin
 from openforms.registrations.admin import RegistrationBackendFieldMixin
+from openforms.typing import StrOrPromise
 from openforms.utils.expressions import FirstNotBlank
 
 from ..models import Category, Form, FormDefinition, FormStep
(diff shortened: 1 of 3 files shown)


vaszig (commented):
Is it a problem on my end that I keep on seeing a link even when the form is inactive?
If I click it I get a 404.

Viicos:
> Is it a problem on my end that I keep on seeing a link even when the form is inactive?
> If I click it I get a 404.

Hum probably you need to run `npm run start/build` first?

vaszig:
> > Is it a problem on my end that I keep on seeing a link even when the form is inactive?
> > If I click it I get a 404.
> 
> Hum probably you need to run `npm run start/build` first?

As we discussed in the office I was talking about the wrong button!It's' working for the `form_change` page but we have to do the same for the list page of the forms.

Viicos:
I'll first need to know if this worth the effort, as the list view list is defined as a Django `list_display` field, making it hard to customize depending on the `active` value.

src/openforms/forms/admin/form.py
```diff
@@ -235,12 +236,14 @@ def get_queryset(self, request):
         )
 
     @admin.display(description=_("Actions"))
-    def get_object_actions(self, obj) -> str:
-        links = ((obj.get_absolute_url(), _("Show form")),)
+    def get_object_actions(self, obj: Form) -> str:
+        links: list[tuple[str, StrOrPromise]] = []
+        if obj.active:
+            links.append((obj.get_absolute_url(), _("Show form")))
```
Viicos, line 242: The link could also be made disabled: I made a draft implementation but this adds complexity to the code so let me know if it is necessary.

sergei-maertens: this is fine!

sergei-maertens (approved):

PR #4136, opened 2024-04-08
Fix plugin validators being AND-ed instead of OR-ed in the backend validation.

sergei-maertens:
Closes #4121

I am schtewpid


src/openforms/validations/drf_validators.py
```diff
@@ -22,18 +24,35 @@ class PluginValidator:
 
     requires_context = True
 
-    def __init__(self, plugin: str):
-        assert plugin in register, "Invalid plugin identifier specified"
-        self.plugin = plugin
+    def __init__(self, plugins: Iterable[str]):
```
Viicos, line 27: I wish we had a way to deal with https://github.com/python/typing/issues/256

Viicos (approved):
I'm wondering why it started appearing in #4121

src/openforms/validations/drf_validators.py
sergei-maertens: could've been `list[str]` but I followed your workshop advice :P

sergei-maertens:
Backported to 2.6.x via 5e8bb37f2e263aa228eed0895f690074dd956e6e

PR #4137, opened 2024-04-08
:whale: Use uv to install dependencies in docker build

sergei-maertens:
Shaves off about ~2 minutes of the docker build time :)


vaszig (approved):

Viicos (approved):
:partying_face:

Issue #4080, opened 2024-03-29
Registration PDF does not update to display the payment status correctly


LaurensBurger:
### Product versie / Product version

2.5.4
Vught 68
RID 195

### Omschrijf het probleem / Describe the bug

When the environment is set to: `wait_for_payment_to_register=True`
The PDF included in the registration should refelct the successes of the payment. But at the moment it can only display the "old" status:
![image](https://github.com/open-formulieren/open-forms/assets/112621537/1911228b-bb22-454e-a7ff-7a90e61b694f)

since it was generated before the payment status was updated.
![image](https://github.com/open-formulieren/open-forms/assets/112621537/64f7a575-05d5-47a2-9804-8cfab00ddca8)

It should be regenerated at this point to reflect the right payment status.

---

We're including the user-download PDF here to reflect the same status (so, you download a document that states you paid).

* [x] Add timestamp to PDF when it is generated: #4139 
* [ ] Handle the case where registration is done before and after payment.
* [ ] The PDF after payment needs to reflect the updated payment status and this updated PDF needs to be sent to the registration backend. Whether the PDF is regenerated always during registration or just after payment, is up to the dev, but I (Joeri) wants that the PDF that the client can download is the same as the one sent to the registration backend (with a preference for the same physical file).

joeribekker:
Refinement: Maybe we should drop support for payment after creating the registration. But, that would make the "betaalstatus" field useless so technically it seems still relevant.

The issue is how to deal with a new version of the PDF that says payment is complete, for all registration backends.

Joeri will ask if the feature "payment after registration" is still needed.

alextreme:
@joeribekker voorstel om dit een stuk kleiner te maken: 

Nu staat er "Totaalbedrag te voldoen"

Ik stel voor om dit te hernoemen naar: "Totaalbedrag"

Dan werkt het voor beide situaties en is het issue verholpen

sergei-maertens:
> [@joeribekker](https://github.com/joeribekker) voorstel om dit een stuk kleiner te maken:
> 
> Nu staat er "Totaalbedrag te voldoen"
> 
> Ik stel voor om dit te hernoemen naar: "Totaalbedrag"
> 
> Dan werkt het voor beide situaties en is het issue verholpen

@vaszig let's make a small PR with this language change and that should relieve the pressure on this ticket

joeribekker:
removing from 2.8 because in 2.7 we made the text more generic that should at least it is correct (however its not detailed).

PR #4139, opened 2024-04-09
[#4080] Add timestamp to PDF report

Viicos:
Part of #4080

### DIFF
--- a/src/openforms/submissions/templates/report/submission_report.html
+++ b/src/openforms/submissions/templates/report/submission_report.html
@@ -36,6 +36,7 @@
                 <p>{% blocktrans with submission_date=report.submission.completed_on trimmed %}
                     Submitted on: {{ submission_date }}
                 {% endblocktrans %}</p>
+                <p>{% trans "Report created on:" %} {% now "DATETIME_FORMAT" %}</p>
                 <p>{% blocktrans with ref=report.submission.public_registration_reference trimmed %}
                     Your reference is: {{ ref }}
                 {% endblocktrans %}</p>
--- a/src/openforms/submissions/tests/test_tasks_pdf.py
+++ b/src/openforms/submissions/tests/test_tasks_pdf.py
@@ -4,6 +4,7 @@ from django.test import TestCase, override_settings
 from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 
+from freezegun import freeze_time
 from privates.test import temp_private_root
 from pyquery import PyQuery as pq
 from testfixtures import LogCapture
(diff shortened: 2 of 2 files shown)


src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -36,6 +36,7 @@ <h1 class="title">{{ report.form.name }}</h1>
                 <p>{% blocktrans with submission_date=report.submission.completed_on trimmed %}
                     Submitted on: {{ submission_date }}
                 {% endblocktrans %}</p>
+                <p>{% trans "Report created on:" %} {% now "DATETIME_FORMAT" %}</p>
```
sergei-maertens, line 39: please apply the blocktrans pattern like above so that we can arrange the order of sentences in the translations.

You also want to let the normal rendering of django template engine apply for the timestamp so that its properly localized for the respective languages.

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -236,10 +237,32 @@ def test_public_reference_included(self):
...
+    def test_timestamp_included(self):
+        submission = SubmissionFactory.from_components(
+            [
+                {
+                    "key": "input",
+                    "label": "Input",
+                    "type": "textfield",
+                },
+            ],
+            with_report=True,
```
sergei-maertens, line 255: I think you can remove this overhead if you're rendering it later on anyway.

```suggestion
```

sergei-maertens (changes requested):

src/openforms/submissions/templates/report/submission_report.html
Viicos: I couldn't find how to make it work (`{% blocktrans with creation_date={% now ... %} %}` is invalid and no workarounds are mentioned in docs).

> You also want to let the normal rendering of django template engine apply for the timestamp so that its properly localized for the respective languages.

This is already picked up by `DATETIME_FORMAT`, which follows the current locale

sergei-maertens: docs: https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#now

Use `{% now '...' as now_str %}` and then use `{{ now_str }}` in the blocktranslate.

> This is already picked up by DATETIME_FORMAT, which follows the current locale

I missed that you apparently *must* provide a format string to the tag. Kinda stupid.

Viicos: > docs: https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#now

:facepalm: 

(a bit weird to have `with` and `as` as two different options to do aliases)

sergei-maertens (approved):

PR #4142, opened 2024-04-09
Disable sending hidden fields to objects API

### LINKED ISSUE #4140: Opt-out into sending hidden fields to Objects API
Bugfix for #3890 is causing issues for other environments, so on 2.6.x (ONLY) we add an envvar to opt out of this behaviour.

sergei-maertens:
Closes #4140

Enabling the envvar DISABLE_SENDING_HIDDEN_FIELDS opts out of the bugfix that sends hidden fields to the objects API registration backend, restoring the old (but broken) behaviour. This buys organizations time to update their json schemas.

This fix will only be present in Open Forms 2.6.3+, newer versions enforce the earlier bugfix behaviour.

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/tests/test_template.py
+++ b/src/openforms/registrations/contrib/objects_api/tests/test_template.py
@@ -402,3 +402,94 @@ class JSONTemplatingRegressionTests(SubmissionsMixin, TestCase):
                 },
             },
         )
+
+    @tag("dh-673", "gh-4140")
+    @override_settings(DISABLE_SENDING_HIDDEN_FIELDS=True)
+    def test_opt_out_of_sending_hidden_fields(self):
+        submission = SubmissionFactory.from_components(
+            components_list=[
+                {
+                    "type": "radio",
+                    "key": "radio",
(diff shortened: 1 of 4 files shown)


src/openforms/registrations/contrib/objects_api/tests/test_template.py
```diff
@@ -402,3 +402,94 @@ def test_object_nulls_regression(self):
...
+                    "objecttype_version": 300,
+                    # skip document uploads
+                    "informatieobjecttype_submission_report": "",
+                    "upload_submission_csv": False,
+                    "informatieobjecttype_attachment": "",
+                },
+            )
+
+        _objects_client.create_object.mock_assert_called_once()
+        _objects_client.create_object.call_args
```
Viicos, line 481: Is this empty statement required?

Viicos (approved):

src/openforms/registrations/contrib/objects_api/tests/test_template.py
sergei-maertens: nope, will clean up in master, it was from my debugger

PR #4144, opened 2024-04-09
Apply partial input validation on submission step put

### LINKED ISSUE #4143: Apply validations (except for required) on submission step PUT endpoint
This should help enforce invalid data input (e.g. according to regex) while still allowing pausing of forms (which must support incomplete data).

sergei-maertens:
Closes #4143

### DIFF
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -1,3 +1,4 @@
+import copy
 import logging
 from dataclasses import dataclass
 from datetime import timedelta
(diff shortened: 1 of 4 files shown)


src/openforms/submissions/api/serializers.py
```diff
@@ -221,6 +224,8 @@ class SubmissionStepSerializer(NestedHyperlinkedModelSerializer):
         "submission_uuid": "submission__uuid",
     }
 
+    instance: SubmissionStep
```
Viicos, line 227: `class SubmissionStepSerializer(NestedHyperlinkedModelSerializer[SubmissionStep]):` can be used to infer the `instance` type as well (but it seems to be either `SubmissionStep | Sequence[SubmissionStep]` from stubs so not really useful)

Viicos (approved):
So if I understand correctly, this will allow:
- _Not_ filling all the fields in a step (to allow pausing the form)

and enforce:
- validation of fields that have been filled in

?

So users won't be able to pause a submission if they have set invalid values on some fields?

Edit: seems that this broke some e2e tests

sergei-maertens:
> So if I understand correctly, this will allow:
> 
> * _Not_ filling all the fields in a step (to allow pausing the form)
> 
> and enforce:
> 
> * validation of fields that have been filled in
> 
> ?
> 
> So users won't be able to pause a submission if they have set invalid values on some fields?
> 
> Edit: seems that this broke some e2e tests

Correct. I will look into the e2e failures

sergei-maertens:
Backport: 2310f9e03831ec7748944ab57fc64a190a412bcb

PR #4148, opened 2024-04-10
[#4052] Don't trigger 'on completion' task chain if payment was unsuccessful

### LINKED ISSUE #4052: Payment e-mail reminder after cancel arrives after 15 minutes and 3 times.

2.5.4 - 2.6.0

image

start payment process
Cancel payment
wait 15 minutes
image

Update 29-03: 
On success there also seems to be 2 confirmations. 1 directly after payment and again after 15 minutes a identical message. 

To Fix:
- [x] Confirmation email being sent also when payment is cancelled. https://github.com/open-formulieren/open-forms/pull/4148
- [x] Confirmation emails being sent immediately AND after 15 min timeout. https://github.com/open-formulieren/open-forms/pull/4157

SilviaAmAm:
Fixes #4052

This fixes the issues that multiple emails are sent if a payment is cancelled.

It does NOT fix the fact that if you complete the payment successfully you get both an immediate email and one after the timeout.

### DIFF
--- a/src/openforms/payments/tests/test_views.py
+++ b/src/openforms/payments/tests/test_views.py
@@ -1,15 +1,18 @@
 from decimal import Decimal
 from unittest.mock import patch
 
-from django.http import HttpResponseRedirect
-from django.test import TestCase, override_settings
+from django.http import HttpRequest, HttpResponse, HttpResponseRedirect
+from django.test import TestCase, override_settings, tag
 from django.test.client import RequestFactory
+
+from rest_framework import status
 
 from openforms.config.models import GlobalConfiguration
 from openforms.submissions.constants import PostSubmissionEvents
 from openforms.submissions.tests.factories import SubmissionFactory
 
 from ..base import BasePlugin, PaymentInfo
+from ..constants import PaymentStatus
 from ..models import SubmissionPayment
 from ..registry import Registry
 from .factories import SubmissionPaymentFactory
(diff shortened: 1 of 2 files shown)


Viicos (approved):

src/openforms/payments/tests/test_views.py
```diff
@@ -205,3 +212,148 @@ def test_start_plugin_not_enabled(
...
+        webhook_url = self.plugin.get_webhook_url(base_request)
+
+        with self.captureOnCommitCallbacks(execute=True) as callbacks:
+            response = self.client.post(
+                webhook_url, data={"orderID": payment.public_order_id}
+            )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        # Payment was successful, so 'transaction.on_commit' is called and triggers the on_post_submission_event
+        self.assertEqual(len(callbacks), 1)
```
sergei-maertens, line 306: I have some mixed feelings about these assertions - they will break as soon as another `on_commit` handler is added anywhere in the stack (so it could be a call downstream, or even a call via middleware that adds an on_commit handler completely unrelated to the problem at hand here).

I think I'd rather opt for a `patch("openforms.payments.views.on_post_submission_event")` followed by a `mock_on_post_submission_event.assert_called_once_with(submission.pk, PostSubmissionEvents.on_payment_complete)`, and for the test above an assertion that it's *not* called.

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.6.x: f63762d0e929c282dd6a746fffcef0a6dfb63fc6
* 2.5.x: a5a209eaa58950a04c73eefa79c9ac30c46b8c2a

Order versions aren't affected.

PR #4154, opened 2024-04-11
[#3725] Add prefill failures to email digest

### LINKED ISSUE #3725: Detect more problems and add them to the email digest

vaszig:
Partly fixes #3725 

This PR handles the prefill plugins (second subtask)

### DIFF
--- a/src/openforms/emails/tasks.py
+++ b/src/openforms/emails/tasks.py
@@ -10,6 +10,7 @@ from openforms.celery import app
 from openforms.config.models import GlobalConfiguration
 from openforms.logging.service import (
     collect_failed_emails,
+    collect_failed_prefill_plugins,
     collect_failed_registrations,
 )
 
@@ -23,17 +24,17 @@ class Digest:
     def get_context_data(self) -> dict[str, Any]:
         failed_emails = collect_failed_emails(self.since)
         failed_registrations = collect_failed_registrations(self.since)
-
-        if not (failed_emails or failed_registrations):
-            return {}
+        failed_prefill_plugins = collect_failed_prefill_plugins(self.since)
 
         return {
             "failed_emails": failed_emails,
             "failed_registrations": failed_registrations,
+            "failed_prefill_plugins": failed_prefill_plugins,
         }
 
     def render(self) -> str:
-        if not (context := self.get_context_data()):
+        context = self.get_context_data()
+        if not any(context.values()):
             return ""
 
         return render_to_string("emails/admin_digest.html", context)
--- a/src/openforms/emails/templates/emails/admin_digest.html
+++ b/src/openforms/emails/templates/emails/admin_digest.html
@@ -22,15 +22,38 @@
     <ul>
         {% for registration in failed_registrations %}
             <li>
-                {% blocktranslate with form_name=registration.form_name counter=registration.failed_submissions_counter first_failure_at=registration.initial_failure_at|time:"H:i" last_failure_at=registration.last_failure_at|time:"H:i" trimmed %}
-                    Form '{{ form_name }}' failed {{ counter }} times between {{ first_failure_at }} and {{ last_failure_at }}.</br>
-                {% endblocktranslate %}
-            
+                <p>
+                    {% blocktranslate with form_name=registration.form_name counter=registration.failed_submissions_counter first_failure_at=registration.initial_failure_at|time:"TIME_FORMAT" last_failure_at=registration.last_failure_at|time:"TIME_FORMAT" trimmed %}
+                        Form '{{ form_name }}' failed {{ counter }} time(s) between {{ first_failure_at }} and {{ last_failure_at }}.</br>
+                    {% endblocktranslate %}
+                </p>
                 <a href="{{ registration.admin_link }}">
                     {% blocktranslate with form_name=registration.form_name %}
                         View failed '{{ form_name }}' submissions
                     {% endblocktranslate %}
                 </a>
+            </li>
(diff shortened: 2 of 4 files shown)


src/openforms/emails/tasks.py
```diff
@@ -23,13 +24,15 @@ def __init__(self, since: datetime) -> None:
     def get_context_data(self) -> dict[str, Any]:
         failed_emails = collect_failed_emails(self.since)
         failed_registrations = collect_failed_registrations(self.since)
+        failed_prefill_plugins = collect_failed_prefill_plugins(self.since)
 
-        if not (failed_emails or failed_registrations):
+        if not (failed_emails or failed_registrations or failed_prefill_plugins):
```
Viicos, line 29: I'm wondering if this code path is necessary: if we remove it, the next return statement will be something like:

```python
{
    "failed_emails": {},
    "failed_registrations": {},
    "failed_prefill_plugins": {},
}
```

And the rendered template content will be the same as the template checks for truthy values on these variables

src/openforms/logging/service.py
```diff
@@ -75,3 +90,60 @@ def collect_failed_registrations(
...
+    query_params = {
+        "content_type": content_type,
+        "object_id__in": ",".join(submission_ids),
+        "extra_data__log_event__in": "prefill_retrieve_empty,prefill_retrieve_failure",
+    }
+    submissions_admin_url = furl(reverse("admin:logging_timelinelogproxy_changelist"))
+    return submissions_admin_url.add(query_params).url
+
+
+def collect_failed_prefill_plugins(since: datetime) -> list[dict[str, FailedPrefill]]:
```
Viicos, line 105:
```suggestion
def collect_failed_prefill_plugins(since: datetime) -> dict[str, FailedPrefill]:
```

src/openforms/logging/service.py
```diff
@@ -75,3 +90,60 @@ def collect_failed_registrations(
...
+    logs = TimelineLogProxy.objects.filter(
+        timestamp__gt=since,
+        extra_data__log_event__in=[
+            "prefill_retrieve_empty",
+            "prefill_retrieve_failure",
+        ],
+    )
```
Viicos, line 112: I seem to remember Sergei said we should filter on the template name instead of the log event, but my memory might be playing tricks on me

src/openforms/logging/service.py
```diff
@@ -75,3 +90,60 @@ def collect_failed_registrations(
...
+
+    if not logs:
+        return []
```
Viicos, line 115: This will be inconsistent with the other return value type, I think it can be removed if the rest of the code safely returns an empty dict.

src/openforms/logging/service.py
```diff
@@ -75,3 +90,60 @@ def collect_failed_registrations(
...
+
+    content_type = ContentType.objects.get_for_model(Submission).id
+
+    sorted_prefill_plugins = sorted(logs, key=lambda x: x.extra_data["plugin_label"])
```
Viicos, line 119: Can't this sorting be done at the database level?

Viicos: Maybe the name of the variable should be `sorted_logs`?

src/openforms/logging/service.py
```diff
@@ -75,3 +90,60 @@ def collect_failed_registrations(
...
+        timestamps = []
+        forms = set()
+        submission_ids = []
+        errors = False
+
+        for log in logs:
+            timestamps.append(log.timestamp)
+            forms.add(log.content_object.form.name)
+            submission_ids.append(str(log.content_object.id))
+            errors = True if log.extra_data.get("error") else False
```
Viicos, line 138:
```suggestion
            errors = bool(log.extra_data.get("error"))
```

src/openforms/logging/service.py
```diff
@@ -75,3 +90,60 @@ def collect_failed_registrations(
...
+
+        failed_prefill_plugins[prefill_plugin] = {
```
Viicos, line 140: Maybe you meant to use the dataclass defined in the same file

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -31,6 +31,34 @@ <h5>{% trans "Registrations" %}</h5>
...
+        {% endfor %}
+    </ul>
+{% endif %}
+
+{% if failed_prefill_plugins %}
+    <h5>{% trans "Prefill plugins" %}</h5>
+    <ul>
+        {% for failed_plugin, failed_plugin_data in failed_prefill_plugins.items %}
+            <li>
+                {% blocktranslate with counter=failed_plugin_data.counter initial_failure_at=failed_plugin_data.initial_failure_at|time:"H:i" last_failure_at=failed_plugin_data.last_failure_at|time:"H:i" trimmed %}
```
Viicos, line 47:
```suggestion
                {% blocktranslate with counter=failed_plugin_data.counter initial_failure_at=failed_plugin_data.initial_failure_at|time:"TIME_FORMAT" last_failure_at=failed_plugin_data.last_failure_at|time:"TIME_FORMAT" trimmed %}
```

So that it can be localized?

src/openforms/logging/service.py
vaszig: Sergei realised that filtering the log_event is okay because it's a generic wrapper taking care of it and that his earlier remark can/should be discarded :) (Sergei is also typing this)

sergei-maertens (approved):

PR #4163, opened 2024-04-12
[#3725] Add detection of BRP broken configuration concerning addressNL component

### LINKED ISSUE #3725: Detect more problems and add them to the email digest

vaszig:
Partly fixes #3725

This handles the third subtask


src/openforms/forms/models/form.py
```diff
@@ -603,6 +603,11 @@ def deactivate(self):
         logger.debug("Deactivating form %s", self.admin_name)
         self.save(update_fields=["active", "deactivate_on"])
 
+    def form_uses_component(self, component_type: str) -> bool:
```
vaszig, line 606: "form" in the name is a bit redundant, I will fix that in the feedback, if you agree, and this method will probably be updated since we may need to handle multiple types in other clients.

src/openforms/config/service.py
```diff
@@ -0,0 +1,39 @@
...
+from openforms.plugins.exceptions import InvalidPluginConfiguration
+
+
+@dataclass
+class BrokenConfiguration:
+    config_name: str
+    exception_message: str | None
+
+
+def check_BRK_config_for_addressNl() -> str | None:
```
Viicos, line 14:
```suggestion
def check_brk_config_for_addressnl() -> str | None:
```

I think it's better to stick with snake case, although `addressnl` feels a bit weird; so maybe `addressNL` is fine

src/openforms/config/service.py
```diff
@@ -0,0 +1,39 @@
...
+
+    for form in live_forms:
+        if form.form_uses_component("addressNL"):
+            try:
+                BRKValidatorCheck.check_config()
+            except InvalidPluginConfiguration as e:
+                return e.args[0]
+            break
+
+    return
```
Viicos, line 25: I think something like this might be more readable:

```suggestion
    if any(form.form_uses_component("addressNL") for form in live_forms):
        try:
            BRKValidatorCheck.check_config()
        expect InvalidPluginConfiguration as e:
            return e.args[0]
```

`any` will stop as soon as it encounters a truthy value so performance is the same here

Viicos: On a side note, the way `InvalidPluginConfiguration` is being defined makes it unsafe to use `e.args[0]`, probably at some point we could define a proper `__init__` for this exception class

src/openforms/config/service.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def collect_broken_configurations() -> list[BrokenConfiguration]:
+    check_BRK_configuration = check_BRK_config_for_addressNl()
```
Viicos, line 29:
```suggestion
    check_brk_configuration = check_BRK_config_for_addressNl()
```

src/openforms/config/service.py
```diff
@@ -0,0 +1,39 @@
...
+
+    broken_configurations = []
+    if check_BRK_configuration:
+        broken_configurations.append(
+            BrokenConfiguration(
+                config_name="BRK Client", exception_message=check_BRK_configuration
```
Viicos, line 35: Maybe the `config_name` should be a translatable string?

src/openforms/forms/models/form.py
Viicos: I also think `uses_component` is better. Or maybe even `has_component`?

src/openforms/config/service.py
vaszig: I think the best would be what you suggest (`addressNL`), in order to stick to the component's type as well.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -63,3 +63,17 @@ <h5>{% trans "Prefill plugins" %}</h5>
...
     </ul>
 {% endif %}
+
+{% if broken_configurations %}
+    <h5>{% trans "Broken configuration" %}</h5>
+    <ul>
+        {% for config in broken_configurations %}
+            <li>
+                {% blocktranslate with name=config.config_name message=config.exception_message trimmed %}
+                    Configuration for '{{ name }}' is broken (<i>{{ message }}</i>)<br/>
```
Viicos, line 73: In the `BRKValidatorCheck` class, I see some error messages with and without a dot. So we probably need to find a way to add one if required

Viicos (changes requested):

src/openforms/config/service.py
vaszig: Why is it unsafe?Could you please elaborate?

Viicos: It is defined as a simple subclass:
https://github.com/open-formulieren/open-forms/blob/fac58af05a9685d68068ba1f4609c748079d96ac/src/openforms/plugins/exceptions.py#L1-L2

Meaning I can do `e = InvalidPluginConfiguration()` and `e.args == tuple()`

src/openforms/config/service.py
```diff
@@ -0,0 +1,39 @@
...
+from django.utils.translation import gettext as _
+
+from openforms.contrib.brk.checks import BRKValidatorCheck
+from openforms.forms.models.form import Form
+from openforms.plugins.exceptions import InvalidPluginConfiguration
+
+
+@dataclass
+class BrokenConfiguration:
+    config_name: str
```
Viicos, line 12: You can use `StrOrPromise` from `openforms.typing`

```suggestion
    config_name: StrOrPromise
```

`gettext_lazy` returns a `Promise` object, so our `StrOrPromise` is a convenient type alias that accepts both `str` and `Promise` objects.

You'll also have to update the `gettext` import to use `gettext_lazy`

sergei-maertens (dismissed):

Viicos (approved):

PR #4167, opened 2024-04-12
[#3725] Add broken BAG configuration detection to email digest

### LINKED ISSUE #3725: Detect more problems and add them to the email digest

vaszig:
Partly closes #3725

This handles the 4th subtask


vaszig:
This should wait for #4163 to be approved because it touches some files in common.

src/openforms/emails/tests/test_tasks.py
```diff
@@ -399,3 +401,162 @@ def test_no_email_sent_when_brk_congiguration_is_invalid_but_unused(self):
...
+        sent_email = mail.outbox[0]
+
+        self.assertEqual(
+            sent_email.subject, "[Open Forms] Daily summary of detected problems"
+        )
+        self.assertEqual(sent_email.recipients(), ["user@example.com"])
+        self.assertIn(
+            _(
+                "The configuration for 'BAG Client' is invalid (bag_service endpoint is not configured)."
+            ),
```
sergei-maertens, line 460: This is not a proper translation message - you need to use the same placeholders like in the original template: https://github.com/open-formulieren/open-forms/blob/3f6e71284b92eefe014a82d3f5c00bc3d956d076/src/openforms/emails/templates/emails/admin_digest.html#L72

Something along the lines of:

```suggestion
            _(
                "The configuration for '%(name)s' is invalid (%(message)s)."
            ),
```
but this is flawed anyway, because this exact translation string doesn't exist, as it was created by stripping tags from the HTML. Since you force the language to be english in these tests, skip the marking-translatable here and just do a literal string assertion because you control the language of the message explicitly already:

```suggestion
            "The configuration for 'BAG Client' is invalid (bag_service endpoint is not configured).",
```

src/openforms/emails/tests/test_tasks.py
```diff
@@ -399,3 +401,162 @@ def test_no_email_sent_when_brk_congiguration_is_invalid_but_unused(self):
...
+                        "label": "Street name",
+                        "deriveCity": False,
+                        "derivePostcode": "postCode",
+                        "deriveStreetName": True,
+                        "deriveHouseNumber": "houseNumber",
+                    },
+                ],
+            }
+        )
+        FormStepFactory.create(form=form, form_definition=form_definition)
```
sergei-maertens, line 532: the preferred way to set up this data is to use the factory traits:

```py
form = FormFactory.create(
    generate_minimal_setup=True,
    formstep__form_definition__configuration={"components": [...]},
)
```

sergei-maertens (changes requested):

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -0,0 +1,227 @@
...
+        submission = SubmissionFactory.create(
+            form=form, registration_status=RegistrationStatuses.failed
+        )
+
+        with freeze_time("2023-01-03T01:00:00+01:00"), patch(
+            "openforms.emails.tasks.GlobalConfiguration.get_solo",
+            return_value=GlobalConfiguration(
+                recipients_email_digest=["tralala@test.nl"]
+            ),
+        ):
```
sergei-maertens, line 123: I suppose these can be moved to a decorator of the test?

```py
@patch("openforms.emails.tasks.GlobalConfiguration.get_solo", return_value=...)
@freeze_time("2023-01-03T01:00:00+01:00")
@override_settings(...)
def test_email_sent_when_there_are_failures(self, *mocks):
    ...
```
this can considerably flatten the test :)

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -0,0 +1,227 @@
...
+                with freeze_time("2023-01-02T12:30:00+01:00"):
+                    logevent.email_status_change(
+                        submission,
+                        event="registration",
+                        status=Message.STATUS_FAILED,
+                        status_label="Failed",
+                        include_in_daily_digest=True,
+                    )
+
+                send_email_digest()
```
sergei-maertens, line 134: I'd set up everything in the beginning of the test, then send the digest once and then make assertions:

```
create form and failed submission
create email status change (failure)
create failed registration log
create failed prefill log

send_email_digest()

assert outgoing email with correct recipient(s)
sent_email = mail.outbox[-1]

with self.subTest("failed registrations")
    assert failed registration

with self.subTest(...)
    assert other thing ...
```

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -0,0 +1,209 @@
...
+            ),
+        ):
+            send_email_digest()
+
+        self.assertEqual(0, len(mail.outbox))
+
+    @patch(
+        "openforms.emails.tasks.GlobalConfiguration.get_solo",
+        return_value=GlobalConfiguration(recipients_email_digest=["tralala@test.nl"]),
+    )
```
sergei-maertens, line 107: I think this can be moved to the test case class as a decorator, since it applies for all tests. For the `test_no_email_sent_if_no_recipients` test, you can then do:

```py
def test_no_email_sent_if_no_recipients(self, mock_global_config):
    mock_global_config.return_value.recipients_email_digest = []
```

src/openforms/emails/tests/test_digest_functions.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def test_successful_registrations_are_not_collected(self):
+        form = FormFactory.create()
+        SubmissionFactory.create(form=form, registration_success=True)
+
+        failed_registrations = collect_failed_registrations(
+            since=datetime(2023, 1, 2, 14, 30, 0).replace(tzinfo=utc)
+        )
+
+        self.assertEqual(failed_registrations, [])
```
sergei-maertens, line 134: I don't think any log events are created here for succesful registration, so this test would always pass.

src/openforms/emails/tests/test_digest_functions.py
```diff
@@ -0,0 +1,405 @@
...
+        logevent.registration_failure(
+            submission,
+            RegistrationFailed("Registration plugin is not enabled"),
+        )
+
+        failed_registrations = collect_failed_registrations(
+            since=datetime(2023, 1, 4, 1, 0, 0).replace(tzinfo=utc)
+        )
+
+        self.assertEqual(failed_registrations, [])
```
sergei-maertens, line 149: I don't see this 24h aspect in the `collect_failed_registrations` function, it's just querying based on `since`. 

I don't think that this aspect is really testable, it all comes down to "are we passing the right `since`" parameter.

You *could* perhaps do:

```py
digest = Digest(since=datetime(2023, 1, 4, 1, 0, 0).replace(tzinfo=utc))
context = digest.get_context_data()

assert context["failed_emails"] == []
```
which checks that the contraint is passed along properly.

src/openforms/emails/tests/test_digest_functions.py
```diff
@@ -0,0 +1,405 @@
...
+        form = FormFactory.create()
+        submission = SubmissionFactory.create(form=form)
+
+        logevent.prefill_retrieve_empty(submission, hc_plugin, ["burgerservicenummer"])
+
+        failed_plugins = collect_failed_prefill_plugins(
+            since=datetime(2023, 1, 4, 2, 0, 0).replace(tzinfo=utc)
+        )
+
+        self.assertEqual(failed_plugins, [])
```
sergei-maertens, line 203: Similar remark as above

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #4168, opened 2024-04-15
[#4156] Use a template for payment reference

### LINKED ISSUE #4156: Payment transaction should be customizable using a template
Discussed with Utrecht (Taiga 200), and the solution as described with the number of attempts wasn't done conform spec of this Github issue:

"we kunnen de prefix met het jaartal er wel afhalen, maar wat na het zaaknummer komt staat nu als volgt gedefinieerd OF-nummer/xxx. Dat moet OF-nummer_xxx zoals ook in het betreffende github issue stond"

_[We can remove the prefix with the year, but what comes after the case number is now defined as follows OF-number/xxx. This must be OF-number_xxx as stated in the relevant GitHub issue.]_

_Originally posted by @alextreme in https://github.com/open-formulieren/open-forms/issues/3793#issuecomment-2049229362_

---

Since 2.6.0, the payment reference now follows this pattern:

`{prefix}/{public_registration_reference}/{payment.pk}`

For example, with a prefix of `{year}`: `2024/OF-123456/945`
- `OF-123456` is a registration reference generated in OF, but can be a zaak number as well.
- `945` is a unique incrementing ID (the database PK), which creates confusion as this is NOT the payment retry attempt number.

> This must be OF-number_xxx

According to https://github.com/open-formulieren/open-forms/issues/3793 this wasn't a strong requirement, and we had a checkbox stating _Determine if an attempt postfix is needed (for example when you cancel the first payment, and make another attempt)._

@alextreme is there a specific reason to have it in the `_XXX` format? Should it be the retry attempt number? I think we went for the payment database PK to be certain it would generate a unique number, but probably this could work as well with the payment retry?

Viicos:
Closes #4156
Replaces #4166
Replaces #4160


src/openforms/config/migrations/0001_initial_to_v250.py
```diff
@@ -24,6 +26,17 @@
...
 
+# Function removed from the code, moved here to not break the migration:
+def validate_payment_order_id_prefix(value: str):
+    value = value.replace("{year}", "")
+    if value and not value.isalnum():
+        raise ValidationError(
+            _(
+                "Prefix must be alpha numeric, no spaces or special characters except {year}"
+            )
+        )
```
sergei-maertens, line 37: I think you can remove the implementation entirely in historical migrations with just a `pass`, they don't do anything anyway.

src/openforms/config/migrations/0002_squashed_to_of_v230.py
```diff
@@ -17,6 +19,17 @@
 import openforms.utils.translations
 
 
+# Function removed from the code, moved here to not break the migration:
+def validate_payment_order_id_prefix(value: str):
```
sergei-maertens, line 23: same as above

src/openforms/config/migrations/0058_migrate_to_order_id_template.py
```diff
@@ -0,0 +1,46 @@
...
+    Otherwise, the default value is adapted to have the ``{uid}`` placeholder included.
+    """
+    GlobalConfiguration = apps.get_model("config", "GlobalConfiguration")
+    VersionInfo = apps.get_model("upgrades", "VersionInfo")
+    config = GlobalConfiguration.objects.first()
+    if config is None:
+        return
+
+    version_info = VersionInfo.objects.first()
+    assert version_info is not None
```
sergei-maertens, line 25: because we have the system check that creates it, right?

Technically you can end up in a situation without record when system checks are completely disabled. Edge case, but not impossible. In that case however, you can also conclude that it's a fresh installation (because the post-migrate hook has never run).

src/openforms/config/migrations/0058_migrate_to_order_id_template.py
```diff
@@ -0,0 +1,46 @@
...
+
+    is_new_deploy = version_info.current == settings.RELEASE
```
sergei-maertens, line 27:
```suggestion
    is_new_deploy = version_info is None or version_info.current == settings.RELEASE
```

src/openforms/config/migrations/0058_migrate_to_order_id_template.py
```diff
@@ -0,0 +1,46 @@
...
+    if is_new_deploy:
+        # The default value of the `payment_order_id_template` field
+        # will take effect
+        return
+
+    config.payment_order_id_template = config.payment_order_id_prefix + "{uid}"
+    config.save()
```
sergei-maertens, line 34: I think we have some migrations that flush the solo cache when saving it, since this `save` method is not the one with the cache-update hooks from django-solo.

src/openforms/config/migrations/0059_remove_globalconfiguration_payment_order_id_prefix.py
sergei-maertens, line 1: any chance you can merge these three migration files into a single file? Makes it easier to forward-port.

src/openforms/config/models/config.py
```diff
@@ -321,15 +321,18 @@ class GlobalConfiguration(SingletonModel):
...
+        _("Payment Order ID template"),
+        max_length=48,
+        default="{year}/{reference}/{uid}",
+        help_text=_(
+            "Template to use when generating payment order IDs. It should be alpha-numerical and can contain the '/._-' characters. "
+            "The following placeholders are supported:\n"
+            "  - {year}: The current year\n"
+            "  - {reference}: The submission reference\n"
+            "  - {uid}: A unique incrementing payment ID."
+        ),
```
sergei-maertens, line 334: Newlines don't translate to HTML. Please write your help texts in a more simple form, and add verbose Sphinx documentation to provide more context/background. Your last three lines can be replaced with: "You can use the placeholder tokens: {year}, {public_reference}, {uid}."

src/openforms/config/models/config.py
```diff
@@ -321,15 +321,18 @@ class GlobalConfiguration(SingletonModel):
...
-        max_length=16,
-        default="{year}",
-        blank=True,
-        help_text=_(
-            "Prefix to apply to generated numerical order IDs. Alpha-numerical only, supports placeholder {year}."
-        ),
-        validators=[validate_payment_order_id_prefix],
+    payment_order_id_template = models.CharField(
+        _("Payment Order ID template"),
+        max_length=48,
```
sergei-maertens, line 326: what's the reason for this magic number?

src/openforms/config/models/config.py
```diff
@@ -321,15 +321,18 @@ class GlobalConfiguration(SingletonModel):
...
+        default="{year}/{reference}/{uid}",
```
sergei-maertens, line 327:
```suggestion
        default="{year}/{public_reference}/{uid}",
```
I'd really prefer using the same variable names consistently across the project. All other templates use `public_reference`.

src/openforms/payments/validators.py
```diff
@@ -2,11 +2,12 @@
 from django.utils.translation import gettext as _
 
 
-def validate_payment_order_id_prefix(value: str):
-    value = value.replace("{year}", "")
+def validate_payment_order_id_template(value: str) -> None:
+    if "{uid}" not in value:
+        raise ValidationError(_("The template should include the {uid} placeholder."))
```
sergei-maertens, line 7:
```suggestion
        raise ValidationError(_("The template must include the {uid} placeholder."))
```

src/openforms/payments/validators.py
```diff
@@ -2,11 +2,12 @@
...
+    for allowed in ["{year}", "{reference}", "{uid}", "/", ".", "_", "-"]:
+        value = value.replace(allowed, "")
+
     if value and not value.isalnum():
-        raise ValidationError(
-            _(
-                "Prefix must be alpha numeric, no spaces or special characters except {year}"
-            )
-        )
+        raise ValidationError(_("The template should be alphanumeric."))
```
sergei-maertens, line 13:
```suggestion
        raise ValidationError(_("The template may only consist of alphanumeric, /, ., _ and - characters."))
```

sergei-maertens (changes requested):

src/openforms/config/models/config.py
Viicos: It's a reasonable to default, not too long to avoid having runtime errors with ogone

src/openforms/config/migrations/0058_migrate_to_order_id_template.py
Viicos: Fixed with the suggestion you gave below

src/openforms/config/models/config.py
sergei-maertens: I'm trying to understand it, because the placeholder tokens will shrink/expand depending on the values

e.g. `{year}` will always be 4 chars, but `{public_reference}` can be a long `ZAAK-123456789` or short `OF-123456` number, while `uid` will realistically take 1-5 chars. But the tokens/placeholders don't cleanly map to lengths of their values, so I think we can't really make any assumptions/guarantees about avoiding runtime Ogone errors.

Viicos: Yeah I was thinking the same it's a bit hard to manage. Should we go with 255?

sergei-maertens: no need, we can increase it if it's too small - just add a comment that there's no particular reason for 48 and that the template substitutions make this tricky anyway.

sergei-maertens (approved):

Issue #3154, opened 2023-06-13
Robuuste datastructuren voor het verwerken van inzendingsgegevens


CharString:
### Thema / Theme

Other

### Omschrijving / Description

`SubmissionState` and `SubmissionValueVariablesState` are "cache" classes that optimize read access to the database during API calls like "Store submission step data" and "Apply/check form logic". Their current implementation has grown to be a bit problematic.

Invalidation of these states isn't handled correctly and since references to them (and their innards) are copied, we might bump into more bugs like #3114 

One short sighted possible solution for `SubmissionState` is have `forms_steps` and `submission_steps` not be of `List`, but of `Sequence` that has implementations of that protocol that check cache staleness.

Viicos:
We will track all our wishes regarding the usage of this "public API service" in this meta-ticket

sergei-maertens:
I think roughly all `DataMapping` type hints should support `FormioData` instead, so that we don't need raw `glom` usage for Formio data structures anymore.

The goal here is really that you can just use `component["key"]` to access the values without having to worry about any nesting (and the same for assigning values).

sergei-maertens:
Some thoughts from me on what is needed/desired, completely ignoring what we currently have.

**First, a history lesson**

* Open Forms 1.x did not have explicit variables, only implicit variables derived from
  the formio configuration. So, we needed tools to extract the python values and
  string representations of those values from the submitted JSON data

* Open Forms 1.x stored whatever user input for a step was provided in a `data` JSON
  field, getting all the data for a submission was then taken by merging those objects
  for each step

* Open Forms 2.0 introduced variables via a model, while keeping the API interface the
  same (submit JSON `data` key and then it gets translated to the underlying DB structure)

* We also had some mechanisms/logic to figure out where in the form we are exactly (
  step completed, optional... paused etc.). This is the submission state.

* To handle the transition in code from OF 1.x to 2.x we added helpers to deal with the
  variables state. We also gradually moved more stuff to raw variables instead of relying
  on the formio components

Because variables were essentially retrofitted and we continued to iterate on them,
datastructures grew and got expanded as needs came to be. E.g. static variables were not
a part of Open Forms 2.0 (I think), they were added later. Prefill used to run for every
step, but with variables we could move that to the submission start and save the value
in a variable and not touch it anymore.

We tried to anticipate future needs/logic etc. by adding some type information, based on
our knowledge/experience of Formio at the time. In hindsight, this was not complete
[...]

PR #4179, opened 2024-04-16
[#4141] Experimental fix

### LINKED ISSUE #4141: Objects API v2 registration crashes with fieldsets

2.6.2

With the following configuration:

```python
    [
        {
            "key": "fieldset",
            "type": "fieldset",
            "components": [
                {
                    "key": "fieldset.textfield",
                    "type": "textfield"
                }
            ]
        },
    ],
    completed=True,
    submitted_data={
        "fieldset": {
            "textfield": "some_string",
        },
    },
)
```

The objects API v2 registration crashes:

```sh
openforms/submissions/models/submission_value_variable.py line 57, in get_variable
    return self.variables[key]
KeyError: 'fieldset'
```

Due to how the submission values variable state is being used

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

Viicos:
This is an experiment to fix the `FormioData` structure, it fixes #4141 and only makes around 20 tests to fail.

More details about the fix are added in the meta issue https://github.com/open-formulieren/open-forms/issues/3154.

I'm closing this one as this leads to dirty fixes for existing code (i.e. using the underlying `FormioData.data` dict as before) and some tests are tricky to fix.

### DIFF
--- a/src/openforms/formio/datastructures.py
+++ b/src/openforms/formio/datastructures.py
@@ -1,7 +1,7 @@
 import re
 from collections import UserDict
-from collections.abc import Hashable
-from typing import Iterator, cast
+from collections.abc import Hashable, ItemsView, Iterable, Iterator, KeysView
+from typing import Any, cast
 
 from glom import PathAccessError, assign, glom
 
@@ -88,7 +88,7 @@ class FormioConfigurationWrapper:
         return all(is_visible_in_frontend(node, values) for node in nodes)
 
 
-class FormioData(UserDict):
+class FormioData(UserDict[str, JSONValue]):
     """
     Handle formio (submission) data transparently.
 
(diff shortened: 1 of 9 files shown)


src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -242,7 +242,7 @@ def register_submission(
 
             extra_data = get_unmapped_data(
                 submission, self.zaak_mapping, REGISTRATION_ATTRIBUTE
-            )
+            ).data
             # The extraElement tag of StUF-ZDS expects primitive types
             extra_data = flatten_data(extra_data)
```
Viicos, line 247: Note that this `flatten_data` would become unnecessary, as this flattening is already done by the `FormioData` structure

src/openforms/formio/datastructures.py
```diff
@@ -107,12 +107,24 @@ class FormioData(UserDict):
...
 
-    def __getitem__(self, key: Hashable):
+        def unnest_inner(
+            data: dict[str, JSONValue], parent_key: str | None
+        ) -> Iterable[tuple[str, Any]]:
+            for k, v in data.items():
+                key = f"{parent_key}.{k}" if parent_key is not None else k
+                if isinstance(v, dict):
+                    yield from unnest_inner(v, key)
+                else:
```
Viicos, line 119: Probably lists could be handled as well.

src/openforms/formio/datastructures.py
```diff
@@ -121,3 +133,12 @@ def __contains__(self, key: Hashable) -> bool:
         except PathAccessError:
             return False
         return True
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self._unnest(self.data))
```
sergei-maertens, line 138: I don't think this is a good idea because it doesn't take editgrid into account at all.

I never had iterating over `FormioData` in mind as an intended usage, it was always intended to be used with the assumption that you know which key you need (so you know the key before-hand). I'm afraid this is blowing up the scope of this data structure way too much and will lead to other problems in the future.

Viicos: Hum makes sense, I added my reasoning in the meta issue, but I'll update it with this in mind

PR #4183, opened 2024-04-16
[#3964] Fix old forms with simple conditionals

### LINKED ISSUE #3964: FE logic: When the value of number field is used to trigger visibility, components are emptied when "Clear on hide" is checked

2.4.3 - 2.6.0

When the value of number/currency field is used in FE logic and the component that is made visible has: "Clear on hide" checked, the values entered into this field will be removed at validation (check_logic)

`if Getal = 0 show Tekstvlak`

pooof
Kooha-2024-03-01-21-23-53

If a different logic is used on the same Number component like: `if not Getal hide Tekstvlak`
This bug does not occur, so it seems to be specific to the value in the number field

# Discussion:

We changed the formio-builder so that for new forms the conditional value is not saved as a string but as a number.

However, this will not fix existing forms that have a configuration where the conditional value is already a string. For fixing these forms there were 2 possibilities:
- Migration: find forms that have this problem and fix the configuration. This migration is quite complex and would also have to run on import.
- Bandaid fix in the logic (like for checkboxes: https://github.com/open-formulieren/open-forms/blob/1d22bdc733f5fa854f4d07cfe357e62e35a82db0/src/openforms/formio/utils.py#L358). This is quite easy, although care needs to be taken with the int/float casting.

Considerations:
- Writing the migration feels like the 'right thing':  new components have fixed behaviour and you don't want to carry along a legacy fix forever. It's better to root it out as soon as possible.
- Backporting the migration is tricky

Conclusion:
- Migration for master
- Bandaid for backport

In addition:
- See if you can fix checkboxes so we can remove the bandaid fix in master

SilviaAmAm:
Fixes #3964 

To do:
- [x] Handle importing old forms
- [x] Check if you can fix checkboxes (needs https://github.com/open-formulieren/formio-builder/pull/148)
- [x] Write bandaid fix for older versions (other PR) https://github.com/open-formulieren/open-forms/pull/4193
- [x] Update Formio builder once checkboxes fix is merged


sergei-maertens:
A rebase on `master` will give you a shiny new formio-builder version :)

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1127,3 +1127,161 @@ def test_import_form_with_old_service_fetch_config(self):
...
+
+        self.assertTrue(isinstance(fixed_components[3]["conditional"]["eq"], int))
+        self.assertTrue(isinstance(fixed_components[4]["conditional"]["eq"], float))
+        self.assertTrue(isinstance(fixed_components[5]["conditional"]["eq"], str))
+        self.assertTrue(isinstance(fixed_components[6]["conditional"]["eq"], float))
+        self.assertTrue(isinstance(fixed_components[7]["conditional"]["eq"], float))
+        self.assertTrue(isinstance(fixed_components[8]["conditional"]["eq"], str))
+        self.assertTrue(
+            isinstance(fixed_components[9]["components"][0]["conditional"]["eq"], int)
+        )
```
sergei-maertens, line 1287: I think `self.assertIsInstance` exists, so you can simplify these assertions

src/openforms/forms/tests/test_migrations.py
```diff
@@ -287,3 +287,165 @@ def test_migration_fixes_issues(self):
...
+
+        self.assertTrue(isinstance(fixed_components[3]["conditional"]["eq"], int))
+        self.assertTrue(isinstance(fixed_components[4]["conditional"]["eq"], float))
+        self.assertTrue(isinstance(fixed_components[5]["conditional"]["eq"], str))
+        self.assertTrue(isinstance(fixed_components[6]["conditional"]["eq"], float))
+        self.assertTrue(isinstance(fixed_components[7]["conditional"]["eq"], float))
+        self.assertTrue(isinstance(fixed_components[8]["conditional"]["eq"], str))
+        self.assertTrue(
+            isinstance(fixed_components[9]["components"][0]["conditional"]["eq"], int)
+        )
```
sergei-maertens, line 377: can't we compare with the exact value rather than isinstance checks?

What if a number accidentally gets misinterpreted as hex or octals (wild guessing here on my part)?

src/openforms/forms/migrations/0098_fix_forms_conditionals.py
```diff
@@ -0,0 +1,42 @@
...
+
+def fix_forms_simple_conditionals(apps, schema_editor):
+    FormDefinition = apps.get_model("forms", "FormDefinition")
+
+    all_definitions = FormDefinition.objects.all()
+    if not len(all_definitions):
+        return
+
+    # Check for simple conditionals where the "when" is not an empty string
+    raw_pattern = r'"conditional": \{[\w\s",:]*"when": "(\w+)"[\w\s",:]*\}'
```
sergei-maertens, line 19: this looks like the exact same regex - why don't we define it as a module constant? You can even use `from django.utils.regex_helper import _lazy_re_compile` but I honestly don't expect a big import-time overhead from this.

I'm also not sure about this pattern - the spaces here seem to be an implementation detail, `"conditional":{"when": ...` would still be valid JSON, and the ordering of keys in JSON is undefined, so it feels a bit fragile. Do we even need this regex at all? Since the Formio.js SDK/builder adds all the keys (like `conditional`) anyway, I think we can assume that 90% of the components will have this present and would need to be checked anyway.

src/openforms/forms/tests/test_migrations.py
SilviaAmAm: I wanted to avoid doing equalities with floats, and then I kept the other consistent. But I can compare with the actual value.

src/openforms/forms/migrations/0098_fix_forms_conditionals.py
```diff
@@ -0,0 +1,35 @@
...
+
+def fix_forms_simple_conditionals(apps, schema_editor):
+    FormDefinition = apps.get_model("forms", "FormDefinition")
+
+    all_definitions = FormDefinition.objects.all()
+    if not len(all_definitions):
+        return
+
+    definitions_to_update = []
+    for definition in all_definitions:
```
sergei-maertens, line 18: (for future, this is fine for this PR)

1. try to favour `.iterator()` over `.all()` in data migrations when operating on potentially large datasets. it prevents the entire queryset from being loaded in memory and thus won't crash your application with OOM errors
2. the `len(all_definitions)` check isn't needed - you can replace it with a `queryset.exists()` for an optimized query if you absolute must prevent work being done, but in this case, I would just loop over the (empty) qs and only do the `bulk_update` if `definitions_to_update` is not empty. I wouldn't be surprised if django itself already skips any work to be done if an empty list is passed.

sergei-maertens (approved):
Let's wrap this up :)

PR #4193, opened 2024-04-17
[#3964] Bandaid to fix old forms with simple conditionals

### LINKED ISSUE #3964: FE logic: When the value of number field is used to trigger visibility, components are emptied when "Clear on hide" is checked

SilviaAmAm:
Fixes #3964 (partially)

### DIFF
--- a/src/openforms/submissions/tests/form_logic/test_modify_components.py
+++ b/src/openforms/submissions/tests/form_logic/test_modify_components.py
@@ -1046,6 +1046,75 @@ class ComponentModificationTests(TestCase):
 
         self.assertEqual(submission_step.data["textField"], "Test value")
 
+    @tag("gh-3964")
+    def test_number_frontend_logic(self):
+        form = FormFactory.create()
+        step = FormStepFactory.create(
+            form=form,
+            form_definition__configuration={
+                "components": [
+                    {
+                        "key": "number",
+                        "type": "number",
+                    },
+                    {
+                        "key": "currency",
+                        "type": "currency",
+                    },
+                    {
+                        "key": "anotherNumber",
+                        "type": "number",
+                    },
+                    {
+                        "key": "textField1",
+                        "type": "textfield",
+                        "clearOnHide": True,
+                        "conditional": {"eq": "1", "show": True, "when": "number"},
+                    },
+                    {
+                        "key": "textField2",
+                        "type": "textfield",
+                        "clearOnHide": True,
+                        "conditional": {"eq": "1", "show": True, "when": "currency"},
+                    },
+                    {
+                        "key": "textField3",
+                        "type": "textfield",
+                        "clearOnHide": True,
+                        "conditional": {
+                            "eq": "1.5",
+                            "show": True,
+                            "when": "anotherNumber",
+                        },
+                    },
+                ]
+            },
+        )
+
+        submission = SubmissionFactory.create(form=form)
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/tests/form_logic/test_modify_components.py
```diff
@@ -1046,6 +1046,57 @@ def test_checkbox_frontend_logic(self):
...
+                        "key": "textField1",
+                        "type": "textfield",
+                        "clearOnHide": True,
+                        "conditional": {"eq": "1", "show": True, "when": "number"},
+                    },
+                    {
+                        "key": "textField2",
+                        "type": "textfield",
+                        "clearOnHide": True,
+                        "conditional": {"eq": "1", "show": True, "when": "currency"},
```
Viicos, line 1074:
```suggestion
                        "conditional": {"eq": "1.0", "show": True, "when": "currency"},
```
Maybe a test with a `float`?

Viicos (approved):

src/openforms/submissions/tests/form_logic/test_modify_components.py
SilviaAmAm: I was worried of potential flakiness due to comparing floats :thinking:

Viicos: I think it should be alright, as long as you don't perform any calculations on the floating numbers

And if we have flakyness in the test we will have it at runtime as well so I think this is a good thing :)

SilviaAmAm: Yes! Although I remeber at some point we had discouraged clients to do comparisons on with the floating points, so hopefully it shouldn't happen too often.

Viicos: If this becomes an issue we could probably use decimals to do the comparison

SilviaAmAm:
Backports:
- 2.5.x: https://github.com/open-formulieren/open-forms/pull/4195
- 2.4.x: https://github.com/open-formulieren/open-forms/pull/4196

PR #4198, opened 2024-04-18
[#3725] Add invalid forms (configuration and logic rules) to the email digest

### LINKED ISSUE #3725: Detect more problems and add them to the email digest

vaszig:
Fixes #3725 partially

### DIFF
--- a/src/openforms/emails/digest.py
+++ b/src/openforms/emails/digest.py
@@ -12,14 +12,24 @@ from django.utils.translation import gettext_lazy as _
 
 from django_yubin.models import Message
 from furl import furl
+from rest_framework import serializers
 from simple_certmanager.models import Certificate
 
 from openforms.contrib.brk.service import check_brk_config_for_addressNL
 from openforms.contrib.kadaster.service import check_bag_config_for_address_fields
+from openforms.forms.models import Form
+from openforms.forms.models.form_registration_backend import FormRegistrationBackend
+from openforms.forms.models.logic import FormLogic
+from openforms.logging import logevent
 from openforms.logging.models import TimelineLogProxy
+from openforms.plugins.exceptions import InvalidPluginConfiguration
+from openforms.registrations.registry import register
 from openforms.submissions.models.submission import Submission
 from openforms.submissions.utils import get_filtered_submission_admin_url
 from openforms.typing import StrOrPromise
+from openforms.utils.json_logic.introspection import introspect_json_logic
+from openforms.variables.constants import FormVariableDataTypes
+from openforms.variables.service import get_static_variables
 
 
 @dataclass
@@ -88,6 +98,36 @@ class InvalidCertificate:
             kwargs={"object_id": self.id},
         )
         return form_admin_url
+
+
+@dataclass
+class InvalidRegistrationBackend:
+    config_name: StrOrPromise
+    exception_message: str
+    form_name: str
+    form_id: int | None = None
+
+    @property
+    def admin_link(self) -> str:
+        form_admin_url = reverse(
+            "admin:forms_form_change", kwargs={"object_id": self.form_id}
+        )
+        return form_admin_url
+
+    @property
+    def form_level(self) -> bool:
+        return bool(self.form_id)
+
+
+@dataclass
+class InvalidLogicRule:
+    variable: str
+    form_name: str
+    form_id: int
+
+    @property
+    def admin_link(self) -> str:
+        return reverse("admin:forms_form_change", kwargs={"object_id": self.form_id})
 
 
 def collect_failed_emails(since: datetime) -> Iterable[FailedEmail]:
@@ -257,3 +297,119 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
             )
 
     return invalid_certs
+
+
+def collect_invalid_registration_backends() -> list[InvalidRegistrationBackend]:
+    registration_backends = (
+        backend
+        for backend in FormRegistrationBackend.objects.select_related("form").iterator()
+        if backend.form.is_available
+    )
+
+    checked_plugins = set()
+    invalid_registration_backends = []
+    for registration_backend in registration_backends:
+        form_name = registration_backend.form.admin_name
+        form = registration_backend.form
+        backend_options = registration_backend.options
+        backend_type = registration_backend.backend
+        plugin = register[backend_type]
+
+        # errors in general configuration are more straightforward for the user,
+        # so we show the exact ones
+        plugin_cls = type(plugin)
+        if plugin_cls not in checked_plugins:
+            try:
+                plugin.check_config()
+            # we always raise an InvalidPluginConfiguration exception even when we have
+            # errors like HTTPError
+            except InvalidPluginConfiguration as e:
+                invalid_registration_backends.append(
+                    InvalidRegistrationBackend(
+                        config_name=plugin.verbose_name,
+                        exception_message=e.args[0],
+                        form_name=form_name,
+                        form_id=form.id,
+                    )
+                )
+                checked_plugins.add(plugin_cls)
+
+        # errors in the form level can be more detailed so here we want to show
+        # a more general error and the link to the broken form to investigate
+        serializer = plugin.configuration_options(
+            data=backend_options,
+            context={"validate_business_logic": True},
+        )
+
+        try:
+            serializer.is_valid(raise_exception=True)
+        except serializers.ValidationError:
+            invalid_registration_backends.append(
+                InvalidRegistrationBackend(
+                    config_name=plugin.verbose_name,
+                    exception_message=_(
+                        "Invalid registration backend configuration detected"
+                    ),
+                    form_name=form_name,
+                    form_id=form.id,
+                )
+            )
+
+    return invalid_registration_backends
+
+
+def collect_invalid_logic_rules() -> list[InvalidLogicRule]:
+    forms = Form.objects.live().prefetch_related("formvariable_set")
+    static_variables = {
+        var.key: {"source": var.source, "type": var.data_type}
+        for var in get_static_variables()
+    }
+
+    invalid_logic_rules = []
+    for form in forms:
+        form_variables = {
+            form_variable.key: {
+                "source": form_variable.source,
+                "type": form_variable.data_type,
+            }
+            for form_variable in form.formvariable_set.all()
+        }
+
+        all_keys = list(static_variables) + list(form_variables)
+
+        form_logics = FormLogic.objects.filter(form=form)
+        form_logics_vars = [
+            var
+            for logic in form_logics
+            for var in introspect_json_logic(logic.json_logic_trigger).get_input_keys()
+        ]
+
+        for var in form_logics_vars:
+            outer, *nested = var.key.split(".")
+
+            if var.key not in all_keys:
+                if nested and (
+                    (
+                        outer in form_variables
+                        and form_variables[outer]["type"]
+                        in [FormVariableDataTypes.object, FormVariableDataTypes.array]
+                        and form_variables[outer]["source"] == "user_defined"
+                    )
+                    or (
+                        outer in static_variables
+                        and static_variables[outer]["type"]
+                        in [FormVariableDataTypes.object, FormVariableDataTypes.array]
+                    )
+                ):
+                    logevent.invalid_variable_reference_in_logic(form, var.key)
+
+                else:
+                    invalid_logic_rules.append(
+                        InvalidLogicRule(
+                            variable=var.key,
+                            form_name=form.admin_name,
+                            form_id=form.id,
+                        )
+                    )
+
+    return invalid_logic_rules
--- a/src/openforms/emails/digest.py
+++ b/src/openforms/emails/digest.py
@@ -1,3 +1,4 @@
+import logging
 import uuid
 from dataclasses import dataclass
 from datetime import datetime, timedelta
@@ -17,19 +18,22 @@ from simple_certmanager.models import Certificate
 
 from openforms.contrib.brk.service import check_brk_config_for_addressNL
 from openforms.contrib.kadaster.service import check_bag_config_for_address_fields
+from openforms.forms.constants import LogicActionTypes
 from openforms.forms.models import Form
 from openforms.forms.models.form_registration_backend import FormRegistrationBackend
 from openforms.forms.models.logic import FormLogic
-from openforms.logging import logevent
 from openforms.logging.models import TimelineLogProxy
 from openforms.plugins.exceptions import InvalidPluginConfiguration
 from openforms.registrations.registry import register
 from openforms.submissions.models.submission import Submission
 from openforms.submissions.utils import get_filtered_submission_admin_url
 from openforms.typing import StrOrPromise
+from openforms.utils.json_logic.datastructures import InputVar
 from openforms.utils.json_logic.introspection import introspect_json_logic
 from openforms.variables.constants import FormVariableDataTypes
 from openforms.variables.service import get_static_variables
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -105,7 +109,7 @@ class InvalidRegistrationBackend:
     config_name: StrOrPromise
     exception_message: str
     form_name: str
-    form_id: int | None = None
+    form_id: int
 
     @property
     def admin_link(self) -> str:
@@ -113,10 +117,6 @@ class InvalidRegistrationBackend:
             "admin:forms_form_change", kwargs={"object_id": self.form_id}
         )
         return form_admin_url
-
-    @property
-    def form_level(self) -> bool:
-        return bool(self.form_id)
 
 
 @dataclass
@@ -332,7 +332,7 @@ def collect_invalid_registration_backends() -> list[InvalidRegistrationBackend]:
                         form_id=form.id,
                     )
                 )
-                checked_plugins.add(plugin_cls)
+            checked_plugins.add(plugin_cls)
 
         # errors in the form level can be more detailed so here we want to show
         # a more general error and the link to the broken form to investigate
@@ -359,7 +359,7 @@ def collect_invalid_registration_backends() -> list[InvalidRegistrationBackend]:
 
 
 def collect_invalid_logic_rules() -> list[InvalidLogicRule]:
-    forms = Form.objects.live().prefetch_related("formvariable_set")
+    forms = Form.objects.live().iterator()
     static_variables = {
         var.key: {"source": var.source, "type": var.data_type}
         for var in get_static_variables()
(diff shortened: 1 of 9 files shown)


vaszig:
This should wait for #4163 to be approved because it also touches some files from there.

src/openforms/emails/digest.py
```diff
@@ -179,3 +214,73 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+                        form_id=str(form.id),
+                        form_level=True,
+                    )
+                )
+
+    return failed_registration_backends
+
+
+def collect_failed_logic_rules() -> list[FailedLogicRule]:
+    logic_for_active_forms = FormLogic.objects.filter(form__active=True)
```
Viicos, line 265: Same as above, I think the iterating logic can be improved and we need to take into account static/user defined variables. We can discuss this tomorrow

src/openforms/emails/digest.py
```diff
@@ -179,3 +214,73 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+            if broken_general_config := check_general_config(plugin):
+                failed_registration_backends.append(
+                    FailedRegistrationBackend(
+                        config_name=plugin.verbose_name,
+                        exception_message=broken_general_config,
+                        form_name=form_name,
+                    )
+                )
+
+            # errors in the form level can be more detailed so here we want to show
```
Viicos, line 246:
```suggestion
            # errors at the form level can be more detailed so here we want to show
```

Viicos (changes requested):

src/openforms/emails/digest.py
```diff
@@ -179,3 +257,80 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
         )
 
     return broken_configurations
+
+
+def collect_invalid_registration_backends() -> list[InvalidRegistrationBackend]:
```
Viicos, line 262: Maybe another implementation could be to split the detection of invalid registration backend instances (where we do `plugin.configuration_options(data=options)`) and the config checks (where we do `plugin.check_config()`).

This would avoid:
- The duplicate check logic you implemented with `already_detected`
- Having everything in the single `InvalidRegistrationBackend` class (the `form_level` boolean attribute makes things confusing, and `form_name/id` is not relevant for the plugin check).

So I think `collect_invalid_registration_backends` could only check for invalid instances (with the `configuration_options` check), and another collect function can be defined to iterate over the registration plugins (using `iter_enabled_plugins`) return a list of a separate dataclass representing invalid plugins.

src/openforms/emails/digest.py
```diff
@@ -69,6 +77,76 @@ class BrokenConfiguration:
...
+    variable: str
+    form_name: str
+    form_id: str
+
+    @property
+    def admin_link(self) -> str:
+        return reverse("admin:forms_form_change", kwargs={"object_id": self.form_id})
+
+
+def is_form_config_valid(plugin: BasePlugin, options: dict[str, str]) -> bool:
```
Viicos, line 107: I think this one and `check_general_config` are simple enough to be done in the collect functions directly

src/openforms/emails/digest.py
```diff
@@ -69,6 +77,76 @@ class BrokenConfiguration:
...
+        return e.args[0]
+
+    return ""
+
+
+def get_variables_from_logic(form_logic: FormLogic) -> list[str]:
+    data = form_logic.json_logic_trigger
+    var_values = []
+
+    def _recursive_find(data):
```
Viicos, line 134: As we discussed, this will need to take into account some edge cases:

<details>

<summary>cases</summary>

```jsonl
{
    "var": "path.to.some.data"
}
{
    "var": ["path.to.some.data", 9]
}
{
    "var": {
        "var": "a"
    }
}
```
</details>

(adding them here so that others can see)

src/openforms/emails/digest.py
```diff
@@ -69,6 +77,76 @@ class BrokenConfiguration:
...
+
+    # we always raise an InvalidPluginConfiguration exception even when we have
+    # errors like HTTPError
+    except InvalidPluginConfiguration as e:
+        return e.args[0]
+
+    return ""
+
+
+def get_variables_from_logic(form_logic: FormLogic) -> list[str]:
```
Viicos, line 130:
```suggestion
def _get_variables_from_logic(form_logic: FormLogic) -> list[str]:
```
To explicitly state this as being a private util

src/openforms/emails/digest.py
```diff
@@ -179,3 +257,80 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+            form_variable.key for form_variable in form.formvariable_set.all()
+        ]
+
+        all_keys = static_variables + form_variables_keys
+        form_logics = FormLogic.objects.filter(form=form)
+        form_logics_vars = [
+            var for logic in form_logics for var in get_variables_from_logic(logic)
+        ]
+
+        for var in form_logics_vars:
```
Viicos, line 326: As we discussed, it might be that a variable is of type `object`, and a `"var"` json logic operator is being used to access a (nested) property of this object, e.g.:

`someStaticVar` is of type `object`, and we have a json rule with the json logic trigger:

```json
{"var": "someStaticVar.someProperty"}
```

In this case, this will add an `InvalidLogicRule` entry as a false positive. However, this case might be really uncommon.

src/openforms/emails/digest.py
sergei-maertens: also `{"var": ""}` is valid, it returns the whole object

src/openforms/emails/digest.py
sergei-maertens, line 130: this should probably be a method on the `FormLogic` model, I have some uses for it too in building a logic evaluation dependency tree :)

sergei-maertens, line 130: Also, instead of this whole new implementation, it's probably better to use the existing introspection utilities: `openforms.utils.json_logic.introspect_json_logic` which returns a `ExpressionIntrospection` instance that has the method `get_input_components` that can serve as inspiration or even be adapted (it seems to not be used anymore).

src/openforms/emails/digest.py
sergei-maertens: checking a plugin by itself doesn't make sense if the plugin is not actually used in any forms, so looping over `iter_enabled_plugins` will just complicate stuff since you need to look up in which forms it's used and that's a lot harder than starting from the registration backends in the first place.

I think starting from the forms itself is the best approach here to not report on things that don't cause problems and ensure that receivers of the email digest know which form(s) are affected by this problem.

src/openforms/emails/digest.py
```diff
@@ -179,3 +257,80 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+
+def collect_invalid_logic_rules() -> list[InvalidLogicRule]:
+    forms = Form.objects.live()
+    static_variables = [var.key for var in get_static_variables()]
+
+    invalid_logic_rules = []
+    for form in forms:
+        form_variables_keys = [
+            form_variable.key for form_variable in form.formvariable_set.all()
+        ]
```
sergei-maertens, line 318: I think you can get those more efficiently via a clever `.prefetch_related` call on the `forms` queryset.

src/openforms/emails/digest.py
```diff
@@ -179,3 +257,80 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+
+        all_keys = static_variables + form_variables_keys
+        form_logics = FormLogic.objects.filter(form=form)
```
sergei-maertens, line 321: should also be included in `Form.objects.live.prefetch_related(...)`

src/openforms/emails/digest.py
sergei-maertens: 1. this should handle the array (for default value) too indeed
2. let's add some logging for when a variable key has dots in it and it's not found so that we can monitor for false positives. we even expose a `auth` static variable which is of type object that can cause problems here. Unfortunately, we're not far enough yet to provide more detailed schemas of the variables that we could make use of for valid/invalid variable specifications.

I think it'd be wise to not report on missing variables if a "parent" variable (`someStaticVar`) of type object exists.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -77,3 +77,45 @@ <h5>{% trans "Configuration problems" %}</h5>
...
+    <h5>{% trans "Registration backends problems" %}</h5>
+    <ul>
+        <ul>
+        {% for failed_backend in invalid_registration_backends %}
+            <li>
+                <p>
+                    {% blocktranslate with name=failed_backend.config_name message=failed_backend.exception_message trimmed %}
+                        The configuration for plugin '{{ name }}' is invalid.<br>
+                        ({{ message }})<br>
+                    {% endblocktranslate %}
```
sergei-maertens, line 91: I think you'd still want to report in which forms this plugin is used so that the recipient can properly judge the impact of this problem.

src/openforms/emails/tests/test_tasks.py
```diff
@@ -399,3 +435,201 @@ def test_no_email_sent_when_brk_congiguration_is_invalid_but_unused(self):
...
+            send_email_digest()
+
+        sent_email = mail.outbox[0]
+
+        self.assertEqual(
+            sent_email.subject, "[Open Forms] Daily summary of detected problems"
+        )
+        self.assertEqual(sent_email.recipients(), ["user@example.com"])
+        self.assertIn(
+            _("The configuration for plugin 'Invalid backend' is invalid."),
```
sergei-maertens, line 462: as remarked in another PR, it doesn't make a lot of sense to mark these strings translatable.

sergei-maertens:
this

src/openforms/emails/digest.py
```diff
@@ -257,3 +297,119 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+
+        for var in form_logics_vars:
+            outer, *nested = var.key.split(".")
+
+            if var.key not in all_keys:
+                if nested and (
+                    (
+                        outer in form_variables
+                        and form_variables[outer]["type"]
+                        in [FormVariableDataTypes.object, FormVariableDataTypes.array]
```
sergei-maertens, line 395: type `array` should only be allowed if `nested[1]` is numeric, e.g. `foo.0.bar` -> `nested[1] == '0'` and `'0'.isnumeric() == True`

something like:

```py
if var.key not in all_keys:
    expected_container_types = {FormVariableDataTypes.object}
    if nested:
        if nested[1].isnumeric():
            expected_container_types.add(FormVariableDataTypes.array)
        # look up the var
        parent_var = form_variables.get(outer) or statci_variables.get(outer)
        # XXX: shouldn't this be *not* in expected_container_types??
        if parent_var is not None and parent_var["type"] in expected_container_types:
            logevent.invalid_variable_reference_in_logic(form, var.key)
    else:
        # report invalid logic rule

```

src/openforms/emails/digest.py
```diff
@@ -257,3 +297,119 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+        }
+
+        all_keys = list(static_variables) + list(form_variables)
+
+        form_logics = FormLogic.objects.filter(form=form)
+        form_logics_vars = [
+            var
+            for logic in form_logics
+            for var in introspect_json_logic(logic.json_logic_trigger).get_input_keys()
+        ]
```
sergei-maertens, line 385: this now only checks the triggers, and not the actions that use json logic expressions to calculate values

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -107,3 +107,42 @@ <h5>{% trans "Found invalid certificates" %}</h5>
...
+                </p>
+                <p>
+                    {% blocktranslate with form_name=failed_backend.form_name trimmed %}
+                        This affects form '{{ form_name }}'
+                    {% endblocktranslate %}
```
sergei-maertens, line 125: this could affect more forms that use the same registration backend, right? In the case of a generic plugin check (without form-specific options). The easiest way to handle this would be a small language change:

```suggestion
                    {% blocktranslate with form_name=failed_backend.form_name trimmed %}
                        This affects form '{{ form_name }}', possibly other forms are affected too.
                    {% endblocktranslate %}
```

src/openforms/utils/json_logic/datastructures.py
```diff
@@ -40,11 +45,26 @@ class ExpressionIntrospection:
...
+                continue
+            if not isinstance(node.arguments[0], str):
+                continue
+
+            key = cast(str, node.arguments[0])
+            inputs.append(InputVar(key=key))
+
+        return inputs
+
     def get_input_components(
```
sergei-maertens, line 63: I think we concluded that this method wasn't used anywhere and can be removed?

src/openforms/logging/logevent.py
```diff
@@ -693,3 +693,12 @@ def email_status_change(
...
     )
+
+
+# - - -
+def invalid_variable_reference_in_logic(form: Form, logic_variable: str):
+    _create_log(
+        form,
+        "invalid_variable_reference_in_logic",
+        extra_data={"logic_variable": logic_variable},
+    )
```
sergei-maertens, line 704: we don't need DB logging for this, just a simple stdlib `logger.info(...)` is enough

src/openforms/emails/digest.py
```diff
@@ -257,3 +296,119 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+        all_keys = list(static_variables) + list(form_variables)
+
+        form_logics = FormLogic.objects.filter(form=form)
+
+        form_logics_vars = []
+        for logic in form_logics:
+            for var in introspect_json_logic(logic.json_logic_trigger).get_input_keys():
+                form_logics_vars.append(var)
+            for action in logic.actions:
+                if action_var := action.get("variable"):
```
sergei-maertens, line 386:
```suggestion
                if action["action"]["type"] == LogicActionTypes.variable:
```

src/openforms/emails/digest.py
```diff
@@ -257,3 +296,119 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+                    form_logics_vars.append(InputVar(key=action_var))
```
sergei-maertens, line 387: you basically also need to introspect the action value because those are json logic expressions too, see `openforms.submissions.logic.actions.VariableAction`.

So something along the lines of:

```py
expression = action["action"]["value"]
form_logics_vars += [var for var in introspect_json_logic(expression).get_input_keys()]
```

src/openforms/emails/digest.py
```diff
@@ -257,3 +296,119 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+                            form.admin_name,
+                        )
+                else:
+                    invalid_logic_rules.append(
+                        InvalidLogicRule(
+                            variable=var.key,
+                            form_name=form.admin_name,
+                            form_id=form.id,
+                        )
+                    )
```
sergei-maertens, line 412: for var in set(form_logics_vars):
            # there is a variable with this exact key, it is a valid reference
            if var.key in all_keys:
                continue

            outer, *nested = var.key.split(".")

            def _report():
                invalid_logic_rules.append(
                    InvalidLogicRule(
                        variable=var.key,
                        form_name=form.admin_name,
                        form_id=form.id,
                    )
                )

            # Before checking the type of the parent/outer bit, check if it exists in
            # the first place. It could also be that it's not a parent at all (nested is
            # empty), so that's a guaranteed broken reference too.
            if outer not in all_keys:
                _report()
                continue

            # Nested cannot be empty now - if it were, either the key exists as a var,
            # which is the first thing we checked above, or it doesn't exist as a var,
            # which is the second thing we checked above and errored out.
            assert nested
            
            # We can only check the outer bit of the variable, which must be a suitable
            # container type to be valid. Container types are objects and arrays, but
            # we only consider arrays as a valid type if the first bit of the nested
            # key is numeric, as this must refer to an array index.
            # E.g. foo.bar -> only objects are possible, but foo.3 -> objects and arrays
            # are possible (you could have a dict with the string "3" as key").
            expected_container_types = {FormVariableDataTypes.object}
            if nested[0].isnumeric():
                expected_container_types.add(FormVariableDataTypes.array)

            # Now, check that the type of the container variable is as expected. If the
            # variable does *not* have any of the container types, then it's a 
            # guaranteed error since lookups inside primitives are not possible.
            parent_var = form_variables.get(outer) or static_variables.get(outer)
            if parent_var["type"] not in expected_container_types:
                _report()
                continue

            # all the other situations - we cannot (yet) conclude anything meaningful,
            # so instead, log information for our own insight into (typical) usage/
            # constructions.
            logger.info(
                "possible invalid variable reference (%s) in logic of form %s",
                var.key,
                form.admin_name,
            )

src/openforms/emails/tests/test_digest_functions.py
```diff
@@ -487,3 +530,215 @@ def test_both_expiring_and_invalid_certificates_are_collected(self):
...
+            actions=[
+                {
+                    "action": {"type": "variable", "value": "d"},
+                    "variable": "foo.wrong",
+                    "component": "",
+                    "form_step": "",
+                    "form_step_uuid": "c490b80f-05fe-4006-884d-f631c8abe968",
+                },
+            ],
+        )
```
sergei-maertens, line 703: I'm missing some cases:

* variable `foo` of type string & logic rule doing `foo.bar`
* variable `foo` of type array & logic rule doing `foo.bar`

sergei-maertens (changes requested):
Almost there!

src/openforms/emails/digest.py
```diff
@@ -257,3 +297,154 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+                if action["action"]["type"] == LogicActionTypes.variable:
+                    form_logics_vars.append(InputVar(key=action["variable"]))
+                    expression = action["action"]["value"]
+                    form_logics_vars += [
+                        var
+                        for var in introspect_json_logic(expression).get_input_keys()
+                    ]
+
+        for var in set(form_logics_vars):
+            # there is a variable with this exact key, it is a valid reference
```
vaszig, line 396: I have also left all your comments in place as I think that they are really helpful for the person who reads the code. I hope it doesn't mind.

sergei-maertens: yeah I hoped you'd do that :)

src/openforms/emails/digest.py
```diff
@@ -257,3 +297,154 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
...
+            for var in introspect_json_logic(logic.json_logic_trigger).get_input_keys():
+                form_logics_vars.append(var)
+            for action in logic.actions:
+                if action["action"]["type"] == LogicActionTypes.variable:
+                    form_logics_vars.append(InputVar(key=action["variable"]))
+                    expression = action["action"]["value"]
+                    form_logics_vars += [
+                        var
+                        for var in introspect_json_logic(expression).get_input_keys()
+                    ]
```
sergei-maertens, line 393: (no need to fix this, just for the future)

the list comprehension isn't necessary, `get_inputs_keys()` already returns a list of input vars:

```suggestion
                    form_logics_vars += introspect_json_logic(expression).get_input_keys()
```

sergei-maertens (approved):

PR #4210, opened 2024-04-19
[#3725] Add failed services and configuration overview to email digest

### LINKED ISSUE #3725: Detect more problems and add them to the email digest

vaszig:
Fixes #3725 partially

Handles the last subtask


src/openforms/config/views.py
```diff
@@ -1,48 +1,21 @@
...
-def is_plugin(plugin: Any) -> TypeGuard[AbstractBasePlugin]:
-    if hasattr(plugin, "identifier"):
-        return True
-    return False
-
-
-class ConfigurationView(UserIsStaffMixin, PermissionRequiredMixin, TemplateView):
+class ConfigurationView(
+    UserIsStaffMixin, PermissionRequiredMixin, ConfigurationCheck, TemplateView
+):
```
sergei-maertens, line 18: I would not add `ConfigurationCheck` as a mixin, can't you just instantiate an instance in the view and use that?

```py
def get_context_data(self, ...):
    ...

    check = ConfigurationCheck(requested_plugin=self.request.GET.get("plugin"))
    sections = check.get_configuration_results(module)
    ...
```
the `ConfigurationCheck` class is completely decoupled from the request/response/view, so there's really no reason to not do simple composition here.

src/openforms/emails/digest.py
```diff
@@ -179,3 +190,36 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+
+        if time_until_expiry <= timedelta(days=14):
+            invalid_certs.append(
+                InvalidCertificate(label=cert.label, expiry_date=cert.expiry_date)
+            )
+
+        if not (is_valid := cert.is_valid_key_pair()):
+            invalid_certs.append(
+                InvalidCertificate(label=cert.label, is_valid_pair=is_valid)
+            )
```
sergei-maertens, line 210: what about when a certificate is both close to expiry, and it's also not valid pair? You'd be reporting the same certificate twice then :)

Probably better to do something like:

```
Found invalid certificates:

    Certificate A: will expire soon, invalid keypair
    Certificate B: will expire soon
```

src/openforms/emails/digest.py
```diff
@@ -179,3 +190,36 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
         )
 
     return broken_configurations
+
+
+def collect_invalid_certificates() -> list[InvalidCertificate]:
+    today = timezone.now()
+
+    invalid_certs = []
+    for cert in Certificate.objects.all():
```
sergei-maertens, line 199: on our test env we have some certificates that are not actually used. I'd really restrict this to certificates that are configured on a zgw-consumers service (or SoapService or other place) to only report on certificates that are in use.

Even better would be to only consider services that are actually used, but I think that's a lot harder :thinking:

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -77,3 +77,53 @@ <h5>{% trans "Configuration problems" %}</h5>
...
+                            '{{ certificate_label }}' certificate is about to expire on {{ expiry_date }}
+                        {% endblocktranslate %}
+                    {% else %}
+
+                        <!-- No private key provided -->
+                        {% if invalid_certificate.is_valid_pair == None %}
+                            {% blocktranslate with certificate_label=invalid_certificate.label trimmed %}
+                                No private key for certificate '{{ certificate_label }}' was provided.
+                            {% endblocktranslate %}
+                        {% endif %}
```
sergei-maertens, line 99: `is_valid_pair` `None` means that nothing can be said about validity. If the certificate is of type `cert_only`, then this is fine because no private key is expected.

You can only say something meaningful about `is_valid_pair == False` because that means something is actually wrong.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -77,3 +77,53 @@ <h5>{% trans "Configuration problems" %}</h5>
...
+                        
+                        <!-- Invalid key-pair -->
+                        {% if invalid_certificate.is_valid_pair == False %}
+                            {% blocktranslate with certificate_label=invalid_certificate.label trimmed %}
+                                The combination of the public key of the certificate and the private key is not valid.
+                                Please look into certificate '{{ certificate_label }}'.
+                            {% endblocktranslate %}
```
sergei-maertens, line 106: we can include admin links here too

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -77,3 +77,53 @@ <h5>{% trans "Configuration problems" %}</h5>
...
+                    {% endif %}
+                </p>
+            </li>
+        {% endfor %}
+    </ul>
+{% endif %}
+
+{% if invalid_configuration_overview %}
+    <h5>{% trans "Invalid configuration overview" %}</h5>
+    {% trans "The following have failed to validate the configuration" %}
```
sergei-maertens, line 117:
```suggestion
    {% trans "The following plugins have failed to validate the configuration" %}
```

src/openforms/emails/digest.py
```diff
@@ -179,3 +190,36 @@ def collect_broken_configurations() -> list[BrokenConfiguration]:
...
+    invalid_configuration_overview = ConfigurationCheck().get_configuration_results()
+
+    invalid_entries = [
+        entry
+        for item in invalid_configuration_overview
+        for entry in item["entries"]
+        if not entry.status
+    ]
+
+    return invalid_entries
```
sergei-maertens, line 225: won't this report on plugins that aren't used at all too? those will definitely not pass config check, because why would you configure something you're not planning on using.

It'd probably be better to start from the forms and figure out:

* which prefill plugins are used
* which registration plugins are used (you already have this in another PR)
* which appointments module is used (this is configured globally)
* ...

and only test those that are actually used. But, we already have more specific detections for these things or they would be better anyway, so I'm not sure how useful it actually is to run the full configuration overview/check.

sergei-maertens (changes requested):
I would also split your single commit message into three commits:

1. Refactor the config check into it's own class & update the view to make use of it
2. Add the additional checks
3. Add the detection of problems

src/openforms/emails/digest.py
vaszig: Discussed on slack, this will not be added since we have more specific checks in place by now.

src/openforms/emails/tests/test_digest_functions.py
```diff
@@ -392,3 +398,83 @@ def test_invalid_bag_configuration_without_required_property_is_not_collected(
...
+
+@override_settings(LANGUAGE_CODE="en")
+class InvalidCertificatesTests(TestCase):
+
+    def test_expiring_certificates_not_used_by_a_service_are_not_collected(self):
+        with open(TEST_FILES / "test.certificate", "r") as client_certificate_f:
+            CertificateFactory.create(
+                label="Test certificate",
+                public_certificate=File(client_certificate_f, name="test.certificate"),
+            )
```
sergei-maertens, line 411: What would make this certificate turn up as invalid? Can you add a comment if it has a particular expiry date? Right now I'd say "yeah but even if this certificate is used by a service, it probably won't show because it's valid anyway".

sergei-maertens (approved):

src/openforms/emails/tests/test_digest_functions.py
vaszig: Yes, I had a comment in each test that was using the certificate concerning the expiry date (which would make it invalid). Is it ok if I add a comment at the class level, or should I add it to each test separately?

sergei-maertens: I prefer to have all information in each test, that's consistent with the principle that each test should be debuggable/isolated by itself. Ideally to understand what's being tested, I only have to look at the test method itself.

PR #4211, opened 2024-04-19
[#4197] Fix late binding bug when resizing multiple images

### LINKED ISSUE #4197: image resize for uploads only works for 1 image when multiple are uploaded

2.6.4 

upload 3 images: 
image

component resize settings:
image

result:
image

Viicos:
Fixes #4197.

I got curious and saw that `ruff check --select B023 src/` ([doc](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/)) caught this issue and a couple more, so I'll provide a follow up PR with more fixes.

### DIFF
--- a/src/openforms/submissions/attachments.py
+++ b/src/openforms/submissions/attachments.py
@@ -173,7 +173,9 @@ def validate_uploads(submission_step: SubmissionStep, data: dict | None) -> None
         raise ValidationError(validation_errors)
 
 
-def attach_uploads_to_submission_step(submission_step: SubmissionStep) -> list:
+def attach_uploads_to_submission_step(
+    submission_step: SubmissionStep,
+) -> list[tuple[SubmissionFileAttachment, bool]]:
     # circular import
     from .tasks import resize_submission_attachment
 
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/attachments.py
```diff
@@ -257,7 +259,7 @@ def attach_uploads_to_submission_step(submission_step: SubmissionStep) -> list:
             # NOTE there is a possible race-condition if user completes a submission before this resize task is done
             # see https://github.com/open-formulieren/open-forms/issues/507
             transaction.on_commit(
-                lambda: resize_submission_attachment.delay(attachment.id, resize_size)
+                resize_submission_attachment.si(attachment.id, resize_size).delay
```
vaszig, line 262: I thought we were using `si` when we want to have a chain of tasks. is this serving similarly our purpose?

vaszig (approved):

src/openforms/submissions/attachments.py
Viicos: `si` is to make a "partial" signature of the task with some arguments (and is immutable, but not really relevant in our case. `s` is a way to make a signature but mutable).

It's similar to `functools.partial` in some way

Viicos: This was my initial thought, but then thinking about about `s`/`si` I was wondering why they existed if `partial` is doing the same thing? Is there something that `s`/`si` does that `partial` doesn't?

sergei-maertens: `.si` and friends are specific to chains because by default the return value of a previous task in a chain is passed as an argument to the next task, which `.si` prevents from happening.

sergei-maertens (approved):

PR #4213, opened 2024-04-19
Fix late `lambda` binding bugs

Viicos:
Follow up https://github.com/open-formulieren/open-forms/pull/4211:

- The form logevent is an actual bug, but not that big of a deal so not really worth a test
- The zds registration is actually fine (as the callback gets called immediately afterwards) but let's stick to the best practice

### DIFF
--- a/src/openforms/forms/tasks.py
+++ b/src/openforms/forms/tasks.py
@@ -1,6 +1,7 @@
 import copy
 import logging
 from datetime import timedelta
+from functools import partial
 
 from django.db import DatabaseError, transaction
 from django.db.utils import IntegrityError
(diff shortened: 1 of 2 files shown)


src/openforms/forms/tasks.py
```diff
@@ -160,7 +161,7 @@ def activate_forms():
                     extra={"pk": form.pk},
                 )
             else:
-                transaction.on_commit(lambda: logevent.form_activated(form))
+                transaction.on_commit(partial(logevent.form_activated, form))
```
vaszig, line 164: Is there a difference between these two?Isn't the result the same?

Viicos: https://docs.astral.sh/ruff/rules/function-uses-loop-variable/ ;)

sergei-maertens (approved):

PR #4214, opened 2024-04-19
Some lints

Viicos:
Things I noticed over the weeks

### DIFF
--- a/src/openforms/formio/service.py
+++ b/src/openforms/formio/service.py
@@ -45,7 +45,7 @@ __all__ = [
 ]
 
 
-def format_value(component: Component, value: Any, as_html: bool = False):
+def format_value(component: Component, value: Any, *, as_html: bool = False):
     """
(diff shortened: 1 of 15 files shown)


src/openforms/formio/service.py
```diff
@@ -45,7 +45,7 @@
 ]
 
 
-def format_value(component: Component, value: Any, as_html: bool = False):
+def format_value(component: Component, value: Any, *, as_html: bool = False):
```
vaszig, line 48: Is that necessary here?I mean is it not obvious that we have kwargs after `value`?
i guess it's a good practice to do it this way?

vaszig (approved):

src/openforms/formio/service.py
Viicos: Yes: https://adamj.eu/tech/2021/07/10/python-type-hints-how-to-avoid-the-boolean-trap/ :)

PR #4218, opened 2024-04-23
[#4146] Apply soap timeout for Stuf-ZDS client

### LINKED ISSUE #4146: Default timeout is always used for SOAP services

2.6.2

When setting a custom timeout for a soapservice this is not reflected in the real request, timeout will still occur after 10s and this is also showing in the logs:
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='httpstat.us', port=443): Read timed out. (read **timeout=10.0**)
image

This seems to only affect the SOAP services, as other services use the timeout from the configuration:
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='httpstat.us', port=443): Read timed out. (read **timeout=11**)
image

Viicos:
Fixes #4146

### DIFF
--- a/src/stuf/stuf_zds/tests/test_client.py
+++ b/src/stuf/stuf_zds/tests/test_client.py
@@ -155,3 +155,15 @@ class StufZdsRegressionTests(SimpleTestCase):
                 self.fail("Body encoding should succeed")
             except RegistrationFailed:
                 pass
+
+    @tag("gh-4146")
+    @requests_mock.Mocker()
+    def test_custom_timeout_applied(self, m: requests_mock.Mocker):
+        m.get("https://example.com/path", text="ok")
+        stuf_service = StufServiceFactory.build(
+            soap_service__url="https://example.com", soap_service__timeout=1
+        )
+        with StufZDSClient(stuf_service, {}) as client:  # type: ignore
+            client.get("https://example.com/path")
+
+        assert m.last_request.timeout == 1  # type: ignore
--- a/src/stuf/models.py
+++ b/src/stuf/models.py
@@ -117,3 +117,6 @@ class StufService(models.Model):
 
     def get_auth(self) -> tuple[str, str] | None:
         return self.soap_service.get_auth()
+
+    def get_timeout(self) -> int:
+        return self.soap_service.timeout
--- a/src/stuf/service_client_factory.py
+++ b/src/stuf/service_client_factory.py
@@ -23,6 +23,7 @@ class ServiceClientFactory:
             "verify": self.service.get_verify(),
             "cert": self.service.get_cert(),
             "auth": self.service.get_auth(),
+            "timeout": self.service.get_timeout(),
         }
         return kwargs
 


src/stuf/stuf_zds/tests/test_client.py
```diff
@@ -155,3 +156,13 @@ def test_non_latin1_characters(self):
...
+
+    @tag("gh-4146")
+    def test_custom_timeout_applied(self):
+        stuf_service = StufServiceFactory.build(
+            soap_service__url="https://httpstat.us/", soap_service__timeout=1
+        )
+        client = StufZDSClient(stuf_service, {})  # type: ignore
+
+        with self.assertRaises(Timeout):
+            client.get("https://httpstat.us/200?sleep=3000")
```
vaszig, line 168: Can't we test this with a mocked request and not a real one?

vaszig (commented):
I am not sure about this fix (if this is the only solution), maybe we should wait for @sergei-maertens 's opinion as well.

src/stuf/stuf_zds/client.py
```diff
@@ -114,13 +114,15 @@ def StufZDSClient(service: StufService, options: ZaakOptions) -> "Client":
...
-    return Client.configure_from(
+    client = Client.configure_from(
         factory,
         options=options,
         failure_log_callback=partial(logevent.stuf_zds_failure_response, service),
         success_log_callback=partial(logevent.stuf_zds_success_response, service),
         **init_kwargs,
     )
+    client.request = partial(client.request, timeout=service.soap_service.timeout)
+    return client
```
sergei-maertens, line 125: ape-pie supports specifying `timeout` at the "client" level: https://github.com/maykinmedia/ape-pie/blob/979972db0797e0c0b50aa14db801c7e8e9ac170f/tests/test_config_adapter.py#L56

The timeout kwarg should be set in the `get_client_init_kwargs` helper.

src/stuf/stuf_zds/tests/test_client.py
sergei-maertens: yes, see the test in ape-pie: https://github.com/maykinmedia/ape-pie/blob/979972db0797e0c0b50aa14db801c7e8e9ac170f/tests/test_config_adapter.py#L56

You must always be able to run the test suite offline.

sergei-maertens (changes requested):

src/stuf/stuf_zds/tests/test_client.py
Viicos: I got my inspiration from this test:
https://github.com/open-formulieren/open-forms/blob/264e0644f88efe5734889677700cf86df538b7fa/src/soap/tests/test_client.py#L246-L259

Probably it will need to be updated

sergei-maertens: I vaguely remember there being a very specific edge cases that required this approach due to the zeep abstractions

sergei-maertens (approved):

sergei-maertens:
Backport: c79d5f9cbb7c97c8a3f651ca7db3fed2b859b3b9 (to 2.6.x only, older versions don't have this feature)

PR #4223, opened 2024-04-24
[#4205] Add more entries to form-action CSP

### LINKED ISSUE #4205: eHerkenning users with active eHerkenning session blocked by form-action CSP rules.

2.5.x - 2.6.x

When a user has selected "remember my choice" on the eHerkenning page:
image
And starts a form, the user is redirected automatically
https://brk.eid-pp.sitewithprovidercoices.com/brk/DV1CResponder
https://sitewithprovidercoices-eid-pp.sso.eherkenning.nl/brk/EIDIdPDiscoveryProvider
https://brk.eid-pp.sitewithprovidercoices.com/brk/HM1CServiceProvider
to: 
https://admr-a.providerA.nl/ad113_preprod/process
or any other of the providers listed here

But this is not allowed in the CSP from-action rules. They are only set for the "sitewithprovidercoices" domein and not for any of the redirects that are triggered.

This leads to the user being stuck, untill they remove the sessions/cookies and retry. 

(can't reproduce this behavior on signicat environments)

Viicos:
Fixes #4205.

I'm wondering if it makes sense to keep the `ADDITIONAL_CSP_VALUES` hardcoded values?


src/openforms/conf/base.py
```diff
@@ -1027,14 +1027,10 @@
...
-# validates the entire redirect chain, see: https://stackoverflow.com/a/69439102
-CSP_FORM_ACTION = (
-    ["'self'"]
-    + CORS_ALLOWED_ORIGINS
-    + config("CSP_EXTRA_FORM_ACTION", default=[], split=True)
-)
+# Allow any 'https:' host, as we don't know in advance which target is used by eHerkenning.
+# Behavior is also different between browsers regarding redirects:
+# https://github.com/w3c/webappsec-csp/issues/8
+CSP_FORM_ACTION = config("CSP_FORM_ACTION", default=["'self'", "https:"], split=True)
```
Viicos, line 1033: Should we add a sane default for `dev`, allowing `http://localhost`?

src/openforms/conf/base.py
```diff
@@ -1027,13 +1027,11 @@
...
-# CORS_ALLOWED_ORIGINS is included because we (likely) need to redirect back to those
-# third party domains that are embedding the SDK after login. Chrome in particular
-# validates the entire redirect chain, see: https://stackoverflow.com/a/69439102
-CSP_FORM_ACTION = (
-    ["'self'"]
-    + CORS_ALLOWED_ORIGINS
-    + config("CSP_EXTRA_FORM_ACTION", default=[], split=True)
+# Allow any 'https:' host, as we don't know in advance which target is used by eHerkenning.
+# Behavior is also different between browsers regarding redirects:
+# https://github.com/w3c/webappsec-csp/issues/8
```
sergei-maertens, line 1032: Can you please restore the stackoverflow link? I think it illustrates the problem much better with practical examples compared to the github link. It's fine to add the github link for an "authorative" source, but then it should be an additional reference.

src/openforms/conf/base.py
```diff
@@ -1027,13 +1027,11 @@
     "'self'",
 ] + config("CSP_EXTRA_DEFAULT_SRC", default=[], split=True)
 
-# CORS_ALLOWED_ORIGINS is included because we (likely) need to redirect back to those
-# third party domains that are embedding the SDK after login. Chrome in particular
-# validates the entire redirect chain, see: https://stackoverflow.com/a/69439102
-CSP_FORM_ACTION = (
-    ["'self'"]
-    + CORS_ALLOWED_ORIGINS
```
sergei-maertens, line 1035: I don't think that dropping this is a good idea, especially if for example `http://localhost:3000` is listed in `CORS_ALLOWED_ORIGIN` (which I have in my local `.env`).

Dropping the support for `CSP_EXTRA_FORM_ACTION` is also a breaking change, someone may be using this. I think this should also stay for people that want to *add* to the list instead of replacing it alltogether. Since it doesn't appear to be documented, we can treat this as non-public API, but then there should be an entry in the changelog for devops engineers on how to update their infra accordingly.

sergei-maertens (commented):
I'm okay with this changeset, but I also think we can/should revert the mechanism then that digid/eherkenning/ogone add extra CSP items in the database, since realistically, that will all be covered by the `https:` directive...

sergei-maertens:
Discussed with Joeri - we can remove the DigiD/eHerkenning/OgoneMerchant mechanism that updates the `form-action` CSP directive records, it's obsoleted by the setting changes.

src/openforms/conf/base.py
Viicos: Yes I dropped it because I saw it wasn't documented (unlike the others). The only issue I see is if people want to replace the whole list (if for instance they don't want `https:`). Maybe we should do something like:

```python
CSP_FORM_ACTION = config(
    "CSP_FORM_ACTION",
    default=["\"'self'\"", "https:"] + config("CSP_EXTRA_FORM_ACTION", default=[], split=True),
    split=True,
) + CORS_ALLOWED_ORIGINS
```

But is overly complex.

sergei-maertens: and flawed too because it still adds `CORS_ALLOWED_ORIGINS` anyway.

src/openforms/conf/base.py
sergei-maertens: updating the `.env` template is sufficient IMO

src/openforms/conf/base.py
Viicos: So what's the path for this one then? I don't feel like dropping `CSP_EXTRA_FORM_ACTION` is a good idea, it creates an inconsistent state with the other csp settings where such an extra env variable exists.

sergei-maertens: probably that suggestion above is the best?

* for dev, I can add `CSP_EXTRA_FORM_ACTION=http:`
* in prod if I want to add additional domains, infra can use `CSP_EXTRA_FORM_ACTION`
* in prod, if they want to make it more strict, they set `CSP_FORM_ACTION` envvar

sergei-maertens (commented):
The Ogone Merchant CSP update can also be removed :)

src/openforms/conf/base.py
Viicos: Hum, then if we need to support both `CSP_EXTRA_FORM_ACTION` and `CSP_FORM_ACTION` we need to do something similar to https://github.com/open-formulieren/open-forms/pull/4223#discussion_r1589167342.

But supporting extension of settings is hard (what happens if both `CSP_FORM_ACTION` and `CSP_EXTRA_FORM_ACTION` are set?)

src/openforms/payments/contrib/ogone/models.py
```diff
@@ -63,10 +61,3 @@ def clean(self):
 
     def __str__(self):
         return self.label
-
-    def save(self, *args, **kwargs):
-        super().save(*args, **kwargs)
-
-        from openforms.config.models import CSPSetting
-
-        CSPSetting.objects.set_for(self, [(CSPDirective.FORM_ACTION, self.endpoint)])
```
sergei-maertens, line 72: is the `CSPSetting.objects.set_for` method still used anywhere else after this removal? Perhaps we can remove that?

Viicos: It is used in the `update_analytics_tool` function

sergei-maertens:
Backports: 

* 2.6.x: 62e0a9d8300d18284c0dfc4e68abaa6b62ed93ad
* 2.5.x: 6c7abc4966dd8571649930ebd6e492ba984bc991
* 2.4.x: 1b2775be8940e30ee60b3bfb3cbc2f5544182c2a

PR #4227, opened 2024-04-25
[#72] Add backend validation for file component

### LINKED ISSUE #4222: Users can upload too many files

2.6.x

When a upload component has a max files limit set, a user is able to upload more. 

When trying to upload 4 files at once: 
image

but when i first upload 2 files, followed by 4 at once
image

results in +3 uploaded files. Which are successfully processed/submitted.

Viicos:
Partly fixes #72, #4222. I don't know if it should also validate more fields in the submitted values for files?

- SDK: https://github.com/open-formulieren/open-forms-sdk/pull/680


src/openforms/formio/components/vanilla.py
```diff
@@ -311,6 +311,12 @@ def rewrite_for_request(component: FileComponent, request: Request):
                 }
             )
 
+    def build_serializer_field(self, component: FileComponent) -> serializers.Field:
+        return serializers.ListField(
+            max_length=component.get("maxNumberOfFiles"),
+            child=serializers.DictField(),
+        )
```
sergei-maertens, line 318: this definitely has issues with the `component["validate"]["required"]` key and extensive unit tests are needed for all those cases!

We need to see variants with `multiple: true | false` and other possible combinations too.

Happy flow tests are also essential, not just tests that validate that it's invalid.

src/openforms/formio/tests/validation/test_file.py
```diff
@@ -0,0 +1,45 @@
...
+            component,
+            {
+                "foo": [
+                    # truncated for brevity
+                    {
+                        "url": "some://url",
+                        "name": "my-foo-1.bin",
+                        "type": "application/foo",
+                        "originalName": "my-foo-1.bin",
+                    },
```
sergei-maertens, line 31: probably best to define a factory for this that generates a representative data structure since this *will* get copy-pasted and cause problems in the future.

sergei-maertens (changes requested):

sergei-maertens (commented):
Had some time to think about this PR.

This does not go far enough - the nested `DictField` that expects anything does not at all provide proper input validation. We either implement validation for a component entirely or not all, so please extend this to cover all the validations.

Additionally, I'm wondering how much of the file validation that is currently happening in other places (mime type, extensions, virus check...) can be moved here. I'd like to concentrate that in a single place instead of smearing it out over the entire codebase.

Viicos:
> mime type, extensions, virus check...

From what I saw when implementing this, these checks are done during the creation of the temporary file upload, which is done as soon as a file is added in the fileupload component widget. So moving this here might require some refactoring

sergei-maertens:
> > mime type, extensions, virus check...
> 
> From what I saw when implementing this, these checks are done during the creation of the temporary file upload, which is done as soon as a file is added in the fileupload component widget. So moving this here might require some refactoring

Some of them can be moved, some can't - e.g. the virus check and mime type check are fine to keep with the temp upload, but the checks for file types (extensions) for example should be moved.

The "problem" with the temporary uploads is that we have no trusted input pointing to the responsible component for that upload, so we can only perform generic checks there. But once we have the values & the component (so when building the serializer field and running validation), we are able to resolve the (temporary?) uploads belonging to this component and can run the rest of the validation.

sergei-maertens:
As mentioned on Slack, I'd relate this PR to #72 instead :)

src/openforms/formio/components/vanilla.py
```diff
@@ -287,6 +291,70 @@ def build_serializer_field(
...
+
+        if attrs["originalName"] != temporary_upload.file_name:
+            raise serializers.ValidationError(
+                {"originalName": _("Name does not match the uploaded file.")}
+            )
+
+        if attrs["data"]["name"] != temporary_upload.file_name:
+            raise serializers.ValidationError(
+                {"data": {"name": _("Name does not match the uploaded file.")}}
+            )
```
Viicos, line 316: Ideally this error should be raised in the `FileDataSerializer` instead, but would result in duplicate code as I would have to fetch the temporary upload from `temporary_upload_from_url` again (the data structure to validate is complex, I spent my day trying to figure out a good solution so you might have an alternative as I'm unable to think about this anymore).

The reason I'm making this is comment is that the error structure isn't the same when raising a `ValidationError` with a nested detail dict:

`serializer.errors` usually follows the following structure: `{"field": [ErrorDetail(...), ErrorDetail(...)]}`, but when raising with a nested dict, errors are not wrapped in a list: `{"data": {"name": ErrorDetail(...)}}`.

Leading to an inconsistent structure. DRF docs are confusing/wrong, so it doesn't help.

src/openforms/formio/tests/validation/test_file.py
```diff
@@ -0,0 +1,466 @@
...
+            "label": "Test",
+            "storage": "url",
+            "url": "",
+            "useConfigFiletypes": False,
+            "filePattern": "",
+            "file": {"allowedTypesLabels": []},
+            "validate": {"required": True},
+        }
+
+        is_valid, errors = validate_formio_data(component, {})
```
sergei-maertens, line 102: we definitely need to see more cases that represent formio's real behaviour. I suspect it will send a value `[]` or maybe even `[null]` if left blank in the UI

src/openforms/submissions/api/serializers.py
```diff
@@ -259,7 +258,6 @@ def to_representation(self, instance):
         return super().to_representation(instance)
 
     def validate_data(self, data: dict):
-        validate_uploads(self.instance, data=data)
```
sergei-maertens, line 262: We should now do one of two things:

* still call this if the feature flag to disable formio backend validation is enabled
* remove the feature flag to disable formio backend formio validation

src/openforms/formio/tests/test_api_fileupload.py
```diff
@@ -106,21 +106,6 @@ def test_upload_empty(self):
...
-        response = self.client.post(url, {"file": file}, format="multipart")
-
-        self.assertContains(
-            response,
-            _("The provided file is not a {file_type}.").format(
-                **{"file_type": ".png"}
-            ),
-            status_code=status.HTTP_400_BAD_REQUEST,
-        )
-
```
sergei-maertens, line 123: why is this test removed?

How I read this testing code right now is that it seems possible to make a temporary upload with a file whose content doesn't match the declared content type. That should not never be allowed, we need to stop the bogus input as soon as possible.

sergei-maertens (changes requested):

sergei-maertens:
I'm also missing the integration tests that show the frontend validation and backend input blocking are aligned.

src/openforms/formio/tests/validation/helpers.py
```diff
@@ -23,7 +24,7 @@ def validate_formio_data(
     return is_valid, serializer.errors
 
 
-def extract_error(errors: ReturnDict, key: str, index: int = 0):
+def extract_error(errors: ReturnDict, key: str, index: int = 0) -> ErrorDetail:
```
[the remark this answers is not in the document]
Viicos: Hum before changing this I made sure every usage of `extract_error` expected an `ErrorDetail`

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -670,3 +678,60 @@ def test_regex_pattern(self):
...
+
+        async with browser_page() as page:
+            await page.goto(url)
+            await page.get_by_role("button", name="Formulier starten").click()
+
+            async with page.expect_file_chooser() as fc_info:
+                await page.get_by_text("blader").click()
+
+            file_chooser = await fc_info.value
+            # breakpoint()
```
sergei-maertens, line 715:
```suggestion
```

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -670,3 +678,60 @@ def test_regex_pattern(self):
             ui_input="<h2>test</h2>",
             expected_ui_error="Ongeldig Nederlands kenteken",
         )
+
+
+class SingleFileTests(ValidationsTestCase):
```
sergei-maertens, line 683: also needs tests for:

* trying to upload a `.pdf` file that isn't a PDF
* trying to upload a file with an extension that isn't in the allowed file types

sergei-maertens (changes requested):

src/openforms/tests/e2e/input_validation_base.py
```diff
@@ -159,4 +159,6 @@ def _assertBackendValidation(self, form: Form, key: str, value: JSONValue):
                 invalid_params = response.json()["invalidParams"]
                 names = [param["name"] for param in invalid_params]
                 expected_name = f"data.{key}"
-                self.assertIn(expected_name, names)
+                # For complex components (i.e. fileuploads), the error can
+                # be a nested key, so we check for `.startswith`:
+                self.assertTrue(any(name.startswith(expected_name) for name in names))
```
sergei-maertens, line 164: this has collissions if you have keys like `foo` and `foobar` - the `foobar` component would show up while you're testing the `foo` component and that's a false positive.

I'd check that either the exact key is present or, or that `name.startswith(f"{expected_name}.")` to get that path separator in.

sergei-maertens: I'd also like to opt in to this behaviour, the majority of components have exact names, so a classvar `fuzzy_match_invalid_param_names = True` as an override would work nicely.

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -670,3 +681,133 @@ def test_regex_pattern(self):
...
+                    ".pdf",
+                ],
+            },
+            "filePattern": "application/pdf",
+        }
+
+        # The frontend validation will *not* create a TemporaryFileUpload,
+        # as the endpoint will return a 400 because of the forbidden file type.
+        # For this reason, we manually create an invalid TemporaryFileUpload
+        # and use it for the `api_value`:
```
sergei-maertens, line 795: This comment doesn't match what is going on, but either way I would rephrase is - saying that normally the frontend blocks the upload alltogether (because of the allowed file types config), but a malicious user could be using Postman/Bruno/Curl to upload files and try to trick the backend that way.

This tests that the files are not allowed to sneak through.

I'd also add an assertion at the end that only a single temporary file upload exists to ensure no frontend regressions allow uploads after all at some point.

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -670,3 +681,133 @@ def test_regex_pattern(self):
...
+        # The frontend validation will *not* create a TemporaryFileUpload,
+        # as the endpoint will return a 400 because of the bad content type.
+        # For this reason, we manually create an invalid TemporaryFileUpload
+        # and use it for the `api_value`:
+        with open(TEST_FILES / "image-256x256.pdf", "rb") as infile:
+            temporary_upload = TemporaryFileUploadFactory.create(
+                file_name="image-256x256.pdf",
+                content=File(infile),
+                content_type="application/pdf",
+            )
```
sergei-maertens, line 760: This is not realistic - because it's the endpoint itself that blocks the upload (or is supposed to), there is no way that a user can get this mismatched content type uploaded to the backend.

Instead, I would drop this bit and at the end of the test make an assertion that no temporary file uploads exist at all. For the upload itself (`api_value`), you can just generate a random UUID and the backend validation is then expected to report that no such temporary upload exists.

sergei-maertens (changes requested):

src/openforms/tests/e2e/input_validation_base.py
```diff
@@ -159,4 +166,9 @@ def _assertBackendValidation(self, form: Form, key: str, value: JSONValue):
                 invalid_params = response.json()["invalidParams"]
                 names = [param["name"] for param in invalid_params]
                 expected_name = f"data.{key}"
-                self.assertIn(expected_name, names)
+                if self.fuzzy_match_invalid_param_names:
+                    self.assertTrue(
+                        any(name.startswith(f"{expected_name}.") for name in names)
+                    )
```
sergei-maertens, line 172: this doesn't allow for any errors with this exact key, so you need both the `expected_name in names or any(...)` checks

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -670,3 +682,135 @@ def test_regex_pattern(self):
...
+                SubmittedFileFactory.build(
+                    type="application/pdf",
+                    url="http://localhost/api/v2/submissions/files/d4c97feb-68f6-4b85-9b78-7a74ee48b072",
+                    data__url="http://localhost/api/v2/submissions/files/d4c97feb-68f6-4b85-9b78-7a74ee48b072",
+                )
+            ],
+        )
+
+        # Make sure the frontend did not create one:
+        self.assertEqual(TemporaryFileUpload.objects.count(), 0)
```
sergei-maertens, line 771: slightly more elegant/semantic:

```suggestion
        # Make sure the frontend did not create one:
        self.assertFalse(TemporaryFileUpload.objects.exists())
```

src/openforms/tests/e2e/input_validation_base.py
```diff
@@ -133,7 +140,7 @@ def _assertBackendValidation(self, form: Form, key: str, value: JSONValue):
         # step data validation is run *if* a value is provided - it ignores empty data
         # for fields that are required. So we accept an HTTP 400, or if a 200/201 is
         # returned, then we apply additional checks with the _complete endpoint.
-        assert response.status_code in [201, 200, 400]
+        assert response.status_code in (201, 200, 400)
```
Viicos, line 143: So that the match statement correctly exhausts all possible cases

sergei-maertens: TIL

sergei-maertens (commented):
BEFORE we merge this, I want to manually click through it as a final check :)

src/openforms/tests/e2e/input_validation_base.py
Viicos: That's a pyright only thing by the way, mypy will still understand it as a tuple of ints

sergei-maertens (approved):

PR #4233, opened 2024-04-26
[#72] Add backend validation for password component

vaszig:
Closes #72 partially

### DIFF
--- a/src/openforms/formio/tests/validation/test_password.py
+++ b/src/openforms/formio/tests/validation/test_password.py
@@ -1,0 +1,90 @@
+from django.test import SimpleTestCase
+
+from openforms.typing import JSONValue
+
+from ...typing import Component
+from .helpers import extract_error, validate_formio_data
+
+
+class PasswordFieldValidationTests(SimpleTestCase):
+    def test_valid_required_passwordfield(self):
+        component: Component = {
+            "type": "password",
+            "key": "foo",
+            "label": "Foo",
+            "validate": {"required": True},
+        }
+
+        data: JSONValue = {"foo": "test"}
+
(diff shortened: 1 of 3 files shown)


src/openforms/formio/tests/validation/test_password.py
```diff
@@ -0,0 +1,77 @@
...
+            "label": "Foo",
+            "validate": {"required": True},
+        }
+
+        data: JSONValue = {"foo": "test"}
+
+        is_valid, errors = validate_formio_data(component, data)
+
+        self.assertTrue(is_valid)
+        self.assertDictEqual(errors, {})
```
sergei-maertens, line 23:
```suggestion
        self.assertEqual(errors, {})
```
python itself figures out which more specific type assertion it needs to use, always use `self.assertEqual`

src/openforms/formio/tests/validation/test_password.py
```diff
@@ -0,0 +1,77 @@
...
+            "label": "Foo",
+            "multiple": True,
+        }
+
+        data: JSONValue = {"foo": ["ghwgG22", "TRshhe33j"]}
+
+        is_valid, errors = validate_formio_data(component, data)
+
+        self.assertTrue(is_valid)
+        self.assertDictEqual(errors, {})
```
sergei-maertens, line 60:
```suggestion
        is_valid, _ = validate_formio_data(component, data)

        self.assertTrue(is_valid)
```
assertions on errors don't make sense when you establish that it's valid in the first place

sergei-maertens (changes requested):
I am missing the playwright e2e test that guarantees the frontend validations are in sync with the backend validations

sergei-maertens (approved):

PR #4235, opened 2024-04-26
[#4173] Include registration backends when copying a form

### LINKED ISSUE #4173: registration is always removed when copying a form

2.5.6 - 2.6.3
hrlmmr: 101
lv: 164

When copying a form, the registration tab will always be empty. 
It seems to happen for any type of registration.

image

vaszig:
Fixes #4173

### DIFF
--- a/src/openforms/forms/tests/test_copy.py
+++ b/src/openforms/forms/tests/test_copy.py
@@ -19,6 +19,7 @@ from .factories import (
     FormDefinitionFactory,
     FormFactory,
     FormLogicFactory,
+    FormRegistrationBackendFactory,
     FormStepFactory,
     FormVariableFactory,
 )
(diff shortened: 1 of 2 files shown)


src/openforms/forms/tests/test_copy.py
```diff
@@ -263,3 +264,35 @@ def test_copy_form_with_variables(self):
             1, variables_copy.filter(source=FormVariableSources.user_defined).count()
         )
         self.assertEqual(form_copy.formlogic_set.count(), 1)
+
+
+class CopyFormWithRegistrationBackendsTest(APITestCase):
```
sergei-maertens, line 269: why does this have to be a new test case rather than a new test in the existing test case?

vaszig: No specific reason, I just saw that the above for variables was a different TestCase so I thought I should follow the pattern. I will update it and add it to the main one.

sergei-maertens: that may have been because of the variables feature flag, feel free to merge all tests into the `CopyFormAPITests` test case, I don't see a reason why there should be multiple test cases

PR #4271, opened 2024-05-06
🐛 [#4199] Delete authinfo from session after it's stored on submission 

### LINKED ISSUE #4199: Session persistance causes login choice te be ignored

2.6.x

When form A is a form with authentication and i start form B without authentication, form B will ignore this and also be authenticated. 
Complete with prefill and auth_* object being present in the submission. 

This behavior does not seem correct, my choice when starting a form should not be ignored and my authentication object should not be present in form B.

Create 2 forms that share authentication options
Open both front-ends of the forms
Login to first form using open "Inloggen met Demo BAN (test)"
Login to seconde form not using "Inloggen met Demo BAN (test)" but "Start form" 
Check both submissions, the should contain the same Authentication details:
image

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
Closes #4199 partly
Related PR: https://github.com/open-formulieren/open-forms-sdk/pull/688

**Changes**

* Delete authinfo from session after it's stored on submission 
* Add explicit `anonymous` option to submission endpoint

TODO:
- [x] add test for anonymous attribute
- [x] check where `FORM_AUTH_SESSION_KEY` is used
  - [x] set_auth_attribute_on_session
  - [x] set_cosign_data_on_submission (cosign also uses is_authenticated_with_an_allowed_plugin)
  - [x] ResumeFormMixin: is_auth_data_correct / get_redirect_url -> user is always forced to login again (is this an issue?)
      - [x] is_authenticated_with_plugin / meets_plugin_requirements -> run after relogging, so this is fine
      - [x] EHerkenningAuthentication.check_requirements -> used in meets_plugin_requirements which is only used for resume flow
      - [x] VerifyChangeAppointmentLinkView.custom_submission_modifications -> only used in resume flow
- [x] expose function to remove auth info from session in service.py
- [x] call this function at end of add_submission_to_session and other functions that rely on this attr (?)
- [x] update branch in CI of token-exchange lib to see if it passes: https://github.com/open-formulieren/open-forms-ext-token-exchange/pull/16

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/docs/configuration/authentication/eherkenning_eidas.rst
+++ b/docs/configuration/authentication/eherkenning_eidas.rst
@@ -34,7 +34,7 @@ that describes the authenticated entity (person or company).
     EH4         hoog/high       urn:etoegang:core:assurance-class:loa4
     =========== =============== ===========================================
 
-    Source: `Afsprakenstelsel eToegang <https://afsprakenstelsel.etoegang.nl/display/as/Level+of+assurance>`_
+    Source: `Afsprakenstelsel eToegang <https://afsprakenstelsel.etoegang.nl/Startpagina/v1/level-of-assurance>`_
 
 .. note::
 


src/openforms/authentication/signals.py
```diff
@@ -122,6 +122,11 @@ def set_auth_attribute_on_session(
     else:
         store_auth_details(instance, form_auth)
 
+    # After the authentication details have been attached to the submission, the session
+    # no longer has to keep track of this information, because the user could start another
+    # form with different / no authentication
+    del request.session[FORM_AUTH_SESSION_KEY]
```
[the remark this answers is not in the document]
stevenbal: I looked into this, but I'm not entirely sure how I can test it locally (because it seems this is only implemented for BRP?). Judging from the explanation in the [docs](https://open-forms.readthedocs.io/en/latest/developers/extensions.html#keycloak-token-exchange-extension), I don't think it will be an issue because the `request.session[FORM_AUTH_SESSION_KEY]` is deleted after the `submission_start` signal (step 2) is sent and isn't mentioned outside of step 1.

I do however wonder if this won't cause issues for several utils in `openforms.authentication.utils` (e.g. `is_authenticated_with_plugin`), because these utils do rely on `request.session[FORM_AUTH_SESSION_KEY]`.

stevenbal:
@sergei-maertens I'm not 100% sure if this needs backport? What is our policy for when to backport bugfixes?

stevenbal:
@sergei-maertens I checked the usage of the session auth info and I didn't notice any problems if we remove it from the session, the only thing that changes is that when resuming a form, the user will always be forced to login again even if it's within the same session (because the authinfo isn't persisted on the session anymore)

sergei-maertens:
> @sergei-maertens I'm not 100% sure if this needs backport, what is our policy for when to backport bugfixes?

We typically label the ticket as such during refinement, based on how big the impact is and whether a workaround exists or not.

I think that the odds that a end user is filling out multiple forms at the same time where one is authenticated and the other one not are pretty low, so I don't think it warrants a backport. There also doesn't seem to be a stakeholder (except for us) in this, so it doesn't seem to be a big problem.

sergei-maertens:
> @sergei-maertens I checked the usage of the session auth info and I didn't notice any problems if we remove it from the session, the only thing that changes is that when resuming a form, the user will always be forced to login again even if it's within the same session (because the authinfo isn't persisted on the session anymore)

I think that's a good side-effect of this patch, it fits better with the expectations that resuming always forces you to re-authenticate and creates less different authentication flows for the same "mechanism".

src/openforms/authentication/tests/test_signals.py
```diff
@@ -85,6 +86,68 @@ def test_attribute_set_for_demo_plugin_with_staff_user(self):
...
+
+        submission_with_auth.refresh_from_db()
+        self.assertEqual(submission_with_auth.auth_info.value, "123")
+
+        submission_no_auth = SubmissionFactory.create()
+
+        with mock_register(register):
+            set_auth_attribute_on_session(
+                sender=None, instance=submission_no_auth, request=request
+            )
```
sergei-maertens, line 145: is the absence of `anonymous=True` deliberate here?

src/openforms/authentication/tests/test_utils.py
```diff
@@ -71,3 +80,23 @@ def test_get_cosign_info(self):
                 kwargs={"form_slug": "form-with-cosign"},
             ),
         )
+
+    def test_remove_auth_info_from_session(self):
+        factory = APIRequestFactory()
+        request = factory.get("/foo")
+        request.session = self.client.session
+        request = Request(request)
```
sergei-maertens, line 88:
```suggestion
```
`APIRequestFactory` already produces a `Request` instance, no?

I'd expect your type checker to complain about this too btw

src/openforms/submissions/api/viewsets.py
```diff
@@ -166,7 +166,10 @@ def perform_create(self, serializer):
 
         # dispatch signal for modules to tap into
         submission_start.send(
-            sender=self.__class__, instance=serializer.instance, request=self.request
+            sender=self.__class__,
+            instance=serializer.instance,
+            request=self.request,
+            anonymous=serializer.validated_data.get("anonymous", False),
```
sergei-maertens, line 172: maybe provide a default/initial on the serializer field so you can do `serializer.validated_data["anonymous"]`? That's a bit more straight-forward.

sergei-maertens (changes requested):
I'd like to see a test with explicit anonymous start in `openforms.submissions.tests.test_start_submission`.

I don't put too much value on the unit tests of the signal, I'm mostly interested in the integration tests for the `POST` call that those behave as expected, so please add a test that sets up an authentication state and creates an anonymous submission and make assertions on that created submission.

src/openforms/authentication/tests/test_utils.py
stevenbal: I thought so too, that is why I used `APIRequestFactory` instead of `RequestFactory` but strangely enough my type checker complains about this if I don't explicitly cast it:

Screenshot from 2024-05-23 14-23-20

src/openforms/authentication/tests/test_signals.py
stevenbal: I made this test before I added the `anonymous` flag I think, I'll add it when I move these tests to the `start_submission` tests

src/openforms/submissions/api/serializers.py
```diff
@@ -148,6 +148,14 @@ class SubmissionSerializer(serializers.HyperlinkedModelSerializer):
         read_only=True,
     )
 
+    anonymous = serializers.BooleanField(
+        label=_("Anonymous"),
+        help_text=_("Whether the submission was started anonymously or not."),
+        required=False,
+        write_only=True,
+        default=False,  # type: ignore
```
stevenbal, line 156: I think this is an issue with django and typing, but at this point I don't know if I can trust my type checker :stuck_out_tongue: 
Screenshot from 2024-05-23 15-23-06

stevenbal:
Is this a known flaky test? @sergei-maertens https://github.com/open-formulieren/open-forms/actions/runs/9208977538/job/25332413671?pr=4271#step:7:3347

src/openforms/submissions/api/serializers.py
sergei-maertens: yeah it is, maybe that the drf-stubs help with this, but otherwise you can indeed just `type: ignore` it :)

sergei-maertens:
> Is this a known flaky test? @sergei-maertens https://github.com/open-formulieren/open-forms/actions/runs/9208977538/job/25332413671?pr=4271#step:7:3347

Yeah, sporadically it crashes. hypothesis tends to reveal those cases...

sergei-maertens (approved):
I'll leave it up to you to hit the merge button if this is ready to be merged in your opinion :)

src/openforms/authentication/tests/test_utils.py
stevenbal: I'll add `AnyRequest` :+1:

stevenbal: So this means my type checker isn't messed up? :smile:

stevenbal:
> I'll leave it up to you to hit the merge button if this is ready to be merged in your opinion :)

I pushed the typing changed and will merge it once CI passes

PR #4282, opened 2024-05-08
DMN configuration improvements

### LINKED ISSUE #4269: DMN improvements for demo
Some improvements to the DMN UI

- When selecting a decision definition/table to evaluate, the dependencies in the DRD do not seem to be taken into account
- Consequently, only the inputs of a single table, but not as a whole DRD are taken into account
- Check if we can directly provide the necessary inputs instead of a required decision
- Using `a + b` expressions in an input expression is... uncommon. We don't necessarily need to support this right now.
- Invoking a table for each item in an editgrid (repeating group)?
- Try out -> and relay back error message

Useful resource: https://consulting.camunda.com/dmn-simulator/

**Tasks**

- [x] Update DMN parser in django-camunda to correctly resolve requirements and their inputs - this doesn't seem to go right now
- [x] Add JS library (https://github.com/nikku/feelin/blob/main/src/parser.ts) to parse the FEEL expressions client-side to extract the remaining input parameters.
- [x] Ensure that we can configure variable mappings with implicit parameters (e.g. input with `a + b` FEEL expression -> vars `a` and `b`)

sergei-maertens:
Closes #4269

**Changes**

* Updates to new django-camunda version
* Added FEEL parsing on the client-side for better input var mapping
* Improved accessibility

**Checklist**

- Impact on features

  - [x] Checked copying a form
  - [x] Config checks in the configuration overview admin page  (#4284)
  - [x] Problem detection in the admin email digest is handled (#4284)

- Release management

- Commit hygiene


vaszig (approved):
I only have to mention a small thing concerning the styling. Other than that I can't find something weird as far as I can say concerning DMN.

sergei-maertens:
Backported to 2.6.x with 116214ccde906a391e1a5cd317fa6c71e7e14bf8

PR #4305, opened 2024-05-16
 [#4302] Add cosign information for Objects API registration 

### LINKED ISSUE #4302: As municipality I want to pass co-sign information to the Objects API
* [x] Make co-signer BSN available in Object-registration template (legacy) and as variable (new method) that can be mapped
* [x] Make the co-sign date available in Object-registration template (legacy) and as variable (new method) that can be mapped

Viicos:
Closes #4302

**Changes**

Still needs legacy Objects API registration support

Adds the necessary registration variables.

As I had to explore a bit the cosign flow (both legacy and new), noticed that the `co_sign_data` field seems to have a validator incompatible with the new cosign flow:

https://github.com/open-formulieren/open-forms/blob/e780ea782eb2264dc63cf33dd19c67d57a7a15c2/src/openforms/submissions/models/submission.py#L124-L130

`CosignDataSerializer` is defined as:

https://github.com/open-formulieren/open-forms/blob/e780ea782eb2264dc63cf33dd19c67d57a7a15c2/src/openforms/submissions/serializers.py#L12-L22

But the new flow sets this data from the auth plugin data directly:

https://github.com/open-formulieren/open-forms/blob/e780ea782eb2264dc63cf33dd19c67d57a7a15c2/src/openforms/authentication/signals.py#L140-L147

and the data from the auth plugin doesn't have `co_sign_auth_attribute`, `representation`, etc.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

  - [ ] Ran `./bin/makemessages.sh`

- Commit hygiene

### DIFF
--- a/docs/manual/templates.rst
+++ b/docs/manual/templates.rst
@@ -556,6 +556,10 @@ Variabele                              Beschrijving
 ``{{ payment.completed }}``            Indicatie of de betaling voltooid is.
 ``{{ payment.amount }}``               Bedrag dat betaald moet worden.
 ``{{ payment.public_order_ids }}``     Lijst van bestelling IDs die naar de externe betaalprovider meegestuurd zijn.
+``{{ cosign_data.date }}``             De datum waarop de inzending mede is ondertekend.
+``{{ cosign_data.bsn }}``              Het BSN van de medeondertekenaar, indien beschikbaar.
+``{{ cosign_data.kvk }}``              Het KvK van de medeondertekenaar, indien beschikbaar.
+``{{ cosign_data.pseudo }}``           Het pseudo van de medeondertekenaar, indien beschikbaar.
 =====================================  ===========================================================================
 
 
@@ -576,7 +580,11 @@ Voorbeeld
            "attachments": {% uploaded_attachment_urls %},
            "submission_id": "{{ submission.kenmerk }}",
            "language_code": "{{ submission.language_code }}",
-           "public_reference": "{{ submission.public_reference }}"
+           "public_reference": "{{ submission.public_reference }}",
+           {% if cosign_data %}
+           "cosign_date": "{{ cosign_data.date.isoformat }}",
+           "cosign_bsn": "{{ cosign_data.bsn }}"
+           {% endif %}
          }
 
    .. tab:: Resultaat
@@ -594,5 +602,7 @@ Voorbeeld
            "attachments": ["http://some-url.nl/to/attachment1", "http://some-url.nl/to/attachment2"],
            "submission_id": "c305a56f-c56c-49bc-9d94-3e301d0b8bf8",
            "language_code": "nl",
-           "public_reference": "OF-12345"
+           "public_reference": "OF-12345",
+           "cosign_date": "2024-01-01T12:00:00.037672+00:00",
+           "cosign_bsn": "123456783"
          }
--- a/src/openforms/authentication/signals.py
+++ b/src/openforms/authentication/signals.py
@@ -2,6 +2,7 @@ import logging
 
 from django.core.exceptions import PermissionDenied
 from django.dispatch import Signal, receiver
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework.request import Request
@@ -142,4 +143,5 @@ def set_cosign_data_on_submission(
     form_auth = request.session.get(FORM_AUTH_SESSION_KEY)
 
     instance.co_sign_data = form_auth
+    instance.co_sign_data["cosign_date"] = timezone.localtime().isoformat()
     instance.save()
--- a/src/openforms/authentication/signals.py
+++ b/src/openforms/authentication/signals.py
@@ -143,5 +143,5 @@ def set_cosign_data_on_submission(
     form_auth = request.session.get(FORM_AUTH_SESSION_KEY)
 
     instance.co_sign_data = form_auth
-    instance.co_sign_data["cosign_date"] = timezone.localtime().isoformat()
+    instance.co_sign_data["cosign_date"] = timezone.now().isoformat()
     instance.save()
--- a/src/openforms/authentication/tests/test_signals.py
+++ b/src/openforms/authentication/tests/test_signals.py
@@ -4,7 +4,6 @@ from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import PermissionDenied
 from django.test import override_settings, tag
 from django.test.client import RequestFactory
-from django.utils import timezone
 
 from freezegun import freeze_time
 from rest_framework.request import Request
@@ -306,7 +305,7 @@ class SetSubmissionIdentifyingAttributesTests(APITestCase):
         )
 
 
-@freeze_time("2021-11-26T17:00:00+01:00")
+@freeze_time("2021-11-26T17:00:00+00:00")
 class SetCosignDataTests(APITestCase):
     def test_set_cosigner_data(self):
         submission = SubmissionFactory.create(completed=True)
@@ -330,6 +329,6 @@ class SetCosignDataTests(APITestCase):
                 "plugin": "digid",
                 "attribute": "bsn",
                 "value": "123456782",
-                "cosign_date": timezone.localtime().isoformat(),
+                "cosign_date": "2021-11-26T17:00:00+00:00",
             },
         )
--- a/src/openforms/authentication/tests/test_signals.py
+++ b/src/openforms/authentication/tests/test_signals.py
@@ -4,7 +4,9 @@ from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import PermissionDenied
 from django.test import override_settings, tag
 from django.test.client import RequestFactory
+from django.utils import timezone
 
+from freezegun import freeze_time
 from rest_framework.request import Request
 from rest_framework.test import APIRequestFactory, APITestCase
 
@@ -304,6 +306,7 @@ class SetSubmissionIdentifyingAttributesTests(APITestCase):
         )
 
 
+@freeze_time("2021-11-26T17:00:00+01:00")
 class SetCosignDataTests(APITestCase):
     def test_set_cosigner_data(self):
         submission = SubmissionFactory.create(completed=True)
@@ -327,5 +330,6 @@ class SetCosignDataTests(APITestCase):
                 "plugin": "digid",
                 "attribute": "bsn",
                 "value": "123456782",
+                "cosign_date": timezone.localtime().isoformat(),
             },
         )
--- a/src/openforms/registrations/contrib/objects_api/registration_variables.py
+++ b/src/openforms/registrations/contrib/objects_api/registration_variables.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+from datetime import datetime
 from typing import TYPE_CHECKING
 
 from django.utils.translation import gettext_lazy as _
 
+from openforms.authentication.service import AuthAttribute, BaseAuth
 from openforms.plugins.registry import BaseRegistry
 from openforms.variables.base import BaseStaticVariable
 from openforms.variables.constants import FormVariableDataTypes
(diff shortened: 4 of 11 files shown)


src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -77,3 +80,69 @@ def get_initial_value(self, submission: Submission | None = None):
         if submission is None:
             return None
         return submission.payments.get_completed_public_order_ids()
+
+
+@register("cosign")
```
sergei-maertens, line 85:
```suggestion
@register("cosign_data")
```
let's use a proper variable name :)

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -77,3 +80,69 @@ def get_initial_value(self, submission: Submission | None = None):
...
+class Cosign(BaseStaticVariable):
+    name = _("Cosign")
```
sergei-maertens, line 87:
```suggestion
    name = _("Co-sign data")
```

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -1,14 +1,17 @@
 from __future__ import annotations
 
+from datetime import datetime
 from typing import TYPE_CHECKING
 
 from django.utils.translation import gettext_lazy as _
 
+from openforms.authentication.constants import AuthAttribute
```
sergei-maertens, line 8: you're importing from a module other than the `service.py` module, this is not public API. The same goes for the `FormAuth` type - I think you're abusing these types.

co-sign also doesn't have any information about `machtigen` or `loa` afaik, so I'd really define a new typed dict for this instead. I would expect something already exists because of the shape of co sign v1 and v2?

Possibly you can make `BaseAuth` TypedDict public for this, as that captures the co sign v2 correctly.

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -77,3 +80,69 @@ def get_initial_value(self, submission: Submission | None = None):
...
+
+def get_cosign_value(submission: Submission | None, attribute: AuthAttribute) -> str:
+    if not submission or not submission.cosign_complete:
+
+        return ""
+
+    if submission.co_sign_data["attribute"] == attribute:
+        return submission.co_sign_data["value"]
+
+    return ""
```
sergei-maertens, line 107:
```suggestion
def get_cosign_value(submission: Submission | None, attribute: AuthAttribute) -> str:
    if not submission or not submission.cosign_complete:
        return ""

    if submission.co_sign_data["attribute"] == attribute:
        return submission.co_sign_data["value"]

    return ""
```
that blank line looks a bit odd :shrug:

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -77,3 +80,69 @@ def get_initial_value(self, submission: Submission | None = None):
...
+    name = _("Cosign date")
+    data_type = FormVariableDataTypes.datetime
+
+    def get_initial_value(
+        self, submission: Submission | None = None
+    ) -> datetime | None:
+        if not submission or not submission.cosign_complete:
+            return None
+
+        return datetime.fromisoformat(submission.co_sign_data["cosign_date"])
```
sergei-maertens, line 121: this will break on existing submissions that don't have this data yet

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/registration_variables.py
Viicos: Nice catch, I'm wondering how these kind of cases should be handled:

Using a safe `.get` works, but does not reflect "the current behavior" (developers seeing this `.get` might think that this `cosign_date` field might not always be set, while in fact it is from now on).

Should we add a reminder to switch back to a `[...]` access in x months? Or just a comment here is fine?

src/openforms/registrations/contrib/objects_api/registration_variables.py
Viicos: Makes sense for `FormAuth`, however for `AuthAttribute` I think it makes sense to consider it "public" (and maybe in general the `constants` modules from each app).

`AuthAttribute` is imported 16 times in the code base (without taking tests into account).

src/openforms/registrations/contrib/objects_api/registration_variables.py
sergei-maertens:
```py
if (date := cosign_data.get("cosign_date")) is None:
    warnings.warn("Co sign v1 is deprecated", DeprecationWarning)
    return None
```

and add it to the list of things to remove in 3.0 :)

src/openforms/registrations/contrib/objects_api/registration_variables.py
sergei-maertens: yeah I'm trying to get bad practices removed :)

you can export the public constants from the service module

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -77,3 +79,73 @@ def get_initial_value(self, submission: Submission | None = None):
...
+
+    if submission.co_sign_data["attribute"] == attribute:
+        return submission.co_sign_data["value"]
+
+    return ""
+
+
+@register("cosign_date")
+class CosignDate(BaseStaticVariable):
+    name = _("Cosign date")
```
sergei-maertens, line 110: can you fix the renaming names to use a dash? It's "Co-sign", not "Cosign" :)

src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
```diff
@@ -249,6 +254,17 @@ class V2HandlerTests(TestCase):
...
+    BASE_REGISTRATION_OPTIONS: RegistrationOptionsV2 = {
+        "version": 2,
+        "objecttype": "-dummy-",
+        "objecttype_version": 1,
+        "upload_submission_csv": False,
+        "informatieobjecttype_submission_report": "",
+        "informatieobjecttype_attachment": "",
+        "organisatie_rsin": "000000000",
+        "variables_mapping": [],
+    }
```
sergei-maertens, line 266: I don't like this indirectness, and you always end up with new tests that need to modify this constant which then causes other tests to break or become too bloated. Please just specify just the necessary configuration for each test.

If it's really irrelevant, use a factory, but at least isolate tests from each other.

sergei-maertens (changes requested):
Where are the tests and the patches for objects API registration v1?

That's arguably the most important place this needs to work since Utrecht is still depending on this.

sergei-maertens:
The validator on the model field is a problem indeed :(

Can you patch the `save` to call this validator and see how much tests break? We can then decide what to do.

src/openforms/registrations/contrib/objects_api/registration_variables.py
Viicos: Will update other usage in another PR

Viicos:
> Where are the tests and the patches for objects API registration v1?

As I added in the original comment, still needs to implement it for V1

Viicos:
> Can you patch the `save` to call this validator and see how much tests break? We can then decide what to do.

Around 20 tests failing, seems like half of them are caused by the new cosign flow, the other half is invalid tests from the legacy cosign flow (well, not invalid, but with test data incomplete with respect to the validator but not relevant for the test)

src/openforms/authentication/signals.py
```diff
@@ -142,4 +143,5 @@ def set_cosign_data_on_submission(
     form_auth = request.session.get(FORM_AUTH_SESSION_KEY)
 
     instance.co_sign_data = form_auth
+    instance.co_sign_data["cosign_date"] = timezone.localtime().isoformat()
```
sergei-maertens, line 146: why not `timezone.now().isoformat()`? Django's datetimefields store their values in the DB as UTC too and having that consistency is nice.

src/openforms/authentication/tests/test_signals.py
```diff
@@ -327,5 +330,6 @@ def test_set_cosigner_data(self):
                 "plugin": "digid",
                 "attribute": "bsn",
                 "value": "123456782",
+                "cosign_date": timezone.localtime().isoformat(),
```
sergei-maertens, line 333: *please* use explicit assertions, this just repeats the implementation detail

```suggestion
                "cosign_date": "2021-11-26T16:00:00+00:00",
```

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -278,6 +279,24 @@ def get_payment_context_data(submission: Submission) -> dict[str, Any]:
             "public_order_ids": submission.payments.get_completed_public_order_ids(),
         }
 
+    @staticmethod
+    def get_cosign_context_data(submission: Submission) -> dict[str, Any]:
```
sergei-maertens, line 283:
```suggestion
    def get_cosign_context_data(submission: Submission) -> dict[str, str | datetime]:
```

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -278,6 +279,24 @@ def get_payment_context_data(submission: Submission) -> dict[str, Any]:
...
+        if not submission.cosign_complete:
+            return {
+                "bsn": "",
+                "kvk": "",
+                "pseudo": "",
+                "date": "",
+            }
```
sergei-maertens, line 290: maybe `return None` is better to signal that we can't say anything meaningful (because it's legacy co sign flow)?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -278,6 +279,24 @@ def get_payment_context_data(submission: Submission) -> dict[str, Any]:
...
+        else:
+            return {
+                "bsn": get_cosign_value(submission, AuthAttribute.bsn),
+                "kvk": get_cosign_value(submission, AuthAttribute.kvk),
+                "pseudo": get_cosign_value(submission, AuthAttribute.pseudo),
+                # date can be missing on existing submissions, so fallback to an empty string
+                "date": submission.co_sign_data.get("cosign_date", ""),
```
sergei-maertens, line 297: shouldn't this be made a `datetime` to pass to the template engine so that the `date` and `time` filters work correctly? There are no builtins to do this in the template layer, but there are plenty of builtins to do datetime -> str.

sergei-maertens (changes requested):
The extra context variables need to be documented on https://open-forms.readthedocs.io/en/latest/manual/templates.html#objecten-api-registratie too, just like the payment context & include it in the example too please.

docs/manual/templates.rst
```diff
@@ -576,7 +580,11 @@ Voorbeeld
            "attachments": {% uploaded_attachment_urls %},
            "submission_id": "{{ submission.kenmerk }}",
            "language_code": "{{ submission.language_code }}",
-           "public_reference": "{{ submission.public_reference }}"
+           "public_reference": "{{ submission.public_reference }}",
+           {% if cosign_data %}
+           "cosign_date": "{{ cosign_data.date.isoformat }}",
+           "cosign_bsn": "{{  }}"
```
sergei-maertens, line 586: missing some bit?

docs/manual/templates.rst
```diff
@@ -594,5 +602,7 @@ Voorbeeld
            "attachments": ["http://some-url.nl/to/attachment1", "http://some-url.nl/to/attachment2"],
            "submission_id": "c305a56f-c56c-49bc-9d94-3e301d0b8bf8",
            "language_code": "nl",
-           "public_reference": "OF-12345"
+           "public_reference": "OF-12345",
+           "cosign_date": "2024-01-01T12:00:00.037672",
```
sergei-maertens, line 606: Shouldn't this have the timezone indicator?

```suggestion
           "cosign_date": "2024-01-01T12:00:00.037672+00:00",
```

sergei-maertens (approved):

sergei-maertens:
Backported via #4312

PR #4307, opened 2024-05-16
[#1906] Fix duplicate form definitions after importing a form

### LINKED ISSUE #1906: In importing several forms which share (reuse) form definitions, a separate definition is made for each

1.5.0

When I export two forms, both sharing the form definition 'Step1' - and consquently import them in a new OpenForms environment - two separate form definitions called 'Step1' are created. 

For example, these definitions called 'Uw gegevens' would be the same (and shared/reused) in the original environment - but appear separately up when ex- and then imported:
image

_No response_

For OpenForms to recognise identical (reusable) form steps. 
This could be done by hash, in which case there is no chance of overwriting as the form definitions are guaranteed to be identical. Alternatively by ID; In case of recognising duplicate form steps by ID, the options could be offered to 'Keep both versions separately', 'Merge and keep version A' or 'Merge and keep version B'. 

### Screen resolution

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

vaszig:
Closes #1906

**Changes**

- Added regression tests for import and export with a file component as an input
- Removed the dynamic rewrite_formio_components from the export flow as we don't care about the request at this point

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/conf/locale/nl/LC_MESSAGES/django.po
+++ b/src/openforms/conf/locale/nl/LC_MESSAGES/django.po
@@ -4,7 +4,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Open Forms\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-05-21 10:26+0200\n"
+"POT-Creation-Date: 2024-05-21 13:43+0200\n"
 "PO-Revision-Date: 2024-05-17 15:22+0200\n"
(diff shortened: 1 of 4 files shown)


sergei-maertens:
> I am not entirely sure if we want this to be backported. @sergei-maertens what do you think?

While it's annoying, it's not leading to data loss or anything like that, so no need.

sergei-maertens (approved):

PR #4316, opened 2024-05-22
⬆️ Update `typing-extensions` to `4.11.0`

Viicos:
**Changes**

Apart from here:

https://github.com/open-formulieren/open-forms/blob/53a8c845fbf6aced041a2370951a0553c5e3938c/src/openforms/appointments/models.py#L87-L88

It is safe to use `TypeIs` for the other cases.

(Separate PR as I'll need it for another one)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/config/checks.py
+++ b/src/openforms/config/checks.py
@@ -21,9 +21,11 @@ from .data import Action, Entry
 class ConfigCheckable(Protocol):
     verbose_name: str
 
-    def check_config(self) -> None: ...
+    @staticmethod
+    def check_config() -> None: ...
 
-    def get_config_actions(self) -> list[Action]: ...
+    @staticmethod
+    def get_config_actions() -> list[Action]: ...
 
 
 def _subset_match(requested: str | None, checking: str) -> bool:
@@ -39,7 +41,7 @@ def is_plugin(plugin: Any) -> TypeIs[AbstractBasePlugin]:
 
 
 class ConfigurationCheck:
-    def __init__(self, requested_plugin: AbstractBasePlugin = None) -> None:
+    def __init__(self, requested_plugin: str = "") -> None:
         self.requested_plugin = requested_plugin
 
     def get_configuration_results(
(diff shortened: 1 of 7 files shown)


src/openforms/config/checks.py
```diff
@@ -52,9 +54,9 @@ def get_configuration_results(
                     "name": _("Address lookup plugins"),
                     "entries": [
                         self.get_plugin_entry(
-                            BAGCheck,
+                            BAGCheck(),
```
sergei-maertens, line 57: I think it's better to update the protocol to define staticmethods, as that is how the actual checks are implemented. They're not really meant to be instantiated.

src/openforms/config/checks.py
```diff
@@ -119,7 +121,7 @@ def get_plugin_entry(self, plugin: AbstractBasePlugin | ConfigCheckable) -> Entr
         try:
             actions = plugin.get_config_actions()
         except Exception as e:
-            actions = [
+            actions: list[tuple[str, str]] = [
```
sergei-maertens, line 124:
```suggestion
            actions: list[Action] = [
```
there's a type alias for this already

src/openforms/config/checks.py
```diff
@@ -30,14 +32,14 @@ def _subset_match(requested: str | None, checking: str) -> bool:
...
-def is_plugin(plugin: Any) -> TypeGuard[AbstractBasePlugin]:
+def is_plugin(plugin: Any) -> TypeIs[AbstractBasePlugin]:
     if hasattr(plugin, "identifier"):
         return True
     return False
 
 
 class ConfigurationCheck:
-    def __init__(self, requested_plugin: AbstractBasePlugin = None) -> None:
+    def __init__(self, requested_plugin: str | None = None) -> None:
```
sergei-maertens, line 42: can we normalize this to an empty string maybe?

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #4317, opened 2024-05-22
[#4298] Broken admin links and certificate label in email digest

### LINKED ISSUE #4298: Certificate problems are not rendered properly in email digest
The label of the certificate and the admin link are not rendered properly.

vaszig:
Closes #4298

- Handle empty labels in certificates
- Replace admin links with the absolute ones

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/emails/digest.py
+++ b/src/openforms/emails/digest.py
@@ -30,6 +30,7 @@ from openforms.submissions.utils import get_filtered_submission_admin_url
 from openforms.typing import StrOrPromise
 from openforms.utils.json_logic.datastructures import InputVar
 from openforms.utils.json_logic.introspection import introspect_json_logic
+from openforms.utils.urls import build_absolute_uri
 from openforms.variables.constants import FormVariableDataTypes
 from openforms.variables.service import get_static_variables
 
@@ -71,10 +72,12 @@ class FailedPrefill:
             "object_id__in": ",".join(self.submission_ids),
             "extra_data__log_event__in": "prefill_retrieve_empty,prefill_retrieve_failure",
         }
-        submissions_admin_url = furl(
+        submissions_relative_admin_url = furl(
             reverse("admin:logging_timelinelogproxy_changelist")
         )
-        return submissions_admin_url.add(query_params).url
+        submissions_relative_admin_url.add(query_params)
+
+        return build_absolute_uri(submissions_relative_admin_url.url)
 
     @property
     def failed_submissions_counter(self) -> int:
@@ -97,11 +100,12 @@ class InvalidCertificate:
 
     @property
     def admin_link(self) -> str:
-        form_admin_url = reverse(
+        form_relative_admin_url = reverse(
             "admin:simple_certmanager_certificate_change",
             kwargs={"object_id": self.id},
         )
-        return form_admin_url
+
+        return build_absolute_uri(form_relative_admin_url)
 
 
 @dataclass
@@ -113,10 +117,10 @@ class InvalidRegistrationBackend:
 
     @property
     def admin_link(self) -> str:
-        form_admin_url = reverse(
+        form_relative_admin_url = reverse(
             "admin:forms_form_change", kwargs={"object_id": self.form_id}
         )
-        return form_admin_url
+        return build_absolute_uri(form_relative_admin_url)
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


src/openforms/emails/digest.py
```diff
@@ -289,7 +297,7 @@ def collect_invalid_certificates() -> list[InvalidCertificate]:
             invalid_certs.append(
                 InvalidCertificate(
                     id=cert.id,
-                    label=cert.label,
+                    label=cert.__str__(),
```
vaszig, line 300: Firstly I changed this manually in the template (using a default filter for the empty label) but then I thought it's better since we have the `__str__` method doing the same thing. Let me know if you agree.

sergei-maertens, line 300:
```suggestion
                    label=str(cert),
```
you should never call magic methods directly :)

I thought we provided the model instance directly to the template, and then `{{ certificate }}` would have the same effect, but this works too and is the least amount of code changes.

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -207,35 +208,41 @@ def test_email_sent_when_there_are_failures(self, mock_global_config, brk_config
...
             )
-            admin_submissions_url.args = {
+            admin_relative_submissions_url.args = {
                 "form__id__exact": form.id,
                 "needs_on_completion_retry__exact": 1,
                 "registration_time": "24hAgo",
             }
+            admin_submissions_url = build_absolute_uri(
+                admin_relative_submissions_url.url
+            )
```
sergei-maertens, line 221: this just repeats the implementation details, can't you instead use furl to set up the expected state:

```py
admin_path = reverse("admin:submissions_submission_changelist")
admin_submissions_url = furl(f"http://testserver{admin_path}")
```

sergei-maertens (changes requested):

src/openforms/emails/tests/test_tasks_integration.py
vaszig: Shouldn't I use `settings.BASE_URL` instead of the hardcoded url?This is what is taken into account in the method `admin_link`.

src/openforms/emails/digest.py
```diff
@@ -113,10 +118,10 @@ class InvalidRegistrationBackend:
 
     @property
     def admin_link(self) -> str:
-        form_admin_url = reverse(
-            "admin:forms_form_change", kwargs={"object_id": self.form_id}
+        form_relative_admin_url = furl(
+            reverse("admin:forms_form_change", kwargs={"object_id": self.form_id})
         )
-        return form_admin_url
+        return build_absolute_uri(form_relative_admin_url.url)
```
Viicos, line 124:
```suggestion
        return build_absolute_uri(reverse("admin:forms_form_change", kwargs={"object_id": self.form_id}))
```
No need for `furl` here

src/openforms/emails/digest.py
```diff
@@ -97,11 +100,13 @@ class InvalidCertificate:
...
-            "admin:simple_certmanager_certificate_change",
-            kwargs={"object_id": self.id},
+        form_relative_admin_url = furl(
+            reverse(
+                "admin:simple_certmanager_certificate_change",
+                kwargs={"object_id": self.id},
+            )
         )
-        return form_admin_url
+        return build_absolute_uri(form_relative_admin_url.url)
```
Viicos, line 109:
```suggestion
        return build_absolute_uri(reverse("admin:simple_certmanager_certificate_change", kwargs={"object_id": self.id}))
```

No need for `furl` here

src/openforms/emails/digest.py
```diff
@@ -127,7 +132,10 @@ class InvalidLogicRule:
 
     @property
     def admin_link(self) -> str:
-        return reverse("admin:forms_form_change", kwargs={"object_id": self.form_id})
+        form_relative_admin_url = furl(
+            reverse("admin:forms_form_change", kwargs={"object_id": self.form_id})
+        )
+        return build_absolute_uri(form_relative_admin_url.url)
```
Viicos, line 138: Same

src/openforms/emails/tests/test_tasks_integration.py
sergei-maertens: well we control the value of the setting - you can make this more explicit with `@override_settings`, but in general indirection in tests is a bad practice.

Viicos: `testserver` is the default hostname for the django test client. It is not easily configurable, so it's fine to hardcode it in assertions.

Viicos (changes requested):

src/openforms/emails/tests/test_tasks_integration.py
sergei-maertens: there is no test client, this runs in celery tasks without request context :)

sergei-maertens (approved):

Viicos (approved):

PR #4322, opened 2024-05-22
[#4291] Fix logic triggers with boolean user defined variables

### LINKED ISSUE #4291: Logic triggers with boolean variables don't work correctly

latest (fdb19fdd6625e0281bdaed4a39a90f4f9150ec26)

When I use a simple logic rule and use a boolean variable (checkbox component, boolean user defined variable...) as a trigger, this rule is not correctly serialized to JSON.

The resulting expression is:

```json
{
  "==": [
    {
      "var": "kanAfvalOphalen"
    },
    "true"
  ]
}
```

and it should be:

```json
{
  "==": [
    {
      "var": "kanAfvalOphalen"
    },
    true
  ]
}
```

UI representation:

image

The workaround is to use advanced logic instead of simple logic.

1. Add user defined variable `foo` with data type boolean
2. Add a simple logic rule
3. Select the variable `foo` in the "Als" dropdown
4. Select de operation "is gelijk aan"
5. Select "de waarde"
6. Select the comparison value "Ja"
7. Now, click the cog icon and click "Toon/verberg DSL weergave"

Any actions associated with this rule don't execute (for both true/false cases) because a string will never be equal to a boolean.

Selecting Yes/No in the dropdown configures an actual boolean in the JSON logic expression, and my actions execute properly.

### Screen resolution

None

### Device

None

### OS

None

### Browser

Chrome

vaszig:
Closes #4291

**Changes**

- Convert user defined boolean variables values from string to boolean type

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/Trigger.js
+++ b/src/openforms/js/components/admin/form_design/logic/Trigger.js
@@ -293,6 +293,14 @@ const Trigger = ({name, logic, onChange, error, withDSLPreview = false, children
       default:
         firstOperand = {var: triggerVariableKey};
     }
+  } else if (
+    triggerVariable?.source === VARIABLE_SOURCES.userDefined &&
+    triggerVariable.dataType === 'boolean'
+  ) {
+    firstOperand = {var: triggerVariableKey};
+    // cast from string to actual boolean
+    if (compareValue === 'true') compareValue = true;
+    if (compareValue === 'false') compareValue = false;
   } else {
     firstOperand = {var: triggerVariableKey};
   }


src/openforms/js/components/admin/form_design/logic/Trigger.js
```diff
@@ -293,6 +294,14 @@ const Trigger = ({name, logic, onChange, error, withDSLPreview = false, children
...
         firstOperand = {var: triggerVariableKey};
     }
+  } else if (
+    triggerVariable?.source === VARIABLE_SOURCES.userDefined &&
+    triggerVariable.dataType === 'boolean'
+  ) {
+    firstOperand = {var: triggerVariableKey};
+    // cast from string to actual boolean
+    if (compareValue === 'true') compareValue = true;
+    if (compareValue === 'false') compareValue = false;
```
Viicos, line 304:
```suggestion
    compareValue = STRING_TO_TYPE.bool(compareValue);
```

Or if you prefer to keep it this way, the import is unused

PR #4327, opened 2024-05-24
:zap: [#4255] Speed up PDF generation by replacing flexbox with float

### LINKED ISSUE #4255: Text area with 3500+(?) characters crashes pdf generation

2.5.x - 2.6.x

sentry: 353356
RID 192
Venray 103

When a submission contains a textarea with large amounts of text the submission crashes because after 5 minutes the pdf is not generated. 
user will see a error:
image
image

stevenbal:
Turns out that flexbox was the problem here (also mentioned at the end of this comment https://github.com/Kozea/WeasyPrint/issues/578#issuecomment-1058889147). I didn't want to use floats, but all the alternatives I tried were just as slow :disappointed:.

For comparison, this is what a PDF currently looks like if I enter 3500 characters in the textarea (takes about a minute to render locally, if I double the characters it exceeds the maximum): 
[original.pdf](https://github.com/open-formulieren/open-forms/files/15433138/original.pdf). I'm not sure if it's intentional that the first page does not contain any of the steps yet, but I could change that as well if needed

This is what it looks like after these CSS changes (takes about 2-3 seconds to render): [fixes.pdf](https://github.com/open-formulieren/open-forms/files/15433164/fixes.pdf).

Closes #4255

**Changes**
* Speed up PDF generation by replacing flexbox with float

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/scss/pdfs/_submission-step-row.scss
+++ b/src/openforms/scss/pdfs/_submission-step-row.scss
@@ -1,10 +1,6 @@
 @import '~microscope-sass/lib/typography';
 
 .submission-step-row {
-  display: flex;
-  justify-content: flex-start;
-  align-items: flex-start;
-
   & + & {
     margin-top: 1mm;
   }
@@ -42,12 +38,17 @@
     @include body();
     width: 40%;
     padding-right: 2em;
+    // float is used here, because flexbox/grid/tables are really slow with
+    // weasyprint, especially if large textareas are used
+    // see: https://github.com/open-formulieren/open-forms/issues/4255
+    float: left;
   }
 
   &__value {
     @include body();
     width: 60%;
     word-break: break-all;
+    margin-left: 40%;
 
     // wysiwyg content
     p {


sergei-maertens (commented):
Interesting cause and solution!

I'm not happy with the layout yet, especially the wrapping after a page break and the lack of information on the first page. Perhaps that `page-break-inside` is the culprit here?

I'm wondering if we should maybe apply a more column-oriented layout instead of the tabular layout now :thinking:

stevenbal:
@sergei-maertens if I remove the `page-break-inside`, it looks like this [no_avoid.pdf](https://github.com/open-formulieren/open-forms/files/15433588/no_avoid.pdf)

sergei-maertens:
Reference of what a "normal" PDF looks like.
[dmn-demo-15-mei.pdf](https://github.com/open-formulieren/open-forms/files/15433648/dmn-demo-15-mei.pdf)

stevenbal:
The word-break issue only occurs in the dev view, if I render the PDF it works as expected
Screenshot from 2024-05-24 13-51-56

sergei-maertens (approved):

sergei-maertens:
Backports: d8e5a8ac0658f73b09ee79caf00e2d0d48e06620, 555e45f17a0baac099d7df3fdd68ec45c2f46146 and 2b0a933fe8ca1a21b6e7f1ecd64332b971d8c0a5

PR #4331, opened 2024-05-27
:bug: [#4012] Fix content component link popup in form builder

### LINKED ISSUE #4012: CKEditor: cannot open "link" when i have previously opened another link in another content component (missing pop-up)

2.5.2

CKEditor does not show a pup-up for links when the user has previously used the functionality in another component. 

Create 2 content components
Create link in each component
Save the form
Open content1 and click the link 
image

now close content1 and open content2 and click that link:
image

No edit pop-up is visible, but the button does incicate the edit mode is "acitive"

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
Closes #4012 

**Changes**

* remove the ckeditor node after closing the formbuilder editcomponent modal

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/formio_builder/WebformBuilder.js
+++ b/src/openforms/js/components/formio_builder/WebformBuilder.js
@@ -184,6 +184,7 @@ class WebformBuilder extends WebformBuilderFormio {
         // Clean up.
         this.removeEventListener(this.dialog, 'close', dialogClose);
         this.dialog = null;
+        root.unmount();
       };
       this.addEventListener(this.dialog, 'close', dialogClose);
 


stevenbal:
@sergei-maertens does this need backport? It doesn't seem like a huge issue, though I guess it would be nice to have it patched in previous versions

sergei-maertens:
> @sergei-maertens does this need backport? It doesn't seem like a huge issue, though I guess it would be nice to have it patched in previous versions

Nah, it's annoying but doesn't seem to be too common?

What I'm more interested in, is if we can't apply this patch in the form-builder repository instead of here - as I understand it, this node should be destroyed whenever the WYSIWYG editor gets unmounted? So it sounds more like it needs to be patched in a useEffect cleanup in https://github.com/open-formulieren/formio-builder/blob/main/src/registry/content/rich-text.tsx

stevenbal:
@sergei-maertens I have tried to fix this in the formbuilder with a useEffect hook, but haven't been able to get it to work. It seems that `.destroy` is called by default when switching translation tabs for instance, but not when closing the modal. The code in the return of my useEffect hook also doesn't seem to be called when the modal is closed

src/openforms/js/components/formio_builder/WebformBuilder.js
```diff
@@ -184,6 +184,13 @@ class WebformBuilder extends WebformBuilderFormio {
         // Clean up.
         this.removeEventListener(this.dialog, 'close', dialogClose);
         this.dialog = null;
+
+        // We have to reset the ckeditor BodyCollection to allow link popups to work
+        // after closing the editComponent modal, because it seems that `ClassicEditor.destroy` isn't being called
+        // https://ckeditor.com/docs/ckeditor5/latest/api/module_ui_editorui_bodycollection-BodyCollection.html
+        // See https://github.com/open-formulieren/open-forms/issues/4012
+        if (document.querySelector('.ck-body-wrapper'))
+          document.querySelector('.ck-body-wrapper').remove();
```
sergei-maertens, line 193: I was curious - perhaps the problem is that `root.unmount()` is not called, and therefore, the components cleanup functions from useEffect are not being called? This would effectively mean we are currently memory-leaking too :thinking: 

As I understand it, the the modal close by clicking outside of it or on the X basically removes that whole DOM node and since Formio is not aware of our react stuff, it can't properly clean that up.

I'd like to validate that theory before accepting this workaround.

stevenbal: You were right! Calling `root.unmount` fixes in `dialogClose` fixes it

sergei-maertens: :tada: thanks for doing the hard work and pointing us in the right direction!

sergei-maertens (approved):

Issue #4267, opened 2024-05-03
Allow specifying the documents API in the objects API registration


joeribekker:
**Reference:** [PF-114](https://dimpact.atlassian.net/browse/PF-114)

### Thema / Theme

Frontend

### Omschrijving / Description

**See below for an updated list of tasks**

- [x] Restructuring of configuration of Objects API into groups (like ZGW groups) including the Documents API
- [x] Remove global objecttype/documenttype configuration.
- [x] Add group selection in Objects API registration-tab on form-level
- [x] Restructuring of Objecttype selection from URL to dropdown with objecttype UUIDs, after selection of chosen group
- [x] Adjusting variable mapping, taking into account group choice (clear after changing choice - can be a popup).
- [x] #4577
- [ ] Make the ZGW APIs registration consistent #4606

---

We should store the unique-attributes if possible, instead of the UUID, and not store URLs at all (except for the services ofcourse).

- Objecttype: uuid 
- InformatieObjecttype: omschrijving + catalogus
- Zaaktype: omschrijving + catalogus
- RolType: omschrijving + zaaktype

To do so, it is proposed to add a management command to migrate from URLs to `omschrijving` + `catalogus`/`zaaktype` (as it requires API calls that can't be made in data migrations). This command should be run when upgrading from 2.7 to 2.8. There are two ways we could deal with this:
- Explicitly state the management command need to be run in the changelog, and we hope they will.
- Add an `2.7` -> `2.8` upgrade check iterating over the registration backend configurations and checking the management command was run (e.g. no more `zaaktype` field, but `zaaktype_omschrijving` + `zaaktype_catalogus`).
  - Should we add a mechanism to prevent the management command from running two times?
---

Tasks:
- [x] Backend should send a "display name" for the catalogus to the frontend
- [x] https://github.com/open-formulieren/open-forms/issues/4534
  - https://github.com/open-formulieren/open-forms/pull/4539 
- [x] Update the Objects API configuration serializer (V1 and V2): add the `catalogus_domein` and `catalogus_rsin` fields, update business validation logic.
  - https://github.com/open-formulieren/open-forms/pull/4541
- [ ] Add a management command to upgrade existing Objects API configurations
  - https://github.com/open-formulieren/open-forms/pull/4536
- [x] Update Objects API registration plugin logic, to fetch the corresponding IOT URL based on the domein, RSIN and omschrijving. TypedDict definitions should also be updated.
[...]

PR #4332, opened 2024-05-27
Implement Objects API sets

Viicos:
Closes https://github.com/open-formulieren/open-forms/issues/4267

**Changes**

This is best reviewed commit per commit.

- Add a Objects API group model
- The ObjectsAPIConfig model now holds a reference to a default group
- Data migrations to update existing instances
- Update the Objects API registration configuration to be able to specify a group to use (fallback to the default group)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/registrations/api/filters.py
```diff
@@ -23,33 +26,54 @@ class ListInformatieObjectTypenQueryParamsSerializer(serializers.Serializer):
...
+            "in this group will be returned. Otherwise, the Catalogi API in the default Objects API group will be used to find "
+            "informatieobjecttypen."
+        ),
+        label=_("Objects API group"),
+        required=False,
+    )
     registration_backend = serializers.ChoiceField(
         help_text=_("The ID of the registration backend to use."),
         label=_("Registration backend ID"),
-        required=False,
```
Viicos, line 29: I think it doesn't really make sense to call this endpoint without a registration backend ID, so I removed it (the frontend always passes the query param)

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -498,7 +499,12 @@ def register_submission_to_objects_api(
             submission, object_mapping, REGISTRATION_ATTRIBUTE, object_data
         )
 
-        with get_objects_client() as objects_client:
+        objects_api_config = ObjectsAPIConfig.get_solo()
+        assert isinstance(objects_api_config, ObjectsAPIConfig)
```
Viicos, line 503: I assumed we use the default objects API group when registering to the Objects API from the ZGW registration.

Should we add support for group selection as well for ZGW?

sergei-maertens: Yeah, selection seems best and consistent, but this can be a separate PR

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -49,7 +52,14 @@ class ObjecttypeVariableMappingSerializer(serializers.Serializer):
...
 
 class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
-
+    objects_api_group = PrimaryKeyRelatedAsChoicesField(
+        queryset=ObjectsAPIGroupConfig.objects.all(),
+        label=("Objects API group"),
+        help_text=_(
+            "Which Objects API group to use. If not provided, the default Objects API group will be used."
+        ),
+        required=False,
```
sergei-maertens, line 61: I'd make this `required=True` for simplicity.

src/openforms/registrations/contrib/objects_api/migrations/0017_move_singleton_data.py
```diff
@@ -0,0 +1,89 @@
...
+        catalogi_service=solo_model.catalogi_service,
+        objecttype=solo_model.objecttype,
+        objecttype_version=solo_model.objecttype_version,
+        productaanvraag_type=solo_model.productaanvraag_type,
+        informatieobjecttype_submission_report=solo_model.informatieobjecttype_submission_report,
+        informatieobjecttype_submission_csv=solo_model.informatieobjecttype_submission_csv,
+        informatieobjecttype_attachment=solo_model.informatieobjecttype_attachment,
+        organisatie_rsin=solo_model.organisatie_rsin,
+        content_json=solo_model.content_json,
+        payment_status_update_json=solo_model.payment_status_update_json,
```
sergei-maertens, line 35: so looking at the zgw api group, I notice here the models specify a default `objecttype` and `objecttype_version`, plus the `document_type` for the attachments, submission report and CSV. The equivalents for that don't exist on the ZGW API group - you must specify zaaktype/documenttype at the individual form level.

I think this applies for objecttype/objecttype_version for sure - there's no point IMO to provide those at the API group level. They are also easily migrated in this data migration from the solo-config to all the registration backend instances that don't have them specified explicitly.

For some of the other fields I think the answer is a bit harder:

* `productaanvraag_type` -> this is legacy objects API only, so I would tackle this with a data migration on the relevant registration backend instances. It definitely does not belong in the API group model
* `informatieobjecttype_submission_report`, `informatieobjecttype_submission_csv` and `informatieobjecttype_attachment` - these are tied to the catalogi API of the group, so I think it makes sense to configure them here, because the UX impact of every new registration backend with the same API group would become too bad, having to select the same document type over and over again
* `organisatie_rsin` also exists on zgw API group, can stay
* `content_json` and `payment_status_update_json` are only relevant for legacy objects API - I'd keep those on the solo model and remove them entirely in Open Forms 3.0 when we drop support for the legacy flavour. Then we can likely drop the solo-model entirely.

sergei-maertens: The goal of such an API group should really be treat a set of connection details as one unit, and I suppose I can live with some configuration defaults that don't get overridden frequently. But the objecttype should really be a different object type for every form, so there's no point in providing a default for that and all the related implementation complexities.

src/openforms/registrations/contrib/objects_api/migrations/0017_move_singleton_data.py
```diff
@@ -0,0 +1,89 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("registrations_objects_api", "0016_objectsapigroupconfig"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            move_singleton_data_to_objects_api_set_model,
+            reconfigure_objects_api_solo_model,
```
sergei-maertens, line 87: I can live with a `migrations.RunPython.noop`, I make sure to make a database backup before checking out branches like this locally :)

The approach of taking the first group is also fine, but if the above remarks are taken into account, for the content_json defaults etc. you can just fall back to the model field defaults.

src/openforms/registrations/contrib/objects_api/migrations/0019_add_default.py
sergei-maertens, line 1: this can be dropped entirely - and the zgw config/api group should get the same treatment.

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -25,34 +25,52 @@ def get_payment_status_update_text() -> str:
...
     """
 
-    objects_service = models.OneToOneField(
+    default_objects_api_group = models.ForeignKey(
+        to="ObjectsAPIGroupConfig",
+        on_delete=models.PROTECT,
+        verbose_name=_("default Objects API group."),
+        help_text=_("Which Objects API group should be used by default."),
+        null=True,
+    )
```
sergei-maertens, line 37: see earlier remarks - probably best to only keep legacy config here and then drop that too in Open Forms 3.0

src/openforms/registrations/contrib/objects_api/models.py
```diff
@@ -153,7 +171,11 @@ class ObjectsAPIConfig(SingletonModel):
     )
 
     class Meta:
-        verbose_name = _("Objects API configuration")
+        verbose_name = _("Objects API group")
+        verbose_name_plural = _("Objects API groups")
+
+    def __str__(self) -> str:
+        return self.name
```
sergei-maertens, line 178: Admin smoketests for models are a great way to also get this code covered :)

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -49,6 +49,15 @@ class ObjectsAPIRegistration(BasePlugin):
...
     configuration_options = ObjectsAPIOptionsSerializer
 
+    @staticmethod
+    def get_objects_api_config(options: RegistrationOptions) -> ObjectsAPIGroupConfig:
+        objects_api_group = options.get("objects_api_group")
+        if objects_api_group is None:
+            config = ObjectsAPIConfig.get_solo()
+            assert isinstance(config, ObjectsAPIConfig)
+            objects_api_group = config.default_objects_api_group
+        return objects_api_group  # type: ignore | can it really be None?
```
sergei-maertens, line 59: dropping the default simplifies this a lot

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/registrations/objectsapi/LegacyConfigFields.js
```diff
@@ -32,6 +34,33 @@ const LegacyConfigFields = ({index, name, formData, onChange}) => {
...
+        id="root_objectsApiGroup"
+        label={intl.formatMessage({
+          defaultMessage: 'Objects API group',
+          description: 'Objects API group',
+        })}
+        rawDescription={intl.formatMessage({
+          description: 'Objects API group selection',
+          defaultMessage:
+            'Which Objects API group to use. If not provided, the default Objects API group will be used.',
+        })}
```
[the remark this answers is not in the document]
Viicos: Is it fine if I do this as a feedback PR, and not by rebasing and rewriting migrations? :smiling_face_with_tear:

src/openforms/js/components/admin/form_design/registrations/objectsapi/utils.js
```diff
@@ -3,6 +3,15 @@ import React from 'react';
...
 // with the backend UI refactor.
 
+const getChoicesFromSchema = (enums, enumNames) => {
+  let finalChoices = {};
+  Object.keys(enums).forEach(key => {
+    finalChoices[enums[key]] = enumNames[key];
+  });
+
+  return finalChoices;
+};
```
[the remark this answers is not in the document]
Viicos: This was copy pasted from the zgw utils, I'll take a look for both

src/openforms/registrations/contrib/objects_api/migrations/0017_move_singleton_data.py
Viicos: Seems like you changed your mind from https://github.com/open-formulieren/open-forms/pull/3146/#discussion_r1228382591 :)

src/openforms/js/components/admin/form_design/registrations/objectsapi/LegacyConfigFields.js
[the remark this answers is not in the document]
sergei-maertens: I'm tempted to say "yes, if you can guarantee that the outcome of the data migrations is the same as if it was rebased", but on the other  hand, I find it quite wasteful to carry around (data) migrations and tests for them (that are super expensive in run time) that we already know are obsoleted...

Can't you append the necessary changes, and when everything is passing, delete the migration files and re-generate them again? That should give you the operations to add and remove models/fields, and you can then insert the `migrations.RunPython` operation in between that. If necessary, you can re-order the operations generated by Django, but that way you can have a single migration file that runs in a single DB transaction. I did that here: https://github.com/maykinmedia/mozilla-django-oidc-db/blob/master/mozilla_django_oidc_db/migrations/0002_migrate_to_claim_field.py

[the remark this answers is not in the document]
Viicos: Yeah looking at the necessary changes seems to be better to redo some migrations

src/openforms/registrations/contrib/objects_api/migrations/0017_move_singleton_data.py
sergei-maertens: ah, the growth of a person in a year's time :stuck_out_tongue: 

I don't have strong feelings about it - looking back I think the only ones actually doing rollbacks are developers, sometimes, when checking out branches locally to review/test things out. Making a DB dump and restoring that is probably less work than writing the reverse operation and writing tests for it :thinking:

Viicos: I did it as it was heavily inspired from Silvia's work on the ZGW groups, I'll see if I keep it when rewriting the migration history

src/openforms/registrations/api/filters.py
Viicos: Yes

src/openforms/registrations/contrib/zgw_apis/plugin.py
Viicos: Opened https://github.com/open-formulieren/open-forms/issues/4344

sergei-maertens:
Closing to declutter :)

src/openforms/registrations/contrib/objects_api/checks.py
```diff
@@ -10,93 +10,105 @@
...
-                _("Client error: {exception}").format(exception=exc)
+        urls = [
+            "objecttype",
+            "informatieobjecttype_submission_report",
+            "informatieobjecttype_submission_csv",
+            "informatieobjecttype_attachment",
+        ]
+        # we don't know for sure that these URLs share the same base, so instead we use a
+        # raw requests session to get connection pooling without tying this to the configured
+        # services.
```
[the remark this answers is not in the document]
Viicos: done in the new PR

PR #4336, opened 2024-05-29
DMN fixes

### LINKED ISSUE #4278: (Suspected) issues in DMN configuration admin
Business topic: DMN

* Configuring in/output variable mappings and then changing the decision definition to use -> doesn't clear the mapping. The dropdowns can't find the input/output variables and render in "empty" state, but there still *is* configuration behind the scenes.
* Removing a form variable likely does not remove it from the DMN input/output variable mappings
* A parameter that was removed in the decision definition (in a new version, for example, or renamed) is not cleared (similar to the two cases above)

Some UI controls to show there is still data with a dead reference is probably in order, the user can then decide to remove or update the mapping?

- [x] Clear mapping when decision definition is changed
- [x] Reset mapping for version changes (in modal)
- [x] Handle removing/editing of form variables (component + user defined)
    - [x] Reset form variable for input/output variable when a form variable is deleted
    - [x] Detect incomplete mappings and display a warning icon on:
        - the logic tab
        - the logic rule affected
        - the logic action for DMN affected
        - inside the modal when the DMN configuration modal is opened
    - [x] Update the key for changes in variables/components

sergei-maertens:
Closes #4278

**Changes**

* Display warning icons on logic tab & next to logic actions for invalid/incomplete DMN parameter mappings
* When a variable/component is deleted, clear the DMN mappings that have a reference (which triggers the above)
* When a variable/component key is renamed, update the DMN mappings to point to the new key
* Added a bunch more stories and refactored some components to be more/better re-usable

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/actions/Actions.js
+++ b/src/openforms/js/components/admin/form_design/logic/actions/Actions.js
@@ -21,6 +21,7 @@ import VariableSelection from 'components/admin/forms/VariableSelection';
 import Modal from 'components/admin/modals/Modal';
 
 import DMNActionConfig from './dmn/DMNActionConfig';
+import {detectMappingProblems as detectDMNMappingProblems} from './dmn/utils';
 import {ActionError, Action as ActionType} from './types';
 
 const ActionProperty = ({action, errors, onChange}) => {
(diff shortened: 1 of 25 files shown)


src/openforms/js/components/admin/form_design/logic/actions/Actions.js
```diff
@@ -347,4 +348,37 @@ ActionComponent.propTypes =
       onChange: PropTypes.func.isRequired,
     };
 
+export const detectProblems = (action, intl) => {
+  switch (action.action.type) {
```
vaszig, line 352: I would use a simple if statement here since we have only one case, do we use switch for possible update in the code?

src/openforms/js/components/admin/form_design/logic/actions/dmn/VariableMapping.js
```diff
@@ -85,7 +91,7 @@ const VariableMapping = ({
   alreadyMapped = [],
 }) => {
   const intl = useIntl();
-  const {getFieldProps, values} = useFormikContext();
+  const {values} = useFormikContext();
 
   const confirmationMessage = intl.formatMessage({
```
vaszig, line 96: Already there before your PR but if you change something else we can remove this I guess.

vaszig (approved):

src/openforms/js/components/admin/form_design/logic/actions/Actions.js
sergei-maertens: yeah, I'm doing it for consistency and future problem detection in other action types!

PR #4348, opened 2024-06-03
Refactor feature flags

### LINKED ISSUE #4346: Move feature flags into their own model
To declutter the global configuration & make managing feature flags in migrations easier.

- [ ] Adopt django-flags instead of our own model
- [ ] Add data migration for existing flags
- [ ] Add feature flag for OIDC callback URL (legacy vs. new) for #4246

sergei-maertens:
Closes #4346

**Changes**

* Moved the feature flags from the GlobalConfiguration to django-flags concepts (including data migration)
* Moved some fields around in the GlobalConfiguration that were in the wrong section (registration, payment related)
* Updated tests with settings-based feature flag enabling so we don't need to mock GlobalConfiguration anymore

More feature flag(s) will come with the DigiD/eHerkenning rework, so I figured it's nicer to be able to query just what is needed in terms of feature flags instead of always grabbing the whole solo model. There is a pointer from the old section to
the new one.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/authentication/contrib/digid_eherkenning_oidc/tests/digid/test_auth_plugin_endpoint.py
```diff
@@ -21,7 +21,7 @@ def setUp(self):
 
         self.client.force_authenticate(user=self.user)
 
-    @patch("openforms.plugins.registry.GlobalConfiguration.get_solo")
+    @patch("openforms.plugins.plugin.GlobalConfiguration.get_solo")
```
vaszig, line 24: Have we changed this in all the relative files in order to do the mock where the import is done for the first time?

vaszig (approved):

src/openforms/utils/tests/feature_flags.py
```diff
@@ -0,0 +1,19 @@
+from copy import deepcopy
+
+from django.conf import settings
+from django.test import override_settings
+
+
+def enable_feature_flag(flag: str):
+    assert flag in settings.FLAGS, f"Flag {flag} is not defined in settings.FLAGS"
+    flags = {
+        **deepcopy(settings.FLAGS),
```
Viicos, line 10: Is there a specific reason a `deepcopy` is necessary?

src/openforms/authentication/contrib/digid_eherkenning_oidc/tests/digid/test_auth_plugin_endpoint.py
sergei-maertens: yeah, it's a bit shitty because normally you don't mock things where they are defined, but this happens to work because the method is actually defined in another place.

Before this patch, the `GlobalConfiguration` model was imported in the `registry` module, but that's no longer the case, so the original mocks don't work anymore.

src/openforms/utils/tests/feature_flags.py
sergei-maertens: it's a better-safe-than-sorry approach, I really want to avoid the settings possibly being mutated accidentally which would break test isolation and cause hard-to-debug (flaky) test failures

PR #4349, opened 2024-06-03
Nudge people to the organisation login if OIDC enabled.

### LINKED ISSUE #4347: Make logging in with username/password less prominent when OIDC is enabled
### Thema / Theme

Admin

Currently, the username/password form is shown even when OIDC is enabled. When enabled, we assume this is the preferred way to login. Hence, the form should be less visible.

This feature can also be ported to the OIDC lib.

Less helpdesk support for new accounts.

_No response_

joeribekker:
Closes #4347

**Changes**

Hides the form when OIDC is disabled. Note that this is not the best code, but it shows the intent of what I want. It has 2 aspects: 1) nicer and better visible -or-, 2) less nudge towards form login

msedge_yrdXJ7k9O5

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/login.js
+++ b/src/openforms/js/components/admin/login.js
@@ -1,9 +1,16 @@
-const defaultLogin = document.querySelector('.admin-login-password');
-const loginForm = document.querySelector('#login-form');
+import {onLoaded} from 'utils/dom';
 
-if (defaultLogin && loginForm) {
-  defaultLogin.addEventListener('click', e => {
-    loginForm.style.display = 'block';
-    defaultLogin.style.display = 'none';
-  });
-}
+const addAdminLoginExpand = () => {
+  const defaultLoginToggle = document.querySelector('.admin-login-option--password');
+  const loginForm = document.querySelector('#login-form');
+
+  if (defaultLoginToggle && loginForm) {
+    defaultLoginToggle.addEventListener('click', e => {
+      e.preventDefault();
+      loginForm.classList.toggle('login-form--enabled');
+      defaultLoginToggle.classList.toggle('admin-login-option--disabled');
+    });
+  }
+};
+
+onLoaded(addAdminLoginExpand);
--- a/src/openforms/js/components/admin/login.js
+++ b/src/openforms/js/components/admin/login.js
@@ -1,0 +1,9 @@
+const defaultLogin = document.querySelector('.admin-login-password');
+const loginForm = document.querySelector('#login-form');
+
+if (defaultLogin && loginForm) {
+  defaultLogin.addEventListener('click', e => {
+    loginForm.style.display = 'block';
+    defaultLogin.style.display = 'none';
+  });
+}
--- a/src/openforms/scss/admin/_app_overrides.scss
+++ b/src/openforms/scss/admin/_app_overrides.scss
@@ -97,6 +97,7 @@ $djai-padding--tablet: 30px;
   clear: both;
   padding-top: 1em;
 }
+
 .admin-login-divider {
   text-align: center;
   clear: both;
@@ -106,13 +107,21 @@ $djai-padding--tablet: 30px;
   text-align: center;
   border-bottom: 1px solid #ccc;
   line-height: 0.1em;
-  margin: 10px 0 20px;
+  margin: 10px 0 10px;
 
   span {
     background: #fff;
     padding: 0 10px;
   }
-} 
+}
+
+/**
+ * mozilla-django-oidc-db
+ * hide the regular admin login form by default if OIDC is enabled
+ */
+#login-form:has(~ .admin-login-oidc) {
+  display: none;
+}
 
 /* Styling for react-jsonschema-form */
 .rjsf {
--- a/src/openforms/scss/admin/_app_overrides.scss
+++ b/src/openforms/scss/admin/_app_overrides.scss
@@ -97,6 +97,22 @@ $djai-padding--tablet: 30px;
   clear: both;
   padding-top: 1em;
 }
+.admin-login-divider {
+  text-align: center;
+  clear: both;
+  padding-top: 1em;
+
+  width: 100%;
+  text-align: center;
+  border-bottom: 1px solid #ccc;
+  line-height: 0.1em;
+  margin: 10px 0 20px;
+
+  span {
+    background: #fff;
+    padding: 0 10px;
+  }
+} 
 
 /* Styling for react-jsonschema-form */
 .rjsf {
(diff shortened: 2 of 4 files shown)


joeribekker:
@stevenbal could you prettify this code and port it?

src/openforms/js/components/admin/login.js
```diff
@@ -0,0 +1,9 @@
+const defaultLogin = document.querySelector('.admin-login-password');
+const loginForm = document.querySelector('#login-form');
+
+if (defaultLogin && loginForm) {
+  defaultLogin.addEventListener('click', e => {
+    loginForm.style.display = 'block';
+    defaultLogin.style.display = 'none';
+  });
+}
```
sergei-maertens, line 9: I find this makes it hard to track where style rules come from, the preferred approach is to toggle a class name, or, with our BEM approach, a modifier. You can then fully control the hide/display logic in CSS, and we can also (optionally) incorporate it in Storybook should we wish to do so.

sergei-maertens: you'll also want to put this code in `js/utils/dom::onLoaded` so that it doesn't matter if you load the script in the head tag or at the end of the body tag.

src/openforms/templates/maykin_2fa/login.html
```diff
@@ -1,11 +1,14 @@
 {% extends "maykin_2fa/login.html" %}
-{% load solo_tags i18n %}
+{% load solo_tags i18n static %}
 
 {% block extra_login_options %}
     {% get_solo 'mozilla_django_oidc_db.OpenIDConnectConfig' as oidc_config %}
     {% if oidc_config.enabled %}
-        <div class="admin-login-option">{% trans "or" %}</div>
-        <div class="admin-login-option">
+        <div class="admin-login-option admin-login-password">
```
sergei-maertens, line 7:
```suggestion
        <div class="admin-login-option admin-login-option--password">
```
would be the proper BEM approach - adding the `password` modifier lets you target it properly and it makes clear it's just a variant of a login option

src/openforms/templates/maykin_2fa/login.html
```diff
@@ -1,11 +1,14 @@
...
+            <a href="#">{% trans "Login with application account" %}</a>
```
sergei-maertens, line 8: I think a `<button>` element is semantically more correct, but I'm not sure if there are easy CSS classes available that make it look like a link. Either way, you want to call `event.preventDefault()` with links so that the `#` doesn't end up in the URL.

src/openforms/scss/admin/_app_overrides.scss
```diff
@@ -98,6 +98,31 @@ $djai-padding--tablet: 30px;
...
+  }
+}
+
+/**
+ * mozilla-django-oidc-db
+ * hide the regular admin login form by default if OIDC is enabled
+ */
+#login-form:has(~ .admin-login-oidc) {
+  display: none;
+}
```
sergei-maertens, line 124: I think I'd make these selectors more component-oriented with the proposed class names to follow BEM methodology.

```scss
.admin-login-option {
  @include bem.modifier('username') {
    ~ .admin-login-option--oidc {
      display: none;
    }
  }
}
```

src/openforms/scss/admin/_app_overrides.scss
```diff
@@ -98,6 +98,31 @@ $djai-padding--tablet: 30px;
...
+  text-align: center;
+  border-bottom: 1px solid #ccc;
+  line-height: 0.1em;
+  margin: 10px 0 10px;
+
+  span {
+    background: #fff;
+    padding: 0 10px;
+  }
+}
```
sergei-maertens, line 116: This doesn't work in dark mode:

image

You can basically never use literal values for colors anymore, but should always point to the relevant CSS variable. See `django/contrib/admin/static/admin/css/base.css` for a reference of built in django vars, and our own stylesheets define some extra variables you can use.

sergei-maertens, line 116: Alternative:

```suggestion
.admin-login-divider {
  display: flex;
  flex-direction: row;
  align-items: center;
  padding-block-start: 1em;

  &:after,
  &:before {
    content: '';
    display: block;
    border-bottom: 1px solid #ccc;
    flex-grow: 1;
  }

  > span {
    flex-shrink: 0;
    padding-inline: 10px;
  }
}
```

the flexbox makes this automatically centered without need for magic numbers, and you don't need the background color to create an overlay so it works in whatever theme you have (just need to fix the color reference for the border color).

sergei-maertens (changes requested):

src/openforms/scss/admin/_app_overrides.scss
stevenbal: But the rule at line 122 is meant to set `display: none;` on the `#login-form` itself, whereas your suggestion hides the OIDC option instead (if I understand it correctly)?

src/openforms/scss/admin/_app_overrides.scss
stevenbal: I'm not sure if I should add a new variable for the border color, or if I can use `$color_dark_light`?

src/openforms/templates/maykin_2fa/login.html
stevenbal: I'll keep it as a link and add `event.preventDefault` to the handler

src/openforms/scss/admin/_app_overrides.scss
sergei-maertens: I didn't try out the actual CSS, so yeah, probably should throw in a `:has(...)` in there! My point was more about using the BEM methodology :)

src/openforms/scss/admin/_app_overrides.scss
```diff
@@ -96,6 +96,42 @@ $djai-padding--tablet: 30px;
   text-align: center;
   clear: both;
   padding-top: 1em;
+
+  &--disabled {
+    display: none;
+  }
```
sergei-maertens, line 102: can you please use the mixins like in the sample I gave? These exist because the resulting selector now is `.admin-login-option--disabled`, and with the mixin it is `.admin-login-option.admin-login-option--disabled`.

A classname with modifier may not exist without the block/element itself, these sass mixins help enforce this.

stevenbal: I added it for `.admin-login-option--disabled`, though I'm not sure if this works for the `#login-form` block, because it's an id instead of a class :thinking:

sergei-maertens: yeah it doesn't work with IDs, since the whole idea of BEM is class-names :) for those you can spell them out manually entirely. ID's are an anti-pattern anyway for styling, I dislike that Django uses them so much in their CSS :(

sergei-maertens (approved):

PR #4353, opened 2024-06-04
✨ [#4337] Use form name in ZGW registration Zaak.omschrijving

### LINKED ISSUE #4337: Update ZGW registration to include form name in Zaak.omschrijving
The zaak is created here:

https://github.com/open-formulieren/open-forms/blob/cb6d2bf2c2c211f50fabb756cc080f0a854e98be/src/openforms/registrations/contrib/zgw_apis/plugin.py#L171

and the client passed the `omschrijving` like this:

https://github.com/open-formulieren/open-forms/blob/cb6d2bf2c2c211f50fabb756cc080f0a854e98be/src/openforms/contrib/zgw/clients/zaken.py#L34

Instead of this hardcoded `omschrijving` property, use the ``submission.form.name`` in the omschrijving field. Make sure to truncate this to max 80 chars.

stevenbal:
Closes #4337 

**Changes**

* Use form name in ZGW registration Zaak.omschrijving

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/registrations/contrib/zgw_apis/plugin.py
+++ b/src/openforms/registrations/contrib/zgw_apis/plugin.py
@@ -3,6 +3,7 @@ from functools import partial, wraps
 from typing import Any, TypedDict
 
 from django.urls import reverse
+from django.utils.text import Truncator
 from django.utils.translation import gettext, gettext_lazy as _
 
 import requests
@@ -171,6 +172,7 @@ class ZGWRegistration(BasePlugin):
             _create_zaak = partial(
                 zaken_client.create_zaak,
                 zaaktype=options["zaaktype"],
+                omschrijving=Truncator(submission.form.name).chars(80),
                 bronorganisatie=options["organisatie_rsin"],
                 vertrouwelijkheidaanduiding=options.get(
                     "zaak_vertrouwelijkheidaanduiding", ""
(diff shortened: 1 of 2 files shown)


src/openforms/utils/helpers.py
```diff
@@ -7,12 +9,19 @@ def get_charfield_max_length(
     return model_or_instance._meta.get_field(field_name).max_length
 
 
-def truncate_str_if_needed(original_str: str, new_str: str, max_length: int) -> str:
+def truncate_str_if_needed(
+    original_str: str, new_str: Optional[str], max_length: int
+) -> str:
```
sergei-maertens, line 14: I already had trouble reading this function to figure out what exactly it does, and these modifications don't make it easier.

Can't you instead use `django.utils.text.Truncator`?

```py
>>> from django.utils.text import Truncator
>>> Truncator("a string that is too long").chars(20)
'a string that is to…'
```

Probably this existing helper could be converted into that too, but let's leave it untouched for now.

sergei-maertens (changes requested):

src/openforms/utils/helpers.py
stevenbal: I didn't know about it either :open_mouth:. I changed it to use `Truncator` instead

sergei-maertens: it's what powers the template filters like https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#truncatechars :)

sergei-maertens (approved):

PR #4387, opened 2024-06-14
Prevent crash due to array-value for textfield|textarea defaultValue when multiple is true

### LINKED ISSUE #4362: Form designer crashes when selecting "multiple values"

`4c62dcf7ff4b9341c7b8be10619829a92a69d2ed`

### Customer reference

None

When selecting "multiple vlaues" in a textfield, textarea the form builder crashes: 
`TypeError: e.literal.localeCompare is not a function`

Not affecting other components, just the text components.

* Enable translations on the form
* Add textfield to form, check "allow multiple" and form builder should crash on "save"

## Formio-builder broken

After checking what we can do in the formio-builder, it turns out that entering the translations for the `multiple: true` variant doesn't work properly in the translation UI, so that needs fixing too. That might make backporting a lot harder though :(

sergei-maertens:
Closes #4362

Textfield and textarea have the multiple: true option, which turns the defaultValue into an array of strings rather than a single string, and that doesn't play well with the translations machinery or missing translation detection.

Open Forms 2.7.0 removes support for translations for this property, and older versions get a hotfix to prevent the crash from happening during detection, where we just don't look at `defaultValue` to extract missing translations.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/formio_builder/translation.js
+++ b/src/openforms/js/components/formio_builder/translation.js
@@ -30,8 +30,6 @@ const ALWAYS_TRANSLATABLE_PROPERTIES = ['label', 'description', 'placeholder', '
  *   test the property path against.
  */
 const ADDITIONAL_PROPERTIES_BY_COMPONENT_TYPE = {
-  textfield: ['defaultValue'],
-  textarea: ['defaultValue'],
   content: ['html'],
   fieldset: ['legend'],
   editgrid: ['groupLabel'],


vaszig (approved):

sergei-maertens:
Other backports:

* 2.5.x: 9b09ec16f55589ddf96d360f6ecf0c1f8a8108b3
* 2.4.x: 1be2c48b7e1036bd33e02988cc57dbc2331663bb

PR #4422, opened 2024-06-19
Disallow using temporary file uploads not owned

Viicos:
Closes https://github.com/open-formulieren/security-issues/issues/30

**Changes**

We use the request from the context to validate that the user owns the temporary file uploads.

I'm not too happy with the test implementation but I don't think we can do any better.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/formio/tests/validation/test_file.py
+++ b/src/openforms/formio/tests/validation/test_file.py
@@ -1,18 +1,24 @@
 from pathlib import Path
 from unittest.mock import Mock, patch
 
+from django.contrib.sessions.middleware import SessionMiddleware
 from django.core.files import File
 from django.test import TestCase, tag
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework.settings import api_settings
+from rest_framework.test import APIRequestFactory
+from rest_framework.utils.serializer_helpers import ReturnDict
 
 from openforms.config.models import GlobalConfiguration
+from openforms.submissions.attachments import temporary_upload_uuid_from_url
+from openforms.submissions.constants import UPLOADS_SESSION_KEY
 from openforms.submissions.tests.factories import TemporaryFileUploadFactory
+from openforms.typing import JSONValue
 
 from ...typing import FileComponent
 from ..factories import SubmittedFileFactory
-from .helpers import extract_error, validate_formio_data
+from .helpers import extract_error, validate_formio_data as _validate_formio_data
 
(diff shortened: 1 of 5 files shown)


src/openforms/formio/tests/validation/test_file.py
```diff
@@ -29,6 +35,19 @@
...
 
+def validate_formio_data(
+    component: FileComponent, data: JSONValue, session_uploads: list[str] = []
+) -> tuple[bool, ReturnDict]:
+    request = APIRequestFactory().put("/irrelevant")
+
+    # Add a session to the request:
+    middleware = SessionMiddleware(lambda x: None)  # type: ignore
+    middleware.process_request(request)
+    request.session[UPLOADS_SESSION_KEY] = [str(uuid) for uuid in session_uploads]
```
sergei-maertens, line 46: I took a different approach in other tests: https://github.com/maykinmedia/mozilla-django-oidc-db/blob/master/tests/conftest.py#L120

src/openforms/submissions/api/validation.py
```diff
@@ -71,7 +71,10 @@ def _run_formio_validation(self) -> None:
                 step_data_serializer = build_serializer(
                     configuration["components"],
                     data=data,
-                    context={"submission": submission},
+                    context={
+                        "submission": submission,
+                        "request": self.context["request"],
+                    },
```
sergei-maertens, line 77: I'm not sure what's wise here - this addresses the problem indeed, but it does tie formio validation to a request-response cycle and as you noticed, it makes test code not particularly elegant.

I think I wrote down somewhere that IMO there's no reason why the uploads need to be tracked individually in the session - I think we should be able to:

* send the submission UUID/URL during the file upload in the SDK
* link the temporary upload to the submission record. the security check here then only needs to validate ownership of the submission
* when validating, we already have access to the submission instance, so we should only consider temporary uploads within that submission

This has a number of benefits:

* less session polution (and much nicer: race conditions when updating the array of values!)
* all information is contained within a submission, which doesn't require a request/response cycle for processing (that includes validation)
* when submissions are pruned, temporary uploads still left over (because of whatever reason) can be cascade-deleted, so we don't need the periodic cleanup of temporary uploads anymore

sergei-maertens: Uploads are created before submission steps are.

Viicos:
Closing in favor of https://github.com/open-formulieren/open-forms/pull/4469

PR #4443, opened 2024-06-26
 [#4414] Remove default ZGW API group 

### LINKED ISSUE #4414: Remove default ZGW group config
### Thema / Theme

Other

To align with the new Objects API Group, the `ZgwConfig` model should be removed, with proper data migrations involved.

_No response_

_No response_

Viicos:
Closes #4414

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/registrations/contrib/zgw_apis/migrations/0013_set_zgw_api_group.py
```diff
@@ -0,0 +1,45 @@
...
+    else:
+        default_group = None
+
+    for registration_backend in FormRegistrationBackend.objects.filter(
+        backend="zgw-create-zaak"
+    ):
+        zgw_api_group = registration_backend.options.get("zgw_api_group")
+        if zgw_api_group is None:
+            # This should not happen and is guaranteed by an upgrade check:
+            assert default_group is not None
```
sergei-maertens, line 27: it can and will happen if a form was imported and the configs are deleted, or the config was never set up but a form from another env was imported (on our test env).

The upgrade checks can also be bypassed since they run on the system check framework, so let's be a bit more defensive here - if no default group exists but there are matching registration backends -> just create an empty group, the rest will be picked up by digest email and/or runtime crashes on the form.

The latter is preferred over not being able to `migrate`, since the problem can be fixed in the UI once the instance is updated.

sergei-maertens (changes requested):
Can you rebase on master and re-generate API spec/translations?

Otherwise this also cleans up some code quite nicely, good improvement.

src/openforms/registrations/contrib/zgw_apis/migrations/0013_set_zgw_api_group.py
```diff
@@ -0,0 +1,47 @@
...
+        default_group = None
+
+    for registration_backend in FormRegistrationBackend.objects.filter(
+        backend="zgw-create-zaak"
+    ):
+        zgw_api_group = registration_backend.options.get("zgw_api_group")
+        if zgw_api_group is None:
+            if default_group is None:
+                # This can only happen if upgrade checks were bypassed.
+                default_group = ZGWApiGroupConfig.objects.create(name="Default group")
```
sergei-maertens, line 29: can you use the literal name I gave in my earlier PR feedback? This was deliberate so that attention is drawn to it in the admin and we can clearly see it was a system-generated group.

sergei-maertens (changes requested):

src/openforms/registrations/contrib/zgw_apis/migrations/0013_set_zgw_api_group.py
sergei-maertens, line 29:
```suggestion
                default_group = ZGWApiGroupConfig.objects.create(name="AUTO_GENERATED - FIXME",)
```
let's do the same thing like in `0019_add_default_objects_api_group`

PR #4445, opened 2024-06-27
[#4400] Fix email digest concerning invalid JSON logic expressions

### LINKED ISSUE #4400: Invalid JSON logic expressions crash the digest email
Sentry 357891

Traceback:

```
ValueError: Operator 'object' is unknown (unregistered in 'json_logic.operations' nor in 'json_logic.scoped_operations').
  File "celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
  File "openforms/emails/tasks.py", line 63, in send_email_digest
    content = digest.render()
  File "openforms/emails/tasks.py", line 48, in render
    context = self.get_context_data()
  File "openforms/emails/tasks.py", line 35, in get_context_data
    invalid_logic_rules = collect_invalid_logic_rules()
  File "openforms/emails/digest.py", line 399, in collect_invalid_logic_rules
    form_logics_vars += introspect_json_logic(
  File "openforms/utils/json_logic/introspection.py", line 89, in introspect_json_logic
    tree = JSONLogicExpression.from_expression(expression).as_tree()
  File "json_logic/meta/expressions.py", line 64, in as_tree
    arguments = [
  File "json_logic/meta/expressions.py", line 65, in <listcomp>
    JSONLogicExpression.from_expression(value).as_tree() for value in values
  File "json_logic/meta/expressions.py", line 67, in as_tree
    return Operation.for_operator(
  File "json_logic/meta/base.py", line 72, in for_operator
    return operator_cls(operator, *args, **kwargs)
  File "<string>", line 7, in __init__
    OperationArgument = Union[Primitive, "Operation"]
  File "json_logic/meta/base.py", line 45, in __post_init__
    raise ValueError(
```

Offending JsonLogic expression:

image

vaszig:
Closes #4400

**Changes**

- Added regression tests
- Handle exceptions in broken JSON logic rules

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/emails/digest.py
+++ b/src/openforms/emails/digest.py
@@ -13,6 +13,7 @@ from django.utils.translation import gettext_lazy as _
 
 from django_yubin.models import Message
 from furl import furl
+from json_logic.typing import JSON
 from rest_framework import serializers
 from simple_certmanager.models import Certificate
 
@@ -128,6 +129,8 @@ class InvalidLogicRule:
     variable: str
     form_name: str
     form_id: int
+    exception: bool = False
+    rule_index: int = 0
 
     @property
     def admin_link(self) -> str:
@@ -366,7 +369,35 @@ def collect_invalid_registration_backends() -> list[InvalidRegistrationBackend]:
     return invalid_registration_backends
 
 
+def introspect_json_logic_wrapper(
+    expression: JSON, form_name: str
+) -> list[InputVar] | None:
+    try:
+        introspection_result = introspect_json_logic(expression).get_input_keys()
+    except Exception as e:
+        logger.error(
+            "malformed/unsupported JsonLogic expression in form %s: %r",
+            form_name,
+            expression,
+            exc_info=e,
+        )
+        return None
+
+    return introspection_result
+
+
 def collect_invalid_logic_rules() -> list[InvalidLogicRule]:
+    def _report(var_key: str = "", exception: bool = False, rule_index: int = 0):
+        invalid_logic_rules.append(
+            InvalidLogicRule(
+                variable=var_key,
+                form_name=form.admin_name,
+                form_id=form.id,
+                exception=exception,
+                rule_index=rule_index,
+            )
+        )
+
     forms = Form.objects.live().iterator()
     static_variables = {
         var.key: {"source": var.source, "type": var.data_type}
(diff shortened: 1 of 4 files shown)


src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -136,11 +136,19 @@ <h5>{% trans "Logic rules problems" %}</h5>
         <ul>
         {% for logic_rule in invalid_logic_rules %}
             <li>
+                {% if logic_rule.exception %}
+                    <p>
+                        {% blocktranslate with form_name=logic_rule.form_name trimmed %}
+                            We couldn't process the logic rule for '{{ form_name }}' because it is completely broken. <br>
```
sergei-maertens, line 142:
```suggestion
                            We couldn't process logic rule {{ index }} for '{{ form_name }}' because it appears to be invalid. <br>
```
and we'd also have to keep track of the index/order of the logic rule to display it. I think this is the best we can do to point them to the problematic rule.

I've also softened the language a bit :P

src/openforms/utils/json_logic/introspection.py
```diff
@@ -85,8 +85,12 @@
 }
 
 
-def introspect_json_logic(expression: JSON) -> ExpressionIntrospection:
-    tree = JSONLogicExpression.from_expression(expression).as_tree()
+def introspect_json_logic(expression: JSON) -> ExpressionIntrospection | str:
```
sergei-maertens, line 88: I wouldn't do the exception handling here, but in a wrapper function that you define in `digest.py`

Converting exceptions into strings for code that is used in other places for other purposes seems like a bad idea - and a type checker would also start complaining here for all the other call-sites to `introspect_json_logic`!

As a rule of thumb - if you're modifying function signatures, you need to be damn sure it's really needed :)

src/openforms/emails/digest.py
```diff
@@ -389,16 +400,37 @@ def collect_invalid_logic_rules() -> list[InvalidLogicRule]:
 
         form_logics_vars = []
         for logic in form_logics:
-            form_logics_vars += introspect_json_logic(
-                logic.json_logic_trigger
-            ).get_input_keys()
+            logic_introspection_result = introspect_json_logic(logic.json_logic_trigger)
+
+            if isinstance(logic_introspection_result, str):
```
sergei-maertens, line 405: so this approach still doesn't really work in terms of code elegance - you've now essentially traded a `try: except` with a `isinstance` check - and the handling code is still duplicated (reporting, logging...)

sergei-maertens (changes requested):

src/openforms/emails/digest.py
```diff
@@ -366,7 +368,33 @@ def collect_invalid_registration_backends() -> list[InvalidRegistrationBackend]:
...
+    expression: JSON, form_name: str
+) -> list[InputVar] | None:
+    try:
+        introspection_result = introspect_json_logic(expression).get_input_keys()
+    except Exception as e:
+        logger.info(
+            "exception in json logic (rule - action) of form %s: %s",
+            form_name,
+            str(e),
+        )
```
sergei-maertens, line 381:
```suggestion
        logger.error(
            "malformed/unsupported JsonLogic expression in form %s: %r",
            exprssion,
            form_name,
            exc_info=e,
        )
```
logging is an art :) but if you pass the exception via `exc_info`, then that will be added to the traceback/logging infrastructure and Sentry for example should show the exception information.

Additionally, since this causes a proper crash, a log level of `error` is not too much. By default `warning` and higher end up in Sentry.

Stringified exceptions in log messages tend not to be too useful, especially because causes of exceptions get lost (if you do `raise X from Y`, the `from Y` information gets lost I think, which doesn't happen with the `exc_info=e` mechanism.

src/openforms/emails/digest.py
```diff
@@ -389,16 +417,30 @@ def collect_invalid_logic_rules() -> list[InvalidLogicRule]:
...
+                logic.json_logic_trigger, form.admin_name
+            )
+
+            if logic_introspection_result is None:
+                _report(exception=True)
+
+                # no need for checking actions since the logic rule is already broken
+                continue
+            else:
+                form_logics_vars += logic_introspection_result
```
sergei-maertens, line 430: you can drop the `else` and remove a level of indentation - the `continue` above already skips the rest of the loop execution.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -136,11 +136,19 @@ <h5>{% trans "Logic rules problems" %}</h5>
         <ul>
         {% for logic_rule in invalid_logic_rules %}
             <li>
+                {% if logic_rule.exception %}
+                    <p>
+                        {% blocktranslate with index=forloop.counter form_name=logic_rule.form_name trimmed %}
+                            We couldn't process logic rule {{ index }} for '{{ form_name }}' because it appears to be invalid. <br>
```
sergei-maertens, line 142: ah - you don't want to use the template for loop counter, but the python code forloop counter! If rule 1 and 2 are fine, but rule 3 has an error, then this would be reported as if it's a problem in rule 1 which is wrong.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #4448, opened 2024-06-27
Bump DRF & tinymce dependencies

sergei-maertens:
django-tinymce upgrade aligns the tinymce versions used in React and the admin, so that's nice. The breaking changes don't appear to affect us (but CI can say otherwise of course).

The DRF upgrade is a YOLO - we'll see what breaks in this PR :grimacing:


Viicos (approved):

sergei-maertens:
I don't understand why CI is not passing after retries. The e2e tests on chromium do pass locally for me :/

PR #4449, opened 2024-06-27
Add registration variables for auth context (objects API registration v2)

### LINKED ISSUE #4246: Capture authentication context data
Part of #3623

Currently the "machtigen" context data is stored as a simple JSON field, which doesn't capture everything that should be captured, and the shape is also not well documented. For this, some changes are needed:

- [x] Patch our OIDC library so that we can map properties with <kbd>.</kbd> characters too
- [x] Release new version of mozilla-django-oidc-db
- [x] Investigate removing the "django-solo" nature of the OIDC config model(s) (and do more with JSONField for config)
- [x] Establish structure of authentication context in models (`authentication.AuthInfo`)
    - Record/translate means (digid/eherkenning/...)
    - Record LOA
    - Record authorizee
    - Record representee
    - Record mandate context (service(s), role...)
- [x] Implement `AuthInfo.to_auth_context_data()` method and validate that against the JSON schema's (for robustness)
- [x] Extract necessary claims/context data in authentication plugins and save the data
- [x] Add context var to Objects API registration v1
- [x] Add registration var to Objects API registration v2
- [x] Update static variables to read from the appropriate place(s)
- [x] Update prefill to read from the appropriate place(s)

sergei-maertens:
Partly closes #4246

**Changes**

* Added registration variable for full auth context data
* Added registration variables for the nested properties inside (until we can handle complex schema's with better typing support)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
```diff
@@ -376,6 +376,33 @@ def test_hidden_component(self):
...
+        submission = SubmissionFactory.create(
+            with_public_registration_reference=True,
+            public_registration_reference="OF-911",
+            auth_info__plugin="irrelevant",
+            auth_info__is_eh_bewindvoering=True,
+        )
+        ObjectsAPIRegistrationData.objects.create(submission=submission)
+        v2_options: RegistrationOptionsV2 = {
+            "version": 2,
+            # "objects_api_group": self.objects_api_group.pk,
```
sergei-maertens, line 389: @Viicos these are because of your PR - ideally yours can get merged first and then I can rebase these tests

src/openforms/registrations/contrib/objects_api/registration_variables.py
Viicos, line 1: That's a lot of registration variables, I hope it won't get too messy in the UI. I think a nice future improvement would be to have a description field on these variables, that could be displayed as a hover pop up in the admin

src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
Viicos: Sounds good. Isn't the `objects_api_group` available on `master`?

Viicos (approved):

src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
sergei-maertens: no, your PR updates the `setUpTestData`

src/openforms/registrations/contrib/objects_api/registration_variables.py
sergei-maertens: I want to go further - other than description, the JSON schema of each variable should also be included. But out of scope for now.

I just realized that these new variables belong in the static variables, not just in objects API so I'm moving them.

sergei-maertens:
@stevenbal can you review the documentation changes, as they're in Dutch? :grimacing:

src/openforms/authentication/models.py
```diff
@@ -290,7 +296,7 @@ def to_auth_context_data(
             # EHerkenning without machtigen/mandate
             # Presence of an acting subject confirms eHerkenning (or eidas? TODO),
             # as opposed to a registrator entering the KVK number.
-            case (AuthAttribute.kvk, "") if self.acting_subject_identifier_value:
+            case (AuthAttribute.kvk, "") if not has_registrator:
```
Viicos, line 299: Why not:

```suggestion
            case (AuthAttribute.kvk, "") if not submission.has_registrator:
```

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -61,8 +72,7 @@ class AuthKvK(BaseStaticVariable):
     name = _("Authentication KvK")
     data_type = FormVariableDataTypes.string
 
-    @staticmethod
```
Viicos, line 64: Nice, I had this on my list of TODO cleanups

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+
+De "authenticatie context" bevat alle beschikbare informatie over *hoe* iemand ingelogd
+is op een formulier. Deze bundel informatie is beschikbaar in de vaste variabele
+``auth_context``.
+
+Wanneer er niet ingelogd is op het formulier, dan is de waarde van deze variabele
+``null``.
+
+.. note:: Het is een nieuwere en volledigere variant van de ``auth`` variabele.
+   Er zijn geen plannen om deze laatste te verwijderen.
```
stevenbal, line 87: Does this mean that `auth.plugin` and `auth.value` (which are both in the static variables table) could also be taken from `auth_context`? Wouldn't that be the preference since it's mentioned that `auth_context` is newer and more complete?

Also it is mentioned that we will not delete `auth`, but above it is mentioned that parts of `auth` (like `machtigen` and maybe `attribute` will be deleted. Perhaps we should put all this in the same info/warning box, because if you just read this info from the note it might seem like using everything from `auth` is fine

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+
+De variabele bevat een bak aan informatie, gestructureerd volgens het
+authenticatiecontextdatamodel (TODO: add link). De structuur is als volgt:
```
stevenbal, line 90: will this be added later?

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    "id": "string",
+                    "uuid": "string"
+                }
+            ]
+        }
+    }
+
+Merk op dat niet alle attributen aanwezig zijn, dit hangt af van het inlogmiddel (
+DigiD, eHerkenning) en of er wel/niet sprake is van een machtiging én de soort
```
stevenbal, line 123:
```suggestion
Merk op dat niet alle attributen aanwezig zijn, dit hangt af van het inlogmiddel (DigiD, eHerkenning) en of er wel/niet sprake is van een machtiging én de soort
```

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+* ``source``, mogelijk lege string als waarde
+* ``levelOfAssurance``, mogelijk lege string als waarde
+* ``authorizee``
+* ``authorizee.legalSubject``
+* ``authorizee.legalSubject.identifierType``, mogelijk lege string als waarde
+* ``authorizee.legalSubject.identifier``, mogelijk lege string als waarde
+
+De onderdelen van deze structuur worden ook als individuele variabelen aangeboden:
+
+``auth_context_source``
```
stevenbal, line 137: I'm not sure if it's practical here (because the descriptions are quite long for some variables), but it might be good to put these in a similar table as the static variables described above. If you glance through these docs it might not be immediately clear that these are extra static variables on top of the ones described in the table

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+    Middel van inloggen: de waarde is ``"digid"`` of ``eherkenning``, of een lege string
```
stevenbal, line 138:
```suggestion
    Middel van inloggen: de waarde is ``"digid"`` of ``"eherkenning"``, of een lege string
```

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+    is van vertegenwoordiging). Leeg indien het formulier zonder inloggen gestart is.
+
+``auth_context_legal_subject_identifier``
+    Identificatie van de (wettelijke) vertegenwoordiger. Leeg indien het formulier
+    zonder inloggen gestart is.
+
+``auth_context_acting_subject_identifier_type``
+    In de praktijk zal de waarde altijd ``opaque`` of leeg zijn. Geeft aan hoe de
+    identificatie van de handelende persoon ("de persoon aan de knoppen")
+    geïnterpreteerd moet worden.
```
stevenbal, line 164: I'm not sure what this means, does this mean that the value literally is `"opaque"`? Is this like a convenience variable to not have to check if machtigen is used when determining the acting subjects auth info?

docs/manual/templates.rst
```diff
@@ -686,57 +686,16 @@ Je kan doorgeven op welke manier een gebruiker al dan niet ingelogd was tijdens
...
-    * ``{{ auth_context.authorizee.actingSubject }}``: object die de handelende persoon
-      identificeert bij vertegenwoordiging. Dit is enkel aanwezig wanneer het
-      eHerkenning betreft.
-
-      De attributen ``{{ auth_context.authorizee.actingSubject.identifierType }}`` en
-      ``{{ auth_context.authorizee.actingSubject.identifier }}`` hebben een
-      gelijkaardige betekenis als bij de ``representee``.
+Zie :ref:`manual_forms_variables_auth_context` voor een voorbeeld van de structuur, en
+een overzicht van alle "onderdelen" waaruit de ``auth_context`` variabele bestaat. Je
+kan deze allemaal individueel gebruiken in de sjablonen.
```
stevenbal, line 701: Is there a preference when it comes to  using these variables and does it make a difference (aside from number of characters), e.g. `auth_context_loa` vs `auth_context.levelOfAssurance`?

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -26,6 +26,17 @@ class Registry(BaseRegistry[BaseStaticVariable]):
 """The Objects API registration variables registry."""
 
 
+@register("public_reference")
+class PublicReference(BaseStaticVariable):
```
stevenbal, line 30: I don't see this variable mentioned in the `Vaste variabelen` section, should this be mentioned there? Or is this mentioned in a specific page for Objects API registration?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -336,11 +336,6 @@ def get_object_data(
                 "pdf_url": registration_data.pdf_url,
                 "csv_url": registration_data.csv_url,
             },
-            "auth_context": (
-                submission.auth_info.to_auth_context_data()
-                if submission.is_authenticated
-                else None
-            ),
```
stevenbal, line 343: Just for my understanding, this is no longer needed because the `auth_context` variable is always available for the registration backend now?

stevenbal (changes requested):
Minor comments regardings the docs

src/openforms/authentication/models.py
sergei-maertens: completely missed that property exists :-) will update!

src/openforms/authentication/static_variables/static_variables.py
sergei-maertens: I hate these type error though, semantically this *is* perfectly fine to be a static method.

docs/manual/forms/variables.rst
```diff
@@ -0,0 +1,170 @@
...
+auth               object                                Een verzameling van authenticatie gegevens. Zie hieronder.
+auth.plugin        string    ``digid``                   De systeemnaam van de gebruikte authenticatie plugin.
+auth.value         string    ``111222333``               De identificerende waarde in het ``attribute`` van de
+                                                         authenticatie plugin.
+auth_type          string    ``bsn``                     Kan de waarden ``bsn``, ``kvk`` of ``pseudo`` hebben.
+auth_bsn           string    ``111222333``               De waarde van ``auth.value`` indien ``auth_type`` als waarde
+                                                         ``bsn`` heeft. Anders leeg.
+auth_kvk           string    ``90001354``                De waarde van ``auth.value`` indien ``auth_type`` als waarde
+                                                         ``kvk`` heeft. Anders leeg.
+auth_pseudo        string    ``a8bfe7a293dd``...         De waarde van ``auth.value`` indien ``auth_type`` als waarde
```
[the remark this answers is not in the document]
sergei-maertens: indeed - what exactly it means and what you can do with it is still unclear, thought it's now an active topic at Dimpact :)

docs/manual/forms/variables.rst
sergei-maertens: no, that's also why they aren't removed yet :) IMO there should not be a reason to use `auth.plugin` - the actual plugin used does not matter for semantic checks, instead you want to know if digid/eherkenning auth was used so it's more about the `auth.attribute`. Unfortunately, logic rules are now built on top of `auth.plugin` because people don't know any better.

`auth.value` is a bit more nuanced - it will always contain the identifier of the ultimate beneficiary of the service/product being requested with the form. With `auth_context`, the property holding this value is either `authorizee.legalSubject.identifier` or `representee.identifier`, depending on whether it's digid/eherkenning and/or machtigen is involved or not, which is more complicated to express in form logic.

> because if you just read this info from the note it might seem like using everything from auth is fine

yeah, I'll revisit it tomorrow, the language can definitely be improved!

docs/manual/forms/variables.rst
sergei-maertens: yes, though I need to check if I can share those gitbook links and people outside of our team probably need to review/tweak the content, so I don't think we should block this PR on that TODO.

docs/manual/forms/variables.rst
sergei-maertens: I don't see the difference or mistake :sweat_smile:

docs/manual/forms/variables.rst
sergei-maertens: I tried putting them in a table initially and then pulled them out because RST tables suck especially with long un-breakable words/code. It makes the actual RST unreadable due to line wrapping/horizontal scrollbars.

docs/manual/forms/variables.rst
sergei-maertens: the identifier type value is indeed `"opaque"`, as opposed to `"bsn"` or `"kvkNummer"` (or the future `"RSIN"`).

Currently it's the only possible type, but in the future more types could be possible (this could for example be used for a registrator at the service desk too, at some point). It basically means that you have zero guarantees about the meaning of the identifier value except that it's a value tied to a particular staff member. It *could* look like a BSN, but if you use it to consult the BRP, it may sometimes work, it will most likely not. It is not expected to be decrypted.

docs/manual/templates.rst
sergei-maertens: there is a slight difference in that the static variables will always yield a `string`, while missing other attributes could theoretically return `None` and template authors would have to check for that.

In practice, I think the DTL fallback to empty string will result in the same outcome (most of the time).

src/openforms/registrations/contrib/objects_api/registration_variables.py
sergei-maertens: it's not a static variable, but a registration variable and visible in the UI when you select an objects API registration backend :) and there the title + type are displayed (it's "self-documenting" via the registry).

src/openforms/registrations/contrib/objects_api/submission_registration.py
sergei-maertens: yeah, precisely. I assumed wrongly that the `auth_context` should be provided only as objects registration variables, but it acts at the same level as the existing `auth` static variable, so we can expose it directly to all registration backends automatically :)

stevenbal (approved):

PR #4469, opened 2024-06-28
Disallow using temporary file uploads not owned

Viicos:
Closes https://github.com/open-formulieren/security-issues/issues/30, replaces https://github.com/open-formulieren/open-forms/pull/4422

**Changes**

Alternative solution as suggested in https://github.com/open-formulieren/open-forms/pull/4422#discussion_r1646472465

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


Viicos:
I've tried looking at how to implement this in the SDK: we should add the submission uuid https://github.com/open-formulieren/open-forms-sdk/blob/7058f3609be5ac6736d8df026cb8e2324cb16389/src/formio/components/FileField.js#L123-L125, but this is way easier said than done :/

src/openforms/formio/api/serializers.py
```diff
@@ -25,6 +25,13 @@ class TemporaryFileUploadSerializer(serializers.Serializer):
         use_url=False,
         validators=[MimeTypeValidator(), NoVirusValidator()],
     )
+    submission = serializers.SlugRelatedField(
+        slug_field="uuid",
+        queryset=Submission.objects.filter(completed_on=None),
+        label=_("Submission"),
+        write_only=True,
+        required=True,
+    )
```
sergei-maertens, line 34: we use hyperlinked model serializers everywhere else, please use a hyperlink instead of just the UUID.

src/openforms/formio/api/views.py
```diff
@@ -59,13 +63,19 @@ def post(self, request, *args, **kwargs):
                 content_type="text/plain",
             )
 
+        submission = serializer.validated_data["submission"]
+
+        if not owns_submission(request, submission.uuid):
+            raise PermissionDenied()
```
sergei-maertens, line 69: missing tests for permission check, but alternatively you can also make the queryset for the `submission` field in the serializer dynamic based on `self.context["request"].session` to filter out only submissions that are owned.

I find permission checks outside of permission classes not the most elegant pattern.

sergei-maertens, line 69: additionally, this doesn't result in an response with `text/plain` content type, which is required for Formio upload fields.

src/openforms/formio/api/serializers.py
```diff
@@ -3,12 +3,12 @@
...
 from rest_framework.reverse import reverse
 
-from openforms.submissions.models import TemporaryFileUpload
+from openforms.submissions.models import Submission, TemporaryFileUpload
 
 from .validators import MimeTypeValidator, NoVirusValidator
 
 
-class TemporaryFileUploadSerializer(serializers.Serializer):
+class TemporaryFileUploadSerializer(serializers.ModelSerializer):
```
sergei-maertens, line 11: I don't see the benefit of changing this into a `ModelSerializer` if the record is created directly in the view anway.

What breaks if you don't do this?

src/openforms/submissions/tests/factories.py
```diff
@@ -312,12 +312,19 @@ class Meta:
...
     content = factory.django.FileField(filename="file.dat", data=b"content")
 
     class Meta:
         model = TemporaryFileUpload
 
+    class Params:
+        is_legacy = factory.Trait(
+            legacy=True,
+            submission=None,
+        )
```
sergei-maertens, line 326: I recently learned about `factory.Maybe` :) so there's a way to automatically make this so if `legacy=True` is provided.

src/openforms/formio/rendering/tests/test_vanilla_formio_components.py
```diff
@@ -539,9 +539,13 @@ def test_nested_files(self):
...
-        upload_in_repeating_group_1 = TemporaryFileUploadFactory.create()
-        upload_in_repeating_group_2 = TemporaryFileUploadFactory.create()
-        nested_upload = TemporaryFileUploadFactory.create()
+        upload_in_repeating_group_1 = TemporaryFileUploadFactory.create(
+            submission=submission
+        )
+        upload_in_repeating_group_2 = TemporaryFileUploadFactory.create(
+            submission=submission
+        )
+        nested_upload = TemporaryFileUploadFactory.create(submission=submission)
```
sergei-maertens, line 548: could also go for:

```py
upload_in_repeating_group_1, upload_in_repeating_group_2, nested_upload = TemporaryFileUploadFactory.create_batch(
    3, submission=submission
)
```

src/openforms/formio/tests/test_api_fileupload.py
```diff
@@ -43,14 +43,18 @@ def test_upload_view_requires_active_submission(self):
             "my-file.txt", b"my content", content_type="text/plain"
         )
 
-        response = self.client.post(url, {"file": file}, format="multipart")
+        response = self.client.post(
+            url, {"file": file, "submission": self.submission.uuid}, format="multipart"
+        )
```
sergei-maertens, line 48: the SDK compatibility matrix in the docs requires updating for this, it means that the current SDK is no longer compatible with the 2.7 API version.

src/openforms/formio/tests/validation/helpers.py
```diff
@@ -13,12 +14,12 @@
 
 
 def validate_formio_data(
-    component: Component, data: JSONValue
+    component: Component, data: JSONValue, extra_context: dict[str, Any] = {}
```
sergei-maertens, line 17: can we not provide properly defined/typed kwargs for extra context? The structure of context is well-known since it's built when the validation is invoked.

src/openforms/formio/tests/validation/test_file.py
```diff
@@ -204,35 +250,65 @@ def test_no_temporary_upload(self):
...
+    def test_temporary_file_upload_invalid_submission(self):
+        temporary_file_upload = TemporaryFileUploadFactory.create()
+        data = SubmittedFileFactory.create(temporary_upload=temporary_file_upload)
+
+        # Not giving any extra context, so that `validate_formio_data` will create
+        # a dummy submission for us, which will be != `temporary_file_upload.submission`:
+        is_valid, errors = validate_formio_data(
+            DEFAULT_FILE_COMPONENT,
+            {"foo": [data]},
+        )
```
sergei-maertens, line 296: and explicitly created different submission would be better for that purpose instead of relying on implicit shims

src/openforms/submissions/migrations/0006_set_legacy_true.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def set_legacy_true(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    """Alter existing ``TemporaryFileUpload`` instances to have ``legacy`` set to ``True`` by default."""
+
+    TemporaryFileUpload = apps.get_model("submissions", "TemporaryFileUpload")
+
+    for file_upload in TemporaryFileUpload.objects.iterator():
+        file_upload.legacy = True
+        file_upload.save()
```
sergei-maertens, line 16:
```suggestion
    TemporaryFileUpload = apps.get_model("submissions", "TemporaryFileUpload")

    TemporaryFileUpload.objects.bulk_update(legacy=True)
```

src/openforms/submissions/models/submission_files.py
```diff
@@ -48,6 +49,20 @@ def select_prune(self, age: timedelta):
 
 class TemporaryFileUpload(DeleteFileFieldFilesMixin, models.Model):
     uuid = models.UUIDField(_("UUID"), unique=True, default=uuid.uuid4)
+    legacy = models.BooleanField(
+        _("legacy"),
+        default=False,
+        help_text=_("Whether the instance is linked to a submission instance."),
+    )
+    # TODO null=True is a transitional state, and should be removed
+    # at some point:
```
sergei-maertens, line 58: put a `DeprecationWarning` in the comment so we can find it when preparing Open Forms 3.0

src/openforms/submissions/migrations/0005_temporaryfileupload_legacy_and_more.py
```diff
@@ -0,0 +1,45 @@
...
+            model_name="temporaryfileupload",
+            constraint=models.CheckConstraint(
+                check=models.Q(
+                    models.Q(("legacy", False), ("submission__isnull", False)),
+                    models.Q(("legacy", True), ("submission__isnull", True)),
+                    _connector="OR",
+                ),
+                name="non_legacy_submission_not_null",
+            ),
+        ),
```
sergei-maertens, line 44: so:

1. legacy is added, default is `False` and all `submission` columns are `NULL`
2. the constraint is added and says that an upload needs to either
     * have legacy `False` and submission `NOT NULL`
     * have legacy `True` and submission `NULL`

that will crash during migration when legacy upload records exist - I think this migration needs to be incorporated in the migration tests too to prevent this from happening

The constraint can only be added after the data is fixed through a data migration

src/openforms/submissions/tests/factories.py
```diff
@@ -326,7 +333,12 @@ def content_type(self) -> str:
 
 class SubmissionFileAttachmentFactory(factory.django.DjangoModelFactory):
     submission_step = factory.SubFactory(SubmissionStepFactory)
-    temporary_file = factory.SubFactory(TemporaryFileUploadFactory)
+    temporary_file = factory.SubFactory(
+        TemporaryFileUploadFactory,
+        submission=factory.LazyAttribute(
+            lambda o: o.factory_parent.submission_step.submission
+        ),
+    )
```
sergei-maertens, line 341: is `factory.SelfAttribute` not an option? I think `factory_parent` is private API

src/openforms/submissions/tests/test_models.py
```diff
@@ -584,6 +586,22 @@ def test_clear_execution_state_without_execution_state(self):
...
+                submission=None,
+                legacy=False,
+            )
+
+    def test_non_legacy_check_constraint(self):
+        with self.assertRaises(IntegrityError):
+            TemporaryFileUploadFactory.create(
+                submission=SubmissionFactory.create(),
+                legacy=True,
+            )
```
sergei-maertens, line 602: I find testing constraints on models with factories a bit tricky, I don't know what's happening behind the scenes and which implicit things are set up. I think I'd rather test model classes directly in those cases.

sergei-maertens (changes requested):

src/openforms/formio/api/serializers.py
Viicos: It was expected to be a `ModelSerializer` because fields were already defined in `Meta`, but I'll switch to `Serializer` and remove the irrelevant meta fields instead.

src/openforms/submissions/migrations/0006_set_legacy_true.py
Viicos: Assuming you meant `.update(legacy=True)`

src/openforms/submissions/tests/factories.py
Viicos: They both work and `factory_parent` is documented as an alternative: https://factoryboy.readthedocs.io/en/stable/reference.html#parents

src/openforms/submissions/tests/test_models.py
Viicos: `TemporaryFileUpload` requires some annoying fields (file fields)..

src/openforms/formio/api/serializers.py
```diff
@@ -25,6 +26,14 @@ class TemporaryFileUploadSerializer(serializers.Serializer):
         use_url=False,
         validators=[MimeTypeValidator(), NoVirusValidator()],
     )
+    submission = serializers.HyperlinkedRelatedField(
+        view_name="api:submission-detail",
+        lookup_field="uuid",
+        queryset=Submission.objects.all(),
```
sergei-maertens, line 32: can we set this to `none` and dynamically based on the request set the queryset?

```suggestion
        queryset=Submission.objects.none(),
```

```py
def get_fields(self):
    fields = super().get_fields()
    session = self.context["request"].session
    fields["submission"].queryset = Submission.objects.filter(uuid__in=session.get(SUBMISSIONS_SESSION_KEY, []))
    return fields
```

That makes it safe by default and only explicitly sets the allow-list for valid requests.

src/openforms/formio/api/serializers.py
```diff
@@ -36,18 +45,18 @@ class TemporaryFileUploadSerializer(serializers.Serializer):
...
             request=request,
         )
+
+    def validate_submission(self, submission: Submission) -> Submission:
+        if submission.uuid not in self.context["reques"].session.get(
+            SUBMISSIONS_SESSION_KEY, []
+        ):
+            raise serializers.ValidationError({"submission": _("Invalid submission")})
+
+        return submission
```
sergei-maertens, line 62: this would not be needed with the suggestion above

src/openforms/formio/rendering/tests/test_vanilla_formio_components.py
```diff
@@ -731,8 +731,8 @@ def test_nested_files_with_duplicate_keys(self):
             form__formstep__form_definition__configuration={"components": components},
         )
 
-        upload_1 = TemporaryFileUploadFactory.create()
-        upload_2 = TemporaryFileUploadFactory.create()
+        upload_1 = TemporaryFileUploadFactory.create(submission=submission)
+        upload_2 = TemporaryFileUploadFactory.create(submission=submission)
```
sergei-maertens, line 735: you can also do the `create_batch` approach here

src/openforms/formio/tests/validation/helpers.py
```diff
@@ -12,13 +14,17 @@
...
+
 def validate_formio_data(
-    component: Component, data: JSONValue
+    component: Component, data: JSONValue, extra_context: ExtraContext = {}
 ) -> tuple[bool, ReturnDict]:
     """
     Dynamically build the serializer, validate it and return the status.
     """
-    context = {"submission": SubmissionFactory.build()}
+    context = {"submission": SubmissionFactory.build(), **extra_context}
```
sergei-maertens, line 27: I think you're making it too complicated, just do:

```py
def validate_formio_data(
    component: Component,
    data: JSONValue,
    submission: Submission | None = None,
) -> tuple[bool, ReturnDict]:
    if submission is None:
        submission = SubmissionFactory.build()
    ...
```

src/openforms/submissions/tests/factories.py
```diff
@@ -312,12 +312,23 @@ class Meta:
...
     content = factory.django.FileField(filename="file.dat", data=b"content")
 
     class Meta:
         model = TemporaryFileUpload
 
+    class Params:
+        is_legacy = factory.Trait(
+            legacy=True,
+            submission=None,
+        )
```
sergei-maertens, line 330: I think this can be removed now that `factory.Maybe` is used?

src/openforms/submissions/tests/test_submission_attachment.py
```diff
@@ -381,9 +383,14 @@ def test_attach_malformed_file(self, resize_mock):
...
-        upload_in_repeating_group_2 = TemporaryFileUploadFactory.create()
-        nested_upload = TemporaryFileUploadFactory.create()
+        submission = SubmissionFactory.create()
+        upload_in_repeating_group_1 = TemporaryFileUploadFactory.create(
+            submission=submission
+        )
+        upload_in_repeating_group_2 = TemporaryFileUploadFactory.create(
+            submission=submission
+        )
+        nested_upload = TemporaryFileUploadFactory.create(submission=submission)
```
sergei-maertens, line 393: could also use `create_batch`

src/openforms/submissions/tests/test_submission_attachment.py
```diff
@@ -1060,9 +1087,13 @@ def test_attach_upload_larger_than_configured_max_size_raises_413(self):
                 "fileMaxSize": "10B",
             },
         ]
+
+        # Using `legacy` is the only way to create an upload without a submission.
+        # It will be set later.
         upload = TemporaryFileUploadFactory.create(
-            file_name="aaa.txt", content__data=b"a" * 20, file_size=20
+            file_name="aaa.txt", content__data=b"a" * 20, file_size=20, legacy=True
         )
```
sergei-maertens, line 1095: why can't we create it with a submission from the get-go?

src/openforms/formio/api/serializers.py
Viicos: I initially went for this but results in the Spectacular API view to fail as it probably accesses the `fields` property (which call `get_fields`), but without any request available.

src/openforms/submissions/tests/test_submission_attachment.py
Viicos: Because the submission is created afterwards with `.from_components(components, data)`, and `data` is created using `upload`

src/openforms/formio/api/serializers.py
sergei-maertens: `self.context["view"].swagger_fake_view`, see https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-get-queryset-depends-on-some-attributes-not-available-at-schema-generation-time

src/openforms/submissions/tests/test_submission_attachment.py
sergei-maertens: you can just call the submission creation before without any data, and then set the submission step data, no? That also better reflects what actually happens with the SDK - the submission must exist before uploads/step data submission can happen.

sergei-maertens (approved):

PR #4470, opened 2024-07-01
Fix type annotations for prefill (plugin) signatures

sergei-maertens:
No issue, but came across it as part of #4246 and decided to open a separate PR.

No typing errors anymore with Pyright in Sublime - though I suspect some call-sites may now be messed up. Those should probably be updated rather than messing with the base classes again.

I'm not particularly pleased yet with the type annotation for `attributes` - I think it'd be nice if we could dump in a constant/`models.TextChoices`, but first let's see how #4266 develops first, as that one has dynamic attributes that are not known upfront.


src/openforms/config/data.py
```diff
@@ -3,7 +3,9 @@
 
 from django.contrib.admin.templatetags.admin_list import _boolean_icon
 
-Action: TypeAlias = tuple[str, str]
+from openforms.typing import _StrOrPromise
```
Viicos, line 6: I think I redefined `StrOrPromise` explicitly, without the `_`

src/openforms/plugins/plugin.py
```diff
@@ -43,7 +46,7 @@ def check_config(self):
         """
         raise NotImplementedError()
 
-    def get_config_actions(self) -> list[tuple[str, str]]:
+    def get_config_actions(self) -> Sequence[Action]:
```
Viicos, line 49: Altough I usually like to use the most precise type for return annotations, it makes sense in this case as the way to say "I only guarantee the `Sequence` protocol to be respected" :)

Viicos (approved):

src/openforms/plugins/plugin.py
sergei-maertens: there's actually a warning about covariance here if you *don't* include the `list[Action]` type annotation, and for how we use it, list/tuple/sequence doesn't make any difference so this was the easiest fix

src/openforms/config/data.py
sergei-maertens: right - why do we have both? :P

Viicos: `_StrOrPromise` is defined as a private alias in the django stubs but preferred to export it in a more explicit way

PR #4471, opened 2024-07-01
Experiment: pyright as type checker

### LINKED ISSUE #4513: Add type checker to CI pipeline
To prevent regressions in type annotations, we need to enforce "known good" typed code and run the type checks in CI.

sergei-maertens:
Closes #4513

Let's see if we can gradually get type checking added... I'm particularly interested in the `include` option :)

Goal: lock in some type-safety easy-wins - clean up some pure python (without too much django) type annotations, and hook them up to CI so that we don't get regressions on that department anymore.

Whatever work can be done in this PR would be nice to have, and we can then squash-rebase and add the resulting commit to the `.git-blame-ignore-revs` file to keep the git history clean.

Future typing improvements can then be added to the config file to prevent regressions, and this way we gradually add type-checking to the project as a whole.


src/openforms/plugins/registry.py
```diff
@@ -1,19 +1,19 @@
...
 if TYPE_CHECKING:
     from .plugin import AbstractBasePlugin  # noqa: F401
 
 
-PluginT_co = TypeVar("PluginT_co", bound="AbstractBasePlugin", covariant=True)
-
-# Ideally, this should be bound to "type[PluginT_co]", but is not possible as of today.
-PluginTypeT = TypeVar("PluginTypeT", bound="type[AbstractBasePlugin]")
+PluginT_co = TypeVar("PluginT_co", bound="AbstractBasePlugin")
+PluginTypeT: TypeAlias = type[PluginT_co]
```
sergei-maertens, line 16: doesn't seem to trip the type checker with sub-classes *so far* (the covariance was definitely a problem though)

sergei-maertens: it was broken, didn't complain for sub-classes inheriting from the wrong base class :)

Viicos: There's a tradeoff between the two solutions, but I personally prefer the existing one (this was changed in https://github.com/open-formulieren/open-forms/pull/3837):

- Existing one: [pyright-play](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDCAhgDYlEBGJApgDRQDi1K1ISAxvQCqLUBqREPSIBnEaxgB9eAmoAoOe3JioAQQoiYIIuxgAhUdQAKJAK5pUALjlRbUBKJEK5J86i6T2YKAF4oPWQEQAAoAIlcLFA8vUPoKMFMUABMfdU1tXQNxCNR6LwA3QSQiFBgfLhBTagBKFzNIgOouX39eILCcqN4uWKh4xJSZagBtNK0dfUNOgF1ahSVHKCzqACVqC3S4YKYWNnZhzuiwWes7KCTqYChJT1ISG%2BDxEmB6RKQARyrJJAvSpGAkKxLFB0vQAPTVKAAWgAfIQ7pQaMMDvV3N1pvRDujgQA6PHORTKERQABiYDAnWCYwyk2yqJQ1VOdgcYgUAGIoAB5ADS1gWKjJYDWGy0W2WwqQm2GgpmjJszMc7KgEs0rFQaGscmA5MkIHWkpgrBagpVouCcwAAtqwLr9aqQqEIHAodbQrV%2BcSALJwGX04K%2BtwMpm2FlOOQcnnWRwSaS8YLegORehDYYJ8my8P%2BAAWkswxKwODwJXwFGo7CIpnEmBQsGomhEeT1RBg6vOpggTvs9OJJSSUD1IrV6FgWeoEEzAHdcABrHvE6gAD1kumofeCSBx1BxUAAklByzWkt4AAZpil%2B6o4kSQaiSCDUGBZsBJYJ4nHVY9zQmLAwgSnUiZlk6OUzlDeYiSWQRTRAMVDGguBhl-WVg3sRU5AoQRbUHEAWl-eDzSVHk4lMfARCfUwSCSFAAHIYGsC0MJALCDVYMInShRi3W-FRvSQv1E1QECFVZZwgA)
- With proposed changes: [pyright-play](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDCAhgDYlEBGJApgDRQDi1K1ISAxvQCqLUBqREACgh7cgGdxUAIIVxMEEXYwAQkXHUACiQCuaVAC4hUE1ATrxIodr2ouAfXZgoAXig8E-QQAoARDf0UBydfegowHRQAExdZeUVlNQ0A1ABKKzELKCTqACVqfXi4byYWNnYAbRSgxzAAXVSjUygo6mAoe0dSEk7vDRJgekikAEcdanskVpQYJGAkVgMoePoAelSoAFoAPkJuyhoKivhPKt1A4Pq6%2BhPqM9sapzq6pYA6d6tRCSkAMTAwareOIKJSqdRac5pJqmcySEQAYigAHkANJGTKSKB-MD5QoKYo5XFIIoVbHVBrQkywyxCRFE%2BSsVBoIxCYD-ewgArEmCsVxY-70-HedJCAACbLAHK5DJAfggcE2Et86QxUgAsnAyZCUN4tQ9GsZTK12vL7BL7BBqDAABZgKJ9agDDY7TAzN4fBFQAByYHwqKMnIAbtRSPZbt4NXrAhsoIjfLdSf9yaEoNQAB6eZTUKKwXgVSNJ7V1KAiAsA7XC16m82Wm124UmREeaipkDgYQiVXZQSA4EJMHJbUG5rUzvfbsgQUgAngqdwCpqEDk4cwiwiCiCKV43luRdz4Wen1%2BlH0dSpjPULNRIyijcgLfc1hyhV35VfLIaxeAqNQw1UtdWEAA)
:shrug:

sergei-maertens: the original one didn't pass the type check, but maybe I was confusing it with another change?

but I personally don't think you should be calling/testing interfaces that exist on a particular plugin that are not part of the base plugin interface - the former is never called by application code, while the latter is, and tests should be testing the black-box interface/public API, not private API/implementation details.

I'll tinker with it more in the off-week though.

sergei-maertens: yeah so I tried reverting it and it just doesn't type check:

```

/home/bbt/code/open-forms/src/openforms/plugins/registry.py
  /home/bbt/code/open-forms/src/openforms/plugins/registry.py:55:31 - error: Argument of type "PluginTypeT@__call__" cannot be assigned to parameter "plugin" of type "PluginT@BaseRegistry" in function "check_plugin"
    Type "PluginTypeT@__call__" is incompatible with type "PluginT@BaseRegistry" (reportArgumentType)
  /home/bbt/code/open-forms/src/openforms/plugins/registry.py:57:13 - error: Argument of type "PluginTypeT@__call__" cannot be assigned to parameter "value" of type "PluginT@BaseRegistry" in function "__setitem__"
    Type "PluginTypeT@__call__" is incompatible with type "PluginT@BaseRegistry" (reportArgumentType)
2 errors, 0 warnings, 0 informations 
```
image

Viicos: Yes this definitely can't be type checked with the current features available.

> but I personally don't think you should be calling/testing interfaces that exist on a particular plugin that are not part of the base plugin interface - the former is never called by application code, while the latter is, and tests should be testing the black-box interface/public API, not private API/implementation details.

While from a typing perspective I really don't like having:

```python
@some_dec
class B(A): pass

reveal_type(B)  # "A"
```

it makes more sense if you are dealing with an ABC and subclasses. So as you wish. Let's just remove the `_co` suffix also :)

sergei-maertens: can't you narrow it with `assert isinstance(...)` at that point, if the exact type of the plugin is relevant?

Generic code does not benefit from auto-completion on the concrete plugin classes, and if you're unit testing the plugins, I think the `isinstance` check would be a small price to pay - in the end the type checker is doing its job and enforcing that you only call the shared/generic API.

I did indeed remove the `_co` suffix :)

.github/workflows/code_quality.yml
```diff
@@ -144,3 +144,30 @@ jobs:
...
+          uv pip install \
+            --system \
+            -r requirements/ci.txt \
+            django-stubs \
+            django-stubs-ext \
+            djangorestframework-stubs \
+            types-lxml \
+            types-pytz \
+            types-pyyaml \
+            types-requests
```
Viicos, line 169: Should we have an extra `.in` file for these instead?

Viicos (approved):

.github/workflows/code_quality.yml
sergei-maertens: yeah, I'll set that up, will make things easier for local checks too

PR #4472, opened 2024-07-01
Adapt prefill to new auth context

### LINKED ISSUE #4246: Capture authentication context data

sergei-maertens:
Closes #4246

**Changes**

* Updated existing tests to mention legacy behaviour
* Updated attribute name for authorizee prefill (which points to the legal subject)
* Added tests for prefill of authorizee data with new format

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/prefill/contrib/stufbg/plugin.py
```diff
@@ -143,7 +143,18 @@ def get_identifier_value(
             return submission.auth_info.value
 
         if identifier_role == IdentifierRoles.authorizee:
-            return submission.auth_info.machtigen.get("identifier_value")
+            legacy_fallback = submission.auth_info.machtigen.get("identifier_value")
```
vaszig, line 146: I see the code is duplicated but I guess we want that in this explicit way.  I would have this in one place to be honest, in contrib for example.

vaszig (approved):

src/openforms/prefill/contrib/stufbg/plugin.py
sergei-maertens: IMO that breaks the isolation of plugins and ties them too much to generic open-forms code. in 3.0 we will probably remove the fallback behaviour and then it's just a similar-but-not-the-same pattern. maybe it could become a method on `AuthInfo` at some point, but the extra checks if it's a BSN/KVK kind of make it plugin-specific what needs to be checked.

In essence this is one-time only code, but because StUF-BG and HaalCentraal BRP plugins are alternatives for the same functionality, this code is duplicated.

PR #4476, opened 2024-07-02
Fix another flaky test when intercepting/spying on API calls

sergei-maertens:
Webkit seems to make requests to some 'srcdoc' somewhere that we have no idea about... Surpressing it should be fine, but logging it to make the problem more visible in CI seems appropriate.


stevenbal (approved):
Not sure if the CI failure is related?

sergei-maertens:
> Not sure if the CI failure is related?

nope, that's another random flaky thing :(

PR #4478, opened 2024-07-02
Keep username/password form/2fa form visible

### LINKED ISSUE #4429: Regular admin login doesn't stay expanded after attempting to log in

latest (pre 2.7)

### Customer reference

_No response_

If admin login via OIDC is enabled, the default Django login is collapsed and can be expanded by clicking `Login with application account`. If there are any errors or extra steps (such as 2FA) the default login is collapsed again, which is annoying

1. Go to `/admin/`
2. Try to log in with incorrect credentials
3. Observe that the login form is collapsed again:
image

1. Go to `/admin/`
2. Log in with 2FA enabled
3. Observe that the token field is collapsed:
image

When entering incorrect credentials or when you have to enter your 2FA token, the form should stay expanded.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
The JS approach to toggle this does not play well with the markup in the library templates, so we need to resort to programmatically clicking the show link. IMO this feature request was rushed and it leads to this kind of fragile code to deal with that fallout from that :/

Closes #4429 

**Changes**

* Added meta-information about login-phase to DOM
* Added JS triggers to toggle visibility depending on meta-information/state of login flow

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/login.js
+++ b/src/openforms/js/components/admin/login.js
@@ -5,11 +5,22 @@ const addAdminLoginExpand = () => {
   const loginForm = document.querySelector('#login-form');
 
   if (defaultLoginToggle && loginForm) {
-    defaultLoginToggle.addEventListener('click', e => {
-      e.preventDefault();
+    const showLoginForm = () => {
       loginForm.classList.toggle('login-form--enabled');
       defaultLoginToggle.classList.toggle('admin-login-option--disabled');
+    };
+
+    // bind click event
+    defaultLoginToggle.addEventListener('click', e => {
+      e.preventDefault();
+      showLoginForm();
     });
+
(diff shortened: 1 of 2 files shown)


stevenbal (commented):
I think if we really want to do this without JS, we could use selectors like this

```css
.errornote ~ #content-main #login-form {
  display: block;
}

.errornote ~ #content-main .admin-login-option--password {
  display: none;
}
```

This should open the login option in case of errors (I'm sure we could do the same for 2FA). @sergei-maertens thoughts?

sergei-maertens:
Discussed during standup, let's at least get the minimal patch in and we can improve on it later.

PR #4481, opened 2024-07-02
[#4267] Fix migration crash

Viicos:


sergei-maertens (approved):

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -86,7 +86,7 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
         label=_("objecttype version"),
         help_text=_("Version of the objecttype in the Objecttypes API"),
     )
-    informatieobjecttype_submission_report = serializers.URLField(
+    informatieobjecttype_submission_report_url = serializers.URLField(
```
sergei-maertens, line 89: I'm reverting this, tests are failing and I don't see why this is added.

Viicos: My mistake

Issue #4418, opened 2024-06-19
Figure out a "transform" mechanism for (AddressNL) component(s) when registering them in the Objects API


sergei-maertens:
Sparked by https://github.com/open-formulieren/open-forms/issues/3993#issuecomment-2178591720

This should assist in transitioning from the deprecated location derivation without breaking (existing) Object Type API contracts.

Currently textfields are manually wired together, so you end up with a bunch of form variables that map to certain Objects API properties, e.g.:

* `postcode` -> `address.postcode`
* `street` -> `address.streetName`

Replacing this with an AddressNL component would instead create:

* ` address` -> `address`, but the nested keys inside may/will be mis-aligned

Because the form designer doesn't have any control over the nested keys, this would require them to update the objecttypes, which creates problems in other applications.

So, I'd propose that next to the mapping of form var -> target path, you can also specify some transformation for the value, so that you can rename the nested keys during registration to make this align properly.

This should in theory also enable us to transition from deprecated -> new in an automated, backwards compatible way, or at least build some tooling to assist in this.

sergei-maertens:
Note that we already do this (in a hardcoded way) to handle attachments and map components.

joeribekker:
Refinement: Victorien can point in the right direction.

Issue #4416, opened 2024-06-19
ZGW & objects API registraton backends: add the "ontvangstdatum" attribute when adding attachments


sergei-maertens:
### Thema / Theme

Plugins

### Omschrijving / Description

User uploads/attachments are now uploaded to the Documenten API for both the objects API and the ZGW API registration plugins.

These uploads do not have the `ontvangstdatum` attribute set, even though we know on which day the attachments were uploaded/sent (the same day the form submission is created).

Looking at the API standard, we should do two things:

* (easy) populate the `EnkelvoudigInformatieObject.ontvangstdatum` field with the date of the submission completion timestamp
* [ ] ON HOLD: (harder) because the above attribute is deprecated, we should check if the Documents API version is one that supports the `Verzending` [resource](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/VNG-Realisatie/gemma-documentregistratiecomponent/1.5.0/src/openapi.yaml#tag/verzendingen) and create such a record. I think that:
    * `betrokkene` should point to the initiator of the zaak
    * `informatieobject` should point to the uploaded document
    * `aardRelatie` needs to be `afzender`
    * `ontvangstDatum` -> same as bullet above
    * `contactPersoon` ???

### Added value / Toegevoegde waarde

* Better information management
* Happy customers that happen to use Rx.Mission (and other ZGW API providers)

sergei-maertens:
Asked some refinement here:  https://github.com/VNG-Realisatie/gemma-zaken/issues/2297

joeribekker:
Johannes Batjes is verantwoordelijk voor dit dat VERZENDING object - vraag hem vooral hoe het gebruikt moet worden.

joeribekker:
Estimate: Let's pick up the ontvangtsdatum property now, and after clarification, we can decide to that part if needed.

sergei-maertens:
Part 2 requires a stakeholder

PR #4489, opened 2024-07-04
Fix docker compose setup for local development

sergei-maertens:
Reported via #4416 and #4418

The docker-compose for objects/objecttypes no longer allowed logging in to the admin because of CSRF issues after adding the nginx setup. The root cause for this is/was that the request `Host` and `Origin` no longer matched.

This is addressed in the nginx config & envvars of the services, and additionally some other improvements are made:

* Fixtures are now automatically loaded when the compose services are brought up
* Documented the commands to use to dump fixtures after making changes
* A superuser is now automatically created
* Updated to the latest available versions of Open Zaak, objecttypes and objects API

As part of the 2.7 release process, I will re-record the VCR cassettes and hopefully nothing broke there :grimacing:


docker/docker-compose.objects-apis.yml
```diff
@@ -20,28 +20,53 @@ services:
...
+      - ./objects-apis/fixtures/objecttypes_api_fixtures.json:/app/fixtures/objecttypes_api_fixtures.json
     depends_on:
       - objects-objecttypes-db
+      - objects-redis
     networks:
       - open-forms-dev
 
   objects-web:
-    image: maykinmedia/objects-api:${OBJECTS_VERSION:-2.3.0}
+    image: maykinmedia/objects-api:${OBJECTS_VERSION:-2.3.2}
```
vaszig, line 48: I don't see the `v2.3.2` in the repo. Can we assume that from somewhere else?

vaszig (approved):

docker/docker-compose.objects-apis.yml
Viicos: It is available here: https://hub.docker.com/r/maykinmedia/objects-api/tags

Viicos (approved):

docker/docker-compose.objects-apis.yml
sergei-maertens: yeah I had to look at docker hub / git tags too, it's not in the changelog

PR #4493, opened 2024-07-04
[#4416] Add ontvangstdatum attribute for attachments in ZGW and ObjectsAPI backends

### LINKED ISSUE #4416: ZGW & objects API registraton backends: add the "ontvangstdatum" attribute when adding attachments

vaszig:
Closes #4416 (first bullet)

- Added `ontvangstdatum` to the data sent to ObjectsAPI and ZGW
- Tested with both registration backends and the date is sent as expected

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/contrib/zgw/clients/documenten.py
+++ b/src/openforms/contrib/zgw/clients/documenten.py
@@ -28,6 +28,7 @@ class DocumentenClient(NLXClient):
         content: ContentFile | BinaryIO,
         status: DocumentStatus,
         filename: str,
+        received_date: str | None = None,
         description: str = "",
         vertrouwelijkheidaanduiding: str = "",
     ):
@@ -46,6 +47,7 @@ class DocumentenClient(NLXClient):
             "inhoud": base64_body,
             "status": status,
             "bestandsnaam": filename,
+            "ontvangstdatum": received_date,
             "beschrijving": description,
             "indicatieGebruiksrecht": False,
             "bestandsomvang": (
--- a/src/openforms/contrib/zgw/service.py
+++ b/src/openforms/contrib/zgw/service.py
@@ -84,6 +84,7 @@ def create_attachment_document(
             status="definitief",
             filename=submission_attachment.get_display_name(),
             description="Bijgevoegd document",
+            received_date=options["ontvangstdatum"],
             vertrouwelijkheidaanduiding=options.get(
                 "doc_vertrouwelijkheidaanduiding", ""
             ),
--- a/src/openforms/registrations/contrib/objects_api/submission_registration.py
+++ b/src/openforms/registrations/contrib/objects_api/submission_registration.py
@@ -31,6 +31,7 @@ from openforms.submissions.models import (
     SubmissionReport,
 )
 from openforms.typing import JSONObject
+from openforms.utils.date import datetime_in_amsterdam
 from openforms.variables.constants import FormVariableSources
 from openforms.variables.service import get_static_variables
 from openforms.variables.utils import get_variables_for_context
@@ -147,6 +148,14 @@ def register_submission_attachment(
             "doc_vertrouwelijkheidaanduiding": "doc_vertrouwelijkheidaanduiding",
         },
     )
+
+    # this helps the type checker in narrowing the type
+    # and expressing that it should not happen that submission.completed_on is not set
+    assert submission.completed_on is not None
+    attachment_options["ontvangstdatum"] = (
+        datetime_in_amsterdam(submission.completed_on).date().isoformat()
+    )
+
     component_overwrites = {
         "doc_vertrouwelijkheidaanduiding": attachment.doc_vertrouwelijkheidaanduiding,
         "titel": attachment.titel,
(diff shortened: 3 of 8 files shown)


src/openforms/contrib/zgw/clients/documenten.py
```diff
@@ -28,6 +28,7 @@ def create_document(
         content: ContentFile | BinaryIO,
         status: DocumentStatus,
         filename: str,
+        received_date: Optional[str] = None,
```
Viicos, line 31:
```suggestion
        received_date: str | None = None,
```

src/openforms/contrib/zgw/service.py
```diff
@@ -84,6 +84,11 @@ def create_attachment_document(
             status="definitief",
             filename=submission_attachment.get_display_name(),
             description="Bijgevoegd document",
+            received_date=(
+                options["ontvangstdatum"].date().isoformat()
+                if options["ontvangstdatum"]
+                else None
+            ),
```
Viicos, line 91: To be consistent with the other `create_document` args, why not do `received_date=options.get("ontvangstdatum")` here, and set the correct date before like so:

```python
if submission.is_completed:
    attachment_options["ontvangstdatum"] = submission.completed_on.date().isoformat()
```

I'm also wondering if this check is necessary, as submissions should be completed when registration plugins are called. I think it's better to assume `submission.completed_on` is set, so that if it isn't it crashes and at least we know something's wrong

vaszig: I agree that during the registration the submission should be completed https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/tasks.py#L166 but I have to investigate why these tests are failing. They are not if I add an if-else statement and set `ontvangstdatum` either to a string or `None`

sergei-maertens: `completed_on` is set when the `/api/v2/submissions/:uuid/_complete` endpoint succeeds, and that kicks off the celery tasks in the background that run the registration.

There is a factory trait to make sure this is set in a consistent way, probably the existing tests were written before this trait existed or they weren't aware of it at the time of writing the test/feature.

Feel free to just write the registration plugin code as `submission.completed_on.isoformat()` - any failing tests need to have their initial state fixed. Registering a submission that isn't completed does not make sense and should crash hard.

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -147,6 +147,7 @@ def register_submission_attachment(
             "doc_vertrouwelijkheidaanduiding": "doc_vertrouwelijkheidaanduiding",
         },
     )
+    attachment_options["ontvangstdatum"] = submission.completed_on.date().isoformat()
```
sergei-maertens, line 150: you'll want to make sure this is localize to Amsterdam time, see `openforms.utils.date.datetime_in_amsterdam`.

Otherwise a submission that was finished at 00:45 AM will be reported as submitted the day before :sparkles: because tiemzones are fun :sparkles:

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -147,6 +148,11 @@ def register_submission_attachment(
             "doc_vertrouwelijkheidaanduiding": "doc_vertrouwelijkheidaanduiding",
         },
     )
+
+    attachment_options["ontvangstdatum"] = (
+        datetime_in_amsterdam(submission.completed_on).date().isoformat()
```
vaszig, line 153: I get an error here because of the definition of function `datetime_in_amsterdam` which accepts `datetime` and the definition of `create_document` which accepts `str|None`. Can or should this be fixed?

sergei-maertens: discussed on slack, `assert` statement will help

vaszig: As discussed in slack this will be fixed in the next commit with an assertion (type checker happy  - developers will know that `submission.completed_on` will be set)

PR #4501, opened 2024-07-08
:bug: Fix how the list of plugins is specified

sergei-maertens:
Should be either a string of (space separated) plugins or an actual JSON/JS array of strings for each plugin, as tinymce tries to load the plugin files on-demand.


vaszig (approved):

sergei-maertens:
Needs an extra check on our deployed test env to rule out unexpected breakage

PR #4507, opened 2024-07-08
Feature/4267 api endpoints

Viicos:
Partly closes #4267

**Changes**

Edited the IOT endpoint to filter by catalog. Made sure no duplicates are returned.
Added a catalogus endpoint.
Rewritten tests to use VCR.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/registrations/api/urls.py
```diff
@@ -9,6 +14,11 @@
         AllAttributesListView.as_view(),
         name="registrations-attribute-list",
     ),
+    path(
+        "catalogi",
+        CatalogiListView.as_view(),
+        name="catalogi-list",
+    ),
```
sergei-maertens, line 21: can you move this down to the plugin urlpatterns, and also move the `informatieobjecttypen` entry?

These are endpoints that are tightly coupled to particular registration plugins and they fit the TODO that's present already.

src/openforms/registrations/api/views.py
```diff
@@ -56,6 +60,35 @@ def get_objects(self):
...
+
+    def get_objects(self):
+        filter_serializer = APIGroupQueryParamsSerializer(
+            data=self.request.query_params
+        )
+        filter_serializer.is_valid(raise_exception=True)
+
+        client = filter_serializer.get_ztc_client()
+        if not client:
+            return []
```
sergei-maertens, line 86: please address the missing test coverage 

Some more questions about this flow:

1. don't we enforce that either a ZGW API group or objects API group needs to be provided?
2. if the above is "no", what's stopping us from doing so?

It would be nicer if this code can just assume that the catalogi client will never be `None`. IMO this endpoint doesn't make sense to be called without a group/catalogi API reference and an HTTP 400 is more appropriate.

src/openforms/registrations/api/views.py
```diff
@@ -86,13 +123,48 @@ def get_objects(self):
...
 
-        iotypen_data = client.get_all_informatieobjecttypen()
-        iotypen = [
-            {
-                "informatieobjecttype": factory(InformatieObjectType, iotype),
-                "catalogus": factory(Catalogus, catalogus_mapping[iotype["catalogus"]]),
-            }
-            for iotype in iotypen_data
-        ]
+        if "catalogus_domein" in filter_serializer.validated_data:
```
sergei-maertens, line 126: we get back a `url` already when we retrieve a list of catalogi - can we not just use the catalogus URL directly as a filter parameter rather than having to pass the domein + rsin?

We'll need to store those values in the options, but for these filter endpoints, we can directly forward the URL query param, since these don't need to be stored, right?

src/openforms/registrations/api/views.py
```diff
@@ -86,13 +123,48 @@ def get_objects(self):
...
+            )
+
+            if matching_catalogus is None:
+                return []
+
+            iotypen_data = client.get_all_informatieobjecttypen(
+                catalogus=matching_catalogus
+            )
+        else:
+            iotypen_data = client.get_all_informatieobjecttypen()
```
sergei-maertens, line 147: can you add a comment to describe in which situations/code-paths we *don't* filter based on catalogus yet? That way we have a list of things to still improve to clean up this branching.

src/openforms/registrations/api/views.py
```diff
@@ -86,13 +123,48 @@ def get_objects(self):
...
+                existing_iotype["informatieobjecttype"].omschrijving == iotype["omschrijving"]
+                and existing_iotype["informatieobjecttype"].catalogus == iotype["catalogus"]
+                for existing_iotype in iotypen
+            )
+            # fmt: on
+
+            if not exists:
+                iotypen.append(
+                    {
+                        "informatieobjecttype": factory(InformatieObjectType, iotype),
```
sergei-maertens, line 163: I think you want to use a different data model for `InformatieObjectType` since now this still includes a URL and version data, and we're basically only interested in the combination of `(domein, rsin, iot_omschrijving)` tuples, so let's just make the data non-confusing and discard what isn't relevant.

Probably that way you can also use a nicer way to build the collection of unique document types, maybe a set of tuples and then you sort it based on catalogus.rsin / catalogus.domein / iot.omschrijving for a stable ordering.

src/openforms/registrations/tests/test_views.py
```diff
@@ -170,19 +95,101 @@ def test_missing_objects_api_group(self, m):
...
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_retrieve_filter_by_catalogus(self):
+        user = StaffUserFactory.create()
+        url = furl(reverse("api:iotypen-list"))
+        url.args["objects_api_group"] = self.objects_api_group.pk
+        url.args["registration_backend"] = "objects_api"
+        url.args["catalogus_domein"] = "TEST"
+        url.args["catalogus_rsin"] = "000000000"
```
sergei-maertens, line 132: please use the simple django API where possible, `furl` should only be used where we have to process incoming URL strings ourselves or when doing complex URL mutations. The much simpler equivalent is:

```py
response = self.client.get(
    reverse(""api:iotypen-list"),
    {
        "objects_api_group": self.objects_api_group.pk,
        "registration_backend": "objects_api",
        "catalogus_domein": "TEST",
        "catalogus_rsin": "000000000",
    },
)
```

src/openforms/registrations/tests/test_views.py
```diff
@@ -1,119 +1,46 @@
...
+)
 from openforms.registrations.contrib.zgw_apis.tests.factories import (
     ZGWApiGroupConfigFactory,
 )
+from openforms.utils.tests.vcr import OFVCRMixin
+
+
+class GetInformatieObjecttypesViewTests(OFVCRMixin, APITestCase):
 
+    VCR_TEST_FILES = Path(__file__).parent / "files"
```
sergei-maertens, line 22: please define that as a module constant like in the rest of the tests

sergei-maertens (changes requested):

src/openforms/registrations/api/urls.py
Viicos: These two endpoints are used by both the ZGW and Objects API registration plugins, that's why I put them here.

src/openforms/registrations/api/views.py
Viicos: Hopefully this won't complicate things in the frontend

src/openforms/registrations/api/views.py
Viicos: I think https://github.com/open-formulieren/open-forms/pull/4507#discussion_r1677592130 simplifies things already

Viicos:
I think the E2E tests should be fixed once this and the frontend PR are merged, will make things easier

Viicos:
A commit will update the formio-builder to take into account the new structure of IOTs

src/openforms/registrations/api/serializers.py
```diff
@@ -38,21 +38,14 @@ class RegistrationAttributeSerializer(serializers.Serializer):
...
 
 
-class InformatieObjectTypeChoiceSerializer(serializers.Serializer):
-    informatieobjecttype = InformatieObjectTypeSerializer()
-    catalogus = CatalogusDomainSerializer()
+class InformatieObjectTypeSerializer(serializers.Serializer):
+    catalogus_domein = serializers.CharField(label=_("catalogus domein"))
+    catalogus_rsin = serializers.CharField(label=_("catalogus rsin"))
+    url = serializers.URLField(label=_("url"))
+    omschrijving = serializers.CharField(label=_("omschrijving"))
```
sergei-maertens, line 51: discussed in meeting - probably best to move this to `openforms.contrib.zgw`, and subclass the `Catalogus` model to add an extra field for `naam`

src/openforms/registrations/api/views.py
```diff
@@ -56,43 +59,77 @@ def get_objects(self):
...
 
 
-@extend_schema_view(
-    get=extend_schema(
-        summary=_("List available InformatieObjectTypen"),
-        parameters=[ListInformatieObjectTypenQueryParamsSerializer],
-    ),
-)
-class InformatieObjectTypenListView(ListMixin, APIView):
+class BaseCatalogiListView(ListMixin, APIView):
```
sergei-maertens, line 62: discussed in meeting - probably best to move this to `openforms.contrib.zgw`

src/openforms/registrations/api/views.py
```diff
@@ -56,43 +59,77 @@ def get_objects(self):
...
+            data=self.request.query_params,
+        )
+        filter_serializer.is_valid(raise_exception=True)
+
+        client = filter_serializer.get_ztc_client()
+        catalogus_data = client.get_all_catalogi()
+        return factory(Catalogus, catalogus_data)
+
+
+class BaseInformatieObjectTypenListView(ListMixin, APIView):
```
sergei-maertens, line 79: discussed in meeting - probably best to move this to `openforms.contrib.zgw`

src/openforms/registrations/api/views.py
```diff
@@ -56,43 +59,77 @@ def get_objects(self):
...
+        iotypen = []
+
+        for iotype in iotypen_data:
+            # fmt: off
+            exists = any(
+                existing_iotype["catalogus_domein"] == catalogus_mapping[iotype["catalogus"]]["domein"]
+                and existing_iotype["catalogus_rsin"] == catalogus_mapping[iotype["catalogus"]]["rsin"]
+                and existing_iotype["omschrijving"] == iotype["omschrijving"]
+                for existing_iotype in iotypen
+            )
```
sergei-maertens, line 117: discussed in meeting, we can probably simplify this check by considering the `(catalogus_url, omschrijving)` unique tuples and only expand the domein/RSIN information after all the data was gathered/de-duplicated.

src/openforms/registrations/contrib/zgw_apis/api/filters.py
```diff
@@ -0,0 +1,33 @@
...
+from ..client import get_catalogi_client
+from ..models import ZGWApiGroupConfig
+
+
+class APIGroupQueryParamsSerializer(BaseAPIGroupQueryParamsSerializer):
+    zgw_api_group = PrimaryKeyRelatedAsChoicesField(
+        queryset=ZGWApiGroupConfig.objects.exclude(drc_service=None),
+        help_text=_(
+            "The primary key of the Objects API group to use. The informatieobjecttypen from the Catalogi API "
+            "in this group will be returned."
```
sergei-maertens, line 18: some copy-paste leftovers, s/objects/zgw

sergei-maertens (changes requested):

sergei-maertens (approved):
Let's get this merged :tada:

PR #4521, opened 2024-07-12
[#4397] Add query parameter to pass reference to existing object in the Objects API

### LINKED ISSUE #4397: Add query parameter to pass reference to existing object and send this to the backend on submissionCreate
This reference will be used in https://github.com/open-formulieren/open-forms/issues/4395, https://github.com/open-formulieren/open-forms/issues/4396 and https://github.com/open-formulieren/open-forms/issues/4398

This reference might be stored as a form variable or on the `Submission` itself, I'm not sure yet which is more appropriate

vaszig:
Closes #4397

**Changes**

- Added `object_reference` to the `Submission` model in order to save the reference we get from the frontend.
  This will be used to retrieve further information for the real object

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/submissions/api/serializers.py
```diff
@@ -200,6 +210,13 @@ def to_representation(self, instance):
         check_submission_logic(instance, unsaved_data=self.context.get("unsaved_data"))
         return super().to_representation(instance)
 
+    def validate_object_reference(self, value: str) -> str:
+        re_match = re.search(r"^[a-zA-Z0-9\-]+$", value)
```
Viicos, line 214: This is almost the same regex as the `SlugField` one, so I think we should either:
- use `SlugField` at the model level, and just add `object_reference` in `Meta.fields` here (no need to define it explicitly)
- use `CharField` at the model level with a `RegexValidator`, and use a `RegexField` here in the serializer (I think DRF doesn't reuse model fields validators).

src/openforms/submissions/models/submission.py
```diff
@@ -256,6 +256,14 @@ class Submission(models.Model):
             "Indicates whether sensitive data (if there was any) has been removed from this submission."
         ),
     )
+    object_reference = models.CharField(
+        _("object reference"),
+        blank=True,
+        help_text=_(
+            "The object reference-id which will be used to modify the product in the registration backend."
```
Viicos, line 263:
```suggestion
            "The object reference-id which will be used to modify the product in the registration backend. "
```

Viicos (changes requested):

src/openforms/submissions/api/serializers.py
vaszig: Discussed this with Sergei as well on slack and decided to remove this for now since we do not know exactly the structure of the query parameter value. It will be a string so the `CharField` will do the needed validation.

src/openforms/submissions/api/serializers.py
```diff
@@ -155,6 +155,14 @@ class SubmissionSerializer(serializers.HyperlinkedModelSerializer):
         write_only=True,
         default=False,  # type: ignore
     )
+    object_reference = serializers.CharField(
+        label=_("Object reference"),
+        required=False,
+        help_text=_(
+            "The object reference-id which will be used to modify the product in the registration backend."
```
Viicos, line 162:
```suggestion
            "The object reference-id which will be used to modify the product in the registration backend. "
```

Viicos: But I think adding `object_reference` to `Meta.fields` is enough?

vaszig: Yes you are right, I am fixing that now.

src/openforms/submissions/admin.py
```diff
@@ -265,6 +266,7 @@ class SubmissionAdmin(admin.ModelAdmin):
                     "finalised_registration_backend_key",
                     "pre_registration_completed",
                     "registration_result",
+                    "object_reference",
```
sergei-maertens, line 269: this is not part of the registration process, I'd put it under `Misc` instead

src/openforms/submissions/admin.py
```diff
@@ -230,6 +230,7 @@ class SubmissionAdmin(admin.ModelAdmin):
         "payment_complete_confirmation_email_sent",
         "get_price",
         "get_payment_required",
+        "object_reference",
```
sergei-maertens, line 233: let's leave it editable instead of making it read-only, all these readonly fields are making problem solving hard for Laurens and me

src/openforms/submissions/models/submission.py
```diff
@@ -256,6 +256,14 @@ class Submission(models.Model):
...
         ),
     )
+    object_reference = models.CharField(
+        _("object reference"),
+        blank=True,
+        help_text=_(
+            "The object reference-id which will be used to modify the product in the registration backend. "
+            "(For example the object reference for the Object in the Objects API)"
+        ),
+    )
```
sergei-maertens, line 266:
```suggestion
    initial_data_reference = models.CharField(
        _("initial data reference"),
        blank=True,
        help_text=_(
            "An identifier that can be passed as a querystring when the form is started. Initial form field values are pre-populated from the retrieved data. During registration, the object may be updated again (or a new record may be created). This can be an object reference in the Objects API, for example."
        ),
    )
```

* I'd make the scope of this field clearer by naming it `initial_data_reference` - there will be other cases in the future where data is retrieved somewhere else from than the objects API (e.g. an existing zaak/case).
* generalized the help text and meaning - it's not guaranteed that we will update the same object again, you could also load data from the objects API but then write the registration to another place.

src/openforms/submissions/tests/test_start_submission.py
```diff
@@ -231,3 +231,14 @@ def test_start_submission_with_prefill(self, mock_logevent):
...
+    def test_start_submission_with_object_reference(self):
+        body = {
+            "form": f"http://testserver.com{self.form_url}",
+            "formUrl": "http://testserver.com/my-form",
+            "objectReference": "of-or-3452fre3",
+        }
+
+        response = self.client.post(self.endpoint, body)
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
sergei-maertens, line 244: I also want to see an assertion that the database record has the right value

sergei-maertens (changes requested):

src/openforms/submissions/models/submission.py
vaszig: Yes, makes sense I will update it. This means I should change it everywhere right?I mean serializers and SDK

sergei-maertens: yep!

src/openforms/submissions/tests/test_start_submission.py
```diff
@@ -231,3 +231,19 @@ def test_start_submission_with_prefill(self, mock_logevent):
...
+        }
+
+        response = self.client.post(self.endpoint, body)
+
+        submission = Submission.objects.get()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(
+            submission.initial_data_reference, body["initialDataReference"]
+        )
```
sergei-maertens, line 249: in tests, please make assertions with the literal values instead of referring to variables in tests, that makes it clear what is being compared to what.

sergei-maertens (approved):

Issue #4330, opened 2024-05-27
Exporting a form from 'test' and importing on 'acceptance' leads to invalid URLs to informatieobjecttypes


alextreme:
### Thema / Theme

Admin

### Omschrijving / Description

Discussed with Laurens. When exporting a form in OF and importing it in another environment, the URL references to informatieobjecttypes still point to the test environment. This leads to forms that don't work after importing them.

(I'm limiting the scope of this issue on purpose to only IOTs, as certain OF DH forms have dozens of IOT references)

There are a number of ways how we can improve this, some suggestions:
- Ensuring that when importing a form, that the admin can see and modify the imported IOT URLs
- Ensuring that there's a page to check (GET) the IOTs configured for the form and show the URLs
- Ability to configure test/acc/prod configuration within the form (suggestion Laurens)

This has been discussed between Laurens and Joeri earlier, I'm attempting to broaden the discussion and see what's possible with a more limited scope

### Tasks

* [ ] Make sure import/export uses natural keys (problem identified: we store primary keys for object api/zgw api groups, service fetch)
* [ ] MORE

sergei-maertens:
we should store the omschrijving/identificatie instead of the full URLs, so that the latest version of that object can be resolved dynamically. The same approach is planned for zaaktype too.

joeribekker:
Refinement: I think this issue will be part of workpackage that includes NEVER to save URLs but only the unique attributes, and lookup the URL when needed. This workpackage will be incoming.. in 2.8 I think.

joeribekker:
Related: #4267

joeribekker:
This ticket is about import and export. The work Viicos is doing is essential for this.

sergei-maertens:
> (I'm limiting the scope of this issue on purpose to only IOTs, as certain OF DH forms have dozens of IOT references)

This particular issue is gone as of 4.0.0 - those URLs are no longer stored on the components or at all - we now use indirect references at the registration plugin level.

There are broader import/export issues touched upon here as well, but Robin has done substantial work to improve that situation and that's under review, likely to be included in 4.1.

PR #4522, opened 2024-07-12
[#4516] Fix import of forms with Objects API registration backends without any group set

### LINKED ISSUE #4516: Import of legacy Objects API registration forms is not possible anymore
Version: 2.7.0

importing form with v1 Object Api registration causes error: "Er is iets foutgegaan bij het importeren van formulier: {'registration_backends': [{'options': {'objects_api_group': [ErrorDetail(string='Dit veld is vereist.', code='required')], 'objecttype': [ErrorDetail(string='Must be a valid UUID.', code='invalid')]}}"

I have an export of such a form provided by Laurens.

Culprit seems to be 4acd2f4

Viicos:
Closes #4516

- Impact on features

  - [ ] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/forms/tests/test_import_export.py
+++ b/src/openforms/forms/tests/test_import_export.py
@@ -1401,10 +1401,8 @@ class ImportObjectsAPITests(TempdirMixin, OFVCRMixin, TestCase):
 
     VCR_TEST_FILES = PATH / "files"
 
-    def setUp(self):
-        super().setUp()
-
-        self.objects_api_group = ObjectsAPIGroupConfig.objects.create(
+    def _create_objects_api_group(self):
+        return ObjectsAPIGroupConfig.objects.create(
             objecttypes_service=ServiceFactory.create(
                 api_root="http://localhost:8001/api/v2/",
                 api_type=APITypes.orc,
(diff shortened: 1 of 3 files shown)


src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -191,7 +203,35 @@ def to_internal_value(self, data: Any) -> Any:
...
+            if existing_group is None:
+                raise serializers.ValidationError(
+                    {
+                        "objects_api_group": _(
+                            "You must create a valid Objects API Group config (with the necessary services) before importing."
+                        )
+                    }
+                )
+            else:
+                attrs["objects_api_group"] = existing_group
```
sergei-maertens, line 230: if there are multiple candidate groups, we should grab one that matches the provided objecttype URL if possible, now you can end up in a situation that you pick "the wrong" group.

Keep in mind this will especially happen on environments like DH where they have old exports that are being imported into an environment that has (an) existing group(s)

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1401,10 +1401,8 @@ class ImportObjectsAPITests(TempdirMixin, OFVCRMixin, TestCase):
 
     VCR_TEST_FILES = PATH / "files"
 
-    def setUp(self):
-        super().setUp()
-
-        self.objects_api_group = ObjectsAPIGroupConfig.objects.create(
+    def _create_objects_api_group(self):
+        return ObjectsAPIGroupConfig.objects.create(
```
sergei-maertens, line 1405: wouldn't it make more sense to define a factory with a trait for all this boilerplate? these helper methods in tests are quite annoying to track down/context switch to then infer what they're setting up

something like `ObjectsAPIGroupFactory.create(for_test_docker_compose=True)`

sergei-maertens (changes requested):

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -183,54 +183,68 @@ def get_fields(self):
 
         # If in an import context, we don't want to error on missing groups.
         # Instead, we will try to set one if possible in the `validate` method.
+        # We also change the `objecttype` field to be any string, as it could be an URL.
```
Viicos, line 186: Not ideal, but I can't handle this in `to_internal_value` anymore as I need the `objecttype` to be an URL so that I can select the correct group based on it (and `to_internal_value` runs before `validate`)

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1401,44 +1401,117 @@ class ImportObjectsAPITests(TempdirMixin, OFVCRMixin, TestCase):
...
-                client_id="test_client_id",
-                secret="test_secret_key",
-                auth_type=AuthTypes.zgw,
-            ),
+        with zipfile.ZipFile(self.filepath, "w") as zip_file:
+            for name, data in resources.items():
+                zip_file.writestr(f"{name}.json", json.dumps(data))
+
+        with self.assertRaises(CommandError):
+            call_command("import", import_file=self.filepath)
```
sergei-maertens, line 1434: I would like some more assertions about what exactly the error is, this can happen for a dozen different reasons. The only assertion expressed here now is "importing does not work", which is dangerously close to reported bug in the first place.

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1401,44 +1401,117 @@ class ImportObjectsAPITests(TempdirMixin, OFVCRMixin, TestCase):
...
+
+        # no matching groups:
+        ObjectsAPIGroupConfigFactory.create_batch(2)
+
+        with zipfile.ZipFile(self.filepath, "w") as zip_file:
+            for name, data in resources.items():
+                zip_file.writestr(f"{name}.json", json.dumps(data))
+
+        with self.assertRaises(CommandError):
+            call_command("import", import_file=self.filepath)
```
sergei-maertens, line 1469: please be more specific

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1401,44 +1401,117 @@ class ImportObjectsAPITests(TempdirMixin, OFVCRMixin, TestCase):
...
+                                "objecttype_version": 1,
+                            },
+                        }
+                    ],
+                }
+            ]
+        }
+
+        # no matching groups:
+        ObjectsAPIGroupConfigFactory.create_batch(2)
```
sergei-maertens, line 1462: please set up groups with explicit values that don't match

at some point maybe someone can set a factory default that happens to match and then this test will start failing and it will not be obvious.

When using factories, the fields/values that you rely on in your test should always be specified explicitly.

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1502,7 +1577,9 @@ def test_import_form_with_objecttype_uuid_objects_api_registration_backend(self)
                             "name": "Test backend",
                             "backend": "objects_api",
                             "options": {
-                                "objects_api_group": self.objects_api_group.pk,
+                                "objects_api_group": ObjectsAPIGroupConfigFactory.create(
+                                    for_test_docker_compose=True
+                                ).pk,
```
sergei-maertens, line 1582: I thought we had an issue somewhere for the primary keys in export artifacts, but I can't seem to find it?

src/openforms/registrations/contrib/objects_api/config.py
sergei-maertens: it's documented, that's the important part :)

src/openforms/forms/tests/test_import_export.py
Viicos: iirc it was added as a checkbox in this issue: https://github.com/open-formulieren/open-forms/issues/4330

sergei-maertens (approved):
Will merge and backport tomorrow!

PR #4526, opened 2024-07-15
Remove storing upload IDs in the session

### LINKED ISSUE #4492: Remove storing file upload IDs in the session
This is no longer necessary since d959ff5b22, as having the submission ID in the session is now sufficient.

This affects:

* `openforms.submissions.api.views.TemporaryFileView`
* `openforms.formio.api.views.TemporaryFileUploadView`

`OwnsTemporaryUploadPermission` needs to be updated to check that the `file.submission` is in the session

sergei-maertens:
Closes #4492

**Changes**

This is no longer necessary now that we relate the temporary upload to a submission (session) - we can use the submission authorization checks for upload ownership checks.

This removes a source of race conditions, as we have one less array of UUIDs to keep track of in the session, and especially file uploads were susceptible to this when multiple uploads are performed in parallel.

Additionally, it turns out there was more dead code in the permissions (`filter_queryset`) that was never called, and coverage reports also show it never was, so I deleted it too.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/api/permissions.py
+++ b/src/openforms/submissions/api/permissions.py
@@ -6,9 +6,9 @@ from rest_framework.views import APIView
 
 from openforms.api.permissions import TimestampedTokenPermission
 
-from ..constants import SUBMISSIONS_SESSION_KEY, UPLOADS_SESSION_KEY
+from ..constants import SUBMISSIONS_SESSION_KEY
 from ..form_logic import check_submission_logic
-from ..models import SubmissionStep
+from ..models import SubmissionStep, TemporaryFileUpload
 from ..tokens import (
     submission_report_token_generator,
     submission_status_token_generator,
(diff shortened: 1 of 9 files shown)


src/openforms/submissions/api/permissions.py
```diff
@@ -81,30 +81,16 @@ def filter_queryset(self, request: Request, view: APIView, queryset):
...
-    def filter_queryset(self, request: Request, view: APIView, queryset):
-        active_uploads = request.session.get(UPLOADS_SESSION_KEY)
-        if not active_uploads:
-            return queryset.none()
-        return queryset.filter(uuid__in=active_uploads)
+    def has_object_permission(
+        self, request: Request, view: APIView, obj: TemporaryFileUpload
+    ) -> bool:
+        submission_uuid = str(obj.submission.uuid)
+        return owns_submission(request, submission_uuid)
```
Viicos, line 93: I'm wondering if `super().has_object_permission` should be called just for safety? Currently `AnyActiveSubmissionPermission` doesn't override it, but it might in the future?

Viicos (approved):

src/openforms/submissions/api/permissions.py
sergei-maertens: that doesn't make sense? the scope of the super class is that you are allowed to proceed as soon as you have started any submission at all, there are no object level permissions in play there. And, there is no guarantee that IF they were to be added, the objects will share a common interface, they can be completely different models with a completely different relation to a submission.

PR #4529, opened 2024-07-16
Initiate download of submission report PDF

### LINKED ISSUE #4488: Force downloading of PDF from confirmation page, instead of letting the browser open.
### Thema / Theme

Frontend

Browsers are very eager to open pdf's for their users, but this behavior does not really fit with the kinds of PDF's that OF generates, which are definitely more of the kind you would want to actually save on your disk.

Since the user loses the ability to download the pdf again after the session is gone, the 
"Download PDF" should force a actual download of the PDF via the "download" attribute which seems to be supported by all [modern browsers.](https://caniuse.com/mdn-html_elements_a_download)

- [x] Set up proper URL without exposing API endpoint details
- [x] Add Content-Disposition headers to force download

sergei-maertens:
Closes #4488

**Changes**

* Refactored to use the PrivateMediaView generic CBV
* URL moved to non-API url
* Deleted the API-related code that is now obsolete

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/views.py
+++ b/src/openforms/submissions/views.py
@@ -1,11 +1,13 @@
 import logging
 import uuid
+from pathlib import Path
 
 from django.contrib.auth.hashers import check_password as check_salted_hash
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.core.exceptions import PermissionDenied
 from django.http import HttpRequest
 from django.shortcuts import get_object_or_404
+from django.utils import timezone
 from django.utils.crypto import constant_time_compare
 from django.utils.translation import gettext_lazy as _
 from django.views.generic import FormView, RedirectView
@@ -28,9 +30,9 @@ from openforms.tokens import BaseTokenGenerator
 from .constants import RegistrationStatuses
 from .exceptions import FormDeactivated, FormMaintenance
 from .forms import SearchSubmissionForCosignForm
-from .models import Submission, SubmissionFileAttachment
+from .models import Submission, SubmissionFileAttachment, SubmissionReport
 from .signals import submission_resumed
-from .tokens import submission_resume_token_generator
+from .tokens import submission_report_token_generator, submission_resume_token_generator
 from .utils import add_submmission_to_session, check_form_status
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


src/openforms/submissions/views.py
```diff
@@ -192,6 +194,46 @@ def get_form_resume_url(self, submission: Submission) -> str:
         )
 
 
+class DownloadSubmissionReportView(PrivateMediaView):
+    """
+    Download the PDF report containing the submission data.
+
+    This URL requires a token which is tied to the submission from the session. The
+    token automatically expired after
```
vaszig, line 202:
```suggestion
    token automatically expires after
```

src/openforms/submissions/views.py
```diff
@@ -192,6 +194,46 @@ def get_form_resume_url(self, submission: Submission) -> str:
...
+            "mimetype": "application/pdf",
+        }
+        return opts
+
+    def get(self, request: HttpRequest, *args, **kwargs):
+        self.object = self.get_object()
+
+        response = super().get(request, *args, **kwargs)
+
+        self.object.last_accessed = timezone.now()
```
vaszig, line 232: Do we need local time here, like `datetime_in_amsterdam` ?And basically is there a standard for when we need the local time?

vaszig (approved):
Only one small typo.

src/openforms/submissions/views.py
Viicos: I think the difference between using `timezone.now` vs `datetime_in_amsterdam` is whether you want the datetime displayed (i.e. string formatted or not). `last_accessed` is a db field and they are stored with UTC timezones so it's good to use `timezone.now()`

Sergei might confirm

src/openforms/submissions/views.py
```diff
@@ -192,6 +194,46 @@ def get_form_resume_url(self, submission: Submission) -> str:
...
+    def has_permission(self):
+        report: SubmissionReport = self.get_object()
+        token = self.kwargs["token"]
+        return submission_report_token_generator.check_token(report, token)
+
+    def get_sendfile_opts(self) -> dict:
+        path = Path(self.object.content.name)
+        opts = {
+            "attachment": True,
+            "attachment_filename": path.name,
```
Viicos, line 222: What is the reason to go through `Path(self.object.content.name).name` and not just `self.object.content.name`?

Viicos (approved):
I guess the SDK needs to be updated with the new API endpoint? Or the URL is sent by the backend?

src/openforms/submissions/views.py
sergei-maertens: no, django converts everything to UTC anyway when it's being saved in the DB. The `datetime_in_amsterdam` is only needed in edge cases:

* presentation to the end-use, usually this is handled already in the django template layer automatically
* when interfacing with other systems that don't use ISO-8601 datetimes and expect naive, local times

So usually these are external factors outside of our control.

sergei-maertens:
> I guess the SDK needs to be updated with the new API endpoint? Or the URL is sent by the backend?

URL is sent by the backend, no updates needed. HATEOAS baby :tada:

src/openforms/submissions/views.py
sergei-maertens: the former only returns `[basename].[extension]`, the latter returns `some/subpath/from/upload/to/[basename].[extension]`

PR #4538, opened 2024-07-18
[#4395] Add flag for updating an object to the ObjectsAPI registration

### LINKED ISSUE #4395: Add a flag on Objects API registration to update existing objects instead of creating new ones

vaszig:
Closes #4395

- Added `update_existing_object` flag to the ObjectsAPI (both versions). This will be used in combination with `submission.initial_data_reference` in order to update an existing object and not recreate it.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/contrib/objects_api/clients/objects.py
+++ b/src/openforms/contrib/objects_api/clients/objects.py
@@ -15,3 +15,17 @@ class ObjectsClient(NLXClient):
         response.raise_for_status()
 
         return response.json()
+
+    def update_object(
+        self, objecttype_url: str, record_data: dict, initial_data_reference: str
+    ) -> dict:
+        endpoint = f"objects/{initial_data_reference}"
+        json = {
+            "type": objecttype_url,
+            "record": record_data,
+        }
+
+        response = self.patch(endpoint, json=json, headers=CRS_HEADERS)
+        response.raise_for_status()
+
+        return response.json()
--- a/src/openforms/contrib/objects_api/clients/objects.py
+++ b/src/openforms/contrib/objects_api/clients/objects.py
@@ -16,12 +16,17 @@ class ObjectsClient(NLXClient):
 
         return response.json()
 
-    def update_object(
-        self, objecttype_url: str, record_data: dict, initial_data_reference: str
-    ) -> dict:
-        endpoint = f"objects/{initial_data_reference}"
+    def get_object(self, object_uuid: str) -> dict:
+        endpoint = f"objects/{object_uuid}"
+
+        response = self.get(endpoint, headers=CRS_HEADERS)
+        response.raise_for_status()
+
+        return response.json()
+
+    def update_object(self, record_data: dict, object_uuid: str) -> dict:
+        endpoint = f"objects/{object_uuid}"
         json = {
-            "type": objecttype_url,
             "record": record_data,
         }
 
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/V2ConfigFields.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/V2ConfigFields.js
@@ -23,6 +23,7 @@ const V2ConfigFields = ({index, name, schema, formData, onFieldChange, onChange}
     objecttype = '',
     objecttypeVersion = '',
     informatieobjecttypeSubmissionReport = '',
+    updateExistingObject = false,
     uploadSubmissionCsv = false,
     informatieobjecttypeSubmissionCsv = '',
     informatieobjecttypeAttachment = '',
@@ -167,6 +168,31 @@ const V2ConfigFields = ({index, name, schema, formData, onFieldChange, onChange}
         />
       </CustomFieldTemplate>
       <CustomFieldTemplate
+        id="root_updateExistingObject"
+        label={intl.formatMessage({
+          defaultMessage: 'Update existing object',
+          description: 'Objects API registration options "Update existing object" label',
+        })}
+        rawDescription={intl.formatMessage({
+          defaultMessage:
+            'Indicates whether the (existing) object should be updated and not created',
+          description: 'Objects API registration options "Upload submission CSV" description',
+        })}
+        rawErrors={getFieldErrors(name, index, validationErrors, 'updateExistingObject')}
+        errors={buildErrorsComponent('updateExistingObject')}
+        displayLabel
+      >
+        <Checkbox
+          id="root_updateExistingObject"
+          name="updateExistingObject"
+          value={updateExistingObject}
+          checked={updateExistingObject}
+          onChange={e =>
+            onFieldChange({target: {name: 'updateExistingObject', value: !updateExistingObject}})
+          }
+        />
+      </CustomFieldTemplate>
+      <CustomFieldTemplate
         id="root_informatieobjecttypeSubmissionReport"
         label={intl.formatMessage({
           defaultMessage: 'Submission report PDF informatieobjecttype',
@@ -306,6 +332,7 @@ V2ConfigFields.propTypes = {
     objecttype: PropTypes.string,
     objecttypeVersion: PropTypes.string,
     informatieobjecttypeSubmissionReport: PropTypes.string,
+    updateExistingObject: PropTypes.bool,
     uploadSubmissionCsv: PropTypes.bool,
     informatieobjecttypeSubmissionCsv: PropTypes.string,
     informatieobjecttypeAttachment: PropTypes.string,
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/V2ConfigFields.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/V2ConfigFields.js
@@ -175,8 +175,8 @@ const V2ConfigFields = ({index, name, schema, formData, onFieldChange, onChange}
         })}
         rawDescription={intl.formatMessage({
           defaultMessage:
-            'Indicates whether the (existing) object should be updated and not created',
-          description: 'Objects API registration options "Upload submission CSV" description',
+            'Indicates whether the existing object (retrieved from an optional initial data reference) should be updated, instead of creating a new one. If no existing object exists, a new one will be created instead',
+          description: 'Objects API registration options "Update existing object" description',
         })}
         rawErrors={getFieldErrors(name, index, validationErrors, 'updateExistingObject')}
         errors={buildErrorsComponent('updateExistingObject')}
@@ -185,7 +185,6 @@ const V2ConfigFields = ({index, name, schema, formData, onFieldChange, onChange}
         <Checkbox
           id="root_updateExistingObject"
           name="updateExistingObject"
-          value={updateExistingObject}
           checked={updateExistingObject}
           onChange={e =>
             onFieldChange({target: {name: 'updateExistingObject', value: !updateExistingObject}})
--- a/src/openforms/registrations/contrib/objects_api/config.py
+++ b/src/openforms/registrations/contrib/objects_api/config.py
@@ -86,6 +86,13 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
         label=_("objecttype version"),
         help_text=_("Version of the objecttype in the Objecttypes API"),
     )
+    update_existing_object = serializers.BooleanField(
+        label=_("Update existing object"),
+        help_text=_(
+            "Indicates whether the (existing) object should be updated and not created."
+        ),
+        default=False,
+    )
     informatieobjecttype_submission_report = serializers.URLField(
         label=_("submission report PDF informatieobjecttype"),
         help_text=_(
--- a/src/openforms/registrations/contrib/objects_api/config.py
+++ b/src/openforms/registrations/contrib/objects_api/config.py
@@ -84,12 +84,12 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
     )
     objecttype_version = serializers.IntegerField(
         label=_("objecttype version"),
-        help_text=_("Version of the objecttype in the Objecttypes API"),
+        help_text=_("Version of the objecttype in the Objecttypes API."),
     )
     update_existing_object = serializers.BooleanField(
         label=_("Update existing object"),
         help_text=_(
-            "Indicates whether the (existing) object should be updated and not created."
+            "Indicates whether the existing object should be updated (if it exists), instead of creating a new one."
         ),
         default=False,
     )
@@ -97,7 +97,7 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
         label=_("submission report PDF informatieobjecttype"),
         help_text=_(
             "URL that points to the INFORMATIEOBJECTTYPE in the Catalogi API "
-            "to be used for the submission report PDF"
+            "to be used for the submission report PDF."
         ),
         required=False,
     )
@@ -105,7 +105,7 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
         label=_("Upload submission CSV"),
         help_text=_(
             "Indicates whether or not the submission CSV should be uploaded as "
-            "a Document in Documenten API and attached to the ProductAanvraag"
+            "a Document in Documenten API and attached to the ProductAanvraag."
         ),
         default=False,
     )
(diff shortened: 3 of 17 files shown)


src/openforms/contrib/objects_api/clients/objects.py
```diff
@@ -4,14 +4,21 @@
 
 
 class ObjectsClient(NLXClient):
-    def create_object(self, objecttype_url: str, record_data: dict) -> dict:
-
+    def create_or_update_object(
```
vaszig, line 7: I had some doubts if I should add a new method here `update_object` but I realized that it's almost the same code, so feel free to give an opinion.

Viicos: I think it's best to have the client classes not know about specific logic and flows and be as close to the API spec in some way, and instead the logic should leave in the caller's code:

```python
if initial_data_reference:
    client.update_object(...)
else:
    client.create_object(...)
```

src/openforms/contrib/objects_api/clients/objects.py
```diff
@@ -15,3 +15,17 @@ def create_object(self, objecttype_url: str, record_data: dict) -> dict:
         response.raise_for_status()
 
         return response.json()
+
+    def update_object(
+        self, objecttype_url: str, record_data: dict, initial_data_reference: str
```
Viicos, line 20: I think in this case we can use `object_uuid` instead of `itinal_data_reference` (as it is scoped to the Objects API)

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -86,6 +86,13 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
         label=_("objecttype version"),
         help_text=_("Version of the objecttype in the Objecttypes API"),
     )
+    update_existing_object = serializers.BooleanField(
+        label=_("Update existing object"),
+        help_text=_(
+            "Indicates whether the (existing) object should be updated and not created."
```
Viicos, line 92:
```suggestion
            "Indicates whether the existing object should be updated (if it exists), instead of creating a new one"
```

Or something similar?

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -92,15 +92,31 @@ def register_submission(
...
-                    record_data=record_data,
-                ),
-                submission,
-                "intermediate.objects_api_object",
-            )
+            # update or create the object
+            if not (
+                options.get("update_existing_object")
+                and submission.initial_data_reference
+            ):
```
Viicos, line 99: Is this condition right? I would expect the opposite:

```suggestion
            if (
                options.get("update_existing_object")
                and submission.initial_data_reference
            ):
                # update object
            else:
                # create object
```

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +226,199 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+        super().setUp()
+
+        config_patcher = patch(
+            "openforms.registrations.contrib.objects_api.models.ObjectsAPIConfig.get_solo",
+            return_value=ObjectsAPIConfig(),
+        )
+        self.mock_get_config = config_patcher.start()
+        self.addCleanup(config_patcher.stop)
+
+        self.objects_api_group = ObjectsAPIGroupConfig.objects.create(
```
Viicos, line 244: You may need to rebase, but you can now use `ObjectsAPIGroupConfigFactory.create(for_test_docker_compose=True)`

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +226,199 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+                    },
+                ],
+                submitted_data={"voornaam": "Foo"},
+                form_definition_kwargs={"slug": "fd-create"},
+            )
+            object_create_result = plugin.register_submission(
+                submission_create,
+                {
+                    "version": 1,
+                    "objecttype": UUID("8e46e0a5-b1b4-449b-b9e9-fa3cea655f48"),
```
Viicos, line 301: In https://github.com/open-formulieren/open-forms/pull/4548, I created an objecttype which accepts anything. `8e46e0a5-b1b4-449b-b9e9-fa3cea655f48` still works because I think by default `additionalProperties` is `true`, but it might be better to use the new one I added for clarity (once it gets merged).

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +226,199 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+
+            assert object_create_result
+
+            self.assertEqual(
+                object_create_result["record"]["data"]["data"]["fd-create"]["voornaam"],
+                "Foo",
+            )
+
+        with self.subTest("Update the above existing object"):
+            submission_update = SubmissionFactory.from_components(
```
Viicos, line 316: Question: `SubmissionFactory.from_components` will create a new form, different from the first one. Does this reflect the expected flow? From what I read in the feature request, seems to be the case (a different form is used but they want to update the same registration object) but just wanted to be sure I'm not mistaken.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -92,15 +92,31 @@ def register_submission(
...
-                    objects_client.create_object,
-                    objecttype_url=objecttype_url,
-                    record_data=record_data,
-                ),
-                submission,
-                "intermediate.objects_api_object",
-            )
+            # update or create the object
+            if not (
+                options.get("update_existing_object")
```
Viicos, line 97:
```suggestion
                options["update_existing_object"]
```

I think it's fine as the serializer as a default value. I'm also wondering if we should add a data migration to add this default value at the DB level? Probably not necessary.

src/openforms/js/components/admin/form_design/registrations/objectsapi/V2ConfigFields.js
```diff
@@ -166,6 +167,31 @@ const V2ConfigFields = ({index, name, schema, formData, onFieldChange, onChange}
...
       </CustomFieldTemplate>
+      <CustomFieldTemplate
+        id="root_updateExistingObject"
+        label={intl.formatMessage({
+          defaultMessage: 'Update existing object',
+          description: 'Objects API registration options "Update existing object" label',
+        })}
+        rawDescription={intl.formatMessage({
+          defaultMessage:
+            'Indicates whether the (existing) object should be updated and not created',
```
Viicos, line 178: May need to be changed depending on the outcome of my other comment

src/openforms/registrations/contrib/objects_api/typing.py
```diff
@@ -15,6 +15,7 @@ class _BaseRegistrationOptions(TypedDict, total=False):
     objects_api_group: Required[ObjectsAPIGroupConfig]
     objecttype: Required[UUID]
     objecttype_version: Required[int]
+    update_existing_object: bool
```
Viicos, line 18: Depending on whether we add a data migration, this might be `Required[bool]`

Viicos (changes requested):

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
vaszig: I don't think that affects the desired result but indeed a nice test would be one form with multiple submissions I guess. I have updated this.

src/openforms/registrations/contrib/objects_api/plugin.py
vaszig: I don't see the difference here. I understand that we are doing the same check in your suggestion but we start from the update. Do you mean it's better in this order?

src/openforms/registrations/contrib/objects_api/typing.py
vaszig: I don't see the reason for a data migration for this, maybe Sergei has a different opinion though..

sergei-maertens: fine as is, missing/`None` is equivalent to explicit `False` which should be the default

sergei-maertens: actually, looking at https://github.com/open-formulieren/open-forms/pull/4538#discussion_r1687768751 - if the serializer field indeed has a `default=` specified, then you can make the type hint `Required[...]` since the serializer will ensure the key is always provided, even if it was not initially provided.

Viicos: Ah yes it is `Required` even without the migration, as this typed dict represents the data from the serializer (so with the default value)

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
Viicos: Hum I don't think it had to be changed then, the initial flow you had created different forms, so I just wanted to make sure it reflected the real use case hence my question

Viicos: I discussed with Sergei about this:

In fact it doesn't matter who/what created the initial data record (it could be from another form from OF, but also an external party). So to simplify the test I could just use the objects client to create the initial data, and then test the flow we are interested in (this is done in a similar way [here](https://github.com/open-formulieren/open-forms/blob/6fe0b1ef35709c9772fd1936e2ffb98caf5ba1e3/src/openforms/registrations/contrib/objects_api/tests/test_update_payment_status_v2.py#L43-L69))

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +228,174 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+                                "voornaam": "My last name",
+                            },
+                        },
+                        objecttype_version=3,
+                    ),
+                    objecttype_url="http://objecttypes-web:8000/api/v2/objecttypes/8e46e0a5-b1b4-449b-b9e9-fa3cea655f48",
+                )
+                # Because of the nginx reverse proxy, we need to set the correct
+                # host as this URL will be used by the plugin to update the payment status:
+                data["url"].replace("objects-web:8000", "localhost:8002")
```
Viicos, line 285: I think this can be moved outside the subtest

Viicos, line 285: No need for this if `data` isn't used in the rest of the test

src/openforms/contrib/objects_api/clients/objects.py
```diff
@@ -15,3 +15,17 @@ def create_object(self, objecttype_url: str, record_data: dict) -> dict:
         response.raise_for_status()
 
         return response.json()
+
+    def update_object(
+        self, objecttype_url: str, record_data: dict, object_uuid: str
+    ) -> dict:
+        endpoint = f"objects/{object_uuid}"
+        json = {
+            "type": objecttype_url,
```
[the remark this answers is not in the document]
vaszig: I did it because of the specification body but indeed it's not needed.

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -86,6 +86,13 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
         label=_("objecttype version"),
         help_text=_("Version of the objecttype in the Objecttypes API"),
     )
+    update_existing_object = serializers.BooleanField(
+        label=_("Update existing object"),
+        help_text=_(
+            "Indicates whether the existing object should be updated (if it exists), instead of creating a new one"
```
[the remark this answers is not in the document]
vaszig: Sorry for this but I thought I should follow the rest of the fields structure, where none of them has a dot at the end!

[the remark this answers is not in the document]
sergei-maertens: ah yeah, didn't pay attention to that. Please do add a period to all of them! Help texts should be proper sentences and follow those grammar rules.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -92,14 +92,27 @@ def register_submission(
...
                     objecttype_url=objecttype_url,
                     record_data=record_data,
-                ),
+                )
+            )
+            response = execute_unless_result_exists(
+                update_or_create,
                 submission,
-                "intermediate.objects_api_object",
+                "intermediate.objects_api_object_create_or_update",
```
sergei-maertens, line 115: you can keep using `intermediate.objects_api_object`, otherwise we'll get problems with retries right after the update :)

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +227,178 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+                    "objects_api_group": objects_api_group,
+                    "update_existing_object": True,
+                },
+            )
+
+            assert object_update_result
+
+            with get_objects_client(objects_api_group) as client:
+                obj_retrieved = client.get_object(
+                    object_uuid=object_update_result["uuid"]
```
sergei-maertens, line 331:
```suggestion
                    object_uuid=created_obj["uuid"]
```
otherwise the UUID assertion below doesn't make any sense

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +227,178 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+            )
+
+            assert object_create_result2
+
+            self.assertEqual(
+                object_create_result2["record"]["data"]["data"]["fd-test"]["voornaam"],
+                "Updated value",
+            )
+            self.assertNotEqual(
+                object_create_result2["uuid"], object_create_result["uuid"]
```
sergei-maertens, line 372:
```suggestion
                object_create_result2["uuid"], created_obj["uuid"]
```
no?

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +227,178 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+            )
+            self.assertEqual(obj_retrieved["uuid"], object_update_result["uuid"])
+
+        with self.subTest(
+            "Existing object not updated when update_existing_object=False"
+        ):
+            submission_update2 = SubmissionFactory.create(
+                form=form,
+                completed=True,
+                initial_data_reference=object_create_result["uuid"],
```
sergei-maertens, line 346:
```suggestion
                initial_data_reference=created_obj["uuid"],
```
we want to use the object we initially created, right?

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +227,178 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+            )
+
+            assert object_create_result3
+
+            self.assertEqual(
+                object_create_result3["record"]["data"]["data"]["fd-test"]["voornaam"],
+                "Updated value",
+            )
+            self.assertNotEqual(
+                object_create_result3["uuid"], object_create_result["uuid"]
```
sergei-maertens, line 403:
```suggestion
                object_create_result3["uuid"], created_obj["uuid"]
```

src/openforms/contrib/objects_api/clients/objects.py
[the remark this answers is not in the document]
sergei-maertens: yeah drf-spectacular/drf was a bit weird about that, but it's been fixed in recent releases!

src/openforms/registrations/contrib/objects_api/tests/test_backend.py
```diff
@@ -220,3 +227,172 @@ def test_registration_works_after_failure(self, m: requests_mock.Mocker):
...
+                    "objecttype": UUID("8e46e0a5-b1b4-449b-b9e9-fa3cea655f48"),
+                    "objecttype_version": 3,
+                    "objects_api_group": objects_api_group,
+                    "update_existing_object": False,
+                },
+            )
+
+            assert object_create_result is not None
+
+            self.assertNotEqual(object_create_result["uuid"], created_obj)
```
sergei-maertens, line 299: this compares a dict with a string, so I think it should be:

```suggestion
            self.assertNotEqual(object_create_result["uuid"], created_obj["uuid"])
```

sergei-maertens (approved):

PR #4547, opened 2024-07-22
Prevent saving the registration backend key too soon

### LINKED ISSUE #4502: Pausing form triggers "Submission.finalised_registration_backend_key" causing incorrect registration selection

2.6.x

### Customer reference

_No response_

Pausing a form causes the registration to be set, which can lead to submissions registering to the wrong back-end when it's changed after un-pausing. 

create a form with 2 registration option
create a checkbox 
create logic for this checkbox to change registration on select. 
start form and pause submission
continue submission and click checkbox 
watch submission use registration for unchecked checkbox

registration backend should not be saved when pausing a form

sergei-maertens:
Closes #4502 

**Changes**

* Added a bandaid fix that can be cherrypicked to stable branches - block the registration backend key from being saved if the submission is not completed yet
* Improved on the fix by making it a hard database constraint for the master branch

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/submissions/models/submission.py
```diff
@@ -347,6 +347,24 @@ def __str__(self):
...
+    def save(self, *args, **kwargs):
+        update_fields = kwargs.get("update_fields")
+        save_all_fields = update_fields is None
+        _will_save_backend_key = (
+            save_all_fields or "finalised_registration_backend_key" in update_fields
+        )
+        if (
+            _will_save_backend_key
+            and self.finalised_registration_backend_key
+            and not self.completed_on
```
Viicos: How is it possible that `self.finalised_registration_backend_key` is set (with `self.completed_on` falsy) in the first place? This seems to contradict with the `ValueError` message below, no?

Viicos: Ah no makes sense, `self.finalised_regisration_backend_key` was set before saving

src/openforms/submissions/models/submission.py
```diff
@@ -339,6 +339,20 @@ class Meta:
                 ),
                 name="registration_status_consistency_check",
             ),
+            models.CheckConstraint(
```
Viicos, line 342: Should we have a test for this constraint, or are we confident it will work as expected?

src/openforms/submissions/models/submission.py
```diff
@@ -339,6 +339,20 @@ class Meta:
...
+                check=(
+                    models.Q(finalised_registration_backend_key="")
+                    | (
+                        ~models.Q(finalised_registration_backend_key="")
+                        & models.Q(completed_on__isnull=False)
+                    )
```
Viicos, line 348: Doesn't the ` | ` part takes into account the fact that `~models.Q(finalised_registration_backend_key="")`? So maybe it can be simplified as:

```suggestion
                    models.Q(finalised_registration_backend_key="")
                    | (
                        models.Q(completed_on__isnull=False)
                    )
```

src/openforms/submissions/models/submission.py
```diff
@@ -339,6 +339,20 @@ class Meta:
                 ),
                 name="registration_status_consistency_check",
             ),
+            models.CheckConstraint(
+                check=(
+                    models.Q(finalised_registration_backend_key="")
+                    | (
+                        ~models.Q(finalised_registration_backend_key="")
+                        & models.Q(completed_on__isnull=False)
```
Viicos, line 347:
```suggestion
                        & models.Q(completed_on__isnull=True)
```

Or am I missing something?

src/openforms/submissions/models/submission.py
sergei-maertens: indeed, it's set from the logic action evaluation. tracking that one down was not fun

src/openforms/submissions/models/submission.py
sergei-maertens: we have tests that verify the registration backend key is properly persisted after logic evaluation, and we have plenty of tests that otherwise save the submission, so if it didn't work, there would've been a lot of test failures :P it works ;)

src/openforms/submissions/models/submission.py
sergei-maertens: yes, I had that initially but then opted for a more explicit variant to really illustrate the two different cases (by including the common factor between the two cases)

I'd expec the ORM/database to optimize this anyway

src/openforms/submissions/models/submission.py
sergei-maertens: two cases:

1. either there is no registration backend key set at all - the value of `completed_on` doesn't matter
2. there is a non-empty registration backend key, in that case `completed_on` must be set (otherwise the submission is not completed) so it must not be `null` -> `completed_on__isnull=False`

sergei-maertens: another way to read this is `completed_on__isnotnull`

sergei-maertens:
Backports:

* stable/2.7.x: ad926c7153b54c99f7ef607b7f90d117a5790e3f
* stable/2.6.x: a74d0fc6e1b93e72be97970a8418b85cc7823bfd
* stable/2.5.x: e293eb9038f25f454a63bc045e1464f96ab3c7fd

PR #4549, opened 2024-07-22
Ensure payment email is sent, irrespective of wait_for_payment registration mode

sergei-maertens:
Closes #4334 

**Changes**

* If wait_for_payment option is enabled and the submission is paid, immediately schedule the payment update email too

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

PR #4550, opened 2024-07-22
Add a generic type var for registration plugin options

sergei-maertens:
Noticed type errors and mismatches in the type definition vs. serializer while working on #4549, so this follows that PR.

### DIFF
--- a/pyright.pyproject.toml
+++ b/pyright.pyproject.toml
@@ -24,6 +24,8 @@ include = [
     # Formio tooling
     "src/openforms/formio/typing/",
     "src/openforms/formio/formatters/",
+    # Payments
+    "src/openforms/payments/models.py",
     # Registrations
     "src/openforms/registrations/contrib/email/config.py",
     "src/openforms/registrations/contrib/email/plugin.py",
--- a/pyright.pyproject.toml
+++ b/pyright.pyproject.toml
@@ -24,6 +24,9 @@ include = [
     # Formio tooling
     "src/openforms/formio/typing/",
     "src/openforms/formio/formatters/",
+    # Registrations
+    "src/openforms/registrations/contrib/email/config.py",
+    "src/openforms/registrations/contrib/email/plugin.py",
     # our own template app/package on top of Django
     "src/openforms/template",
     # generic typing helpers
@@ -31,7 +34,7 @@ include = [
 ]
 exclude = [
     "**/__pycache__",
-    "src/openforms/formio/formatters/tests/"
+    "src/openforms/formio/formatters/tests/",
 ]
 ignore = []
 
--- a/src/openforms/registrations/base.py
+++ b/src/openforms/registrations/base.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Generic, TypeVar
 
 from rest_framework import serializers
 
@@ -29,7 +29,10 @@ class PreRegistrationResult:
     data: dict | None = None
 
 
-class BasePlugin(ABC, AbstractBasePlugin):
+OptionsT = TypeVar("OptionsT", bound=dict)
+
+
+class BasePlugin(Generic[OptionsT], ABC, AbstractBasePlugin):
     configuration_options: SerializerCls = EmptyOptions
     """
     A serializer class describing the plugin-specific configuration options.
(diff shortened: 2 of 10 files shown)


pyright.pyproject.toml
sergei-maertens, line 1: SQUASH MERGE and then add the commit to the git-rev-parse-ignore file!

src/openforms/registrations/base.py
```diff
@@ -29,7 +29,14 @@ class PreRegistrationResult:
...
 
-class BasePlugin(ABC, AbstractBasePlugin):
+class Options(TypedDict):
+    pass
+
+
+OptionsT = TypeVar("OptionsT", bound=Options)
+
+
+class BasePlugin(Generic[OptionsT], ABC, AbstractBasePlugin):
```
Viicos, line 39: Checked and pyright seems to work and enforce the parametrization of the type var to be a typed dict

src/openforms/registrations/contrib/email/config.py
```diff
@@ -91,3 +94,30 @@ class Meta:
...
+    email_content_template_text: NotRequired[str]
+
+
+# sanity check for development - keep serializer and type definitions in sync
+_serializer_fields = EmailOptionsSerializer._declared_fields.keys()
+_options_keys = Options.__annotations__.keys()
+assert (
+    _serializer_fields == _options_keys
+), "Mismatch between serializer fields and options dictionary keys!"
+del _serializer_fields, _options_keys
```
Viicos, line 123: Interesting :thinking:

src/openforms/registrations/contrib/email/config.py
```diff
@@ -91,3 +94,30 @@ class Meta:
...
+class Options(TypedDict):
+    """
+    Shape of the email registration plugin options.
+
+    This describes the shape of :attr:`EmailOptionsSerializer.validated_data`, after
+    the input data has been cleaned/validated.
+    """
+
+    to_emails: list[str]
+    attachment_formats: NotRequired[list[AttachmentFormat | str]]
```
Viicos, line 108: Why not just `list[AttachmentFormat]`?

src/openforms/submissions/models/submission.py
```diff
@@ -321,7 +323,10 @@ class Submission(models.Model):
     _prefilled_data = None
     _total_configuration_wrapper = None
 
+    # type hints for (reverse) related fields
     auth_info: AuthInfo
+    report: SubmissionReport
+    payments: SubmissionPaymentManager
```
Viicos, line 329: iirc the reverse related manager is dynamically created :/ Not much we can do

src/openforms/submissions/exports.py
```diff
@@ -80,7 +80,7 @@ def create_submission_export(queryset: SubmissionQuerySet) -> tablib.Dataset:
 
 
 def export_submissions(
-    queryset: SubmissionQuerySet, file_type: FileType
+    queryset: models.QuerySet[Submission], file_type: FileType
```
Viicos, line 83: Why not keep `SubmissionQuerySet`?

src/openforms/registrations/contrib/email/config.py
sergei-maertens: it's disgusting and it will slightly affect the application startup, but it's a hit worth taking :grimacing:

src/openforms/registrations/contrib/email/config.py
sergei-maertens: because the key is absent in `serializer.validated_data` if you do this:

```py
>>> from openforms.registrations.contrib.email.config import *
>>> serializer = EmailOptionsSerializer(data={"to_emails": []})
>>> serializer.is_valid()
True
>>> serializer.validated_data
{'to_emails': [], 'attach_files_to_email': None}

```

sergei-maertens: Oh, misunderstood! it's about the `AttachmentFormat` vs. `str` union - restoring from the JSON in the DB will have strings and not the enum at runtime. They pass equality checks, but they are a different type.

Ideally we can produce a `Literal[...]` from an enum, but that's currently not possible in Python.

src/openforms/registrations/contrib/email/config.py
Viicos: Even hackier, this check could be moved to the plugin class, by using the `serialization_options` class var and the parametrized options class

src/openforms/submissions/exports.py
sergei-maertens: type error in the email plugin because it can only infer that a queryset of `Submission` is passed. The way the queryset is consumed here is compatible with a broader type and fixes the type error. I could not for the life of me make the manager return the `SubmissionQuerySet` type, it only wants to produce `models.QuerySet[Submission, Submission]` and it baffles me because I did the setup the same way like the payments manager and that one seems to work properly :thinking: 

So it's a compromise and I hope in a year from now there will be better built-in support

src/openforms/registrations/contrib/email/config.py
sergei-maertens: good shout, though we can only do that once all plugins are updated with the options annotation :thinking: I'll keep it in mind for when we get there!

Viicos (approved):

PR #4551, opened 2024-07-23
Bump dependencies

sergei-maertens:
**Changes**

* Switched to `uv`, following our update in default-project
* `setuptools` is dropped from the requirements because of the above
* Bumped sentry-sdk which has a dependabot alert (we're not affected)


Viicos (approved):
Great to see `setuptools` removed

PR #4554, opened 2024-07-23
:recycle: Reduce test flakiness

sergei-maertens:
Replaced 4xx and 5xx status code fuzzy with hypothesis with factory boy fuzzing, which also has options to make it reproducible if needed. This should reveal potential issues over time the same way that hypothesis does, except it will only run one case in each CI run of course where hypothesis could potentially cover the entire range in a single test run by generating 100 (or less) samples.


Viicos (approved):
This will also reduce the time it takes to run these tests

PR #4565, opened 2024-07-26
Address flakiness in tests

sergei-maertens:
* Report the random state of factory body - this may help in reproducing flaky tests.
* Fixed a cache isolation/reset causing flaky tests


sergei-maertens:
well, that test flakiness does not appear to be due to factory boy random state, can't reproduce locally yet :(

sergei-maertens:
Going to merge this one as-is, and other issues that pop up I will try to investigate asap.

PR #4571, opened 2024-07-31
:whale: Allow configuring log-outgoing-requests through envvars

sergei-maertens:
Some customers request outgoing request logging to be enabled on test instances (permanently). We can parametrize this via envvars now.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/conf/base.py
+++ b/src/openforms/conf/base.py
@@ -626,10 +626,6 @@ MAX_UNTRUSTED_JSON_PARSE_SIZE = config(
 ESCAPE_REGISTRATION_OUTPUT = config("ESCAPE_REGISTRATION_OUTPUT", default=False)
 DISABLE_SENDING_HIDDEN_FIELDS = config("DISABLE_SENDING_HIDDEN_FIELDS", default=False)
 
-LOG_OUTGOING_REQUESTS_MAX_AGE = config(
-    "LOG_OUTGOING_REQUESTS_MAX_AGE", default=7
-)  # number of days
-
 # TODO: convert to feature flags so that newly deployed instances get the new behaviour
 # while staying backwards compatible for existing instances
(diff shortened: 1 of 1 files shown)


sergei-maertens:
@sjoerdie FYI, more environment variables that you can configure :) I'm not documenting these since I am against making it too easy to deviate from safe-by-default practices

src/openforms/conf/base.py
```diff
@@ -1165,15 +1161,27 @@
...
+LOG_OUTGOING_REQUESTS_EMIT_BODY = config(
+    "LOG_OUTGOING_REQUESTS_EMIT_BODY", default=True
+)
 LOG_OUTGOING_REQUESTS_MAX_CONTENT_LENGTH = 524_288  # 0.5MB
+# number of days
+LOG_OUTGOING_REQUESTS_MAX_AGE = config("LOG_OUTGOING_REQUESTS_MAX_AGE", default=7)
+# number of minutes, set to 0 to disable reset
+LOG_OUTGOING_REQUESTS_RESET_DB_SAVE_AFTER = config(
+    "LOG_OUTGOING_REQUESTS_RESET_DB_SAVE_AFTER", default=60
+)
```
stevenbal, line 1184: Do I understand it correctly that having `LOG_OUTGOING_REQUESTS_DB_SAVE=True` makes this setting obsolete?

stevenbal (approved):

src/openforms/conf/base.py
sergei-maertens: not really - you can disable saving in the admin, and then after 60 mins (by default), it will be reset to the "settings default", which in your example would re-enable it. It's a bit of a weird flow - nobody is likely to do this, but it does highlight the implementation details.

PR #4576, opened 2024-08-02
:wrench: [#4575] added SENDFILE_BACKEND configuration option

### LINKED ISSUE #4575: Send file backend is not configurable

2.7.1

### Customer reference

_No response_

I would expect the SENDFILE_BACKEND option would be configurable using an environment variable. This is for open forms installations that do not run behind an nginx proxy and want to use different [send file backends](https://django-sendfile2.readthedocs.io/en/latest/backends.html)

See https://github.com/open-formulieren/open-forms/blob/master/src/openforms/conf/base.py#L346
I'll create a PR to adres this issue.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

marnixdessing:
Closes #4575 

**Changes**

Added SENDFILE_BACKEND option to conf/base.py and set ci and dev defaults using the environment variable. Included update to documentation. 

Note of caution: I haven't been able to test the changes.

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/docs/installation/config.rst
+++ b/docs/installation/config.rst
@@ -365,6 +365,8 @@ Other settings
 
 * ``SUBPATH``: A string with a prefix for all URL paths, for example ``/openforms``. Typically used at the infrastructure level to route to a particular application on the same (sub)domain. Defaults to empty string meaning that Open Forms is hosted at the root (``/``).
 
+* ``SENDFILE_BACKEND``: which backend to use for authorization-secured upload downloads. Defaults to sendfile.backends.nginx. See `django-sendfile2 <https://pypi.org/project/django-sendfile2/>`_` for available backends.
+
 .. _`Django DATABASE settings`: https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-DATABASE-ENGINE
 
 Specifying the environment variables
--- a/docs/installation/config.rst
+++ b/docs/installation/config.rst
@@ -365,7 +365,14 @@ Other settings
 
 * ``SUBPATH``: A string with a prefix for all URL paths, for example ``/openforms``. Typically used at the infrastructure level to route to a particular application on the same (sub)domain. Defaults to empty string meaning that Open Forms is hosted at the root (``/``).
 
-* ``SENDFILE_BACKEND``: which backend to use for authorization-secured upload downloads. Defaults to sendfile.backends.nginx. See `django-sendfile2 <https://pypi.org/project/django-sendfile2/>`_` for available backends.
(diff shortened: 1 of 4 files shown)


docs/installation/config.rst
```diff
@@ -365,6 +365,8 @@ Other settings
 
 * ``SUBPATH``: A string with a prefix for all URL paths, for example ``/openforms``. Typically used at the infrastructure level to route to a particular application on the same (sub)domain. Defaults to empty string meaning that Open Forms is hosted at the root (``/``).
 
+* ``SENDFILE_BACKEND``: which backend to use for authorization-secured upload downloads. Defaults to sendfile.backends.nginx. See `django-sendfile2 <https://pypi.org/project/django-sendfile2/>`_` for available backends.
```
sergei-maertens, line 368: Slight re-wording and added a disclaimer:

```suggestion
* ``SENDFILE_BACKEND``: which backend to use to serve the content of non-public files. The value depends on the
  reverse proxy solution used with Open Forms. For available backends, see the `django-sendfile documentation`_.
  Defaults to ``sendfile.backends.nginx``.
  
  .. note:: Open Forms only considers nginx to be in scope. You can deviate from using nginx, but we cannot offer any
    support on other backends.
  
.. _django-sendfile documentation: https://django-sendfile2.readthedocs.io/en/stable/backends.html
```

src/openforms/conf/ci.py
```diff
@@ -16,6 +16,9 @@
 
 os.environ.setdefault("IS_HTTPS", "no")
 os.environ.setdefault("SECRET_KEY", "dummy")
+
+os.environ.setdefault("SENDFILE_BACKEND", "django_sendfile.backends.development")
```
sergei-maertens, line 20:
```suggestion
os.environ.setdefault("SENDFILE_BACKEND", "django_sendfile.backends.development")
```

sergei-maertens (changes requested):
Thanks for the contribution!

I suggested some cosmetic changes and more importantly, we need to protect our scope since our team can only offer support on nginx.

sergei-maertens:
Which backend are you planning to use by the way? Since some backends require additional settings to be defined, and for that we currently don't have a mechanism.

marnixdessing:
For now we just want to use the `django_sendfile.backends.simple` backend as we are still in the experimentation phase with open forms and ZGW in general. We are running the container on AWS without an reverse proxy for simplicity so we need this configuration option for downloads to work.

PR #4574, opened 2024-07-31
Allow lookup of document types within specified catalogue (backend)

sergei-maertens:
Partly closes #4577

**Changes**

- [x] Add defaults field for catalogue selection
- [x] Validate the specified catalogue at the model layer
- [x] Add field(s) for catalogue selection at the form level
- [x] Validate the specified catalogue at the serializer level
- [x] Validate that the documenttype URLs are in the specified catalogue (if provided)
- [x] Allow specifying documenttype references by 'omschrijving'
- [x] Validate the documenttype with omschrijving exists in the specified catalogue

For a different PR, as this one has grown quite large:

- Dynamically resolve the correct documenttype to use within the specified catalogue and applicable version (select based on valid from/until combined with the submission completion timestamp)
- Selection of document types -> feature flag to include/exclude draft/concept document types

Likely for a different PR to keep things as small as possible...

- Update the UI to allow selecting catalogue in dropdown
- Update the UI to allow selecting the documenttype in a dropdown

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

PR #4578, opened 2024-08-02
Update plugin registration behaviour to take document type resolution into account

### LINKED ISSUE #4577: Specify document type to use in Objects API registration in a more natural way
Allow specifying the document type as a combination of `(catalogue, description)` so that the appropriate version can be selected dynamically at runtime.

Split off from #4267

**Tasks**

- [x] Update configuration aspect in backend #4574
    - [x] Add defaults field for catalogue selection
    - [x] Validate the specified catalogue at the model layer
    - [x] Add field(s) for catalogue selection at the form level
    - [x] Validate the specified catalogue at the serializer level
    - [x] Validate that the documenttype URLs are in the specified catalogue (if provided)
    - [x] Allow specifying documenttype references by 'omschrijving'
    - [x] Validate the documenttype with omschrijving exists in the specified catalogue
- [x] Dynamically resolve the correct documenttype to use within the specified catalogue and applicable version (select based on valid from/until combined with the submission completion timestamp) #4578
- [x] Add a feature flag to allow/disallow draft ("concept") document types during validation & registration phase
- [x] Update UI to allow users to use the new configuration
    - [x] allow selecting catalogue in dropdown
    - [x] allow selecting the documenttype in a dropdown
    - [x] Move the existing URL-based fields to a "deprecated" fieldset

sergei-maertens:
~~Depends on #4574~~ (it's merged)

---

Partly closes #4577

**Changes**

* Refactored the informatieobjecttype determination
* Cleaned up some typing information and added more files to the Pyright config
* Made the informatieobjecttype resolution dynamic with a fallback to the specified URLs

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -61,43 +67,86 @@ def _point_coordinate(value: Any) -> dict[str, Any] | object:
...
+    if not document_type:
         return ""
 
     submission_report = SubmissionReport.objects.get(submission=submission)
-    submission_report_options = build_options(
-        options,
-        {
-            "informatieobjecttype": "informatieobjecttype_submission_report",
-            "organisatie_rsin": "organisatie_rsin",
-            "doc_vertrouwelijkheidaanduiding": "doc_vertrouwelijkheidaanduiding",
```
stevenbal, line 92: Is `doc_vertrouwelijkheidaanduiding` no longer needed? I don't see it in the new `options`

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -107,30 +156,28 @@ def register_submission_csv(
...
+    if not document_type:
         return ""
 
-    submission_csv_options = build_options(
-        options,
-        {
-            "informatieobjecttype": "informatieobjecttype_submission_csv",
-            "organisatie_rsin": "organisatie_rsin",
-            "doc_vertrouwelijkheidaanduiding": "doc_vertrouwelijkheidaanduiding",
-            "auteur": "auteur",
```
stevenbal, line 123: Same as above, these two seem to be removed, are they no longer needed?

src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
```diff
@@ -316,7 +309,17 @@ def test_retrieve_with_explicit_objects_api_group(self):
 
         data = response.json()
 
-        self.assertEqual(len(data), 3)
+        num_document_types = len(data)
+        # at least 6 distinct document types are defined in the fixture
+        self.assertGreaterEqual(num_document_types, 6)
```
stevenbal, line 314: Is there a reason to use `assertGreaterEqual` instead of `assertEqual` here?

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -41,7 +46,43 @@ class InformatieObjectType(TypedDict):
...
+
+
 class CatalogiClient(NLXClient):
+    _api_version: CatalogiAPIVersion | None = None
+
+    @property
+    def api_version(self) -> CatalogiAPIVersion:
+        if self._api_version is None:
+            # hit a known endpoint and parse the response headers
+            response = self.get("catalogussen", params={"domein": "VRSN"})
```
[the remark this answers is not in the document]
sergei-maertens: made up - it is not needed but I want to limit the result set as much as possible and keep load on the API limited :) VRSN = VERSION, just a dummy, expected to not yield any results

preferably I'd hit the root endpoint, but that endpoint is not documented/required to return anything, so it's unreliable :(

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -87,26 +128,54 @@ def find_informatieobjecttypen(
...
         If no results are found, ``None`` is returned.
         """
-        response = self.get(
-            "informatieobjecttypen",
-            params={
-                "catalogus": catalogus,
-                "omschrijving": description,
-            },
-        )
+        _supports_filtering_valid_on = self.api_version > (1, 2)
```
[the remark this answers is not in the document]
sergei-maertens: in a way I hoped to convey that the patch version is irrelevant, only the minor version determines the behaviour, but it's a small effort to make it `>= (1, 2, 0)` if that's clearer!

src/openforms/registrations/contrib/objects_api/submission_registration.py
sergei-maertens: the "funny" thing with these options was that they were never specified anywhere, so they were also not being applied as an override. it was just copy-pasted from the ZGW API's.

AFAIK, only the file component allows you to override the confidentiality level, and the ZGW API's allow you to configure a default.

src/openforms/registrations/contrib/objects_api/tests/test_api_endpoints.py
sergei-maertens: future proofing in case more document types get added in the fixture and not having to update this exact assertion.

these tests are more tests that "anything" is returned and there may be more data (but rather we check that only unique descriptions are returned etc. instead of the exact count)

stevenbal (approved):

PR #4580, opened 2024-08-03
Improve UX by using react-select (Objects API registration options)

sergei-maertens:
Related to #4577 

**Changes**

* Refactored existing selects to use react-select under the hood. This provides a search field which is useful for large lists to filter down in the options.
* Updated to new formio-builder version so we can re-use the dark-mode support styles
* Added base react-select component with some additional django-admin style overrides
* Refactored some duplicated code into single component so that we can easily deprecate it next

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/ObjectTypeVersionSelect.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/ObjectTypeVersionSelect.js
@@ -4,7 +4,7 @@ import useAsync from 'react-use/esm/useAsync';
 
 import Field from 'components/admin/forms/Field';
 import FormRow from 'components/admin/forms/FormRow';
-import Select, {LOADING_OPTION} from 'components/admin/forms/Select';
+import ReactSelect from 'components/admin/forms/ReactSelect';
 import {get} from 'utils/fetch';
 
 import {useSynchronizeSelect} from './hooks';
@@ -29,7 +29,6 @@ const getAvailableVersions = async (uuid, apiGroupID) => {
 };
 
 const ObjectTypeVersionSelect = () => {
-  const [fieldProps, , {setValue}] = useField('objecttypeVersion');
   const {
     values: {objectsApiGroup = null, objecttype = ''},
   } = useFormikContext();
@@ -45,11 +44,12 @@ const ObjectTypeVersionSelect = () => {
   if (error) throw error;
 
   const choices = loading
-    ? LOADING_OPTION
+    ? []
     : versions.map(version => [version.version, `${version.version} (${version.status})`]);
 
   useSynchronizeSelect('objecttypeVersion', loading, choices);
 
+  const options = choices.map(([value, label]) => ({value, label}));
   return (
     <FormRow>
       <Field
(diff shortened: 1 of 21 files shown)


src/openforms/js/components/admin/forms/ReactSelect.js
```diff
@@ -0,0 +1,65 @@
+import {getReactSelectStyles} from '@open-formulieren/formio-builder/esm/components/formio/select';
+import {useField} from 'formik';
+import PropTypes from 'prop-types';
+import ReactSelect from 'react-select';
+
+const initialStyles = getReactSelectStyles();
```
stevenbal, line 6: I noticed that the size of the selectbox scales with the search input, while it isn't major, I guess it would be nicer if it stayed the same size
Screenshot from 2024-08-06 12-31-18

src/openforms/js/components/admin/forms/Field.js
```diff
@@ -44,11 +45,15 @@ const Field = ({
 
   const htmlFor = `id_${name}`;
 
-  const childProps = {id: htmlFor, name: originalName};
-  if (disabled) {
-    childProps.disabled = true;
+  const manageChildProps = !noManageChildProps;
```
stevenbal, line 48: Is there a reason why `noManageChildProps = false` is used instead of `manageChildProps` with default `true`?

src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/ObjectTypeVersionSelect.js
```diff
@@ -61,19 +61,14 @@ const ObjectTypeVersionSelect = () => {
...
-          choices={choices}
-          id="id_objecttypeVersion"
-          {...fieldProps}
-          onChange={event => {
-            // overridden to handle the proper data type, since every <option value>
-            // turns into a string in HTML
-            const newVersion = parseInt(event.currentTarget.value, 10);
-            setValue(newVersion);
-          }}
+          options={options}
```
stevenbal, line 69: Locally for me it seems that the version select is overlapped by the fieldset below it
Screenshot from 2024-08-06 12-41-11

stevenbal (changes requested):

src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/ObjectTypeVersionSelect.js
sergei-maertens: I ran into it in another PR where it's fixed - django's `.form-row` hides overflow: https://github.com/open-formulieren/open-forms/pull/4586/files#diff-a95d74cf85e72c9e17c4281231439a672cb7d4d77646f77889f0e77499b76cfcR14

I'll cherry-pick that CSS fix to this branch :)

src/openforms/js/components/admin/forms/ReactSelect.js
sergei-maertens: I can't seem to reproduce this - maybe browser CSS cache? I've tested on Chromium and Firefox.

src/openforms/js/components/admin/forms/Field.js
sergei-maertens: as an extra - at some point we will only have Formik based forms and then the `manageChildProps` construct will be obsolete anyway - React documentation advises against using `React.Children.(...)` and `React.cloneElement` etc. in favour of just explicitly creating the components, so we're mostly dealing with legacy code here.

I'm sure we can construct better wrappers, but I'm hoping we won't need these wrappers at all in the future when we have a design decoupled from the Django admin (see #3631)

PR #4585, opened 2024-08-05
:construction_worker: Set up CI job to test upgrade checks

### LINKED ISSUE #4584: Add smoke test for upgrade checks
Releases 2.7.2 and 2.7.3 were made because:

* the check script were not included in the docker image
* there was a typo in the check script name, leading to import errors

We need some CI check to verify that (inside a docker container from the built image) this configuration is correct and complete.

sergei-maertens:
Closes #4584 

* Added CI check so that our upgrade checks run before publishing a release
* Patched the check script crash that happens in 2.7.3


sergei-maertens:
Backport: f07de75ddb421af5d9cd55f30a9cf42ee17cfa07

PR #4595, opened 2024-08-09
Add feature flag for draft document types (and prepare for ZGW similar mechanics)

### LINKED ISSUE #4577: Specify document type to use in Objects API registration in a more natural way

sergei-maertens:
Closes #4577

**Changes**

* [x] Implement a feature flag to allow draft document types
* [x] Test using a draft document type during registration
* [x] Document the feature flag in the objects API & ZGW APIs registration docs

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/docs/installation/config.rst
+++ b/docs/installation/config.rst
@@ -294,16 +294,6 @@ Other settings
   reasons. New installations should opt-out. If ``False``, the OIDC callback URL is
   ``/auth/oidc/callback/``, if ``True``, it is ``/org-oidc/callback/``.
 
-* ``DIGID_EHERKENNING_OIDC_STRICT``: Enable strict claim processing/validation when
-  using :ref:`configuration_authentication_oidc_digid`,
-  :ref:`configuration_authentication_oidc_eherkenning` or
-  :ref:`configuration_authentication_oidc_machtigen`. Defaults to ``False``. You can
-  also add a feature flag in the admin interface to enable this behaviour.
-
-  .. versionadded:: 2.7.0
-     A formal and more complete authentication context data model is used - existing
-     installations likely do not provide all this information yet.
-
 * ``SESSION_EXPIRE_AT_BROWSER_CLOSE``: Controls if sessions expire at browser close.
   This applies to both the session of end-users filling out forms and staff using the
   administrative interface. Enabling this forces users to log in every time they open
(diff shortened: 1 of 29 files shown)


src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -139,17 +159,19 @@ def find_informatieobjecttypen(
...
+        params: InformatieObjectTypeListParams = {
             "catalogus": catalogus,
             "omschrijving": description,
         }
-
         # datumGeldigheid was added in v1.2.0 of the APIs & the APIs reject query string
         # parameters that are unkonwn, so we can not just optmistically send it :(
         if valid_on and _supports_filtering_valid_on:
             params["datumGeldigheid"] = valid_on.isoformat()
+        if self.allow_drafts:
```
stevenbal, line 170: I'm not sure if this is an issue, but here I saved with unpublished types, then changed the feature flag to `False` and the select shows no selected options. When I save it, it does appear to still have the `Unpublished` types selected judging from the errors:

image

docs/installation/config.rst
```diff
@@ -376,6 +366,33 @@ Other settings
...
 
+.. _installation_environment_config_feature_flags:
+
+Feature flags
+=============
+
+Open Forms sometimes supports a layered approach for feature flags, where some
+behaviours can be enabled at deploy-time through environment variables already. If
+this option is not available, you can still enable/disable the feature flag in the
+admin interface, via **Admin** > **Configuration** > **Flag states**.
```
stevenbal, line 377: For my understanding: does setting the feature flag via the admin always override the environment variable behavior?

stevenbal (approved):
Small question/remark, looks good otherwise

src/openforms/contrib/zgw/clients/catalogi.py
sergei-maertens: Yeah, bit of an odd one :thinking: 

It's definitely a design choice that remove systems being down should not break existing configuration - so clearing the value because it cannot be found is out of the question.

Perhaps if a value is set but no valid option found, displaying a warning of sorts would be wise.

docs/installation/config.rst
sergei-maertens: depends on the `required` key being set or not - if you define it as `required` in code and the envvar disables it, then whatever you do in the admin will not matter.

src/openforms/contrib/zgw/clients/catalogi.py
sergei-maertens: Turned it into this: 
image

The message takes feature flag on/off into account.

PR #4598, opened 2024-08-12
✨ [#4420] Backend validation for AddressNL components

### LINKED ISSUE #4420: Validation for fields within AddressNL component
### Thema / Theme

Form designer

to add to: #4418 

Currently forms might be validating on the "city" and "postcode" fields. The AddressNL component does not seem to support validation on the fields within the component. 

If this validation is lost these forms would have no way of using both the AddressNL component and BAG plugin.

stevenbal:
Closes #4420 partly

**Changes**

* Implement backend validation for AddressNL components

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from datetime import datetime
 from typing import Protocol
 
@@ -8,6 +9,7 @@ from django.utils.crypto import constant_time_compare
 from django.utils.html import format_html
 from django.utils.translation import gettext as _
 
+from glom import glom
 from rest_framework import ISO_8601, serializers
 from rest_framework.request import Request
 
@@ -411,10 +413,30 @@ class AddressValueSerializer(serializers.Serializer):
 
     def __init__(self, **kwargs):
         self.derive_address = kwargs.pop("derive_address", None)
+        self.component = kwargs.pop("component", None)
         super().__init__(**kwargs)
+
+    def validate_city(self, value: str) -> str:
+        if city_regex := glom(
+            self.component, "openForms.components.city.validate.pattern", default=""
+        ):
+            if not re.fullmatch(city_regex, value):
+                raise serializers.ValidationError(
+                    _("City does not match the specified pattern."),
+                    code="invalid",
+                )
+        return value
 
     def validate_postcode(self, value: str) -> str:
         """Normalize the postcode so that it matches the regex from the BRK API."""
+        if postcode_regex := glom(
+            self.component, "openForms.components.postcode.validate.pattern", default=""
+        ):
+            if not re.fullmatch(postcode_regex, value):
+                raise serializers.ValidationError(
+                    _("Postcode does not match the specified pattern."),
+                    code="invalid",
+                )
         return value.upper().replace(" ", "")
 
     def validate(self, attrs):
@@ -469,6 +491,7 @@ class AddressNL(BasePlugin[AddressNLComponent]):
             derive_address=component["deriveAddress"],
             required=required,
             allow_null=not required,
+            component=component,
             **extra,
         )
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@fortawesome/fontawesome-free": "^6.1.1",
         "@open-formulieren/design-tokens": "^0.51.0",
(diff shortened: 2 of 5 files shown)


stevenbal:
I noticed this large amount of whitespace inside tabs, but it's an issue on master as well:

image

Removing this line does fix it, but I'm not sure what else it could break :grimacing:

src/openforms/formio/components/custom.py
```diff
@@ -422,10 +425,28 @@ def validate(self, attrs):
 
         city = attrs.get("city", "")
         street_name = attrs.get("streetName", "")
+        postcode = attrs.get("postcode", "")
+
+        if postcode_regex := glom(
+            self.component, "ofComponents.postcode.validate.pattern", default=""
+        ):
+            if not re.fullmatch(postcode_regex, postcode):
```
stevenbal, line 433: Should we do something here to avoid possible ReDoS attacks with specific patterns?

sergei-maertens: hrm, is there a way to detect those easily?

There are plans to offer some "pre-baked" validations like house number/house letter that users can select in a dropdown, with a final option "custom" that would let them write the regex themselves (as they do now), so that should minimize the risk of accidentally writing a bad regex.

stevenbal: I checked, there is a library `re2` by Google which is supposed to mitigate this https://github.com/google/re2/wiki/WhyRE2

There is also discussion in Django to use `re2` instead of `re`: https://forum.djangoproject.com/t/using-google-re2-instead-of-re-in-django/31841

src/openforms/formio/components/custom.py
```diff
@@ -411,10 +413,30 @@ class AddressValueSerializer(serializers.Serializer):
...
+        self.component = kwargs.pop("component", None)
         super().__init__(**kwargs)
 
+    def validate_city(self, value: str) -> str:
+        if city_regex := glom(
+            self.component, "openForms.components.city.validate.pattern", default=""
+        ):
+            if not re.fullmatch(city_regex, value):
+                raise serializers.ValidationError(
+                    _("Invalid city, does not match custom regex"),
```
sergei-maertens, line 425:
```suggestion
                    _("City does not match the specified pattern."),
```

src/openforms/formio/components/custom.py
```diff
@@ -411,10 +413,30 @@ class AddressValueSerializer(serializers.Serializer):
...
+        return value
+
     def validate_postcode(self, value: str) -> str:
         """Normalize the postcode so that it matches the regex from the BRK API."""
+        if postcode_regex := glom(
+            self.component, "openForms.components.postcode.validate.pattern", default=""
+        ):
+            if not re.fullmatch(postcode_regex, value):
+                raise serializers.ValidationError(
+                    _("Invalid postcode, does not match custom regex"),
```
sergei-maertens, line 437:
```suggestion
                    _("Postcode does not match the specified pattern."),
```

src/openforms/formio/typing/custom.py
```diff
@@ -9,5 +9,19 @@ class DateComponent(Component):
...
+    postcode: NotRequired[ComponentValidation]
+    city: NotRequired[ComponentValidation]
+
+
+class CustomOpenFormsComponent(TypedDict):
+    components: AddressValidationComponents
+
+
 class AddressNLComponent(Component):
     deriveAddress: bool
```
sergei-maertens, line 27: I'm not a big fan of bare `dict`'s as they don't give any additional useful information, especially when we know the shape upfront.

I also don't think you have `CustomOpenFormsComponent` hooked up anywhere? It might not be possible due to the inheritance rules of TypedDict. The type annotations/descriptions of formio components kinda suck, `TypedDict` isn't really flexible enough IMO.

stevenbal: Oh whoops, I changed it and added it to the `AddressNLComponent`

stevenbal: I'm not sure what is best to do here, since `AddressNLComponent` inherits from `Component` it already has `openForms` defined as `OpenFormConfig`, but I'm not sure if it makes sense (or even is possible) to define something like `AddressValidationComponents` in a generic way?

sergei-maertens (approved):

PR #4607, opened 2024-08-17
Convert ZGW registration options to modal + Formik

### LINKED ISSUE #4606: Make ZGW registration options consistent with Objects API

sergei-maertens:
Partially closes #4606 - does the groundwork before we can actually add the real UX improvements with nice dropdowns.

**Changes**

* Refactored form fields to Formik
* Moved fields/config form into modal
* Moved case property mappings from modal to a tab inside the new modal
* Fixed some behaviours in the Objects API that were uncovered through testing ZGW
* Updated some error-state styling for react-select

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/js/components/admin/form_design/registrations/objectsapi/LegacyConfigFields.js
```diff
@@ -132,7 +132,7 @@ JSONTemplateField.propTypes = {
   helpText: PropTypes.node,
 };
 
-const ContentJSON = () => (
+export const ContentJSON = () => (
```
sergei-maertens, line 135: re-using this here as-is because it's legacy - I'm hoping we can remove all of this in 3.0 :)

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
```diff
@@ -1,20 +1,117 @@
...
+              <>
+                <ZGWFormFields
+                  index={index}
+                  name={name}
+                  apiGroupChoices={apiGroupChoices}
+                  confidentialityLevelChoices={confidentialityLevelChoices}
+                />
+
+                <SubmitRow>
+                  <SubmitAction
```
vaszig, line 103: Why do we need this since the `SubmitRow` contains `SubmitAction` and `event.preventDefault()`?It's a bit confusing to me..

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
```diff
@@ -1,20 +1,117 @@
...
+            onSubmit={(values, actions) => {
+              onChange({formData: values});
+              actions.setSubmitting(false);
+              setModalOpen(false);
+            }}
+          >
+            {({handleSubmit}) => (
+              <>
+                <ZGWFormFields
+                  index={index}
```
vaszig, line 96: I don't see `index` being used somewhere anymore

vaszig (commented):
And a comment concerning frontend..not so important imo but if it's an easy fix it would be nice (It happens both in storybook and in the admin)

image

sergei-maertens:
> And a comment concerning frontend..not so important imo but if it's an easy fix it would be nice (It happens both in storybook and in the admin)
> 
> image

yeah, this is a known issue - I've looked into fixing it but overlaying the icon on the tab is also ugly. I think in a future iteration we'll just have to include the icon inline with the text, fully inside the tab and that will be the best solution. Out of scope for now!

src/openforms/js/components/admin/form_design/registrations/objectsapi/LegacyConfigFields.js
sergei-maertens: I think for 3.0 we need to either remove the objects API integration in ZGW APIs (this was a temporary solution for Utrecht if I remember correctly), or convert it to the Objects API variable mapping approach. Either way, the `contentJson` template is gone in 3.0. That's why I didn't find it worth it to move this to a shared module instead, considering it should only exist here for another 3 months and then it's gone.

src/openforms/js/components/admin/form_design/registrations/zgw/ManageVariableToPropertyMappings.js
```diff
@@ -0,0 +1,169 @@
...
+              defaultMessage="All simple variables are mapped to a case property."
+            />
+          )}
+        </>
+      )}
+    </FieldArray>
+  );
+};
+
+ManageVariableToPropertyMappings.propTypes = {};
```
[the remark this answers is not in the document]
sergei-maertens: indeed - the component does not require/take any props

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
sergei-maertens: we need to make sure that Formik's `handleSubmit` gets called, since it will run (client-side) validation en then perform the `onSubmit` if everything is fine. So we don't let the browser default behaviour submit to the backend, as we need to handle everything in JS (that's what the `event.preventDefault()` is for).

PR #4617, opened 2024-08-23
Add endpoint to verify an email address

### LINKED ISSUE #4542: Implement email verification procedure for the email component
Taiga DH 684, business case "WMV", needs to be added to intranet still

To address mistakes in provided email addresses, we need to add an (optional) email verification flow for forms where the email address is crucial. The goals are:

* ensure that the user filling out the form does not enter wrong data and actually has access to the email account
* prevent forms from being submitted if the email is not verified

The verification process happens *during* the form filling, not after the form is submitted (different from typical user signup flows).

**User interaction**

1. User encounters an email account that requires verification, they enter `myname@example.com`
2. They click the button "send verification code" next to the email input field
3. a modal opens with a text input to enter the verification code + a descriptive text "that an email is being sent and it can take a couple of seconds for it to arrive. Please make sure to check your spam box yada yada..."
4. User enters verification code - we keep track of it in the submission in the SDK and send it in with the form step submit
5. if the code is wrong or not provided (modal dismissed), then we show a validation error with the email component that verification is required

**Imlementation**

- [x] Add an option to the email component `requireVerification` (can go in the `validation` nest obj I suppose)
- [x] Ensure verification button is displayed if a (valid) email is entered and the option is enabled
- [x] When verification is triggered, generate a code in the backend and save it along with the value of the email address, linked to the submission (note: we verify a particular email, not a component, consider changing the email value after verifying it!)
- [x] Ensure that a verification email is sent (add endpoints, templates, celery task...)
- [ ] Include backend validation of the verification code

sergei-maertens:
Closes #4542 partially - ~~requires #4615 to be merged first~~

**Changes**

* Added endpoint to verify the code obtained via the verification email

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -626,3 +626,37 @@ class EmailVerificationSerializer(serializers.ModelSerializer):
             )
 
         return attrs
+
+
+class VerifyEmailData(EmailVerificationData):
+    code: str
+
+
+class VerifyEmailSerializer(EmailVerificationSerializer):
+    code = serializers.CharField(
+        label=_("verification code"),
+        min_length=6,
+        max_length=6,
+        write_only=True,
+    )
+
+    class Meta(EmailVerificationSerializer.Meta):
+        fields = EmailVerificationSerializer.Meta.fields + ("code",)
+
+    def validate(self, attrs: VerifyEmailData):  # type: ignore
+        verification = EmailVerification.objects.filter(
+            submission=attrs["submission"],
+            component_key=attrs["component_key"],
+            email=attrs["email"],
+            verification_code=attrs["code"],
+        ).first()
+        if verification is None:
+            raise serializers.ValidationError(
+                {"code": _("The verification code does not match.")}
+            )
+
+        # if all is well, store the verification in the instance for the view to
+        # consume
+        self.instance = verification
+
+        return attrs
--- a/src/openforms/submissions/api/views.py
+++ b/src/openforms/submissions/api/views.py
@@ -1,11 +1,14 @@
 from django.conf import settings
 from django.db import transaction
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from django_sendfile import sendfile
 from djangorestframework_camel_case.render import CamelCaseJSONRenderer
 from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework.generics import CreateAPIView, DestroyAPIView
+from rest_framework.response import Response
+from rest_framework.views import APIView
 
 from openforms.api.authentication import AnonCSRFSessionAuthentication
 from openforms.api.serializers import ExceptionSerializer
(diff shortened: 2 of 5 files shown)


src/openforms/submissions/api/serializers.py
```diff
@@ -626,3 +626,37 @@ def validate(self, attrs: EmailVerificationData) -> EmailVerificationData:
...
+class VerifyEmailSerializer(EmailVerificationSerializer):
+    code = serializers.CharField(
+        label=_("verification code"),
+        min_length=6,
+        max_length=6,
+        write_only=True,
+    )
+
+    class Meta(EmailVerificationSerializer.Meta):
+        fields = EmailVerificationSerializer.Meta.fields + ("code",)
```
vaszig, line 644: Don't we want to explicitly define-show the fields here?I haven't seen this before for fields that's why I am asking (another question concerning best practices)

src/openforms/submissions/api/views.py
```diff
@@ -96,3 +99,34 @@ def perform_create(self, serializer):
...
+        "Using the code obtained from the verification email, mark the email address "
+        "as verified. Using an invalid code results in validation errors in the error "
+        "response."
+    ),
+)
+class VerifyEmailView(APIView):
+    authentication_classes = (AnonCSRFSessionAuthentication,)
+    permission_classes = (AnyActiveSubmissionPermission,)
+
+    def get_serializer_class(self):
```
vaszig, line 116: Why don't we add this as an attribute of the class and we override the method?I thought that we do that when we want to have more than one serializers depending on a specific situation.

src/openforms/submissions/tests/test_email_verification.py
```diff
@@ -190,3 +190,121 @@ def test_custom_email_subject_and_body(self):
             """,
             message_html,
         )
+
+
+class VerifyEmailTests(SubmissionsMixin, APITestCase):
```
vaszig, line 195: Should we add a test for multiple codes?For example if the user asks for a verification code more than once in the same submission?Or maybe I don't remember right, is that possible anyway?

src/openforms/submissions/api/serializers.py
sergei-maertens: it's the "easiest" way for inheritance. I may regret inheriting the serializer in the future, but then I could easily just define it separately entirely :) this automatically brings along the fields from the parent serializer and keeps things in sync.

src/openforms/submissions/api/views.py
sergei-maertens: for drf-spectacular API documentation reasons :) otherwise it fails to properly process the serializer, because it's not a generic API view

src/openforms/submissions/tests/test_email_verification.py
sergei-maertens: it is possible, I will add a test case :)

src/openforms/submissions/api/views.py
vaszig: Nice, I wasn't aware of this!

vaszig (approved):

PR #4619, opened 2024-08-23
[#4396] Add basic Objects API prefill plugin

### LINKED ISSUE #4396: Create a prefill plugin to prefill with values from an external reference (initially only for Objects API)
- what Objecttype to use

vaszig:
Closes #4396 partially

**Changes**

- Added documentation and the basic prefill plugin for the ObjectsAPI
- Moved ObjectsAPI related api to the contrib module
- Added necessary endpoints for the ObjectsAPI prefill plugin

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


docs/configuration/prefill/objects_api.rst
```diff
@@ -0,0 +1,42 @@
+.. _configuration_prefill_objects_api:
+
+===========
+Objects API
+===========
+
+`Objects API`_ can store objects which are defined (their schema and properties) in the `Objecttypes API`_.
+With this prefill plugin we can have components/form fields which have pre-filled the values taken from the Objects API.
```
sergei-maertens, line 8:
```suggestion
The `Objects API`_ stores data records of which the structure and shape are defined by a particular object type
definition in the `Objecttypes API`_. These records can be used to pre-fill form fields if an object reference is
passed when the form is started.
```

docs/configuration/prefill/objects_api.rst
```diff
@@ -0,0 +1,42 @@
...
+
+.. note::
+
+   This service contains sensitive data and requires a connection to a specific
+   client system. The forms which make use of this prefill plugin require DigiD/Eherkenning authentication
+   of the user.
```
sergei-maertens, line 14:
```suggestion
   The service may contain sensitive data. It is advised to require DigiD/eHerkenning authentication on the form. You
   should be careful with how you pass the object references to the customers and set up the object type in a way that
   makes authentication checks possible (e.g. by storing the expected BSN or KVK number).
```

docs/configuration/prefill/objects_api.rst
```diff
@@ -0,0 +1,42 @@
...
+4. Navigate to: **Variables** tab
+5. Navigate to: **User defined** subtab
+6. Click **Add variable** and fill in the data from the available options:
+
+   * **Plugin**: *Choose the Objects API plugin*
+   * **API Group**: *Choose the desired API Group *
+     (There must be at least one added-configured via **Miscellaneous** > **Objects API groups**)
+   * **Objecttype**: *Based on the selected API group above, you can choose the objecttype from the auto populated list*
+   * **Mappings**: *On the left we have the available form variables and on the right the available attributes which are
+     defined in the selected objecttype's schema properties. You can do the mappings accordingly*
```
sergei-maertens, line 35:
```suggestion
6. Click **Add variable** and fill in the data from the available options:

   * **Plugin**: Choose the *Objects API* plugin
   * **API Group**: Select the appropriate API group. These API groups should be set up by an administrator,
     via **Admin** > **Configuration** > **Prefill plugins** > **Objects API** > **Manage API groups**
   * **Objecttype**: Select the expected object type from the dropdown.
   * **Mappings**: Configure which property from the Objects API record needs to be assigned to which form variable.
     For each form variable you want to pre-fill, add a new mapping. Then, on the left select the desired form variable,
     and on the right you can specify which property from the object type contains the value.
```

This will also require you to add the respective action(s) to the prefill plugin config actions :)

We usually do not refer to the "Miscellaneous" menu in the admin, instead you are supposed to get there via the Configuration overview page.

docs/configuration/prefill/objects_api.rst
sergei-maertens, line 1: Can you add a TODO to indicate this feature is not finished yet:

```rst
.. todo:: This feature is still in development and not ready for production yet.
```

docs/developers/plugins/prefill_plugins.rst
```diff
@@ -25,6 +25,9 @@ You can find an example implementation in :mod:`openforms.prefill.contrib.demo`.
 Implementation
 --------------
 
+Plugins must be added to the INSTALLED_APPS :mod:`openforms.conf.base`. See the demo app as an example
```
sergei-maertens, line 28:
```suggestion
Plugins must be added to the ``INSTALLED_APPS`` setting in :mod:`openforms.conf.base`. See the demo 
app (:class:`openforms.prefill.contrib.demo.apps.DemoApp`) as an example.
```

docs/developers/plugins/prefill_plugins.rst
```diff
@@ -25,6 +25,9 @@ You can find an example implementation in :mod:`openforms.prefill.contrib.demo`.
...
+("openforms.prefill.contrib.demo.apps.DemoApp")
```
sergei-maertens, line 29:
```suggestion
```

src/openforms/prefill/api/urls.py
```diff
@@ -1,8 +1,29 @@
...
+        name="prefill-objects-api-objecttype-version-list",
+    ),
+    path(
+        "plugins/objects-api/objecttypes/<uuid:objects_api_objecttype_uuid>/versions/<int:objects_api_objecttype_version>/attributes",
+        PluginObjectsAPIAttributesListView.as_view(),
+        name="prefill-objects-api-objecttype-attribute-list",
+    ),
     path("plugins", PluginListView.as_view(), name="prefill-plugin-list"),
     path(
         "plugins/<slug:plugin>/attributes",
```
sergei-maertens, line 29: can you please separate these URL patterns in core/shared patterns + plugin specific patterns, just like we do with registrations: 
https://github.com/open-formulieren/open-forms/blob/beacb481e2dc31ec5fc9cfe56c157420b1f42d80/src/openforms/registrations/api/urls.py#L14

And please do make use of the `include` for urls and define the actual plugin URLs/views inside the `objects_api` package.

src/openforms/prefill/api/urls.py
```diff
@@ -1,8 +1,29 @@
...
+)
 
 urlpatterns = [
+    path(
+        "plugins/objects-api/objecttypes",
+        PluginObjectsAPIObjecttypeListView.as_view(),
+        name="prefill-objects-api-objecttype-list",
+    ),
+    path(
+        "plugins/objects-api/objecttypes/<uuid:objects_api_objecttype_uuid>/versions",
```
sergei-maertens, line 18:
```suggestion
        "plugins/objects-api/objecttypes/<uuid:objecttype_uuid>/versions",
```
I find the parameter names a bit too verbose.

src/openforms/prefill/api/urls.py
```diff
@@ -1,8 +1,29 @@
 from django.urls import path
 
-from .views import PluginAttributesListView, PluginListView
+from .views import (
+    PluginAttributesListView,
+    PluginListView,
+    PluginObjectsAPIAttributesListView,
+    PluginObjectsAPIObjecttypeListView,
+    PluginObjectsAPIObjecttypeVersionListView,
```
sergei-maertens, line 8: Please define these views in their plugin package `openforms.prefill.contrib.object_api.api.views`. We should *not* pollute the plugin-agnostic python code/modules with plugin-specific features.

src/openforms/prefill/api/urls.py
```diff
@@ -1,8 +1,29 @@
...
+        PluginObjectsAPIObjecttypeListView.as_view(),
+        name="prefill-objects-api-objecttype-list",
+    ),
+    path(
+        "plugins/objects-api/objecttypes/<uuid:objects_api_objecttype_uuid>/versions",
+        PluginObjectsAPIObjecttypeVersionListView.as_view(),
+        name="prefill-objects-api-objecttype-version-list",
+    ),
+    path(
+        "plugins/objects-api/objecttypes/<uuid:objects_api_objecttype_uuid>/versions/<int:objects_api_objecttype_version>/attributes",
```
sergei-maertens, line 23:
```suggestion
        "plugins/objects-api/objecttypes/<uuid:objecttype_uuid>/versions/<int:version>/properties",
```
Property/properties is the terminology used in JSONSchema which is the standard behind objecttypes API.

src/openforms/prefill/api/views.py
```diff
@@ -6,6 +6,15 @@
 from rest_framework.views import APIView
 
 from openforms.api.views import ListMixin
+from openforms.registrations.contrib.objects_api.api.serializers import (
+    ObjectsAPIGroupInputSerializer,
+)
```
sergei-maertens, line 11: The open forms development guidelines forbid importing stuff from one django app in another django app (with the exception of the `models` and `service` modules). This principle is even stronger when considering Open Forms modules (registrations, prefill, appointments...) and plugins within modules (objects_api, stuf-bg, ...).

The reason for that is that a plugin (and django app) must be as self-contained as possible - the documentation on [RTD](https://open-forms.readthedocs.io/en/stable/developers/backend/index.html) tries to explain this.

In this case, it's better to duplicate the serializer implementation. Yes, they currently do similar/the same things, but that doesn't make them the same. The prefill and registration plugins must be evolvable independently from each other.

If you really want to share this, the serializer should be moved to `openforms.contrib.objects_api` instead, so that it can be re-used by the registrations and prefill plugins, and the hierarchy is clear.

src/openforms/prefill/api/views.py
```diff
@@ -6,6 +6,15 @@
...
+from openforms.registrations.contrib.objects_api.api.views import (
+    OBJECTS_API_GROUP_QUERY_PARAMETER,
+    ObjecttypesListView,
+    ObjecttypeVersionsListView,
+)
```
sergei-maertens, line 16: same remarks as above - in this case I would really move these views to `openforms.contrib.objects_api`

src/openforms/prefill/api/views.py
```diff
@@ -67,3 +76,73 @@ def get_objects(self):
...
+@extend_schema(
+    summary=_("List available objecttype versions for Objects API"),
+    parameters=[OBJECTS_API_GROUP_QUERY_PARAMETER],
+)
+class PluginObjectsAPIObjecttypeVersionListView(ObjecttypeVersionsListView):
+    """
+    List the available prefill objecttype versions for Objects API plugin.
+    """
+
+    def get_objects(self):
```
sergei-maertens, line 102: I don't get why the implementation is overridden here?

src/openforms/prefill/api/views.py
```diff
@@ -67,3 +76,73 @@ def get_objects(self):
...
+            reference={
+                "objects_api_group": config_group,
+                "objects_api_objecttype_uuid": self.kwargs[
+                    "objects_api_objecttype_uuid"
+                ],
+                "objects_api_objecttype_version": self.kwargs[
+                    "objects_api_objecttype_version"
+                ],
+            }
+        )
```
sergei-maertens, line 146: I don't think this needs to be a method on the `Plugin` class.

You want something similar here like `TargetPathsListView` in the registration plugin - all the possible properties from the Object Type JSON schema definition should be returned here so you can select it in a dropdown for the mappings. It's important that the schema definition (or part of it at least) is also returned, so that on the client side you can validate/match up things (e.g. a property in the objecttype that is `type: object` cannot be assigned to a textfield form variable which has `type: string`).

I'm really expecting a list of objects that serialize with https://github.com/open-formulieren/open-forms/blob/beacb481e2dc31ec5fc9cfe56c157420b1f42d80/src/openforms/registrations/contrib/objects_api/api/serializers.py#L40 (`target_path` and `json_schema` are the relevant fields here)

You will need to do JSON schema processing here in the view, but I think you can just re-use https://github.com/open-formulieren/open-forms/blob/beacb481e2dc31ec5fc9cfe56c157420b1f42d80/src/openforms/registrations/contrib/objects_api/api/views.py#L119

The big difference is that we don't have a schema on our end yet to filter down the possible options, so those target paths input serializer usages are not relevant for prefill.

src/openforms/prefill/base.py
```diff
@@ -19,9 +19,14 @@ class BasePlugin(AbstractBasePlugin):
...
     @staticmethod
-    def get_available_attributes() -> Iterable[tuple[str, str]]:
+    def get_available_attributes(
+        reference: dict[str, str] | None = None,
+    ) -> Iterable[tuple[str, str]]:
         """
         Return a choice list of available attributes this plugin offers.
+
+        :param reference: a dict based on which we retrieve the available attributes.
+          Can be used when we have dynamic lists of attributes.
```
sergei-maertens, line 29: please revert this - you will update this interface in another PR, but with the remarks from `src/openforms/prefill/api/views.py`, you don't need to touch this method (yet) in this PR.

src/openforms/prefill/contrib/objects_api/apps.py
```diff
@@ -0,0 +1,12 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class ObjectsApiApp(AppConfig):
+    name = "openforms.prefill.contrib.objects_api"
+    label = "objects_api"
```
sergei-maertens, line 7:
```suggestion
    label = "prefill_objects_api"
```
please namespace django app labels, as they must be unique across the project

src/openforms/prefill/contrib/objects_api/plugin.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+
+PLUGIN_IDENTIFIER = "objects_api"
+
+
+@register(PLUGIN_IDENTIFIER)
+class ObjectsAPIPrefill(BasePlugin):
+    verbose_name = _("Objects API")
+    requires_auth = AuthAttribute.bsn
```
sergei-maertens, line 22: I see this being used for both BSN/KVK authentication methods. Maybe requiring auth at the plugin level is too much, but instead when the prefill runs, we can decide at runtime which auth (BSN or KVK) should be used and validated.

```suggestion
```

src/openforms/prefill/contrib/objects_api/plugin.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @staticmethod
+    def get_available_attributes(
+        reference: dict[str, Any] | None = None,
+    ) -> Iterable[tuple[str, str]]:
+        pass
```
sergei-maertens, line 28: let's revert this interface but leave it as a `raise NotImplementedError("not yet")` here

sergei-maertens (changes requested):

src/openforms/prefill/api/views.py
vaszig: It was weird and I forgot to ask this in this PR. I saw that the same view in the registrations uses `submission_uuid` which I don't get and that's the only reason I did it.

https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/objects_api/api/views.py#L91

In the client:
https://github.com/open-formulieren/open-forms/blob/master/src/openforms/contrib/objects_api/clients/objecttypes.py#L56

vaszig: Discussed this and it is a copy paste mistake, so the name will be updated accordingly.

src/openforms/prefill/api/urls.py
```diff
@@ -10,3 +10,10 @@
         name="prefill-attribute-list",
     ),
 ]
+
+urlpatterns += [
```
sergei-maertens, line 14: can you add the comment about registering plugin-specific endpoints and the TODO that this should be made properly dynamic, same as with the registrations module?

src/openforms/prefill/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    List the available prefill objecttype versions for Objects API plugin.
+    """
+
+
+@extend_schema(
+    summary=_("List available attributes for Objects API"),
+    parameters=[OBJECTS_API_GROUP_QUERY_PARAMETER],
+)
+class PrefillObjectsAPIAttributesListView(views.APIView):
```
sergei-maertens, line 56:
```suggestion
class ObjecttypePropertiesListView(views.APIView):
```
we're already in the `prefill.contrib.objects_api` package, so we can omit that information and instead be more specific: we're retrieving a list of available JSON properties defined on a particular objecttype

src/openapi.yaml
```diff
@@ -3451,6 +3451,207 @@ paths:
...
+          headers:
+            X-Session-Expires-In:
+              $ref: '#/components/headers/X-Session-Expires-In'
+            X-CSRFToken:
+              $ref: '#/components/headers/X-CSRFToken'
+            X-Is-Form-Designer:
+              $ref: '#/components/headers/X-Is-Form-Designer'
+            Content-Language:
+              $ref: '#/components/headers/Content-Language'
+  /api/v2/objects-api/target-paths:
```
sergei-maertens, line 3614: I think this endpoint should remain specific to the objects API registration plugin, because it requires the particular input prameters.

src/openforms/contrib/objects_api/api/views.py
```diff
@@ -88,7 +88,7 @@ def get_objects(self) -> list[dict[str, Any]]:
         config_group = input_serializer.validated_data["objects_api_group"]
 
         with get_objecttypes_client(config_group) as client:
-            return client.list_objecttype_versions(self.kwargs["submission_uuid"])
+            return client.list_objecttype_versions(self.kwargs["objecttype_uuid"])
 
 
 class TargetPathsListView(views.APIView):
```
sergei-maertens, line 94: so to address the comment on the API schema file:

this view belongs in `openforms.registrations.contrib.objects_api.api.views`, since it's specific to the registration plugin - we can't re-use it for the prefill plugin.

src/openforms/prefill/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,93 @@
...
+    location=OpenApiParameter.QUERY,
+    description=_("Which Objects API group to use."),
+)
+
+
+@extend_schema(
+    summary=_("List available objecttypes for Objects API"),
+    parameters=[OBJECTS_API_GROUP_QUERY_PARAMETER],
+)
+class PrefillObjectsAPIObjecttypeListView(ObjecttypesListView):
```
sergei-maertens, line 36: IMO this should use the shared view/endpoint from the `openforms.contrib.objects_api` package and should not be redefined here

src/openforms/prefill/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    List the available prefill objecttypes for Objects API plugin.
+    """
+
+
+@extend_schema(
+    summary=_("List available objecttype versions for Objects API"),
+    parameters=[OBJECTS_API_GROUP_QUERY_PARAMETER],
+)
+class PrefillObjectsAPIObjecttypeVersionListView(ObjecttypeVersionsListView):
```
sergei-maertens, line 46: IMO this should use the shared view/endpoint from the `openforms.contrib.objects_api` package and should not be redefined here

src/openforms/prefill/contrib/objects_api/api/views.py
```diff
@@ -0,0 +1,93 @@
...
+    parameters=[OBJECTS_API_GROUP_QUERY_PARAMETER],
+)
+class PrefillObjectsAPIAttributesListView(views.APIView):
+    """
+    List the available attributes for Objects API plugin.
+    """
+
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+    serializer_class = PrefillTargetPathsSerializer
```
sergei-maertens, line 63: this is now incorrectly picked up by the API schema generation tooling and doesn't document that it's a list of these kind of objects (because of the `many=True)`, so the implementation is correct, but the documentation isn't. I think the endpoint for the registrations may suffer from the same bug.

Either way - I'd expect this view to use the `ListMixin` from `openforms.api.views` and only implement `get_objects` instead of implementing the full `def get(...)` method. If you do that, then the API documentation will also be picked up correctly.

src/openforms/registrations/contrib/objects_api/tests/test_json_schema.py
sergei-maertens, line 1: shouldn't these tests be moved to `openforms.contrib.objects_api` instead of just updating the imports?

src/openforms/registrations/contrib/objects_api/tests/test_objecttypes_client.py
sergei-maertens, line 1: shouldn't these tests be moved to `openforms.contrib.objects_api` instead of just updating the imports?

sergei-maertens (commented):
Let's go through this together on Wednesday

src/openforms/prefill/contrib/objects_api/plugin.py
```diff
@@ -0,0 +1,39 @@
+import logging
+
+from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
+
+from openforms.contrib.objects_api.checks import check_config
+from openforms.registrations.contrib.objects_api.models import ObjectsAPIConfig
```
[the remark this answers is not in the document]
vaszig: Another problem..Should we do the same for this model?Move it to the contrib?I see that it doesn't have any FKs but it does affect old migration tests which make use of the following:
- [https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/objects_api/models.py#L11](url) 
- [https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/objects_api/models.py#L17](url)

vaszig:
Rebased  (after moving `ObjectsAPIGroupConfig` to contrib) and processed feedback. Now, there is one similar problem:

https://github.com/open-formulieren/open-forms/pull/4619#discussion_r1749966296

src/openforms/prefill/contrib/objects_api/plugin.py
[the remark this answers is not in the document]
sergei-maertens: let's not, but document that this is scheduled for removal.

it's a legacy model :)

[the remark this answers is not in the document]
vaszig: Yeah, but I do import the model again from the registrations.Should we leave it as is because of your comment?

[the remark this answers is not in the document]
sergei-maertens: yes, leave it as is :)

sergei-maertens (approved):
Let's get this merged and included in the beta.

PR #4620, opened 2024-08-26
[#4396] Objects API prefill plugin

### LINKED ISSUE #4396: Create a prefill plugin to prefill with values from an external reference (initially only for Objects API)

vaszig:
Closes #4396 partially 

**Changes**

- Added necessary model field `prefill_options`  to the `FormVariable` model (for storing all the needed configuration) and updated constraints according to new field
- Moved public prefill functions to `prefill.service`
- Added prefill functionality for ObjectsAPI

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/formio/tests/test_component_translations.py
```diff
@@ -28,7 +28,10 @@ def disable_prefill_injection():
     """
     Disable prefill to prevent prefill-related queries.
     """
-    return patch("openforms.formio.service.inject_prefill", new=MagicMock)
+    return patch(
+        "openforms.prefill.service.inject_prefill",
+        new=MagicMock,
+    )
```
sergei-maertens, line 34: isn't this broken now because of the local import?

src/openforms/forms/models/form_variable.py
```diff
@@ -245,7 +276,29 @@ def check_data_type_and_initial_value(self):
...
+
+        mappings = self.prefill_options["variables_mapping"]
+        form_variable_keys = [item["variable_key"] for item in mappings]
+        form_variables_for_update = FormVariable.objects.filter(
+            key__in=form_variable_keys
+        )
+        for variable in form_variables_for_update:
+            variable.prefill_plugin = self.prefill_plugin
+
+        FormVariable.objects.bulk_update(form_variables_for_update, ["prefill_plugin"])
```
sergei-maertens, line 293: this seems wrong - `prefill_options` does not have a fixed structure, the shape and data inside can only be handled by the respective plugin.

I think I understand what you're trying to do here - make sure that the mapped variables get their data from the retrieved object even though no explicit prefill configuration was assigned to them. However, I think that logic belongs in the prefill plugin (objects API) implementation itself.

I now also understand the check constraint above.

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -16,6 +24,64 @@
...
+        label=_("objecttype version"),
+        required=False,
+        help_text=_("Version of the objecttype in the Objecttypes API."),
+    )
+    variables_mapping = ObjecttypeVariableMappingSerializer(
+        label=_("variables mapping"),
+        many=True,
+        required=False,
+    )
+
```
sergei-maertens, line 83: yeah so this is too naive :D while it achieves what we want in the ticket, it's really something that must be defined in the objects API prefill plugin itself. It's a very similar thing like the registrations plugin `configuration_options` serializer, which must be defined on the plugin itself and called dynamically during validation.

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -110,6 +176,7 @@ class FormVariableSerializer(serializers.HyperlinkedModelSerializer):
     service_fetch_configuration = ServiceFetchConfigurationSerializer(
         required=False, allow_null=True
     )
+    prefill_options = FormVariableOptionsSerializer(required=False)
```
sergei-maertens, line 179: at this level, this must remain a plain `JSONField`, but we should add validators that it must always be a dict (so list, int, string etc. JSON values are not allowed).

In the `def validate(...)` method we can/must then look up the specific validation options based on the value of the `plugin` field so that we can perform deep validation. You can take a look at `FormRegistrationBackendSerializer` for inspiration, it does the same from the registration plugin perspective

src/openforms/prefill/base.py
```diff
@@ -50,6 +51,26 @@ def get_prefill_values(
...
+    ) -> dict[str, str]:
+        """
+        Given the saved form variable, which contains the prefill_options, look up the appropriate
+        values and return them.
+
+        :param submission: an active :class:`Submission` instance, which can supply
+          the required initial data reference to fetch the correct prefill values.
+        :param form_variable: The form variable for which we want to retrieve the data. Its
+        atribute prefill_options contains all the mappings that are needed for retrieving
+        and returning the values.
```
sergei-maertens, line 66:
```suggestion
        :param form_variable: The form variable for which we want to retrieve the data. Its
          attribute :attr:`FormVariable.prefill_options` contains plugin-specific configuration
          options.
```

src/openforms/prefill/base.py
```diff
@@ -50,6 +51,26 @@ def get_prefill_values(
...
+        :return: a key-value dictionary, where the key is the mapped property and
+          the value is the prefill value to use for that property.
```
sergei-maertens, line 68:
```suggestion
        :return: a mapping where the keys are form variable keys, and the values are the 
          initial/default values to assign to the matching form variable. The variable keys
          can point to both component and user defined variables.
```
I think this is how it's meant?

src/openforms/prefill/contrib/objects_api/plugin.py
```diff
@@ -20,6 +26,30 @@
...
 
+    @classmethod
+    def get_prefill_values_from_mappings(
+        cls,
+        submission: Submission,
+        form_variable: FormVariable,
+    ) -> dict[str, str]:
+        variables_mappings = form_variable.prefill_options.get("variables_mapping")
+        config_group_id = form_variable.prefill_options.get("objects_api_group")
+        config_group = ObjectsAPIGroupConfig.objects.get(id=config_group_id)
```
sergei-maertens, line 37: this can have quite some runtime crashes if the data in `prefill_options` is not as expected, especially when using `.get` where you really expect the config option to always be provided.

Take a look at `openforms.registrations.base.BasePlugin`, particularly with regard to the `Generic[OptionsT]` - you can pass a `TypedDict` what the configuration options dict is supposed to look like. A concrete example is `openforms.registrations.contrib.zgw_apis.plugin.ZGWRegistration.register_submission`, where you have a properly typed `options` argument.

I think applying a similar interface also makes more sense - why should the plugin do the lookup of `prefill_options` when the generic tooling can already provide that? So as a method signature I'd rather expect:

```py
class Options(TypedDict):
    pass


class BasePlugin[OptionsT: Options](AbstractBasePlugin):
    def get_prefill_values_from_options(
        cls, submission: Submission, options: OptionsT,
    ) -> Mapping[str, str]: ...
```

This defines the `BasePlugin` to have a type var `OptionsT` which is bound by `Options` (basically saying it *must* be a dict).

Your concrete plugin would then have:

```py
class VariableMapping(TypedDict):
    variable_key: str
    target_path: list[str]

class ObjectsAPIOptions(TypedDict):
    objects_api_group: ObjectsAPIGroupConfig
    variables_mapping: list[VariableMapping]


class Plugin(BasePlugin[ObjectsAPIOptions]):
    ...
```

The beauty here is that when you specify the serializer on the plugin class too, that you can use that already turn the JSON options into their appropriate python/django objects, [just like the registration plugins](https://github.com/open-formulieren/open-forms/blob/d0a504f523def6226ac1433a30f315dfb8523642/src/openforms/registrations/tasks.py#L277).

src/openforms/forms/models/form_variable.py
```diff
@@ -197,10 +208,30 @@ class Meta:
...
+                        & EMPTY_PREFILL_ATTRIBUTE
+                        & ~EMPTY_PREFILL_OPTIONS
+                        & USER_DEFINED
+                    )
+                    | (
+                        ~EMPTY_PREFILL_PLUGIN
+                        & EMPTY_PREFILL_ATTRIBUTE
+                        & EMPTY_PREFILL_OPTIONS
+                        & ~USER_DEFINED
+                    )
```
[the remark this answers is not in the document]
vaszig: This will remain (if I am not mistaken) because of the comment below
https://github.com/open-formulieren/open-forms/pull/4620#discussion_r1796732872

src/openforms/formio/tests/test_component_translations.py
vaszig: I don't see a problem here in the tests and if I remove this the tests are failing.

sergei-maertens: I still don't understand why this works but I'll accept it :P

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -177,18 +180,47 @@ def validate(self, attrs):
...
-            not prefill_plugin and prefill_attribute
-        ):
+        prefill_options = attrs.get("prefill_options")
+
+        if prefill_plugin and prefill_options and prefill_attribute:
             raise ValidationError(
                 {
-                    "prefill_attribute": _(
-                        "Prefill plugin and attribute must both be specified."
+                    "prefill_attribute_options": _(
```
sergei-maertens, line 192:
```suggestion
                    "prefill_attribute": _(
```
either attach it to one particular field, or just `raise ValidationError(message)` to make it a non-field error

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -177,18 +180,47 @@ def validate(self, attrs):
...
+                        "Prefill plugin, attribute and options can not be specified at the same time."
                     ),
                 }
             )
 
+        if source == FormVariableSources.component:
+            if prefill_options:
+                raise ValidationError(
+                    {
+                        "component_prefill_attribute": _(
```
sergei-maertens, line 202:
```suggestion
                        "prefill_options": _(
```

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -177,18 +180,47 @@ def validate(self, attrs):
...
+                            "Prefill options should not be specified for component variables."
+                        ),
+                    }
+                )
+            if not prefill_plugin and prefill_attribute:
```
sergei-maertens, line 207: this case should not be limited to component variables - and I think that's the original one. When no prefill options are specified, you need to ensure that both plugin/attribute are set or not set at the same time. It's possible to prefill a user defined variable from plugin + attribute without the options.

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -177,18 +180,47 @@ def validate(self, attrs):
...
+                raise ValidationError(
+                    {
+                        "component_prefill_attribute": _(
```
sergei-maertens, line 210: keep in mind that these fields must map to serializer fields for the validation errors to properly show up!

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -177,18 +180,47 @@ def validate(self, attrs):
...
+                )
+
+        # check the specific validation options of the prefill plugin
+        if prefill_plugin:
+            plugin = register[prefill_plugin]
+            serializer = plugin.options(data=prefill_options)
+            try:
+                serializer.is_valid(raise_exception=True)
+            except serializers.ValidationError as e:
+                detail = {"options": e.detail}
```
sergei-maertens, line 223:
```suggestion
                detail = {"prefill_options": e.detail}
```
needs to map to the serializer field name

src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -177,18 +180,47 @@ def validate(self, attrs):
...
+                raise ValidationError(
+                    {
+                        "component_prefill_attribute": _(
+                            "Prefill attribute cannot be specified without prefill plugin for component variables."
+                        ),
+                    }
+                )
+
+        # check the specific validation options of the prefill plugin
+        if prefill_plugin:
```
sergei-maertens, line 217:
```suggestion
        if prefill_plugin and prefill_options is not None:
```
should be we not only validate this if options are provided?

src/openforms/prefill/base.py
```diff
@@ -14,9 +17,22 @@ def __contains__(self, thing):
...
+
+class Options(TypedDict):
+    pass
+
+
+SerializerCls = type[serializers.Serializer]
+OptionsT = TypeVar("OptionsT", bound=Options)
+
+
+class BasePlugin(Generic[OptionsT], AbstractBasePlugin):
```
sergei-maertens, line 32: I think with the new Python 3.12 syntax we can do this instead:

```suggestion
class BasePlugin[OptionsT: Options](AbstractBasePlugin):
```

src/openforms/prefill/contrib/objects_api/api/serializers.py
```diff
@@ -1,7 +1,13 @@
+from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
 
+from openforms.api.fields import PrimaryKeyRelatedAsChoicesField
+from openforms.formio.api.fields import FormioVariableKeyField
+from openforms.registrations.contrib.objects_api.models import ObjectsAPIGroupConfig
```
sergei-maertens, line 8:
```suggestion
from openforms.contrib.objects_api.models import ObjectsAPIGroupConfig
```

src/openforms/prefill/contrib/objects_api/api/serializers.py
```diff
@@ -15,3 +21,50 @@ class PrefillTargetPathsSerializer(serializers.Serializer):
...
+    )
+
+
+class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
+    objects_api_group = PrimaryKeyRelatedAsChoicesField(
+        queryset=ObjectsAPIGroupConfig.objects.exclude(
+            Q(objects_service=None)
+            | Q(objecttypes_service=None)
+            | Q(drc_service=None)
+            | Q(catalogi_service=None)
```
sergei-maertens, line 50: I don't think that Documents API/Catalogi API are used at all in the prefill layer, so these filters can be removed. We only need a working Object Types API and Objects API to perform the actual lookup at some point.

src/openforms/prefill/contrib/objects_api/plugin.py
```diff
@@ -3,22 +3,50 @@
...
+            obj = client.get_object(submission.initial_data_reference)
+
+        obj_record = obj.get("record", {})
+        obj_record_data = glom(obj, "record.data")
+
+        results = {}
+        for mapping in options["variables_mapping"]:
+            path = Path(*mapping["target_path"])
+            if value := find_in_dicts(obj_record, obj_record_data, path=path):
+                results[mapping["variable_key"]] = value
```
sergei-maertens, line 47: I don't really see the need for `find_in_dicts` if we have `glom`?

```py
try:
    value = glom(obj_record_data, path)
except PathAccessError:
    continue
results[mapping["variable_key"]] = value
```
should do the trick, no?

src/openforms/prefill/contrib/objects_api/tests/test_prefill.py
```diff
@@ -0,0 +1,287 @@
...
+                "variables_mapping": [
+                    {"variable_key": "lastName", "target_path": ["name", "last.name"]},
+                    {"variable_key": "age", "target_path": ["age"]},
+                ],
+            },
+        )
+
+        prefill_variables(submission=submission)
+
+        # Update the prefilled data
```
sergei-maertens, line 251: I probably misunderstood you when you asked a question about registration and that resulted in this part of the test being added. Let's not test that here, sorry for the confusion!

src/openforms/prefill/service.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+    ):
+        prefill_data.update(**user_defined_results)
+
+    total_config_wrapper = submission.total_configuration_wrapper
+    for variable_key, prefill_value in prefill_data.items():
+        component = total_config_wrapper[variable_key]
+        normalized_prefill_value = normalize_value_for_component(
+            component, prefill_value
+        )
```
sergei-maertens, line 155: since `prefill_data` includes user defined variables, do we have guards here for missing `component` for a given key?

src/openforms/prefill/sources/user_defined.py
```diff
@@ -0,0 +1,47 @@
...
+    SubmissionValueVariable,
+)
+from openforms.typing import JSONEncodable
+
+from ..registry import Registry
+
+logger = logging.getLogger(__name__)
+
+
+def fetch_prefill_values(
```
sergei-maertens, line 16: does this still work if I have a user defined variable configured with plugin + attribute rather than plugin + options?

src/openforms/prefill/utils.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any
+
+from glom import Path, PathAccessError, glom
+
+
+def find_in_dicts(*dicts: dict[str, Any], path: Path) -> str | None:
```
sergei-maertens, line 6: this return type is also wrong - the JSON can contain `null` or `number`, and that is what would be returned. So I think you can only say `JSONValue` as type annotation

Furthermore, using `None` to signify a missing value is ambiguous, since `null` in JSON can be defined but that will be returned as `None` and then ignored. It's better to rely on the glom exception here or use a special sentinel value to signify the key is missing.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -174,15 +174,20 @@ def get_prefill_variables(self) -> list[SubmissionValueVariable]:
...
+        variables = self.variables.values()
+        for variable in list(variables):
             if variable.key not in data:
                 continue
 
             variable.value = data[variable.key]
             variable.source = SubmissionValueVariableSources.prefill
 
-        SubmissionValueVariable.objects.bulk_create(variables_to_prefill)
+        SubmissionValueVariable.objects.bulk_create(variables)
```
sergei-maertens, line 190: this also saves the variables for which the key was not in `data` - maybe we should keep a list of `variables_to_create` and only add them in there if there's actually something to fill? That's closer to the original behaviour while still fulfilling the new requirements.

src/openforms/submissions/tests/test_start_submission.py
```diff
@@ -203,11 +199,20 @@ def test_start_submission_bad_form_url(self):
...
+            form_definition__configuration={
+                "components": [
+                    {
+                        "type": "textfield",
+                        "key": "test-key",
+                        "label": "Test label",
+                        "prefill": {"plugin": "demo", "attribute": "random_string"},
+                    }
+                ]
+            },
```
sergei-maertens, line 215: does this result in the same variable being created?

src/openforms/forms/models/form_variable.py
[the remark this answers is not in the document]
vaszig: This was removed as there is no need to update the component separately with the plugin, so there is no need for this constraint.

src/openforms/prefill/utils.py
vaszig: This was removed as we decided to use `glom` for it.

src/openforms/submissions/tests/test_start_submission.py
vaszig: I realized that in the way we test things here (getting the submission variables by `SubmissionValueVariable.objects.filter(submission=submission)` )gives us one submission variable when the form variables are two. I would do it with  `submission.load_submission_value_variables_state()` tbh which gives 2. I am a bit confused here..

src/openforms/prefill/service.py
```diff
@@ -0,0 +1,188 @@
...
+        except KeyError:
+            # The component to prefill is not in this step
+            continue
+
+        if not (prefill := component.get("prefill")):
+            continue
+        if not prefill.get("plugin"):
+            continue
+        if not prefill.get("attribute"):
+            continue
```
sergei-maertens, line 90: Lines were not covered before the code was moved, accepting this as-is and I'm ignoring the codecov annotations.

src/openforms/prefill/service.py
```diff
@@ -0,0 +1,188 @@
...
+    ):
+        prefill_data.update(**results_from_options)
+
+    total_config_wrapper = submission.total_configuration_wrapper
+    for variable_key, prefill_value in prefill_data.items():
+        if key_source_mappings[variable_key] == FormVariableSources.component:
+            try:
+                component = total_config_wrapper.component_map[variable_key]
+            except KeyError:
+                return None
```
sergei-maertens: This looks a bit weird - if something is wrong we seem to discard everything. Does that make sense? The lack of coverage on the error handling also makes this seem like a what-if.

This could occur if someone manually creates a `FormVariable` in the admin/shell without it being backed by an actual form definition component, which shouldn't happen. The API endpoint synchronizes the components and derived variables, so I'm leaning towards removing the try-except and if it crashes, it crashes and at least it will become visible in Sentry instead of failing silently, no?

In the old situation this wouldn't happen anyway because we looped only over the prefill variables.

src/openforms/submissions/tests/test_start_submission.py
sergei-maertens: I traced back why the additional variable was created and patched that behaviour, so the test changes are reverted too :)

sergei-maertens (approved):

PR #4633, opened 2024-09-04
[#4396] Prefill variables by using ObjectsAPI

### LINKED ISSUE #4396: Create a prefill plugin to prefill with values from an external reference (initially only for Objects API)

vaszig:
Closes #4396 partially,
Needs #4619, #4620

**Changes**

- Added basic variables prefill functionality
- Modified prefill plugin to comply ObjectsAPI requirements

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Commit hygiene


vaszig:
This was initially created without having decided that ObjectsAPIGroupConfig would be restructured (moved to contrib).So, now PR 4620 is not that big and this PR can be part of it.

PR #4638, opened 2024-09-06
ZGW Registration UX improvements: zaaktype dropdown

### LINKED ISSUE #4606: Make ZGW registration options consistent with Objects API

sergei-maertens:
Closes #4606 (partially

**Changes**

* [x] Added API endpoint to look up  zaaktypen (closes #4498)
* [x] Added dropdown to select zaaktype and store identification instead
* [x] Validate document type agains case type reference (instead of URL/catalogue)
* [x] Validate property mappings/properties against the selected case type reference (instead of URL)
* [x] Validate roltype (medewerker) against the selected case type reference (instead of URL)
* [x] Dynamically select correct case type version during registration if case type identification is set

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


sergei-maertens:
@vaszig given the size of this PR, it might be better to go commit by commit for the review. Sorry, it was a lot more work than I expected :(

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.js
```diff
@@ -106,6 +107,17 @@ const ZGWFormFields = ({name, apiGroupChoices, confidentialityLevelChoices}) =>
             >
               <CatalogueSelect />
             </ErrorBoundary>
+            <ErrorBoundary
+              errorMessage={
+                <FormattedMessage
+                  description="ZGW APIs registrations options: case type select error"
+                  defaultMessage={`Something went wrong retrieving the available case types.
```
vaszig: Should we add _`while`_ here?

setup.cfg
```diff
@@ -61,4 +61,4 @@ exclude_also =
     raise NotImplementedError
     raise RuntimeError
     \.\.\.
-    pass
+    pass$
```
vaszig, line 64: Is this intended?

src/openforms/js/components/admin/form_design/registrations/zgw/fields/CatalogueSelect.js
```diff
@@ -1,45 +1,15 @@
...
-    value: catalogueOptionGroups = [],
-    error: cataloguesError,
-  } = useAsync(async () => {
-    if (!zgwApiGroup) return [];
-    return await getCatalogues(zgwApiGroup);
-  }, [zgwApiGroup]);
-  if (cataloguesError) throw cataloguesError;
+import {CatalogueSelect as GenericCatalogueSelect} from 'components/admin/forms/zgw';
 
+const CatalogueSelect = ({loading, optionGroups}) => {
```
vaszig, line 7: This component is missing the help text

src/openforms/registrations/utils.py
```diff
@@ -7,13 +7,15 @@
 unset = object()
 
 
-def execute_unless_result_exists(
-    callback: Callable,
+def execute_unless_result_exists[
```
vaszig, line 10: Could you please elaborate a bit?I haven't seen this structure

bin/locate_vcr_cassettes.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
sergei-maertens: yeah, makes it easier to locate them when re-recording cassettes :) otherwise I keep going through my shell history to find them

src/openforms/authentication/tasks.py
```diff
@@ -20,5 +22,6 @@ def update_saml_metadata() -> None:
     `digid_eherkenning.management.commands` module of django-digid-eherkenning library.
     Updates the stored metadata file and prepopulates the db fields.
     """
-    call_command("update_stored_metadata", "digid")
-    call_command("update_stored_metadata", "eherkenning")
+    stdout = StringIO()
```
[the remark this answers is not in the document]
sergei-maertens: I'm using this to suppress the output in tests - when running the full test suite locally I don't want do say anything else other than `.`, `E`, `S` and `F`.

Otherwise we indeed do not store the result in the celery result store (as that requires you to clean it up at some point), but it's really just for the test output.

setup.cfg
sergei-maertens: yeah, otherwise it was also skipping coverage checks on lines that contain the word  `password` or similar, which was wrong. So now it's only matching lines that end in `pass`, so that should cover the

```py
def foo():
    pass
```

cases

src/openforms/js/components/admin/form_design/registrations/zgw/fields/CatalogueSelect.js
sergei-maertens: there is none for the objects API registration either: https://open-formulieren.github.io/open-forms/?path=/story/form-design-registration-objects-api--default

I can't think of a help text that's useful that doesn't just repeat the form label. Something like "Select the catalogue" is stating the obvious, so IMO it's better to not specify one.

src/openforms/registrations/utils.py
sergei-maertens: it's new syntax available in python 3.12!

You can specify the generic typevar between square brackets now, instead of having to do:

```py
R = TypeVar("R")

def execute_unless_result_exists(callback: Callable[[], R]) -> R:
    ...

src/openforms/js/components/admin/form_design/registrations/zgw/fields/CatalogueSelect.js
vaszig: Yes, it's obvious indeed, the only reason I mentioned that was because of possible connected matters like accessibility

PR #4675, opened 2024-09-17
Use react-select in variable dropdown for better UX

### LINKED ISSUE #4524: Make a React dropdown out of the choose form variable-dropdown
### Thema / Theme

Frontend

Separated from #4519 

_No response_

_No response_

robinmolen:
Closes #4524 

**Changes**

The variable dropdown now uses the react-select component. Allowing for searching within the options

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/forms/ReactSelect.js
+++ b/src/openforms/js/components/admin/forms/ReactSelect.js
@@ -2,6 +2,7 @@ import {getReactSelectStyles} from '@open-formulieren/formio-builder/esm/compone
 import {useField} from 'formik';
 import PropTypes from 'prop-types';
 import ReactSelect from 'react-select';
+import classNames from 'classnames';
 
 const initialStyles = getReactSelectStyles();
 const styles = {
@@ -32,18 +33,51 @@ const styles = {
 };
 
 /**
- * A select dropdown backed by react-select for Formik forms.
+ * A select dropdown backed by react-select for legacy usage.
  *
- * Any additional props are forwarded to the underlyng ReactSelect component.
+ * Any additional props are forwarded to the underlying ReactSelect component.
+ *
+ * @deprecated - if possible, refactor the form to use Formik and use the Formik-enabled
+ * variant.
  */
-const Select = ({name, options, ...props}) => {
-  const [fieldProps, , fieldHelpers] = useField(name);
-  const {value} = fieldProps;
-  const {setValue} = fieldHelpers;
+const SelectWithoutFormik = ({name, options, value, className, onChange, ...props}) => {
+  const classes = classNames("admin-react-select", {
+    [`${className}`]: className
+  });
   return (
     <ReactSelect
       inputId={`id_${name}`}
-      className="admin-react-select"
+      name={name}
+      className={classes}
+      classNamePrefix="admin-react-select"
+      styles={styles}
+      menuPlacement="auto"
+      options={options}
+      value={options.find(opt => opt.value === value) || null}
+      onChange={selectedOption => {
+        onChange(selectedOption === null ? undefined : selectedOption.value);
+      }}
+      {...props}
+    />
+  );
+};
+
+/**
+ * A select dropdown backed by react-select for Formik forms.
+ *
+ * Any additional props are forwarded to the underlying ReactSelect component.
+ */
+const SelectWithFormik = ({name, options, className, ...props}) => {
+  const [fieldProps, , fieldHelpers] = useField(name);
+  const {value} = fieldProps;
+  const {setValue} = fieldHelpers;
+  const classes = classNames("admin-react-select", {
+    [`${className}`]: className
+  });
+  return (
+    <ReactSelect
+      inputId={`id_${name}`}
+      className={classes}
       classNamePrefix="admin-react-select"
       styles={styles}
       menuPlacement="auto"
@@ -63,9 +97,15 @@ const Select = ({name, options, ...props}) => {
   );
 };
 
-Select.propTypes = {
+SelectWithoutFormik.propTypes = {
   name: PropTypes.string.isRequired,
   options: PropTypes.arrayOf(PropTypes.object),
 };
 
-export default Select;
+SelectWithFormik.propTypes = {
+  name: PropTypes.string.isRequired,
+  options: PropTypes.arrayOf(PropTypes.object),
+};
+
+export default SelectWithFormik;
+export {SelectWithFormik, SelectWithoutFormik};
--- a/src/openforms/js/components/admin/forms/ReactSelect.js
+++ b/src/openforms/js/components/admin/forms/ReactSelect.js
@@ -1,8 +1,8 @@
 import {getReactSelectStyles} from '@open-formulieren/formio-builder/esm/components/formio/select';
+import classNames from 'classnames';
 import {useField} from 'formik';
 import PropTypes from 'prop-types';
 import ReactSelect from 'react-select';
-import classNames from 'classnames';
 
 const initialStyles = getReactSelectStyles();
 const styles = {
@@ -32,6 +32,32 @@ const styles = {
   }),
 };
 
+const getValueRecursively = (options, value) => {
+  if (!value) {
+    return null;
+  }
+
+  const option = options.find(opt => opt.value === value);
+  if (option) {
+    return option;
+  }
+
+  // Search for the option recursively
+  for (let i = 0; i < options.length; i++) {
+    const opt = options[i];
+    if (!opt.options) {
+      continue;
+    }
+
+    const option = getValueRecursively(opt.options, value);
+    if (option) {
+      return option;
+    }
+  }
+
+  return null;
+};
+
 /**
  * A select dropdown backed by react-select for legacy usage.
  *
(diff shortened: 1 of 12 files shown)


src/openforms/js/components/admin/forms/ReactSelect.js
```diff
@@ -31,19 +31,43 @@ const styles = {
   }),
 };
 
+/**
+ * A select dropdown backed by react-select for regular usage.
+ *
+ * Any additional props are forwarded to the underlying ReactSelect component.
+ */
```
sergei-maertens, line 38:
```suggestion
/**
 * A select dropdown backed by react-select for legacy usage.
 *
 * Any additional props are forwarded to the underlying ReactSelect component.
 *
 * @deprecated - if possible, refactor the form to use Formik and use the Formik-enabled
 * variant.
 */
```

Our ambition is to manage all form state via Formik - not a reality yet, but let's make clear that this is an escape hatch :)

src/openforms/js/components/admin/forms/ReactSelect.js
```diff
@@ -31,19 +31,43 @@ const styles = {
...
-const Select = ({name, options, ...props}) => {
+const SelectWithFormik = ({name, options, className, ...props}) => {
   const [fieldProps, , fieldHelpers] = useField(name);
   const {value} = fieldProps;
   const {setValue} = fieldHelpers;
   return (
     <ReactSelect
       inputId={`id_${name}`}
-      className="admin-react-select"
+      className={`admin-react-select ${className || ''}`}
```
sergei-maertens, line 70: preferably use `import classNames from 'classnames';` for dynamic class name building. In the typescript projects, this is the `clsx` library instead :)

src/openforms/js/components/admin/forms/ReactSelect.js
```diff
@@ -63,9 +87,15 @@ const Select = ({name, options, ...props}) => {
   );
 };
 
-Select.propTypes = {
+SelectWithoutFormik.propTypes = {
+  name: PropTypes.string.isRequired,
+  options: PropTypes.arrayOf(PropTypes.object),
+};
```
sergei-maertens, line 93: I think we should make `value` and `onChange` required props

src/openforms/js/components/admin/forms/VariableSelection.js
```diff
@@ -30,20 +30,16 @@ const VariableSelection = ({
       const label = formDefinitionsNames[variable.formDefinition]
         ? `${formDefinitionsNames[variable.formDefinition]}: ${variable.name} (${variable.key})`
         : `${variable.name} (${variable.key})`;
-      return [variable.key, label];
+      return {value: variable.key, label};
```
sergei-maertens, line 33: This doesn't appear to implement the `label` JSX with nicer markup/styling yet, right? Something like:

```jsx
<div className="variable-option">
    <span className="variable-option__name">{variable.name}<span>
    <span className="variable-option__key">({variable.key}}<span>
    {formDefinitionsNames[variable.formDefinition] && <span className="variable-option__context">{formDefinitionsNames[variable.formDefinition]}<span>}
</div>
```

src/openforms/js/components/admin/forms/VariableSelection.js
```diff
@@ -30,20 +30,16 @@ const VariableSelection = ({
...
   return (
-    <Select
+    <SelectWithoutFormik
       id={id}
       className="form-variable-dropdown"
       name={name}
-      choices={choices}
-      allowBlank
-      onChange={onChange}
+      options={choices}
```
sergei-maertens, line 41: How should we approach grouping the choices into optgroups by type of variable: component / user defined / static?

src/openforms/scss/components/admin/_select.scss
```diff
@@ -1,3 +1,4 @@
 .form-variable-dropdown {
+  min-inline-size: var(--of-admin-select-min-inline-size);
   max-inline-size: var(--of-admin-select-max-inline-size);
```
sergei-maertens, line 3: we could use the modern `clamp` property here: `inline-size: clamp($min, 100%, $max);` perhaps?

sergei-maertens (changes requested):

src/openforms/js/components/admin/forms/ReactSelect.js
robinmolen: Ah yes, good one! I'll add it 👍

src/openforms/js/components/admin/forms/ReactSelect.js
robinmolen: Ah yes, will do!

src/openforms/js/components/admin/forms/VariableSelection.js
robinmolen: Ah yes, React-select has a default solution for this. If we change the structure of the choices variable/object, it should work.

I'll make a setup

src/openforms/js/components/admin/forms/ReactSelect.js
robinmolen: Mhhh I don't think value can be a required prop. Seeing how it can be null/undefined

src/openforms/js/components/admin/forms/VariableSelection.js
robinmolen: I've made it into something fancy ✨

src/openforms/js/components/admin/forms/VariableSelection.js
robinmolen: The variableSelections are now grouped by variable source (i.e. user defined, components and static). In order of user, component, static.

The ReactSelect now uses a recursive function to define its value. So in the future more ReactSelect components can use grouped options

src/openforms/js/components/admin/forms/ReactSelect.js
robinmolen: I've added it :)

src/openforms/js/components/admin/forms/VariableSelection.stories.js
```diff
@@ -36,6 +37,20 @@ export default {
...
+        slug: '',
+        url: '',
+        _generatedId: '',
+        isNew: true,
+        validationErrors: [],
+      },
+      {
+        configuration: {
+          display: 'form',
+        },
```
sergei-maertens, line 50: this looks a bit out of place, but I could be wrong about it!

robinmolen: Hmm yeah I think that that can be removed.. Will test it 👍

robinmolen: Yeah seems to work fine without it.. Removing it

robinmolen:
The chromatic tests are failing because either the DMN variable options aren't loaded yet, or for some reason the DMN select cannot be done correctly (This second fail is also inconsistent. Sometimes it succeeds, sometimes it fails...)

Maybe we should remove the dmn variable select check from the "Empty" DMN Action configuration test scenario

src/openforms/scss/components/admin/_select.scss
```diff
@@ -1,3 +1,35 @@
...
+
+    &__form-definition {
+      width: 100%;
+      font-size: 12px;
+      color: var(--body-quiet-color);
+    }
+  }
+
+  // The selected option only shows the label
+  .admin-react-select__value-container & {
```
sergei-maertens, line 27: we have `src/openforms/scss/vendor/_react-select.scss`, but I need to take a proper look at how this sass is structured still! more feedback will come

src/openforms/scss/vendor/_react-select.scss
```diff
@@ -20,5 +20,13 @@
...
-  inline-size: clamp(200px, 100%, var(--of-admin-select-max-inline-size));
+  inline-size: clamp(
+    var(--of-admin-select-min-inline-size),
+    100%,
+    var(--of-admin-select-max-inline-size)
+  );
+
+  .admin-react-select__menu {
+    z-index: 10;
+  }
```
sergei-maertens, line 31: magic z-index numbers usually hint at the [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) not being as expected - perhaps there's a better solution?

@stevenbal this may also be what you are running into with the prefill modal

src/openforms/scss/components/admin/_select.scss
robinmolen: Yeah, I put it here to keep all the `.form-variable-dropdown` styling contained to this file :)

src/openforms/scss/vendor/_react-select.scss
robinmolen: I'll take a look if this can be solved in another way 👍

robinmolen: This was needed, because the `.submit-row` would be on-top of the react-select menu. The z-index on the `.submit-row` was added because of an issue with the z-index of the tinymce.

I've changed the styling of the `.submit-row` and the tinymce to prevent this stacking issue

sergei-maertens:
Closing in favour of #4730

PR #4679, opened 2024-09-17
[#4652] Form builder error messages positions

### LINKED ISSUE #4652: Errors shown at strange position in form builder

2.7.x - latest

### Customer reference

_No response_

image
image
image

robinmolen:
Closes #4652 

**Changes**

Ensuring that the errorlist component is shown above the respective input.

Fixing the issues where the errorlist is shown in front of the label

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/scss/admin/_vendor-fixes.scss
+++ b/src/openforms/scss/admin/_vendor-fixes.scss
@@ -1,0 +1,16 @@
+.form-row {
+  // bootstrap styles compete with django .form-row styles...
+  display: block !important;
+
+  > div > .flex-container > label {
+    box-sizing: initial;
+  }
+}
+
+div:has(> .help):not(:has(> .fieldBox)) {
+  flex-wrap: wrap;
+
+  > .errorlist {
+    width: 100%;
+  }
+}
--- a/src/openforms/scss/admin/admin_overrides.scss
+++ b/src/openforms/scss/admin/admin_overrides.scss
@@ -2,3 +2,4 @@
 @import 'app_overrides';
 @import 'hijack';
 @import 'widgets';
+@import 'vendor-fixes';


src/openforms/scss/admin/_tooltip.scss
sergei-maertens, line 1: this sass module is intended to handle the styles of the help text/tooltip only. By default, Django displays the field help text below the input element (textinput, select, checkbox...). We customized this to display as a help icon besides the input, and hovering over it would display the help text in a popover.

All the sass here, is required to achieve this.

BUT - that means that your fix doesn't belong in this file, because ultimately it has nothing to do with this help text / tooltip. I think it'd be best to create an additional module `_vendor-fixes.scss` in the `src/openforms/scss/admin/` directory, and then make sure it's imported in https://github.com/open-formulieren/open-forms/blob/master/src/openforms/scss/admin/admin_overrides.scss

You can then include the relevant CSS rules in that module and then it's nicely isolated.

In the short term, I'm going to remove this `_tooltip.scss` module entirely and revert back to Django's default behaviour. Our default-project boilerplate has already undergone this change :smile:

sergei-maertens (changes requested):
Tested it and tried to break it, didn't succeed :) Good job

The only remark is about the file where the patch is applied - this tooltip is actually going to be removed so the fix belongs in a different place.

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.8.x: e1a360fa056d6e1ae8f09a54e7614fe132431231
* 2.7.x: 7185dd1fececef95b74b33f1dbe0381ac465ac2f
* 2.6.x: be301d2ed0599e72cd8aae4ff33df502e737c9fb

PR #4685, opened 2024-09-19
[#4658] Allow zip files (extra mimetypes) from Windows users

### LINKED ISSUE #4658: Windows users cannot upload .zip

2.6 - latest

### Customer reference

_No response_

When a windows user uploads a zip file they will always get a error about the file not being a zip:
image

Testen in browserstack 
Windows 10 + 11
Chrome, Edge and Firefox. 

This does not occur for Mac or Linux systems

vaszig:
Closes #4658

**Changes**

- Added regression tests for the extra mimetypes
- Update `UploadFileType` choices
- Modify `MimeTypeValidator` to accept the deprecated version of the mimetype and the new one 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/config/constants.py
```diff
@@ -85,7 +85,10 @@ class UploadFileType(models.TextChoices):
         "application/vnd.oasis.opendocument.*,application/vnd.stardivision.*,application/vnd.sun.xml.*",
         _("Open Office"),
     )
-    zip = "application/zip", _(".zip")
+    zip = (
+        "application/zip," "application/zip-compressed," "application/x-zip-compressed",
```
sergei-maertens, line 89:
```suggestion
        "application/zip,application/zip-compressed,application/x-zip-compressed",
```
unless this was deliberate, but in that case I don't understand why it was written that way :P

src/openforms/config/migrations/0057_alter_globalconfiguration_form_upload_default_file_types.py
```diff
@@ -0,0 +1,67 @@
...
+                blank=True,
+                default=list,
+                help_text="Provide a list of default allowed file upload types. If empty, all extensions are allowed.",
+                size=None,
+                verbose_name="Default allowed file upload types",
+            ),
+        ),
+        ConvertComponentsOperation(
+            "file", "ensure_extra_zip_mimetypes_exist_in_file_type"
+        ),
```
sergei-maertens, line 66: this belongs in a migration in the `forms` app, not the `config` app

src/openforms/formio/api/validators.py
```diff
@@ -65,6 +65,17 @@ def __call__(self, value: UploadedFile) -> None:
...
+        # gh #4658
+        # Windows use application/x-zip-compressed as a mimetype for .zip files, which
+        # is deprecated but still we need to support it. Instead, the common case for
+        # zip files is application/zip or application/zip-compressed mimetype.
+        if (
+            value.content_type
+            in ["application/x-zip-compressed", "application/zip-compressed"]
+            and mime_type == "application/zip"
+        ):
+            value.content_type = "application/zip"
```
sergei-maertens, line 77: why is this being assigned here? we are not supposed to mutate `value` inside a validator

Instead the condition should maybe be done similar like line 110?

```py
        elif mime_type == "application/zip" and value.content_type in (
            "application/zip-compressed",
            "application/x-zip-compressed",
        ):
            return
```

src/openforms/formio/migration_converters.py
```diff
@@ -127,6 +127,22 @@ def fix_file_default_value(component: Component) -> bool:
...
+
+    if "application/zip" in file_type:
+        if "application/x-zip-compressed" not in file_type:
+            file_type.append("application/x-zip-compressed")
+        if "application/zip-compressed" not in file_type:
+            file_type.append("application/zip-compressed")
+            "application/zip-compressed"
+
+        assign(component, "file.type", file_type)
+        return True
```
sergei-maertens, line 143: I'm not sure this is sufficient, since there's also the `filePattern` key which is sort of derived? See https://github.com/open-formulieren/open-forms/blob/ea5282deac7704e1d6d6f57b13fcc7aa0ad7b558/src/openforms/formio/components/vanilla.py#L413

sergei-maertens (changes requested):

vaszig:
This PR contains migrations changes and it needs backport, which makes it a bit hard to do it. This will be closed and tackled by basing the branch on the stable 2.6.x version in a new one.

PR #4696, opened 2024-09-23
:sparkles: [#4398] Check object ownership when creating submission

### LINKED ISSUE #4398: Add validation to the Objects API prefill plugin to check if the user is the owner of the object
The prefill plugin should check if the value of the auth attribute (this path will be configurable after https://github.com/open-formulieren/open-forms/issues/4396) is the same as the value of this authattribute for the authenticated user, if not, it should raise a `PermissionDenied` which should result in a 403 on the `Submission` create endpoint

We might want to remove the `Submission` in case this 403 happens, but I'm not 100% sure about this

We might also have to add a check to verify that the object is of the same objecttype as part of the config

stevenbal:
Closes #4398 

**Changes**

* Check object ownership when creating submission

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/contrib/objects_api/validators.py
```diff
@@ -78,3 +91,43 @@ def validate_document_type_references(
...
+            pk=backend.options.get("objects_api_group")
+        ).first()
+        if not api_group:
+            continue
+
+        try:
+            with get_objects_client(api_group) as client:
+                try:
+                    object = client.get_object(initial_data_reference)
+                    break
```
vaszig, line 116: Why is this needed here?I mean, don't we exit the loop if there is no exception?

src/openforms/contrib/objects_api/validators.py
```diff
@@ -78,3 +91,43 @@ def validate_document_type_references(
...
+                except RequestException:
+                    logger.exception(
+                        "Something went wrong while trying to retrieve "
+                        "object for product prefill permission check"
```
vaszig, line 120: Will this be specific to the product prefill exception?I had in mind that checking the owner of the object could be useful to other things in the future as well.

src/openforms/contrib/objects_api/validators.py
stevenbal: It's needed because we should exist the loop once an object is found, to not do unnecessary fetches to services from other registration backends.

src/openforms/contrib/objects_api/validators.py
stevenbal: Hmm yeah you're right, I will change it to not mention prefill

sergei-maertens (changes requested):
I think this is happening at the wrong place with the wrong assumptions :grimacing: 

For starters, `session[FORM_AUTH_SESSION_KEY]` should not be relied on - ideally once the submission is created/started, this should be cleaned from the session data entirely since we have a hook/signal that records it on the submission itself.

Second - we should always pass a `submission` instance when performing this kind of validation, exactly because different credentials could apply to a submission vs. what's in the session (the session data is a data race, since you can have different authentication details there compared to a submission that is also still in your session, for example when you are filling out two forms at the same time with different login options).

I think the prefill plugin itself should perform this access control, since:

1. in `submissions.api.viewsets.SubmissionViewSet.perform_create` we dispatch the submission start signal, that ensures we have the auth details stored
2. after that, we call `prefill_variables`, which receives the `submission` instance and passes it to the plugin
3. inside the plugin, you should be able to `raise PermissionDenied` if anything is fishy
4. if that exception is raised, it should bubble up to the API endpoint, will result in an HTTP 403 *and* the DB transaction should be rolled back so that the submission record is never persisted in the database

stevenbal:
@sergei-maertens it was mentioned in https://github.com/open-formulieren/open-forms/issues/4721 that the mechanism to pass `initial_data_reference` and update existing objects for Objects API registration (instead of creating new ones) should also function if prefill is not used. What would be the best place to do this permission check without having to implement it both in the registration backend and the prefill plugin?

sergei-maertens:
> @sergei-maertens it was mentioned in #4721 that the mechanism to pass `initial_data_reference` and update existing objects for Objects API registration (instead of creating new ones) should also function if prefill is not used. What would be the best place to do this permission check without having to implement it both in the registration backend and the prefill plugin?

uh... right. I'll have to think about that :sweat_smile:

stevenbal:
@sergei-maertens maybe there should be some kind of hook in `SubmissionViewSet.perform_create` that runs any validation for `initial_data_reference` for the configured registration backend and prefill plugins? That way it's generic, though I'm not sure what we could do to avoid performing the same validation twice, perhaps we could skip the validation for the prefill plugin if validation with the same `PLUGIN_IDENTIFIER` (in this case `objects_api`) has already been performed for a registration backend?

sergei-maertens:
@stevenbal I still don't see a simple/elegant way to do this, especially because data can be mutated externally and be outdated. E.g., if you verify it at `perform_create` time and track that it was okay at that time, some other system *could* make corrections to the data and by the time it's being registered, it could possibly no longer be verified (because a BSN was updated due to a mistake, for example).

I'm kind of leaning towards implementing the permission check itself once in `openforms.contrib.objects_api` and then re-using that functionality in both prefill and registration. This means that the check is done twice, but it does make it very explicit *and* I'd rather check too often than not often enough.

It's again a case of two different systems doing very similar things, but that doesn't make them the same. The prefill mapping and registration mapping are technically independent from each other too, we'll just offer convenience UI options to use one as a source for the other, but in the backend, they don't know of each other's existence and that's fine.

The meaning of `initial_data_reference` only becomes clear when interpreted in the context of the plugin using it anyway - so I don't see an agnostic "generic" mechanism. There's also a risk involved with that though - you may accidentally forget to implement access controls.

Perhaps for registration plugins we define a new hook on the base plugin which raises `NotImplementError` by default (so it causes a crash if forgotten). Let's call it `verify_initial_data_ownership` and it *only* gets called if `submission.initial_data_reference` is populated, so it won't break existing forms. You can then implement this calling logic in `openforms.registrations.tasks.pre_registration` so that it's plugin-agnostic. A similar pattern could then probably be applied to prefill?

stevenbal:
@sergei-maertens that sound like a good idea, I'll see if I can implement this

src/openforms/authentication/views.py
```diff
@@ -339,6 +345,14 @@ def _handle_return(self, request: Request, slug: str, plugin_id: str):
...
             authentication_success.send(sender=self.__class__, request=request)
 
+        if initial_data_reference:
+            # Pass the initial data reference back to the SDK, to allow sending it
+            # with the submission create call
+            return HttpResponseRedirect(
+                furl(response.url)
+                .add({"initial_data_reference": initial_data_reference})
+                .url
+            )
```
sergei-maertens, line 355: IMO this is too naive and we should probably look at a better/proper way to capture some state to re-use when redirecting back to the frontend.

If you look at `BasePlugin.handle_return`, the only guarantee is that it will return an `HttpResponse` - you have no guarantees about it being a redirect. Perhaps we must take a critical look at this, it's quite old code and formalize a couple of things.

I'm also a bit concerned about the amount of data we put in the session, it is sensitive to race conditions if multiple authentication requests start and concurrency on session data is quite bad itself.

Maybe an approach where the generic code prepares the return URL and passes it as a parameter to the plugin would be much more robust, that makes it properly stateless and removes the duplication of URL-encoding parameters that happens in plugins.

Ultimately I was surprised to see changes to the `authentication` module to handle this ticket - the responsibility of this module is authenticating users and it should not have any knowledge about the rest of the system. Giving the `initial_data_reference` special treatment doesn't fit the separation of concerns, and it's also not documented in the API documentation for these endpoints (nor should it, in my opinion).

The authentication start view gets the `form_url = request.GET.get("next")` parameter, which should contain *all* the information already (including GET parameters) to send the user back to the form after authenticating.

sergei-maertens: To add to this - nowhere do we seem to add the `_start` parameter in the backend, this is passed in from the frontend code, so I would really expect the `initial_data_reference` to work the same way, it's also just a query string parameter :thinking:

src/openforms/contrib/objects_api/validators.py
```diff
@@ -78,3 +92,52 @@ def validate_document_type_references(
...
         raise ValidationError(errors)
+
+
+def validate_object_ownership(submission: Submission) -> None:
+    form = submission.form
+
+    try:
+        auth_info = submission.auth_info
+    except ObjectDoesNotExist:
+        raise PermissionDenied("Cannot pass data reference as anonymous user")
```
sergei-maertens, line 103: what happens for a submission without data reference and no auth info?

src/openforms/contrib/objects_api/validators.py
```diff
@@ -78,3 +92,52 @@ def validate_document_type_references(
 
     if errors:
         raise ValidationError(errors)
+
+
+def validate_object_ownership(submission: Submission) -> None:
```
sergei-maertens, line 97: I'm not sure that this file is the correct place, we have validators here for configuration validation, but not stuff for runtime submissions. I'd be inclined to just put this in its own module, `ownership_verification.py` or something like that

src/openforms/contrib/objects_api/validators.py
```diff
@@ -78,3 +92,52 @@ def validate_document_type_references(
...
+    form = submission.form
+
+    try:
+        auth_info = submission.auth_info
+    except ObjectDoesNotExist:
+        raise PermissionDenied("Cannot pass data reference as anonymous user")
+
+    object = None
+    for backend in form.registration_backends.filter(backend="objects_api"):
```
sergei-maertens, line 106: We really can't have hardcoded plugin IDs in "generic" code, even if it's in the right domain. Custom extensions can subclass behaviour and they would be registered with a different ID, the ID could change in the future... And then this code won't properly work.

The plugin model is deliberately designed so that we don't need `if plugin == "foo"` code and to drop all the implementation details for a specific plugin in that plugin code. The only place where you can target a particular ID, is in data migrations.

On top of that - the initial data reference mechanism is generic - it could (and will at some point) also apply to ZGW API's. So I really expect a pattern along the lines of:

```
# during (pre?)-registration and prefill
if submission.initial_data_reference:
    plugin = lookup_plugin(submission)
    plugin.verify_ownership(submission)
```

and then the `plugin.verify_ownership` implementation can make use of the helpers in the shared functionality, but ultimately, when you have the plugin (registration) or variable (prefill) reference/instance, you have access to the necessary configuration:

* the API group to use
* the objecttype and its version
* the configured path/location that stores the identifier in the object

Architecture wise, everything in Open Forms works by:

1. invoke module (auth, prefill, registration)
2. resolve relevant plugin from form configuration + module
3. defer implementation to plugin instance

So, I'd expect helper functions in this package with the signature:

```py
def verify_ownership(
    submission: Submission, 
    client: ObjectsClient, 
    object_attribute: list[str],  # or a glom.Path
):
    ...
```

the `get_objects_client` context manager would be invoked somewhere in the caller, so that you can properly optimize connection pooling.

src/openforms/contrib/objects_api/validators.py
```diff
@@ -78,3 +92,52 @@ def validate_document_type_references(
...
+    if not object:
+        # If the object cannot be found, we cannot consider the ownership check failed
+        # because it is not verified that the user is not the owner
+        logger.info(
+            "Could not find object for initial_data_reference: %s",
+            submission.initial_data_reference,
+        )
+        return
+
+    # TODO should this path be configurable?
```
sergei-maertens, line 141: yes

src/openforms/prefill/base.py
```diff
@@ -129,3 +129,14 @@ def get_identifier_value(
...
+    def verify_initial_data_ownership(self, submission: Submission) -> None:
+        """
+        Hook to check if the authenticated user is the owner of the object
+        referenced to by `initial_data_reference
+
+        :param submission: an active :class:`Submission` instance
+        """
+        raise NotImplementedError(
+            "You must implement the 'verify_initial_data_ownership' method."
+        )
```
sergei-maertens, line 142: please put this method close to `get_prefill_values_from_options` as they belong together, and preferably put it *before* since it should be called first anyway.

src/openforms/prefill/contrib/objects_api/plugin.py
```diff
@@ -79,3 +81,6 @@ def configuration_context(cls) -> JSONObject | None:
                 for group in ObjectsAPIGroupConfig.objects.iterator()
             ]
         }
+
+    def verify_initial_data_ownership(self, submission: Submission) -> None:
+        validate_object_ownership(submission)
```
sergei-maertens, line 86: this breaks the isolation between prefill and registration modules because the implementation of `validate_object_ownership` looks at registration backends.

The configuration of which objects API to use is already present in the prefill options, they are passed to `get_prefill_values_from_options` so you need to make sure this also happens when calling `verify_initial_data_ownership`

src/openforms/contrib/objects_api/validators.py
stevenbal: `plugin.verify_initial_data_ownership` is only called if `submission.initial_data_reference` is not empty (for both prefill and registration)

src/openforms/contrib/objects_api/validators.py
stevenbal: What is the right place to configure this, should it be on both registration backend and prefill options? I don't know if there is a way around that, since I think it could be possible to use one without the other

sergei-maertens: Yeah indeed, and the UI copy button is then the convenience so that users can setup prefill from their registration configuration

src/openforms/contrib/objects_api/validators.py
sergei-maertens: Can you put in an explicit assert statement and docstring so that the expectations are made explicit?

src/openforms/authentication/views.py
stevenbal: you're right, I added it to the wrong queryParams in the SDK (I should have added it to the query params of the nextUrl :facepalm:). Made a PR for it https://github.com/open-formulieren/open-forms-sdk/pull/727 and I'll remove the backend changes

src/openforms/contrib/objects_api/validators.py
stevenbal: I've modified it to create the client in the respective plugins

src/openforms/contrib/objects_api/validators.py
stevenbal: I've added a TODO in the PR to add the field to the prefill config form as well (since this is probably reliant on #4799)

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -349,6 +357,16 @@ def validate(self, attrs: RegistrationOptions) -> RegistrationOptions:
                     {"version": _("Unknown version: {version}").format(version=version)}
                 )
 
+        if attrs.get("update_existing_object") and not attrs.get("auth_attribute_path"):
```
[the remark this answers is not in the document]
sergei-maertens: you should be able to set up the necesary frontend state & reproduce this problem in storybook! might be that we don't have validation error display hooked up into this arrayfield (?) yet

src/openforms/contrib/objects_api/validators.py
stevenbal: has been added

src/openforms/registrations/contrib/objects_api/config.py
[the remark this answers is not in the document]
stevenbal: fixed this and added a story for it

src/openforms/js/components/admin/form_design/registrations/objectsapi/V2ConfigFields.js
```diff
@@ -30,6 +34,13 @@ const onApiGroupChange = prevValues => ({
 });
 
 const V2ConfigFields = ({apiGroupChoices}) => {
+  const intl = useIntl();
+  const {REGISTRATION_OBJECTS_API_ENABLE_EXISTING_OBJECT_INTEGRATION = false} =
```
stevenbal, line 38: feature flag can be removed

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiOptionsForm.js
```diff
@@ -56,6 +56,7 @@ ObjectsApiOptionsForm.propTypes = {
     objecttype: PropTypes.string,
     objecttypeVersion: PropTypes.number,
     updateExistingObject: PropTypes.bool,
+    authAttributePath: PropTypes.array,
```
stevenbal, line 59: arrayOf

src/openforms/js/components/admin/form_design/registrations/objectsapi/LegacyConfigFields.js
```diff
@@ -30,88 +33,94 @@ const onApiGroupChange = prevValues => ({
...
+          errorMessage={
+            <FormattedMessage
+              description="Legacy Objects API registrations options: unknown error"
+              defaultMessage={`Something went wrong retrieving the available object types and/or versions.
+              Please check that the services in the selected API group are configured correctly.`}
+            />
+          }
+        >
+          <ObjectTypeSelect
+            label={intl.formatMessage({
```
stevenbal, line 58: use formattedmessage and fix proptypes

src/openforms/js/components/admin/form_design/variables/prefill/objects_api/CopyConfigurationFromRegistrationBackend.js
```diff
@@ -58,7 +62,9 @@ const CopyConfigurationFromRegistrationBackend = ({backends, setShowCopyButton})
             onClick={async e => {
               e.preventDefault();
               const confirmSwitch = await openCopyConfigurationConfirmationModal();
-              if (confirmSwitch) {
+              if (confirmSwitch && selectedBackend) {
+                // Avoid resetting the values because `objectsApiGroup` is being changed
+                setIsProgrammaticChange(true);
```
stevenbal, line 67: create a separate issue for this

src/openforms/prefill/base.py
```diff
@@ -65,6 +65,20 @@ def get_prefill_values(
...
+        """
+        Hook to check if the authenticated user is the owner of the object
+        referenced to by `initial_data_reference`
+
+        :param submission: an active :class:`Submission` instance
+        :param prefill_options: a dictionary containing the configuration options
+        """
+        raise NotImplementedError(
+            "You must implement the 'verify_initial_data_ownership' method."
+        )
```
sergei-maertens, line 80: Discussed - if any error happens in a prefill plugin when initial_data_reference is populated, then we want this to bubble up as a HTTP 403 in the submission start. It's important that the end-user is confronted with this so that they can contact the municipality that something is wrong in their configuration, rather than continuing with the form without prefilled information.

The prefill module now suppresses a lot of exceptions - it logs them, but continues without prefill values.

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -170,3 +174,29 @@ def update_payment_status(
     @override
     def get_variables(self) -> list[FormVariable]:
         return get_static_variables(variables_registry=variables_registry)
+
+    def verify_initial_data_ownership(self, submission: Submission) -> None:
+        assert submission.registration_backend
+        backend = submission.registration_backend
```
stevenbal, line 180: assert initial data ref

src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -170,3 +174,29 @@ def update_payment_status(
     @override
     def get_variables(self) -> list[FormVariable]:
         return get_static_variables(variables_registry=variables_registry)
+
+    def verify_initial_data_ownership(self, submission: Submission) -> None:
```
stevenbal, line 178: add options param (deserialized)

src/openforms/contrib/objects_api/ownership_validation.py
```diff
@@ -0,0 +1,77 @@
...
+        ObjectsAPIRegistration,
+    )
+    from openforms.submissions.models import Submission
+
+
+def validate_object_ownership(
+    submission: Submission,
+    client: ObjectsClient,
+    object_attribute: list[str],
+    plugin: ObjectsAPIPrefill | ObjectsAPIRegistration,
```
stevenbal, line 28: use the base plugins instead

src/openforms/contrib/objects_api/ownership_validation.py
```diff
@@ -0,0 +1,77 @@
...
+            "Could not find object for initial_data_reference: %s",
+            submission.initial_data_reference,
+        )
+        return
+
+    if (
+        glom(object["record"]["data"], Path(*object_attribute), default=None)
+        != auth_info.value
+    ):
+        logger.exception(
```
stevenbal, line 70: .warning instead

src/openforms/contrib/objects_api/ownership_validation.py
```diff
@@ -0,0 +1,77 @@
...
+        logger.info(
+            "Could not find object for initial_data_reference: %s",
+            submission.initial_data_reference,
+        )
+        return
+
+    if (
+        glom(object["record"]["data"], Path(*object_attribute), default=None)
+        != auth_info.value
+    ):
```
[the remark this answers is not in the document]
stevenbal: catch glom path error and raise from

src/openforms/frontend/frontend.py
```diff
@@ -26,6 +26,9 @@ def get_frontend_redirect_url(
     f = submission.cleaned_form_url
     f.query.remove("_of_action")
     f.query.remove("_of_action_params")
+    # Remove any data references, since this should already be stored on the
+    # Submission
+    f.query.remove("initial_data_reference")
```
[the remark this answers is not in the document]
stevenbal: I checked this, the `initial_data_reference` is still part of the URL, so it probably has to be treated the same as `_start` which is removed in `Submission.cleaned_form_url`?

src/openforms/js/components/admin/form_design/variables/prefill/objects_api/ObjectsAPIFields.js
```diff
@@ -118,12 +143,15 @@ const ObjectsAPIFields = ({errors, showCopyButton, setShowCopyButton}) => {
   if (error) throw error;
   const prefillProperties = loading ? LOADING_OPTION : value;
 
+  const [, authAttributePathErrors] = normalizeErrors(errors.options?.authAttributePath, intl);
```
stevenbal, line 146: display errors for other fields as well and add story

src/openforms/prefill/base.py
stevenbal: I've changed this, so now the user gets an error if prefill fails due to permissiondenied

image

though I do not see the Submission show up in the admin, is this as expected?

stevenbal:
@sergei-maertens I think I processed all comments except https://github.com/open-formulieren/open-forms/pull/4696#discussion_r1863209778, I'm not sure how to implement that and keep the logging

src/openforms/frontend/frontend.py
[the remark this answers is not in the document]
sergei-maertens: I think it's fine to keep it here too

sergei-maertens (approved):
I'm happy with it being in this state, let's hope CI passes too and then it's finally done! :tada:

PR #4700, opened 2024-09-24
:zap: [#4695] Do not fetch all IOtypen for objects API validation

### LINKED ISSUE #4695: Performance regression in Objects API (legacy) validation

master / 70771b792a183889247dfbe914547d1986505012

### Customer reference

_No response_

Noticed this while testing something completely different.

If you have the Objects API registration plugin configured with legacy document type URLs, then the backend will fetch *all* available document types in the Catalogi API to validate these URLs. This is really bad performance if there are multiple pages of document types.

Part of the waterfall in APM:

image

Due to the pagination handling, these requests are performed sequentially too, compounding the problem. The total time of the request handling is 2 minutes in this particular instance.

Instead, the validation should be simplified:

1. Check that the document type resource URL has the same prefix as the configured Catalogi API in the selected Objects API group
2. Once that is okay, retrieve the document type (you can even use a `HEAD` request if API version is 1.1+ to further optimize) and check that the status code is `200`

With 3 potential fields, that should give at most 3 requests, but all of them tiny.

1. Add an Objects API registration plugin to a form
2. Configure a legacy Document type URL (not the Catalogue + Document type dropdown!)
3. Save the form
4. Observe the API calls made to the Catalogi API

Saving a form is performant again.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
Closes #4695 

**Changes**

* Do not fetch all IOtypen for objects API validation

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Commit hygiene


src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -453,20 +458,40 @@ def _validate_catalogue_and_document_types(attrs: RegistrationOptions) -> None:
...
+        )
+
+        err_tpl = (
+            _("The provided {field} does not exist in the Catalogi API.")
+            if catalogus is None
+            else _("The provided {field} does not exist in the selected catalogue.")
+        )
+        err_msg = err_tpl.format(field=field)
+
+        if informatieobjecttypen_urls:
```
sergei-maertens, line 477:
```suggestion
        if informatieobjecttypen_urls is None:
```
it's possible the `catalogus["informatieobjecttypen"]` is equal to `[]` which would cause the code to skip this branch, which would be wrong :)

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -453,20 +458,40 @@ def _validate_catalogue_and_document_types(attrs: RegistrationOptions) -> None:
...
+            if url not in informatieobjecttypen_urls:
+                raise serializers.ValidationError({field: err_msg}, code="not-found")
+        else:
+            assert api_group.catalogi_service
+            iotypen_endpoint = (
+                furl(api_group.catalogi_service.api_root) / "informatieobjecttypen"
```
sergei-maertens, line 483:
```suggestion
                furl(api_group.catalogi_service.api_root) / "informatieobjecttypen/"
```
just in case :grimacing: this checks that it's not the collection endpoint

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -453,20 +458,40 @@ def _validate_catalogue_and_document_types(attrs: RegistrationOptions) -> None:
...
+            ).url
+
+            if not url.startswith(iotypen_endpoint):
+                raise serializers.ValidationError({field: err_msg}, code="not-found")
```
sergei-maertens, line 487: missing coverage for this case

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -453,20 +458,40 @@ def _validate_catalogue_and_document_types(attrs: RegistrationOptions) -> None:
...
+            with get_catalogi_client(api_group) as client:
```
sergei-maertens, line 488: I think we have to indent the whole code block to fit the context manager on line 396, since now you're opening and closing HTTP connections for each field. This would benefit from sharing the same connection pool as the initial Catalogi API interaction.

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -453,20 +458,40 @@ def _validate_catalogue_and_document_types(attrs: RegistrationOptions) -> None:
...
+                response = client.head(url)
```
sergei-maertens, line 489: I think we can do this only if we know that `client.api_version >= (1, 1, 0)` since in 1.0 the `HEAD` method is not required to be provided.

```suggestion
                response = client.head(url) if client.api_version >= (1, 1, 0) else client.get(url)
```

sergei-maertens (changes requested):

stevenbal:
2.7.x backport: https://github.com/open-formulieren/open-forms/pull/4705

sergei-maertens:
Backports:

* 2.8.x: 58e7047f42e1798baacc07227e67b77a51ede86a
* 2.7.x via #4705

PR #4702, opened 2024-09-24
[#4701] Add missing data migration for the global config and zip mimetypes

### LINKED ISSUE #4701: Forms which use the globally configured file types break because of .zip files

2.7.7

### Customer reference

_No response_

This is referring to Sentry 372143

When a user chooses to use the globally configured file types, we have a `'application/zip' is not a valid UploadFileType` because of missing data migrations.

This can be solved by adding an extra data migration for the general config.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #4701

**Changes**

- Added missing migration for the global configuration. When we define a global setting for the allowed file types we need to make sure the extra .zip mimetypes are in the list as well.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/config/migrations/0059_alter_globalconfiguration_form_upload_default_file_types.py
+++ b/src/openforms/config/migrations/0059_alter_globalconfiguration_form_upload_default_file_types.py
@@ -3,6 +3,27 @@
 from django.db import migrations, models
 import django_jsonform.models.fields
 
+from openforms.config.constants import UploadFileType
+
+
+def add_extra_zip_mimetypes(apps, _):
+    """
+    Set up the correct zip mimetypes.
+
+    This ensures all the allowed mimetypes concerning zip files are included.
+    """
+    GlobalConfiguration = apps.get_model("config", "GlobalConfiguration")
+    if not GlobalConfiguration.objects.exists():
+        return
+
+    config = GlobalConfiguration.objects.get()
+    if "application/zip" not in config.form_upload_default_file_types:
+        return
+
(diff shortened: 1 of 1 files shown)


src/openforms/config/migrations/0059_alter_globalconfiguration_form_upload_default_file_types.py
```diff
@@ -4,6 +4,35 @@
...
+            "application/zip-compressed",
+        )
+
+        def add_if_missing(current_list: list[str]):
+            for item in needed_mime_types:
+                if item not in current_list:
+                    current_list.append(item)
+
+        if "application/zip" in current_mimetypes:
+            add_if_missing(current_mimetypes)
```
sergei-maertens, line 30: this isn't right - the global config is an arrayfield in which each item has a possible choice value - the correct choice value is `UploadChoices.zip` (which is a single string with comma-separated mime types). In this patch, it looks like you're adding individual items to the arrayfield.

I believe I have the correct patch in my PR: https://github.com/open-formulieren/open-forms/pull/4703/files#diff-c3b529919eccd3ced5e9d359d132192403872ba6355a07979000ee7597b3d4e3R21-R25

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Cherrypicked the commit to stable/2.7.x, stable/2.8.x and master.

Issue #1018, opened 2021-12-02
WCAG-checkup


joeribekker:
**Tasks**

* [ ] Add alt-text to header logo
* [ ] Set disabled link (in the steps for example) color from #9ea4a9 to #767676 (see [reason](https://jxnblk.com/grays/) )
* [ ] Set "Sign above" text to from color #c3c3c3 to #767676
* [ ] Set default line-height to 1.5
* [ ] h1 to h4 are not styled (explicitly) and can be used in the content field. Please add some default styling.

sergei-maertens:
https://www.w3docs.com/tools/color-analyzer is useful to check color contrast

joeribekker:
@sergei-maertens antwoord op je vraag bij taak 1: "Terug naar de hoofdsite" oid

sergei-maertens:
ah, that was a question by Bart :-) the exact Dutch text will be handled in translations, but it's something like that.

PR #4747, opened 2024-10-10
Feature/4717 accessibility improvements aria

### LINKED ISSUE #4717: Accessibility improvements - Aria tags
### Thema / Theme

Frontend

This issue is related to #4622

This issue targets the aria improvements of the accessibility rapport:

- [x] 1.1.1
- [x] 4.1.2
- [x] 4.1.3

_No response_

_No response_

robinmolen:
Closes #4717

**Changes**

The site logo now provides a better textual alternative, the error message element now has the appropriate `role` and the PDF's are now screenreader friendly :)

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/formio/formatters/base.py
+++ b/src/openforms/formio/formatters/base.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from typing import Any, Generic, Iterable, Sequence, TypeVar
 
 from django.utils.encoding import force_str
-from django.utils.html import format_html_join
+from django.utils.html import format_html, format_html_join
 
 from ..typing import Component
 
@@ -51,11 +51,17 @@ class FormatterBase(Generic[ComponentT]):
     def join_formatted_values(
         self, component: Component, formatted_values: Iterable[str]
     ) -> str:
-        if self.as_html:
-            args_generator = ((formatted,) for formatted in formatted_values)
-            return format_html_join(self.multiple_separator, "{}", args_generator)
-        else:
+        if not self.as_html:
             return self.multiple_separator.join(formatted_values)
+
+        args_generator = ((formatted,) for formatted in formatted_values)
+        if component.get("multiple", False):
+            return format_html(
+                "<ul>{values}</ul>",
+                values=format_html_join("", "<li>{}</li>", args_generator),
+            )
+
+        return format_html_join(self.multiple_separator, "{}", args_generator)
 
     def process_result(self, component: ComponentT, formatted: str) -> str:
         return formatted
--- a/src/openforms/formio/rendering/tests/test_component_node.py
+++ b/src/openforms/formio/rendering/tests/test_component_node.py
@@ -533,3 +533,412 @@ class ComponentNodeTests(TestCase):
             "Input 6",
         ]
         self.assertEqual(labels, expected_labels)
+
+    def test_render_mode_pdf_with_list_values(self):
+        config = {
+            "components": [
+                {
+                    "type": "bsn",
+                    "key": "bsn",
+                    "label": "BSN",
+                    "multiple": True,
+                },
+                {
+                    "type": "date",
+                    "key": "date",
+                    "label": "Date",
+                    "multiple": True,
+                },
+                {
+                    "type": "datetime",
+                    "key": "datetime",
+                    "label": "DateTime",
+                    "multiple": True,
+                },
+                {
+                    "type": "email",
+                    "key": "email",
+                    "label": "Email",
+                    "multiple": True,
+                },
+                {
+                    "type": "iban",
+                    "key": "iban",
+                    "label": "IBAN",
+                    "multiple": True,
+                },
+                {
+                    "type": "licenseplate",
+                    "key": "licenseplate",
+                    "label": "LicensePlate",
+                    "multiple": True,
+                },
+                {
+                    "type": "phoneNumber",
+                    "key": "phoneNumber",
+                    "label": "PhoneNumber",
+                    "multiple": True,
+                },
+                {
+                    "type": "postcode",
+                    "key": "postcode",
+                    "label": "Postcode",
+                    "multiple": True,
+                },
+                {
+                    "type": "textarea",
+                    "key": "textarea",
+                    "label": "Textarea",
+                    "multiple": True,
+                },
+                {
+                    "type": "time",
+                    "key": "time",
+                    "label": "Time",
+                    "multiple": True,
+                },
+                {
+                    "type": "selectboxes",
+                    "key": "selectboxes",
+                    "label": "Selectboxes",
+                    "values": [
+                        {
+                            "value": "option1",
+                            "label": "Option 1",
+                        },
+                        {
+                            "value": "option2",
+                            "label": "Option 2",
+                        },
+                        {
+                            "value": "option3",
+                            "label": "Option 3",
+                        },
+                    ],
+                },
+                {
+                    "type": "radio",
+                    "key": "radio",
+                    "label": "Radio",
+                    "values": [
+                        {
+                            "value": "option1",
+                            "label": "Option 1",
+                        },
+                        {
+                            "value": "option2",
+                            "label": "Option 2",
+                        },
+                    ],
+                },
+                {
+                    "type": "select",
+                    "key": "select-single",
+                    "label": "Select single",
+                    "multiple": False,
+                    "openForms": {
+                        "dataSrc": "manual",
+                    },
+                    "data": {
+                        "values": [
+                            {"label": "Option 1", "value": "option1"},
+                            {"label": "Option 2", "value": "option2"},
+                        ]
+                    },
+                },
+                {
+                    "type": "select",
+                    "key": "select-multiple",
+                    "label": "Select multiple",
+                    "multiple": True,
+                    "openForms": {
+                        "dataSrc": "manual",
+                    },
+                    "data": {
+                        "values": [
+                            {
+                                "label": "Option 1",
+                                "value": "option1",
+                            },
+                            {
+                                "label": "Option 2",
+                                "value": "option2",
+                            },
+                            {
+                                "label": "Option 3",
+                                "value": "option3",
+                            },
+                        ]
+                    },
+                },
+                {
+                    "type": "textfield",
+                    "key": "textfield",
+                    "label": "Textfield",
+                    "multiple": True,
+                },
+            ]
+        }
+        data = {
+            "bsn": ["123456782", "987654321"],
+            "date": ["2024-11-11", "2024-11-13", "2024-11-16"],
+            "datetime": ["2022-09-08T12:00:00+02:00", "2022-10-08T15:00:00+02:00"],
+            "email": ["foo@mail.com", "bar@mail.com"],
+            "iban": ["RO09 BCYP 0000 0012 3456 7890", "RO09 BCYP 0000 0056 3456 7890"],
+            "licenseplate": ["123-aa-1", "12-aa-31"],
+            "phoneNumber": ["0612345678", "0645678923"],
+            "postcode": ["1234aa", "5678bb"],
+            "textarea": ["foo", "bar"],
+            "time": ["12:12:00", "14:02:00", "19:21:00"],
+            "selectboxes": {"option1": True, "option2": True},
+            "radio": "option1",
+            "select-single": "option1",
+            "select-multiple": ["option2", "option3"],
+            "textfield": ["Foo", "Bar"],
+        }
+
+        submission = SubmissionFactory.from_components(
+            components_list=config["components"], submitted_data=data
+        )
+        step = submission.steps[0]
+        register = Registry()
+
+        nodelist = []
+        with patch("openforms.formio.rendering.registry.register", new=register):
+            renderer = Renderer(submission, mode=RenderModes.pdf, as_html=True)
+            for component in step.form_step.form_definition.configuration["components"]:
+                component_node = ComponentNode.build_node(
+                    step_data=step.data, component=component, renderer=renderer
+                )
+                nodelist += list(component_node)
+
+        self.assertEqual(len(nodelist), 15)
+
+        expected = [
+            ("BSN", "<ul><li>123456782</li><li>987654321</li></ul>"),
+            (
+                "Date",
+                "<ul><li>11 november 2024</li><li>13 november 2024</li><li>16 november 2024</li></ul>",
+            ),
+            (
+                "DateTime",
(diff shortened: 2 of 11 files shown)


src/openforms/formio/rendering/tests/test_list_components.py
```diff
@@ -0,0 +1,160 @@
...
+from openforms.submissions.tests.factories import SubmissionFactory
+
+from ..default import ChoicesNode, SelectNode, TextNode
+from ..registry import Registry
+
+
+class FormNodeTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
```
sergei-maertens, line 15: Preferably set up the test data inside the test itself rather than creating context switches. You only have a single test, so this test data is only relevant for that single test anyway.

src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -63,7 +67,14 @@ <h2 class="subtitle">{{ submission_step_node.render }}</h2>
...
                                             <div class="submission-step-row__value{% if not component_node.display_value %} submission-step-row__value--empty{% endif %}">
-                                                {{ component_node.display_value|default:"" }}
+                                                {% if not component_node.display_value %}
+                                                {% elif component_node.component.type == "textfield" and component_node.component.multiple %}
+                                                    {% include 'report/submission_report_node_list.html' with values=component_node.value_list %}
+                                                {% elif component_node.component.type == "selectboxes" or component_node.component.type == "select" %}
+                                                    {% include 'report/submission_report_node_list.html' with values=component_node.value_list %}
+                                                {% else %}
+                                                    {{ component_node.display_value|default:"" }}
+                                                {% endif %}
```
sergei-maertens, line 77: hardcoding component types in these templates breaks the entire concept of a renderer/tree visitor where each node type controls its own behaviour. It spreads implementation details across multiple places and tracking down which code results in which display becomes much harder.

All this means that the `display_value` attribute is responsible for outputting the correct markup given the current context (render mode, html on/off). So, these template changes should be reverted IMO in favour of another approach.

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -262,6 +262,135 @@ def test_timestamp_included(self):
 
         self.assertEqual(reference_node.text, "Report created on: Jan. 1, 2024, 1 a.m.")
 
+    def test_list_fields_are_rendered_as_html(self):
```
sergei-maertens, line 265: Good test - you can use this to change the implementation details :)

src/openforms/templates/includes/page-header.html
```diff
@@ -17,8 +17,7 @@
         class="utrecht-link utrecht-link--openforms utrecht-page-header__openforms-return-url"
         href="{% firstof config.main_website '#' %}"
         title="{{ logo_alt }}"
-        aria-label="{{ logo_alt }}"
-    >{% if config.main_website %}{% trans "Back to website" %}{% endif %}</a>
+    >{% if config.main_website %}{{ logo_alt }}{% endif %}</a>
```
sergei-maertens, line 20: what happens now if the organization name is not specified? This will lead to a broken sentence, no?

> Back to website of

In those situations I'd expect the string "Back to website" to be used.

src/openforms/utils/pdf.py
```diff
@@ -106,5 +106,5 @@ def render_to_pdf(template_name: str, context: dict) -> tuple[str, bytes]:
         url_fetcher=UrlFetcher(),
         base_url=settings.BASE_URL,
     )
-    pdf: bytes = html_object.write_pdf()
+    pdf: bytes = html_object.write_pdf(pdf_variant="pdf/ua-1")
```
sergei-maertens, line 109: can you contribute this patch back to our default-project template?

src/openforms/ui/tests/test_header.py
```diff
@@ -24,9 +24,7 @@ def test_logo_label(self, get_solo_mock):
 
         header_node = response.html.find("header")
         link_node = header_node.find("a")
-        aria_label = link_node.get("aria-label")
```
sergei-maertens, line 27: These were explicit a11y patches: #2428 and #1018

So please explain to me why this patch is correct and the other one wasn't, because now we're just flipping back and forwards between approaches.

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -403,7 +403,13 @@ def test_report_is_generated_in_same_language_as_submission(self):
 
         for component_type, localised_input in fields:
             with self.subTest(f"FormIO label for {component_type}"):
-                self.assertIn(localised_input, html_report)
+                if component_type == "selectboxes":
+                    values = localised_input.split("; ")
+                    for value in values:
+                        self.assertIn(value, html_report)
+                else:
+                    self.assertIn(localised_input, html_report)
```
sergei-maertens, line 411: can't you update the `fields` pairs instead? If we're no longer displaying lists with semi-colons (which is a change in behaviour), the premise of the test no longer applies so the test needs to change. This approach makes the test even more unreadable :(

I think changing the respective line to:

```py
("selectboxes", "<li>This</li>"),
("selectboxes", "<li>That</li>"),
("selectboxes", "<li>The Other</li>"),
```
would already be sufficient

src/openforms/formio/rendering/default.py
```diff
@@ -69,6 +76,23 @@ def apply_to_labels(self, f: Callable[[str], str]) -> None:
...
+        for value, isSelected in super().value.items():
+            if isSelected:
+                value_to_add = value
+                # Search the label for the value
+                for choice in self.component["values"]:
+                    if choice["value"] == value:
+                        value_to_add = choice["label"]
+
+                values.append(value_to_add)
+        return values
```
sergei-maertens, line 94: This doesn't seem the right approach, `display_value` goes through `openforms.formio.service.format_value` which again uses the component registry and ultimately calls `openforms.formio.registry.BasePlugin.formatter` - see for example `src/openforms/formio/components/vanilla.py`.

If unspecified, the default formatter is used, and I think all of them are based on `FormatterBase`, which has access to the component and whether HTML is enabled or not. So, IMO the solution lies there to output `<ul>` HTML if HTML is enabled, and stick to the semi colon for plain text rendering.

Patching the code there should remove the need for the template changes and this new `value_list` concept.

FYI, the `value` is supposed to be the properly typed value, and whether it's a list or not is already encapsulated in there. The presentation layer `display_value` is its counterpart, which basically gives a "string representation of the value", optionally HTML capable if the context demands it.

sergei-maertens (changes requested):

src/openforms/ui/tests/test_header.py
robinmolen: The aria-label was removed as a attempt to remove the duplicate calling out of the title (the screenreader would announce the anchor as "Back to website Back to website").

In the rapport it was suggested to remove the `title` and `aria-label` and change the content to "Back to website of [OrganisationName]", to solve this issue. Due to the tooltip functionality of `title`, I decided to only remove the `aria-label`, because the content and `title` attribute would give the same needed information to screen readers. But i just noticed that this doesn't solve the "double title calling" issues.

After some digging I found out that this apparently is a issues with specific screen readers in combination with specific browsers. Some browsers don't/won't play nice with some screen readers..? I our case, the Apple VoiceOver screenreader will read the anchor content twice regardless of the presence of `label` and `aria-title` (tested with Chrome Version 129.0.6668.101 and Safari Version 17.6). I haven't yet tested this with other screenreader.

For now i will return the `aria-label`, seeing how this doesn't remove information or disrups the screenreader.

src/openforms/submissions/tests/test_tasks_pdf.py
robinmolen: I've changed it slightly. I've spread them out by content type (textfield, select, selectboxes and radio), to make it a bit more transparent what the expected behavior is

src/openforms/submissions/tests/test_submission_report.py
robinmolen: Ah, yeah that is a 100 times beter..

I've changed the `fields` to reflect the expected output for selectboxes :)

src/openforms/formio/rendering/default.py
robinmolen: Check, yeah I thought it would be best to keep the html tags inside the html file. This whole `value_list` concept was a way to supplement that. But you're right, it just creates confusion and unexpected behavior/outputs...

I've changed it now so that the html is added in the formatters. This is a much nicer solution, and gives a lot more control. Most of the changes are in the `join_formatted_values` of `formio/formatters/base.py`. Here is decided which values should be returned as a html list. I've added it here, instead of in the individual formatters, because the formatters treat (in most cases) all values of a component individually (so to add the `<ul>` tags, we still would have to use the `join_formatted_values`).
I do have my doubts with this approach for the selectboxes component. The formatter for the selectboxes receives the entire values dict, and transforms it to a `;` separated string. Whereas the other components use the `join_formatted_values` to join the values together. To make this work, the selectboxes formatter now returns a string of `li` elements (when `as_html=True`)

src/openforms/templates/includes/page-header.html
robinmolen: I've changed the `logo_alt` like you suggested. So when no organization name is specified, the `logo_alt` will return "Back to website"

src/openforms/formio/rendering/default.py
sergei-maertens: Sounds like what I initially assessed, nice to see that confirmed. I will take a closer look at the selectboxes later - but they have an odd data structure so I wouldn't be surprised if special treatment is still needed.

src/openforms/utils/pdf.py
robinmolen: Yes, i've added a PR to the default-project project 😄

src/openforms/submissions/templates/report/submission_report.html
robinmolen: I've removed the whole `value_list` concept, so i've reverted this change.

It felt weird to add html on any other level than the html file and I thought only about the use case surrounding the submission report, but i understand your point. It did create a weird situation and, with the second implementation, i've found some other places that benefit from the html list view

src/openforms/formio/rendering/default.py
robinmolen: Yeah i tried to figure out why the selectboxes formatter receives the entire dict, instead of 1 value from the dict, but i haven't found it out yet. Maybe we can change that, to keep all the formatters somewhat simular. 

But for now this works fine..
(Maybe the selectboxes formatter should also add the `ul` tags, to keep it simple/clear.. 🤔)

src/openforms/formio/rendering/tests/test_list_components.py
robinmolen: Okay check, i'll keep that in mind.. 👍

src/openforms/formio/rendering/tests/test_list_components.py
```diff
@@ -0,0 +1,160 @@
...
+                    )
+                elif component["type"] == "select":
+                    component_node = SelectNode(
+                        step_data=self.step.data, component=component, renderer=renderer
+                    )
+                elif component["type"] == "textfield":
+                    component_node = TextNode(
+                        step_data=self.step.data, component=component, renderer=renderer
+                    )
+                nodelist += list(component_node)
```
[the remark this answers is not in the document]
robinmolen: Yeah... this is a part of the `value_list` concept..
With my changes to remove the `value_list` and to add the expected html to the `display_value`, i've also changed this test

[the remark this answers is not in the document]
robinmolen: I don't fully understand your comment about the test case naming... but I think it should be fine now..

src/openforms/formio/formatters/base.py
```diff
@@ -51,12 +51,25 @@ def normalise_value_to_list(self, component: ComponentT, value: Any):
...
-        if self.as_html:
-            args_generator = ((formatted,) for formatted in formatted_values)
-            return format_html_join(self.multiple_separator, "{}", args_generator)
-        else:
+        if not self.as_html:
             return self.multiple_separator.join(formatted_values)
 
+        args_generator = ((formatted,) for formatted in formatted_values)
+        match component.get("type"):
+            case "textfield" | "select":
```
robinmolen, line 59: Don't limit the `ul` html logic to textfield and select. All multiple components should be displayed as html list

src/openforms/formio/rendering/tests/test_component_node.py
```diff
@@ -527,3 +527,181 @@ def test_render_mode_confirmation_email(self):
...
         ]
         self.assertEqual(labels, expected_labels)
+
+
+class FormNodeListsTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        config = {
```
robinmolen, line 537: Move test data to tests

src/openforms/templates/includes/page-header.html
```diff
@@ -8,17 +8,23 @@
...
-Back to website of {{ name }}
-{% endblocktranslate %}
+{% if config.organization_name %}
+    {% blocktranslate with name=config.organization_name asvar logo_alt trimmed %}
+        Back to website of {{ name }}
+    {% endblocktranslate %}
+{% else %}
+    {% blocktranslate asvar logo_alt trimmed %}
+        Back to website
+    {% endblocktranslate %}
```
robinmolen, line 18:
```suggestion
    {% trans "Back to website" %}
```

src/openforms/submissions/tests/test_tasks_pdf.py
```diff
@@ -262,6 +262,205 @@ def test_timestamp_included(self):
...
+                },
+            ],
+            submitted_data={
+                "textfield-single": "foo",
+                "textfield-multiple": ["foo", "bar"],
+            },
+            with_report=True,
+        )
+        html = submission.report.generate_submission_report_pdf()
+        print(html)
```
sergei-maertens, line 288: Can you remove the `print` statements to avoid the test output clutter?

robinmolen: Ah check, yeah that shouldn't be there... I'll remove it 👍

src/openforms/formio/formatters/base.py
robinmolen: All components that can be multiple are now shown as html lists

sergei-maertens (approved):

PR #4749, opened 2024-10-10
Make timestamp representation consistent in model string representation.

### LINKED ISSUE #3705: __str__ representation of a Submission uses GMT time instead of Europe/AMS time

sergei-maertens:
Closes #3705

**Changes**

I've grepped the project for `def __str__(self):` and checked all of them for their timestamp usages.

* Removed the custom `__str__` from `SubmissionStep`, as it adds no value. As a counter point, the timestamps are now included in the admin inline.
* Updated the timestamps that do make sense to ensure they're localized and using the `localtime` representation
* Cleaned up `SubmissionStep` and removed the obsoleted `_data` JSONField, which made it possible to use a tabular inline and that made it possible to display the timestamps in their own table columns without cluttering the submissions admin too much.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


vaszig (approved):
What about this one, why don't we do it here as well?
- [https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/models/form_version.py#L94](url)

sergei-maertens:
> * https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/models/form_version.py#L94

oh I missed that one! I'll update

PR #4751, opened 2024-10-11
[#4418] Add addressNl specific mappings to Objects API registration (variables tab)

### LINKED ISSUE #4418: Figure out a "transform" mechanism for (AddressNL) component(s) when registering them in the Objects API

vaszig:
Closes #4418 partly

**Changes**

This PR handles the frontend part of the task

- Added specific addressNL fields/mappings to the Objects API registration 
- Added stories

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
@@ -1,7 +1,7 @@
 import {FieldArray, useFormikContext} from 'formik';
 import isEqual from 'lodash/isEqual';
 import PropTypes from 'prop-types';
-import React, {useContext} from 'react';
+import React, {useContext, useState} from 'react';
 import {FormattedMessage} from 'react-intl';
 import {useAsync, useToggle} from 'react-use';
 
@@ -38,6 +38,9 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
   const {components} = useContext(FormContext);
   const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
   const {values: backendOptions, getFieldProps, setFieldValue} = useFormikContext();
+
+  const hasAnyAddressNlValue = Object.values(variable.initialValue).some(value => value !== '');
+  const [isSpecificTargetsChecked, setSpecificTargets] = useState(hasAnyAddressNlValue);
 
   /** @type {ObjectsAPIRegistrationBackendOptions} */
   const {
@@ -110,6 +113,13 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
       </ErrorMessage>
     );
 
+  const isAddressNlComponent = components[variable?.key].type === 'addressNL';
+  const deriveAddress = components[variable?.key]['deriveAddress'];
+
+  const onSpecificTargetsChange = event => {
+    setSpecificTargets(event.target.checked);
+  };
+
   return (
     <Fieldset>
       <FormRow>
@@ -126,33 +136,35 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
           />
         </Field>
       </FormRow>
-      <FormRow>
-        <Field
-          label={
-            <FormattedMessage
-              defaultMessage="Map to geometry field"
-              description="'Map to geometry field' checkbox label"
+      {!isAddressNlComponent && (
+        <FormRow>
+          <Field
+            label={
+              <FormattedMessage
+                defaultMessage="Map to geometry field"
+                description="'Map to geometry field' checkbox label"
+              />
+            }
+            helpText={
+              <FormattedMessage
+                description="'Map to geometry field' checkbox help text"
+                defaultMessage="Whether to map this variable to the {geometryPath} attribute"
+                values={{geometryPath: <code>record.geometry</code>}}
+              />
+            }
+            name="geometryVariableKey"
+            disabled={!!mappedVariable.targetPath}
+          >
+            <Checkbox
+              checked={isGeometry}
+              onChange={event => {
+                const newValue = event.target.checked ? variable.key : undefined;
+                setFieldValue('geometryVariableKey', newValue);
+              }}
             />
-          }
-          helpText={
-            <FormattedMessage
-              description="'Map to geometry field' checkbox help text"
-              defaultMessage="Whether to map this variable to the {geometryPath} attribute"
-              values={{geometryPath: <code>record.geometry</code>}}
-            />
-          }
-          name="geometryVariableKey"
-          disabled={!!mappedVariable.targetPath}
-        >
-          <Checkbox
-            checked={isGeometry}
-            onChange={event => {
-              const newValue = event.target.checked ? variable.key : undefined;
-              setFieldValue('geometryVariableKey', newValue);
-            }}
-          />
-        </Field>
-      </FormRow>
+          </Field>
+        </FormRow>
+      )}
       <FormRow>
         <Field
           name={`${namePrefix}.targetPath`}
@@ -173,6 +185,30 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
           />
         </Field>
       </FormRow>
+      {isAddressNlComponent && (
+        <FormRow>
+          <Field
+            name={`${namePrefix}.nestedMapping`}
+            label={
+              <FormattedMessage
+                defaultMessage="Define specific targets"
+                description="Define specific targets for all the fields of AddressNL component"
+              />
+            }
+          >
+            <Checkbox checked={isSpecificTargetsChecked} onChange={onSpecificTargetsChange} />
+          </Field>
+        </FormRow>
+      )}
+      {isSpecificTargetsChecked && (
+        <SpecificTargetsDisplay
+          namePrefix={namePrefix}
+          index={index}
+          choices={choices}
+          mappedVariable={mappedVariable}
+          deriveAddress={deriveAddress}
+        />
+      )}
       <div style={{marginTop: '1em'}}>
         <a href="#" onClick={e => e.preventDefault() || toggleJsonSchemaVisible()}>
           <FormattedMessage
@@ -200,7 +236,7 @@ ObjectsApiVariableConfigurationEditor.propTypes = {
   }).isRequired,
 };
 
-const TargetPathSelect = ({name, index, choices, mappedVariable, disabled}) => {
+const TargetPathSelect = ({id = 'targetPath', name, index, choices, mappedVariable, disabled}) => {
   // To avoid having an incomplete variable mapping added in the `variablesMapping` array,
   // It is added only when an actual target path is selected. This way, having the empty
   // option selected means the variable is unmapped (hence the `arrayHelpers.remove` call below).
@@ -217,7 +253,7 @@ const TargetPathSelect = ({name, index, choices, mappedVariable, disabled}) => {
       name="variablesMapping"
       render={arrayHelpers => (
         <Select
-          id="targetPath"
+          id={id}
           name={name}
           allowBlank
           choices={choices}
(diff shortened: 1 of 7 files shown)


src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,160 @@
...
+  objectsApiGroup,
+  objecttypeVersion,
+  TargetPathDisplay,
+}) => {
+  const {csrftoken} = useContext(APIContext);
+  const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
+
+  const {
+    loading,
+    value: targetPaths,
```
robinmolen, line 35: Providing `targetPaths` with a default value allows it to be used, even when its still loading (this way, things like `getTargetPath` cannot break on the `.find(...)` function)
```suggestion
    value: targetPaths = [],
```

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,160 @@
...
+    // Load only once:
+    []
+  );
+
+  const getTargetPath = pathSegment => targetPaths.find(t => isEqual(t.targetPath, pathSegment));
+
+  const choices =
+    loading || error
+      ? LOADING_OPTION
+      : targetPaths.map(t => [JSON.stringify(t.targetPath), <TargetPathDisplay target={t} />]);
```
robinmolen, line 60: With the default value of `[]` for the `targetPaths`, we don't need the LOADING_OPTION choices :)
```suggestion
  const choices = targetPaths.map(t => [JSON.stringify(t.targetPath), <TargetPathDisplay target={t} />]);
```

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,160 @@
...
+          id="targetPath"
+          name={name}
+          allowBlank
+          choices={choices}
+          {...props}
+          disabled={disabled}
+          value={JSON.stringify(props.value)}
+          onChange={event => {
+            if (event.target.value === '') {
+              arrayHelpers.remove(index);
```
robinmolen, line 140: When you have multiple fields filled in, and you clear one, all fields will be cleared. This happens because you remove the entire dataset of `index`.

I'll try to make a proof of concept solution...

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,347 @@
...
+          id={id}
+          name={name}
+          allowBlank
+          choices={choices}
+          {...props}
+          disabled={disabled}
+          value={JSON.stringify(props.value)}
+          onChange={event => {
+            if (event.target.value === '') {
+              arrayHelpers.remove(name);
```
robinmolen, line 328: Same as https://github.com/open-formulieren/open-forms/pull/4751/files#r1832285232. The entire dataset will be removed if one value is changed back to empty

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,347 @@
...
+              JSON.stringify(getTargetPath(mappedVariable.targetPath).jsonSchema, null, 2)
+            )}
+          </pre>
+        )}
+      </div>
+    </>
+  );
+};
+
+const TargetPathSelect = ({id, name, index, choices, mappedVariable, disabled}) => {
```
robinmolen, line 302: Perhaps we can move this to a separate file. Seeing how AddressNlObjectsApiVariableConfigurationEditor and GenericObjectsApiVariableConfigurationEditor both use a simular component.

If it is indeed the same, it would be best to make 1 component that both AddressNL and Generic can use

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,347 @@
...
+          description="Objects API variable registration configuration API error"
+          defaultMessage="Something went wrong when fetching the available target paths"
+        />
+      </ErrorMessage>
+    );
+
+  const onSpecificTargetPathsChange = event => {
+    setIsSpecificTargetPaths(event.target.checked);
+
+    // update Formik's state in order to be able to have this value as the default one for the React's state
```
robinmolen, line 99: I don't think we need the `isSpecificTargetPaths` when you also update the `specificTargetPaths`. Updating `specificTargetPaths` should trigger a reload, that updates the UI, so `isSpecificTargetPaths` is probably obsolete.

robinmolen: So i think you can just remove the state, and use `specificTargetPaths` for the checks

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
robinmolen: Its a bit big, and we can probability clean it up a bit, but this works.

So, when `event.target.value === ''`, we check if there are multiple selected values. We get all the keys of the mappedVariable, and check if there are more than 1 specific fields present (i.e., more that 1 field has a selected value). If there are multiple specific values, we set the current field to undefined, and otherwise we remove the entire arrayHelpers index
```suggestion
              if (
                Object.keys(mappedVariable).filter(variable =>
                  [
                    'postcodeTargetPath',
                    'houseLetterTargetPath',
                    'houseNumberTargetPath',
                    'houseNumberAdditionTargetPath',
                    'cityTargetPath',
                    'streetNameTargetPath',
                  ].includes(variable)
                ).length > 1
              ) {
                setFieldValue(name, undefined);
              } else {
                arrayHelpers.remove(name);
              }
```

robinmolen: I think you should also move this `onChange` logic to a separate function, to keep thinks clear :)

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
sergei-maertens: Adding to this:

1. If using native selects, it's nice to get the visual feedback in the dropdown that the options are still loading
2. If using react-select, you can forward the `isLoading` prop and then it the variant suggested by Robin is better

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
sergei-maertens: I think this is what you're referring to @vaszig ?

This is/was deliberate - because dealing with the backend validation errors on the variables tab is far from straight forward, the idea was that only knowingly valid items may be present in the mapping array. So if one configuration aspect being cleared means the item becomes invalid and would not pass backend validation, it is removed from the mapping entirely.

While in the modal that may seem not obvious, but when saving the modal the registration summary would accurately reflect that the form variable is not mapped to any registration aspect because the item is absent from the mapping, so it's not happening in an invisble way.

I may be lacking a bit more context here, so let me know if I miss anything.

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
sergei-maertens: > both use a simular component

Similar != the same

This was deliberate to cut down on generic complexity - rather than have one component that must be able to support everything, it was opted to co-locate code and behaviour. We will for sure have more component types that require additional/flexible behaviours, and once a pattern emerges from that we can refactor/revisit.

Alex likes to mention the rule of three here - doing it once is okay, doing it a second time is also okay, once you're doing it for the the third time -> abstract it

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
vaszig: @sergei-maertens  if I understand correctly  we want to remove the item, as I have i in my PR, right?

sergei-maertens: yes, I think so

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
vaszig: Using native selects so this remains the same.

vaszig:
FYI, we have to take into consideration the old forms that they already may have addressNL `variable_mappings` configured. We may need a data migration for this.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -71,126 +89,72 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
...
-              JSON.stringify(getTargetPath(mappedVariable.targetPath).jsonSchema, null, 2)
-            )}
-          </pre>
-        )}
-      </div>
-    </Fieldset>
+          </Field>
+        </FormRow>
+      </Fieldset>
+      <Fieldset>
```
sergei-maertens, line 143:
```suggestion
```

src/openforms/js/components/admin/form_design/variables/VariablesEditor.stories.js
```diff
@@ -680,3 +680,227 @@ export const WithValidationErrors = {
...
+    },
+  },
+
+  play: async ({canvasElement, step}) => {
+    const canvas = within(canvasElement);
+
+    const editIcons = canvas.getAllByTitle('Registratie-instellingen bewerken');
+    await userEvent.click(editIcons[0]);
+
+    const modalForm = await screen.findByTestId('modal-form');
```
sergei-maertens, line 753: with the recent changes on master, you can use `canvas` instead of `screen`

src/openforms/js/components/admin/forms/Inputs.js
```diff
@@ -119,8 +119,8 @@ const Checkbox = ({name, label, helpText, noVCheckbox = false, ...extraProps}) =
 };
 
 Checkbox.propTypes = {
-  name: PropTypes.string.isRequired,
-  label: PropTypes.node.isRequired,
+  name: PropTypes.string,
+  label: PropTypes.node,
```
sergei-maertens, line 123: with the remarked layout issues, I think the `name` and `label` really should stay required. Any form input element always should have a `name`, even if we don't really use it in the `onChange` handler, let's make sure we follow best HTML practices.

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -56,7 +56,10 @@ class ObjecttypeVariableMappingSerializer(serializers.Serializer):
         help_text=_(
             "Representation of the JSON target location as a list of string segments."
         ),
+        required=False,
     )
+    # specific options according to the variable key/type
+    options = JSONFieldWithSchema(default={}, required=False)
```
sergei-maertens, line 62:
```suggestion
    options = JSONFieldWithSchema(default=dict, required=False)
```
I'm always scared when I see a mutable default argument :)

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -482,16 +485,40 @@ def _get_data(
                 # This should only happen for payment status update,
                 # where only a subset of the variables are updated.
                 continue
-            target_path = mapping["target_path"]
 
-            # Type hint is wrong: currently some static variables are of type date/datetime
```
sergei-maertens, line 487: we want to keep this comment :)

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -482,16 +485,40 @@ def _get_data(
...
-            glom.assign(record_data, glom.Path(*target_path), value, missing=dict)
+            target_path = mapping.get("target_path")
+            if (
+                component_key_type_mappings
+                and component_key_type_mappings.get(variable_key) == "addressNL"
+            ):
+                assert isinstance(value, dict)
+                filtered_value = {k: v for k, v in value.items() if v}
+                if target_path:
+                    value.pop("secretStreetCity")
```
sergei-maertens, line 503: shouldn't this be `filtered_value`?

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -482,16 +485,40 @@ def _get_data(
...
+                        filtered_value,
+                        missing=dict,
+                    )
+
+                elif subfields_mapping := mapping.get("options"):
+                    for subfield_key, subfield_value in subfields_mapping.items():
+                        glom.assign(
+                            record_data,
+                            glom.Path(*subfield_value),
+                            filtered_value[underscore_to_camel(subfield_key)],
```
sergei-maertens, line 516: I'd rather we don't have to deal with snake/camel case conversion, but that would require re-structuring the data model from a simple key-value object to an array of mapping objects, which will impact the frontend code again. So, maybe just put a `# TODO` here to remark that and let's leave it as it is then

sergei-maertens: what's more concerning, is that `filtered_value` excludes empty-ish values, but the `subfield_key` access doesn't consider that. E.g. if you only fill out postal code + house number and leave house letter empty, while house letter is mapped, this will crash. So I'd definitely write a test for this case too :)

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -482,16 +485,40 @@ def _get_data(
...
             # Comply with JSON Schema "format" specs:
             if isinstance(value, (datetime, date)):
                 value = value.isoformat()
 
-            glom.assign(record_data, glom.Path(*target_path), value, missing=dict)
+            target_path = mapping.get("target_path")
+            if (
+                component_key_type_mappings
+                and component_key_type_mappings.get(variable_key) == "addressNL"
+            ):
```
sergei-maertens, line 499: this is not yet a proper generic solution because it requires us to write `if component_type == '...'` code which gets very messy very quickly, so please put a `TODO` comment saying this will need to be addressed at some point

sergei-maertens (commented):
Functional remarks:

* Is it possible to disable/remove the "map to geometry field" row/option when configuring an AddressNL component?
* The checkbox looks a bit odd because the label is specified before the checkbox itself:
   image

    Standard Django form styling puts this in a different order:
    
    image
    
    If we change this (please do), then I think the "map geometry field" needs the same treatment :)

* The "show json schema" component doesn't seem to make a lot of sense when mapping specific sub-components

   image

   "Normally" this shows the json schema of the selected property from the object type when selecting a target path. This isn't relevant when mapping in more detail, unless you use this component for each nested target path, which will probably clutter too much, so I'd just not display it at all when mapping the sub components.

sergei-maertens (approved):

vaszig:
Just a reminder for the translations and for https://github.com/open-formulieren/open-forms/pull/4751#issuecomment-2482412560

sergei-maertens:
> Just a reminder for the translations and for [#4751 (comment)](https://github.com/open-formulieren/open-forms/pull/4751#issuecomment-2482412560)

I will do translations separately and I'll take a look at the data migration remark.

PR #4752, opened 2024-10-11
Ensure submissions tablib.Dataset generation doesn't crash

### LINKED ISSUE #3629: Bulk Export of submission crash when submission objects with different amount of items in them (repeating groups, uploads)

2.2.6

Bodegr 39
nopolder: 83
venr: 86
RID: 130

sentry: 340715
Bulk export will crash when 2 submissions have a big difference in keys: 

Export generates the headers to merge from submission A:
```python
[
    "Formuliernaam",
    "Inzendingdatum",
    "voornaam",
    "voorvoegsel",
    "achternaam",
    "bsn",
    "straatnaam",
    "huisnummer1",
    "huisletter_persoon",
    "huisnummertoevoegingPersoon",
]

```
But submission B looks like this:
```python	
[
    "Evenementenvergunning aanvragen",
    datetime.datetime(2023, 11, 14, 11, 3, 42, 407842),
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
]
```
Because 
```
    "voornaam",
    "voorvoegsel",
    "achternaam",
    "bsn",
    "straatnaam",
    "huisnummer1",
    "huisletter_persoon",
    "huisnummertoevoegingPersoon",
```
are all empty, these components are all part of a step that was skipped for submission B.

This causes the mass export to fail.

sergei-maertens:
Closes #3629

**Changes**

* Added regression tests for repeating group and file upload components that can cause the submission export to crash.
* Ensure that (hidden/skipped) submission steps because of logic are not skipped during export
* Ensure that editgrid components are considered a single component during export instead of recursing into them

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/formio/rendering/default.py
+++ b/src/openforms/formio/rendering/default.py
@@ -229,7 +229,15 @@ class EditGridNode(ContainerMixin, ComponentNode):
     display_value: str = ""
 
     @property
+    def is_layout(self) -> bool:
+        if self.mode == RenderModes.export:
+            return False
+        return True
+
+    @property
     def value(self):
+        if self.mode == RenderModes.export:
+            return super().value
         return None
 
     @property
@@ -268,6 +276,10 @@ class EditGridNode(ContainerMixin, ComponentNode):
         So we need to repeat the child nodes of the configuration and associate them with the data
         provided by the user.
         """
+        # during exports, treat the entire repeating group as a single component/column
+        if self.mode == RenderModes.export:
+            return
+
         repeats = len(self._value) if self._value else 0
 
         for node_index in range(repeats):
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -235,10 +235,7 @@ class ComponentNode(Node):
             return
 
         # in export mode, only emit if the component is not a layout component
-        if self.mode != RenderModes.export or (
-            not is_layout_component(self.component)
-            and self.component["type"] != "editgrid"
-        ):
+        if self.mode != RenderModes.export or (not is_layout_component(self.component)):
             yield self
 
         for child in self.get_children():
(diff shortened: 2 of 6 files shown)


sergei-maertens:
Looks like only the `openforms.formio.rendering.tests.test_component_node.FormNodeTests.test_export_always_emits_all_nodes` test broke because of this patch, and I'm fairly certain that test is just wrong :)

src/openforms/formio/rendering/default.py
```diff
@@ -228,8 +228,16 @@ class EditGridNode(ContainerMixin, ComponentNode):
     layout_modifier: str = "editgrid"
     display_value: str = ""
 
+    @property
+    def is_layout(self) -> bool:
+        if self.mode == RenderModes.export:
+            return False
+        return True
```
stevenbal, line 235: I don't know too much about these rendering classes yet, what does it mean for rendering if `is_layout` is `True` vs `False`? Does `True` imply that the children are rendered individually?

src/openforms/formio/rendering/default.py
```diff
@@ -268,6 +276,10 @@ def get_children(self) -> Iterator["ComponentNode"]:
         So we need to repeat the child nodes of the configuration and associate them with the data
         provided by the user.
         """
+        # during exports, treat the entire repeating group as a single component/column
+        if self.mode == RenderModes.export:
+            return
```
stevenbal, line 281: This probably answers my previous question with yes :p

stevenbal (approved):

src/openforms/formio/rendering/default.py
sergei-maertens: yeah these nodes are skipped/ignored in some modes because they don't have an intrinsic value - for example in the email summary only the child elements are displayed, but the column/fieldset/editgrid itself aren't. Similarly, in the export code we have:

```py
        for node in data_nodes.get_children():
            if node.is_layout:
                continue
            yield node
```

sergei-maertens:
Backports:

* 2.8.x: ed7de9dd9a041399c95f5dde0046bd7f3a954d3e
* 2.7.x: 99ccd27de9cdadad7dc6bef0f9adc2710cb7c549

PR #4755, opened 2024-10-14
Enable soft-required validation in the form designer

### LINKED ISSUE #4546: Add "soft required" validation/hint to document uploads component
Taiga DH 661, business case "ondernemers"

Approval for the "dynamic variant" is given.

Forms require documents to be uploaded that provide information whether the permit can be granted or not. However, any permit request must be taken into consideration because of legal reasons, even if not all (required) documents are provided. To improve the user experiences and avoid unnecessary costs, we want to be able to mark a (document) field as "soft-required" so that if no documents are uploaded, a warning can be displayed to the end users that they may incur additional costs. This will not block progress to the next step / submission completion.

image

The technical implementation will be challenging for this.

**Tasks**

- [x] Add validation option `softRequired` in the form builder/form designer. This cannot be enabled both at the same time as `validate.required` is checked (so toggling one should uncheck the other).
- [x] Add a new custom component "SoftRrequiredErrors", which is a variant of the Content component. There should be a placeholder in here for the actual errors that will be programmatically filled, something like `{{ field_errors }}`. This must work properly with the CSP.
- [x] Implement the detection of empty values - in a generic way, not just targetting file upload components. This mechanism is likely to get extended in the future to other component types. For "invalid" components, we need to extract the label to display it in the placeholder location, and for accesibility reasons the "invalid" components should have a pointer to the error node linking everything together (TODO Sergei: double check this with notes from NL DS meeting)
- [x] Document on RTD how to use this

sergei-maertens:
Partially closes #4546 - depends on https://github.com/open-formulieren/formio-builder/pull/188

**Changes**

* [x] Upgraded formio-builder to enable soft required validation option in file components
* [x] Enabled the new `softRequiredErrors` component in the form designer
* [x] Added documentation on how to use these components

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/docs/manual/forms/basics.rst
+++ b/docs/manual/forms/basics.rst
@@ -142,7 +142,7 @@ stap.
 
 In deze sectie kunt u velden (ook wel componenten genoemd) naar de formulier
 stap slepen en configureren. Kijk voor alle mogelijkheden naar het overzicht van
-:ref:`formulier velden <manual_form_fields>` en naar de
+:ref:`formuliervelden <manual_form_fields>` en naar de
 :ref:`voorbeelden <manual_examples>`.
 
 Bevestiging
--- a/docs/manual/forms/form_fields.rst
+++ b/docs/manual/forms/form_fields.rst
@@ -108,7 +108,7 @@ Validatie
 ---------
 
 * **Verplicht**: Indien aangevinkt dan is dit veld verplicht voor de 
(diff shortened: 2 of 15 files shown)


robinmolen (approved):
Yeah looks good 👍

PR #4758, opened 2024-10-15
:sparkles: [#4344] Configurable objects API group in ZGW registration

### LINKED ISSUE #4344: Add option to select Objects API Group to use in the ZGW registration
### Thema / Theme

API

Once https://github.com/open-formulieren/open-forms/pull/4343 is merged, we should add the option to select the Objects API Group to use in the ZGW registration. It is currently implicit, the first group is used.

See https://github.com/open-formulieren/open-forms/pull/4332#discussion_r1619040850

_No response_

_No response_

stevenbal:
previously this was not configurable and always used the ObjectsAPIGroup with the lowest pk to create the client when registering the submission

Closes #4344

**Changes**

* Configurable objects API group in ZGW registration

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.js
```diff
@@ -25,7 +26,21 @@ import {
...
 
-const ZGWFormFields = ({name, apiGroupChoices, confidentialityLevelChoices}) => {
+/**
+ * Callback to invoke when the API group changes - used to reset the dependent fields.
+ */
+const onApiGroupChange = prevValues => ({
+  ...prevValues,
+  objecttype: '',
+  objecttypeVersion: undefined,
+});
```
stevenbal, line 36: I'm not sure if this is actually needed, since this form still uses the old Objecttype and ObjecttypeVersion fields (simple inputs instead of select)

sergei-maertens: this will get refactored as part of #4606, I'm gradually making my way down the various fields :D

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
```diff
@@ -18,8 +18,12 @@ const ZGWOptionsForm = ({index, name, label, schema, formData, onChange}) => {
   const [modalOpen, setModalOpen] = useState(false);
   const validationErrors = useContext(ValidationErrorContext);
 
-  const {zgwApiGroup, zaakVertrouwelijkheidaanduiding} = schema.properties;
+  const {zgwApiGroup, zaakVertrouwelijkheidaanduiding, objectsApiGroup} = schema.properties;
```
sergei-maertens, line 21: please update the component prop types accordingly

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -578,7 +578,9 @@ def register_submission_to_objects_api(
         )
 
         # In a follow up PR: the group will be configurable:
-        api_group = ObjectsAPIGroupConfig.objects.order_by("pk").first()
+        api_group = options.get(
+            "objects_api_group", ObjectsAPIGroupConfig.objects.order_by("pk").first()
+        )
```
sergei-maertens, line 583: I'd prefer to remove this default/fallback behaviour entirely so that all configuration is explicit, and we can then also define this accordingly in the options type declarations.

```py
api_group = options["objects_api_group"]
assert api_group is not None
```

src/openforms/registrations/contrib/zgw_apis/typing.py
```diff
@@ -37,6 +39,7 @@ class RegistrationOptions(TypedDict):
     organisatie_rsin: NotRequired[str]
     zaak_vertrouwelijkheidaanduiding: NotRequired[VertrouwelijkheidAanduiding]
     medewerker_roltype: NotRequired[str]
+    objects_api_group: NotRequired[ObjectsAPIGroupConfig]
```
sergei-maertens, line 42:
```suggestion
    objects_api_group: ObjectsAPIGroupConfig | None
```
I think we can enforce this key being set at all times. If the value is `None`, then no objects API registration is enabled (and we expect `objecttype` and `objecttype_version` to not be set at all, but either way they're ignored)

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.js
```diff
@@ -112,6 +127,10 @@ const ZGWFormFields = ({name, apiGroupChoices, confidentialityLevelChoices}) =>
             collapsible
             fieldNames={['objecttype', 'objecttypeVersion', 'contentJson']}
           >
+            <ObjectsAPIGroup
+              apiGroupChoices={objectsApiGroupChoices}
+              onApiGroupChange={onApiGroupChange}
+            />
```
sergei-maertens, line 133: I think this one should be `isClearable` to disable objects API integration again? Though clearing the objecttype field has the same effect

src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -104,6 +105,14 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
     )
 
     # Objects API
+    objects_api_group = PrimaryKeyRelatedAsChoicesField(
+        queryset=ObjectsAPIGroupConfig.objects.exclude(
+            Q(objects_service=None) | Q(objecttypes_service=None)
+        ),
+        help_text=_("Which Objects API set to use."),
+        label=_("Objects API set"),
+        required=False,
```
sergei-maertens, line 114: I think we may need an `allow_null=True` to make it explicit that no objects API integration is done.

There's some more steps involved in this PR I think:

* A data migration that updates existing registration backends
    - if no `objecttype` is specified, ensure that `objects_api_group` is set to `None`
    - if an objecttype is specified, set the `objects_api_group` to the lowest PK (use that queryset that is now used at runtime)
    - this behaviour also needs to be applied during form import time when legacy forms are being imported (!), we can probably do something with the [`DEFINITION_CONVERTERS`](https://github.com/open-formulieren/open-forms/blob/ad7ead2aa3267c23ecda3063cc570992ab91ca47/src/openforms/formio/migration_converters.py#L295)
* Validation! Once we know which API group is used, we know the API root URl of the objecttypes API, and can check the URL provided. We can even look up the provided object type and check that it exists, but let's not do _that_ yet because credentials to the object types API may not be configured yet. This should only come up when people edit the form again, and should not affect existing runtime behaviour of forms.

sergei-maertens (changes requested):
You'll also need to update some stories since they now fail to render in chromatic :)

src/openforms/registrations/contrib/zgw_apis/plugin.py
stevenbal: Would it also work if I add it to the if statement here (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/zgw_apis/plugin.py#L443-L447), so that it will only try to register the submission if the API group is part of the options?

src/openforms/registrations/contrib/zgw_apis/options.py
stevenbal: I added a new type of converter for the registration backends, that runs before the data is passed into the serializer

stevenbal:
I tested the import locally, I did notice that there seems to be an issue when exporting/importing a form that has the legacy zaaktype empty, it is not present in the export and causes errors when importing:

image

I think it's unrelated, so I'll create a separate issue for it

src/openforms/registrations/contrib/zgw_apis/plugin.py
sergei-maertens: yes, sort of - the type checker will not be able to narrow the type of `options["objects_api_group"]` so for that you'll still need the assert

src/openforms/forms/utils.py
```diff
@@ -387,6 +394,11 @@ def apply_definition_conversions(configuration: JSONObject) -> None:
         converter(configuration)
 
 
+def apply_registration_backend_conversions(configuration: list[dict]) -> None:
+    for converter in REGISTRATION_BACKEND_CONVERTERS:
+        converter(configuration)
```
sergei-maertens, line 399: Looking at the data migration which doesn't make use of this constant, I don't see a benefit in adding an abstraction for this here and would much favour explicitness in the options serializers to perform the normalization.

The `CONVERTERS` is different, because we have a custom migration operation that allows you to refer to a converter operation and you don't have to code data migrations anymore, while ensuring that the same logic runs during import/export of forms. Those conditions do not apply here, so let's simplify the implementation.

src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -245,6 +257,41 @@ def _validate_against_catalogi_api(attrs: RegistrationOptions) -> None:
...
+
+    objects_api_group = attrs.get("objects_api_group")
+    if not objects_api_group:
+        raise serializers.ValidationError(
+            {
+                "objects_api_group": _(
+                    "Objects API group must be specified if an objecttype is specified."
+                )
+            },
+            code="invalid",
```
sergei-maertens, line 277:
```suggestion
            code="required",
```
or which code does DRF use for missing/blank/empty values?

src/openforms/registrations/contrib/zgw_apis/tests/test_migrations.py
```diff
@@ -0,0 +1,38 @@
...
+            options={"objecttype": "https://objecttypes.com/foo/bar"},
+        )
+
+    def test_set_explicit_objects_api_groups_on_zgw_api_group_configs(self):
+        self.backend_without_api_group.refresh_from_db()
+        self.backend_with_api_group.refresh_from_db()
+
+        self.assertEqual(
+            self.backend_without_api_group.options["objects_api_group"], None
+        )
```
sergei-maertens, line 34:
```suggestion
        self.assertIsNone(self.backend_without_api_group.options["objects_api_group"])
```

src/openforms/registrations/contrib/zgw_apis/tests/test_migrations.py
```diff
@@ -0,0 +1,38 @@
...
+        self.backend_with_api_group = FormRegistrationBackend.objects.create(
+            form=form2,
+            backend="zgw-create-zaak",
+            options={"objecttype": "https://objecttypes.com/foo/bar"},
+        )
+
+    def test_set_explicit_objects_api_groups_on_zgw_api_group_configs(self):
+        self.backend_without_api_group.refresh_from_db()
+        self.backend_with_api_group.refresh_from_db()
```
sergei-maertens, line 30: one day this will bite you in the a$$ :) this model is now at the state it was before migrations ran. You typically want to use `self.apps.get_model(...)` to get the snapshot/state of the model *after* migrations ran, and then do `Model.objects.get(pk=self.backend_without_api_group.pk)` to get an accurate representation.

It mostly goes right because it's not often we combine schema migrations and data migrations, but it *does* happen.

sergei-maertens (changes requested):
Almost there!

src/openforms/registrations/contrib/zgw_apis/options.py
sergei-maertens: this is the last thing to pick up, barring any CI nopes

sergei-maertens (approved):

PR #4768, opened 2024-10-17
Fix performance regression in FormioData data structure

### LINKED ISSUE #4744: Performance regression in 2.8.0

2.8.0

### Customer reference

Taiga DH 700

2.8.0 appears to be measurable slower compared to 2.7.8 - notably in the `_check_logic` operations.

This manifests in the automated (playwright) tests running into timeouts now. The measured/perceived extra duration for check logic calls is 2-3 seconds, which is not an acceptable user experience.

Unknown - need to investigate with APM and compare performance with 2.7.x

Performance is equal or better than 2.7

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #4744

**Changes**

Optimized key access in the `FormioData` data structure.

Performance results - the last column are the numbers for this PR. The gains are massive compared to `master` and they improve even on 2.7.x.

| Form step                   | version: master (0ea7bfa) | version: 2.7.x | Perf patch |
| --------------------------- | ------------------------- | -------------- | ---------- |
| Inleiding                   | 1540                      | 736            | 448        |
| Verklaring van waarheid     | 1520                      | 751            | 464        |
| Uw gegevens                 | 1727                      | 879            | 551        |
| Uw situatie                 | 1730                      | 827            | 483        |
| Gezinssamenstelling         | 1660                      | 823            | 558        |
| Uw inkomsten                | 1790                      | 991            | 637        |
| Uw inkomsten overzicht      | 1740                      | 836            | 565        |
| Inkomsten partner           | 1840                      | 909            | 585        |
| Inkomsten partner overzicht | 1720                      | 864            | 540        |
| Bezit en schulden           | 1620                      | 859            | 508        |
| Bezit en schulden overzicht | 1600                      | 817            | 637        |
| Post                        | 1560                      | 769            | 485        |
| Overzicht                   | 10350                     | 4850           | 3330       |

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/submission_registration.py
+++ b/src/openforms/registrations/contrib/objects_api/submission_registration.py
@@ -9,7 +9,6 @@ from typing import Any, Generic, Iterator, Literal, TypeVar, cast, override
 from django.db.models import F
 
 import glom
-from glom import PathAccessError
 
 from openforms.authentication.service import AuthAttribute
 from openforms.contrib.objects_api.clients import (
@@ -537,7 +536,7 @@ class ObjectsAPIV2Handler(ObjectsAPIRegistrationHandler[RegistrationOptionsV2]):
         for key, variable in state.variables.items():
             try:
                 submission_value = dynamic_values[key]
-            except PathAccessError:
+            except KeyError:
                 continue
 
             # special casing documents - we transform the formio file upload data into
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -363,7 +363,7 @@ django-sessionprofile==2.0.0
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
     #   django-digid-eherkenning
-django-silk==5.1.0
+django-silk==5.2.0
     # via -r requirements/dev.in
 django-simple-certmanager==2.4.1
     # via
--- a/src/openforms/formio/datastructures.py
+++ b/src/openforms/formio/datastructures.py
@@ -1,6 +1,5 @@
 import re
 from collections import UserDict
-from collections.abc import Hashable
 from typing import Iterator, cast
 
 from glom import PathAccessError, assign, glom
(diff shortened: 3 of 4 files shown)


sergei-maertens:
So one particularly interesting observation is that in the hottest code path and logic rule evaluation there seems to be much more room for improvement.

With the `master` branch, 58% of time was spent in `glom`, and there were about 18,5K calls to it. In this patch, 34% of time is spent in `glom`, and there are still about 15K calls - so only about 3K calls are stripped away. It could be that many of those led to exception handling and that caused the slowdown, but I can't even begin to imagine what performance is left uncovered if we can bring those 15K to 1K calls, for example.

sergei-maertens:
Validated and approved by DH - it's not up to OF 2.6.x levels yet, but a vast improvement over 2.7.x.

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -537,7 +536,7 @@ def get_record_data(
         for key, variable in state.variables.items():
             try:
                 submission_value = dynamic_values[key]
-            except PathAccessError:
+            except KeyError:
```
stevenbal, line 539: Was this changed to KeyError because the variable keys cannot have `.` in them?

stevenbal (approved):

src/openforms/registrations/contrib/objects_api/submission_registration.py
sergei-maertens: no, before it would implicitly rely on `glom` being used internally, which raises `PathAccessError` if you access the key (without checking that it's present). So that was a leaky abstraction.

However, you highlight a bug - this only applies for keys without period, I also need to make sure these path access errors get transformed into a proper KeyError when using dotted keys.

sergei-maertens: heh, it already works correctly because `PathAccessError` subclasses `KeyError`... I've added a test and made things more explicit.

sergei-maertens:
Backports:

* 2.8.x: 6ed8a2d23803c59239f51b9ffd07b746a22a1a6d, bd44c1338604bb0ba659a824e747724103fbf026
* 2.7.x: 63528a8a94ffba2d51ac41c9b8e4155bf95fde57

PR #4770, opened 2024-10-18
Obtain submission price from variable

### LINKED ISSUE #4764: Make it possible to set the submission price from a form variable
### Thema / Theme

Form designer

Currently the submission price is determined by either:

* setting the price of the related product, which is a fixed price
* setting up logic rules with associated prices, but the values are "fixed numbers"

A third variant is desired: point to a form variable holding the price value. This should be limited to variables of the type `number`.

For complex calculations, specifying out the logic rules is tedious and error prone, and another system may be the source of truth for the prices which can be queried using service fetch.

This allows managing the price information in its own respective system, while still being able to handle the payments through Open Forms.

DH would need this still this year, which doesn't fit our release schedule (3.0 is for the end of the year).

If possible, an approach where we can include this in a bugfix release would be greatly appreciated.

Additionally, the priority has been determined as *more* important than the soft-required validation for horecavergunning.

sergei-maertens:
Closes #4764

Needs forward-porting due to new migrations.

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/submissions/tests/test_price_calculation.py
```diff
@@ -0,0 +1,139 @@
...
+            submission.refresh_from_db()
+
+            with self.assertRaisesMessage(
+                InvalidPrice,
+                "No variable 'badVariable.reference' present in the submission data, "
+                "refusing the temptation to guess.",
+            ):
+                get_submission_price(submission)
+
+        invalid_values = (
```
vaszig, line 106: Don't we need the other data types as well?I guess because they would be of the same type (str)?

src/openforms/submissions/pricing.py
```diff
@@ -32,8 +48,32 @@ def get_submission_price(submission: "Submission") -> Decimal:
...
 
     form = submission.form
-    data = submission.data
 
+    #
+    # 1. If a variable key is configured, try that.
+    #
+    try:
+        price = _price_from_variable(submission)
+    except InvalidPrice as exc:
```
vaszig, line 57: When would this be raised?In `_price_from_variable` I see we return `None`, `Decimal(value)` and otherwise we raise an exception. I am a bit confused here, I thought that this exception would be raised before we reach this point and I do not see a test for this tbh. I guess I m missing something here..

src/openforms/submissions/pricing.py
```diff
@@ -51,10 +91,51 @@ def get_submission_price(submission: "Submission") -> Decimal:
...
+        var_key,
+        value,
+    )
+
+    invalid_type_error = InvalidPrice(
+        f"Got an incompatible value type for the price variable '{var_key}': "
+        f"{type(value)}. We require a value that can be cast to a decimal.",
+        variable=var_key,
+    )
+    if not isinstance(value, (str, float, int)) or isinstance(value, bool):
```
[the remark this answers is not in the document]
vaszig: Ok, I got this..

src/openforms/submissions/tests/test_price_calculation.py
sergei-maertens: The rationale behind these types was:

* `boolean` is a subclass of `int`, so that needs special treatment on our end
* `string` *could* be in some cases valid, if it's a number like `123.45`, which can be cast to a decimal, but non-numeric strings will cause crashes
* the implementation applies allow-list rather than block-list logic, so the test is geared towards those edge cases. other types could be included, but the details about how the initial value is changed depending on the data type are a bit tricky and peculiar and distract from what's actually being tested, so I opted to leave those out

src/openforms/submissions/pricing.py
sergei-maertens: E.g. `Decimal("")` results in `decimal.InvalidOperation` in `_price_from_variable`, which is caught and re-raised as `InvalidPrice`

While the UI should only allow selecting variables of numeric type, I don't doubt for a moment that someone will mess with the JSON and end up with prices pointing to a string variable which will be unset/not initialized properly due to broken form logic, and that's why this defensive programming is here - I want to make those error cases explicit as quickly as possible instead of letting the problem bubble through the system.

PR #4797, opened 2024-10-28
Bug/4727 crashing form after changing user variable to array

### LINKED ISSUE #4727: Changing variable type of user variable can crash the form editor

any

### Customer reference

DH 698

When a user variable is changed from: object to array the form builders crashes:

`Uncaught TypeError: n.some is not a function`

it seems the value of the saved previous type causes this crash, since changing it to [] before switching to array does not crash. 

Add a user variable "object" with a value of {} (default)
Save form 
Change user variable type to array

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #4727

**Changes**

Fixes the crashes on user variable dataType change, when a initialValue is defined. This crash happend when changing any dataType to the array dataType, because the component for the array dataType used the `.some` js array function. (It expected that the value was an array, so this function was used without any checks)

To resolve this situation, and possible similar situations in the future, the initialValue now gets transformed when the dataType changes. This transformation could result into losing the previous initialValue. To make sure that the dataType change is intentional, a confirmation message will be shown.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/form-creation-form.js
+++ b/src/openforms/js/components/admin/form_design/form-creation-form.js
@@ -70,6 +70,7 @@ import {
   getUniqueKey,
   parseValidationErrors,
   slugify,
+  transformInitialValue,
   updateKeyReferencesInLogic,
   updateRemovedKeyInLogic,
 } from './utils';
@@ -784,6 +785,14 @@ function reducer(draft, action) {
         draft.formVariables[index][propertyName] = propertyValue;
       }
 
+      // When dataType changes, a data transformation is needed
+      if (propertyName === 'dataType') {
+        draft.formVariables[index]['initialValue'] = transformInitialValue(
+          propertyValue,
+          originalVariable.initialValue
+        );
+      }
+
       // Check if there are errors that need to be reset
       if (draft.formVariables[index].errors) {
         const errorKeys = propertyName === '' ? Object.keys(propertyValue) : [propertyName];
@@ -1229,6 +1238,42 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl, outgoingRequestsUr
       payload: pluginId,
     });
   };
+  const onUserDefinedVariableChange = async (key, propertyName, propertyValue) => {
+    const originalVariable = state.formVariables.find(variable => variable.key === key);
+    // Just dispatch if anything other than dataType changes
+    // or if the initialValue is null/undefined
+    if (
+      propertyName !== 'dataType' ||
+      originalVariable?.initialValue == null ||
+      originalVariable?.initialValue === ''
+    ) {
+      dispatch({
+        type: 'CHANGE_USER_DEFINED_VARIABLE',
+        payload: {key, propertyName, propertyValue},
+      });
+      return;
+    }
+
+    // Check if the dataType change is intentional.
+    if (
+      propertyName === 'dataType' &&
+      !window.confirm(
+        intl.formatMessage({
+          description:
+            'Changing user variable data type and transforming initial value confirmation message',
+          defaultMessage:
+            'Changing the data type requires the initial value to be changed. This will reset the initial value back to the empty value. Are you sure that you want to do this?',
+        })
+      )
+    ) {
+      return;
+    }
+
+    dispatch({
+      type: 'CHANGE_USER_DEFINED_VARIABLE',
+      payload: {key, propertyName, propertyValue},
+    });
+  };
 
   if (loading || state.submitting) {
     return <Loader />;
@@ -1268,8 +1313,8 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl, outgoingRequestsUr
         <div className="fetch-error">
           <FormattedMessage
             description="Generic admin error message"
-            defaultMessage={`Sorry! Something unexpected went wrong.<br></br>Contact your 
-              technical administrator to investigate, or perhaps more information is 
+            defaultMessage={`Sorry! Something unexpected went wrong.<br></br>Contact your
+              technical administrator to investigate, or perhaps more information is
               available in the <link>outgoing request logs</link>.`}
             values={{
               br: () => <br />,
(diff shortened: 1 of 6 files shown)


src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1229,6 +1243,39 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl, outgoingRequestsUr
...
+    // Check if the dataType change is intentional.
+    if (
+      propertyName === 'dataType' &&
+      checkIfInitialValueTransformationIsNeeded(propertyValue, originalVariable) &&
+      !window.confirm(
+        intl.formatMessage({
+          description:
+            'Changing user variable data type and transforming initial value confirmation message',
+          defaultMessage:
+            'Changing the data type requires the initial value to be changed, in some cases this could mean result in the initial value being reset to the default value. Are you sure that you want to do this?',
```
sergei-maertens, line 1267:
```suggestion
            'Changing the data type requires the initial value to be changed. This will reset the initial value back to the empty value. Are you sure that you want to do this?',
```

src/openforms/js/lang/nl.json
```diff
@@ -271,6 +271,11 @@
     "isTranslated": true,
     "originalDefault": "minus"
   },
+  "5zEEiB": {
+    "defaultMessage": "Om het Datatype te veranderen, moet ook de beginwaarde aangepast worden. In sommige gevallen kan dit betekenen dat de beginwaarde wordt veranderd naar de standaard waarde van het Datatype. Weet je zeker dat je dit wilt doen?",
```
sergei-maertens, line 275: this is a bit too technical I think, the english version was simplified a bit.

Also please lower case "datatype"

src/openforms/js/lang/nl.json
```diff
@@ -2622,6 +2627,11 @@
     "description": "JSON editor: object entry has empty key name",
     "originalDefault": "Set a key name before you can configure this variable"
   },
+  "vowYMF": {
+    "defaultMessage": "Weet je zeker dat je het Datatype wilt veranderen?",
+    "description": "Changing user variable data type and transforming initial value confirmation title",
+    "originalDefault": "Are you sure that you want to change the data type?"
+  },
```
sergei-maertens, line 2634: where is this message coming from? I only see one new `intl.formatMessage` call in this PR

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -57,6 +57,74 @@ const checkKeyChange = (mutationType, newComponent, oldComponent) => {
   return newComponent.key !== oldComponent.key;
 };
 
+const checkIfInitialValueTransformationIsNeeded = (newType, originalVariable) => {
```
sergei-maertens, line 60: we only use the `initialValue` of the variable - do we need to pass the entire variable then? Looks like `originalValue` would be sufficient as an argument.

sergei-maertens, line 60: I can't really picture situations where transformation is NOT needed, except for cases that are already broken like an array of strings but the variable type is set to "string".

Shouldn't we just always reset? Like even from string -> datetime: "foo" is not a valid datetime anyway, but a valid string, so maybe we can cut out some complexity?

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -57,6 +57,74 @@ const checkKeyChange = (mutationType, newComponent, oldComponent) => {
...
+      return false;
+  }
+};
+
+const transformInitialValue = (newType, originalVariable) => {
+  switch (newType) {
+    case 'array':
+      if (
+        originalVariable?.initialValue == null ||
+        typeof originalVariable.initialValue === 'object'
```
sergei-maertens, line 97: `Array` is also object so I'm not sure how robust this is.

src/openforms/js/components/admin/form_design/utils.js
```diff
@@ -57,6 +57,74 @@ const checkKeyChange = (mutationType, newComponent, oldComponent) => {
...
+      ) {
+        return [];
+      }
+      return [originalVariable.initialValue];
+
+    case 'boolean':
+      return undefined;
+
+    case 'int':
+      return Number.parseInt(originalVariable.initialValue);
```
sergei-maertens, line 107: if `initialValue` is `'a'`, this will result in NaN and then we have an entirely different class of problems. I'm in favour of just always resetting to default empty value when you change type.

src/openforms/js/components/admin/form_design/utils.js
robinmolen: Ah yeah, it should have a `Array.isArray(...)` to make sure..

src/openforms/js/components/admin/form_design/utils.js
robinmolen: Mhh, i that thought also crossed my mind.. Maybe it is indeed best to always reset..

The only cases a transformation could be useful/nice is when a value changes to datatype `object` or when a date becomes a datetime (or something simular..). Or a simple value changes to array..

sergei-maertens: let's just keep it simple and always reset then!

src/openforms/js/lang/nl.json
robinmolen: Ah yeah, I added this as the title for the new confirmation modal. At this moment this message isn't being used.
I'll remove it

src/openforms/js/components/admin/form_design/form-creation-form.js
robinmolen: I've changed this message and the dutch translation

src/openforms/js/components/admin/form_design/utils.js
robinmolen: I've removed it 👍

src/openforms/js/components/admin/form_design/utils.js
robinmolen: Yeah for all dataType's the transformation now just resets the value. So array becomes `[]`, object becomes `{}`, bool, int, and float return to `undefined`, and string, date, datetime and time return to `''`.

src/openforms/js/components/admin/form_design/utils.js
robinmolen: Yeah, the entire object is indeed not needed.. I've changed it 👍

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.8.x: 3f955b622184719155a92d6e340e2013e3368e67
* 2.7.x: 43272cd17e4966106a1102602b969def44114849

PR #4799, opened 2024-10-28
:alien: [#4693] Integrate Objects API prefill modal with backend

### LINKED ISSUE #4693: Product prefill configuration modal: integrate with backend
Remaining points from https://github.com/open-formulieren/open-forms/pull/4610 (issue https://github.com/open-formulieren/open-forms/issues/4608) to ensure the PR doesn't hang for too long
- [x] once https://github.com/open-formulieren/open-forms/pull/4620/files is merged
  - [x] implement save
  - [x] uses `objecttypeUuid` instead of `objecttype` (which is used by the registration endpoint and in the ObjectType fields)
- [x] modal objecttype select menuPlacement makes it open top, leaving 5th option under header
- [x] implement copy from button
- [x] default selected objecttype doesnt cause options to be populated because wrong objecttype is loaded?
- [x] check if on error data is still filled in
- [x] failure when fetching objecttypes doesnt go away when switching group
- [x] add storybook for copy button

stevenbal:
Closes #4693 

**Changes**

* Integrate Objects API prefill modal with backend

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


stevenbal:
@sergei-maertens I'm trying to fix the issue with the react-select menu being clipped, but I'm not sure how to fix it. I've tried setting a z-index, but that alone doesn't fix the issue. If I try using `overflow: visible`, I do see the full menu, but the rest of the modal breaks :/

current situation
image
with overflow
image

Any ideas on how this can be fixed? Changing the menuPlacement to `bottom` will likely cause the same issue in the future, if the list of choices gets long enough

stevenbal:
@sergei-maertens I've made the placement bottom and set a maxMenuHeight for the selects used in the modal to avoid overflow issues (https://github.com/open-formulieren/open-forms/pull/4799/commits/efb307d44275fbfdc6ab70e1dd87a68e8e181a18)

src/openforms/js/components/admin/form_design/variables/prefill/ObjectsAPIFields.js
```diff
@@ -147,11 +156,49 @@ const ObjectsAPIFields = ({errors}) => {
...
+                  backend will clear the existing configuration. Are you sure you want to continue?`,
+                })
+              );
+              if (confirmed) {
+                setFieldValue('prefillOptions.objectsApiGroup', backend.options.objectsApiGroup);
+                setFieldValue('prefillOptions.objecttypeUuid', backend.options.objecttype);
+                setFieldValue(
+                  'prefillOptions.objecttypeVersion',
+                  backend.options.objecttypeVersion
+                );
```
sergei-maertens, line 186: you can also `setValues((prevValues) => ...)` to do this in a single callback and reduce the risk of race conditions

stevenbal: I'll add it, where is this mapping configured for the registration backend actually? I do not see them in the objects API registration modal

src/openforms/js/components/admin/forms/objects_api/ObjectTypeSelect.js
```diff
@@ -94,6 +95,7 @@ const ObjectTypeSelect = ({
             const okToProceed = onChangeCheck === undefined || onChangeCheck();
             if (okToProceed) setValue(selectedOption.value);
           }}
+          {...selectProps}
```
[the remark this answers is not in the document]
robinmolen: Yesterday this issue popped back into my mind, and took a look. Maybe [Portaling](https://react-select.com/advanced#portaling) is a good solution. It allows us to move the dropdown to a different part of the dom tree, solving parent-child styling issues (and would probably save us some z-index magic styling).

(As i'm writing this, I do wonder if this would be a big accessibility issue... 🤔. I haven't looked yet how this works for screen readers and tab navigation... (I think the menu receives focus when opened, so maybe it won't be an issue...?))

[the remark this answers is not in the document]
sergei-maertens: I can't imagine it *not* messing with the tab order :/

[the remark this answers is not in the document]
sergei-maertens: Keyboard navigation is fine, and Robin verified the screenreader support which also appears to be fine.

Portaling appears to be the only viable solution here, so it's been adapted, let's see if it breaks anything else :(

src/openforms/js/components/admin/forms/ReactSelect.js
```diff
@@ -82,6 +82,7 @@ const SelectWithoutFormik = ({name, options, value, className, onChange, ...prop
     classNamePrefix="admin-react-select"
     styles={styles}
     menuPlacement="auto"
+    menuPortalTarget={document.body}
```
sergei-maertens, line 85: this is too simple as it breaks storybook, which renders in another DOM node and the `within(canvasElement)` queries break. Instead, we should probably set up a custom context that defaults to `document.body`, but which we can override in storybook to point it to the right node, similar to what we do with the modals.

sergei-maertens:
Rebased on master - I will now look at the state again and do a thorough review!

sergei-maertens:
I'm not happy yet with the "copy configuration" button and this is a tough nut to crack. After some discussion/debate with Joeri, we've settled on something like this:

1. When you open the prefill modal, you get the current state (with dropdowns likely empty if this is the initial configuration:
    
    image

    * A link is injected next to the 'plugin' field. This will require some interesting changes which we can discuss tomorrow, since you currently can't provide additional children for the plugin field... We can add a tooltip here that describes how it works/what it does for better documentation.
    * The registration backend dropdown is not visible at all

2. The user clicks the link to copy the settings - this makes the registration dropdown visible:
    
    image

3. Clicking the link again just collapse the form row again
4. Selecting a backend and then clicking the 'copy' button performs the copy as it is right now, but also automatically collapses the form row again at the end to declutter the UI again.

Please also set up the styling properly so that the link and button are nicely aligned.

src/openforms/js/components/admin/form_design/variables/prefill/PrefillConfigurationForm.js
```diff
@@ -23,14 +23,16 @@ const PrefillConfigurationForm = ({
...
     objectsApiGroup: '',
-    objecttype: '',
+    objecttypeUuid: '',
     objecttypeVersion: null,
     variablesMapping: [],
-  },
-  errors,
-}) => {
+  };
+  options = {...defaults, ...options};
```
sergei-maertens, line 35: I think this needs to be pushed down to `ObjectsAPIFields` and have it set the default values on initial render - you can get the `initialValues` from the Formik state. This generic component should not be polluted with plugin-specific options and we want to keep the relevant code co-located.

src/openforms/js/components/admin/forms/ReactSelect.js
sergei-maertens: we now have this modal context for storybook from @robinmolen, so we can apply the same pattern.

sergei-maertens (commented):
We should go through these remarks tomorrow :)

Let's get this PR wrapped up before we tackle the other one.

sergei-maertens (approved):
Let's merge this so we can get it tested in the alpha release!

PR #4803, opened 2024-10-30
Add Product to the ZGW API registration

### LINKED ISSUE #4796: Als gemeente wil ik na het selecteren van een zaaktype, ook het product selecteren en meesturen met de inzending
### Thema / Theme

Frontend

In the ZGW API configuration for a form, allow configuration of the product. Available products can be retrieved from the Zaaktype (producten_en_diensten).

**Tasks**

* [x] When the zaaktype is configured on form-level, retrieve products and show them in the admin as dropdown
* [x] When the form is submitted, also submit the product in the Zaak-POST.

Allows a single zaaktype to be used with multiple products.

_No response_

robinmolen:
Closes #4796 

**Changes**

Added product as a configuration option to the ZGW API registration. The product select gets populated from the selected case type. If there are multiple versions of the case type, then the case type version that is active at the moment of configuring is chosen.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/docker/docker-compose.rx-mission.yml
+++ b/docker/docker-compose.rx-mission.yml
@@ -1,0 +1,15 @@
+version: '3.8'
+
+name: rx-mission
+
+services:
+  flask_app:
+    build: ./rx-mission
+    ports:
+      - "80:80"
+    volumes:
+      - ./rx-mission/:/app/
+
+networks:
+  open-forms-dev:
+    name: open-forms-dev
--- a/docker/rx-mission/app.py
+++ b/docker/rx-mission/app.py
@@ -1,0 +1,23 @@
+import json
+
+from flask import Flask, jsonify
+
+app = Flask(__name__)
+products = {}
+
+
+@app.route("/product/<string:product_uuid>", methods=["GET"])
+def handle_request(product_uuid):
+    if product_uuid not in products:
+        return jsonify({"message": "Product not found"}), 404
+
+    product = products[product_uuid]
+    return jsonify(product)
+
+
+if __name__ == "__main__":
+    with open("./fixtures/rx-mission-products.json", encoding="utf-8") as json_file:
+        parsed_json = json.load(json_file)
+        products = {product["id"]: product for product in parsed_json["products"]}
+
+    app.run(host="0.0.0.0", port=80, debug=True)
--- a/src/openforms/contrib/zgw/clients/catalogi.py
+++ b/src/openforms/contrib/zgw/clients/catalogi.py
@@ -48,6 +48,7 @@ class CaseType(TypedDict):
     beginGeldigheid: str  # ISO 8601 date string
     eindeGeldigheid: NotRequired[str | None]  # ISO 8601 date string or empty
     concept: NotRequired[bool]
+    productenOfDiensten: list[str]  # URL pointers to products
     informatieobjecttypen: NotRequired[list[str]]  # URL pointers to document types
 
 
--- a/src/openforms/contrib/zgw/clients/zaken.py
+++ b/src/openforms/contrib/zgw/clients/zaken.py
@@ -22,6 +22,7 @@ class ZakenClient(NLXClient):
         vertrouwelijkheidaanduiding: str = "",
         payment_required: bool = False,
         existing_reference: str = "",
+        product_url: str = "",
         **overrides,
     ):
         today = get_today()
@@ -31,6 +32,7 @@ class ZakenClient(NLXClient):
             "verantwoordelijkeOrganisatie": bronorganisatie,
             "registratiedatum": today,
             "startdatum": today,
+            "productenOfDiensten": [product_url] if product_url else [],
             "omschrijving": "Zaak naar aanleiding van ingezonden formulier",
             "toelichting": "Aangemaakt door Open Formulieren",
             "betalingsindicatie": "nog_niet" if payment_required else "nvt",
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/BasicOptionsFieldset.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/BasicOptionsFieldset.js
@@ -5,23 +5,11 @@ import {useAsync} from 'react-use';
 
 import useConfirm from 'components/admin/form_design/useConfirm';
 import Fieldset from 'components/admin/forms/Fieldset';
-import {getCatalogueOption, groupAndSortCatalogueOptions} from 'components/admin/forms/zgw';
+import {getCatalogueOption} from 'components/admin/forms/zgw';
 import ErrorBoundary from 'components/errors/ErrorBoundary';
-import {get} from 'utils/fetch';
 
 import {CaseTypeSelect, CatalogueSelect, ZGWAPIGroup} from './fields';
-
-// Data fetching
-
-const CATALOGUES_ENDPOINT = '/api/v2/registration/plugins/zgw-api/catalogues';
-
-const getCatalogues = async apiGroupID => {
-  const response = await get(CATALOGUES_ENDPOINT, {zgw_api_group: apiGroupID});
-  if (!response.ok) {
-    throw new Error('Loading available catalogues failed');
-  }
-  return groupAndSortCatalogueOptions(response.data);
-};
+import {getCatalogues} from './utils';
 
 // Components
 
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
@@ -45,6 +45,7 @@ const ZGWOptionsForm = ({name, label, schema, formData, onChange}) => {
         zaakVertrouwelijkheidaanduiding: '',
         medewerkerRoltype: '',
         propertyMappings: [],
+        productUrl: '',
         // Ensure that this is explicitly set to null instead of undefined,
         // because the field is required by the serializer
         objectsApiGroup: null,
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.stories.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.stories.js
@@ -9,7 +9,7 @@ import {
 } from 'components/admin/form_design/story-decorators';
 
 import ZGWFormFields from './ZGWOptionsFormFields';
-import {mockCaseTypesGet, mockCataloguesGet} from './mocks';
+import {mockCaseTypesGet, mockCataloguesGet, mockProductsGet} from './mocks';
 
 const NAME = 'form.registrationBackends.0.options';
 
@@ -54,7 +54,7 @@ export default {
   },
   parameters: {
     msw: {
-      handlers: [mockCataloguesGet(), mockCaseTypesGet()],
+      handlers: [mockCataloguesGet(), mockCaseTypesGet(), mockProductsGet()],
     },
   },
 };
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
@@ -1,0 +1,79 @@
+import {useFormikContext} from 'formik';
+import PropTypes from 'prop-types';
+import {FormattedMessage} from 'react-intl';
+import useAsync from 'react-use/esm/useAsync';
+
+import Field from 'components/admin/forms/Field';
+import FormRow from 'components/admin/forms/FormRow';
+import ReactSelect from 'components/admin/forms/ReactSelect';
+import {get} from 'utils/fetch';
+
+const PRODUCTS_ENDPOINT = '/api/v2/registration/plugins/zgw-api/products';
+
+const getAvailableProducts = async (apiGroupID, catalogueUrl, caseTypeIdentification) => {
+  const response = await get(PRODUCTS_ENDPOINT, {
+    zgw_api_group: apiGroupID,
+    catalogue_url: catalogueUrl,
+    case_type_identification: caseTypeIdentification,
+  });
+  if (!response.ok) {
+    throw new Error('Loading available case type products failed');
+  }
+  return response.data.map(({url, description}) => ({
+    value: url,
+    label: description || url,
+  }));
+};
+
+const ProductSelect = ({catalogueUrl = ''}) => {
+  const {
+    values: {zgwApiGroup = null, caseTypeIdentification = ''},
+  } = useFormikContext();
+
+  const {
+    loading,
+    value: options = [],
+    error,
+  } = useAsync(async () => {
+    if (!zgwApiGroup || !catalogueUrl || !caseTypeIdentification) return [];
+    return await getAvailableProducts(zgwApiGroup, catalogueUrl, caseTypeIdentification);
+  }, [zgwApiGroup, catalogueUrl, caseTypeIdentification]);
+  if (error) throw error;
+
+  return (
+    <FormRow>
+      <Field
+        name="productUrl"
+        required={false}
+        label={
+          <FormattedMessage
+            description="ZGW APIs registration options 'product' label"
+            defaultMessage="Product"
+          />
+        }
+        helpText={
+          <FormattedMessage
+            description="ZGW APIs registration options 'product' helpText"
+            defaultMessage="This product will be set on the created case."
+          />
+        }
+        noManageChildProps
+      >
+        <ReactSelect
+          name="productUrl"
+          options={options}
+          isLoading={loading}
+          isDisabled={!caseTypeIdentification}
+          required={false}
+          isClearable
+        />
+      </Field>
+    </FormRow>
+  );
+};
+
+ProductSelect.propTypes = {
+  catalogueUrl: PropTypes.string,
+};
+
+export default ProductSelect;
--- a/src/openforms/registrations/contrib/zgw_apis/api/views.py
+++ b/src/openforms/registrations/contrib/zgw_apis/api/views.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass, field
 
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.utils import extend_schema, extend_schema_view
@@ -7,16 +8,22 @@ from rest_framework import authentication, permissions
 from rest_framework.views import APIView
 
 from openforms.api.views.mixins import ListMixin
-from openforms.contrib.zgw.api.serializers import CaseTypeSerializer
+from openforms.contrib.zgw.api.serializers import (
+    CaseTypeProductSerializer,
+    CaseTypeSerializer,
+)
 from openforms.contrib.zgw.api.views import (
     BaseCatalogueListView,
     BaseInformatieObjectTypenListView,
 )
+from openforms.contrib.zgw.products_and_services import resolve_case_type_products
+from openforms.utils.date import datetime_in_amsterdam
 
 from .filters import (
     APIGroupQueryParamsSerializer,
     ListCaseTypesQueryParamsSerializer,
     ListInformatieObjectTypenQueryParamsSerializer,
+    ListProductsQueryParamsSerializer,
 )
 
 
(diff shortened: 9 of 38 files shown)


src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsFormFields.stories.js
```diff
@@ -9,7 +9,7 @@ import {
 } from 'components/admin/form_design/story-decorators';
 
 import ZGWFormFields from './ZGWOptionsFormFields';
-import {mockCaseTypesGet, mockCataloguesGet} from './mocks';
+import {mockCaseTypesGet, mockCataloguesGet, mockProductsGet} from './mocks';
```
robinmolen, line 12: The `mockProductsGet` isn't used yet, causing storybook errors

robinmolen: Added it

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -227,6 +234,37 @@ def find_case_types(
...
+                description = ""
+
+                response = requests.get(
+                    product_uri, headers={"Content-Type": "application/json"}
+                )
+                if response.ok:
+                    description = response.json().get("description")
+                products.append(Product(uri=product_uri, description=description))
+
+        return products
```
sergei-maertens, line 266: I'm trying to figure out if this method should exist on the Catalogi client or not. Technically, the product/service URLs are not part of the Catalogi API, so it should be a different client because it's a different domain. However, there is no formal specification yet, so we don't really have a good place for the code.

So, I'd propose this:

```py
# in file src/openforms/contrib/zgw/products_and_services.py
"""
Support resolving products and services referenced by case types.

.. warning:: Products and services integration is experimental. There is no standard for it yet.
"""

def resolve_case_type_products(case_type_versions: list[CaseType]) -> list[Product]:
    all_urls: set[str] = set(sum(version["productenOfDiensten"] for version in case_type_versions, []))
    ...  # look up content of each URL
    return products
```

and then in `src/openforms/contrib/zgw/service.py` expose this helper:

```py
from .products_and_services import resolve_case_type_products
```

You can then combine these things at the places where you need them:

```py
case_type_versions = client.find_case_types(...)
products = resolve_case_type_products(case_type_versions) if case_type_versions else []
```

And this way, we keep the clients/domains pure.

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -227,6 +234,37 @@ def find_case_types(
...
+            case_versions
+            and len(case_versions) > 0
+            and "productenOfDiensten" in case_versions[0]
+        ):
+            for product_uri in case_versions[0]["productenOfDiensten"]:
+                description = ""
+
+                response = requests.get(
+                    product_uri, headers={"Content-Type": "application/json"}
+                )
```
sergei-maertens, line 261: Couple remarks here:

1. This is basically server-side request forgery, since the code blindly trusts the URL returned from the zaaktype response
2. using bare `requests` will never add any authorizations configured in the service admin, and will likely return 403 for those product URLs since these APIs are not public
3. we don't send a request body, so `Content-Type` header is not relevant. I think you mean to say `Accept: application/json`

You can resolve 1 and 2 by making use of our client/service libraries: 

```
from zgw_consumers.models import Service
from zgw_consumers.client import build_client

service = Service.get_service(url)
client = build_client(service)
```

`get_service` looks up the URL in the configured services in the DB, so it has all the necessary credentials and configuration. `build_client` understands that information and gives you an [ape-pie](https://pypi.org/project/ape-pie/) client, which is really just a `requests.Session` subclass with the footguns removed.

This means that any API roots that aren't configured in the admin won't yield a service, and we are not vulnerable to SSRF (since we crash/abort instead).

src/openforms/contrib/zgw/clients/zaken.py
```diff
@@ -22,6 +22,7 @@ def create_zaak(
         vertrouwelijkheidaanduiding: str = "",
         payment_required: bool = False,
         existing_reference: str = "",
+        productenOfDiensten: list[str] = [],
```
sergei-maertens, line 25: *never ever* use mutable default arguments in python (like lists and dicts) - if the function/method mutates them, the default argument for the next call is mutated.

instead, use `None` and handle the empty list default (if needed) in the function body.

src/openforms/js/components/admin/form_design/registrations/zgw/BasicOptionsFieldset.js
sergei-maertens, line 1: the frontend (and backend) code has considerable overlap with https://github.com/open-formulieren/open-forms/pull/4817. I'll need to handle that in my PR, this will probably get merged first.

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
```diff
@@ -41,6 +41,7 @@ const ZGWOptionsForm = ({name, label, schema, formData, onChange}) => {
         zaakVertrouwelijkheidaanduiding: '',
         medewerkerRoltype: '',
         propertyMappings: [],
+        product: '',
```
sergei-maertens, line 44: if we only support a single product selection, why do we have to pass a list of product URLs to the case creation? wouldn't it make more sense in the backend to only accept a single URL `str` argument instead of a list then?

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
```diff
@@ -0,0 +1,99 @@
...
+    description: description,
+  }));
+};
+
+const ProductOption = props => {
+  const {label, description} = props.data;
+  return (
+    <components.Option {...props}>
+      {description && <span className="catalogi-type-option">{description}</span>}
+      <code>{label}</code>
```
sergei-maertens, line 35: I don't really see much benefit in displaying the full URI here. Instead, if there's no description, you can use the URI as fallback, but otherwise just showing the description is sufficient.

Having `getAvailableProducts` return `{value: uri, label: description || uri}` seems like the much simpler approach here, then you also don't need a custom option component.

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
```diff
@@ -0,0 +1,99 @@
...
+    </components.Option>
+  );
+};
+
+const ProductSingleValue = props => {
```
sergei-maertens, line 40: would be obsoleted with the above remark

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
```diff
@@ -0,0 +1,99 @@
...
+  const {label, description} = props.data;
+  return (
+    <components.SingleValue {...props}>{description ? description : label}</components.SingleValue>
+  );
+};
+
+const ProductSelect = ({catalogueUrl = ''}) => {
+  const {
+    values: {zgwApiGroup = null, caseTypeIdentification = null},
```
sergei-maertens, line 49:
```suggestion
    values: {zgwApiGroup = null, caseTypeIdentification = ''},
```
we're not using typescript yet, but still it's good practice to think about the data types :) this field is not nullable, ever.

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
```diff
@@ -0,0 +1,99 @@
...
+          isClearable
+        />
+      </Field>
+    </FormRow>
+  );
+};
+
+ProductSelect.propTypes = {
+  caseTypeIdentification: PropTypes.string,
+};
```
sergei-maertens, line 97: proptypes don't match the actual props

src/openforms/registrations/contrib/zgw_apis/api/views.py
```diff
@@ -73,6 +80,49 @@ def get_objects(self):
...
+        filter_serializer = ListProductsQueryParamsSerializer(
+            data=self.request.query_params
+        )
+        filter_serializer.is_valid(raise_exception=True)
+
+        catalogue_url = filter_serializer.validated_data["catalogue_url"]
+        case_type_identification = filter_serializer.validated_data[
+            "case_type_identification"
+        ]
+        version_valid_on = datetime_in_amsterdam(timezone.now()).date()
```
sergei-maertens, line 112:
```suggestion
        today = datetime_in_amsterdam(timezone.now()).date()
```

src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -72,6 +72,14 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
         ),
         default="",
     )
+    product = serializers.CharField(
```
sergei-maertens, line 75: Should be URLField, no?

```suggestion
    product = serializers.URLField(
```

src/openforms/registrations/contrib/zgw_apis/typing.py
```diff
@@ -32,6 +32,7 @@ class RegistrationOptions(TypedDict):
     zgw_api_group: ZGWApiGroupConfig
     catalogue: NotRequired[CatalogueOption]
     case_type_identification: str
+    product: NotRequired[str]
```
sergei-maertens, line 35: if you do `default=""` on the serializer, you can drop the `NotRequired` and that makes code less verbose

```suggestion
    product: str  # URL reference to a product in the case type
```

sergei-maertens (changes requested):
I'd suggest renaming `product` -> `product_url` everywhere so that it's clear we're dealing with a URL. Especially because we're moving literal URL-configuration options to their semantic 'description' and resolve URLs dynamically, I suspect we will go through this at some point too (where maybe a product slug/description will be stored instead of the exact URL).

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
robinmolen: Mhh okay check. Yeah I thought it might be nice to see which product it precisely is..

Yeah if only the description or the url is shown, then that would be a better approach

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
robinmolen: Yeah, you're right. I've fixed it

src/openforms/registrations/contrib/zgw_apis/api/views.py
robinmolen: Done

src/openforms/registrations/contrib/zgw_apis/options.py
robinmolen: Ah yeah, thats a lot better. I've also renamed it to `product_url`

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -48,9 +49,15 @@ class CaseType(TypedDict):
     beginGeldigheid: str  # ISO 8601 date string
     eindeGeldigheid: NotRequired[str | None]  # ISO 8601 date string or empty
     concept: NotRequired[bool]
+    productenOfDiensten: NotRequired[list[str]]  # URL pointers to products
     informatieobjecttypen: NotRequired[list[str]]  # URL pointers to document types
 
 
+class Product(TypedDict):
+    uri: str
+    description: NotRequired[str]
```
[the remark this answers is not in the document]
robinmolen: I've added it

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -48,9 +49,15 @@ class CaseType(TypedDict):
     beginGeldigheid: str  # ISO 8601 date string
     eindeGeldigheid: NotRequired[str | None]  # ISO 8601 date string or empty
     concept: NotRequired[bool]
+    productenOfDiensten: NotRequired[list[str]]  # URL pointers to products
```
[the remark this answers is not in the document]
robinmolen: Ah chips, i thought i saw somewhere that it was option.. My bad.

Good catch!

docker/docker-compose.open-zaak.yml
```diff
@@ -64,6 +64,13 @@ services:
     networks:
       - open-forms-dev
 
+  flask_app:
+    build: ./open-zaak
+    ports:
+      - "80:80"
+    volumes:
+      - ./open-zaak/:/app/
```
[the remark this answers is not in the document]
robinmolen: Ah yeah fair enough. I've added a new docker env for the `rx_mission`

src/openforms/contrib/zgw/clients/catalogi.py
robinmolen: Yeah this thought also crossed my mind. I decided this would be an okay place, seeing how it uses the catalogi service to fetch the case_type data, but fetching the products was a bit weird..

I've moved this logic like you suggested, and removed the `find_case_types_products` function. The view now first fetches the case type versions, and then gets the related products with extra information

src/openforms/contrib/zgw/clients/catalogi.py
robinmolen: Yeah this is 1000% better. I wasn't sure how i could/should do requests, but yeah `requests.get()` was pretty un-safe...

I've changed it to this service/client setup, which seems to work fine

src/openforms/js/components/admin/form_design/registrations/zgw/ZGWOptionsForm.js
robinmolen: The backend already expected a single product, but i've changed it so that the `ZakenClient` turns it into an array before assigning it to the `productenOfDiensten` attribute

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
robinmolen: Yep, this has been removed

src/openforms/registrations/contrib/zgw_apis/typing.py
robinmolen: I've changed this. The test code now does give warnings (in the editor, not while running the tests) that `product_url` (i've also renamed it) is missing. So to keep the editor UI clean, i've added `"product_url": ""` to some tests

sergei-maertens: yes, good - another PR also enabled/enables type checking on the test modules to make those things fail CI

we want to keep types as simple as possible and that usually means that they're strict rather than lax :)

docker/docker-compose.rx-mission.yml
sergei-maertens, line 1: I think we need a disclaimer that this is not the actual Rx.Mission product but a mock service intended to replicate some interfaces, and that this is in no way affiliated with or endorsed by Roxit.

docker/rx-mission/app.py
```diff
@@ -0,0 +1,19 @@
...
+
+from flask import Flask, jsonify
+
+app = Flask(__name__)
+
+
+@app.route("/product/<string:product_id>", methods=["GET"])
+def handle_request(product_id):
+    with open("./fixtures/rx-mission-products.json", encoding="utf-8") as json_file:
+        parsed_json = json.load(json_file)
```
sergei-maertens, line 11: possibly we can do this in the `if __name__` block so that it's performed once at startup time instead of for every request?

docker/rx-mission/app.py
```diff
@@ -0,0 +1,19 @@
...
+
+    for product in parsed_json["products"]:
+        if product["id"] == product_id:
+            return jsonify(product)
```
sergei-maertens, line 15: if you parse and process the objects once you can use a simple hash map lookup:

```py
# module level
products = {product["id"]: product for product in parsed_json["products"]}

# in request handler
product = products[product_id]
return jsonify(product)
```

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -48,9 +48,17 @@ class CaseType(TypedDict):
     beginGeldigheid: str  # ISO 8601 date string
     eindeGeldigheid: NotRequired[str | None]  # ISO 8601 date string or empty
     concept: NotRequired[bool]
+    productenOfDiensten: list[str]  # URL pointers to products
     informatieobjecttypen: NotRequired[list[str]]  # URL pointers to document types
 
 
+class Product(TypedDict):
```
sergei-maertens, line 55: I'd move this to `src/openforms/contrib/zgw/products_and_services.py` too

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
```diff
@@ -0,0 +1,80 @@
...
+        label={
+          <FormattedMessage
+            description="ZGW APIs registration options 'product' label"
+            defaultMessage="Product"
+          />
+        }
+        helpText={
+          <FormattedMessage
+            description="ZGW APIs registration options 'product' helpText"
+            defaultMessage="This product will be send allong form submits."
```
sergei-maertens, line 58:
```suggestion
            defaultMessage="This product will be set on the created zaak."
```

src/openforms/registrations/contrib/zgw_apis/tests/test_api_endpoints.py
```diff
@@ -265,3 +266,83 @@ def test_retrieve_filter_by_catalogus(self):
...
+                ),
+                "case_type_identification": "ZT-001",
+            },
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        results = response.json()
+        self.assertEqual(len(results), 1)
+
+        expectedProducts = [
```
sergei-maertens, line 342: please use snake_case for variable names

src/openforms/registrations/contrib/zgw_apis/tests/test_api_endpoints.py
```diff
@@ -265,3 +266,83 @@ def test_retrieve_filter_by_catalogus(self):
...
+        invalid_param_names = {
+            item["name"] for item in response.json()["invalidParams"]
+        }
+        self.assertEqual(
+            invalid_param_names,
+            {"zgwApiGroup", "catalogueUrl", "caseTypeIdentification"},
+        )
+
+    @freeze_time("2024-10-31T18:00:00+02:00")
+    def test_find_case_types_deduplicates_versions(self):
```
sergei-maertens, line 322: I don't see how this test name relates to the endpoint being tested?

sergei-maertens (changes requested):

docker/docker-compose.rx-mission.yml
robinmolen: I've added a README to the docker environment, describing that the environment is a dummy set-up that very loosely depicts Roxit interfaces and products.

docker/rx-mission/app.py
robinmolen: I've moved the reading and processing of the json file to the `if __name__` block

docker/rx-mission/app.py
robinmolen: Yeah that's a bit nicer. I've changed it

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -214,6 +214,7 @@ def pre_register_submission(
                 vertrouwelijkheidaanduiding=options.get(
                     "zaak_vertrouwelijkheidaanduiding", ""
                 ),
+                product_url=options.get("product_url", ""),
```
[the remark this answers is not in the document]
robinmolen: Ah yeah ofcource, by the `default=""` in the serializer. I'm changing it

src/openforms/registrations/contrib/zgw_apis/api/views.py
```diff
@@ -73,6 +81,60 @@ def get_objects(self):
         return case_types
 
 
+@dataclass
+class Product:
+    url: str
+    description: Optional[str]
```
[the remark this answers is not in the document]
robinmolen: Ah check, changing it

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ProductSelect.js
robinmolen: Yeah thats a lot better..

src/openforms/registrations/contrib/zgw_apis/api/views.py
```diff
@@ -73,6 +81,60 @@ def get_objects(self):
...
+            case_type_products = (
+                resolve_case_type_products(case_type_versions)
+                if case_type_versions
+                else []
+            )
+            for case_type_product in case_type_products:
+                products.append(
+                    Product(
+                        url=case_type_product["url"],
+                        description=case_type_product["description"],
```
sergei-maertens, line 131: Can you address this type checker error? I would opt to make the key required in the product, if it's missing from the upstream API, you can default it to an empty string.

sergei-maertens (approved):
I'm not 100% happy with how the `catalogueUrl` now is passed around/calculated, but I won't let it block this PR since my #4606 UX ticket will touch that code anyway, so I'll play around with it in my future PR.

Once the type checking error is addressed, we can merge this.

src/openforms/registrations/contrib/zgw_apis/api/views.py
robinmolen: Mhh, yeah making it required is also an option..
I was thinking `description=case_type_product.get("description", "")`

PR #4812, opened 2024-11-06
Broken react select styling for large values

### LINKED ISSUE #4802: React select styling issue with large values

3.0.0-alpha.0

### Customer reference

_No response_

When the ReactSelect has a large value selected, the help icon will be pushed away. I think the element containing the selected value doesn't have a max-width, and just grows outside the ReactSelect element. 

Screenshot 2024-10-30 at 12 31 49

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #4802

**Changes**

- Fixing the bug by ensuring that the `.admin-react-select` element can't grow beyond its max-inline-size
- Replaced the `DocumentTypes` and `CaseTypeSelect` select option html labels with simple text labels. The html labels are now added as a custom Option component for the ReactSelect. This means that the html labels will only be used when the option is presented in the dropdown, otherwise the simple text label will be used.
- Updated the storybook mock data to showcase large option labels for DocumentTypes and CaseTypeSelect

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/DocumentTypes.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/DocumentTypes.js
@@ -3,6 +3,7 @@ import {useField, useFormikContext} from 'formik';
 import PropTypes from 'prop-types';
 import {useContext, useEffect} from 'react';
 import {FormattedMessage, useIntl} from 'react-intl';
+import {components} from 'react-select';
 import {useAsync, usePrevious} from 'react-use';
 
 import {FeatureFlagsContext} from 'components/admin/form_design/Context';
@@ -42,7 +43,17 @@ const getDocumentTypes = async (apiGroupID, catalogueUrl) => {
   const documentTypes = response.data.sort((a, b) => a.omschrijving.localeCompare(b.omschrijving));
   return documentTypes.map(({omschrijving, isPublished}) => ({
     value: omschrijving,
-    label: (
+    label: omschrijving,
+    isPublished: isPublished,
+  }));
+};
+
+// Components
+
+const DocumentTypeSelectOption = props => {
+  const {isPublished, label} = props.data;
+  return (
+    <components.Option {...props}>
       <span
         className={classNames('catalogi-type-option', {
           'catalogi-type-option--draft': !isPublished,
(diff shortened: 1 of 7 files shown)


src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/DocumentTypes.js
```diff
@@ -1,8 +1,9 @@
 import classNames from 'classnames';
 import {useField, useFormikContext} from 'formik';
 import PropTypes from 'prop-types';
-import {useContext, useEffect} from 'react';
+import React, {useContext, useEffect} from 'react';
```
sergei-maertens, line 4: we use the automatic jsx runtime, so the `React` import is not necessary :) (it's also not wrong)

src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/DocumentTypes.js
```diff
@@ -40,30 +41,37 @@ const getDocumentTypes = async (apiGroupID, catalogueUrl) => {
     throw new Error('Loading available document types failed');
   }
   const documentTypes = response.data.sort((a, b) => a.omschrijving.localeCompare(b.omschrijving));
-  return documentTypes.map(({omschrijving, isPublished}) => ({
+  return documentTypes.map(({omschrijving}) => ({
     value: omschrijving,
-    label: (
+    label: omschrijving,
+  }));
```
sergei-maertens, line 47: isn't `isPublished` missing here?

src/openforms/scss/vendor/_react-select.scss
```diff
@@ -25,4 +25,5 @@
     100%,
     var(--of-admin-select-max-inline-size)
   );
+  max-inline-size: var(--of-admin-select-max-inline-size);
```
sergei-maertens, line 28: I've had to do this in more places in other projects and now I no longer understand how `clamp` works

src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/DocumentTypes.js
robinmolen: Ah okay, check. I'm also not sure why this was added... 😅
But i'll remove it :)

src/openforms/js/components/admin/form_design/registrations/objectsapi/fields/DocumentTypes.js
robinmolen: Oh shit, your right..

src/openforms/scss/vendor/_react-select.scss
robinmolen: Yeah i also dont fully understand why the max-inline-size was needed.. Maybe because a child element was growing, instead of the `.admin-react-select`. So that the max-inline-size prevent all growth?

I think the `clamp` just allows the parent to grow between a min and max value, but it doesn't work fully for flexbox child elements? That they don't respect the set value, and grow beyond? The `.admin-react-select` is used in a `flex-wrap` environment, which might also complicate things..

But I haven't checked/investigated it so deeply..

sergei-maertens (approved):

PR #4814, opened 2024-11-08
Feature/4798 custom confirmation modal

### LINKED ISSUE #4798: Using custom confirmation modals
### Thema / Theme

Frontend

Replacing the `window.confirm(...)` confirmations with a custom confirmation that uses our Modal component

_No response_

_No response_

robinmolen:
Closes #4798

**Changes**

Replacing all `window.confirm(...)` confirmations with a new custom React hook (`useConfirm`). This new hook is using our Modal component to display the confirmation message. These changes happend to keep the application UI uniform.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/zgw/fields/ZGWAPIGroup.js
+++ b/src/openforms/js/components/admin/form_design/registrations/zgw/fields/ZGWAPIGroup.js
@@ -48,9 +48,9 @@ const ZGWAPIGroup = ({apiGroupChoices, onChangeCheck}) => {
           name="zgwApiGroup"
           options={options}
           required
-          onChange={selectedOption => {
-            const okToProceed = onChangeCheck === undefined || onChangeCheck();
-            if (okToProceed) setValue(selectedOption.value);
+          onChange={async selectedOption => {
+            const okToProceed = onChangeCheck === undefined || (await onChangeCheck());
+            if (okToProceed) setValue(selectedOption?.value);
           }}
         />
       </Field>
--- a/src/openforms/js/components/admin/form_design/useConfirm.js
+++ b/src/openforms/js/components/admin/form_design/useConfirm.js
@@ -1,4 +1,5 @@
 import {useState} from 'react';
+import {useIntl} from 'react-intl';
 
 import ActionButton from 'components/admin/forms/ActionButton';
 import {Modal} from 'components/admin/modals';
@@ -25,20 +26,31 @@ const useConfirm = (message, title = '') => {
     handleClose();
   };
 
-  const ConfirmationModal = () => (
-    <Modal
-      title={title}
-      isOpen={promise !== null}
-      contentModifiers={['with-form', 'small']}
-      closeModal={handleCancel}
-    >
-      <div style={{flexGrow: 1}}>{message}</div>
-      <div className="button-group">
-        <ActionButton text="confirm" className="default" onClick={handleConfirm} />
-        <ActionButton text="cancel" onClick={handleCancel} />
-      </div>
-    </Modal>
-  );
+  const ConfirmationModal = () => {
+    const intl = useIntl();
+    const confirmBtnText = intl.formatMessage({
+      description: 'Confirmation modal confirm button',
+      defaultMessage: 'Confirm',
+    });
+    const cancelBtnText = intl.formatMessage({
+      description: 'Confirmation modal cancel button',
+      defaultMessage: 'Cancel',
+    });
+    return (
+      <Modal
+        title={title}
+        isOpen={promise !== null}
+        contentModifiers={['with-form', 'small']}
+        closeModal={handleCancel}
+      >
+        <div style={{flexGrow: 1}}>{message}</div>
+        <div className="button-group">
+          <ActionButton text={confirmBtnText} className="default" onClick={handleConfirm} />
+          <ActionButton text={cancelBtnText} onClick={handleCancel} />
+        </div>
+      </Modal>
+    );
+  };
   return [ConfirmationModal, confirm];
 };
 
--- a/src/openforms/js/components/admin/form_design/useConfirm.js
+++ b/src/openforms/js/components/admin/form_design/useConfirm.js
@@ -1,13 +1,13 @@
 import {useState} from 'react';
 
-import ActionButton from '../forms/ActionButton';
-import {Modal} from '../modals';
+import ActionButton from 'components/admin/forms/ActionButton';
+import {Modal} from 'components/admin/modals';
 
-const useConfirm = (title, message) => {
+const useConfirm = (message, title = '') => {
   const [promise, setPromise] = useState(null);
 
   const confirm = () =>
-    new Promise((resolve, reject) => {
+    new Promise(resolve => {
       setPromise({resolve});
     });
 
@@ -29,7 +29,7 @@ const useConfirm = (title, message) => {
     <Modal
       title={title}
       isOpen={promise !== null}
-      contentModifiers={['with-form']}
+      contentModifiers={['with-form', 'small']}
       closeModal={handleCancel}
     >
       <div style={{flexGrow: 1}}>{message}</div>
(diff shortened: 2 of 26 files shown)


stevenbal:
@robinmolen can you fix the merge conflicts? Also not sure if the failing tests are related to the code changes?

robinmolen:
There is a new window.confirm added in a recent commit, in the `/form_design/form-creation-form.js`. I'll replace that one now with the new `useConfirm`

robinmolen:
I've replaced the `.../form_design/form-creation-form.js` `window.confirm` with the `useConfirm`. With that, all window.confirm's have been replaced :)

src/openforms/js/components/admin/form_design/useConfirm.js
```diff
@@ -0,0 +1,57 @@
...
+      defaultMessage: 'Cancel',
+    });
+    return (
+      <Modal
+        title={title}
+        isOpen={promise !== null}
+        contentModifiers={['with-form', 'small']}
+        closeModal={handleCancel}
+      >
+        <div style={{flexGrow: 1}}>{message}</div>
```
stevenbal, line 46: Maybe a class for this should be added where this styling is set instead?

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ZGWAPIGroup.js
```diff
@@ -44,13 +44,16 @@ const ZGWAPIGroup = ({apiGroupChoices, onChangeCheck}) => {
         }
         noManageChildProps
       >
-        <ReactSelect
+        <SelectWithoutFormik
```
stevenbal, line 47: Are these changes intentional? I'm not sure if we want to remove formik for this field

src/openforms/js/components/admin/form_design/useConfirm.stories.js
```diff
@@ -0,0 +1,37 @@
...
+      ) : null}
+      <ConfirmationModal />
+    </div>
+  );
+};
+
+export default {
+  title: 'Admin / Custom / UseConfirm',
+  render: () => <ButtonWithUseConfirm />,
+  component: useConfirm,
```
stevenbal, line 32: It might be good to add a basic test here to open the modal and check that it has the expected text

src/openforms/js/components/admin/form_design/useConfirm.js
```diff
@@ -0,0 +1,57 @@
...
+        <div className="button-group">
```
stevenbal, line 47: We could maybe reuse `submit-row`, to remove the need for extra styling for button-group

image

stevenbal (changes requested):

src/openforms/js/components/admin/form_design/useConfirm.js
```diff
@@ -0,0 +1,57 @@
...
+    promise?.resolve(true);
+    handleClose();
+  };
+
+  const handleCancel = () => {
+    promise?.resolve(false);
+    handleClose();
+  };
+
+  const ConfirmationModal = () => {
```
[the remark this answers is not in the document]
robinmolen: Yeah this is the normal `small` modal modifier with a height of 25vh, but for the confirmations this is indeed a bit much..
I'll a `dialog` modifier option to the modal component with a bit nicer styling

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ZGWAPIGroup.js
robinmolen: Yes this is intentional. The ReactSelect should set the value after the confirmation is accepted. The `SelectWithFormik` (the component that is exported as the default, aka `ReactSelect`) doesn't allow custom onChange handling, seeing how Formik should control this component.

So another solution for this would be to add a `onChange` handler to the `SelectWithFormik`, but that would (in my opinion) kinda defeat the purpose of the component..

This field still uses Formik, but the value/change handling is now done by the component, instead of the Select

src/openforms/js/components/admin/form_design/useConfirm.stories.js
robinmolen: Ah yeah check, will do!

src/openforms/js/components/admin/form_design/useConfirm.js
robinmolen: Yeah i don't really like the styling of the `submit-row`. The weird extra border and background, extra padding and the uppercase button text.

I get your point, and maybe a modal action bar would be nice (in my head more like a bootstrap modal, where the padding between the `submit-row` and the modal would be removed). But then i would like to change that for all the modals.

For now i think the button group looks quite a bit better

src/openforms/js/components/admin/form_design/useConfirm.js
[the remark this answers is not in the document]
robinmolen: I've made it a bit differently. The modal component now has a new `confirmation` modifier. This applies some styling changes to make the modal "grow" with the content (i.e. no fixed height). This also allows the inline flexGrow styling to be removed

src/openforms/js/components/admin/form_design/useConfirm.js
robinmolen: With some modal styling changes is this inline styling no longer needed 👍 

Extra info: https://github.com/open-formulieren/open-forms/pull/4814#discussion_r1838513591

src/openforms/js/components/admin/form_design/useConfirm.js
robinmolen: I've removed the `button-group` styling and added some modal specific styling. So this still doesn't use the `submit-row` class

src/openforms/js/components/admin/form_design/useConfirm.js
```diff
@@ -0,0 +1,57 @@
...
+      >
+        <p>{message}</p>
+        <div className="react-modal__actions">
+          <ActionButton text={confirmBtnText} className="default" onClick={handleConfirm} />
+          <ActionButton text={cancelBtnText} onClick={handleCancel} />
+        </div>
+      </Modal>
+    );
+  };
+  return [ConfirmationModal, confirm];
```
sergei-maertens, line 54: I remember reading that dynamically defining React components is an anti-pattern, and have tried this in the past in other places too and the conclusion was that it changes all the time/re-renders because there's no stable identity.

Utlimately, the proper solution is to return props to use for the component, and not return the whole component as is.

I'm afraid this approach here suffers from those same issues, and it may have been one of the reasons you were ending up in infinite render loops :thinking:

robinmolen: Ah okay, that sound very interesting... Yeah then this very well might have played a bigger part in the whole infinite render issue..
I'll look into the article, and change the hook

sergei-maertens (commented):
Holding of *a little* longer because of the SDK context/decorator for storybook I mentioned on Slack. I think it would clean up the new/updated interaction tests, and I'm curious if that formik onChange stuff is still necessary, since what it currently does is stop tracking the `isTouched` state and once we refactor to not roll our validation errors/mechanics 100% on our own, that will definitely cause some problems then :)

src/openforms/js/components/admin/form_design/registrations/zgw/fields/ZGWAPIGroup.js
robinmolen: Yeah, its still the case that the `SelectWithFormik` (at the moment) doesn't allow custom onChange handling. This is needed, to trigger the confirmation modal onChange, and to save the data to formik after the confirmation is accepted.

robinmolen: Nevermind... I just noticed (with a comment from Sergei) that you can overwrite the props of the `SelectWithFormik`..
So yeah, this isn't needed and i'll revert this

sergei-maertens (approved):
bellissimo

PR #4819, opened 2024-11-11
Change minimum submission removal limit to 0

### LINKED ISSUE #4815: Entering '0' days for the submission pruning appears to have no effect

2.8.1

### Customer reference

Taiga DH #715

There are global and form-specific configuration options to prune submissions (in various states, like completed/suspended/ongoing/...) after a certain number of days. The amount of days can be configured in the admin interface. A celery task (via beat) picks up the submissions that need to be pruned.

Setting the number of days to '0' seems to cause these submissions not to be pruned, while the expectation is that these would be removed on the first next run of the cleanup job.

(unconfirmed)

1. Create a form, set the cleanup days to '0'
2. Start a submission and complete it
3. Trigger the cleanup task/job
4. Observe that the submission is not deleted

Submission is deleted after next scheduled cleanup job run.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #4815

**Changes**

Allowing submissions to be deleted after 0 days.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene


sergei-maertens (approved):

sergei-maertens (commented):
I would like to see a test that creates a submission on timestamp t0 with a form configured to delete submissions after 0 days, and then the `delete_submissions` is invoked a couple of hours later (e.g. submission create at 18:00 and delete runs at 01:00 AM) to confirm that the submission is then indeed deleted.

I'm seeing some code with the annotations of how old a submission is that I'm not 100% convinced yet this patch fully resolves the problem.

src/openforms/data_removal/tests/test_tasks.py
```diff
@@ -58,6 +58,34 @@ def test_successful_submissions_correctly_deleted(self):
...
+        self,
+    ):
+        form = FormFactory.create(
+            successful_submissions_removal_limit=0,
+            incomplete_submissions_removal_limit=7,
+            errored_submissions_removal_limit=7,
+            all_submissions_removal_limit=7,
+        )
+
+        with freeze_time(lambda: datetime(2024, 11, 12, 18)):
```
sergei-maertens, line 71: what's up with the lambda? Is there a reason you can't do:

```py
with freeze_time("2024-11-12T18:00:00+01:00"):
```

furthermore, *never ever* use naive datetimes in python/django (yours is naive, because no `tzinfo` has been specified). Generally you want to use the django wrappers defined in `django.utils.timezone`.

robinmolen: Ah yeah, that lambda isn't needed 👍 

Ah okay, check. I'll keep it in mind

PR #4837, opened 2024-11-19
Bug/4810 changing textcase on form variable update

### LINKED ISSUE #4810: Case of select boxes values changed when step is never applicable during submission

2..8.1

### Customer reference

DH

When a step is non-applicable from the start of the form and never made applicable, the select box component in this step (mapped with objecttype v2) submits values that are different from what can be seen in the form builder:
image
image

This does not occur when the step is applicable. 
One submission with 2 identical components in different steps can produce this result: 
```json
                "stap-uno": {
                    "ditZijnCheckboxes": {
                        "HOOFDLETTERS": true,
                        "IS_IN_HOOFDLETTERS": true,
                        "JAWEL_SCREAMING_SNAKES": false
                    }
                },
                "stap-dos": {
                    "ditZijnCheckboxes1": {
                        "hoofdletters": false,
                        "is_in_hoofdletters": false,
                        "jawel_screaming_snakes": false
                    }
                }
```

Which causes issues with jsonschema validation.

robinmolen:
Closes #4810

**Changes**

The JSON parser on the form variables endpoint now ignores the `initial_value` field. This ensures that the form variable initial value, and the default value of the component, stay exactly the same.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/forms/tests/variables/test_viewset.py
+++ b/src/openforms/forms/tests/variables/test_viewset.py
@@ -1018,3 +1018,56 @@ class FormVariableViewsetTest(APITestCase):
                     "Prefill plugin, attribute and options can not be specified at the same time."
                 ),
             )
+
+    def test_bulk_create_and_update_with_non_camel_case_initial_values(self):
+        user = StaffUserFactory.create(user_permissions=["change_form"])
+        self.client.force_authenticate(user)
+
+        form = FormFactory.create(generate_minimal_setup=True)
+        form_definition = form.formstep_set.get().form_definition
+        form_path = reverse("api:form-detail", kwargs={"uuid_or_slug": form.uuid})
+        form_url = f"http://testserver.com{form_path}"
+        form_definition_path = reverse(
+            "api:formdefinition-detail", kwargs={"uuid": form_definition.uuid}
+        )
+        form_definition_url = f"http://testserver.com{form_definition_path}"
+
+        data = [
+            {
+                "form": form_url,
+                "form_definition": form_definition_url,
+                "key": form_definition.configuration["components"][0]["key"],
+                "name": "Test",
+                "service_fetch_configuration": None,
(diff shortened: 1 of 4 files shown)


src/openforms/forms/tests/variables/test_viewset.py
```diff
@@ -1018,3 +1018,57 @@ def test_bulk_create_and_update_with_prefill_constraints(self):
                     "Prefill plugin, attribute and options can not be specified at the same time."
                 ),
             )
+
+    def test_bulk_create_and_update_with_non_camel_case_initial_values(self):
+        user = StaffUserFactory.create(user_permissions=["change_form"])
+        self.client.force_authenticate(user)
+
+        form = FormFactory.create()
+        form_step = FormStepFactory.create(form=form)
```
vaszig, line 1027: FYI we have a trait for the minimal setup in the `Form` factory, so you could do something like this `form = FormFactory.create(generate_minimal_setup=True)`, which will create a form step as well . 

Of course no need to change something here, just to be aware of.

vaszig (changes requested):
It looks nice to me, only a couple of remarks (changes are only needed for the type checking if you agree as well)

For future bug fixes, it's nice to have a first commit with the regression test (which should fail) and then a second with the real fix of the bug.

src/openforms/forms/tests/variables/test_viewset.py
robinmolen: Ah okay cool!

To keep things clean, i've changed this to use the `generate_minimal_setup`

vaszig (approved):

vaszig:
Backports:

- stable 2.8.x: ceb3dad33741b38a269faf991009f8ea362bb28d
- stable 2.7.x: be2514ad30f414f24f7e64ba0d5f8070c8af9b97

PR #4842, opened 2024-11-20
Remove password Formio component

sergei-maertens:
Part of #3283

This component never made sense, it was only added reluctantly because the original programme of requirements specified it.

Passwords being stored in plain text are a terrible, terrible idea. Nobody should have every actually used this in a real form.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


vaszig (approved):
What about the existing Password components?Could we have a problem (I cannot think of something but just to make sure)?

sergei-maertens:
> What about the existing Password components?Could we have a problem (I cannot think of something but just to make sure)?

Formio will render "Unknown component type 'password'" or something along those lines. But we've never actually seen anyone use this component (which restores my faith in humanity a little)

PR #4850, opened 2024-11-22
Add test for pre 3.0 addressNL migration check / improve backwards compatibility

sergei-maertens:
Part follow up of #4418

**Changes**

* Added test for registration options without nested options
* Added small compatibility layer to retain the old behaviour for old configuration

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

Issue #4431, opened 2024-06-24
AddressNL - overarching epic


sergei-maertens:
* Add layout option: `singleColumn` / `doubleColumn` (#4423)
* Validate (or show validation errors) only when a field has been touched (#4420)
* Allow specifying validation regular expressions for nested fields - we can leverage the JSON nature for this:
  
  ```json
  {
    "type": "addressNL",
    "key": "address",
    "label": "Your address",
    "components": {
      "postcode": {
        "validate": {"pattern": "1015 [a-zA-Z]{2}"},
        "translatedErrors": {
          "en": {
            "pattern": "Must be at 1015 XX"
          }
        }
      }
    }
  ``` 
  ![image](https://github.com/open-formulieren/open-forms/assets/5518550/4b780257-6399-46bf-9d5b-c25d92844024)

* #4418
* #4430 
*

joeribekker:
@sergei-maertens can you let your mind go over this. I tried to incorporate all features.

The AddressNL-component features these fields:

A. postalcode, housenumber
B. houseletter, housenumber-addition
C. city, streetname

The letters are referenced in the text below.


## Tasks

* [ ] Add dropdown for "mode" in the component editor, allowing you to choose: Regular, Autofill, Prefill (see Mode-behaviour). This replaces the "adres afleiden" checkbox.
* [ ] Autofill can ONLY be selected if the BAG API is configured. Give an error if its not or hide option completely.
* [ ] Add an option to hide C-fields. By default, they are shown.
* [ ] Add a Prefill-tab to the component editor, allowing you to select a plugin, attribute-set (see Prefill attribute-sets) and source

To discuss: Registration

* [ ] Change the Registration-tab in the componenent editor, allowing to select an attribute-set (see Registration attribute-sets)

--OR--

* [ ] Make sure all A, B and C-fields are shown as variables: `<property name>.postalcode, <property name>.housenumber, ... etc.`
* [ ] Remove the registration tab, and connect them INDIVIDUALLY via the variables-tab

--OR --

* [ ] Make sure the component is shown as variable
* [ ] Optionally add all A, B and C-fields are shown as variables: `<property name>.postalcode, <property name>.housenumber, ... etc.`
* [ ] Remove the registration tab, and connect them AS A SET via the variables


### Prefill attribute-sets:

* KvK: Bezoekadres, correspondentie adres
* StUF-BG/HC BRP: Verblijfsadres
* HC HR: Bezoeklocatie, postlocatie
[...]

PR #4851, opened 2024-11-22
Refactored the record-data building + post-processing in Objects API registration v2

sergei-maertens:
Part of #4431 - builds on top of #4850

**Changes**

Refactored the Objects API v2 registration data building/transformations after the mappings were implemented by @vaszig. I was also able to add more files to the type checker and finally some type definitions / hints could also be cleaned up, hurray!

This had been on my mind for a while when we were fixing the bugs specifically to the file upload types w/r to single/multiple uploads, and now the additional options lead to more :spaghetti: code in this registration handler, so I figured it was time to put my thoughts into code.

### DIFF
--- a/src/openforms/contrib/objects_api/clients/objects.py
+++ b/src/openforms/contrib/objects_api/clients/objects.py
@@ -1,12 +1,14 @@
 from zgw_consumers.nlx import NLXClient
 
+from ..typing import Object, Record
+
 CRS_HEADERS = {"Content-Crs": "EPSG:4326"}
 
 
 class ObjectsClient(NLXClient):
-    def create_object(self, objecttype_url: str, record_data: dict) -> dict:
+    def create_object(self, objecttype_url: str, record_data: Record) -> dict:
 
-        json = {
+        json: Object = {
             "type": objecttype_url,
             "record": record_data,
         }
@@ -24,9 +26,9 @@ class ObjectsClient(NLXClient):
 
         return response.json()
 
-    def update_object(self, record_data: dict, object_uuid: str) -> dict:
+    def update_object(self, record_data: Record, object_uuid: str) -> dict:
         endpoint = f"objects/{object_uuid}"
-        json = {
+        json: Object = {
             "record": record_data,
         }
 
--- a/src/openforms/registrations/contrib/objects_api/handlers/v2.py
+++ b/src/openforms/registrations/contrib/objects_api/handlers/v2.py
@@ -1,0 +1,120 @@
+"""
+Implementation details for the v2 registration handler.
+"""
+
+from collections.abc import Sequence
+from dataclasses import dataclass
+from datetime import date, datetime
+
+from glom import Assign, Path, glom
+
+from openforms.api.utils import underscore_to_camel
+from openforms.formio.typing import Component
+from openforms.typing import JSONObject, JSONValue
+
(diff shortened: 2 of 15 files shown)


src/openforms/registrations/contrib/objects_api/handlers/v2.py
```diff
@@ -0,0 +1,120 @@
...
+    if isinstance(value, (date, datetime)):
+        value = value.isoformat()
+
+    # transform the value within the context of the component
+    # TODO: convert this in a proper registry in due time so we can use better type
+    # annotations
+    match component:
+        case {"type": "addressNL"}:
+            assert isinstance(value, dict)
+            value = value.copy()
```
vaszig, line 73: Why don't we use a new variable here?I thought this is the way when we copy one.

src/openforms/typing.py
```diff
@@ -14,19 +16,19 @@
 else:
     _StrOrPromise = str
 
-JSONPrimitive: TypeAlias = str | int | float | bool | None
+type JSONPrimitive = str | int | float | bool | None
```
vaszig, line 19: Is it the new way of declaring a type alias or are we doing this to avoid the using the library's class?

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -157,14 +157,27 @@ def remove_variables(self, keys: list) -> None:
...
+            for variable in get_static_variables(
+                submission=self.submission,
+                variables_registry=other_registry,
+            )
+        }
+
+    def get_static_data(
+        self, other_registry: VariablesRegistry | None = None
+    ) -> JSONObject:
+        # Enseure we do not accidentally cache the non-default registry
```
vaszig, line 172: If anything else needs update in the code let's fix the typo as well

vaszig (approved):

src/openforms/registrations/contrib/objects_api/handlers/v2.py
sergei-maertens: I'm only doing it here to avoid mutating the original object, and the type stays the same. I think either way is fine?

src/openforms/typing.py
sergei-maertens: this is indeed a Python 3.12 feature! the `type` keyword can now be used to declare type aliases, and in particular it makes it easier to reference symbols only defined later so you don't need to wrap them in a string! The added benefits are of course that the syntax is less verbose and less imports indeed.

src/openforms/submissions/models/submission_value_variable.py
sergei-maertens: I'll aply the suggestions!

src/openforms/contrib/objects_api/clients/objects.py
```diff
@@ -24,7 +26,7 @@ def get_object(self, object_uuid: str) -> dict:
 
         return response.json()
 
-    def update_object(self, record_data: dict, object_uuid: str) -> dict:
+    def update_object(self, record_data: Record, object_uuid: str) -> dict:
         endpoint = f"objects/{object_uuid}"
         json = {
```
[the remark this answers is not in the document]
sergei-maertens: Added it, and made the `type` key `NotRequired` with a comment providing more context.

Issue #4320, opened 2024-05-22
Various co-sign improvements


joeribekker:
### Thema / Theme

Frontend

#### To do now

For actual text changes, see https://github.com/open-formulieren/open-forms/issues/4320#issuecomment-2359019567

**Globaal**

* [x] Het woord "mede-ondertekener" moet worden "mede-ondertekenaar"

* Bij het gebruik van een link in de mede-onderteken e-mail:
    * [x] Verberg mede-onderteken blok op het formulier
    * [x] Stuur een link in de mede-onderteken e-mail, die een code in de URL bevat dat het om mede-ondertekenen gaat.
    * [x] Toon enkel het mede-onderteken blok als de URL de code van hierboven bevat (verberg dus de mogelijkheid om gewoon in te loggen)
    * [x] Geen tussenstap om code in te voeren

**Startpagina**

* [x] Tekstwijziging bij mede-onderteken blok: Heeft u een e-mail ontvangen om mede te ondertekenen? Start het mede-ondertekenen door in te loggen.

**Bevestigingspagina**

* [x] Tekstwijziging titel
* [x] Tekstwijziging body
* [x] Toon NIET het nummer indien links worden gebruikt
* [x] Toon direct een knop "Nu mede-onderteken" (die de link gebruikt om mede te ondertekenen)

**Bevestigings e-mail naar de indiener**

* [x] Tekstwijziging: Onderwerp moet niet "bevestiging" bevatten en ook geen referentienummer. Dit geeft de indruk dat het klaar is.
* [x] Verwijder kopje "samenvatting" als er geen samenvatting is
* [x] Tekstwijziging: Body

**Mede-onderteken e-mail**

* [x] Tekstwijziging: Body (meer B1)
* [x] Betere layout (kopje)

**Bevestiging mede-ondertekening**

* [x] Tekstwijziging. Wellicht meer zoals de standaard bevestiging

* [x] Bij co-signing wordt de 'ingezonden op...' datum al gegenereerd voordat het formulier is ingezonden. De behandelaar heeft de inzending nog niet ontvangen op dat moment. Hier kunnen juridische issues door ontstaan. -> gaat om PDF, input DH

#### Later

**Globaal**

* [ ] Meerdere PDFs met onderscheid in wat relevant is (tevens wel/niet betaald, wel/niet mede-ondertekend). Deze als versies opslaan.

**Mede-onderteken component**

* [ ] Vraag om BSN (of selecteer partner in latere fase)

[...]

PR #4852, opened 2024-11-25
:children_crossing: [#4320] Use more accessible language

sergei-maertens:
Simplified the language down to B1 level.

Needs to be backported to 2.7.x too!


vaszig (approved):
I will leave the CI to finish and I will do the backport as well

PR #4867, opened 2024-11-28
Removing single backend registration config from form

robinmolen:
Part of #3283

**Changes**

The deprecated fields `registration_backend` and `registration_backend_options` have been removed. The field `registration_backends` should now be used in all situations.

These fields have also been removed from the import and export. This means, when you import a form using the `registration_backend` config, the registration will no longer be converted to `registration_backends` and the form will be imported without registration configurations.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene


docs/installation/upgrade-300.rst
```diff
@@ -67,3 +67,21 @@ Removal of legacy translations conversion
...
+The legacy format (from before Open Forms 2.3) for registration backend will no longer be
+converted to the current standard. When importing a form with this configuration,
+the form will be created without registration backends.
+
+Removal of single registration on forms
+=======================================
+
+The old registration backend configuration (from before Open Forms 2.3) is now completely
+removed. This removes `registrationBackend` and `registrationBackendOptions` from all api
+endpoints. You should be using the `registrationBackends` property.
```
sergei-maertens, line 83: No need to document this - the API endpoints were labeled private API a while ago. The external world shouldn't have been using this in the first place :)

robinmolen: I've removed this part of upgrade note :)

src/openforms/forms/tests/test_copy.py
```diff
@@ -33,7 +33,10 @@ def test_form_copy_with_reusable_definition(self):
         self.assertNotEqual(copied_form.pk, form.pk)
         self.assertNotEqual(copied_form.uuid, str(form.uuid))
         self.assertEqual(copied_form.active, form.active)
-        self.assertEqual(copied_form.registration_backend, form.registration_backend)
+        self.assertEqual(
```
vaszig, line 36: Maybe we should test the exact backends  and not the number of them(since we are testing the copy functionality)?Like we do with all the assertions above.

src/openforms/forms/tests/test_import_export.py
```diff
@@ -281,10 +281,9 @@ def test_import(self):
         self.assertNotEqual(imported_form.pk, form_pk)
         self.assertNotEqual(imported_form.uuid, str(form.uuid))
         self.assertEqual(imported_form.active, False)
-        self.assertEqual(imported_form.registration_backend, form.registration_backend)
         self.assertEqual(
-            imported_form.registration_backend_options,
-            form.registration_backend_options,
+            imported_form.registration_backends.count(),
```
vaszig, line 285: I would say the same like the comment for the copy.

vaszig (dismissed):
I am a bit confused here and maybe I am missing something..

I see `registration_backend` and `registration_backend_options` are still used in a lot of places, shouldn't we remove them from everywhere now that we deprecated these two? There are factories as well that make use of them so they need updating and then the tests themselves.

Like here for instance:
https://github.com/open-formulieren/open-forms/blob/master/src/openforms/forms/tests/factories.py#L50

https://github.com/open-formulieren/open-forms/blob/master/src/openforms/registrations/contrib/email/tests/test_email_rendering.py#L14

**And a general comment..It helps a lot if you add next to the subtask (in cases like meta ticket 3283) the link for the pull request

robinmolen:
> I see `registration_backend` and `registration_backend_options` are still used in a lot of places, shouldn't we remove them from everywhere now that we deprecated these two? There are factories as well that make use of them so they need updating and then the tests themselves.

Yeah this is something i discussed with Sergei, keeping the `registration_backend` and `registration_backend_options` in the `FormFactory`, to keep the registration simple for the test. But yeah, maybe best to completely remove this way of working.. I'll take a crack at it :)

> **And a general comment..It helps a lot if you add next to the subtask (in cases like meta ticket 3283) the link for the pull request

Ah check, yeah will do 👍

src/openforms/forms/tests/test_copy.py
robinmolen: I've done a number comparison because both have 0 backends :)
But yeah, its a bit ambiguous. I'll make it more explicit 👍

src/openforms/submissions/admin.py
```diff
@@ -361,7 +361,7 @@ def successfully_processed(self, obj) -> bool | None:
 
     @admin.display(description=_("Registration backend"))
     def get_registration_backend(self, obj):
-        return obj.form.registration_backend or "-"
+        return backend if (backend := obj.form.registration_backends.first()) else "-"
```
sergei-maertens, line 364: looks like this was bugged in the old code, it should really use `submission.registration_backend` so it takes into account the dynamic backend selection from logic rules (!)

src/openforms/forms/tests/test_copy.py
sergei-maertens: let's make it very clear what the expected result is, both the old and new assertion confuse me in what the actual initial and final state are

sergei-maertens:
> > I see `registration_backend` and `registration_backend_options` are still used in a lot of places, shouldn't we remove them from everywhere now that we deprecated these two? There are factories as well that make use of them so they need updating and then the tests themselves.
> 
> Yeah this is something i discussed with Sergei, keeping the `registration_backend` and `registration_backend_options` in the `FormFactory`, to keep the registration simple for the test. But yeah, maybe best to completely remove this way of working.. I'll take a crack at it :)
> 
> > **And a general comment..It helps a lot if you add next to the subtask (in cases like meta ticket 3283) the link for the pull request
> 
> Ah check, yeah will do 👍

quick search on master branch for `registration_backend=` usage in `test_*.py` files:

> 90 matches across 17 files

which will turn the currently one liner into more lines because a separate factory needs to be called. On a PR with 11 changed files, I expect this to make the PR twice as large. IMO the convenience in the factory (which is precisely the *reason* why we use factories) is not a problem, it's just a shortcut, like how `SubmissionFactory(auth_info__identifier="123")` is a shortcut that makes the tests a lot more readable and maintainable.

So, I suggest *not* to rip this out of the factories.

src/openforms/submissions/admin.py
robinmolen: Check, the `obj.registration_backend` is now used for this admin display. I've also added a test that makes sure that the admin correctly shows the registration_backend of the submission

src/openforms/submissions/tests/form_logic/test_submission_logic.py
```diff
@@ -369,6 +372,114 @@ def test_check_logic_hide_with_default_value(self):
             data["step"]["formStep"]["configuration"]["components"][0]["hidden"],
         )
 
+    def test_check_logic_with_registration_backend_correctly_displayed_in_admin(self):
```
sergei-maertens, line 375: can you please put the tests in `src/openforms/submissions/tests/test_admin.py` with the rest of the admin tests, and use webtest?

These tests here are tests for API endpoints

sergei-maertens (changes requested):

src/openforms/submissions/tests/form_logic/test_submission_logic.py
robinmolen: Hmm yeah i wasn't sure about the placement of this test.

On the one hand, it is about the admin, so should be placed with the admin tests. On the other hand, this test heavily focuses on the form logic, and the effects it has..

robinmolen: But I'll move it to the `test_admin` 👍

sergei-maertens (approved):

src/openforms/submissions/tests/form_logic/test_submission_logic.py
sergei-maertens: > On the one hand, it is about the admin, so should be placed with the admin tests. On the other hand, this test heavily focuses on the form logic, and the effects it has..

It's testing what is displayed on the admin page and that's what the initial comment was about :-) that the logic engine needs to be used is an implementation detail, but the original test case was `APITestCase` and that's _definitely_ the wrong location for admin tests :stuck_out_tongue:

PR #4869, opened 2024-11-28
Refactor ZGW options components structure

sergei-maertens:
Required for #4606

Refactored the components and hooks so we can lift up the retrieval of available catalogues in a ZGW API group.

The parent component is now responsible for fetching the available catalogues and relays the loading/error state from useAsync. It also encapsulated the derived catalogueUrl when a valid value is available and selected, which other components need to look up related objects (like case types, document types, products and the future role types).

I've opted to pass down the error into a component that just throws for the existing error boundary behaviour and location, as it's important that you can still change the API group to trigger a new attempt and we want to keep the code nicely organised.

Finally, some more fieldsets were abstracted into their own components for readability reasons.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/js/compiled-lang/en.json
+++ b/src/openforms/js/compiled-lang/en.json
@@ -2517,6 +2517,12 @@
       "value": "Add variable"
     }
   ],
+  "LbiXCA": [
+    {
+      "type": 0,
+      "value": "Something went wrong while retrieving the available products defined in the selected case. Please check that the services in the selected API group are configured correctly."
+    }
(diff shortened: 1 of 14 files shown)


sergei-maertens:
This was briefly discussed on Slack with @robinmolen and I *think* the conclusion was that this PR can be merged. If there's still feedback after all, I'll handle it separately in another PR so that I can wrap this up in our off-week.

PR #4870, opened 2024-11-28
Display available role types in dropdown for ZGW registration plugin

### LINKED ISSUE #4606: Make ZGW registration options consistent with Objects API

sergei-maertens:
Closes #4606

**Changes**

These are the final UX improvements for this registration backend.

Once a case type has been selected in the dropdown, the available role type options are retrieved and displayed in a
dropdown themselves. The runtime registration behaviour was already in place, i.e. it already works with an indirect reference rather than a fixed URL.

The legacy configuration is also still supported - if a zaaktype URL is specified and no catalogue + case type identification reference provided, the form field degrades back into the text field that is was before this PR. The same semantic data is stored in both cases.

I've also added more fixture data to Open Zaak to account for the possible edge cases that the client/API endpoint need to handle.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -325,6 +357,46 @@ def list_statustypen(self, zaaktype: str) -> list[dict]:
...
+            or []
+        )
+        all_valid_roltype_urls: list[str] = sum(
+            (
+                case_type_version.get("roltypen", [])
+                for case_type_version in case_type_versions
+            ),
+            [],
+        )
+        if not all_valid_roltype_urls:
```
robinmolen, line 382: Is there a reason to use `if not all_valid_roltype_urls` instead of `if len(all_valid_roltype_urls) == 0`?

Just for my own understanding :) (I'm reading some mixed preferences/reasons online. So i'm wondering what your reasoning is)

src/openforms/js/components/admin/form_design/registrations/zgw/fields/MedewerkerRoltype.js
```diff
@@ -1,12 +1,87 @@
...
+    case_type_identification: caseTypeIdentification,
+  });
+  if (!response.ok) {
+    throw new Error('Loading available object types failed');
+  }
+  const roleTypes = response.data.sort((a, b) => a.description.localeCompare(b.description));
+  return roleTypes.map(({description, descriptionGeneric}) => ({
+    value: description,
+    label: description,
+    descriptionGeneric: descriptionGeneric, // omschrijvingGeneriek, which is an enum
```
robinmolen, line 29: Is `descriptionGeneric` this needed?

robinmolen:
I don't think we have a test for creating a case with `roltypen`. Maybe nice to add one :)

src/openforms/contrib/zgw/clients/catalogi.py
sergei-maertens: it's considered the "Pythonic" way and relies on a value being falsy. In general, it's considered more readable, but I've seen some recent arguments/talks that this is actually going against the "explicit is better than implicit" guideline.

So in short - "this is how it's usually done in Python".

src/openforms/js/components/admin/form_design/registrations/zgw/fields/MedewerkerRoltype.js
sergei-maertens: not used at the moment, though I do anticipate that it'll possibly be displayed similarly to a variable key. I've opted to include it here so that it's obvious that we *do have* that information.

sergei-maertens:
> I don't think we have a test for creating a case with `roltypen`. Maybe nice to add one :)

We have existing tests:

* `test_submission_with_registrator`
* `test_submission_roltype_initiator_not_found`

if we didn't have a test, the [coverage](https://app.codecov.io/github/open-formulieren/open-forms/blob/master/src%2Fopenforms%2Fregistrations%2Fcontrib%2Fzgw_apis%2Fplugin.py#L402) check would complain too :wink:

src/openforms/contrib/zgw/clients/catalogi.py
robinmolen: Yeah exactly, i've read some similar messages about the "explicit is better than implicit", so i was wondering what is "best" for Open Forms.

Okay, check 👍

robinmolen (approved):
Looks good to me

PR #4872, opened 2024-11-29
[#3283] Remove oldstyle appointments flow

### LINKED ISSUE #3283: Deprecations to remove in 3.0

vaszig:
Closes #3283 partly

**Changes**

- Updated code to adhere to the new appointments flow
- Now we do not extract the appointment data from a submission
- Removed frontend/admin code to manage appointment (tab)
- Removed support for fixed customer fields
- Removed "invalid input serializer -> return empty result list" in favour of raising a validation error exception

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/js/components/admin/form_design/Appointments.js
robinmolen, line 1: I this file can be removed entirely :)

src/openforms/js/components/admin/form_design/form-creation-form.js
robinmolen, line 1: Lets just remove these items :)

src/openforms/emails/templates/emails/templatetags/appointment_information.html
```diff
@@ -53,17 +53,12 @@ <h2 style="margin: 1em 0; font-size: 1.3em;">{% trans "Your contact details" %}<
     </table>
 
     <p>
-        {% trans "If you want to cancel or change your appointment, you can do so below." %}
+        {% trans "If you want to cancel your appointment, you can do so below." %}
     </p>
     <p>
-        <a href="{{ appointment_cancel_link }}" rel="nofollow">{% trans "Cancel appointment" %}</a>
+        {% trans "In case you want to modify it you have to cancel it and then create a new one." %}
```
robinmolen, line 59: I think the previous message was a bit nicer/more descriptive.

```suggestion
        {% trans "If you wish to change your appointment, please cancel it using the link below and create a new one." %}
```

src/openforms/appointments/tasks.py
```diff
@@ -21,7 +18,7 @@
     ignore_result=False,
     once={"graceful": True},  # do not spam error monitoring
 )
-def maybe_register_appointment(submission_id: int) -> None | str:
+def register_appointment(submission_id: int) -> None | str:
```
robinmolen, line 21: Shouldn't this task be removed?

vaszig: No, this was discussed with Sergei and we decided that this is the correct way to do it (as a celery task). We just remove the old-flow parts from the task itself.

robinmolen: Ahhh okay, check 👍

src/openforms/appointments/management/commands/appointment.py
```diff
@@ -5,6 +5,7 @@
 from django.core.management.base import CommandError
 from django.db.models import Q
 
+from openforms.appointments.base import CustomerDetails
```
robinmolen, line 8: Import for CustomerDetails already exists

docs/manual/forms/examples/appointment_creation.rst
```diff
@@ -1,83 +0,0 @@
...
-
-12. (Optioneel) Sleep een **Telefoonnummer** component op het witte vlak, vul de volgende
-    gegevens in en druk daarna op **Opslaan**:
-
-   * **Label**: Telefoonnummer
-
-13. Klik op het tabblad **Afspraken**.
-14. Kies de juiste componenten voor elke veld in dit tabblad.
-
-.. image:: _assets/appointment_creation_2_2.png
```
sergei-maertens, line 73: this image can be deleted too

src/openforms/appointments/tasks.py
```diff
@@ -21,37 +18,20 @@
...
```
sergei-maertens, line 21: I'd stick with the original task name though, the `maybe_` prefix indicates that the task itself may not do anything at all. This is done so that the logic to act or not act is contained in the task itself (in this task `NoAppointmentForm` can be raised and then nothing is done).

The location where the task is scheduled then reads more natural:

* maybe we need to register an appointment, then
* maybe run pre-registration, then
* (definitely) generate the summary PDF, then
* maybe run registration, then
* maybe send a confirmation email

etc.

From the scheduling point of view, you don't know what will run or not, and while you can check it in there, that greatly breaks the loose coupling because you end up with a "god function" that understands the whole system. We don't do that, instead we set up the chain of tasks to execute and use naming conventions to indicate what will or might happen.

src/openforms/appointments/core.py
```diff
@@ -116,5 +115,4 @@ def book_for_submission(submission: Submission) -> str:
         logevent.appointment_register_failure(appointment_info, plugin, e)
         raise AppointmentRegistrationFailed("Unable to create appointment") from e
 
-    cancel_previous_submission_appointment(submission)
```
sergei-maertens, line 119: considering this whole function is new-style appointments only, I'm not so sure about this line being removed :thinking:

src/openforms/submissions/query.py
sergei-maertens, line 1: if we remove previous_submission, do we need the copy method at all? Seems like it's only used in the appointments app. That would be a really nice cleanup if we can get rid of it!

src/openforms/emails/templatetags/appointments.py
```diff
@@ -22,7 +22,7 @@ def appointment_information(context):
     else:
         template_name = "emails/templatetags/appointment_information.html"
 
-    # check for new style appointments
+    # check for appointments
     submission = context["_submission"]
     appointment = get_appointment(submission)
     plugin_id = appointment.plugin if appointment else ""
```
sergei-maertens, line 28: Shouldn't this now basically be an `assert appointment` because it doesn't make sense to use this tag if there's no appointment?

sergei-maertens (commented):
OMG so much code being removed it's like it's almost Christmas!

src/openforms/appointments/core.py
vaszig: I am not sure that I understand your point. I removed it because we do not have to deal with an extra function in order to cancel an appointment (we have `CancelAppointmentView`) right?And basically I have removed the whole function from `appointments/utils.py`.

sergei-maertens: I was wrong - this was incorrectly added when the new-style appointments was created and the removal is correct

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1,6 +1,6 @@
 import {produce} from 'immer';
 import cloneDeep from 'lodash/cloneDeep';
-import getObjectValue from 'lodash/get';
+// import getObjectValue from 'lodash/get';
```
sergei-maertens, line 3:
```suggestion
```

sergei-maertens (changes requested):
and obviously the CI failures need to be looked into!

sergei-maertens (approved):

sergei-maertens:
YEET

image

PR #4873, opened 2024-11-29
Reinstate some default styling (tooltip, fieldsets)

sergei-maertens:
* Removed the custom tooltip styling to be aligned with default-project again
* Fixed the checkbox markup in our react components to be consistent with rest of django admin
* Removed CSS overrides that have been obsoleted now that RJSF has been dropped.


robinmolen:
I really like the changes (the big red minus and the accessibility improvements 👌)

There are some weird spacing styling issues with the help texts that, in my opinion, should be 'fixed'. Things like:
- alignment with inputs
- spacing between fields
- weird extra space between input and help text

I've added some comments in the chromatic to highlight these items

src/openforms/js/components/admin/forms/Field.js
sergei-maertens, line 1: This was... very broken since a long time :) I can't even remember if we had storybook back when we were on Django 3.2

sergei-maertens:
@robinmolen this should now be ready for re-review. Note that I dismissed some valid remarks due to it being the way it is in Django markup/CSS and I'd rather not introduce custom CSS again with all the pain it brings when upgrading Django :/

robinmolen (approved):
Yeah i get your disdain to replace custom styling with other custom styling...

The changes look good to me!

PR #4881, opened 2024-12-03
Add new leaflet map background configuration

robinmolen:
Part of #2173

**Changes**

Adds a new admin config page/model for Leaflet map backgrounds. The map components can use these new background definitions to display custom backgrounds on the tile layer (for the formio-builder and sdk).

When a map component is fetched (by the formio-builder and/or sdk), the url for the background is fetched and passed along. This allows admins to update the Leaflet map background urls and makes sure that all map components use the new configuration.

This PR also adds some tests for the existing `rewrite_for_request` map component logic (updating the component, if global config should be used).

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Commit hygiene


src/openforms/formio/components/custom.py
```diff
@@ -198,6 +198,16 @@ def rewrite_for_request(component, request: Request):
             component["initialCenter"]["lat"] = config.form_map_default_latitude
             component["initialCenter"]["lng"] = config.form_map_default_longitude
 
+        if component.get("backgroundIdentifier", False):
+            try:
+                leaflet_map_background = LeafletMapBackground.objects.get(
+                    identifier=component["backgroundIdentifier"]
+                )
+                # Write the background url information
+                component["url"] = leaflet_map_background.url
```
vaszig, line 207: Let's move this out from the try block and emphasize only on the line which will cause the exception.

vaszig (changes requested):

src/openforms/config/tests/test_admin.py
```diff
@@ -27,3 +27,24 @@ def test_color_detail(self):
         response = self.app.get(url, user=user)
 
         self.assertEqual(response.status_code, 200)
+
+
+@disable_admin_mfa()
+class LeafletMapBackgroundTests(WebTest):
```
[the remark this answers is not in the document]
robinmolen: Check, yeah this just tests the creation and pages for the new map object.

The tests aren't particular exciting, but maybe nice as "sanity check" tests :)

src/openforms/formio/components/custom.py
robinmolen: Yeah sounds good. Will do 👍

robinmolen:
This PR is depended on https://github.com/open-formulieren/formio-builder/pull/197. When the formio-builder PR is merged, the `@open-formulieren/formio-builder` version in the PR should be updated.

Once that has been done, this PR can be merged

robinmolen:
The formio-builder has been updated. This PR can now be fully reviewed

src/openforms/js/components/form/map.js
```diff
@@ -70,8 +72,26 @@ export default class Map extends TextFieldComponent {
...
+    );
     map.addLayer(tiles);
   }
+
+  // Try to get the tile layer url for the component.
+  // If it cannot be found, return the default url.
+  getTileLayerUrl(tileLayerIdentifier) {
+    const MAP_TILE_LAYERS = jsonScriptToVar('config-MAP_TILE_LAYERS', {
+      default: [],
+    });
```
sergei-maertens, line 87: I'd move this outside of the method onto the module level - since the data is supposed to already be in the DOM before the JS starts to run, we can evaluate this at load-time rather than doing the same evaluation every time a map gets rendered.

src/openforms/formio/tests/test_dynamic_config.py
```diff
@@ -0,0 +1,292 @@
...
+
+        configuration = {
+            "components": [
+                {
+                    "type": "map",
+                    "key": "map",
+                    "defaultZoom": "3",
+                    "initialCenter": {
+                        "lat": "43.23",
+                        "lng": "41.23",
```
sergei-maertens, line 36: are these stored as string or floats with the actual formio-builder? The types say `number`: https://github.com/open-formulieren/types/blob/5e2e2a219b4280c162c6eeffa6a51e333706bf49/src/formio/components/map.ts#L40

src/openforms/formio/components/custom.py
```diff
@@ -189,6 +189,20 @@ def build_serializer_field(
...
+        if component.get("tileLayerIdentifier", False):
+            try:
+                tile_layer = MapTileLayer.objects.get(
+                    identifier=component["tileLayerIdentifier"]
+                )
+            except MapTileLayer.DoesNotExist:
+                return None
+
+            # Add the tile layer url information
+            component["tileLayerUrl"] = tile_layer.url
```
sergei-maertens, line 204: With some shortcuts and our beloved :walrus: operator

```suggestion
        if (identifier := component.get("tileLayerIdentifier")) is not None:
            tile_layer = MapTileLayer.objects.filter(identifier=identifier).first()
            if tile_layer is not None:
                # Add the tile layer url information
                component["tileLayerUrl"] = tile_layer.url
```

the fallback to `False` introduces yet another type to handle (now we have `str | None` rather than `str | None | bool`)

src/openforms/formio/components/custom.py
```diff
@@ -189,6 +189,20 @@ def build_serializer_field(
 class Map(BasePlugin[Component]):
```
sergei-maertens, line 189: it may be a good idea to define a `MapComponent` type that's more specific and includes the extra properties that are component specific.

src/openforms/fixtures/admin_index_unlisted.json
```diff
@@ -1,5 +1,6 @@
 [
     "config.RichTextColor",
+    "config.MapTileLayer",
```
sergei-maertens, line 3: These should go in the configuration group, since the unlisted stuff doesn't get displayed to non-superusers, I think

(and that's probably a problem for the rich text colors)

src/openforms/config/models/map.py
```diff
@@ -0,0 +1,35 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class MapTileLayer(models.Model):
+    identifier = models.SlugField(
+        _("identifier"),
+        unique=True,
+        max_length=50,
+        help_text=_("A unique identifier for the tile layer"),
```
sergei-maertens, line 10:
```suggestion
        help_text=_("A unique identifier for the tile layer."),
```

src/openforms/config/models/map.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    url = models.URLField(
+        _("tile layer url"),
+        max_length=255,
+        help_text=_(
```
sergei-maertens, line 15: I think you need a disclaimer that the tiler layers must use EPSG:28992 (Rijksdriehoek coordinate system), using other systems will probably lead to weird results.

src/openforms/config/admin.py
```diff
@@ -221,6 +221,20 @@ class RichTextColorAdmin(admin.ModelAdmin):
...
+    fields = [
+        "label",
+        "identifier",
+        "url",
+    ]
+    list_display = [
+        "label",
+        "identifier",
+        "url",
+    ]
```
sergei-maertens, line 235: I tend to favour tuples rather than lists for these properties, because:

* tuples are immutable
* lists defined at the module level are mutable and people tend to try nasty things where they dynamically patch the admin class and add stuff (in tests) and then they get surprises 3 months later

In general I avoid using mutable datastructures at the module-level

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
sergei-maertens, line 1: Can you revert this? We'll handle translations close to release.

src/openforms/fixtures/admin_index_unlisted.json
robinmolen: Ahhh okay, yeah that would be a big problem..

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
robinmolen: Yes, will do 👍

src/openforms/formio/tests/test_dynamic_config.py
robinmolen: Ah check, yeah these should be floats. Changing it 👍

src/openforms/formio/components/custom.py
robinmolen: Ah okay cool. Yeah that `objects.filter` is also pretty nice

src/openforms/fixtures/admin_index_unlisted.json
robinmolen: For now i've left RichTextColors in the unlisted file. I'll add another PR to change that

src/openforms/formio/components/custom.py
robinmolen: yeah definitely, seeing as this component will get a bit bigger with the other changes..
I've added a `MapComponent` type and replaced some instances of `Component` with `MapComponent`

sergei-maertens (approved):

PR #4885, opened 2024-12-03
[#4321] Allow max number of submissions per form

### LINKED ISSUE #4321: Als formulier editor wil ik een maximum aantal inzendingen toestaan.
### Thema / Theme

Form designer

* [x] Set a maximum number of submissions to be allowed.
* [x] Keep track of the number of submissions (separate from the statistics)
* [x] Allow to reset the count to 0 again, without resetting the statistics.
* [x] If the maximum is reached, the form should not start (ideally it would also show directly on the start page).

_No response_

_No response_

vaszig:
Closes #4321 partly

**Changes**

- Added fields to Form model in order to track max submissions and the counter
- New tab on the admin page of the form for setting max allowed submissions

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


src/openforms/forms/admin/form.py
```diff
@@ -48,6 +48,21 @@ class FormStepInline(OrderedTabularInline):
...
+
+    def lookups(self, request, model_admin):
+        return [("unavailable", "Unavailable for submission")]
+
+    def queryset(self, request, queryset):
+        if self.value() == "unavailable":
+            return queryset.filter(
+                ~Q(submission_maximum_allowed=None)
+                & Q(submission_maximum_allowed=F("submission_counter"))
+            )
```
sergei-maertens, line 63:
```suggestion
            return queryset.filter(
                submission_maximum_allowed__isnull=False,
                submission_counter__gte=F("submission_maximum_allowed"),
            )
```

* `isnull` is a bit more explicit *and* allows you to drop the `Q` wrapper
* race conditions happen and unless you set up a database constraint, at some point there will be situations where the submission counter has  value larger than `submission_maximum_allowed`

An example of the second case happening in an innocent way:

1. Admin user sets submission limit of 100
2. Form gets 75 submissions
3. Admin user changes submission limit to 50

Now you have 75 submissions which is more than 50 - the form is not available.

src/openforms/forms/admin/form.py
```diff
@@ -48,6 +48,21 @@ class FormStepInline(OrderedTabularInline):
     extra = 1
 
 
+class FormReachedSubmissionLimitListFilter(admin.SimpleListFilter):
+    title = _("has reached submission limit")
+    parameter_name = "submission_limit"
+
+    def lookups(self, request, model_admin):
+        return [("unavailable", "Unavailable for submission")]
```
sergei-maertens, line 56: label must be marked translatable

src/openforms/forms/admin/form.py
```diff
@@ -48,6 +48,21 @@ class FormStepInline(OrderedTabularInline):
     extra = 1
 
 
+class FormReachedSubmissionLimitListFilter(admin.SimpleListFilter):
+    title = _("has reached submission limit")
+    parameter_name = "submission_limit"
```
sergei-maertens, line 53: Can we not use a filter that's simpler like the filters for a boolean field? it's either Submission limit reached or not.

src/openforms/forms/models/form.py
```diff
@@ -380,12 +397,27 @@ def __str__(self):
...
+                self._is_deleted,
+                not self.active,
+                self.maintenance_mode,
+                self.has_reached_submissions_limit(),
+            )
+        ):
             return False
         return True
 
+    def has_reached_submissions_limit(self) -> bool:
```
sergei-maertens, line 414: should be a `@property` instead of a method, it's a very simple calculated property.

src/openforms/forms/models/form.py
```diff
@@ -380,12 +397,27 @@ def __str__(self):
...
+        if (
+            limit := self.submission_maximum_allowed
+        ) and limit == self.submission_counter:
```
sergei-maertens, line 417: This check is duplicated in many places - you should use this property instead of repeating the equality check, and as mentioned before, check for greater than or equal rather than just equality.

src/openforms/js/components/admin/form_design/SubmissionFields.js
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  const resetCounter = async () => {
+    try {
+      const resetResult = await patch(
+        `/api/v2/forms/${formUuid}/reset_submission_counter`,
+        csrftoken,
+        {},
+        true
+      );
```
sergei-maertens, line 37: I'd expect a patch call to the detail endpoint here rather than a separate endpoint being added.

src/openforms/js/components/admin/form_design/SubmissionFields.js
```diff
@@ -0,0 +1,112 @@
...
+      return null;
+    }
+    return resetResult.data;
+  };
+
+  useEffect(() => {
+    if (confirmationResult === true) {
+      resetCounter();
+    }
+  }, [confirmationResult]);
```
sergei-maertens, line 54: don't use effects to call API endpoints because of user actions (like clicking a button), instead, perform the side effect directly as part of the `onClick` handler. Effects only run on the next render and are harder to debug/follow what's happening.

This also removes the artificial state for `confirmationResult` and generally makes the implementation simpler.

src/openforms/js/components/admin/form_design/SubmissionFields.js
```diff
@@ -0,0 +1,112 @@
...
+              />
+            }
+            helpText={
+              <FormattedMessage
+                description="Successful Submissions Removal Limit help text"
+                defaultMessage="The maximum number of allowed submissions for this form. Leave this empty if no limit is needed."
+              />
+            }
+          >
+            <NumberInput value={submissionMaximumAllowed || null} onChange={handleChange} min="0" />
```
sergei-maertens, line 75: self-reminder: check for react warnings about component changing from uncontrolled to controlled - this happens when the value goes from `undefined` or `null` to a number/string

src/openforms/js/utils/fetch.js
```diff
@@ -113,6 +113,11 @@ const put = async (url, csrftoken, data = {}, throwOn400 = false) => {
   return resp;
 };
 
+const patch = async (url, csrftoken, data = {}, throwOn400 = false) => {
```
sergei-maertens, line 116: new function -> let's go for the optimal interface from the start!

```suggestion
const patch = async (url, csrftoken, data = {}, throwOn400 = true) => {
```

src/openforms/submissions/exceptions.py
```diff
@@ -11,3 +11,10 @@ class FormDeactivated(UnprocessableEntity):
 class FormMaintenance(ServiceUnavailable):
     default_detail = _("The form is currently disabled for maintenance.")
     default_code = "form-maintenance"
+
+
+class FormMaximumSubmissions(ServiceUnavailable):
+    default_detail = _(
+        "The form has currently reached the maximum amount of submissions."
+    )
```
sergei-maertens, line 19:
```suggestion
    default_detail = _("The form has reached the maximum number of submissions.")
```

src/openforms/submissions/exceptions.py
```diff
@@ -11,3 +11,10 @@ class FormDeactivated(UnprocessableEntity):
...
+    default_code = "form-maximum-submissions"
```
sergei-maertens, line 20:
```suggestion
    default_code = "form-submission-limit-reached"
```

src/openforms/submissions/signals.py
```diff
@@ -103,3 +103,12 @@ def increment_form_counter(sender, instance: Submission, **kwargs):
...
         form_statistics.save()
+
+
+@receiver(
+    submission_complete, dispatch_uid="submission.increment_submissions_form_counter"
+)
+def increment_submissions_form_counter(sender, instance: Submission, **kwargs):
+    if instance.form.submission_maximum_allowed:
+        instance.form.submission_counter += 1
+        instance.form.save()
```
sergei-maertens, line 114: Use `F` objects for atomic, race-condition-safe updates

```suggestion
    if instance.form.submission_maximum_allowed:
        instance.form.submission_counter = F("submission_counter") + 1
        instance.form.save()
```

sergei-maertens (changes requested):

src/openforms/forms/admin/form.py
vaszig: So, reading this and the comment below for the serializer's field validation should I add a db constraint?
Or the solution is just to make the form unavailable?

sergei-maertens: just make the form unavailable, the above situation is a valid case when the number of submissions could be higher than the limit.

src/openforms/forms/models/form.py
```diff
@@ -485,6 +517,7 @@ def copy(self):
         )
         copy.slug = _("{slug}-copy").format(slug=self.slug)
         copy.product = self.product
+        copy.submission_counter = 0
```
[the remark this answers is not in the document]
vaszig: I guess we should have the same approach right?Resetting the counter during export.

src/openforms/submissions/tests/test_resume_form_view.py
```diff
@@ -456,3 +457,46 @@ def test_redirects_to_auth_if_form_does_not_require_login_but_user_logged_in_the
...
+            "submissions:resume",
+            kwargs={
+                "token": submission_resume_token_generator.make_token(submission),
+                "submission_uuid": submission.uuid,
+            },
+        )
+
+        response = self.client.get(endpoint)
+
+        self.assertNotEqual(response.status_code, 302)
```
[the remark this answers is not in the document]
vaszig: Basically here the `response.status_code` is 200 (TemplateResponse) and we show just the error message. Only if we click the Start button we get a 503. Is it wrong?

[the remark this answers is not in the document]
sergei-maertens: Ah no, that's okay I think!

src/openforms/forms/api/serializers/form.py
```diff
@@ -513,6 +518,12 @@ def get_cosign_has_link_in_email(self, obj: Form) -> bool:
         config = GlobalConfiguration.get_solo()
         return config.cosign_request_template_has_link
 
+    def get_submission_limit_reached(self, obj: Form) -> bool:
+        if not obj:
+            return False
```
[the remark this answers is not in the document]
vaszig: I removed the check for None. While I was trying to write a test for this I realized that the code is not accessed when the object (Form) is None. Then I read that it's only accessed when an instance exists and saw this in DRF as well. Do you agree?

https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L1839

src/openforms/forms/tests/test_submission_counter.py
sergei-maertens, line 1: I think it makes more sense to put these tests in `/src/openforms/submissions/tests/test_submission_completion.py`

vaszig:
I do not know why test coverage doesn't like the `src/openforms/submissions/utils.py#L306 ` but these lines are covered both for starting and resuming a form.

src/openforms/forms/api/serializers/form.py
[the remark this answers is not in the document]
sergei-maertens: Yes :) these sort of things are quite nice if they integrate well with type-checkers/type annotations!

sergei-maertens (approved):

PR #4893, opened 2024-12-09
🐛 [#4824] Schedule task to create FormVariables after saving FormStep

### LINKED ISSUE #4824: error on PUT to /api/v2/forms/{uuid_or_slug}/variables issues (investigating)

2.7.x and higher (reproduced on our test env)

### Customer reference

_No response_

Seeing cases where errors while saving a form ("variables") lead to a form where no input will be stored in de submission for the new variables. 

This can bee seen in the form itself as navigating back to these steps will simply show empty fields that previously held data. 
These specific variables are also missing in the variables tab of the form. 

These variables are "present" in the form builder, they also show in the front-end of the form but they are missing in the submission, and consequently trigger a validation error if they are required. 

This is either showing up more frequent now because of BE validation, or 2.7 introduced a bug where saving of variables has changed?

suspected ValidationErrors look something like this
`
[{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {'key': [ErrorDetail(string='Ongeldige componentvariabele: er bestaat geen component met de opgegeven sleutel in de formulierdefinitie.', code='invalid')]}, {'key': [ErrorDetail(string='Ongeldige componentvariabele: er bestaat geen component met de opgegeven sleutel in de formulierdefinitie.', code='invalid')]}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
`
This would suggest this form now contains 2 "ghost" variables that unless refreshed by the from builder and saved again, will be missing from every submission

maykin-test form 608 is currently in this state - it contains 2 fields "dummyveld" and "dummyveld1" these were added with a "bad" component that triggered a error. 

Both fields show up in the form designer and front-end, but are not present when creating a submission. 

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
Closes #4824 partially

**Changes**

* Schedule task to create FormVariables after saving FormStep

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/forms/tests/test_api_formsteps.py
+++ b/src/openforms/forms/tests/test_api_formsteps.py
@@ -856,6 +856,103 @@ class FormsStepsAPITests(APITestCase):
         slugs = {form_step.slug for form_step in form_steps}
         self.assertEqual(len(slugs), 2)  # we expect two unique slugs
 
+    @patch(
+        "openforms.forms.api.serializers.form_step.on_formstep_save_event",
+        autospec=True,
+    )
+    @tag("gh-4824")
+    def test_create_form_step_schedules_task_chain_to_fix_form_variables(
+        self, mock_on_formstep_save_event
+    ):
+        """
+        Regression test for https://github.com/open-formulieren/open-forms/issues/4824
+        """
+        assign_change_form_permissions(self.user)
+        form = FormFactory.create()
+        url = reverse("api:form-steps-list", kwargs={"form_uuid_or_slug": form.uuid})
+        form_definition = FormDefinitionFactory.create(
+            configuration={
+                "display": "form",
+                "components": [
+                    {
+                        "key": "lastName",
+                        "type": "textfield",
+                        "label": "Last Name",
+                    },
+                ],
+            }
+        )
+        formdefinition_detail_url = reverse(
+            "api:formdefinition-detail",
+            kwargs={"uuid": form_definition.uuid},
+        )
+        data = {
+            "formDefinition": f"http://testserver{formdefinition_detail_url}",
+            "index": 0,
+        }
+        self.client.force_login(user=self.user)
+
+        with self.captureOnCommitCallbacks(execute=True):
+            response = self.client.post(url, data=data)
+
+        mock_on_formstep_save_event.assert_called_once_with(form.id, 60)
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(
+            FormStep.objects.filter(form_definition=form_definition).count(),
+            1,
+        )
+
+    @patch(
+        "openforms.forms.api.serializers.form_step.on_formstep_save_event",
+        autospec=True,
+    )
+    @tag("gh-4824")
+    def test_update_form_step_schedules_task_chain_to_fix_form_variables(
+        self, mock_on_formstep_save_event
+    ):
+        """
+        Regression test for https://github.com/open-formulieren/open-forms/issues/4824
+        """
+        assign_change_form_permissions(self.user)
+        form_step = FormStepFactory.create()
+        form_definition = FormDefinitionFactory.create(
+            configuration={
+                "display": "form",
+                "components": [
+                    {
+                        "key": "lastName",
+                        "type": "textfield",
+                        "label": "Last Name",
+                    },
+                ],
+            }
+        )
+        url = reverse(
+            "api:form-steps-detail",
+            kwargs={"form_uuid_or_slug": form_step.form.uuid, "uuid": form_step.uuid},
+        )
+        formdefinition_detail_url = reverse(
+            "api:formdefinition-detail",
+            kwargs={"uuid": form_definition.uuid},
+        )
+        data = {
+            "formDefinition": f"http://testserver{formdefinition_detail_url}",
+            "index": 0,
+        }
+        self.client.force_login(user=self.user)
+
+        with self.captureOnCommitCallbacks(execute=True):
+            response = self.client.put(url, data=data)
+
+        mock_on_formstep_save_event.assert_called_once_with(form_step.form.id, 60)
+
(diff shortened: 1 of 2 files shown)


src/openforms/forms/api/viewsets.py
```diff
@@ -102,6 +103,34 @@ def get_serializer_context(self):
...
+
+        form = get_object_or_404(Form, uuid=self.kwargs["form_uuid_or_slug"])
+        transaction.on_commit(
+            lambda: create_form_variables_for_components.apply_async(
+                args=(form.id,),
+                countdown=60,
+            )
+        )
+
+        return response
```
sergei-maertens, line 132: Would it make sense to implement this on the `FormStepSerializer.save` instead? That one calls out to `create/update`, so you'd only have to do it once, and there you have access to `self.instance.form` to look up the form ID.

Also - lambda's have late binding, it's safer to use `partial` instead.

src/openforms/forms/tests/test_api_formsteps.py
```diff
@@ -15,6 +15,7 @@
     TokenFactory,
     UserFactory,
 )
+from openforms.forms.tasks import create_form_variables_for_components
```
sergei-maertens, line 18:
```suggestion
from ..tasks import create_form_variables_for_components
```

src/openforms/forms/tasks.py
```diff
@@ -96,6 +96,29 @@ def repopulate_reusable_definition_variables_to_form_variables(form_id: int) ->
...
+    from .models import Form, FormVariable  # due to circular import
+
+    form = Form.objects.get(id=form_id)
+
+    # delete the existing form variables, we will re-create them
+    FormVariable.objects.filter(
+        form=form_id, source=FormVariableSources.component
+    ).delete()
+
+    FormVariable.objects.create_for_form(form)
```
sergei-maertens, line 119: shouldn't this also be followed by the recouple task? E.g. when you're updating a form that has existing submissions.

src/openforms/forms/tests/test_api_formsteps.py
```diff
@@ -856,6 +857,99 @@ def test_with_non_unique_fd_slugs(self):
         slugs = {form_step.slug for form_step in form_steps}
         self.assertEqual(len(slugs), 2)  # we expect two unique slugs
 
+    @tag("gh-4824")
+    def test_create_form_step_schedules_recouple_task(self):
```
sergei-maertens, line 861: wrong test name?

src/openforms/forms/tests/test_api_formsteps.py
```diff
@@ -856,6 +857,99 @@ def test_with_non_unique_fd_slugs(self):
...
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+            self.assertEqual(
+                FormStep.objects.filter(form_definition=form_definition).count(),
+                1,
+            )
+
+            mock_apply_async.assert_called_once_with(args=(form.id,), countdown=60)
+
+    @tag("gh-4824")
+    def test_update_form_step_schedules_recouple_task(self):
```
sergei-maertens, line 905: wrong test name?

sergei-maertens (changes requested):

src/openforms/forms/api/viewsets.py
stevenbal: you're right, I'll move it

src/openforms/forms/tasks.py
stevenbal: Hmm yeah that's right and I think the `repopulate_reusable_definition_variables_task` needs to be run as well, because if the FormDefinition that you're updating is reusable, those other Forms need their FormVariables updated as well

src/openforms/forms/tests/test_api_formsteps.py
```diff
@@ -856,6 +856,133 @@ def test_with_non_unique_fd_slugs(self):
...
+        self.assertEqual(
+            FormStep.objects.filter(form_definition=form_definition).count(),
+            1,
+        )
+
+    @patch("openforms.forms.tasks.create_form_variables_for_components.si")
+    @patch(
+        "openforms.forms.tasks.repopulate_reusable_definition_variables_to_form_variables.si"
+    )
+    @patch("openforms.forms.tasks.recouple_submission_variables_to_form_variables.si")
```
sergei-maertens, line 925: doesn't it make more sense to ensure that the `on_post_step_save` is called with the right arguments, and then separately write a unit test without any mocking/patching that calls that function with `CELERY_TASK_ALWAYS_EAGER` to check that the desired state is achieved?

I get nervous from seeing `unittest.mock` because it doesn't give me any guarantees that the code actually does what it's supposed to do, it's testing implementation details and in tests I really want to stay away from implementation details and be able to freely change/refactor them without having to touch tests.

sergei-maertens (changes requested):

src/openforms/forms/tests/variables/test_tasks.py
```diff
@@ -205,3 +209,143 @@ def race_condition():
...
+                submission__form=form
+            )
+
+            self.assertEqual(submission_vars1.count(), 1)
+
+            [first_name_submission_var1] = submission_vars1
+
+            self.assertEqual(first_name_submission_var1.form_variable, first_name_var1)
+
+        # with self.subTest(
```
stevenbal, line 340: This currently doesn't work, created a separate issue for it https://github.com/open-formulieren/open-forms/issues/4900

PR #4894, opened 2024-12-09
:boom: Make TemporaryFileUpload.submission non-nullable

sergei-maertens:
Part of #3283

We've provided a transition period from Open Forms 2.7, which ensures that a submission is stored. Support for legacy uploads is now removed.

The celery background jobs should by now have deleted all the old unclaimed uploads, unless absurdly large retention values have been set.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene


sergei-maertens:
-> need to add an upgrade check script

docs/installation/upgrade-300.rst
```diff
@@ -103,3 +103,15 @@ Removal of /api/v2/location/get-street-name-and-city endpoint
...
+
+Legacy temporary file uploads no longer supported
+=================================================
+
+Before Open Forms 2.7.0, temporary file uploads (as created by the file upload form
+component) would not be related to the submission they were uploaded in. We call these
+legacy temporary file uploads, and support for them has been removed in Open Forms 3.0.
+
+The setting ``TEMPORARY_UPLOADS_REMOVED_AFTER_DAYS`` controls how long file uploads are
+kept. Ensure that this many days have passed since the last legacy upload before
```
vaszig, line 115:
```suggestion
kept. Ensure that many days have passed since the last legacy upload before
```

vaszig (changes requested):
I think there are some more pieces left like here:
https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/components/vanilla.py#L387

docs/installation/upgrade-300.rst
sergei-maertens: "this" refers to the value of the setting here, e.g. "Ensure that 5 days have passed since..."

sergei-maertens:
> I think there are some more pieces left like here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/formio/components/vanilla.py#L387

is it not showing up in the diff for you? I've remove this check since I came across it when <kbd>CTRL</kbd> + <kbd>F</kbd> on 'legacy'.

vaszig (approved):

Issue #4788, opened 2024-10-24
Add support for Admin-OIDC configuration via django-setup-configuration


sjoerdie:
### Thema / Theme

Frontend

### Omschrijving / Description

Add django-setup-configuration to configure admin OIDC for openforms

stevenbal:
@sergei-maertens I thought a bit about this and here are my thoughts, let me know what you think:

- If I'm correct, all that is needed to configure the Objects API registration backend is part of the `ObjectsAPIGroupConfig` model
- I don't think it makes a lot of sense to split the configuration of Objecten and Objecttypen (and probably also Documenten and Catalogi), since they are also coupled in the `ObjectsAPIGroupConfig` model. So I would probably group that configuration in a single fixture (as well as the underlying Services probably?)
- Using environment variables to configure this doesn't seem like a good solution, because it should be possible to configure any amount of `ObjectsAPIGroupConfig`. I suppose we could add an envvar that points to a fixture (.yaml/.json) and load that in the `ConfigurationStep`. It does feel like setup-configuration is just a layer on top of `loaddata` then, but I suppose if you include it as a regular fixture there is no way to control when the data will be loaded, while with setup_configuration you deliberately use the command to do so
- We should probably add a simple command to easily create the fixture (which should use natural keys)

PR #4897, opened 2024-12-10
:arrow_up: [#4788] Upgrade mozilla-django-oidc-db to 0.21.1

stevenbal:
the previous release specified oidc_op_jwks_endpoint and oidc_op_logout_endpoint outside of endpoint_config


stevenbal:
@sergei-maertens I found out about this after the merge yesterday, two endpoint fields were specified outside of the `endpoint_config` key, so I made a patch release to ensure all endpoints are grouped together: https://github.com/maykinmedia/mozilla-django-oidc-db/blob/master/tests/setupconfig/files/full_setup.yml#L19-L20

sergei-maertens (approved):

PR #4901, opened 2024-12-10
:fire: Delete migration test code

sergei-maertens:
The tests rely too much on the real models, which breaks other PRs that add additional fields. Refactoring the test code to freeze the price calculation is not reasonable, because it relies on the logic engine evaluation, which can't work with historic models and requires access to the real models and their models.

We'll have to trust that the tests were sufficient when the code cleanup was initially authored :fingers_crossed:


vaszig (approved):
After rebasing the tests are passing!

PR #4905, opened 2024-12-11
[#4895] outgoing email logging improvements

### LINKED ISSUE #4895: Outgoing email logging improvements
We support sending custom email headers in all emails sent by Open Forms. Currently, this is only used for registration emails. See

https://github.com/open-formulieren/open-forms/blob/dd38e638d1b00101cc37967c897bb3e6a753d991/src/openforms/emails/constants.py#L4

While debugging a production environment to figure out what type of emails were sent, I had to search on the content of the email which is fragile because different forms can use completely different templates rather than a single global configuration. This would've been easier if the registration emails would have been properly tagged with the custom headers so I could filter on this header instead (I was able to exclude registration emails this way).

Additionally, there's a typo in the content type for the `Submission` model - the app label is singular instead of plural.

**Tasks**

- [x] Ensure that custom email headers are used for confirmation emails
- [x] Ensure that custom email headers are used for cosign request emails
- [x] Fix content type for submissions in email constants

viktorvanwijk:
Closes #4895 

**Changes**

Add custom headers to confirmation and cosign emails

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/tasks/emails.py
+++ b/src/openforms/submissions/tasks/emails.py
@@ -144,7 +144,7 @@ def send_email_cosigner(submission_id: int) -> None:
                     "Content-Language": submission.language_code,
                     X_OF_CONTENT_TYPE_HEADER: EmailContentTypeChoices.submission,
                     X_OF_CONTENT_UUID_HEADER: str(submission.uuid),
-                    X_OF_EVENT_HEADER: EmailEventChoices.cosign_confirmation,
+                    X_OF_EVENT_HEADER: EmailEventChoices.cosign_request,
                 },
             )
         except Exception:
--- a/src/openforms/submissions/tasks/emails.py
+++ b/src/openforms/submissions/tasks/emails.py
@@ -10,6 +10,13 @@ from celery_once import QueueOnce
 
 from openforms.celery import app
 from openforms.config.models import GlobalConfiguration
+from openforms.emails.constants import (
+    X_OF_CONTENT_TYPE_HEADER,
+    X_OF_CONTENT_UUID_HEADER,
+    X_OF_EVENT_HEADER,
+    EmailContentTypeChoices,
+    EmailEventChoices,
+)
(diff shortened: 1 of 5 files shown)


src/openforms/appointments/tests/test_confirmation_emails.py
```diff
@@ -302,3 +303,33 @@ def test_cancel_and_change_instructions(self):
         with self.subTest(type="plain text"):
             self.assertIn(cancel_link, message_text)
             self.assertNotIn(change_link, message_text)
+
+    def test_headers_present_in_confirmation_email(self):
```
sergei-maertens, line 307: I think it's better to have this test in the submissions app instead of appointments, since it's not something exclusive or special to appointments - even for non-appointment submissions we expect these confirmation emails (with custom headers) to be sent.

src/openforms/submissions/tasks/emails.py
```diff
@@ -133,6 +140,12 @@ def send_email_cosigner(submission_id: int) -> None:
                 from_email=settings.DEFAULT_FROM_EMAIL,
                 recipient_list=[recipient],
                 text_message=content,
+                extra_headers={
+                    "Content-Language": submission.language_code,
+                    X_OF_CONTENT_TYPE_HEADER: EmailContentTypeChoices.submission,
+                    X_OF_CONTENT_UUID_HEADER: str(submission.uuid),
+                    X_OF_EVENT_HEADER: EmailEventChoices.cosign_confirmation,
```
sergei-maertens, line 147: This email isn't a cosign confirmation email, but rather a cosign request. To be fair, I think the original code defining the event is wrong and should be renamed to 'cosign-request', since the cosign confirmation is not much different from the "normal" confirmation email (if I remember correctly).

sergei-maertens (changes requested):

sergei-maertens (approved):
Nice one!

PR #4911, opened 2024-12-12
:boom: Remove legacy machtigen context

sergei-maertens:
Part of #3283

The shape of data that we store was badly documented, not in the least because we barely had any idea what kind of information we receive.

Since Open Forms 2.7.0 it has been fleshed out what the mandata context looks like, and properly implemented backed by a proposal for a standard which is implemented all the way to Open Zaak to try out.

This legacy context is now being removed in favour of the new way to track and expose the mandata context.

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Dockerfile/scripts

  - [x] Updated the Dockerfile with the necessary scripts from the `./bin` folder

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

PR #4912, opened 2024-12-12
:boom: Disable legacy OIDC endpoints by default

sergei-maertens:
Part of #3283 - stumbled on this while working on #4911

**Changes**

While the removal of the legacy endpoints is postponed to Open Forms 4.0, I thought we should grab the chance now to switch the default value for the configuration options from legacy-enabled to legacy-disabled. This makes it more likely that fresh instances are set up with the best situation possible and won't have to deal with breaking changes later for 4.0.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/docs/installation/upgrade-300.rst
+++ b/docs/installation/upgrade-300.rst
@@ -14,6 +14,29 @@ be aware of, as they may require additional manual actions.
    :depth: 1
    :local:
 
+Legacy OpenID Connect callback endpoints are now disabled by default
+====================================================================
+
+Before Open Forms 3.0, the legacy endpoints were used by default.
+
+The following environment variables now default to ``False`` instead of ``True``:
+
+* ``USE_LEGACY_OIDC_ENDPOINTS``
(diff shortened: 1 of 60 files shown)


vaszig (approved):
Regarding the VCR tests I made sure that I see in the diff all the directories related to the OIDC.

docs/installation/upgrade-300.rst
```diff
@@ -14,6 +14,17 @@ be aware of, as they may require additional manual actions.
...
+====================================================================
+
+Before Open Forms 3.0, the legacy endpoints were used by default.
+
+The following environment variables now default to ``False`` instead of ``True``:
+
+* ``USE_LEGACY_OIDC_ENDPOINTS``
+* ``USE_LEGACY_DIGID_EH_OIDC_ENDPOINTS``
+* ``USE_LEGACY_ORG_OIDC_ENDPOINTS``
+
```
sergei-maertens, line 27: Make more explicit on what to do if you don't want to deal with this change now.

PR #4915, opened 2024-12-12
Typechecking on API URLs (mostly error suppression)

sergei-maertens:


src/openforms/utils/urls.py
```diff
@@ -13,6 +14,9 @@
 
 from openforms.typing import AnyRequest
 
+if TYPE_CHECKING:
+    from django.conf.urls import _IncludedURLConf
```
vaszig, line 18: Is it good that we are using Django's private API here?

vaszig (approved):

src/openforms/utils/urls.py
sergei-maertens: this is not actually private API, but types defined by `django-stubs`! this import does not exist at runtime, and that's why it's guarded by the `if TYPE_CHECKING:` in combination with `from __future__ import annotations`

PR #4916, opened 2024-12-12
Make service FKs in API group config models not nullable

sergei-maertens:
Part of #3283

**Changes**

* Added check script for data that would violate the new constraints
* Added check script to upgrade path checks
* Made the FK fields on Objects/ZGW API group models `null=False` for the non-optional fields.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


bin/check_api_groups_null.py
```diff
@@ -0,0 +1,80 @@
...
+    print(
+        "Go into the admin to fix their configuration, and then try to upgrade again."
+    )
+    print(
+        tabulate(
+            problems,
+            headers=("API group type", "ID", "Name", "Problem"),
+        )
+    )
+    return True
```
stevenbal, line 66: Not a major thing, but does this have to return something?

stevenbal (approved):

bin/check_api_groups_null.py
sergei-maertens: yeah, the return value is checked by the upgrade check which blocks migrating if problems are detected! See https://github.com/open-formulieren/open-forms/blob/9be9f8c4c812175022a1268e2aabc49c09f64785/src/openforms/upgrades/upgrade_paths.py#L62

PR #4947, opened 2024-12-18
org-oidc auth plugin employee login

### LINKED ISSUE #4863: oidc_org: RuntimeError Unknown attribute: employee_id

2.7.9 + 2.8.1

### Customer reference

Utrecht: 313

I cannot start a form as a employee selecting this option crashes the form:
sentry: 380096

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #4863

**Changes**

This should solve the problem of employee login on forms with the org-oidc auth plugin.

Currently this is a WIP. I'm not sure if this is the right approach, and solves the right issue (this might be a very superficial solution, but im not sure). This currently contains:
- Test that showcases the failing scenario
- Solution for employee_id authentication
   (Which is highly experimental, and probably should/will change)

This that still should be done:
- Add employee id to user
- Use employee_id as identifier for employee case in `to_auth_context_data` 
  (I think this requires passing employee id from the org-oidc plugin `handle_return()`)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/authentication/models.py
+++ b/src/openforms/authentication/models.py
@@ -20,6 +20,7 @@ from .types import (
     DigiDMachtigenContext,
     EHerkenningContext,
     EHerkenningMachtigenContext,
+    EmployeeContext,
 )
 
 logger = logging.getLogger(__name__)
@@ -254,6 +255,7 @@ class AuthInfo(BaseAuthInfo):
         | DigiDMachtigenContext
         | EHerkenningContext
         | EHerkenningMachtigenContext
+        | EmployeeContext
     ):
         if self.attribute_hashed:
             logger.debug(
@@ -347,6 +349,25 @@ class AuthInfo(BaseAuthInfo):
                         "branchNumber"
                     ] = branch_number
                 return ehm_context
+
+            # Employee login
+            # ---
+            # Experimental feature that is different from the defined authentication context.
+            # This will have to be re-worked and re-thought-out, and will most definitely change.
+            # This fixes the issue of #4863
+            case (AuthAttribute.employee_id, ""):
+                employee_context: EmployeeContext = {
+                    "source": "custom",
+                    "levelOfAssurance": "unknown",
+                    "authorizee": {
+                        "legalSubject": {
+                            "identifierType": "opaque",
+                            "identifier": self.value,
+                        }
+                    },
+                }
+                return employee_context
+
             case _:  # pragma: no cover
                 raise RuntimeError(f"Unknown attribute: {self.attribute}")
 
--- a/src/openforms/authentication/types.py
+++ b/src/openforms/authentication/types.py
@@ -80,3 +80,19 @@ class EHMandate(TypedDict):
 class EHerkenningMachtigenContext(EHerkenningContext):
     representee: EHRepresenteeEntity
     mandate: EHMandate
+
+
+class EmployeeLegalSubject(TypedDict):
+    identifierType: Literal["opaque"]
+    identifier: str
+
+
+class EmployeeAuthorizee(TypedDict):
+    legalSubject: EmployeeLegalSubject
+
+
+# This type definition is highly experimental, and will most definitely change
+class EmployeeContext(TypedDict):
+    source: Literal["custom"]
+    levelOfAssurance: Literal["unknown"]
+    authorizee: EmployeeAuthorizee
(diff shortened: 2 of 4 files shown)


src/openforms/authentication/models.py
```diff
@@ -347,6 +349,21 @@ def to_auth_context_data(
...
+            case (AuthAttribute.employee_id, ""):
+                employee_context: EmployeeContext = {
+                    "source": "custom",
+                    "levelOfAssurance": self.loa,
+                    "representee": {
+                        "identifierType": "employee_id",
+                        "identifier": self.value,
+                    },
+                }
+                return employee_context
```
sergei-maertens, line 365: Some background to explain more about how things work in Open Forms!

The model `AuthInfo` (which defines the `to_auth_context_data` method) captures who is authenticated for the submission. This means that if an employee of the organization (after OIDC login) logs in for the form for themselves (the other flow I will describe below), they are the `authorizee.legalSubject` (representing themselves). This is a rather simple setup of employee context that needs to be returned if we follow the established pattern (where representee is implied from the authorizee/legal subject).

```json
{
  "source": "custom",
  "loa": "unknown",
  "authorizee": {
    "legalSubject": {
      "identifierType": "opaque",
      "identifier": self.value
    }
  }
}
```

Identifier type opaque means you can't derive any meaning/format validation from it.

The other flow is where an employee is logged in with OIDC, and they enter the BSN or KVK of the customer. In the code, this brings you to `openforms.authentication.signals.set_auth_attribute_on_session` in the branch with `if registrator_subject`. In that situation, on the `AuthInfo` instance not the employee details are stored, but the BSN/KVK nr of the actual customer, and the `attribute` is then set to `bsn` or `kvk` rather than `employee_id`, so you can never end up in this case in those flows. Meaning, the `to_auth_context_data` only needs to consider the situation where an employee is acting on their own behalf which is nice because it simplifies things a bit.

Looking at the Sentry error, this is the right place to patch this up.

So, key takeaways:

* the context dict/shape here needs to use `authorizee > legalSubject` structure, similar to DigiD without machtigen (first case)
* we don't need to consider "acting on behalf of someone else" for employee_id attribute

src/openforms/authentication/types.py
```diff
@@ -80,3 +80,15 @@ class EHMandate(TypedDict):
 class EHerkenningMachtigenContext(EHerkenningContext):
     representee: EHRepresenteeEntity
     mandate: EHMandate
+
+
+class EmployeeRepresenteeEntity(TypedDict):
+    identifierType: Literal["employee_id"]
```
sergei-maertens, line 86: I'd use either `opaque` or `custom` here. I gravitate towards the former because it's already used for the eHerkenning acting subject too

docker/keycloak/README.md
```diff
@@ -52,6 +52,7 @@ VCR.py). The primary reason this setup exists, is for automated testing reasons.
   attributes from `eherkenning-bewindvoering`.
 - `admin` / `admin`, intended to create as django user (can be made staff). The email address is
   `admin@example.com`.
+- `employee` / `employee`, has the `employee_id` attribute (for org-oidc)
```
sergei-maertens, line 55: this isn't necessary, since you can populate the user model field from already existing claims, see the documentation: https://open-forms.readthedocs.io/en/stable/configuration/authentication/oidc_org.html#configuring-oidc-for-login-of-organization-members

In these cases I would probably just map the `sub` or `username` or `email` claim to the `employee_id` field.

robinmolen: But isn't this needed for the login itself?

sergei-maertens: you can use the `admin/admin` user for that :)

src/openforms/authentication/types.py
robinmolen: Check, yeah i'll use `opaque` then it that case, to keep things consistent 👍

src/openforms/authentication/models.py
robinmolen: Okay.. I think i'm starting to understand this a bit more..

Should the context then not also contain the `representee` data? (Although this would probably be the same information.., right?)

sergei-maertens: if no representee is specified, it is indeed (implicitly) equal to `authorizee.legalSubject`

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.8.x: d428e39e2b9566048db3195790a984524827dd7d
* 2.7.x: a6568b997ea3e1f09d359ff9211033704a5bd937

PR #4952, opened 2024-12-18
[#4821] Fixed email digest for BRK and addressNL component

### LINKED ISSUE #4821: Daily digest always reports BRK issues, for all users/environements regardless of configuration.

2.7.9+

### Customer reference

anyone with a addressNL component in a form

```
De instellingen voor 'BRK client' zijn niet geldig (KVK endpoint is niet geconfigureerd). 
Component of plugins die hiervan gebruik maken zullen niet goed werken.
```
the configuration:
image

While the check is very thorough - at least a BRK service should be set for this check to make sense for the users?

vaszig:
Closes #4821

**Changes**

- We want the email digest to report problems (broken configuration) for the addressNL component only when a valid BRK service and the validator are defined/configured.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/contrib/brk/service.py
+++ b/src/openforms/contrib/brk/service.py
@@ -1,16 +1,28 @@
+from glom import glom
+
 from openforms.forms.models.form import Form
 from openforms.plugins.exceptions import InvalidPluginConfiguration
 
 from .checks import BRKValidatorCheck
+from .models import BRKConfig
+from .validators import BRK_ZAKELIJK_GERECHTIGD_VALIDATOR_ID
 
 
 def check_brk_config_for_addressNL() -> str:
     live_forms = Form.objects.live()
 
-    if any(form.has_component("addressNL") for form in live_forms):
-        try:
-            BRKValidatorCheck.check_config()
-        except InvalidPluginConfiguration as e:
-            return e.args[0]
+    for form in live_forms:
+        components = form.iter_components()
+        for component in components:
+            if (
+                component["type"] == "addressNL"
+                and (plugins := glom(component, "validate.plugins", default=[]))
+                and BRK_ZAKELIJK_GERECHTIGD_VALIDATOR_ID in plugins
+                and BRKConfig.get_solo().service
+            ):
+                try:
+                    BRKValidatorCheck.check_config()
+                except InvalidPluginConfiguration as e:
+                    return e.args[0]
 
     return ""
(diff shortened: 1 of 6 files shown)


src/openforms/contrib/brk/service.py
```diff
@@ -1,16 +1,26 @@
...
-        except InvalidPluginConfiguration as e:
-            return e.args[0]
+    for form in live_forms:
+        components = form.iter_components()
+        for component in components:
+            if (
+                component["type"] == "addressNL"
+                and glom(component, "validate.plugins", default=[])
+                and BRKConfig.get_solo().service
+            ):
```
sergei-maertens, line 20:
```suggestion
            if (
                component["type"] == "addressNL"
                and (plugins := glom(component, "validate.plugins", default=[]))
                and "brk-zakelijk-gerechtigd" in plugins
                and BRKConfig.get_solo().service
            ):
```

only if this particular plugin is used do we need to perform the check, if other plugins are used that don't use the BRK, we don't want this false positive to happen again.

It would be a good idea to define that plugin ID as a constant in `src/openforms/contrib/brk/validators.py` instead of hardcoding it

src/openforms/contrib/brk/tests/base.py
```diff
@@ -19,6 +19,13 @@
     header_value=BRK_API_KEY,
 )
 
+INVALID_BRK_SERVICE = ServiceFactory.build(
+    api_root="https://api.brk.kadaster.nl/invalid",
+    oas="https://api.brk.kadaster.nl/esd-eto-apikey/bevragen/v2/",  # ignored/unused
```
sergei-maertens, line 24: you can omit the `oas`, it's disabled through `settings.ZGW_CONSUMERS_IGNORE_OAS_FIELDS`

src/openforms/contrib/brk/validators.py
```diff
@@ -21,6 +21,8 @@
 
 logger = logging.getLogger(__name__)
 
+BRK_BUSINESS_VALIDATOR = "brk-zakelijk-gerechtigd"
```
sergei-maertens, line 24: this is such a highly specific term, I don't think it's wise to attempt to translate it

```suggestion
BRK_ZAKELIJK_GERECHTIGD_VALIDATOR_ID = "brk-zakelijk-gerechtigd"
```

sergei-maertens (changes requested):

sergei-maertens (approved):
Let's backport this too :thinking:

PR #4953, opened 2024-12-18
Allow more map interactions and refactor map value datastructures

robinmolen:
Part of #2177
Part of #2178

**Changes**

Changing the datatype of the map component from array to geoJson

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/formio/components/custom.py
```diff
@@ -191,6 +192,53 @@ def build_serializer_field(
         return serializers.ListField(child=base) if multiple else base
 
 
+class GeoJsonPointGeometrySerializer(serializers.Serializer):
+    coordinates = serializers.ListField(
+        child=serializers.FloatField(required=True, allow_null=False),
+        min_length=2,
+        max_length=2,
+    )
```
sergei-maertens, line 200: it's probably a good idea to extract this into a custom field and define it in `openforms.api.fields`:

```py
class GeoJSONCoordinatesField(serializers.ListField):
    child = serializers.FloatField(required=True, allow_null=False),

        kwargs["min_length"] = 2
        kwargs["max_length"] = 2
        super().__init__(*args, **kwargs)
```

which you can then easily (re) use as:

```py
class GeoJSONPointGeometrySerializer(serializers.Serializer):
    coordinates = GeoJSONCoordinatesField()

class GeoJSONLineStringGeometrySerializer(serializers.Serializer):
    coordinates = serializers.ListField(child=GeoJSONCoordinatesField())
```
etc.

I'd even opt to define all of this (including the serializers) in `openforms.api.geojson` rather than `openforms.api.fields` :)

src/openforms/formio/components/custom.py
```diff
@@ -191,6 +192,53 @@ def build_serializer_field(
...
+
+class GeoJsonGeometryPolymorphicSerializer(PolymorphicSerializer):
+    type = serializers.RegexField("^Point|LineString|Polygon$", required=True)
+
+    discriminator_field = "type"
+    serializer_mapping = {
+        str("Point"): GeoJsonPointGeometrySerializer,
+        str("LineString"): GeoJsonLineStringGeometrySerializer,
+        str("Polygon"): GeoJsonPolygonGeometrySerializer,
+    }
```
sergei-maertens, line 233: probably better to define a `models.TextChoices` for the `type` and use an enum-approach rather than regex. This avoids magic strings by centralizing them and makes it clear what we support and what we don't.

Additionaly, `str("Point")` is exactly the same as `"Point"`, so this casting is not needed.

src/openforms/formio/components/custom.py
```diff
@@ -191,6 +192,53 @@ def build_serializer_field(
...
+
+
+class GeoJsonSerializer(serializers.Serializer):
+    type = serializers.RegexField("^Feature$", required=True)
+    properties = serializers.DictField()
+    geometry = GeoJsonGeometryPolymorphicSerializer()
```
sergei-maertens, line 239: do we still need the `Feature` container now that circles are dropped? I'd expect that just `GeoJsonGeometryPolymorphicSerializer` is sufficient.

src/openforms/formio/constants.py
```diff
@@ -11,3 +13,9 @@
     "editgrid": "array",
     "datetime": "datetime",
 }
+
+
+class GeoJsonGeometryTypes(models.TextChoices):
+    point = "Point", "Point"
+    polygon = "Polygon", "Polygon"
+    line_string = "LineString", "LineString"
```
sergei-maertens, line 21: please *always* mark labels as translatable. And I think this also belongs in `openforms.api.geojson` rather than in this file, since nothing about this is formio specific :)

src/openforms/formio/formatters/custom.py
```diff
@@ -21,10 +22,26 @@ def format(self, component: Component, value: str) -> str:
         return f"{fmt_date(parsed_value)} {fmt_time(parsed_value, 'H:i')}"
 
 
+class GeoJsonGeometry(TypedDict):
+    type: str
+    coordinates: list[float] | list[list[float]] | list[list[list[float]]]
```
sergei-maertens, line 27: you can create a discriminated union here:

```py
type PointCoordinates = list[float]  # tuple[float, float] would be nicer because that validates the length, list doesn't!

class PointGeometry(TypedDict):
    type: Literal["Point"]
    coordinates: PointCoordinates


class LineGeometry(TypedDict):
    type: Literal["LineString"]
    coordinates: list[PointCoordinates]

...

type GeoJSONGeometry = PointGeometry | LineGeometry | ...
```

src/openforms/formio/components/custom.py
robinmolen: Ah check, yeah that sounds nice! Yeah a separate `openforms.api.geojson` sounds better than `openforms.api.fields` 👍

src/openforms/formio/components/custom.py
robinmolen: Check, okay will do

src/openforms/formio/components/custom.py
robinmolen: Yeah the `type` and `properties` don't add anything right now, so these can be removed. Also the registration backends any need the `geometry`, so we also dont need it for them..

I've already changed this in the sdk (that only the geometry is saved) and will now add it to the backend

src/openforms/formio/constants.py
robinmolen: Ah okay, i thought translatable didn't add anything here, seeing these are constants in the code.. But yeah, i'll add it (and take a look around to see where this is used/shown)

Yeah i'll move it to `openforms.api.geojson` 👍

src/openforms/formio/formatters/custom.py
robinmolen: Ah okay, yeah thats a lot better.. I was struggling with this.

Also a lot nicer that `Literal` is an option here.. I didn't know that this was possible 👌

src/openforms/formio/constants.py
sergei-maertens: API documentation would pick up on these, for instance. But even if they're not shown anywhere right now, they may be in the future and then it's best if the translations are already there.

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -47,5 +127,271 @@ def test_map_field_min_max_length_of_items(self):
 
                 self.assertFalse(is_valid)
                 self.assertIn(component["key"], errors)
-                error = extract_error(errors, component["key"])
+
+                error = extract_error(errors["foo"], "coordinates")
+                error_codes = _recursive_get_error_code(error)
+                self.assertEqual(error_codes, error_code)
```
sergei-maertens, line 133: I don't understand this `_recursive_get_error_code` and that's not good when I'm trying to read/debug a test.

what I expect to see here in the assertions is this:

```py
error = extract_error(errors["foo"], "coordinates")
self.assertEqual(error.code, error_code)
```

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -47,5 +127,271 @@ def test_map_field_min_max_length_of_items(self):
 
                 self.assertFalse(is_valid)
                 self.assertIn(component["key"], errors)
```
sergei-maertens, line 129:
```suggestion
                self.assertIn("foo", errors)
```
I don't like this indirection in tests :) there's only a single component used in all the cases - it'd be a different story if you're looping over components with different keys.

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -47,5 +127,271 @@ def test_map_field_min_max_length_of_items(self):
...
+                    "foo": {"type": "LineString", "coordinates": [1, 1]},
+                },
+                ["not_a_list"],
+            ),
+            (
+                {
+                    "foo": {"type": "LineString", "coordinates": [[[1, 2]]]},
+                },
+                [["invalid"]],
+            ),
```
sergei-maertens, line 227: I'd break these out into separate, explicit `subTest` cases that are not done in the forloop, so you can avoid having to recurse to get error codes. The problem with recursion is that you have no control about the depth here, and we are very specifically looking at a particular depth/location of the error

So the first case would become something like:

```py
coordinate_errors = extract_error(errors["foo"], "coordinates")
self.assertIsInstance(coordinate_errors, list)
self.assertEqual(len(coordinate_errors), 1)
self.assertEqual(coordinate_errors[0].code, "not_a_list")
```
and this yields a much more simple (and thus easier to debug/maintain) test

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -4,6 +4,20 @@
 from .helpers import extract_error, validate_formio_data
 
 
+def _recursive_get_error_code(error):
```
sergei-maertens, line 7: it should be possible to rewrite everything to be more explicit which makes this helper obsolete - see further remarks.

Test code should ideally be readable by absolute beginner developers that barely know to write an if/else statement and give them the feeling they understand what's going on. So, dead-simple!

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -47,5 +127,271 @@ def test_map_field_min_max_length_of_items(self):
...
+                    "foo": {"type": "Polygon", "coordinates": [[1, 1]]},
+                },
+                [["not_a_list"], ["not_a_list"]],
+            ),
+            (
+                {
+                    "foo": {"type": "Polygon", "coordinates": [[[[1, 2]]]]},
+                },
+                [[["invalid"]]],
+            ),
```
sergei-maertens, line 338: same as earlier :)

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -47,5 +127,271 @@ def test_map_field_min_max_length_of_items(self):
...
+        ]
+
+        for data, error_code in invalid_values:
+            with self.subTest(data=data):
+                is_valid, errors = validate_formio_data(component, data)
+
+                self.assertFalse(is_valid)
+                self.assertIn(component["key"], errors)
```
sergei-maertens, line 346:
```suggestion
                self.assertIn("foo", errors)
```

src/openforms/formio/tests/validation/test_map.py
```diff
@@ -47,5 +127,271 @@ def test_map_field_min_max_length_of_items(self):
...
+        ]
+
+        for data, error_code in invalid_values:
+            with self.subTest(data=data):
+                is_valid, errors = validate_formio_data(component, data)
+
+                error = extract_error(errors["foo"]["coordinates"][0], 0)
+
+                self.assertFalse(is_valid)
+                self.assertEqual(error.code, error_code)
```
sergei-maertens, line 382:
```suggestion
                self.assertFalse(is_valid)
                error = extract_error(errors["foo"]["coordinates"][0], 0)
                self.assertEqual(error.code, error_code)
```

src/openforms/formio/tests/validation/test_map.py
sergei-maertens, line 1: I probably missed some spots, but the pattern of remarks is to be applied to all the tests of course.

src/openforms/formio/formatters/custom.py
```diff
@@ -21,10 +21,34 @@ def format(self, component: Component, value: str) -> str:
         return f"{fmt_date(parsed_value)} {fmt_time(parsed_value, 'H:i')}"
 
 
+type Coordinates = tuple[float, float]
+
+
+class PointGeometry(TypedDict):
```
sergei-maertens, line 27: wouldn't it make more sense to define these types in `src/openforms/api/geojson.py` too? Ultimately these are the types that match the serializers.

Otherwise I'd say to put them in `openforms.typing`.

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -69,11 +70,8 @@
 logger = logging.getLogger(__name__)
 
 
-def _point_coordinate(value: Any) -> dict[str, Any] | object:
-    if not isinstance(value, list) or len(value) != 2:
-        return SKIP
-    # Providing the coordinates as [lng, lat] #4955
-    return {"type": "Point", "coordinates": [value[1], value[0]]}
+def _point_coordinate(value: GeoJsonGeometry) -> GeoJsonGeometry:
+    return value
```
sergei-maertens, line 74: that name does not match what the value has - since the value is now any geojson geometry. It's also not doing anything, so isn't it better to just remove it entirely and use the value directly?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -75,11 +76,8 @@ def get_property_mappings_from_submission(
     return property_mappings
 
 
-def _point_coordinate(value):
-    if not value or not isinstance(value, list) or len(value) != 2:
-        return SKIP
-    # Providing the coordinates as [lng, lat] #4955
-    return {"type": "Point", "coordinates": [value[1], value[0]]}
+def _point_coordinate(value: GeoJsonGeometry) -> GeoJsonGeometry:
+    return value
```
sergei-maertens, line 80: obsolete

sergei-maertens (changes requested):
Might be a good idea to disable prettier on JSON files, I had to do the same for markdown and HTML too :wink:

robinmolen:
> Might be a good idea to disable prettier on JSON files, I had to do the same for markdown and HTML too 😉

Ah yeah, probably best to add those to the .prettierignore 🙂 (just to solve it project wide)
I'll make a PR for it

src/openforms/formio/components/custom.py
```diff
@@ -21,6 +21,7 @@
 from openforms.utils.validators import BSNValidator, IBANValidator
 from openforms.validations.service import PluginValidator
 
+from ...api.geojson import GeoJsonGeometryPolymorphicSerializer
```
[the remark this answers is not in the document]
robinmolen: Check, i'll change it. Just wondering, why not just use absolute imports for everything?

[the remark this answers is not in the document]
sergei-maertens: Django apps are supposed to be self-contained and it does happen that we lift out code into a separate library. The imports then usually keep working because you take an entire directory and move it somewhere else. If you use (project-specific) absolute imports, then you have a lot of find-and-replace to do.

That's the official reasons - the unofficial one is that absolute imports and relative imports are different blocks with isort and they make it clear when you're importing stuff from another app vs. importing from the same app, and that makes it easier to check if people respect the convention to not import private API from other apps (those imports should go through `service.py` ideally).

sergei-maertens:
> > Might be a good idea to disable prettier on JSON files, I had to do the same for markdown and HTML too 😉
> 
> Ah yeah, probably best to add those to the .prettierignore 🙂 (just to solve it project wide) I'll make a PR for it

Let's enable prettier on json files inside `src/openforms/js` and disable in other locations.

src/openforms/formio/formatters/custom.py
robinmolen: I thought that these should remain near the `MapFormatter`, seeing how these types are only used for this class.
And also to keep the code uniform to the `AddressNLFormatter` / `AddressValue` implementation.

But sure, i can move them

src/openforms/registrations/contrib/objects_api/submission_registration.py
robinmolen: Yeah these are now obsolete.. Will remove

src/openforms/formio/components/custom.py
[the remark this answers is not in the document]
robinmolen: Okay, check check. I'll keep it in mind

src/openforms/formio/tests/validation/test_map.py
robinmolen: Yeah i thought that this would simplify things for the multi-dimensional errors, but i totally get your point..

I've removed this function and cleaned the tests up

src/openforms/formio/tests/validation/test_map.py
robinmolen: Yeah, this creates a level of obscurity...

I've removed that recursion function and separated the tests a bit. I think/hope this solves/removes the obscurity problems..

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -33,11 +33,12 @@
     create_csv_document,
     create_report_document,
 )
+from openforms.formio.formatters.custom import GeoJsonGeometry
```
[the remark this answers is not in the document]
robinmolen: With the removal of the `_point_coordinate` function, this import was no-longer needed. So this has been removed as well.

src/openforms/formio/formatters/custom.py
```diff
@@ -21,10 +23,16 @@ def format(self, component: Component, value: str) -> str:
         return f"{fmt_date(parsed_value)} {fmt_time(parsed_value, 'H:i')}"
 
 
+type MapValue = PointGeometry | LineStringGeometry | PolygonGeometry
```
sergei-maertens, line 26: this is the ideal compromise - it keeps the stuff that's not specific to map components in a generic place while still refining it sufficiently within the map component context

sergei-maertens (approved):

PR #4956, opened 2024-12-19
Feature/4908 json registration backend

### LINKED ISSUE #4908: Add a registration backend that sends the form/submission variables as JSON to some API endpoint
There is a desire for a "simple" registration backend/plugin which just dumps the submission/form variables with their values as a "flat" JSON structure and sends this data to an API endpoint for further processing. The goal is to keep this super simple and light weight, and let the post-processing be done on the receiving end (with an enterprise service bus or other technologies).

**Requirements**

- [x] A top level object with keys `values` and (optional?) `schema` is sent
- [x] You can specify which service to send the data to (reference to `zgw_consumers.Service`)
- [x] You can configure which endpoint to send the data to
- [x] For each form variable, you can include/exclude it in the data with a simple checkbox in the variables table
- [ ] #4980

**Example**

An example form with the following configuration:

* Authentication via DigiD, capturing a BSN
* First form step with text fields `firstName` and `lastName`
* Second form step with a user upload `attachment`

Should lead to a JSON object being sent with the shape:

```json
{
    "values":
    {
        "auth_bsn": "123456782",
        "firstName": "Donnie",
        "lastName": "Darko",
        "attachment": "<base64 encoded data>"
    },
    "schema":
    {
        "type": "object",
        "properties":
        {
            "auth_bsn":
            {
                "type": "string",
                "pattern": "^\\d{9}$"
            },
            "firstName":
            {
                "type": "string"
            },
            "lastName":
            {
                "type": "string"
            },
            "attachment":
            {
                "type": "string",
                "contentEncoding": "base64"
            }
        },
        "required":
        [
            "auth_bsn",
            "firstName",
            "lastName"
        ],
        "additionalProperties": false
    }
}
```

**Tasks**

- [x] Define new registration plugin - it's probably wise to gate it behind a `if settings.DEBUG` since I don't expect this to be done for 3.0
- [x] Implement plugin configuration options:
    - Service to use
    - Endpoint to send data to (relative to service API root)
    - Form variables (including static variables!) to include
- [x] Add docker-compose mock service (flask app?) to simulate a receiving server
- [x] Implement sending the data to the configured service in the defined format, taking into account the configuration options
- [x] Add configuration checks to configuration overview
- [x] Ensure that we can generate an appropriate json-schema for each form component type
- [ ] Add tests

**Pointers**

Existing code that can be useful:

* service fetch
* email registration plugin
* objects API registration configuration/options

viktorvanwijk:
Closes #4908

**Changes**

Add registration backend plugin that converts submitted form data to JSON and sends it to a specified API endpoint

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/js/components/admin/form_design/registrations/json/fields/FormVariablesSelect.js
sergei-maertens, line 1: This entire file is likely obsoleted by the existing `VariableSelection` component. If I'm wrong, I'll be happy to hear why! :smile:

src/openforms/js/components/admin/form_design/registrations/json/fields/RelativeAPIEndpoint.js
```diff
@@ -0,0 +1,34 @@
...
+          />
+        }
+        helpText={
+          <FormattedMessage
+            description="JSON registration options 'relativeApiEndpoint' helpText"
+            defaultMessage="Relative endpoint to send the data to (will be added to the root endpoint of the service)."
+          />
+        }
+      >
+        <TextInput id="id_relativeApiEndpoint" {...fieldProps} />
```
sergei-maertens, line 28:
```suggestion
        <TextInput {...fieldProps} />
```
I *think* it's automatically managed by the `Field` and `TextInput` components.

src/openforms/js/components/admin/form_design/registrations/json/fields/ServiceSelect.js
```diff
@@ -0,0 +1,42 @@
...
+            defaultMessage="Service"
+          />
+        }
+        helpText={
+          <FormattedMessage
+            description="JSON registration options 'serviceSelect' helpText"
+            defaultMessage="Which service to send the data to"
+          />
+        }
+      >
```
sergei-maertens, line 26:
```suggestion
      <Field
        name="service"
        label={
          <FormattedMessage
            description="JSON registration options 'serviceSelect' label"
            defaultMessage="Service"
          />
        }
        helpText={
          <FormattedMessage
            description="JSON registration options 'serviceSelect' helpText"
            defaultMessage="Which service to send the data to"
          />
        }
        required
        noManageChildProps
      >
```
* label should be bold to indicate it's a required field
* `noManageChildProps` should be the default but we can't do that yet because of all the existing code :grimacing: it avoids some of the `Field` props being manipulated and passed down to the child element (`ReactSelect` in this case)

src/openforms/registrations/contrib/json/apps.py
```diff
@@ -0,0 +1,12 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+# TODO-4908: maybe rename to FVaJ (Form Variables as JSON) or similar?
```
sergei-maertens, line 5: I quite like `JSONDumpAppConfig`

src/openforms/registrations/contrib/json/config.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    relative_api_endpoint = serializers.CharField(
+        max_length=255,
+        label=_("Relative API endpoint"),
+        help_text=_(
+            "The API endpoint to send the data to (relative to the service API root)."
+        ),
+        allow_blank=True,
+        required=False,
+    )
```
sergei-maertens, line 29: I'd name this `path` similar to the `ServiceFetchConfiguration.path` field.

Additionally, we need some input validation here (or at the minimum when making the request), see for example `openforms.variables.validators.validate_path_context_values`. Essentially, if an admin configures an API root of `/foo/bar`, some configuring a form should not be able to perform path traversal to make a call to `/baz` for example.

src/openforms/registrations/contrib/json/config.py
```diff
@@ -0,0 +1,51 @@
...
+    form_variables = serializers.ListField(
+        child=FormioVariableKeyField(max_length=50),
```
sergei-maertens, line 31:
```suggestion
        child=FormioVariableKeyField(),
```
`key` on `FormVariable` has no length limits (it's `models.TextField`), so we can't add such constraints as it would exclude certain variables from being sent.

src/openforms/registrations/contrib/json/config.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    JSON registration plugin options
+
+    This describes the shape of :attr:`JSONOptionsSerializer.validated_data`, after
+    the input data has been cleaned/validated.
+    """
+
+    service: Required[Service]
+    relative_api_endpoint: str
+    form_variables: Required[list[str]]
```
sergei-maertens, line 51:
```suggestion
    service: Service
    relative_api_endpoint: str
    form_variables: list[str]
```
`Required` is the default, unless you specify `total=False` on the `TypedDict`.

If you specify `default=""` on the relative API endpoint field, then you also have the guarantee that the key will always be present, and that gives the most ergonomic type definition (nice and strict).

src/openforms/registrations/contrib/json/models.py
sergei-maertens, line 1: I don't think we need a global configuration at all tbh!

src/openforms/registrations/contrib/json/plugin.py
```diff
@@ -0,0 +1,103 @@
...
+            "type": "object",
+            "properties": {
+                "static_var_1": {"type": "string", "pattern": "^cool_pattern$"},
+                "form_var_1": {"type": "string"},
+                "form_var_2": {"type": "string"},
+                "attachment": {"type": "string", "contentEncoding": "base64"},
+            },
+            "required": ["static_var_1", "form_var_1", "form_var_2"],
+            "additionalProperties": False,
+        }
```
sergei-maertens, line 64: You can grab the static variables from the submission variables state too, and build up what needs to be processed before doing the processing.

```py

all_values: JSONObject = {
    **variables_state.get_static_data(),
    **state.get_data(),  # dynamic values from user input
}
values_to_register = {
    key: value 
    for key, value in all_values.items()
    if key in options["form_variables"]
}
```

For the attachment processing, there's a bug if you have multiple file uploads - now the previous uploads get overwritten. So, depending on if the component is `multiple` True/False, you need to handle an array of uploads or just a single upload. I'd recommend setting up component-specific processing like in the Objects API: https://github.com/open-formulieren/open-forms/blob/a9506354c37e8c04a32104a740f61512fcaa5aa1/src/openforms/registrations/contrib/objects_api/submission_registration.py#L574

src/openforms/registrations/contrib/json/plugin.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Send to the service
+        json = {"values": values, "schema": schema}
+        service = options["service"]
+        submission.registration_result = result = {}
+        with build_client(service) as client:
+            result["api_response"] = res = client.post(
+                options.get("relative_api_endpoint", ""),
```
sergei-maertens, line 72: this requires some run-time validation just to be sure :)

src/openforms/registrations/contrib/json/plugin.py
```diff
@@ -0,0 +1,103 @@
...
+                json=json,
+                headers={"Content-Type": "application/json"},
```
sergei-maertens, line 74: I believe this is done automatically by requests if you specify the `json` kwarg.

src/openforms/registrations/contrib/json/plugin.py
```diff
@@ -0,0 +1,103 @@
...
+    def get_config_actions(self) -> list[tuple[str, str]]:
+        return [
+            (
+                _("Configuration"),
+                reverse(
+                    "admin:registrations_json_jsonconfig_change",
+                    args=(JSONConfig.singleton_instance_id,),
+                ),
+            )
+        ]
```
sergei-maertens, line 103: I can't think of anything meaningful to config check, so it's not relevant for this plugin, I think.

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/RegistrationFields.stories.js
```diff
@@ -977,3 +1014,61 @@ export const STUFZDS = {
...
+      },
+      {
+        dataType: 'file',
+        form: null,
+        formDefinition: null,
+        key: 'attachment',
+        name: 'Attachment',
+        source: 'user_defined',
+      },
+    ],
```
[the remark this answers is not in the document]
viktorvanwijk: Looking at the model  `FormVariable` and serializer `FormVariableSerializer`, it seems like I could actually leave out 'formDefinition' and 'dataFormat' as well, but I will add everything for clarity

src/openforms/js/components/admin/form_design/registrations/json/JSONSummaryHandler.js
```diff
@@ -0,0 +1,37 @@
...
+    );
+  }
+  else {
+    return (
+      <FormattedMessage
+        description="JSON registration summary message"
+        defaultMessage="Not included"
+      />
+    );
+  }
```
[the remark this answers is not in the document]
viktorvanwijk: Storybook complains about STATIC_URL being not defined in `BooleanIcon`. Where should I add this?

src/openforms/js/components/admin/form_design/registrations/json/fields/ServiceSelect.js
viktorvanwijk: Also add both to `FormVariableSelect` then?

src/openforms/js/components/admin/form_design/registrations/json/JSONOptionsForm.js
```diff
@@ -0,0 +1,91 @@
...
+        enumNames: PropTypes.arrayOf(PropTypes.string).isRequired,
+      }).isRequired,
+    }).isRequired,
+  }).isRequired,
+  formData: PropTypes.shape({
+    service: PropTypes.number,
+    relativeApiEndpoint: PropTypes.string,
+    // TODO-4908: might need to rename this to selectedFormVariables to avoid confusion or even
+    //  naming conflicts
+    formVariables: PropTypes.arrayOf(PropTypes.string),
```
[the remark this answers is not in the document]
sergei-maertens: yes! the names in the serializer and data you send in the options have to match :)

src/openforms/registrations/contrib/json/config.py
viktorvanwijk: `validate_path_context_values` only checks if the path equals `'..'`. Wouldn't it be better to do a `startswith('..')`, or isn't that necessary?

src/openforms/registrations/contrib/json/plugin.py
viktorvanwijk: In the current implementation, the attachments are added to a dictionary `{'<original_file_name>': '<base64_content>'}`. I liked the consistency better than in the case of switching between a string and list of strings. I wasn't completely sure about the original file name, though. As far as I understand, they are not unique in the database, but they are unique for a particular file-upload component as you can't upload files with the same name. Any thoughts on this?

sergei-maertens: > but they are unique for a particular file-upload component, as you can't upload files with the same name

I didn't know that is the current behaviour! So it's definitely not a deliberate thing.

viktorvanwijk: After discussion: file data will be sent as an object with `{"file_name": <file_name>, "content": <base64_content>}`. In case of multiple files, the data is a list containing this object for each submitted file.

src/openforms/js/components/admin/form_design/registrations/json_dump/fields/Variables.js
```diff
@@ -0,0 +1,43 @@
...
+        }
+        required
+        noManageChildProps
+      >
+        <VariableSelection
+          {...fieldProps}
+          isMulti
+          required
+          closeMenuOnSelect={false}
+          includeStaticVariables={true}
```
sergei-maertens, line 36:
```suggestion
          includeStaticVariables
```
this is equivalent in JSX - no need to change this now, but for future code this is the idiomatic way to specify true boolean prop values

src/openforms/registrations/contrib/json_dump/config.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    path = serializers.CharField(
+        max_length=255,
+        label=_("Path"),
+        help_text=_("Path relative to the Service API root."),
+        allow_blank=True,
+        required=False,
+        default="",
+        validators=[validate_path],
+    )
```
sergei-maertens, line 41: If I enter a path `../bar` when the service has an api root `example.com/foo`, then there is indeed no path traversal happening which is good, however, when I add a safe path, it's not making the correct API call:

```py
>>> from openforms.registrations.contrib.json_dump.plugin import *
>>> service = Service.objects.get(pk=17)
>>> service.api_root
'https://example.com/api/v2/'
>>> data = {"service": service.pk, "path": "nested/404", "variables": ["today"]}
>>> serializer = JSONDumpOptionsSerializer(data=data)
>>> serializer.is_valid()
True
>>> options = serializer.validated_data
>>> options
{'service': <Service: [ORC (Overige)] Objecten Maykin test>, 'path': 'nested/404', 'variables': ['today']}
>>> plugin.register_submission(submission, options)
  File "<input>", line 1, in <module>
    plugin.register_submission(submission, options)
  File "/home/bbt/code/open-forms/src/openforms/registrations/contrib/json_dump/plugin.py", line 69, in register_submission
    res.raise_for_status()
  File "/home/bbt/.virtualenvs/open-forms/lib/python3.12/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://example.com/api/v2/nested%2F404
```

this is slightly different than service fetch - there we support template expressions for the path, here we don't do that. with service fetch, `{{ some_variable }}` is problematic because it can contain slashes (or literally be `..`) and then they need to be URL quoted, but since we only accept literal values, we shouldn't do URL quoting, and that makes it that any `..` in the `path` should probably be blocked (so the validator can just check for `if ".." in value`).

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@register("json_dump")
+class JSONDumpRegistration(BasePlugin):
+    verbose_name = _("JSON dump registration")
+    configuration_options = JSONDumpOptionsSerializer
+
+    def register_submission(
+        self, submission: Submission, options: JSONDumpOptions
+    ) -> dict:
```
sergei-maertens, line 30: bare dicts are not very useful because you have no idea about the key type or the value type, so at the minimum I would declare this as `dict[str, Any]`. However, we know that the return value must be JSON serializable because it gets stored in a `JSONField`, so you can use `openforms.typing.JSONObject` as type hint which limits the keys to `str` type and the values to primitives & json objects/arrays.

```suggestion
) -> JSONObject:
```

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
+import base64
+
+from django.utils.translation import gettext_lazy as _
+
+from zgw_consumers.client import build_client
+
+from openforms.formio.typing import Component
+from openforms.forms.utils import to_json
```
sergei-maertens, line 8: this is a private utility because it's not exported by `openforms.forms.service`.

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
...
+                "form_var_1": {"type": "string"},
+                "form_var_2": {"type": "string"},
+                "attachment": {"type": "string", "contentEncoding": "base64"},
+            },
+            "required": ["static_var_1", "form_var_1", "form_var_2"],
+            "additionalProperties": False,
+        }
+
+        # Send to the service
+        json = to_json({"values": values, "schema": schema})
```
sergei-maertens, line 62: you can just go for `json.dumps(..., cls=DjangoJSONEncoder)`, DRY is not a problem here

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
...
+        service = options["service"]
+        submission.registration_result = result = {}
```
sergei-maertens, line 64: `submission.registration_result` automatically gets written with the return value of `register_submission`, see https://github.com/open-formulieren/open-forms/blob/794bbdae3c62eb4e4ce429661b402b57f24f4605/src/openforms/registrations/tasks.py#L338

```suggestion
```

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
...
+        with build_client(service) as client:
+            path = validate_path_context_values(options.get("path", ""))
```
sergei-maertens, line 66: so this will need a different sanity check - I don't think the existing validator is really re-usable so fine to just do:

```suggestion
            if ".." in (path := options["path"]):
                raise SuspiciousOperation("Possible path traversal detected")
```

Additionally, the `default = ""` on the `path` serializer field matches the `Options` typedict which guarantees that the `path` key exists in the options, so you don't need to `options.get(...)`, you can access it directly: `options[...]` which simplifies the code a bit and removes some edge cases to handle.

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
...
+
+            res = client.post(path, json=json)
+            res.raise_for_status()
+
+            result["api_response"] = res.json()
```
sergei-maertens, line 71:
```suggestion
            return {"api_response": res.json()}
```

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -0,0 +1,142 @@
...
+            # Note: we check the length of the array instead of making the decision
+            # based on component["multiple"], because the number of attachments can
+            # still be one when the component supports multiple uploads.
+            match len(encoded_attachments):
+                case 0:
+                    values[key] = {}
+                case 1:
+                    values[key] = encoded_attachments[0]
+                case _:
+                    values[key] = encoded_attachments
```
sergei-maertens, line 119: I think this is wrong and the decision needs to be made purely based on `multiple`. Example cases for a key `file` / `files`:

* multiple uploads, none provided: `{"files": []}`
* multiple uploads, one provided: `{"files": [{...}]}`
* multiple uploads, multiple provided: `{"files": {...}, {...}]}`
* single upload, none provided: `{"file": null}`
* single upload, file provided: `{"file": {...}}`

This keeps the data type consistent and you can give guarantees that an object will always have the `file_name` and `content` keys.

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+            _component_data_path="file",
+        )
+
+        json_form_options = dict(
+            service=(ServiceFactory(api_root="http://localhost:80/")),
+            path="json_plugin",
+            variables=["firstName", "file", "auth_bsn"],
+        )
+        json_plugin = JSONDumpRegistration("json_registration_plugin")
+        set_submission_reference(submission)
```
sergei-maertens, line 66: it's better DX to pass the `with_public_registration_reference=True` trait to the `SubmissionFactory` :)

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+            file_name="test_file.txt",
+            original_name="test_file.txt",
+            content_type="application/text",
+            content__data=b"This is example content.",
+            _component_configuration_path="components.2",
+            _component_data_path="file",
+        )
+
+        json_form_options = dict(
+            service=(ServiceFactory(api_root="http://localhost:80/")),
```
sergei-maertens, line 61:
```suggestion
            service=(ServiceFactory.create(api_root="http://localhost:80/")),
```
please always make the `.create` / `.build` explicit so we see if we hit the DB or not. Also, you can create this service in the `setUpTestData` classmethod so it only runs once for the entire test case - the exact service does not influence the behaviour of the test so it's okay to move it in the setup method(s).

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+        submission = SubmissionFactory.from_components(
+            [
+                {"key": "firstName", "type": "textfield"},
+                {"key": "lastName", "type": "textfield"},
+                {"key": "file", "type": "file"},
+            ],
+            completed=True,
+            submitted_data={
+                "firstName": "We Are",
+                "lastName": "Checking",
```
sergei-maertens, line 36: Plan F?

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+from pathlib import Path
+
+from django.core.exceptions import SuspiciousOperation
+from django.test import TestCase
+
+from glom import glom
+from requests import RequestException
+from zgw_consumers.test.factories import ServiceFactory
+
+from openforms.submissions.public_references import set_submission_reference
```
sergei-maertens, line 11: this is not public API and should not be imported outside the submission app :-)

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+    SubmissionFileAttachmentFactory,
+)
+from openforms.utils.tests.vcr import OFVCRMixin
+
+from ..plugin import JSONDumpRegistration
+
+VCR_TEST_FILES = Path(__file__).parent / "files"
+
+
+class JSONDumpBackendTests(OFVCRMixin, TestCase):
```
sergei-maertens, line 23: Can you add a test for a `nested/path` endpoint? You'll have to update the Flask app for that too of course.

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+            submission_step=submission.submissionstep_set.get(),
+            file_name="test_file.txt",
+            original_name="test_file.txt",
+            content_type="application/text",
+            content__data=b"This is example content.",
+            _component_configuration_path="components.2",
+            _component_data_path="file",
+        )
+
+        json_form_options = dict(
```
sergei-maertens, line 60: please don't use the `dict(...)` constructor - instead use literal dicts and add the type annotation for how the option should look like so we type-check them properly.

I'd also like to see the whole `json_dump` app added to the `pyright.pyproject.toml` config file so we type check this code in CI too.

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+                        "attachment": {"type": "string", "contentEncoding": "base64"},
+                    },
+                    "required": ["static_var_1", "form_var_1", "form_var_2"],
+                    "additionalProperties": False,
+                },
+            },
+            "message": "Data received",
+        }
+
+        res = json_plugin.register_submission(submission, json_form_options)
```
sergei-maertens, line 95:
```suggestion
        result = json_plugin.register_submission(submission, json_form_options)
```
let's use proper variable names

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+
+        self.assertEqual(res["api_response"], expected_response)
+
+        with self.subTest("attachment content encoded"):
+            decoded_content = b64decode(
+                glom(res, "api_response.data.values.file.content")
```
sergei-maertens, line 101: please don't use glom if you can get by with simple access - `result["api_response"]["data"]["values"]["file"]["content"]` is much easier to understand the actual data structure.

Glom is great for dynamic/unknown stuff and transforming data structures outside of our control into something workable, but avoid it otherwise and go for predictable and obvious code, especially in tests.

src/openforms/registrations/contrib/json_dump/tests/test_backend.py
```diff
@@ -0,0 +1,238 @@
...
+            variables=["file"],
+        )
+        json_plugin = JSONDumpRegistration("json_registration_plugin")
+        set_submission_reference(submission)
+
+        res = json_plugin.register_submission(submission, json_form_options)
+
+        self.assertEqual(glom(res, "api_response.data.values.file"), {})
+
+    def test_path_traversal_attack(self):
```
sergei-maertens, line 215: you can create some variants here and loop over them, using `self.subTest(...)` E.g. an example with `foo/../bar` or `foo..` should be blocked.

src/openforms/registrations/contrib/json_dump/tests/test_config.py
```diff
@@ -0,0 +1,10 @@
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+
+from ..config import validate_path
+
+
+class JSONDumpConfigTests(TestCase):
+    def test_validate_path(self):
+        with self.assertRaises(ValidationError):
+            validate_path("..")
```
sergei-maertens, line 10: IMO this is too much implementation detail testing - instead test that the serializer blocks suspicious paths as that is our interface with/to the frontend/UI and even the registration tasks use the configuration serializers to create the options dict.

sergei-maertens (changes requested):

src/openforms/registrations/contrib/json_dump/plugin.py
sergei-maertens: for future checking: type checker didn't agree :)

src/openforms/js/components/admin/form_design/registrations/json_dump/fields/Variables.js
```diff
@@ -0,0 +1,43 @@
...
+          />
+        }
+        required
+        noManageChildProps
+      >
+        <VariableSelection
+          {...fieldProps}
+          isMulti
+          required
+          closeMenuOnSelect
```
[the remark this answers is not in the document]
viktorvanwijk: Yeah you're right, removed one too many :grimacing:

sergei-maertens (approved):

Issue #4496, opened 2024-07-05
Improve error message for FileUpload component


vaszig:
When we try to upload an invalid file we get the following error which indicates that the field can not be empty. Instead we should provide a better feedback and mention to the user that the provided file type is not acceptable.

![image](https://github.com/open-formulieren/open-forms/assets/63058071/7660e81e-0d9d-4ace-82c8-bc4cd9922de6)

PR #4961, opened 2024-12-19
.msg file upload

### LINKED ISSUE #4795: .msg files cannot be uploaded

tested on 2.7.x 2.8.x

### Customer reference

Utr 306

Sentry: 313314 

When uploading a .msg file and after **navigating to the next step** the upload component shows a error: 
image

`ValidationError
{'data': {'bestandsupload': {0: {'type': [ErrorDetail(string="Het bestandstype kon niet bepaald worden. Controleer of de bestandsnaam met een extensie eindigt (bijvoorbeel '.pdf' of '.png').", code='blank')]}}}}`

Tested with multiple OSes, and multiple files and environments.

robinmolen:
Closes #4795

**Changes**

The sdk cannot determine what content type should be used for .msg files, which results in the content type `""`. When you try to submit the form, the fileSerializer doesn't recognizes the content type and returns an error.

To allow the submission of .msg files, the content type should be allowed to be empty.

- Turned `type` property on `FileSerializer` optional
- Added validation for .msg files to the `MimeTypeValidator`

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


robinmolen:
The removal of the `FileSerializer` `type` blank message conflicts with #4496. Some more work is required to ensure users receive useful error messages, when needed

robinmolen:
The `MimeTypeValidator` now checks if the file has an extension. If it doesn't, an validation error is raised

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
sergei-maertens, line 1: Can you please exclude backend translations from PRs? They get sorted out during the release preparation.

They're a frequent source of merge conflicts otherwise.

src/openforms/formio/tests/test_validators.py
```diff
@@ -202,6 +202,22 @@ def test_allowed_mime_types_for_csv_files(self):
...
+        validator = validators.MimeTypeValidator(allowed_mime_types=[valid_type])
+
+        # 4795
+        # The sdk cannot determine the content_type for .msg files correctly.
+        # Because .msg is a windows specific file, and linux and MacOS don't know it.
+        # So we simulate the scenario where content_type is unknown
+        sample = SimpleUploadedFile(
+            "test.msg",
+            msg_file.read_bytes(),
+        )
```
sergei-maertens, line 217: This results in the file having `content_type="text/plain"` because it's the default of `SimpleUploadedFile`. You need to explicitly reproduce the case where it's empty as provided by the frontend:

```suggestion
        sample = SimpleUploadedFile(
            "test.msg",
            msg_file.read_bytes(),
            content_type="",
        )
```

src/openforms/formio/components/vanilla.py
```diff
@@ -338,14 +338,7 @@ class FileSerializer(serializers.Serializer):
...
     storage = serializers.ChoiceField(choices=["url"])
-    type = serializers.CharField(
-        error_messages={
-            "blank": _(
-                "Could not determine the file type. Please make sure the file name "
-                "has an extension."
-            ),
-        }
-    )
+    type = serializers.CharField(required=False, allow_blank=True)
```
sergei-maertens, line 341:
```suggestion
    type = serializers.CharField(required=True, allow_blank=True)
```

I would definitely still require the key to be present, but we can accept empty strings and process them during validation indeed.

src/openforms/formio/api/validators.py
```diff
@@ -56,7 +56,8 @@ def __init__(self, allowed_mime_types: Iterable[str] | None = None):
 
     def __call__(self, value: UploadedFile) -> None:
         head = value.read(2048)
-        ext = value.name.split(".")[-1]
+        file_name_parts = value.name.split(".")
```
sergei-maertens, line 59: this way of grabbing the extension is actually quite flawed (it already was): try for example a file name `foo/bar.baz/example.png` :)

Instead, what is more correct to use is `pathlib`:

```py
from pathlib import Path

ext = Path(value.name).suffix[1:]
```
which gives you the actual extension (without dot)

src/openforms/formio/api/validators.py
```diff
@@ -76,6 +77,14 @@ def __call__(self, value: UploadedFile) -> None:
                 _("The provided file is not a valid file type.")
             )
 
+        if len(file_name_parts) == 1:
+            raise serializers.ValidationError(
+                _(
+                    "Could not determine the file type. Please make sure the file name "
+                    "has an extension."
+                )
+            )
```
sergei-maertens, line 86:
```suggestion
        if not ext:
            raise serializers.ValidationError(
                _(
                    "Could not determine the file type. Please make sure the file name "
                    "has an extension."
                )
            )
```
this `file_name_parts` check is really a convoluted way to say "I couldn't determine an extension".

src/openforms/formio/api/validators.py
```diff
@@ -111,6 +120,11 @@ def __call__(self, value: UploadedFile) -> None:
             "image/heif",
         ):
             return
+        # 4795
+        # The sdk cannot determine the file type of .msg files, which result into
+        # content_type "". So we have to validate these for ourselves
+        elif mime_type == "application/vnd.ms-outlook" and ext == "msg":
+            return
```
sergei-maertens, line 127: rather than adding another hardcoded check here, we can do better, but we need to restructure the validator code a bit!

First, start with figuring out the provided content type:

```py
provided_content_type = value.content_type or "application/octet-stream"
```

and replace all `value.content_type` checks with `provided_content_type`.

The `or` clause is important here - in situations where the mime type could not be determined by the browser and it returns an empty string, the backend will fall back to the generic binary data mime type. But, the interesting part is that in the generic binary data check:

```py
if value.content_type == "application/octet-stream": ...
```
the extensions for the binary data are looked up and checked against the provided file extension. So, we invert the process here - given the binary data, what should the extension be, and validate this.

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -966,8 +966,8 @@ def test_unknown_file_type(self):
                 ],
             )
 
-        # Make sure the frontend did not create one:
-        self.assertEqual(TemporaryFileUpload.objects.count(), 1)
+        # Make sure that no temporary files were created
+        self.assertEqual(TemporaryFileUpload.objects.count(), 0)
```
sergei-maertens, line 970: I don't really understand what changed here? Why was there initially a temporary file upload and now not anymore?

sergei-maertens (changes requested):

src/openforms/tests/e2e/test_input_validation.py
sergei-maertens: Figured it out - we didn't have a check for empty file name extensions before and we do now, which rejects the temporary file upload created from the frontend.

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.8.x: 75e9299f6857d17464e525ca8159a08ba08652bd
* 2.7.x: d17b16f5ae55eaed7f10e76ed065aa1cea06bede

PR #4962, opened 2024-12-20
Enable exporting (registered) submission metadata

### LINKED ISSUE #4930: As admin, I want to create an export of all submissions to compare them with registration systems
The goal is to see if Open Forms submitted more or less forms then the number of cases created in backend systems. So, the admin wants to compare the submitted forms to the created cases in another system.

To this end, we'll expand on the existing statistics page, and add an export function.

**Tasks**

1. Add date-range filter to statistics: https://pypi.org/project/django-daterange-filterspec/
2. Add export bulk action: Export statistics
3. Upon export, generate the statistics details based on timeline logger events from succesfull submissions
4. Export the columns: public reference, form name, submission datetime

Note: I know the date in the existing view doesn't match the expectations. It now filters on "last submission" date. In a followup issue, this will be addressed.

sergei-maertens:
Closes #4930

**Changes**

* Added custom admin page and link to it from the form statistics
* Implemented exporting submission statistics based on timeline logs

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/forms/admin/views.py
+++ b/src/openforms/forms/admin/views.py
@@ -1,4 +1,5 @@
 import zipfile
+from datetime import date
 from uuid import uuid4
 
 from django import forms
@@ -8,14 +9,16 @@ from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.contrib.messages.views import SuccessMessageMixin
 from django.contrib.postgres.forms import SimpleArrayField
-from django.http import FileResponse
+from django.http import FileResponse, HttpResponse
 from django.shortcuts import get_object_or_404
 from django.urls import reverse_lazy
 from django.utils.decorators import method_decorator
+from django.utils.http import content_disposition_header
 from django.utils.translation import gettext_lazy as _
 from django.views import View
 from django.views.generic.edit import FormView
 
+from import_export.formats.base_formats import XLSX
 from privates.storages import private_media_storage
 from rest_framework.exceptions import ValidationError
 
@@ -125,6 +128,23 @@ class ExportSubmissionStatisticsView(
 
     # must be set by the ModelAdmin
     media: forms.Media | None = None
+
+    def form_valid(self, form: ExportStatisticsForm) -> HttpResponse:
+        start_date: date = form.cleaned_data["start_date"]
+        end_date: date = form.cleaned_data["end_date"]
+        dataset = form.export()
+        format = XLSX()
+        filename = f"submissions_{start_date.isoformat()}_{end_date.isoformat()}.xlsx"
+        return HttpResponse(
+            format.export_data(dataset),
+            content_type=format.get_content_type(),
+            headers={
+                "Content-Disposition": content_disposition_header(
+                    as_attachment=True,
+                    filename=filename,
+                ),
+            },
+        )
 
     def get_context_data(self, **kwargs):
         assert (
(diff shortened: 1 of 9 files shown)


src/openforms/forms/admin/views.py
```diff
@@ -126,6 +129,23 @@ class ExportSubmissionStatisticsView(
     # must be set by the ModelAdmin
     media: forms.Media | None = None
 
+    def form_valid(self, form: ExportStatisticsForm) -> HttpResponse:
+        start_date: date = form.cleaned_data["start_date"]
+        end_date: date = form.cleaned_data["end_date"]
+        dataset = form.export()
+        format = XLSX()
+        filename = f"submissions_{start_date.isoformat()}_{end_date.isoformat()}.xlsx"
```
robinmolen, line 137: Maybe nice to add the form names, if the limit is used?

src/openforms/forms/forms/form_statistics.py
```diff
@@ -50,3 +52,12 @@ class ExportStatisticsForm(forms.Form):
             "multiple options."
         ),
     )
+
+    def export(self) -> Dataset:
+        start_date: date = self.cleaned_data["start_date"]
+        end_date: date = self.cleaned_data["end_date"]
+        return export_registration_statistics(
```
robinmolen, line 59: Maybe a bit picky, but why do you define the dates as variables? Is it (only) to specify the date typing?
Otherwise, why not just pass the `cleaned_data` directly?

src/openforms/forms/statistics.py
```diff
@@ -34,7 +34,7 @@ def export_registration_statistics(
     )
 
     _start_date = make_aware(datetime.combine(start_date, time.min))
-    _end_date = make_aware(datetime.combine(end_date, time.max))
+    _end_date = make_aware(datetime.combine(end_date, time.min))
```
robinmolen, line 37: Why not till the end of the day?

robinmolen:
I've tried the export, and was missing some submissions.. I'm not yet sure why, i'll take a second look later today..

src/openforms/forms/admin/views.py
sergei-maertens: I won't do that - this will not scale if you pick 10 forms for example.

src/openforms/forms/forms/form_statistics.py
sergei-maertens: A couple of reasons, all summarized as "loose coupling".

1. A simple `export_registration_statistics` with plain parameters is easy to understand when you're in that function, without adding the cognitive overhead of knowing what the form structure looks like and what's in `cleaned_data`.
2. Building on top of 1., I can modify my form implementation and interface without needing to touch the actual export function. If I have to touch the export function, I'm very well aware that I'm changing a Python API and need to check for possible other call sites.
3. The export function can also be called through other user interfaces, e.g. a management command/CLI where the date parameters get passed/parsed differently, or a celery task could be wired up to periodically generate these and automatically prepare/send the exports.
4. Explicit type annotations are nice indeed :)

src/openforms/forms/statistics.py
sergei-maertens: Mostly because of how the export UI is set up - the start date is set as "first of the current month", with the idea that it's likely they'll export this every month. It was easiest to then calculate the end date as first-day-of-next-month, and since that would be included in the next export, the time ranges must work with an open interval rather than closed, otherwise you get duplicated statistics.

Come to think of it, it would probably be better to default the data ranges to the previous month and it might not necessarily be harder to just pick the last day of the month and work with inclusive ranges instead.

PR #4967, opened 2024-12-23
:bug: [#4579] Ensure triggerFromStep uses the correct current_step

### LINKED ISSUE #4579: Block going to next step, also affects previous step?

2.6.13

### Customer reference

Taiga Hoorn 37

Je kunt na een keuze / waarde ervoor zorgen dat je niet verder mag gaan door: blokkeer doorgang naar volgende stap.

Hiervoor moet je aangeven dat dat blokkeren pas ingaat als je in een bepaald stap komt.

Maar als je terug gaat naar de Vorige stap (Uw gegevens) wordt de button daar geblokkeerd. 
Waarschijnlijk omdat je al bij Aanvraag bent geweest. Het werkt blijkbaar niet vanaf Aanvraag

Heb van alles geprobeerd om dit op te heffen, maar het lukt mij niet om dit te de-blokeren.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

stevenbal:
previously the current step was inferred by looking at the last completed step and considering the step after that the current step. This does not work however if you complete a step and go back to the previous step, so instead we explicitly pass the current form step via the serializer context

Closes #4579

**Changes**

* Ensure triggerFromStep uses the correct current_step

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/form_logic.py
+++ b/src/openforms/submissions/form_logic.py
@@ -188,7 +188,9 @@ def evaluate_form_logic(
 
 
 def check_submission_logic(
-    submission: "Submission", unsaved_data: dict | None = None
+    submission: "Submission",
+    unsaved_data: dict | None = None,
+    current_step: "SubmissionStep | None" = None,
 ) -> None:
     if getattr(submission, "_form_logic_evaluated", False):
         return
@@ -198,7 +200,7 @@ def check_submission_logic(
     if not submission_state.form_steps:
         return
 
-    rules = get_rules_to_evaluate(submission)
+    rules = get_rules_to_evaluate(submission, current_step)
 
     # load the data state and all variables
     submission_variables_state = submission.load_submission_value_variables_state()
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -200,7 +200,11 @@ class SubmissionSerializer(serializers.HyperlinkedModelSerializer):
         return super().create(validated_data)
 
     def to_representation(self, instance):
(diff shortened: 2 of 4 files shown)


src/openforms/submissions/form_logic.py
```diff
@@ -198,7 +200,7 @@ def check_submission_logic(
     if not submission_state.form_steps:
         return
 
-    rules = get_rules_to_evaluate(submission)
+    rules = get_rules_to_evaluate(submission, current_step)
```
stevenbal, line 203: I was surprised (and suspicious) about this fix, is there a reason why previously the `current_step` was inferred for this flow, instead of explicitly passed to `get_rules_to_evaluate`?

sergei-maertens: this function exists because the logic for the submission as a whole still needs to be evaluated during registration/processing after it's completed. In those contexts, there isn't a submission step, and it's also not relevant for the "trigger_from_step", because the assumption is that this only gets called for completed submissions that have gone through all the steps anyway.

Clearly this was a wrong assumption - as this does get called in the logic check serialization too which is in the context of a particular step. Nice find!

sergei-maertens (approved):

sergei-maertens:
Backports:

* 2.8.x: 0ad3170452e64517f36a59ab5c154544141a80b8
* 2.7.x: 05f143eec4a6d25bbe08ee40f425239dbdf57209

PR #4996, opened 2025-01-07
:sparkles: [#4993] Implement fetching select(boxes) options from Referentielijsten

stevenbal:
Closes #4993

**Changes**

* Implement fetching select(boxes) options from Referentielijsten
* Add docker-compose file and fixtures for Referentielijsten

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Dockerfile/scripts

  - [x] Updated the Dockerfile with the necessary scripts from the `./bin` folder

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

PR #5007, opened 2025-01-09
Add json schema to submitted data of json dump plugin

### LINKED ISSUE #4980: Add JSON Schema definition of the submitted data to the JSON registration backend plugin
Sub task of #4908

**Example**

An example form with the following configuration:

* Authentication via DigiD, capturing a BSN
* First form step with text fields `firstName` and `lastName`
* Second form step with a user upload `attachment`

Should lead to a JSON object being sent with the shape:

```json
{
    "values":
    {
        "auth_bsn": "123456782",
        "firstName": "Donnie",
        "lastName": "Darko",
        "attachment": "<base64 encoded data>"
    },
    "schema":
    {
        "type": "object",
        "properties":
        {
            "auth_bsn":
            {
                "type": "string",
                "pattern": "^\\d{9}$"
            },
            "firstName":
            {
                "type": "string"
            },
            "lastName":
            {
                "type": "string"
            },
            "attachment":
            {
                "type": "string",
                "contentEncoding": "base64"
            }
        },
        "required":
        [
            "auth_bsn",
            "firstName",
            "lastName"
        ],
        "additionalProperties": false
    }
}
```

viktorvanwijk:
Closes #4980

**Changes**

Added JSON schema definition to the submitted data of the JSON dump registration plugin

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/formio/components/vanilla.py
```diff
@@ -649,6 +751,21 @@ def build_serializer_field(
             **field_kwargs,
         )
 
+    @staticmethod
+    def as_json_schema(component: SelectComponent) -> JSONObject:
+        multiple = component.get("multiple", False)
+        label = component.get("label", "Select")
+
+        choices = [options["value"] for options in component["data"]["values"]]
```
viktorvanwijk, line 759: Currently, if the data source for the `Select`, `SelectBoxes`, and `Radio` components is set to 'Form variable', the `component["data"]["values"]` is not updated. It just contains an a list with one entry `{"value": "", "label": ""}`.

I'm not sure why this is.

sergei-maertens: yeah that's right - the options are fetched and added in `mutate_config_dynamically`, which runs only when there's a submission available.

I think for the JSON Schema generation it's most accurate to use the `enum` information if and only if `component["openForms"]["dataSrc"]` is not set to `"variable"`. So if it *is* set to `variable`, then you just say `{"type": "string"}` and done, if it's not variable, you can add/generate the `enum` and add it as extra constraint.

viktorvanwijk: Thanks :)

viktorvanwijk: Follow-up question. The submission is also available in the plugin. Can we run it there and make all the options available as well? Or is that something we don't want to be doing?

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -95,34 +96,107 @@ def process_variables(submission: Submission, values: JSONObject):
...
+                            # No required properties when there are no attachments
+                        ),
+                        "additionalProperties": False,
+                    }
+
+                    match (
+                        component.get("multiple", False),
+                        n_attachments := len(encoded_attachments),
+                    ):
+                        case False, 0:
```
sergei-maertens, line 133: So you can do much more advanced things with pattern matching, e.g.:

```py
match component:
    case {"type": "file", multiple: False}:
        ...
    case {"type": "file", multiple: True}:
        ...
    case {"openForms": {"dataSrc": "variable"}, "type": "radio" | "selectboxes"}:
        ...
    case {"openForms": {"dataSrc": "variable"}, "type": "select"}:
        ...
```
etc.

See https://github.com/open-formulieren/open-forms/blob/19a3c8af294d26da4c8f2be1f49d06d9083e23e5/bin/report_component_problems.py#L21 for a real example.

It can make the code a lot flatter/easier to follow.

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -95,34 +96,107 @@ def process_variables(submission: Submission, values: JSONObject):
...
+                        case False, int():  # pragma: no cover
+                            raise ValueError(
+                                f"Cannot have multiple attachments ({n_attachments}) "
+                                "for a single file component."
+                            )
+                    values[key] = value
+                    schema["properties"][key] = base_schema
+                case "radio":
+                    data_src = component.get("openForms", {}).get("dataSrc")
+                    if data_src != "variable":
```
sergei-maertens, line 152: it doesn't look like we have constants for this, but please take a look at Steven's PR if he adds some or not, see #4996

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -95,34 +96,107 @@ def process_variables(submission: Submission, values: JSONObject):
...
+                        # Only components where another variable is used as a data
+                        # source need to be processed, so skip this one
+                        continue
+
+                    choices = [options["value"] for options in component["values"]]
```
sergei-maertens, line 157: In these instances I think it's okay to `cast(...)` to the more specific type hint since we already pattern matched it to the right type

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -95,34 +96,107 @@ def process_variables(submission: Submission, values: JSONObject):
...
+                                "properties": properties,
+                                "required": list(properties.keys()),
+                                "additionalProperties": False,
+                            }
+                        )
+
+                    # If the select boxes component is not filled, set required
+                    # properties to empty list
+                    if not values[key]:
+                        schema["properties"][key]["required"] = list()
```
sergei-maertens, line 197: do we not send a value at all?

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -136,13 +210,26 @@ def encode_attachment(attachment: SubmissionFileAttachment) -> str:
...
     config_wrapper = variable.form_variable.form_definition.configuration_wrapper
+
+    # Update components. This is necessary to update the options for Select,
+    # SelectBoxes, and Radio components, which get their options from another form
+    # variable.
+    state = submission.load_submission_value_variables_state()
+    data = DataContainer(state=state)
+    config_wrapper = rewrite_formio_components(
+        config_wrapper, submission=submission, data=data.data
+    )
```
sergei-maertens, line 231: I would expect that we do this only once *before* calling `get_component`, this can potentially be called 10-100s of times and with stevens referentielijsten work there's DB queries and network IO being done to retrieve the options, so let's avoid doing that over and over

src/openforms/registrations/contrib/objects_api/registration_variables.py
```diff
@@ -36,6 +36,10 @@ def get_initial_value(self, submission: Submission | None = None) -> str:
             return ""
         return submission.public_registration_reference
 
+    @staticmethod
+    def as_json_schema():
+        raise NotImplementedError()
```
sergei-maertens, line 41: this is done because the base class defines it as `@abstractmethod` and you *must* implement it, right?

I think we should provide a default implementation instead of crashing, so drop the `@abstractmethod` and instead use the minimal schema generation from `get_json_schema_for_user_defined_variable`

src/openforms/variables/service.py
```diff
@@ -40,3 +46,35 @@ def get_static_variables(
         registered_variable.get_static_variable(submission=submission)
         for registered_variable in variables_registry
     ]
+
+
+# TODO-4980: can be combined with get_json_schema_from_form_variable?
+def get_json_schema_for_user_defined_variable(
```
sergei-maertens, line 52: I'd remove the `user_defined` from this function - it works for any `FormVariable` but it will not be very accurate.

I'd even go for a constant here instead, tbh:

```py
DATA_TYPE_TO_JSON_SCHEMA = {
    FormVariableDataTypes.string: {"type": "string"},
    FormVariableDataTypes.boolean: {"type": "boolean"},
    ...
}
```

it's suitable for use as a fallback rather than crashing.

src/openforms/variables/tests/test_user_defined_variables.py
```diff
@@ -0,0 +1,56 @@
...
+from jsonschema import Draft202012Validator
+
+from ..constants import FormVariableDataTypes
+from ..service import get_json_schema_for_user_defined_variable
+
+
+class UserDefinedVariablesValidJsonSchemaTests(TestCase):
+    validator = Draft202012Validator
+
+    def check_schema(self, properties):
```
sergei-maertens, line 12:
```suggestion
    def assertValidSchema(self, properties):
```
makes the assertions clear, because I had the impression that the tests were not asserting anything and you had forgotten something :)

sergei-maertens (changes requested):
First round, focusing on the architectural aspect and not so much on the actual schemas.

You should probably start from https://github.com/open-formulieren/open-forms/pull/5007#discussion_r1925519013

src/openforms/registrations/contrib/json_dump/plugin.py
viktorvanwijk: 'not filled' is not the best description. In the actual not-filled case, the value is an object with all the options listed as False. Here, I mean in the case it is hidden. Then it's an empty object, so I remove the required list.

src/openforms/variables/service.py
viktorvanwijk:
```python
DATA_TYPE_TO_JSON_SCHEMA = {
    FormVariableDataTypes.string: {"type": "string"},
    FormVariableDataTypes.boolean: {"type": "boolean"},
    ...
}
```
I had this at first, but ran into trouble when assigning additional properties to it such as 'title'. They would get overwritten by the last processed user-defined variable

src/openforms/registrations/contrib/json_dump/plugin.py
viktorvanwijk: At least once for every step then, right? Seems like every step has its own form definition, which in turn has its own config wrapper. Or is there one config wrapper available for everything?

src/openforms/forms/utils.py
```diff
@@ -129,6 +134,104 @@ def form_to_json(form_id: int) -> dict:
...
+
+    # Result
+    var_properties = {**static_var_properties, **form_var_properties}
+    schema = {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": var_properties,
+        "required": required,
+        "additionalProperties": False,
+    }
```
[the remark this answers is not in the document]
viktorvanwijk: Currently, keys with a `.` will also not be added to the `values` in the data that is sent. It tries to look for `foo.bar`, which doesn't exist in the submission data

src/openforms/variables/service.py
sergei-maertens: ah yeah, that's because you're mutating a value then rather than copying it :) `from copy import deepcopy` is your friend! :) Side effects (mutation) are to be avoided, preferably.

src/openforms/registrations/contrib/json_dump/plugin.py
sergei-maertens: sort-of... there's a total_config_wrapper or something like that, and we have a function to run the logic evaluation for the whole form/submission

src/openforms/forms/utils.py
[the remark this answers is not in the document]
sergei-maertens: using `FormioData` as a wrapper around the values should resolve looking up those values :thinking:

src/openforms/forms/utils.py
```diff
@@ -129,6 +134,104 @@ def form_to_json(form_id: int) -> dict:
...
+            schema = get_json_schema_for_user_defined_variable(form_variable.data_type)
+            schema["title"] = form_variable.name
+        case _:  # pragma: no cover
+            raise NotImplementedError("Unexpected form variable source")
+
+    return schema
+
+
+# TODO-4980: add to FormVariable?
+def is_form_variable_required(form_variable: FormVariable) -> bool:
```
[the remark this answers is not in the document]
viktorvanwijk: OK, then I completely misunderstood your [comment](https://github.com/open-formulieren/open-forms/issues/4908#issuecomment-2566460791) before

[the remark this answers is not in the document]
sergei-maertens: No, I just didn't think properly before answering that earlier question, sorry!

src/openforms/registrations/contrib/json_dump/plugin.py
sergei-maertens: You raise some good points 🤔 you probably can add another dict without multiple key to the multiple False match since they're semantically the same, using the pipe symbol

src/openforms/registrations/contrib/json_dump/plugin.py
viktorvanwijk: Doesn't seem like it (for now anyway)

src/openforms/forms/utils.py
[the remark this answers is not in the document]
viktorvanwijk: It doesn't fully solve the problem, unfortunately. Because iterating the `FormioData` with `items()` gives the nested key structure, so it returns `foo` instead of `foo.bar`. This will not pass the `if key in options["variables"]` check, as this contains `foo.bar`. I see three solutions:
* Re-implement `items()` for `FormioData` (not sure if this is the general behaviour we want though)
* Process the keys in `options["values"]` to represent the nested data structure
* Iterate over keys in `options["values"]` instead of the `FormioData.items()`, as the lookup in the other direction does work
Any thoughts on this, or other suggestion(s)?

src/openforms/registrations/contrib/json_dump/plugin.py
viktorvanwijk: Created a new [issue](https://github.com/open-formulieren/open-forms/issues/5044) for this

src/openforms/variables/constants.py
```diff
@@ -54,3 +56,17 @@ class DataMappingTypes(models.TextChoices):
     FormVariableDataTypes.date: check_date,
     FormVariableDataTypes.time: check_time,
 }
+
+
+# TODO-4980: move to openforms.utils.json_schema?
```
sergei-maertens, line 61: Fine to keep it here - it's tightly coupled to the `FormVariableDataTypes` constant so it makes sense to keep that close together.

src/openforms/authentication/static_variables/static_variables.py
```diff
@@ -32,6 +40,13 @@ class LanguageCode(BaseStaticVariable):
     def get_initial_value(self, submission: Submission | None = None) -> str:
         return submission.language_code if submission else ""
 
+    def as_json_schema(self):
+        return {
+            "title": "Language code",
+            "description": "Abbreviation of the used langauge.",
+            "type": "string",
+        }
```
sergei-maertens, line 48: you could add an enum based on `settings.LANGUAGES`

sergei-maertens (changes requested):
Nice work!

src/openforms/formio/components/custom.py
```diff
@@ -109,6 +110,14 @@ def build_serializer_field(
         )
         return serializers.ListField(child=base) if multiple else base
 
+    @staticmethod
+    def as_json_schema(component: DateComponent) -> JSONObject:
+        label = component.get("label", "Date")
+        multiple = component.get("multiple", False)
+
+        base = {"title": label, "format": "date"}
```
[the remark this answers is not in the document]
viktorvanwijk: Probably means my tests aren't good enough then. I was trying to catch these errors with the `check_schema`, but it seems it only checks for incorrect types, not if the type is missing.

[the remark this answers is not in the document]
viktorvanwijk: I'm thinking about revising the json schema validation tests to something like this for each component/static variable:
```python
class TextFieldTests(TestCase):

    @staticmethod
    def assertDataIsValidAgainstSchema(data, schema):
        validate(data, schema)

    def test_data_matches_schema(self):

        for multiple in (False, True):
            component: TextFieldComponent = {
                "label": "Text field",
                "key": "text",
                "type": "textfield",
                "multiple": multiple,
            }
            data = "This is a text."
            if multiple:
                data = [data]
            context = {"submission": SubmissionFactory.build()}

            serializer = build_serializer(
                [component], data={"text": data}, context=context
            )
            schema = as_json_schema(component)

            with self.subTest("serializer validity", multple=multiple):

            with self.subTest("data and schema validation", multiple=multiple):
                self.assertDataIsValidAgainstSchema(data, schema)
```

Any thoughts on this?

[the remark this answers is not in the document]
sergei-maertens: yeah so the thing with these schema's is that they're lax by default rather than strict - not specifying `type` basically means that `anything` goes. So, testing that data matches has the same problem - because anything goes, you're better off testing that wrong data does _not_ pass validation (e.g. putting a number in a text field).

I'm okay with just the `type in schema` assertion though, without giving some dummy data. That creates too much the impression that we use the json schema for validation purposes while in reality we use the serializers for that. The generated JSON schema is informative and not business critical, let's treat it that way too. Though it's probably a good idea to add this to the docstring of `generate_json_schema` - mention it's not used for validation nor can be considered business critical.

src/openforms/formio/service.py
```diff
@@ -105,3 +107,18 @@ def build_serializer(
...
+) -> JSONObject:
+    """Return a JSON schema of a component.
+
+    :param component: Component
+    :param _register: ComponentRegistry | None
+    :returns: JSONObject
+    """
+    registry = _register or register
+
+    component_plugin = registry[component["type"]]
```
sergei-maertens, line 123: this will causes problems at some point, but I'll leave it for now and send you the Sentry link when it happens :-)

sergei-maertens (approved):

src/openforms/formio/service.py
viktorvanwijk: :grimacing:

PR #5010, opened 2025-01-10
[#4825] Fix prefill when authentication is needed

### LINKED ISSUE #4825: Daily digest reporting prefill failures for forms that work with more then 1 auth "flow"

2.7.8 - 2.8.1

### Customer reference

SED 102

Most environments use a common NAW-stap that allows for multiple authentication: DigiD + eHerkenning and anonymous. 

These steps usually consist of 2 or 3 field sets and depending on which form of authentication is used, the right field set will be shown with pre-filled values.

In a situation where a users logs into a form with DigiD a request is made to fetch user data  from the BRP, and not to the KvK to fetch the business data. 

However in the Logs we can see something is going on, but wrongly log a request has been made and the response was empty. These log items also get reported in the daily digest as failures. 

Example with eHerkenning login:
image

Example with DigiD login: 
image

Both submissions show 1 failed and 1 successful pre-fill attempt. But in both causes the failed attempt seems to be "false" (we don't see any outgoing requests, no sentry errors)

As these steps/forms are used on almost all environments, the extra "failures" on the daily digest cause confusion and hinders the visibility of actual issues.

vaszig:
Closes #4825 

**Changes**

- Run prefill plugin only when the authentication type matches the plugin's requirements. 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -9541,11 +9541,12 @@ components:
           type: string
           description: The human-readable name for a plugin.
         requiresAuth:
-          type: string
-          nullable: true
-          title: Required authentication attribute
-          description: The authentication attribute required for this plugin to lookup
-            remote data.
+          type: array
+          items:
+            type: string
+            title: Required authentication attribute
+            description: The authentication attribute required for this plugin to
+              lookup remote data.
         configurationContext:
           nullable: true
           title: Extra configuration context
--- a/src/openforms/js/components/admin/form_design/PluginWarning.js
+++ b/src/openforms/js/components/admin/form_design/PluginWarning.js
@@ -35,7 +35,7 @@ const PluginWarning = ({loginRequired, configuration}) => {
       const authPlugin = availableAuthPlugins.find(plugin => plugin.id === pluginName);
       if (!authPlugin) break;
 
-      if (authPlugin.providesAuth.includes(requiredAuthAttribute)) {
+      if (requiredAuthAttribute.includes(authPlugin.providesAuth)) {
         pluginProvidesAttribute = true;
         break;
       }
@@ -46,12 +46,12 @@ const PluginWarning = ({loginRequired, configuration}) => {
         <FormattedMessage
           description="Prefill plugin requires unavailable auth attribute warning"
           defaultMessage={
-            'Component "{label}" uses a prefill that requires the "{requiredAuthAttribute}" attribute. \
-                        Please select an authentication plugin that provides this attribute.'
+            'Component "{label}" uses a prefill that requires one of the "{requiredAuthAttribute}" attributes. \
+            Please select one or more authentication plugins that provide such an attribute.'
           }
           values={{
             label: configuration.label,
-            requiredAuthAttribute,
+            requiredAuthAttribute: requiredAuthAttribute.join(', '),
           }}
         />
       );
--- a/src/openforms/prefill/api/serializers.py
+++ b/src/openforms/prefill/api/serializers.py
@@ -12,12 +12,13 @@ from openforms.plugins.api.serializers import PluginBaseSerializer
 
 
 class PrefillPluginSerializer(PluginBaseSerializer):
-    requires_auth = serializers.CharField(
-        label=_("Required authentication attribute"),
-        help_text=_(
-            "The authentication attribute required for this plugin to lookup remote data."
+    requires_auth = serializers.ListField(
+        child=serializers.CharField(
+            label=_("Required authentication attribute"),
+            help_text=_(
+                "The authentication attribute required for this plugin to lookup remote data."
+            ),
         ),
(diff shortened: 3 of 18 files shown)


src/openapi.yaml
```diff
@@ -9541,11 +9541,13 @@ components:
...
         requiresAuth:
-          type: string
-          nullable: true
-          title: Required authentication attribute
-          description: The authentication attribute required for this plugin to lookup
-            remote data.
+          type: array
+          items:
+            type: string
+            nullable: true
```
sergei-maertens, line 9547: is this `nullable` right?

src/openapi.yaml
```diff
@@ -9555,7 +9557,6 @@ components:
       - configurationContext
       - id
       - label
-      - requiresAuth
```
sergei-maertens, line 9558: this looks unintentional

src/openforms/js/components/admin/form_design/PluginWarning.js
```diff
@@ -46,12 +46,12 @@ const PluginWarning = ({loginRequired, configuration}) => {
         <FormattedMessage
           description="Prefill plugin requires unavailable auth attribute warning"
           defaultMessage={
-            'Component "{label}" uses a prefill that requires the "{requiredAuthAttribute}" attribute. \
-                        Please select an authentication plugin that provides this attribute.'
+            'Component "{label}" uses a prefill that requires the "{requiredAuthAttribute}" attributes. \
```
sergei-maertens, line 49:
```suggestion
            'Component "{label}" uses a prefill that requires one of the "{requiredAuthAttribute}" attributes. \
```

src/openforms/js/components/admin/form_design/PluginWarning.js
```diff
@@ -46,12 +46,12 @@ const PluginWarning = ({loginRequired, configuration}) => {
...
+                        Please select an authentication plugin that provides these attributes.'
```
sergei-maertens, line 50:
```suggestion
                        Please select one or more authentication plugins that provide such an attribute.'
```

src/openforms/prefill/api/serializers.py
```diff
@@ -12,12 +12,15 @@
...
+    requires_auth = serializers.ListField(
+        child=serializers.CharField(
+            label=_("Required authentication attribute"),
+            help_text=_(
+                "The authentication attribute required for this plugin to lookup remote data."
+            ),
+            allow_null=True,
         ),
-        allow_null=True,
+        default=list,
```
sergei-maertens, line 23: I can't imagine a prefill plugin not needing any auth (except for the demo, but we're changing that behaviour now too), so I'd expect this to be `required=True` without a default, and no `allow_null` for each item.

src/openforms/prefill/fields.py
```diff
@@ -35,7 +35,7 @@ def formfield(self, **kwargs):
     def _get_plugin_choices(self):
         choices = []
         for plugin in register.iter_enabled_plugins():
-            if plugin.requires_auth != self.auth_attribute:
+            if plugin.requires_auth and self.auth_attribute not in plugin.requires_auth:
                 continue
```
sergei-maertens, line 39:
```suggestion
            if self.auth_attribute not in plugin.requires_auth:
                continue
```
simplified - if the collection is empty, the check has the same outcome

src/openforms/prefill/sources.py
```diff
@@ -70,17 +70,22 @@ def invoke_plugin(
...
     invoke_plugin_args = []
     for plugin_id, field_groups in grouped_fields.items():
+        plugin = register[plugin_id]
+
+        # check if we need to run the plugin when the user is authenticated
+        if (required_attribute := plugin.requires_auth) is not None and (
+            (auth_info := getattr(submission, "auth_info", None))
+            and auth_info.attribute not in required_attribute
+        ):
+            continue
```
sergei-maertens, line 85: I think it would be possible to simplify this because `BasePlugin` must define the `requires_auth` attribute and it's an empty collection by default.

```suggestion
        auth_info = getattr(submission, "auth_info", None)
        if auth_info and auth_info.attribute not in plugin.requires_auth:
            continue
```

src/openforms/prefill/sources.py
```diff
@@ -51,10 +52,9 @@ def fetch_prefill_values_from_attribute(
 
     @elasticapm.capture_span(span_type="app.prefill")
     def invoke_plugin(
-        item: tuple[str, IdentifierRoles, list[dict[str, str]]]
+        item: tuple[BasePlugin, IdentifierRoles, list[dict[str, str]]]
```
sergei-maertens, line 55: nice simplification!

Issue #2933, opened 2023-03-24
User rights inconsistent


LaurensBurger:
### Product versie / Product version

2.1.0

### Omschrijf het probleem / Describe the bug

The rights that can be assigned to a user are not as their title describes.
Example:
"forms" + "view" and expect this user to be able to view forms, steps.
![image](https://user-images.githubusercontent.com/112621537/227468722-5a814fae-4f48-4af9-929f-cd8fc40ade94.png)
This user now sees:
No form details
No steps
No variables
No logic
No history

We change the rules to: 
"forms" + "change" and expect this user to be able to edit parts of the form, but not delete or add forms,steps etc.
This user is able to:
Add/remove steps, logic, variables. So delete everything in the form just not the form itself.

### Screen resolution

None

### Device

None

### OS

None

joeribekker:
Refinement: Discussion about the current permission structure. It's too fine grained and doesn't match what users expect. Although they like the granularity, it's probably not needed. Suggested to maybe introduce roles (like the Archiefbeheercomponent). This should also hold the API permissions probably.

joeribekker:
This is not really a bug, it's a misconception but we cant blame the users.

joeribekker:
We should re-evaluate who we give what kind of permissions and how we expose permissions to end-users and the supplier.

joeribekker:
Refinement: Discuss with Utrecht what they want.

PR #5015, opened 2025-01-13
Display form submission statistics based on logs

### LINKED ISSUE #4931: As admin, I want to filter submission statistics based on their timestamp
Followup on #4930

To reflect the number of submissions over a period of time, we need to change the statistics page.

**Tasks**

1. Remove the existing date filters and timeline hierarchy from the statistics admin page.
2. Replace the counters with an aggregated count of the timeline logger events
3. Make sure the date/time filter affects the counts as expected

sergei-maertens:
Closes #4931

**Changes**

* Replaced the model + signals statistics with calculated stats based on log records
* The admin class is customized quite a bit to display the aggregation data
* Moved around the models and templates to replace the previous iteration

I'm a bit concerned about the heavy admin customization again :unamused: 

On top of that - this will lose some historic data if organisations eagerly clean up their submissions - there's a best effort data migration to backfill this data, but it will take some time to get meaningful statistics over time.

:warning: There don't appear to be any default groups that have permissions to view these statistics, @joeribekker - weren't we going to revisit this at some point? See #2933

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/logging/models.py
```diff
@@ -8,9 +9,9 @@
 
 from timeline_logger.models import TimelineLog
 
-from openforms.forms.models import Form
-from openforms.logging.constants import TimelineLogTags
-from openforms.submissions.models import Submission
```
vaszig, line 13: Why did we move this inside the methods?Does it have to do with the models and if they are ready or something else?

vaszig (approved):
- Do we need a test for the data migration?
- I see names in classes like `FormSubmissionStatistics` but the files are named as `form_statistics` for example. Has this been done on purpose?

src/openforms/logging/models.py
sergei-maertens: yeah circular imports, I don't particularly like this :(

sergei-maertens:
I shall add a test for the data migration! I forgor

src/openforms/logging/models.py
vaszig: I guess we cannot avoid it sometimes

sergei-maertens:
> I shall add a test for the data migration! I forgor

Attempted this but it's a world of pain - the database can't be flushed because a table is supposed to be deleted but that goes in a later migration, GFK's make everything harder and swapping the order of migrations leads to django/testing tools not being able to resolve the model state in the migration graph :/

I ran it locally and it produced the results I expected.

src/openforms/forms/templates/admin/forms/formsubmissionstatistics/search_form.html
```diff
@@ -0,0 +1,21 @@
+{% load i18n static %}
```
vaszig, line 1: What is the purpose of overriding this?

vaszig (approved):

src/openforms/forms/templates/admin/forms/formsubmissionstatistics/search_form.html
sergei-maertens: Removing the item result count, as it shows the wrong number. See commit message https://github.com/open-formulieren/open-forms/pull/5015/commits/719dbb4697bcc38ace68c8cca3f1e717d006fcf3 :)

PR #5030, opened 2025-01-22
Showing error messages in variable mapping

### LINKED ISSUE #4871: Display error messages for VariablesMapping (DMN/Objects API prefill)
From https://github.com/open-formulieren/open-forms/pull/4696#discussion_r1863194308

The `VariableMapping` component currently doesn't show error messages that are passed to it (both for DMN configuration and Objects API prefill

robinmolen:
Closes #4871

**Changes**

Showing error messages in the variable mapping for Objects API and DMN. 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/forms/VariableMapping.js
+++ b/src/openforms/js/components/admin/forms/VariableMapping.js
@@ -1,12 +1,13 @@
 import {FieldArray, useFormikContext} from 'formik';
 import get from 'lodash/get';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, {useContext} from 'react';
 import {FormattedMessage, useIntl} from 'react-intl';
 
 import ButtonContainer from 'components/admin/forms/ButtonContainer';
 import Field from 'components/admin/forms/Field';
 import Select from 'components/admin/forms/Select';
+import {ValidationErrorContext} from 'components/admin/forms/ValidationErrors';
 import VariableSelection from 'components/admin/forms/VariableSelection';
 import {DeleteIcon, WarningIcon} from 'components/admin/icons';
 
@@ -47,6 +48,7 @@ const VariableMappingRow = ({
   includeStaticVariables = false,
   alreadyMapped = [],
   rowCheck = undefined,
+  errors = undefined,
 }) => {
   const intl = useIntl();
   const {getFieldProps, setFieldValue} = useFormikContext();
@@ -63,12 +65,14 @@ const VariableMappingRow = ({
   );
 
   const mapping = getFieldProps(prefix).value;
-  const errors = rowCheck?.(intl, mapping).join(', ') ?? '';
+  const rowErrors = rowCheck?.(intl, mapping).join(', ') ?? '';
+  const variableError = errors?.[variableName];
+  const propertyError = errors?.[propertyName];
 
   return (
     <tr>
       <td>
-        <Field name={fullVariableName}>
+        <Field name={fullVariableName} errors={variableError}>
           <VariableSelection
             includeStaticVariables={includeStaticVariables}
             {...getFieldProps(fullVariableName)}
@@ -83,7 +87,7 @@ const VariableMappingRow = ({
       {directionIcon && <td className="mapping-table__direction-icon">{directionIcon}</td>}
 
       <td>
-        <Field name={fullPropertyName}>
+        <Field name={fullPropertyName} errors={propertyError}>
           <Select
             allowBlank
             disabled={loading}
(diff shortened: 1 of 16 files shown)


src/openforms/forms/tests/e2e_tests/test_logic_tab.py
```diff
@@ -86,3 +86,79 @@ def assertState():
...
+                await expect(
+                    node_with_error.get_by_text("This field is required")
+                ).to_be_visible()
+                await expect(
+                    node_with_error.get_by_text(
+                        "DMN configuration:(not configured yet)"
+                    )
+                ).to_be_visible()
+
+            await assertState()
```
sergei-maertens, line 164:
```suggestion

            logic_tab = page.get_by_role("tab", name="Logic")

            await expect(
                logic_tab.get_by_title("There are validation errors")
            ).to_be_visible()

            await logic_tab.click()

            # Logic shows error for DMN config
            node_with_error = page.locator("css=.dsl-editor__node--errors")
            await expect(
                node_with_error.get_by_text("This field is required")
            ).to_be_visible()
            await expect(
                node_with_error.get_by_text(
                    "DMN configuration:(not configured yet)"
                )
            ).to_be_visible()
```
the `assertState` workaround is to be able to make DB queries in an async context - if you're not doing that, you can just inline the rest of the test code

src/openforms/forms/tests/e2e_tests/test_logic_tab.py
```diff
@@ -86,3 +86,79 @@ def assertState():
                 )
 
             await assertState()
+
+    @skip_on_webtest
+    async def test_evaluate_dmn_missing_configuration_in_logic(self):
```
sergei-maertens, line 91: can we do this with a storybook story instead of an e2e tests, since the former are way cheaper to run (locally and in CI)?

src/openforms/js/components/admin/forms/VariableMapping.js
```diff
@@ -47,6 +48,7 @@ const VariableMappingRow = ({
   includeStaticVariables = false,
   alreadyMapped = [],
   rowCheck = undefined,
+  errors = undefined,
```
sergei-maertens, line 51: I reckon the table layout of the fields (instead of the standard field rendering) makes it not easy to grab the error from context instead?

sergei-maertens (changes requested):
I don't think this is the most elegant code but it would be hypocritical to expect that, given how we work with validation errors in this codebase. A thorough cleanup/rewrite is in order for some time - when we build a dedicated admin interface instead of lumping everything in the django admin.

src/openforms/js/components/admin/forms/VariableMapping.js
robinmolen: Yeah this is needed because the TODO in `prepareErrors`. The errors in this case are a multidimensional object, which currently aren't handled in the `prepareErrors` function. I made an attempt to add this functionality, but couldn't find a quick/nice solution.

Also i was worried that adding support for multidimensional objects in the `prepareErrors` would have consequences for other components using the `ValidationErrorContext`.

This seemed the simplest and cleanest solution

src/openforms/forms/tests/e2e_tests/test_logic_tab.py
robinmolen: Mhh i tried it with storybook and had some difficulties getting the validation to trigger..

But i'll give it a second look

src/openforms/forms/tests/e2e_tests/test_logic_tab.py
robinmolen: Check. Yeah if that storybook test approach is a bust, i'll remove that `assertState` function call

src/openforms/forms/tests/e2e_tests/test_logic_tab.py
sergei-maertens: ah right yeah, it needs the API calls. too much hassle, let's stick to the e2e test.

robinmolen: I'm seeing your comment too late...

I've replaced the e2e test with a simple storybook test that displays the correct placement of the errors. This storybook test cannot handle real validation (because, indeed API calls) so the errors are defined in the story configuration.

sergei-maertens: I'm happy about that!

src/openforms/forms/tests/e2e_tests/test_logic_tab.py
robinmolen: Update: I've removed this e2e test and added a storybook test which validates that DMN configuration errors are shown in the logic view.

sergei-maertens (approved):

PR #5078, opened 2025-02-04
[#5077] Improve performance in loading logic rules in admin

### LINKED ISSUE #5077: Improve performance of loading logic rules in admin

vaszig:
Closes #5077

**Changes**

- Logic rules list in the admin was really slow as the logic rules increase because of the same queries happening for the form and the form step. Some joins were added in order to reduce the amount of the queries.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/forms/api/viewsets.py
+++ b/src/openforms/forms/api/viewsets.py
@@ -531,7 +531,7 @@ class FormViewSet(viewsets.ModelViewSet):
         form = self.get_object()
         filterset = FormVariableFilter(
             request.GET,
-            queryset=form.formvariable_set.select_related("form", "form_definition"),
+            queryset=form.formvariable_set.prefetch_related("form", "form_definition"),
         )
 
         serializer = FormVariableSerializer(
@@ -590,7 +590,10 @@ class FormViewSet(viewsets.ModelViewSet):
     @logic_rules_bulk_update.mapping.get
     def logic_rules_list(self, request, *args, **kwargs):
         form = self.get_object()
-        logic_rules = form.formlogic_set.all()
+
+        logic_rules = form.formlogic_set.prefetch_related(
+            "form", "trigger_from_step__form"
+        )
 
         serializer = FormLogicSerializer(
             instance=logic_rules,


vaszig:
**Before**:

logic-rules-performance-old

**After**:

logic-rules-performance-new

src/openforms/forms/api/viewsets.py
```diff
@@ -590,7 +590,12 @@ def logic_rules_bulk_update(self, request, *args, **kwargs):
     @logic_rules_bulk_update.mapping.get
     def logic_rules_list(self, request, *args, **kwargs):
         form = self.get_object()
-        logic_rules = form.formlogic_set.all()
+
+        # Had to apply some joins here to avoid useless queries and improve performance
+        # See github issue https://github.com/open-formulieren/open-forms/issues/5077
+        logic_rules = form.formlogic_set.select_related(
+            "form", "trigger_from_step__form"
+        )
```
sergei-maertens, line 598: have you tried `prefetch_related` too instead of `select_related`?

Select related performs an inner join and results in more data going over the wire between database and application, especially here because the joined form will always be same as the form we already retrieved (meaning that 55 logic rules result in 55 times the same form being joined).

For the trigger_from_step, I expect a similar situation - it's more likely that multiple different logic rules have the same trigger_from_step, so joining can be slower than doing a prefetch. And of course, the related `form` of the `trigger_from_step` has the same problem as the `form` FK of the logic rule.

I think that all three of these fields (form, trigger_from_step and trigger_from_step__form) might benefit from a prefetch - it will result in a couple *more* queries, but it will be a fixed amount and the data going over the wire could be a lot less, so I'm interested in the performance profile of that approach.

sergei-maertens (commented):
I think the variables bulk/list endpoint may have similar issues, iirc that one also showed up in APM. You can give it the same treatment (notably the `FormVariable.form` and `FormVariable.form_definition` foreign keys may cause n+1 queries here)

src/openforms/forms/api/viewsets.py
vaszig: I didn't notice any difference concerning the update endpoints for both logic-rules and variables (tried with both joins and prefetch)

sergei-maertens: You specifically mention the update endpoints, but I'm interested in the list endpoints in particular :P

src/openforms/forms/api/viewsets.py
```diff
@@ -590,7 +590,12 @@ def logic_rules_bulk_update(self, request, *args, **kwargs):
     @logic_rules_bulk_update.mapping.get
     def logic_rules_list(self, request, *args, **kwargs):
         form = self.get_object()
-        logic_rules = form.formlogic_set.all()
+
+        # Had to apply prefetch here as too many queries (n+1) were taking place
+        # See github issue https://github.com/open-formulieren/open-forms/issues/5077
```
sergei-maertens, line 595: Can you put this in the commit message rather than as a comment? This is a clear "why" of the change :)

src/openforms/forms/api/viewsets.py
vaszig: > I think the variables bulk/list endpoint may have similar issues, iirc that one also showed up in APM. You can give it the same treatment (notably the `FormVariable.form` and `FormVariable.form_definition` foreign keys may cause n+1 queries here)

This is why I mentioned it.I understood that we wanted to see what's going on with these as well. But clearly wrong place to write this, it should be as a reply to the above.

PR #5079, opened 2025-02-04
Refactor the submission completion validation

### LINKED ISSUE #4510: it's possible to navigate to the next step under certain conditions, causing validation errors on submit.

2.6.x - 2.7.x

### Customer reference

SED 99
HadM 110
DH (lau)

In certain forms that take a fair amount of time to load/check_logic it is possible for a brief moment to navigate to the next step, before required fields are visible. 
This causes a "call did not validate on the backend" error when submitting. 
image

The issue is reproducible while the check_logic has not yet returned a response, the user is allowed to navigate to the next step. This means if the check_logic returned fields that are required, the user has essentially skipped them. 

The scenario where it's easiest to replicate this is a large form, and a selectboxes component with many answers. each answer should make a field set visible with required fields through back-end logic.  

When the user selects a answer and clicks "Next" right away - the form will navigate to the next step, but the field set will be taken into consideration when submitting. 

Although it's somewhat reproducable - it seems dependant on both the form but also network conditions. So perhaps a solution for this could be a clear indication on what answers still need to be completed - instead of a generic error. 
image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Partly closes #4510 - this is necessary to be able to display proper error messages in the SDK.

The completion validation + communicating the status URL back to the frontend was a weird and unconventional construct. The API documentation was also outdated with what is actually being returned.

This has now been reworked and stabilized to ensure a consistent format of validation errors that can be picked up and processed by the frontend/SDK reliably.

**Changes**

* Reworked how we run final validation of a submission before completing it and processing it
* There's now a properly documented validation error structure for steps
* The statement checkboxes validation remains as is
* The in- and output serializers are now correctly documented
* If the form is configured to not allow submissions, this now results in a 403 instead of a validation error (resolved a TODO)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/api/validation.py
+++ b/src/openforms/submissions/api/validation.py
@@ -1,69 +1,120 @@
 """
 Perform submission-level validation.
-
-TODO: refactor/rework the entire way we _run_ the validations and communicate them back
-to the frontend.
 """
 
-from django.core.validators import MaxLengthValidator
+from typing import TYPE_CHECKING, TypedDict
+
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
+from rest_framework.exceptions import ErrorDetail
+from rest_framework.request import Request
+from rest_framework.settings import api_settings
 
-from openforms.api.utils import mark_experimental
 from openforms.formio.service import build_serializer, get_dynamic_configuration
+from openforms.forms.models import FormDefinition, FormStep
 
-from ..form_logic import check_submission_logic
 from ..models import Submission, SubmissionStep
-from .fields import (
-    PrivacyPolicyAcceptedField,
-    SubmissionAllowedField,
-    TruthDeclarationAcceptedField,
-)
+from .fields import PrivacyPolicyAcceptedField, TruthDeclarationAcceptedField
+
+NON_FIELD_ERRORS_KEY = api_settings.NON_FIELD_ERRORS_KEY
 
 
-@mark_experimental
-class CompletionValidationSerializer(serializers.Serializer):
-    incomplete_steps = serializers.ListField(
-        child=serializers.CharField(),
-        validators=[
-            MaxLengthValidator(
-                limit_value=0,
-                message=_("Not all applicable steps have been completed: %(value)s"),
-            )
-        ],
+def is_step_unexpectedly_incomplete(submission_step: SubmissionStep) -> bool:
+    if not submission_step.completed and submission_step.is_applicable:
+        return True
+    return False
+
+
+class SubmissionCompletionSerializerContext(TypedDict):
+    request: Request
+    submission: Submission
+
+
+type StepValidationErrors = dict[str, ErrorDetail] | dict[str, StepValidationErrors]
+
+
+class SubmissionCompletionSerializer(serializers.Serializer):
+    """
+    Validate the submission completion.
+
+    The input validation requires the statement checkboxes to be provided before
+    completion can be considered. If that is fine, we run validation on the state of
+    the submission, which is a bit "weird" in the sense that we're double-checking
+    earlier user-input and not directly validating ``request.data`` items.
+
+    If all is well, we return the status URL to check the background processing,
+    otherwise the validation errors are raised for the frontend to handle.
+    """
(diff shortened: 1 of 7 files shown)


src/openforms/submissions/api/validation.py
```diff
@@ -5,65 +5,119 @@
...
+    if TYPE_CHECKING:
+
+        @property
+        def context(  # pyright: ignore[reportIncompatibleMethodOverride]
+            self,
+        ) -> SubmissionCompletionSerializerContext: ...
+
+    # statement checkboxes on the overview page
+    privacy_policy_accepted = PrivacyPolicyAcceptedField(required=False)
+    statement_of_truth_accepted = TruthDeclarationAcceptedField(required=False)
```
vaszig, line 63: Why were these changed to required=False?A bit confused with these checkboxes.

src/openforms/submissions/api/validation.py
```diff
@@ -5,65 +5,119 @@
 to the frontend.
```
vaszig, line 5: I guess this TODO cannot be removed right?Your current work is not entirely solving the problem?

src/openforms/submissions/api/viewsets.py
```diff
@@ -297,25 +296,61 @@ def _complete(self, request, *args, **kwargs):
...
         """
         submission = self.get_object()
+        if submission.form.submission_allowed != SubmissionAllowedChoices.yes:
+            raise PermissionDenied(
+                detail=_("Submission is not enabled for this form."),
+                code="submission-not-allowed",
+            )
+
+        # prepare the submission with evaluated logic so that its state can be validated
+        submission.load_execution_state()
```
vaszig, line 329: Do we need this here?The state is loaded again in the `check_submission_logic` (checking because I see a lot of loops inside the method so maybe we could avoid doing this twice)

src/openforms/submissions/api/viewsets.py
```diff
@@ -297,25 +296,61 @@ def _complete(self, request, *args, **kwargs):
...
 
-        The schema of the validation errors response is currently marked as
-        experimental. See our versioning policy in the developer documentation for
-        what this means.
         ---
         """
         submission = self.get_object()
+        if submission.form.submission_allowed != SubmissionAllowedChoices.yes:
+            raise PermissionDenied(
+                detail=_("Submission is not enabled for this form."),
```
vaszig, line 324: Why did we change the error?I think the old one `Submission of this form is not allowed` makes more sense since the checkbox in the admin declares the same (not required, just an opinion).

src/openforms/submissions/api/validation.py
sergei-maertens: before, they were required but it doesn't cause any problems because we were manually setting that data to a default (not very different from what we do know in the view), but setting them to required reflects better in the API docs - it doesn't make sense to mark the field as required if the checkboxes are never displayed in the first place.

src/openforms/submissions/api/validation.py
sergei-maertens: removed!

src/openforms/submissions/api/viewsets.py
sergei-maertens: I prefer to make it explicit - loading the state is idempotent, if it's already cached, it will not load it again :)

src/openforms/submissions/api/viewsets.py
sergei-maertens: The biggest problem was that the original problem resulted in an HTTP 400 with validation errors, but this is not something the person filling out the form can correct. IMO this error message makes it more clear that the user cannot correct the problem.

src/openforms/submissions/api/viewsets.py
vaszig: Perfect, missed that it is cached

vaszig (approved):

PR #5083, opened 2025-02-05
:rewind: Revert removal of DISABLE_SENDING_HIDDEN_FIELDS feature flag

sergei-maertens:
Reverts #4759.

DH is facing challenges with legacy forms while upgrading to 3.0 on a strict timeline. Because the forms haven't been migrated to variables mapping yet, this has a huge impact on downstream systems processing the submissions.


sergei-maertens:
I don't know if we have to put this back in master too or not :/

PR #5085, opened 2025-02-05
:ambulance: Fix performance form definition update

### LINKED ISSUE #5084: Saving a (re-usable) form definition with many components is unreasonably slow

3.0.2 / 2.8.4 / latest

### Customer reference

DB test env

If you have a form with one or more form definitions with many components (observed 95 components), the PUT call in the API to update this is very slow (10s of seconds to minutes), which makes editing forms or form definitions a nightmare.

_No response_

It's fast ⚡ .

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #5084

**Changes**

* Fixed the time spent in Python figuring out which variables to create/update/delete
* Optimized the UPSERT query to avoid flooding it with data we can already ignore in Python

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/forms/models/form_variable.py
+++ b/src/openforms/forms/models/form_variable.py
@@ -1,16 +1,16 @@
 from __future__ import annotations
 
-from copy import deepcopy
-from typing import TYPE_CHECKING, ClassVar
+import itertools
+from copy import copy, deepcopy
+from typing import ClassVar
 
 from django.db import models, transaction
 from django.db.models import CheckConstraint, Q
 from django.utils.translation import gettext_lazy as _
 
-from glom import glom
+import elasticapm
 
 from openforms.formio.utils import (
-    component_in_editgrid,
     get_component_datatype,
     get_component_default_value,
     is_layout_component,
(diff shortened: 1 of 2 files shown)


src/openforms/forms/models/form_variable.py
```diff
@@ -26,29 +26,42 @@
...
+        form_steps = form.formstep_set.select_related(  # pyright: ignore[reportAttributeAccessIssue]
+            "form_definition"
+        )
 
         for form_step in form_steps:
             self.synchronize_for(form_step.form_definition)
 
-    @transaction.atomic
+    @elasticapm.capture_span(span_type="app.core.models")
+    @transaction.atomic(savepoint=False)
```
vaszig, line 64: Was this changed for performance reasons?To avoid savepoints in between?

src/openforms/forms/tests/variables/test_model.py
```diff
@@ -269,3 +293,303 @@ def test_user_defined_variables_cast_initial_value(self):
...
+        )
+        form = FormFactory.create()
+        FormStep.objects.create(form=form, form_definition=fd, order=0, slug="slak")
+        assert not FormVariable.objects.exists()
+
+        with (
+            self.subTest("initial inserts"),
+            # 5 seconds is giving *a lot* of credit, but before the performance patch
+            # this was taking 10+ seconds. In isolation, this takes 100-400ms.
+            assert_execution_time(self, seconds=5),
```
vaszig, line 329: I have not used this kind of tests and I guess sometime will maybe fail right?I mean concerning the time we specify. And maybe this is something deliberate since we can understand that there is a new problem?

vaszig (approved):

src/openforms/forms/tests/variables/test_model.py
sergei-maertens: yes, it's a risk for flakiness, so I'm making sure I'm using different orders of magnitude. i.e. usually this completes in about 0.5s, so I'm taking an upper limit of 10 times that to account for differences in hardware/CPU speed etc.

The performance was so bad that before the patch 10+s was spent, which fit that (terrible) order of magnitude.

I'm also investigating codspeed.io to not have to rely on these kind of flaky constructs, but instead have a dedicated tool in our CI pipeline to catch performance regressions.

sergei-maertens:
Backports:

* stable/3.0.x: 0419a27e7d606b911580c16f41839d1d81ad9d13
* stable/2.8.x: 2fb7263a057a361036fa94a1f36b2037b7d8e4e5

PR #5091, opened 2025-02-07
Fix bug with collapsed metadata fieldset being hidden in JSON dump plugin

### LINKED ISSUE #5088: Metadata fieldset in JSON registration options is not visible when lots of variables are selected
When no variables are selected:

Image

When lots of variables are selected:

Image

Also, might be a good idea to make the variable selection component wider, to account for the number (and also name length) of the variables

viktorvanwijk:
Closes #5088 

**Changes**

Ensure `overflow: visible` is set for the metadata variables fieldset in the JSON Dump plugin

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js
+++ b/src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js
@@ -65,6 +65,7 @@ const JSONDumpOptionsForm = ({name, label, schema, formData, onChange}) => {
               defaultMessage="Metadata variables"
             />
           }
+          extraClassName="openforms-fieldset"
           collapsible
           initialCollapsed
         >
--- a/src/openforms/scss/components/admin/_fieldset.scss
+++ b/src/openforms/scss/components/admin/_fieldset.scss
@@ -1,0 +1,8 @@
+/**
+ * The django admin `.collapsed` styling on fieldset elements sets `overflow: hidden;`.
+ * Seeing how we already hide the fieldset content in the javascript code,
+ * the `overflow: hidden;` only hides the fieldset toggle (which we want to keep visible)
+ */
+.openforms-fieldset.collapsed {
+  overflow: visible !important;
+}
--- a/src/openforms/scss/components/admin/_index.scss
+++ b/src/openforms/scss/components/admin/_index.scss
@@ -19,6 +19,7 @@
 @import './column-field-value';
 @import './confirmation-email-template';
 @import './variablemapping';
+@import './fieldset';
 
 // Form design UI
 @import './select';


src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js
robinmolen, line 1: I'm a bit conflicted, seeing how this is a different way of organizing the plugin configuration (the tabs are currently only used to move/group really different configuration).

I'm leaning towards resetting this to the way it was, and to fix the actual bug (the bootstrap styling).

robinmolen: Just to be completely save (overwriting django styling can be tricky/have unexpected consequences), let's add this fix as a `openforms` custom styling. So i would add a `_fieldset.scss` file with some styling like:

```
/**
 * The django admin `.collapsed` styling on fieldset elements sets `overflow: hidden;`.
 * Seeing how we already hide the fieldset content in the javascript code,
 * the `overflow: hidden;` only hides the fieldset toggle (which we want to keep visible)
 */
.openforms-fieldset.collapsed {
  overflow: auto !important;
}
```

Then add the `openforms-fieldset` class to the Fieldset component, and everything should be fine :)

This would solve the problem of this bug, and also prevent other fieldsets from breaking.

robinmolen (changes requested):

src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js
viktorvanwijk: > I'm leaning towards resetting this to the way it was, and to fix the actual bug (the bootstrap styling).

Yeah I fully agree. That's what I tried at first but couldn't get it to work, so this solution was just a workaround anyway. I had implemented your suggestion at first, but didn't have the `!important` tag, so it wasn't overriding anything. (S)CSS is not my greatest strength :see_no_evil:

src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js
```diff
@@ -65,6 +65,7 @@ const JSONDumpOptionsForm = ({name, label, schema, formData, onChange}) => {
               defaultMessage="Metadata variables"
             />
           }
+          extraClassName="openforms-fieldset"
```
robinmolen, line 68: I was thinking about adding this styling to the component directly (i.e. set it as one of the default Fieldset classes). But this works as well 👍

robinmolen (approved):

src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js
viktorvanwijk: Yeah I figured it's the best to tread carefully here, as I haven't investigated ANY other fieldsets to which this change would also apply :grimacing:

PR #5092, opened 2025-02-07
[#4518] Add log entries for prefill to Submission

### LINKED ISSUE #4518: Add log entries for prefill plugins to submissions
### Thema / Theme

Admin

When investigating something wrong/missing from a submission it would be nice to see the status of the prefill plugin results and if the cause of why i'm looking at the submission was in fact the prefill plugin failing to retrieve data. 

Currently the prefill plugins are logged in the timeline log entries as:

`
[[2024-07-11 11:19:23 CEST] (Submission 123): Prefill plugin "StUF-BG" (stufbg) reported: Successfully retrieved information to prefill fields: BSN (bsn), Geslacht (geslachtsaanduiding), Straatnaam (straatnaam), Voornamen (voornamen), Huisnummer (huisnummer), Voorvoegsels (voorvoegselGeslachtsnaam), Huisletter (huisletter), Achternaam (geslachtsnaam), Toevoeging (huisnummertoevoeging), Postcode (postcode), Woonplaats (woonplaatsNaam), Geboortedatum (geboortedatum)]
`
or 
`
[[2024-07-10 15:28:44 CEST] (Submission 123): Prefill plugin "KvK Bedrijf via KvK-nummer" (kvk-kvknumber) returned empty values]
`

if these were also visible in the submission log, it would save them a lot of time (since the log entries search is very slow and not very accurate)

_No response_

_No response_

vaszig:
Closes #4518

**Changes**

Some AVG logs are needed in the submission admin inline logging so decided to return a queryset based on the included tags and not the other way around. This let us have events with multiple tags and choose which ones we need according to the context.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/logging/logevent.py
+++ b/src/openforms/logging/logevent.py
@@ -111,14 +111,22 @@ def disabling_analytics_tool(
 
 
 def submission_start(submission: Submission):
-    _create_log(submission, "submission_start")
+    _create_log(
+        submission,
+        "submission_start",
+        tags=[TimelineLogTags.submission_lifecycle],
+    )
 
 
 def submission_auth(
     submission: Submission, delegated: bool = False, user: User | None = None
 ):
     _create_log(
-        submission, "submission_auth", user=user, extra_data={"delegated": delegated}
+        submission,
+        "submission_auth",
+        user=user,
+        extra_data={"delegated": delegated},
+        tags=[TimelineLogTags.submission_lifecycle],
     )
 
 
@@ -133,6 +141,7 @@ def submission_step_fill(step: SubmissionStep):
             "step": str(step.form_step.form_definition.name),
             "step_id": step.pk,
         },
+        tags=[TimelineLogTags.submission_lifecycle],
     )
 
 
@@ -155,6 +164,7 @@ def form_submit_success(submission: Submission):
         submission,
         FORM_SUBMIT_SUCCESS_EVENT,
         extra_data=_snapshot_submission_statistics(submission),
+        tags=[TimelineLogTags.submission_lifecycle],
     )
 
 
@@ -162,7 +172,11 @@ def form_submit_success(submission: Submission):
 
 
 def pdf_generation_start(submission: Submission):
-    _create_log(submission, "pdf_generation_start")
+    _create_log(
+        submission,
+        "pdf_generation_start",
+        tags=[TimelineLogTags.submission_lifecycle],
+    )
 
 
 def pdf_generation_success(submission: Submission, submission_report):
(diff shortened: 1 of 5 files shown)


src/openforms/logging/logevent.py
```diff
@@ -478,38 +554,15 @@ def appointment_register_failure(appointment: AppointmentInfo, plugin, error):
...
-        plugin=plugin,
-    )
-
-
-def appointment_update_failure(appointment: AppointmentInfo, plugin, error):
     _create_log(
-        appointment.submission,
-        "appointment_update_failure",
-        plugin=plugin,
-        error=error,
```
vaszig, line 512: Removed this since we do not support update in the new appointments flow (have to cancel first and create a new one).

viktorvanwijk (commented):
Looks good. One extra question: I see that the `stuf_zds_*` and `stuf_bg_*` logging events didn't get the tag `submission_lifecycle`. Perhaps I don't fully understand how it works, but aren't they part of a registration backend and therefore executed upon submission? I guess the question is if they showed up in the submission logging table before this change, because it seems they would not be included now.

vaszig:
> Looks good. One extra question: I see that the `stuf_zds_*` and `stuf_bg_*` logging events didn't get the tag `submission_lifecycle`. Perhaps I don't fully understand how it works, but aren't they part of a registration backend and therefore executed upon submission? I guess the question is if they showed up in the submission logging table before this change, because it seems they would not be included now.

Yes, your question makes sense. I will have a look again to see why I didn't include them.

src/openforms/logging/logevent.py
```diff
@@ -221,14 +238,19 @@ def prefill_retrieve_failure(submission: Submission, plugin, error: Exception):
...
 def registration_start(submission: Submission):
-    _create_log(submission, "registration_start")
+    _create_log(
+        submission,
+        "registration_start",
+        tags=[TimelineLogTags.submission_lifecycle],
+    )
 
 
 registration_debug = partial(_create_log, event="registration_debug")
```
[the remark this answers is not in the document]
sergei-maertens: Let's. It's used when deciding which configured registration backend to use and is currently visible.

src/openforms/logging/logevent.py
```diff
@@ -201,7 +218,7 @@ def prefill_retrieve_success(submission: Submission, plugin, prefill_fields):
         "prefill_retrieve_success",
         extra_data={"prefill_fields": prefill_fields},
         plugin=plugin,
-        tags=[TimelineLogTags.AVG],
+        tags=[TimelineLogTags.submission_lifecycle],
```
sergei-maertens, line 221: does the tag AVG need to be replaced or should be add the submission lifecycle tag to it? I'd expect the latter, because the fact that we retrieve personal details is an AVG event.

sergei-maertens (approved):

PR #5094, opened 2025-02-07
Add API endpoint for default fixed metadata variables of JSON dump plugin

### LINKED ISSUE #5087: Fixed metadata variables are not shown in options of JSON dump registration backend
When adding a new JSON dump backend, all initial options are filled in by the frontend. Because initial value of fixed metadata variables is set to an empty list (see [code](https://github.com/open-formulieren/open-forms/blob/87c96f11a935e28b9cf69acc88c34376522360de/src/openforms/js/components/admin/form_design/registrations/json_dump/JSONDumpOptionsForm.js#L47)) and the fixed metadata variables field in the serializer is not set to read only, they will be overwritten when saving the form, resulting in a permanent empty list of fixed variables.

Image

I see three options:
* Add the default fixed metadata variables to the frontend: doesn't seem like the right thing to do at all
* Make the fixed metadata variables field in the serializer read only: prevents overwriting it with an empty list, but the initial options when adding the backend plugin would still be empty. Also, getting the value for this field in the backend is tricky, because read-only fields are not in the `validated_data` of the serializer.
* Add some extra endpoints to `src/openforms/registrations/api/urls.py` like is done for Camunda/Objects API/ZGW APIs/... to retrieve extra information from the backend: probably the best option

viktorvanwijk:
Closes #5087

**Changes**

Add API endpoint for default fixed metadata variables of JSON dump plugin.

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


vaszig (approved):
I am not completely familiar with the feature but this looks ok to me. 

Having read the possible approaches you mention in the issue itself, I agree with the one that you selected. I don't know though if there was another solution that doesn't come to my mind now.

PR #5096, opened 2025-02-10
Return 'service fetch config' api data with original text case

### LINKED ISSUE #5089: Service fetch automatically changes snake_case to camelCase

3.0.2

### Customer reference

DH

Example HTTP query string with a "data_attrs" key:

```
{"type": ["https://objecttypen.nl/api/v2/objecttypes/23fb3e3e-c313-4be0-b8ed-51cdc5e31449"], "data_attrs": ["training__naam__exact__{{keuzeTraining}},training__status__exact__OPEN_VOOR_INSCHRIJVINGEN"]}
```
turns into "dataAttrs": 

Image

Which breaks the service fetch, until the form builder manually changes the camelCase back to snake_case. 

This causes double actions to be performed after each migration of a form: first set the right service, then change the key's to whatever they need to be. 

(It is also believed these changes get reverted in some scenario's and break the service fetch, because it's using "dataAttrs" again)

The config that exists in the service fetch should not "change" under any circumstance. 

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #5089

**Changes**

Make sure that the GET request to the `api/v2/service-fetch-configurations` call returns `query_params` and `body` data using the correct text case type

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/variables/api/renderers.py
+++ b/src/openforms/variables/api/renderers.py
@@ -1,0 +1,12 @@
+from djangorestframework_camel_case.render import CamelCaseJSONRenderer
+
+
+class ServiceFetchConfigurationRenderer(CamelCaseJSONRenderer):
+    # The field body and query_params needs to be ignored, to prevent accidental snake_case to camelCase changes.
+    # See github issue https://github.com/open-formulieren/open-forms/issues/5089
+    json_underscoreize = {
+        "ignore_fields": (
+            "body",
+            "query_params",
+        )
+    }
--- a/src/openforms/variables/api/viewsets.py
+++ b/src/openforms/variables/api/viewsets.py
@@ -8,6 +8,7 @@ from openforms.api.utils import mark_experimental
 from openforms.forms.api.permissions import FormAPIPermissions
 
 from ..models import ServiceFetchConfiguration
+from .renderers import ServiceFetchConfigurationRenderer
 from .serializers import ServiceFetchConfigurationSerializer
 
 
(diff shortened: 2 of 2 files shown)


vaszig (approved):

src/openforms/forms/tests/e2e_tests/test_service_fetch.py
```diff
@@ -466,6 +466,151 @@ def assertState():
 
             await assertState()
 
+    @skip_on_webtest
+    async def test_service_fetch_configuration_save_existing_as_new_with_snake_case_query_params_and_body(
```
sergei-maertens, line 470: This regression test can be written with a normal API endpoint (unit) test rather than a slow end to end test.

sergei-maertens (changes requested):

src/openforms/forms/tests/e2e_tests/test_service_fetch.py
robinmolen: Check, i've replaced this test with one that just tests the api response

sergei-maertens (approved):

robinmolen:
Backports:

* stable/3.0.x: https://github.com/open-formulieren/open-forms/commit/ddc7ebf9bc6b17f411e9bc71afe9db136c3772d2
* stable/2.8.x: https://github.com/open-formulieren/open-forms/commit/dc66a9d2933b56d2694409122dcaac9919bb9ab6

PR #5097, opened 2025-02-10
Wrapping array data fields in variable table

### LINKED ISSUE #5093: Working with array's in user variables tab is not very nice
### Thema / Theme

Frontend

on 1920 X 1080 resolution working with array's is not very nice, the array itself is rendered just off-screen which requires scrolling 😲 horizontally 😲 to add/remove items. 

Image

_No response_

_No response_

robinmolen:
Closes #5093

**Changes**

Displaying the array fields below the `Use raw JSON input` checkbox in the form variable table.

Screenshot 2025-02-10 at 17 14 50

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/forms/Field.js
+++ b/src/openforms/js/components/admin/forms/Field.js
@@ -59,6 +59,7 @@ const Field = ({
   errorClassPrefix = '',
   errorClassModifier = '',
   noManageChildProps = false,
+  extraModifiers = [],
 }) => {
   const intl = useIntl();
   const originalName = name;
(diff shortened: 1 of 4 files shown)


src/openforms/js/components/admin/forms/Field.js
```diff
@@ -146,6 +151,7 @@ Field.propTypes = {
   ]),
   fieldBox: PropTypes.bool,
   disabled: PropTypes.bool,
+  extraModifiers: PropTypes.arrayOf(PropTypes.oneOf(['flex-wrap'])),
```
vaszig, line 154: Should this be this specific array?I mean, if we want to add something in the future we will have to update this as well, right?

vaszig (approved):

src/openforms/js/components/admin/forms/Field.js
robinmolen: I think this is what we want. This ensures that we only allow this specific modifier to be used.

And indeed if we want to support more modifiers in the future, we will have to update this array. But it ensures that only expected values are used.

PR #5098, opened 2025-02-11
[#5039] Fix email registration errors

### LINKED ISSUE #5039: Not setting plain text in email registration, leads to error with no visible error message (nonFieldErrors worden niet getoond).

vaszig:
Closes #5039

**Changes**

- Email registration used `nonField` errors for a combination of fields and these were not shown on the frontend. Fixed these by showing them on the top of the modal.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
+++ b/src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
@@ -8,6 +8,7 @@ import {
   ValidationErrorsProvider,
   filterErrors,
 } from 'components/admin/forms/ValidationErrors';
+import ErrorMessage from 'components/errors/ErrorMessage';
 import {getChoicesFromSchema} from 'utils/json-schema';
 
 import EmailAttachmentFormatsSelect from './fields/EmailAttachmentFormatsSelect';
@@ -34,8 +35,17 @@ const EmailOptionsFormFields = ({name, schema}) => {
   ).map(([value, label]) => ({value, label}));
 
   const relevantErrors = filterErrors(name, validationErrors);
+  const nonFieldErrors = relevantErrors.filter(([field]) => field === 'nonFieldErrors');
+
   return (
     <ValidationErrorsProvider errors={relevantErrors}>
+      {nonFieldErrors && (
+        <>
+          {nonFieldErrors.map(([field, message], index) => (
+            <ErrorMessage key={`${field}-${index}`}>{message}</ErrorMessage>
+          ))}
+        </>
+      )}
       <Fieldset
         title={
           <FormattedMessage
--- a/src/openforms/js/components/admin/form_design/RegistrationFields.stories.js
+++ b/src/openforms/js/components/admin/form_design/RegistrationFields.stories.js
@@ -1030,6 +1030,51 @@ export const Email = {
   },
 };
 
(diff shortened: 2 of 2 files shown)


src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
```diff
@@ -34,8 +35,17 @@ const EmailOptionsFormFields = ({name, schema}) => {
   ).map(([value, label]) => ({value, label}));
 
   const relevantErrors = filterErrors(name, validationErrors);
+  const nonFieldErrors = relevantErrors.filter(([field]) => field.includes('nonFieldErrors'));
```
viktorvanwijk, line 38: Is there a reason why you included the parentheses and brackets in `([field])`? It would also work without them, right?

src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
```diff
@@ -34,8 +35,17 @@ const EmailOptionsFormFields = ({name, schema}) => {
...
+
   return (
     <ValidationErrorsProvider errors={relevantErrors}>
+      {nonFieldErrors && (
+        <>
+          {nonFieldErrors.map(([_, message], index) => (
```
viktorvanwijk, line 44: `index` doesn't seem to be used. Is there a reason you included it?

viktorvanwijk (commented):
Works as expected :). As you already pointed out, though,  might be nice if there is a solution which will work for all option forms, not only this plugin.

src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
vaszig: No, I had included this because I thought I would use it for the key, I can remove it.

src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
vaszig: The brackets are there because it means that each item in `relevantErrors` is expected to be an array, and we are extracting the first element.

viktorvanwijk: Wouldn't it make more sense to do a string comparison then, instead of a `.includes`?

src/openforms/js/components/admin/form_design/registrations/email/EmailOptionsFormFields.js
sergei-maertens: whenever you `array.map(foo => <component />)`, the component must get the `key`. See https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

sergei-maertens (approved):

sergei-maertens:
Accepting this as it is - I think with the UX rework we'll have the chance to fix this properly so that you don't have to think about it at all anymore.

sergei-maertens:
Backports:

* 3.0.x: 4b1ecdf60b99b54711f1041f790d03d52e4b11f3
* 2.8.x: skipped as it didn't have the UX rework yet

PR #5099, opened 2025-02-11
Accessibility improvements for submission report pdf

### LINKED ISSUE #5047: Accessibility improvements - PDF
Accessibility improvements for the submission summary PDF.

- [x] 1.1.1 (non-textual content) The logo in the pdf contains a textual alternative, but cannot be accessed. (#5052)
- [ ] 1.3.1 (information and relations) Missing relations between (submission rapport) PDF field labels and values. (#5099)
- [ ] 1.3.1 (information and relations) When a (submission rapport) PDF field doesn't have a values, we should provide a "No information provided" alternative. (This is to ensure that users understand that a field was left empty) (#5099)
- [ ] 2.4.2 (navigation - page title) The title in the PDF meta description contains leading and trailing whitespace (#5099)

robinmolen:
Closes #5047

**Changes**

- Added semantic relationships between field label and field content.
- When a field is left empty by the user, the submission report will display 'No information provided' in place of the missing field content.
- Removed leading and trailing whitespace in pdf metadata title

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/formio/formatters/base.py
+++ b/src/openforms/formio/formatters/base.py
@@ -70,9 +70,14 @@ class FormatterBase(Generic[ComponentT]):
         # note all this depends on value not being unexpected type or shape
         values = self.normalise_value_to_list(component, value)
 
-        formatted_values = (
+        formatted_values = list(
             force_str(self.format(component, value)) for value in values
         )
+
+        # Check if formatted_values isn't empty
+        if len(formatted_values) == 0:
+            return self.process_result(component, "")
+
         # logically we'd want a .filter_formatted_values() step here
         return self.process_result(
             component, self.join_formatted_values(component, formatted_values)
--- a/src/openforms/formio/formatters/formio.py
+++ b/src/openforms/formio/formatters/formio.py
@@ -121,6 +121,8 @@ class SelectBoxesFormatter(FormatterBase):
             # For the html output, wrap the values in li tags and put it inside an ul tag.
             # The selectboxes formatter handles all values at the same time,
             # so handle the full html formatting here.
+            if not selected_labels:
+                return ""
             return format_html(
                 "<ul>{values}</ul>",
                 values=format_html_join(
--- a/src/openforms/submissions/templates/report/submission_report.html
+++ b/src/openforms/submissions/templates/report/submission_report.html
@@ -6,11 +6,13 @@
 <html lang="{{ LANG }}" class="{{ theme.get_classname }}">
     <head>
         <meta charset="utf-8">
-        <title>{% block title %}
-            {% blocktrans with form_name=report.form.name trimmed %}
-                {{ form_name }}: PDF report
-            {% endblocktrans %}
-        {% endblock %}</title>
+        <title>{% spaceless %}
+            {% block title %}
+                {% blocktrans with form_name=report.form.name trimmed %}
+                    {{ form_name }}: PDF report
+                {% endblocktrans %}
+            {% endblock %}
+        {% endspaceless %}</title>
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <link href="{% static 'bundles/pdf-css.css' %}" media="all" rel="stylesheet" />
         {% block extra_css %}{% endblock %}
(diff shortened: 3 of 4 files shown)


src/openforms/formio/formatters/base.py
```diff
@@ -73,6 +74,15 @@ def __call__(self, component: ComponentT, value: Any) -> str:
         formatted_values = (
             force_str(self.format(component, value)) for value in values
         )
+
+        # Check if formatted_values isn't empty
+        formatted_values_copy, formatted_values = tee(formatted_values)
```
vaszig, line 79: I haven't used this before, it seems nice instead of a list I guess, but I do not know how it behaves concerning performance and efficiency.

vaszig (commented):
I am not sure if we should have the translations in this PR. I remember we do it in the release flow (let me know if something has changed).

robinmolen:
> I am not sure if we should have the translations in this PR. I remember we do it in the release flow (let me know if something has changed).

Ah yeah, you're right.. I'll remove them from the PR 👍

src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -65,7 +67,7 @@ <h1 class="title">{{ report.form.name }}</h1>
 
             {% for submission_step_node in report.renderer.get_children %}
                 {% if submission_step_node.has_children %}
-                    <div class="submission-step">
+                    <section class="submission-step">
```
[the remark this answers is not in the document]
robinmolen: Yes, with the `section` elements we can convey that the content holds some relation to each other. The `div` elements doesn't convey any semantic value.

In this case we tell that all elements in this `section` are part of the same form step.

src/openforms/formio/formatters/base.py
robinmolen: Looking at the documentation again, i think it would be fine (this setup can be improved though..). It copies the `formatted_values` iterable object, so we can check the content on a copy without moving through the original iterable.

But i'll try to create a different solution

robinmolen: Oh yeah, this can be greatly simplified with list.. Changing it 👍

vaszig: No, I didn't say to change it. I just said that I haven't used it and I am not aware of how it works!If you think though that it should be changed to a list ok. I thought you have discussed this with Sergei and he suggested this

src/openforms/formio/formatters/formio.py
```diff
@@ -121,13 +121,17 @@ def format(self, component: SelectBoxesComponent, value: dict[str, bool]) -> str
...
+                format_html(
+                    "<ul>{values}</ul>",
+                    values=format_html_join(
+                        "",
+                        "<li>{}</li>",
+                        ((selected_label,) for selected_label in selected_labels),
+                    ),
+                )
+                if len(selected_labels) > 0
+                else ""
```
sergei-maertens, line 134: codewise I think a guard clause would be more readable:

```py
if not selected_labels:
    return ""
return format_html(...)
```
this also makes it more easy to see where we're generating (potentially unsafe) HTML that needs to escape inputs.

src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -76,19 +78,23 @@ <h2 class="subtitle">{{ submission_step_node.render }}</h2>
                                         {% if component_node.component.type == "fieldset" %}
                                             <span class="submission-step-row__fieldset-label">{{ component_node.label }}</span>
                                         {% else %}
-                                            <div class="submission-step-row__label">
+                                            <span id="{{ component_node.key }}" class="submission-step-row__label">
```
sergei-maertens, line 81: I'm not sure if these are sufficiently unique if you use repeating groups, have you tested that?

robinmolen: Yeah, this wasn't enough for repeating groups... I've added the `component_node.configuration_path` as prefix, to ensure unique ids. I'm not entirely sure `configuration_path` is the best fit in this situation, but it creates the right result.

I've tried using the `component_node.path` and `component_node.json_renderer_path`, but couldn't get those to render properly...

If you have another solution in mind for this, i would love to hear it

viktorvanwijk: Excellent :upside_down_face:

sergei-maertens (approved):

PR #5100, opened 2025-02-12
Read only submission report content

robinmolen:
Closes open-formulieren/security-issues#34

**Changes**

The submission report content is now read-only in the admin

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/admin.py
+++ b/src/openforms/submissions/admin.py
@@ -500,6 +500,7 @@ class SubmissionReportAdmin(PrivateMediaMixin, admin.ModelAdmin):
     raw_id_fields = ("submission",)
 
     private_media_fields = ("content",)
+    readonly_fields = ("content",)
 
     def has_add_permission(self, request, obj=None):
         return False
--- a/src/openforms/submissions/admin.py
+++ b/src/openforms/submissions/admin.py
@@ -500,6 +500,7 @@ class SubmissionReportAdmin(PrivateMediaMixin, admin.ModelAdmin):
     raw_id_fields = ("submission",)
 
     private_media_fields = ("content",)
+    private_media_view_options = {"attachment": True}
     readonly_fields = ("content",)
 
     def has_add_permission(self, request, obj=None):
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -187,7 +187,7 @@ django-otp==1.3.0
     # via django-two-factor-auth
 django-phonenumber-field==7.3.0
     # via django-two-factor-auth
-django-privates==1.5.0
+django-privates==3.0.0
     # via
     #   -r requirements/base.in
     #   django-simple-certmanager
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -320,7 +320,7 @@ django-phonenumber-field==7.3.0
     #   -c requirements/base.txt
     #   -r requirements/base.txt
     #   django-two-factor-auth
-django-privates==1.5.0
+django-privates==3.0.0
     # via
     #   -c requirements/base.txt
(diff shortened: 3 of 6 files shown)


sergei-maertens (changes requested):
Additionally I'd configure the `PrivateMediaMixin` so that the files are not downloaded with `inline` directives in the `Content-Disposition` header. I *think* you should be able to get there with:

* https://django-privates.readthedocs.io/en/latest/reference.html#privates.admin.PrivateMediaMixin.get_private_media_view_options to pass `django-sendfile2` options to the underlying view. Should be achievable with the `private_media_view_options` attribute on the admin class.
* passing `attachment=True` to https://django-sendfile2.readthedocs.io/en/latest/django_sendfile.utils.html#django_sendfile.utils.sendfile

The `sendfile` options are consumed here: https://github.com/sergei-maertens/django-privates/blob/27c7391ef82479f5d99fb6542647ab999b3ebd2a/privates/views.py#L38

src/openforms/submissions/admin.py
```diff
@@ -498,8 +498,16 @@ class SubmissionReportAdmin(PrivateMediaMixin, admin.ModelAdmin):
         "submission__public_registration_reference",
     )
     raw_id_fields = ("submission",)
+    fields = (
+        "title",
+        "content",
+        "submission",
+        "last_accessed",
+        "task_id",
+    )
```
sergei-maertens, line 507: why do you specify these explicitly? by default all model fields are used iirc. I'd expect this not to be necessary.

robinmolen:
Regarding the `Content-Disposition` header in the SDK (the download link citizens will receive after submitting a form):

I assume the test findings were from a pre-2.8.0 environment. In the release of 2.8.0 the download link shown in the SDK was changed. From that point the correct `Content-Disposition` header were used (change was made in ticket #4488).

src/openforms/submissions/admin.py
```diff
@@ -488,6 +489,13 @@ def retry_processing_submissions(self, request, queryset):
             on_post_submission_event(submission.id, PostSubmissionEvents.on_retry)
 
 
+class SubmissionReportMediaView(PrivateMediaView):
+    def get_sendfile_opts(self):
```
robinmolen, line 493: I've tried defining this as `sendfile_options = { "attachment": True }`, instead of using the `get_sendfile_opts()` function, but this doesn't seem to work. So this is the smallest/most simplistic setup we can do

sergei-maertens:
I'll take a look at django-privates now for the readonly field :)

src/openforms/submissions/admin.py
sergei-maertens: err, looking at the [code](https://github.com/sergei-maertens/django-privates/blob/05fb2b154d4c0544513cb6e0e5a9b5cab4ac6f96/privates/admin.py#L43), it should be sufficient to define on the existing admin class:

```py
class SubmissionReportAdmin(...):
    ...
    private_media_view_options = {"attachment": True}
```
I suspect you were missing the `private_media_` prefix

robinmolen: Ahh yeah that could be it.. I'll take another look 👍

robinmolen: Yeah that does the trick, nice!

sergei-maertens:
Backports:

* 2.8.9: 7a128cba1784f8f7f9719cf60a2c2d3bbbd52fc0
* stable/3.0.x: 0ee925a232b3954ea9ab8f5ecef3ec171934d102

PR #5123, opened 2025-02-26
Components keeping original defaultValue when editing

### LINKED ISSUE #5104: Radio component defaultValue changes to null after saving. Instead of the desired: "".

3.0.4

### Customer reference

reported by: DH, hlm, Ensc

When you add a radio component to a step, the initial defaultValue is set to: `""`. 
By saving the component the defaultValue gets changed to `null`.

The issue with `null` is when this radio is hidden with "empty when hidden" is enabled, the component is included in the check_logic as `"radio": "",` which in turn triggers validation for the step with no user interaction. Unfortunately this issue is pretty prevalent across environments. 

After manually changing the `null` back to `""` in the json of the radio component - this validation triggering is no longer an issue. 

Image

Image

i expect the defaultValue to not cause validation to trigger when a step loads

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #5104

**Changes**

When creating radio component the configuration defined in the component schema will be used. Providing an empty string as the `defaultValue`. If you leave the `defaultValue` as an empty string, save this component and edit it again, the `defaultValue` will change to `null`. This `defaultValue` change happens because of how we construct the component data for the edit modal.

In the `WebformBuilder.editComponent` we receive the `component` property, which is a simplified representation of the component (a sub-set of all the component information). If the `defaultValue` is a falsey value, like in this case an empty string, then `defaultValue` won't be present in this `component` prop. When reconstructing the component to its full representation, the `defaultValue` defined in the Formio component schema will be used (`null`).

To make sure we retain the actual component `defaultValue`, we combine the `original` prop (which is the original component data) and the reconstructed component data.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


sergei-maertens (commented):
I'm terrified of this change because of the earlier issues that happened with it, and on how Formio itself figures out the default value for a component: https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/src/components/_classes/component/Component.js#L2302

Can you please make sure that all the old reported bugs are not re-introduced and test them manually? You should be able to figure out which ones that are from the commit history/git blame.

sergei-maertens:
We should also document the manual JSON editing requirement when backporting the fix. For the master branch, we can do a data migration.

robinmolen:
> I'm terrified of this change because of the earlier issues that happened with it, and on how Formio itself figures out the default value for a component: https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/src/components/_classes/component/Component.js#L2302

Your comment was correct, the change of merging the `original` into the generated component isn't needed. (the change in the `WebformBuilder.js`) The problem comes from the default formio component schema's. Formio doesn't define a `defaultValue` in the radio component schema, and our radio component schema wasn't seen as the schema that formio should use.

So the bug was as follows: you create a form with a radio component. This component has a default value of empty string, so when you save the form the `defaultValue` for the component is set to empty string. If you then edit the component the`defaultValue` is set to `null`. This is done by how formio calculates the default value (the link in your comment). It uses the schema `defaultValue` as base, and if the component `defaultValue` is truety this then will be uses as the real defaultValue. Because the component `defaultValue` was set to empty string, which is falsey, the `defaultValue` of the formio schema was used (i.e. a undefined value, resulting in `null`).

By telling formio to use our schema's (which is the formio schema + some custom stuff) we can ensure that a valid `defaultValue` is used.

sergei-maertens (approved):
I have... strong feelings... about Formio

robinmolen:
For the backport: To ease the pain of having to fix all radio components defaultValues by hand, i've added a fix script `/bin/fix_radio_component_default_values.py`.
This looks at all radio components in all form definitions. Any radio component that has a defaultValue of `None` (this will present in the admin view as a defaultValue of `null`), will have its defaultValue changed to an empty string.

robinmolen:
Backports:

stable/2.8.x: 870fc20731342032d08036b1a9b36fdfbd7a82e8

PR #5127, opened 2025-02-27
Add unique key to Action component

### LINKED ISSUE #5040: When you have 2 logic actions (in the same trigger) that set a variable. And you delete the top one. The 2nd one will get the JSON-logic value of the deleted one.

viktorvanwijk:
Closes #5040

**Changes**

Added a unique key to the Action component.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/FormLogic.stories.js
+++ b/src/openforms/js/components/admin/form_design/FormLogic.stories.js
@@ -161,6 +161,8 @@ export const FullFunctionality = {
               value: '',
             },
 
+            uuid: '',
+            _generatedId: '0',
             component: '',
             formStepUuid: null,
             variable: 'bar',
@@ -197,6 +199,8 @@ export const FullFunctionality = {
               },
             },
 
+            uuid: '',
+            _generatedId: '0',
             component: '',
             formStepUuid: null,
             variable: '',
@@ -237,6 +241,8 @@ export const DeletingOneOfMultipleActionsInSameTrigger = {
 
         actions: [
           {
+            uuid: '',
+            _generatedId: '0',
             component: '',
             variable: 'foo',
             formStepUuid: null,
@@ -246,6 +252,8 @@ export const DeletingOneOfMultipleActionsInSameTrigger = {
             },
           },
           {
+            uuid: '',
+            _generatedId: '1',
             component: '',
             variable: 'bar',
             formStepUuid: null,
--- a/src/openforms/js/components/admin/form_design/FormLogic.stories.js
+++ b/src/openforms/js/components/admin/form_design/FormLogic.stories.js
@@ -1,4 +1,4 @@
-import {fn} from '@storybook/test';
+import {expect, fn, userEvent, within} from '@storybook/test';
 
 import {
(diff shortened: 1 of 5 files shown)


viktorvanwijk:
According to the [docs](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key):

> Rather than generating keys on the fly, you should include them in your data

I have considered doing this by adding a `(uu)id` to `EMPTY_ACTION`, but this would require it to be saved in `FormLogic.actions` field, and I wasn't confident that was the way to go

src/openforms/js/components/admin/form_design/logic/actions/ActionSet.js
```diff
@@ -6,11 +6,14 @@ import {useImmerReducer} from 'use-immer';
...
 import ButtonContainer from 'components/admin/forms/ButtonContainer';
 import useOnChanged from 'hooks/useOnChanged';
+import {getUniqueRandomString} from 'utils/random';
 
 import Action from './Action';
 import {ActionError} from './types';
 
 const EMPTY_ACTION = {
+  uuid: '',
+  _generatedId: '', // consumers should generate this, as it's used for the React key prop when uuid does not exist
```
vaszig, line 16: At the beginning I wasn't sure about this but I saw that we already make use of it in various places in the project.

src/openforms/js/components/admin/form_design/FormLogic.stories.js
```diff
@@ -209,3 +213,78 @@ export const FullFunctionality = {
...
 };
+
+export const DeletingOneOfMultipleActionsInSameTrigger = {
+  name: 'Deleting one of multiple actions in the same trigger',
+
+  args: {
+    logicRules: [
+      {
+        uuid: 'foo',
+        _generatedId: 'foo', // consumers should generate this, as it's used for the React key prop if no uuid exists
```
vaszig, line 224: I this needed here?Since it's a fallback if uuid is not present right?

src/openforms/js/components/admin/form_design/logic/actions/ActionSet.js
viktorvanwijk: Yeah, I've used the same approach for what we have with `EMPTY_RULE`

src/openforms/js/components/admin/form_design/FormLogic.stories.js
viktorvanwijk: You're right, it's not actually needed here. I think I copied this rule from one of the stories above. Do you think we should update or leave them like this?

vaszig: Let's see what Sergei thinks generally about the PR and we can then change it or not. It's not a big deal, I mentioned that because it's nice to reflect the code functionality here as well.

viktorvanwijk: Yeah makes sense :)

sergei-maertens (approved):

PR #5135, opened 2025-03-04
[#5124] Hidden/disabled prefill fields trigger validation

### LINKED ISSUE #5124: unused/hidden read-only prefill fields trigger error

3.x.x

### Customer reference

DH, 's-Hertogenbosch

When a form has multiple field sets for a "uw-gegevens" step e.g:
- DigiD
- eHerkenning
- Anon

And these field sets are hidden and the fields within (being dedicated for pre fill) are disabled: Trigger the "Je mag de prefill-gegevens niet wijzigen" error on "validate" call. 

In this case the prefill did not return any value, and the field set is hidden because the user started the form with another authentication plugin. 

As the fields that do not pass validation are not prefilled at all this should not occur. 

Form 749 on maykin test. 
Start the form anonymous
Select: "Persoon"
Fill in fields and try to navigate to the next step.
See "validate" request.

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #5124 

**Changes**

- When  a component (or its parent is disabled and hidden, the validation in the backend is triggered since there is no data for the specific key. This means, for our validator, that the data has been changed and raises an error.

- Additional  test for the case someone omits the submission data (affects the prefill data).

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/api/validators.py
+++ b/src/openforms/submissions/api/validators.py
@@ -33,6 +33,11 @@ class ValidatePrefillData:
             if not (component_key := component.get("key")):
                 continue
 
+            # in case the component or its parent component is hidden the key will not be
+            # part of the data.
+            if component_key not in data:
+                continue
+
             original_prefill_value = prefill_data.get(component_key)
             if original_prefill_value is None:
                 # the value will be `None` if there is no actual prefill data available, so there is nothing to compare to. This
--- a/src/openforms/submissions/tests/test_submission_step_validate.py
+++ b/src/openforms/submissions/tests/test_submission_step_validate.py
@@ -157,6 +157,93 @@ class SubmissionStepValidationTests(SubmissionsMixin, APITestCase):
 
         self.assertEqual(status.HTTP_204_NO_CONTENT, response.status_code)
 
+    def test_prefill_data_with_hidden_component(self):
+        form = FormFactory.create()
+        step = FormStepFactory.create(
+            form=form,
+            form_definition__configuration={
+                "display": "form",
+                "components": [
+                    {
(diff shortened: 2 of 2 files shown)


viktorvanwijk (approved):

viktorvanwijk:
Just a failing end-to-end test, but probably flakiness by the looks of it :)

vaszig:
> Just a failing end-to-end test, but probably flakiness by the looks of it :)

Yes, I remember this failing in master as well. I re-run the CI one more time to see if this is the case.

sergei-maertens (approved):

src/openforms/submissions/api/validators.py
```diff
@@ -33,6 +33,11 @@ def __call__(self, data: dict, field: JSONField):
             if not (component_key := component.get("key")):
                 continue
 
+            # in case the component or its parent component is hidden the key will not be
+            # part of the data.
+            if component_key not in data:
+                continue
```
sergei-maertens, line 39: What happens now if a field is prefilled but not hidden and the data is omitted from the submission data? It kinda looks now like a user can clear a prefilled field that is read-only.

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Backports:

* 3.0.x: 159e3e0fb8919a226509cb818d248cb5518b0bb6
* 2.8.x: 0f68bbbe4eb9651a6cf7d73f649935bcac7502d0

PR #5021, opened 2025-01-17
Hotfix/broken objects api variable mappings

sergei-maertens:
(no issue created, was reported by Joeri yesterday)

**Changes**

* Fixed being able to configure variable mappings for in-memory added registration backends
* Fixed a prop type warning

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Dockerfile/scripts

  - [x] Updated the Dockerfile with the necessary scripts from the `./bin` folder

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

PR #5150, opened 2025-03-11
Fix missing variables mapping in objects api registration (v2)

### LINKED ISSUE #5031: Variables tab crash after setting Objects API registration.

3.0.1

### Customer reference

DH

After setting registration > Objects API (variable mapping). The variables tab in the from crashes: 

```
react-dom.production.min.js:189 TypeError: Cannot read properties of undefined (reading 'find')
    at summaryHandler (ObjectsApiSummaryHandler.js:30:59)
``` 

Previously mapped, or imported forms with "variables_mapping" do not show this behavior.

robinmolen:
Closes #5031

**Changes**

The bug was already fixed by a previous PR (https://github.com/open-formulieren/open-forms/pull/5021). This PR adds a detection and a fix script, so that we can monitor and fix affected form registration backends.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/bin/fix_objects_api_form_registration_variables_mapping.py
+++ b/bin/fix_objects_api_form_registration_variables_mapping.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+#
+# Fix the `variables_mapping` value in the options configuration of Objects API
+# registration backends.
+#
+# Due to a missing default value, some Objects API registrations don't have
+# `variables_mapping` in their `options` configuration. As this value is used when
+# configuring variable mappings, the missing `variables_mapping` causes javascript
+# errors.
+#
+# This script automatically fixes all options `variables_mapping` values for Objects API
+# registrations. Any missing or incorrect `variables_mapping` is set to `[]`.
+# Related to ticket: https://github.com/open-formulieren/open-forms/issues/5031
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import django
+from django.db.models import Q
+
+SRC_DIR = Path(__file__).parent.parent / "src"
+sys.path.insert(0, str(SRC_DIR.resolve()))
+
+
+def fix_objects_api_registration_variables_mapping():
+    from openforms.forms.models import FormRegistrationBackend
+    from openforms.registrations.contrib.objects_api.config import VersionChoices
+
+    form_registrations_to_update = []
+    registration_backends = (
+        FormRegistrationBackend.objects.filter(backend="objects_api")
+        .filter(~Q(options__version=VersionChoices.V1))
+        .iterator()
+    )
+
+    # Only Objects API registration of version 2, or without a version, will be present
+    for objects_api_registration in registration_backends:
+        should_update = False
+
+        # The objects_api_registration should always have a version, but just to be sure
+        if not objects_api_registration.options.get("version", None):
+            objects_api_registration.options["version"] = VersionChoices.V2
+            should_update = True
+
+        variables_mapping = objects_api_registration.options.get(
+            "variables_mapping", None
+        )
+        if not isinstance(variables_mapping, list):
+            objects_api_registration.options["variables_mapping"] = []
+            should_update = True
+
+        if should_update:
+            form_registrations_to_update.append(objects_api_registration)
+
+    FormRegistrationBackend.objects.bulk_update(
+        form_registrations_to_update, fields=["options"]
+    )
+
+
+def main(**kwargs):
+    from openforms.setup import setup_env
+
+    setup_env()
+    django.setup()
+
+    return fix_objects_api_registration_variables_mapping(**kwargs)
+
+
+if __name__ == "__main__":
+    main()
--- a/bin/report_form_registration_problems.py
+++ b/bin/report_form_registration_problems.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+from __future__ import annotations
+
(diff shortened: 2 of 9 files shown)


src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -257,7 +257,21 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
...
+    def _handle_import(self, attrs) -> None:
+        # we're not importing, nothing to do
+        if not self.context.get("is_import", False):
+            return
+
+        if attrs["version"] == VersionChoices.V2:
+            # Make sure `variables_mapping` has a valid value
+            if (
+                "variables_mapping" not in self.initial_data
+                or type(self.initial_data["variables_mapping"]) is not list
```
sergei-maertens, line 269: I'd also simplify this to:

```py
if not (variables_mapping := self.initial_data.get("variables_mapping")) or not isinstance(variables_mapping, list):
    attrs["variables_mapping"] = []
```

sergei-maertens (changes requested):

bin/report_form_registration_problems.py
```diff
@@ -0,0 +1,95 @@
...
+
+def check_registration_backend(registration: FormRegistrationBackend) -> str | None:
+    match registration.backend:
+        case "objects_api":
+            variables_mapping = registration.options.get("variables_mapping", None)
+
+            if variables_mapping is None:
+                return "'variables_mapping' is missing"
+            if type(variables_mapping) is not list:
+                return "'variables_mapping' is't a list"
```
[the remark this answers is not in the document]
robinmolen: Ah yeah i saw that in the component report. I thought to keep this (at least for now) only backend specific, to keep it a bit easier to understand?

bin/report_form_registration_problems.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def report_problems(backends: Sequence[str]) -> bool:
+    from openforms.forms.models import FormRegistrationBackend
+
+    problems = []
+
+    registration_backends = FormRegistrationBackend.objects.iterator()
```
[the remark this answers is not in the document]
robinmolen: ah yeah thats a bit nicer.. Okay, I'll change it 👍 (I'll also update the PR on the deploy project)

bin/fix_objects_api_form_registration_variables_mapping.py
```diff
@@ -0,0 +1,58 @@
...
+    from openforms.forms.models import FormRegistrationBackend
+
+    form_registrations_to_update = []
+    registration_backends = FormRegistrationBackend.objects.iterator()
+
+    for objects_api_registration in (
+        registration
+        for registration in registration_backends
+        if registration.backend == "objects_api"
+    ):
```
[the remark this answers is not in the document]
robinmolen: Ah chips, yeah i also notice my setup didn't account for v1 config.. (which shouldn't have the `variables_mapping`)

src/openforms/registrations/contrib/objects_api/config.py
robinmolen: Check, yeah i just noticed that the second part cannot happen (if you try to import a Objects API registration with a non list `variables_mapping `, the `variables_mapping` serializer field validation kicks in).

I've added a test to capture/prove this behaviour, and I'm removing this extra condition

bin/report_form_registration_problems.py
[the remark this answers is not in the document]
robinmolen: Im just noticing that this also doesn't check the version of the registration... 🫣 (wondering if i was awake yesterday 😅)

bin/report_form_registration_problems.py
```diff
@@ -0,0 +1,99 @@
...
+
+SRC_DIR = Path(__file__).parent.parent / "src"
+sys.path.insert(0, str(SRC_DIR.resolve()))
+
+
+def check_registration_backend(registration: FormRegistrationBackend) -> str | None:
+    from openforms.registrations.contrib.objects_api.config import VersionChoices
+
+    match (registration.backend, registration.options):
+        case ("objects_api", {"version": int() as version}) if version == VersionChoices.V2:
```
sergei-maertens, line 25:
```suggestion
        case ("objects_api", {"version": 2}):
```
this is equivalent

However, what it doesn't capture is that the `version` key may be absent, and that (currently) defaults to it being v2. I don't know if we should consider that case or not.

sergei-maertens (approved):

bin/report_form_registration_problems.py
robinmolen: Check, yeah i thought: 'we have a constant for this, why not use it :)'

Yeah checking for no version also crossed my mind, for the report and for the fix script. But from the moment we started working with versions, it defaulted to version 1 (which was recently (start of 2025) changed to version 2).
The import does handle this, so this case would be for Open Forms environments upgrading from 2.x to 3.1? But i assume some migration stuff is in place..

Maybe it would be best to play this save... I'll add it to the report and fix scripts

sergei-maertens: > we have a constant for this, why not use it :)

sure, fine - my remark is about the `"version": int() as version if version == 2` part :P

PR #5130, opened 2025-02-28
[#5074] Transform data for `selectboxes` in the registrations (variables tab)

vaszig:
Closes #5074 

**Changes**

- Add checkbox for transforming the data to a list for Objects API and JSON dump registations (variables tab)

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Problem detection in the admin email digest is handled

- Release management

  - [x] I have labelled the PR as "needs-backport" accordingly

- I have updated the translations assets (you do NOT need to provide translations)

  - [x] Ran `./bin/makemessages_js.sh`
  - [x] Ran `./bin/compilemessages_js.sh`

- Dockerfile/scripts

  - [x] Updated the Dockerfile with the necessary scripts from the `./bin` folder

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

PR #5153, opened 2025-03-13
Upgrade django-digid-eherkenning

### LINKED ISSUE #5136: "Dienstcatalogus" for eHerkenning uses current (old) certificate, but should include the future certificate when present

2.8.x

### Customer reference

Vught 89

When new eHerkenning metadata needs to be provided, in this case for a renewal of a PKI certificate. The "Dienstcatalogus" is generated with the old certificate instead of the future certificate which means while the metadata itself does contain the new and old cert, the old cert needs to be removed temporarily to generate the new "Dienstcatalogus" anyway. 

Which defeats the purpose of having both the current and new certificate present in the configuration. 
i believe it would be safe to always use the "future" certificate if one is present because if it is - the new metadata and dienstcatalogus are generated for that purpose. 

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #5136 

**Changes**

* Upgrade to django-digid-eherkenning 0.20.0
* Upgrade to lxml 5.3.1

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -157,7 +157,7 @@ django-csp==3.8
     # via -r requirements/base.in
 django-csp-reports==1.8.1
     # via -r requirements/base.in
-django-digid-eherkenning==0.19.2
+django-digid-eherkenning==0.20.0
     # via -r requirements/base.in
 django-filter==24.2
     # via -r requirements/base.in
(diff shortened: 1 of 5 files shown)


vaszig:
Oh I have the same failure in my PR #5130 , is it something that we are aware of?

sergei-maertens:
> Oh I have the same failure in my PR #5130 , is it something that we are aware of?

Looks like it's test flakiness, but it's gotten bad :/

PR #5154, opened 2025-03-13
Ensure variable bulk endpoint returns list of submitted variables

### LINKED ISSUE #5142: Logica regels worden verwijderd wanneer een keuzelijst wordt verwijderd

2.8.3

### Customer reference

Gemeente Leeuwarden

Een gebruiker heeft een dropdown met deze values erin: 
`"values": [{ "label": "Ouder", "value": "ouder" },{ "label": "Schoonouder, opa of oma", "value": "schoonouderOpaOfOma" },{ "label": "Broer of zus", "value": "broerOfZus" },{ "label": "Kamerhuurder", "value": "iemandDieEenKamerBijMijHuurt" },{ "label": "Kamerverhuurder", "value": "ikHuurEenKamerVanDezePersoon" },{ "label": "Kostganger/kostgever", "value": "kostgangerOfKostgever" },{ "label": "Anders", "value": "anders" }]`
Er zijn maar 4 logicaregels die deze variabele gebruiken, desondanks als de gebruiker de dropdown verwijdert, schoont dit alle logicaregels en gaan we van 261 componentvariabelen naar 0

https://github.com/user-attachments/assets/d0787246-f583-48ed-85b1-67c18a932141

[forms-export_ec8fa249-531e-4adf-bce1-ddde7f373e06.zip](https://github.com/user-attachments/files/19110247/forms-export_ec8fa249-531e-4adf-bce1-ddde7f373e06.zip)
 

Verwijder de watIsDezeMedebewonerVanU2 component en daarna het formulier opslaan

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

viktorvanwijk:
Closes #5142 

**Changes**

Added list of submitted variables to return of variable bulk update endpoint

**Checklist**

- Impact on features

  - [x] Checked copying a form
  - [x Checked import/export of a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/forms/api/viewsets.py
```diff
@@ -503,7 +503,12 @@ def variables_bulk_update(self, request, *args, **kwargs):
         ).exclude(key__in=keys_to_keep)
         stale_user_defined.delete()
 
-        return Response(serializer.data, status=status.HTTP_200_OK)
+        # Create return data
+        var_keys = [variable["key"] for variable in serializer.initial_data]
+        return_data = serializer.to_representation(
+            form.formvariable_set.filter(key__in=var_keys)
+        )
```
sergei-maertens, line 510: I don't like the low level `to_representation` usage, that's not how you're supposed to use serializers. Can't we instead create a new serializer instance?

```py
out_serializer = FormVariableSerializer(
    instance=form.formvariable_set.prefetch_related("form", "form_definition"),   # see variables_list below
    many=True
)
return Response(out_serializer.data, status=status.HTTP_200_OK)
```
?

I'm also not sure why the filtering on keys is needed - aren't we supposed to just return all the variables of the whole form anyways, and the work before ensures this state is consistent?

sergei-maertens (changes requested):

src/openforms/forms/api/viewsets.py
viktorvanwijk: > I'm also not sure why the filtering on keys is needed - aren't we supposed to just return all the variables of the whole form anyways, and the work before ensures this state is consistent?

This is what I had initially, but it caused `test_bulk_create_and_update_with_non_camel_case_initial_values` to fail. This made me think about what the behaviour was previously: which is to only return the variables that were submitted. So I tried to keep it consistent. You think it's better to always return everything?

sergei-maertens: Let's discuss in standup, I think I'm a bit too far removed from the bug/problem that's happening to give a proper answer.

sergei-maertens: Discussed - the expected behaviour is that the "resulting" set of form variables (component + user defined) is returned. The endpoint (via PUT) ensures that the stale component & user defined variables are cleaned up, which effectively means we discard the body data w/r to the component variables. Only user defined variable mutations actually have an effect. It's not pretty.

So, returning `form.formvariable_set.prefetch_related(...)` in a separate serializer instance should do the job.

sergei-maertens (approved):

sergei-maertens:
Backports:

* 3.0.x: ca18e5a5574bfaa57340a0db94177b0d6857548b
* 2.8.x: a13b01d34c8df9cd728db4514fb2f494179e0e01

Issue #5106, opened 2025-02-13
Referentielijsten API: use Accept-Language header when fetching items


stevenbal:
Implement language awareness, ensure the language the form is rendered in is passed along with the API request to fetch the items from Referentielijsten (as an Accept-Language header).

PR #5159, opened 2025-03-14
[#5106] Add language header to the referentielijsten request for items

vaszig:
Closes #...

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


vaszig:
I am not sure,should we add a test here?I mean the API does not support this right now so the only way could be something like a mock I guess

src/openforms/contrib/referentielijsten/client.py
```diff
@@ -39,17 +40,23 @@ def get_tabellen(self) -> list[Tabel]:
         all_data = list(pagination_helper(self, data))
         return all_data
 
-    def get_items_for_tabel(self, code: str) -> list[TabelItem]:
-        response = self.get("items", params={"tabel__code": code}, timeout=10)
+    def get_items_for_tabel(self, code: str, current_language: str) -> list[TabelItem]:
+        response = self.get(
+            "items",
+            params={"tabel__code": code},
+            timeout=10,
```
sergei-maertens, line 47: can we drop the timeout here? it should the taken from the configuration on the `Service` model and ape-pie already takes care of that

sergei-maertens (changes requested):

sergei-maertens:
> I am not sure,should we add a test here?I mean the API does not support this right now so the only way could be something like a mock I guess

Re-recording the VCR cassettes should be sufficient :)

PR #5161, opened 2025-03-17
:art: Remove await from expect

sergei-maertens:
expect is sync and should never be awaited.


vaszig (approved):
Let's wait for the CI and we can merge it.

PR #5164, opened 2025-03-17
[stable/2.8.x] Don't error hard on API standard violations

### LINKED ISSUE #5162: Relax API standard validation (2.8.x ONLY)
When we now parse the API version of ZGW API's, an exception is raised when it's not according to semver. This causes issues with non-standard compliant vendors.

Instead of crashing hard, assume the version is `1.0.0` which drops us into the worst compatibility mode.

We only apply this patch to 2.8.x and 3.0.x+ keeps the strict behaviour - it's an API _standard_ for a reason.

sergei-maertens:
Closes #5162

Before this patch, the behaviour was to outright reject non-standard compliant responses, since the API specifications for the ZGW API's (managed by VNG) are strictly versioned according to semver. Certain features are available only starting with certain versions, and we adapt our client behaviour to be as efficient as possible based on this API version. It is nonsense for an API server to reply with a version number that is not published by VNG on the documentation site https://vng-realisatie.github.io/gemma-zaken/standaard/.

That said, Dimpact is in a pickle because a vendor is not behaving according to the standard, triggering this hard rejection in Open Forms, and said vendor is not able to patch their own systems in time for a particular release time line. We agreed to coalesce the version to the lowest possible supported so that the application can actually be used still, however this triggers all the worst performance and UX paths.

This patch is *only* applied to the stable/2.8.x branch since we really want to bring the ecosystem to standard-compliant implementations.

### DIFF
--- a/src/openforms/contrib/zgw/clients/catalogi.py
+++ b/src/openforms/contrib/zgw/clients/catalogi.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import date
 from functools import cached_property
 from operator import itemgetter
@@ -10,6 +11,8 @@ from zgw_consumers.nlx import NLXClient
 from openforms.utils.api_clients import PaginatedResponseData, pagination_helper
 
 from ..exceptions import StandardViolation
+
+logger = logging.getLogger(__name__)
 
 
 def noop_matcher(roltypen: list) -> list:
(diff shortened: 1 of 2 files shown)


robinmolen (approved):

sergei-maertens:
And it has to go to 3.0.x after all... ba82a9a6c663b4948b9c43e314f486cec17bae97

PR #5167, opened 2025-03-17
Fix test flakiness

sergei-maertens:
Avoids a race condition where the DB transaction was not saved yet by the API call made through the Javascript before asserting the DB state.

The page.wait_for_url was probably a no-op because the page URL is already satisfied when editing a form and reloading that same page. This is mitigated by hitting the save that redirects to the form list page, which has a different URL and provides stronger guarantees that the changes have been persisted in the database.


sergei-maertens:
Labeled with backport since it makes the stable branches CI builds flaky too.

vaszig (approved):

sergei-maertens:
Backports:

* 3.0.x: 31139c469c199f7282197acd98accbed17b4ac54
* 2.8.x: 20876cf47b0988af50a2b5c2b284331383629151

PR #5170, opened 2025-03-18
Hidden map fails form validation

### LINKED ISSUE #5151: Hidden map will error on "validate" call

2.8.5 > 3.1.x

### Customer reference

lv 221

When a map component is hidden, with the option "clear when hidden" is enabled. Navigating to the next step is impossible since the validate call will return a error: 

3.1.x
```
    "type": "https://open-forms/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:17b996e3-3254-42b8-8f90-17bd92be3630",
    "invalidParams": [
        {
            "name": "data.kaartlocatie1.nonFieldErrors",
            "code": "invalid",
            "reason": "Ongeldige data. Verwacht een dictionary, kreeg een list."
        }
    ]
}
```
2.8.6
```
{
    "type": "https://frm.nl/fouten/ValidationError/",
    "code": "invalid",
    "title": "Invalid input.",
    "status": 400,
    "detail": "",
    "instance": "urn:uuid:518b2930-3ecc-404f-b8c3-52de6e9c79db",
    "invalidParams": [
        {
            "name": "data.kaartLocatie",
            "code": "not_a_list",
            "reason": "Een lijst van waarden was verwacht, maar kreeg \"str\"."
        }
    ]
}
```

Maykin test form: 194. 
Don't select the checkbox

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #5151

**Changes**

If a map component is `hidden` (with `clearOnHide` set to `True`), the check-logic call would return an empty value for the map component. This empty value would then in turn be merged into the SDK form data. If we then click on "next step", the form data will be validated and the empty value for the map component won't be conform the map component serializer validation, causing errors the user cannot fix (and locking the form).

The only correct empty value for a map component, is **NO** value.

These changes make sure that missing map components don't receive empty object values. Ensuring that forms cannot be locked by hidden and empty map components.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/bin/report_component_problems.py
+++ b/bin/report_component_problems.py
@@ -20,6 +20,9 @@ sys.path.insert(0, str(SRC_DIR.resolve()))
 
 
 def check_component(component: Component) -> str | None:
+    from openforms.api.geojson import GeoJsonGeometryPolymorphicSerializer
+    from rest_framework.exceptions import ValidationError
+
     match component:
         case {"type": "file", "defaultValue": list() as default_value}:
             if None in default_value:
@@ -68,6 +71,26 @@ def check_component(component: Component) -> str | None:
                 if bool(translation_dict.get("defaultValue")):
                     return "defaultValue has a translation"
 
+        case {"type": "map"}:
+            default_value = component.get("defaultValue")
+            if default_value is None:
+                # No defaultValue is a-okay
+                return None
+
+            try:
+                if isinstance(
+                    default_value, dict
+                ) and GeoJsonGeometryPolymorphicSerializer(data=default_value).is_valid(
+                    raise_exception=True
+                ):
+                    # This object has a valid Geometry shape
+                    return None
+            except ValidationError as error:
+                return f"Default value '{default_value}' is not valid.\nError: {error.detail}"
+
+            # Any other value is automatically invalid
+            return f"Default value '{default_value}' is not valid."
+
 
 def check_component_html_usage(component: Component) -> list[str]:
     messages = []
@@ -80,7 +103,9 @@ def check_component_html_usage(component: Component) -> list[str]:
         property_value = component[property_name]
         property_contains_html = bool(re.search(r"<\w", property_value))
         if property_contains_html:
-            messages.append(f"Component {property_name} contains html: '{property_value}'.")
+            messages.append(
+                f"Component {property_name} contains html: '{property_value}'."
+            )
 
     return messages
 
--- a/src/openforms/tests/e2e/input_validation_base.py
+++ b/src/openforms/tests/e2e/input_validation_base.py
@@ -21,6 +21,8 @@ from openforms.typing import JSONValue
 
 from .base import E2ETestCase, browser_page
 
+omit = object()
+
 
 def create_form(component: Component) -> Form:
     _test_name = f"{component['type']} - {component['label']}"[:50]
@@ -90,7 +92,9 @@ class ValidationsTestCase(SubmissionsMixin, E2ETestCase):
 
         with self.subTest("backend validation"):
             api_value = kwargs.pop("api_value", ui_input)
-            self._assertBackendValidation(form, component["key"], api_value)
+            self._assertBackendValidation(
+                form, component["key"], component["type"], api_value
+            )
 
     def _locate_input(self, page: Page, label: str):
         return page.get_by_label(label, exact=True)
@@ -121,7 +125,9 @@ class ValidationsTestCase(SubmissionsMixin, E2ETestCase):
 
             await expect(page.get_by_text(expected_ui_error)).to_be_visible()
 
-    def _assertBackendValidation(self, form: Form, key: str, value: JSONValue):
+    def _assertBackendValidation(
+        self, form: Form, key: str, type: str, value: JSONValue
+    ):
         submission = SubmissionFactory.create(form=form)
         step = form.formstep_set.get()
         self._add_submission_to_session(submission)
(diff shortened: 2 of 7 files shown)


robinmolen:
Note to self: If an existing map component has a `defaultValue` (like empty array, or something), this would be used as empty value for the check-logic call. Which could still cause issues.

A migration and possible fix script should be included in this change

bin/report_component_problems.py
```diff
@@ -68,6 +68,11 @@ def check_component(component: Component) -> str | None:
                 if bool(translation_dict.get("defaultValue")):
                     return "defaultValue has a translation"
 
+        case {"type": "map"}:
+            default_value = component.get("defaultValue")
+            if default_value is not None:
+                return f"Default value '{default_value}' is not valid."
```
sergei-maertens, line 74: if the default value is `{"type": "Point", "coordinates": [...]}` it would be valid though, right?

src/openforms/tests/e2e/input_validation_base.py
```diff
@@ -133,7 +137,7 @@ def _assertBackendValidation(self, form: Form, key: str, value: JSONValue):
                 "step_uuid": step.uuid,
             },
         )
-        body = {"data": {key: value}}
+        body = {"data": {key: value}} if value is not None else {"data": {}}
```
sergei-maertens, line 140: this will cause issues because for number fields `null` (and thus `None` in Python) is valid and should not be omitted.

sergei-maertens (changes requested):

bin/report_component_problems.py
robinmolen: Euh.. yeah check.. Would be weird to set a default value for a map (I think this would also have to be done via the json editor..)

But i'll update this to allow geojson default values

sergei-maertens: I don't think it's that unexpected if a default value would be set in the builder, but more realistically would be a default value set (dynamically) via prefill. Though that wouldn't show up in the component configuration and is fine :)

robinmolen: Mhh, okay...

I've added a check using the Geometry serializer. It might be a bit over-kill, but it makes sure that the correct `type` and `coordinates` combination is used (seeing how the `coordinates` format is different for each geometry type)

It could be simplified to just checking if a valid `type` is used and if `coordinates` is a list, but then we would lose quite some precision (for example, that `coordinates` is an array (or array of arrays) with 2 float values)

src/openforms/tests/e2e/test_input_validation.py
```diff
@@ -561,6 +561,7 @@ def test_required_field(self):
         self.assertValidationIsAligned(
             component,
             ui_input="",
+            api_value=None,
```
sergei-maertens, line 564: I still don't find this intuitive and the inconsistent behaviour between tests is confusing. Additionally, the component type specific check in the generic base class is a smell and exactly where Formio.js went wrong in their SDK with their `Component` base class.

We can make this explicit quite easily - if we define a singleton in the base module and use that, we effectively decouple it while making it very readable:

in `input_validation_base.py`

```py
omit = object()

...

def _assertBackendValiation(self, ...):
    body_data = {} if value is omit else {key: value}
    body = {"data": body_data}
    ...
```

then here:

```py
from .input_validation_base import omit
```

and
```suggestion
            api_value=omit,
```

bin/report_component_problems.py
```diff
@@ -68,6 +71,27 @@ def check_component(component: Component) -> str | None:
...
+                    raise_exception=True
+                ):
+                    # This object has a valid Geometry shape
+                    return None
+            except ValidationError:
+                return f"Default value '{default_value}' is not valid."
+
+            # Any other value, that doesn't adhere to the Geometry shape, is invalid
+            if default_value is not None:
+                return f"Default value '{default_value}' is not valid."
```
sergei-maertens, line 93:
```suggestion
            if isinstance(
                default_value, dict
            ) and GeoJsonGeometryPolymorphicSerializer(data=default_value).is_valid()
                # This object has a valid Geometry shape
                return None
            elif default_value is not None:
                return f"Default value '{default_value}' is not valid."
```

sergei-maertens (changes requested):

src/openforms/tests/e2e/test_input_validation.py
robinmolen: Ah yeah thats a lot clearer

bin/report_component_problems.py
robinmolen: Won't this just break if a map config isn't set right? I mean, it will report some sort of problem (i.e. the report breaks because component X), but it's not very nice.. (and if there are multiple components with the same problem, you will have to `run report` > `fix component` > `run report` > `fix component` > ...)

sergei-maertens: yeah there's a branch missing for when the serializer is not valid, so we should capture the result of `is_valid`.

I generally don't like the exception-based flow with serializers if you immediately catch it again, you can just use the boolean return value of `is_valid()` instead

robinmolen: Ahh i see the problem..

Yeah but the serializer will still throw an error if using `is_valid()`. For this map defaultValue: 

```
{
    "type": "Point",
    "coordinates": [
      4,
      4,
      5
    ]
  }
```

It throws this error:

```
KeyError: "`GeoJsonGeometryPolymorphicSerializer.serializer_mapping` is missing a corresponding serializer for the `<class 'rest_framework.fields.empty'>` key"
```

So i think we need to throw the error via the `is_valid(raise_exception=True)` and then catch it, right? And the message should be something that reflects the issue of the serializer (i.e. the validation error message)

robinmolen: Further discussed via slack, seems to be an issues originating from https://github.com/maykinmedia/drf-polymorphic/blob/816653385c54cb8e28dfaa34eec8ffa1dbb1e8b1/drf_polymorphic/serializers.py#L124.
Decided to throw and catch the error here, and pass the Validation message to the error messages list

sergei-maertens (approved):

PR #5173, opened 2025-03-21
add registrator omschrijving

olenitsj:
Changes

Added the <ZKN:omschrijving> element to the heeftAlsOverigBetrokkene block to explicitly indicate the role of the related medewerker. This enables generic mapping of the omschrijving field to roles in external systems like XXLNC Zaken.

Previously, it was implicitly assumed that a medewerker related via heeftAlsOverigBetrokkene was always the registrator. With this change, the role is explicitly set using the omschrijving field (<ZKN:omschrijving>registrator</ZKN:omschrijving>), allowing for more flexible and accurate role mapping, especially in cases where the medewerker fulfills a different function within the case.

### DIFF
--- a/src/stuf/stuf_zds/templates/stuf_zds/soap/creeerZaak.xml
+++ b/src/stuf/stuf_zds/templates/stuf_zds/soap/creeerZaak.xml
@@ -160,6 +160,7 @@
                     <ZKN:identificatie>{{ registrator.medewerker.identificatie }}</ZKN:identificatie>
                 </ZKN:medewerker>
             </ZKN:gerelateerde>
+            <ZKN:omschrijving>registrator</ZKN:omschrijving>
         </ZKN:heeftAlsOverigBetrokkene>
         {% endif %}
         {% if zds_zaaktype_status_code or zds_zaaktype_status_omschrijving %}


src/stuf/stuf_zds/templates/stuf_zds/soap/creeerZaak.xml
```diff
@@ -160,6 +160,7 @@
                     <ZKN:identificatie>{{ registrator.medewerker.identificatie }}</ZKN:identificatie>
                 </ZKN:medewerker>
             </ZKN:gerelateerde>
+            <ZKN:omschrijving>registrator</ZKN:omschrijving>
```
sergei-maertens, line 163: Where can I find a reference of possible values for this omschrijving? What if another municipality/StUF-ZDS vendor uses a different value for this?

sergei-maertens (commented):
Can you also please create an issue for this PR so that in the future we can track down what functional requirement initiated this change?

src/stuf/stuf_zds/templates/stuf_zds/soap/creeerZaak.xml
olenitsj: It's a free field. So there is no enum with all the options. xxllnc zaken has some predefined roles, and we map to them. But there is always a translation layer, as there is no predefined list.

sergei-maertens: yes, so anyone not using excellence wanting similar behaviour would likely want different values here, correct?

olenitsj: Correct, but i think it is up to you to define the value of this field. It just should describe the object you are adding. In this case the xml suggests its a registrator. Registrator is something maykin has thought of not xxllnc. We will be mapping it to the value we need. 

It's simpel, just specify the rol like you did with cosign. Right now its 'betrokkene medewerker' but is it the one who picked up the phone, is it the one who registerd the case, is it a internal form where you just refere to your co-worker. So the relation ship is not clear from the xml right now. Bij filling in the omschrijving field with registrator. We will know this specific employee is the one who registered the case. Whatever we call it inside xxllnc its up to us. But its up to openforms to make clear what the employee is.

sergei-maertens: > It's simpel, just specify the rol like you did with cosign

mede_initiator is an actual betrokkene type defined by VNG in standards though, so that's a slightly different case. However I do understand that his is also quite "arbitrary".

I'm trying to be thorough here in the spirit of using existing standards and avoiding new dialects to emerge. So I'm not sure if the value of this field should be something configurable or a hardcoded value. We'll discuss in Monday in the refinement with the rest of the team!

olenitsj: If possible, configurable is always preferred—though of course, that’s up to you.

This change was triggered by my comment here:
https://github.com/open-formulieren/open-forms/issues/4944

You’ll also need to define a child/partner as a betrokkene type.
I haven’t looked into how configurable your betrokkene entities are, but if the omschrijving field can be made configurable for roles like child/partner/mede-ondertekenaar/registrator/etc., that would help.

That said, it would likely require exposing some configuration options per betrokkene in both the data model and the beheerder UI. Feels like a larger change, but perhaps it can be scoped within the Familieleden-component, since you’ll need to address it there anyway.

PR #5182, opened 2025-03-25
Remove `glom` usage from `FormioData`

### LINKED ISSUE #5179: Remove `glom` usage from `FormioData`
Previous fixes (#4768 and #5118) have shown that the usage of `glom` in `FormioData` - especially for large forms - can cause significant performance issues. This is because `glom` does a lot more than just enabling nested-data access, see the [documentation](https://glom.readthedocs.io/en/latest/#glom).

However, the only thing `FormioData` needs to do is to enable this nested-data access through the use of keys with a period. For example:
```python
>>> data = FormioData({"foo": {"bar": "baz"}})
>>> print(data["foo.bar"])
baz
>>> data["nested.key"] = "value"
>>> print(data)
{"foo": {"bar": "baz"}, "nested": {"key": "value"}}
```

Suggestion is to remove the usage of `glom` from `FormioData` and add custom implementations in `__setitem__` and `__getitem__` that enable nested-data access.

Question still to discuss and/or think about: what do we want `FormioData.items()` to return?
* Flat data structure: `dict_keys([foo.bar]), dict_values([baz])`
* Top-level: `dict_keys([foo]), dict_values([{"bar": "baz"}])`
* More `FormioData` instances: `dict_keys([foo]), dict_values([FormioData({"bar": "baz"})])`

viktorvanwijk:
Closes #5179

**Changes**

Remove usage of `glom` from `__getitem__` and `__setitem__` in `FormioData`

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/formio/datastructures.py
+++ b/src/openforms/formio/datastructures.py
@@ -2,9 +2,9 @@ import re
 from collections import UserDict
 from typing import Iterator, cast
 
-from glom import PathAccessError, assign, glom
+from glom import glom
 
-from openforms.typing import DataMapping, JSONValue
+from openforms.typing import DataMapping, VariableValue
 
 from .typing import Component, EditGridComponent, FormioConfiguration
 from .utils import flatten_by_path, is_visible_in_frontend, iter_components
@@ -157,33 +157,78 @@ class FormioData(UserDict):
             data[component["key"]] = ...
 
     without having to worry about potential deep assignments or leak implementation
-    details (such as using ``glom`` for this).
+    details.
     """
 
-    data: dict[str, JSONValue]
-    _keys: set[str]
-    """
-    A collection of flattened key names, for quicker __contains__ access
-    """
+    data: dict[str, VariableValue]
 
-    def __init__(self, *args, **kwargs):
-        self._keys = set()
-        super().__init__(*args, **kwargs)
+    def __getitem__(self, key: str) -> VariableValue:
+        """
+        Get a value from the internal data dict.
 
-    def __getitem__(self, key: str):
+        Keys are expected to be strings and can indicate nested data, e.g.
+        ``variable.key``.
+        """
+        assert isinstance(key, str)
+
         if "." not in key:
             return self.data[key]
-        try:
-            return cast(JSONValue, glom(self.data, key))
-        except PathAccessError as exc:
-            raise KeyError(f"Key '{key}' is not present in the data") from exc
 
-    def __setitem__(self, key: str, value: JSONValue):
+        value = self.data
+        raise_error = False
+        for k in key.split("."):
+            if isinstance(value, dict):
+                try:
+                    value = value[k]
+                except KeyError:
+                    raise_error = True
+            elif isinstance(value, list):
+                try:
+                    value = value[int(k)]
+                except (ValueError, IndexError):
+                    raise_error = True
+            else:
+                raise_error = True
+
+            if raise_error:
+                raise KeyError(f"Key '{key}' is not present in the data")
+
+        return value
+
+    def __setitem__(self, key: str, value: VariableValue):
+        """
+        Set a value to the internal data dict.
+
+        Keys are expected to be strings and can indicate nested data, e.g.
+        ``variable.key``.
+        """
+        assert isinstance(key, str)
+
         if "." not in key:
             self.data[key] = value
-        else:
-            assign(self.data, key, value, missing=dict)
-        self._keys.add(key)
+            return
+
+        data = self.data
+        key_list = key.split(".")
+        for k in key_list[:-1]:
+            if isinstance(data, dict):
+                child = data.get(k, None)
+            elif isinstance(data, list):
+                try:
+                    k = int(k)
+                    child = data[k]
+                except (ValueError, IndexError):
+                    raise KeyError(f"Cannot set an item in a list on index '{k}'")
+            else:
+                raise AttributeError(f"Item '{data}' has no attribute '{k}'")
+
+            if not isinstance(child, (dict, list)):
+                # TODO-5179: should this be configurable?
+                data[k] = {}
+
+            data = data[k]
+
+        data[key_list[-1]] = value
 
     def __contains__(self, key: object) -> bool:
         """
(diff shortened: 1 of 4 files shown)


viktorvanwijk:
# Performance test result

Performance is tested using `FormioDataTests.test_performance`. Below are the results per method.

Result interpretation:
- Current: the current implementation (`FormioData`)
- Full glom: the current implementation without the ‘bandaid’ fixes (`FormioDataWithoutGlom`)
- Custom: the new implementation, presented by the code changes shown here (`FormioDataCustom`)
- Control: regular dict lookup with flat structure (`{“foo.bar”: “baz”}`)

### `__getitem__`
```python
    def __getitem__(self, key: str):
        # TODO-5179: do we want to check the key here as well?
        # self._check_key(key)

        value = self.data
        for k in key.split("."):
            if not isinstance(value, dict):
                raise KeyError(f"Key '{key}' is not present in the data")
            try:
                value = value[k]
            except KeyError:
                raise KeyError(f"Key '{key}' is not present in the data")

        return value
```
image

Comparing 'Current' to 'Custom', we can see a significant increase in performance for nested keys. Note that if we want the non-nested key lookup to be as fast as the current implementation, we need to keep the "bandaid fix":
```python
if "." not in key:
    return self.data[key]
```

### `__setitem__` (for loop)
```python
```

sergei-maertens (changes requested):
Thanks for the thorough analysis, some thoughts:

**Hot code path**

Arguably the non-nested key access (`"foo"` rather than `"foo.bar"`) is the most used in forms, so I'd opt for keeping the "bandaid" fix as an extra optimization. From your benchmarks, not doing it makes the code roughly twice as slow (+100% execution time), and that may add up in large forms. The maintenance overhead of this extra check is minimal and properly contained in the data structure, so I'd reckon this is a good tradeoff.

**Key type checking**

There's now an extra method call `self._check_key(key)` which causes some overhead and slowdown, and I think it mostly just provides developer type safety. I haven't seen any crashes in Sentry. So, we have the `isinstance` check + the method call overhead, and I think we can avoid it:

* short term: change it to an inline `assert isinstance(key, str)` - this enforces it at runtime during development and will be caught in the test suite if people do silly stuff, but allows us to use the [uwsgi optimize](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#optimize) option to strip out those `asserts` in production (same as using the `python -O` option).

* long term: type checking with pyright! we have the type annotation for the key already, and if we get all of our code accessing FormioData properly type checked, this will also be caught before it causes problems and the beauty is that it has zero runtime impact.

**Recursion vs for-loop**

Agree with your conclusion, and it's more obvious if we do the hot-code path optimization.

**Missing cases**

I think one case that we are missing that `glom` covers is index access, e.g. `foo.0.bar`. In Formio itself, this actually still leads to nested dictionaries, but those access paths are ultimately what's driving data access in edit grids - a repeating group with key `foo` has items (accessed by index) which themselves are objects/dicts with keys (`bar). Though this is probably not relevant - I can only imagine some JSON logic triggers with paths to a particular item (which in itself is kind of weird), and that has its own data resolution mechanism separate from `FormioData`.

src/openforms/formio/datastructures.py
```diff
@@ -157,7 +157,7 @@ class FormioData(UserDict):
             data[component["key"]] = ...
 
     without having to worry about potential deep assignments or leak implementation
-    details (such as using ``glom`` for this).
+    details.
     """
 
     data: dict[str, JSONValue]
```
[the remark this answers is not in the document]
viktorvanwijk: Yeah it should be extended to include the Python types :)

viktorvanwijk:
> Arguably the non-nested key access ("foo" rather than "foo.bar") is the most used in forms, so I'd opt for keeping the "bandaid" fix as an extra optimization. From your benchmarks, not doing it makes the code roughly twice as slow (+100% execution time), and that may add up in large forms. The maintenance overhead of this extra check is minimal and properly contained in the data structure, so I'd reckon this is a good tradeoff.

Yeah, fair enough. I was doubting a little if we wanted to get rid of this "bandaid", so I left it out for now. But the performance gain is significant :zap:

> short term: change it to an inline assert isinstance(key, str) - this enforces it at runtime during development and will be caught in the test suite if people do silly stuff, but allows us to use the [uwsgi optimize](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#optimize) option to strip out those asserts in production (same as using the python -O option).

TIL something :)

> I think one case that we are missing that glom covers is index access, e.g. foo.0.bar. In Formio itself, this actually still leads to nested dictionaries, but those access paths are ultimately what's driving data access in edit grids - a repeating group with key foo has items (accessed by index) which themselves are objects/dicts with keys (bar). Though this is probably not relevant - I can only imagine some JSON logic triggers with paths to a particular item (which in itself is kind of weird), and that has its own data resolution mechanism separate from FormioData

Yeah I wasn't sure if this was necessary at the moment, so I left it out. While working on #5221, however, I have found a place where this is used: `ComponentNode.value` and `ComponentNode.is_visible`. They both currently still have raw `glom` usage. Would be nice if we can incorporate this inside `FormioData` then :)

viktorvanwijk:
# Performance results after feedback
image


image

NOTE: during testing, I have noticed variations of up to 0.01 µs when running the same test multiple times, so please take this into consideration when looking at the numbers for non-nested items

src/openforms/formio/tests/test_datastructures.py
```diff
@@ -9,7 +9,7 @@ class FormioDataTests(TestCase):
     def test_mimicks_dict_interface(self):
         formio_data = FormioData({"foo": "bar"})
 
-        self.assertEqual(formio_data, {"foo": "bar"})
```
viktorvanwijk, line 12: I have found usages of `FormioData.items` :)

src/openforms/formio/tests/test_datastructures.py
```diff
@@ -9,7 +9,7 @@ class FormioDataTests(TestCase):
...
+        self.assertEqual(formio_data.data, {"foo": "bar"})
```
viktorvanwijk, line 12: I have found usages of FormioData.items :)

viktorvanwijk:
I started to work on a `Path` class, similar to the one that `glom` has, but we *might* be able to get away with just the list indexing in strings, e.g. `'foo.0.bar'`. Can't tell for sure until I've replaced all the `DataMapping` instances with `FormioData`, so I've left it out for now. It can be added later in #5228 if necessary.
```python
class Path:
    def __init__(self, *args) -> None:
        self.values = ()
        for k in args:
            if isinstance(k, str):
                self.values = self.values + tuple(k.split("."))
            elif isinstance(k, int):
                self.values = self.values + (k,)
            elif isinstance(k, Path):
                self.values = self.values + k.values
            else:
                raise NotImplementedError(f"Unsupported type {type(k)}")

    def __getitem__(self, index: int):
        return self.values[index]

    def __iter__(self):
        yield from self.values

    def __str__(self):
        return str(self.values)

    def __add__(self, other):
        if isinstance(other, Path):
            return Path(self.values + other.values)
```

sergei-maertens (approved):

PR #5183, opened 2025-03-25
[#5178] Do not show expired referentielijsten data to the end user

### LINKED ISSUE #5178: Expired items should not show up in the list (populated from reference lists) and the admin should be notified if all items expired.
### Thema / Theme

Frontend

...or if the table expired.

* [ ] Expired items should not show up in the list (public)
* [ ] If all items expired or the table expired, the admin should get a message in digest email

_No response_

_No response_

vaszig:
Closes #5178

**Changes**

- Check for expired items and remove them from the rendered options
- Add email digest function for informing admin users about expiring (or expired) tabels and items

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/contrib/referentielijsten/helpers.py
```diff
@@ -0,0 +1,10 @@
+from datetime import datetime, timedelta
+
+from django.utils import timezone
+
+
+def tabel_or_item_has_expired(einddatum: str) -> bool:
```
sergei-maertens, line 6:
```suggestion
def check_expired(einddatum: str) -> bool:
```
there's nothing that ties this to a table or item, so we can use a more generic function name. Also, if it returns a boolean, it's nice to use a `is_` prefix or use `check_` in the name to hint about the return type.

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -154,3 +154,44 @@ <h5>{% trans "Logic rules problems" %}</h5>
...
+
+{% if expired_or_near_expiry_referentielijsten_data %}
+    <h4>{% trans "Referentielijsten (tabels-items expired or near expiry)" %}</h4>
+    <ul>
+        {% for service in expired_or_near_expiry_referentielijsten_data %}
+            {% blocktranslate with service_label=service.service %}
+                <h4>Service: {{ service_label }}</h4>
+            {% endblocktranslate %}
+            
+            <!-- Expiring tabels -->
```
sergei-maertens, line 166:
```suggestion
            <!-- Expiring tables -->
```
please use the English "table" :)

src/openforms/emails/templates/emails/admin_digest.html
```diff
@@ -154,3 +154,44 @@ <h5>{% trans "Logic rules problems" %}</h5>
...
+            {% if service.tabels %}
+                <h4>{% trans "Tabels" %}</h4>
+                {% for tabel in service.tabels %}
+                    <li>
+                        <p>
+                            {% blocktranslate with tabel_naam=tabel.naam eindatum=tabel.eindatum|date trimmed %}
+                            Tabel '{{ tabel_naam }}' expires (or expired) in {{ eindatum }}.<br>
```
sergei-maertens, line 173:
```suggestion
                            Table '{{ tabel_naam }}' expires (or expired) in {{ eindatum }}.<br>
```
we'll handle the Dutch translations in another commit.

Additionally, we have the information to check if it has expired or will expire soon, so let's use proper messages for that by combining it with the `timesince` and `timeuntil` template filters - it's much more readable to see "will expire in 4 days" or "has expired 7 days ago" than having to compare a date with the current date.

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -48,18 +49,26 @@ def check_config(self):
...
-    "openforms.emails.tasks.GlobalConfiguration.get_solo",
-    return_value=GlobalConfiguration(
-        recipients_email_digest=["tralala@test.nl", "trblblb@test.nl"]
-    ),
-)
 @override_settings(LANGUAGE_CODE="en")
 class EmailDigestTaskIntegrationTests(TestCase):
 
+    def setUp(self) -> None:
+        self.config = GlobalConfiguration.get_solo()
```
sergei-maertens, line 56: please always call the `super().setUp()` method for possible future mixin behaviour.

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -48,18 +49,26 @@ def check_config(self):
...
+        self.config.recipients_email_digest = ["tralala@test.nl", "trblblb@test.nl"]
+        self.config.save()
```
sergei-maertens, line 58:
```suggestion
        self.config.save()
        self.addCleanup(GlobalConfiguration.clear_cache)
```
let's avoid broken cache isolation leading to flaky tests

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -48,18 +49,26 @@ def check_config(self):
...
+
+        self.referentielijsten_service = ServiceFactory.create(
```
sergei-maertens, line 60: this attribute is not used anywhere, right? Please only expose variables on `self` if they're actually used, otherwise just use a local variable to prevent namespace polution.

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -48,18 +49,26 @@ def check_config(self):
...
+            api_root="http://localhost:8004/api/v1/",
```
sergei-maertens, line 61: can you add a comment and point to the correct `docker` compose file in the repository?

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -48,18 +49,26 @@ def check_config(self):
...
+            slug="referentielijsten",
+            auth_type=AuthTypes.no_auth,
+        )
+        self.config.referentielijsten_services.add(self.referentielijsten_service)
+
+    @patch("openforms.emails.tasks.GlobalConfiguration.get_solo")
     def test_that_repeated_failures_are_not_mentioned_multiple_times(
         self, mock_global_config
     ):
+        mock_global_config.return_value = self.config
```
sergei-maertens, line 71: mocking the config is now pointless here as we are updating the actual real config in tests anyway. The mocking is a mechanism to ensure we don't accidentally mess with the global django-solo cache. Best to just remove the mocks here.

src/openforms/formio/dynamic_config/referentielijsten.py
```diff
@@ -14,7 +15,7 @@
 
 def fetch_options_from_referentielijsten(
     component: Component, submission: Submission
-) -> list[tuple[str, str]] | None:
+) -> list[list[str, str]] | None:
```
sergei-maertens, line 18: let's make the returned data actually tuples instead. A list implies arbitrary length, while a tuple implies a known/fixed length of each item. The type annotation is correct, the implementation not.

Can you check if we can add this file to the type checker config so that it's checked in CI in the future?

src/openforms/formio/dynamic_config/tests/test_referentielijsten_config.py
```diff
@@ -424,6 +453,36 @@ def test_items_not_found(self):
...
+            "openForms": {
+                "code": "not_geldig_option",
+                "dataSrc": DataSrcOptions.referentielijsten,
+                "service": self.service.slug,
+                "translations": {},
+            },
+        }
+
+        with self.assertRaises(ServiceUnavailable):
+            register.update_config(component, submission=self.submission, data={})
```
sergei-maertens, line 472: Is this correct behaviour? This was intended to be returned if the reference lists service is broken, but if it's actually up and running and just returns expired items, then I would expect the form to just be available as usual and not return a 503

sergei-maertens (changes requested):

src/openforms/contrib/referentielijsten/helpers.py
```diff
@@ -0,0 +1,10 @@
+from datetime import datetime, timedelta
+
+from django.utils import timezone
+
+
+def check_expired_or_near_expiry(einddatum: str) -> bool:
+    parsed_datetime = datetime.fromisoformat(einddatum)
+    return (
+        timezone.now() <= parsed_datetime <= timezone.now() + timedelta(days=7)
+    ) or (parsed_datetime < timezone.now())
```
sergei-maertens, line 10: simpler:

```py
in_seven_days = timezone.now() + timedelta(days=7)
return not (parsed_datetime > in_seven_days)
```

src/openforms/emails/digest.py
```diff
@@ -492,3 +523,130 @@ def _report(var_key: str = "", exception: bool = False, rule_index: int = 0):
...
+    for referentielijsten_service in config.referentielijsten_services.all():
+        with build_client(
+            referentielijsten_service, client_factory=ReferentielijstenClient
+        ) as client:
+            try:
+                tabels = client.get_tabellen()
+            except RequestException:
+                tabels = []
+
+            service_label = referentielijsten_service.label
```
sergei-maertens, line 541: why are we checking tables and items that may not even be used in any form?

We should only check the form definitions of (active) forms that make use of reference lists, check only those services and tables and detect if there's stuff (about to) expire(d). The reference can have expired tables, but as long as they aren't used, everything is fine.

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -210,6 +214,64 @@ def test_email_sent_when_there_are_failures(
                 "https://api.brk.kadaster.nl/invalid/kadastraalonroerendezaken?postcode=1234AB&huisnummer=1",
                 status_code=400,
             )
+            m.get(
```
sergei-maertens, line 217: why doesn't this use VCR? You can use `freeze_time` to pin our own timestamps, no?

src/openforms/formio/dynamic_config/referentielijsten.py
```diff
@@ -6,14 +6,14 @@
...
 from openforms.logging import logevent
 from openforms.submissions.models import Submission
-
-from ..typing import Component
+from openforms.typing import JSONObject
 
 
 def fetch_options_from_referentielijsten(
-    component: Component, submission: Submission
+    component: JSONObject, submission: Submission
```
sergei-maertens, line 16: that's a wrong type - it's supposed to represent a Formio component

src/openforms/formio/dynamic_config/referentielijsten.py
```diff
@@ -68,4 +68,14 @@ def fetch_options_from_referentielijsten(
...
+
+        valid_items = []
+        for item in result:
+            if (
+                einddatum := item.get("einddatumGeldigheid")
+            ) and check_expired_or_near_expiry(einddatum):
+                continue
+
+            valid_items.append(item)
+        return [(item["code"], item["naam"]) for item in valid_items]
```
sergei-maertens, line 81:
```suggestion
        return [
            (item["code"], item["naam"]) for item in valid_items
            if not (einddatum := item.get("einddatumGeldigheid")) or not check_expired_or_near_expiry(einddatum)
        ]
```
either this or just push the tuples directly onto `valid_items`, now you're looping twice over a list of items which could be done in a single pass

sergei-maertens (changes requested):

src/openforms/formio/dynamic_config/referentielijsten.py
vaszig: This was failing in pyright that's why I changed it. (Basically all the logs have `component: JSONObject` and then I would have to change them all-don't know if it is what we want)

sergei-maertens: Let's just ignore it for now with a `# pyright: ignore[<errorType>]` comment and add a TODO to clean up the ignore

src/openforms/emails/digest.py
```diff
@@ -492,3 +525,165 @@ def _report(var_key: str = "", exception: bool = False, rule_index: int = 0):
...
+                        try:
+                            response = client.get(
+                                "tabellen",
+                                params={"code": component["openForms"]["code"]},
+                            )
+                        except RequestException:
+                            response = []
+
+                        service_label = service.label
+                        tabels = list(pagination_helper(client, response.json()))
```
sergei-maertens, line 562:
```suggestion
                        tables = list(pagination_helper(client, response.json()))
```
please, English in code.

src/openforms/emails/digest.py
```diff
@@ -492,3 +525,165 @@ def _report(var_key: str = "", exception: bool = False, rule_index: int = 0):
...
+                                                        ),
+                                                        expired=datetime.fromisoformat(
+                                                            tabel_eindatum
+                                                        )
+                                                        <= timezone.now(),
+                                                    )
+                                                ]
+                                            }
+                                        }
+                                    )
```
sergei-maertens, line 602: Use a `defaultdict` for this to avoid the member test

src/openforms/emails/digest.py
```diff
@@ -492,3 +525,165 @@ def _report(var_key: str = "", exception: bool = False, rule_index: int = 0):
...
+                                                    naam=item["naam"],
+                                                    eindatum=datetime.fromisoformat(
+                                                        item_eindatum
+                                                    ),
+                                                    expired=datetime.fromisoformat(
+                                                        item_eindatum
+                                                    )
+                                                    <= timezone.now(),
+                                                )
+                                            )
```
sergei-maertens, line 655: Use `defaultdict`

src/openforms/emails/tests/test_tasks_integration.py
```diff
@@ -31,10 +32,13 @@
...
 from openforms.submissions.models.submission import Submission
 from openforms.submissions.tests.factories import SubmissionFactory
+from openforms.utils.tests.vcr import OFVCRMixin
 
 from ..tasks import send_email_digest
 
-TEST_FILES = Path(__file__).parent / "data"
+TESTS_DIR = Path(__file__).parent.resolve()
+TEST_FILES = TESTS_DIR / "data"
+TEST_FILES_VCR = TESTS_DIR / "files"
```
sergei-maertens, line 41: just use the same directory, we don't have consistency anyway

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #5197, opened 2025-03-26
Fix broken demo prefill plugin

### LINKED ISSUE #5185: Vague warning when using the `demo` prefill plugin.
Image

-> the required attributes don't make sense, might be good to suppress this in the UI as it's an edge case?

This also broke the demo plugin itself from functioning, as nothing is pre-filled any longer 😬

sergei-maertens:
Closes #5185

**Changes**

Added all the available auth attributes so that the plugin always runs.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/prefill/contrib/demo/plugin.py
+++ b/src/openforms/prefill/contrib/demo/plugin.py
@@ -5,6 +5,7 @@ from typing import Any
 from django.utils.crypto import get_random_string
 from django.utils.translation import gettext_lazy as _
 
+from openforms.authentication.constants import AuthAttribute
 from openforms.config.data import Action
 from openforms.submissions.models import Submission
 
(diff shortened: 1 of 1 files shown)


viktorvanwijk (commented):
I have used this demo prefill before and did not encounter any issues with it (other than the vague warning). When did it break for you specifically?

viktorvanwijk:
But won't this also fail for `ObjectsAPIPrefill` then, as `requires_auth` is empty there as well? Seems like we need to fix the line you linked instead of adding all possible authentication attributes, or am I not understanding it correctly?

sergei-maertens:
> But won't this also fail for `ObjectsAPIPrefill` then, as `requires_auth` is empty there as well? Seems like we need to fix the line you linked, and remove the warning for prefills that don't require authentication. Or am I not understanding it correctly?

No, it uses a different code path, using `fetch_prefill_values_from_options`. The plugin + attribute prefill uses `fetch_prefill_values_from_attribute` which caused this breakage.

viktorvanwijk (approved):

PR #5209, opened 2025-03-27
Fix validation crash when a map component value is cleared/empty

### LINKED ISSUE #5191: Map component tests 3.1.0
## Bug with clearOnHide

- [x] `clearOnHide` interaction validation crash

There seems to be a bug with the 'clearOnHide' property set on a map component. To reproduce:
* Create a new form and form step
* Add map component with clearOnHide property set
* Add radio component with options 'A' and 'B'
* Add logic rule that hides the map component when the radio is set to 'A'

When filling out the form:
* Select a point, or draw a line or polygon on the map
* Select radio 'A' to hide the map
* Navigate to the next page
* Observe a form crash

<details>
    <summary>Traceback in the console</summary>

```
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/django/views/decorators/cache.py", line 62, in _wrapper_view_func
    response = view_func(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/src/openforms/utils/decorators.py", line 23, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/viewsets.py", line 124, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/src/openforms/submissions/api/viewsets.py", line 618, in validate
    self._validate_step_input(request)
  File "/home/viktor/open-forms/src/openforms/submissions/api/viewsets.py", line 630, in _validate_step_input
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/serializers.py", line 223, in is_valid
    self._validated_data = self.run_validation(self.initial_data)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/serializers.py", line 442, in run_validation
    value = self.to_internal_value(data)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/serializers.py", line 501, in to_internal_value
    validated_value = validate_method(validated_value)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/src/openforms/submissions/api/serializers.py", line 281, in validate_data
    self._run_formio_validation(data)
  File "/home/viktor/open-forms/src/openforms/submissions/api/serializers.py", line 305, in _run_formio_validation
    step_data_serializer.is_valid(raise_exception=True)
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/serializers.py", line 223, in is_valid
    self._validated_data = self.run_validation(self.initial_data)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/serializers.py", line 442, in run_validation
    value = self.to_internal_value(data)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/serializers.py", line 499, in to_internal_value
    validated_value = field.run_validation(primitive_value)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/drf_polymorphic/serializers.py", line 101, in run_validation
    extra_serializer = self._get_serializer_from_data(data)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/drf_polymorphic/serializers.py", line 124, in _get_serializer_from_data
    discriminator_value = self.discriminator.get_value(data)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/rest_framework/fields.py", line 429, in get_value
    return dictionary.get(self.field_name, empty)
           ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

</details>

## Other bugs (or possibly intended behaviour?)
* [x] (form builder) Sometimes, when deselecting a drawing shape while multiple were selected, all options disappear from the map preview. The options are saved correctly, though. Difficult to consistently reproduce this, unfortunately. edit: Sergei was able to reproduce this in Brave when all three options are enabled and the top option is checked/unchecked repeatedly https://github.com/open-formulieren/formio-builder/pull/222
* [ ] (SDK) When drawing a line or polygon, and only a first point has been placed, clicking "Verwijder laatste punt" does nothing. Is this intended behaviour? edit: Confirmed unintuitive, not a release blocker.
*This is a limitation/behaviour in Leaflet-draw. Will have to discuss what and if we want to invest in this (also the question what we can actually do...). See [Leaflet-draw polyline draw handler](https://github.com/Leaflet/Leaflet.draw/blob/d5dd11781c2e07f9e719308e504fe579000edf54/src/draw/handler/Draw.Polyline.js#L156)*
* [X] ~(SDK) When a shape has been drawn, and the trash can is clicked, there are missing translations for the tooltip (I know translations are part of the release, just listing it here for completeness). edit: not a release blocker
Image open-formulieren/open-forms-sdk#820~
The trash icon will no-longer open a menu (see next point), so these missing translations aren't present anymore.
* [x] (SDK) When clicking "Opslaan" after drawing a shape, it just seems to remove the shape without saving it. Is there a way to get the saved changes back that I haven't found, or is it broken? Also, placing a save button under a trashcan icon seems a bit counterintuitive to me. edit: might be better to relabel this to "confirm", but it's odd anyway since there's only the single shape you can draw anyway. It may make more sense if you can draw multiple shapes, which we don't support. Not a release blocker.
~https://github.com/open-formulieren/open-forms-sdk/pull/821 will change the behaviour of the delete button, in this PR the delete menu will no-longer be shown. Instead, the shapes are immediately deleted from the map. Besides this, we cannot (easily) change the functional behaviour, or which delete menu options are shown, as this is Leaflet-draw core functionality. So if we would like some other delete functionality, we would have to create our own custom delete function.~
https://github.com/open-formulieren/open-forms-sdk/pull/872 provides a simple solution by overwriting the onclick behaviour of the Leaflet-Draw "delete" control.

## Things that were probably already there before #2177 and #2178 
* [x] ~~The tile layer default option is not shown in the component options~~ It's using the default fallback, for backwards compatibility. wontfix
* [x] ~~The initial value of the zoom level is not shown in the component options~~ global configuration defaults are used as fallback, wontfix.
* [x] (form builder) The zoom level has no help text indicating valid values in the component options. Edit: document range between 1-20, but the range may also depend on the tile layer itself (!). not a release blocker
Edit 2: the zoom level range is 1-13, and should remain as such (due to inaccuracies of the proj4 transformation starting beyond zoom level 13). Added a PR for adding a help text to the zoom level field https://github.com/open-formulieren/formio-builder/pull/224.
* [ ] Allowing the map to access your current location will cause it to constantly go back to that position after panning. This also causes the search bar to close automatically. I can only reproduce this on Firefox, with long delays (3-15 seconds). Definitely annoying, also happens on 2.8.x so not a release blocker.
This will be fixed by turning the "use my current location" behaviour into a active user action. By adding a "use my location" button to the map, the map focus point will only be shifted after actively clicking the button open-formulieren/open-forms-sdk#819
* [ ] (form builder) Entering a latitude outside the valid range causes the modal to crash -> error boundary around preview component in generic code would be good. https://github.com/open-formulieren/formio-builder/pull/223
Nice to have addition to error boundary: https://github.com/open-formulieren/formio-builder/pull/239
* [x] (SDK) Might have been the design choice for when a user could only select a point, but the summary always shows the map fully zoomed in, which doesn’t make much sense for lines and polygons. edit: not a release blocker open-formulieren/open-forms-sdk#822

sergei-maertens:
Closes #5191 (partly)

**Changes**

Fixed the underlying bug in drf-polymorphic, and updated the version.

The polymorphic serializer was added after 3.0 was released, so no backports needed for this.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/tests/test_submission_step_validate.py
+++ b/src/openforms/submissions/tests/test_submission_step_validate.py
@@ -14,6 +14,7 @@ from openforms.formio.tests.factories import (
 )
 from openforms.forms.tests.factories import (
     FormFactory,
+    FormLogicFactory,
     FormStepFactory,
     FormVariableFactory,
 )
@@ -680,6 +681,75 @@ class SubmissionStepValidationTests(SubmissionsMixin, APITestCase):
                 },
             ),
             {"data": {"fieldset": [{"fileUpload1": [file]}]}},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    @tag("gh-5191")
+    def test_validate_map_component_hidden_and_clear_on_hide(self):
+        submission = SubmissionFactory.create(
+            form__generate_minimal_setup=True,
+            form__formstep__form_definition__configuration={
+                "components": [
+                    {
+                        "type": "radio",
+                        "key": "radio",
+                        "label": "Radio",
+                        "openForms": {"dataSrc": "manual"},
+                        "values": [
+                            {"value": "a", "label": "A"},
+                            {"value": "b", "label": "B"},
+                        ],
+                    },
+                    {
+                        "type": "map",
+                        "key": "map",
+                        "label": "Map",
+                        "clearOnHide": True,
+                        "interactions": {
+                            "marker": True,
+                            "polygon": False,
+                            "polyline": False,
+                        },
+                    },
+                ]
+            },
+        )
+        FormLogicFactory.create(
+            form=submission.form,
+            json_logic_trigger={"==": [{"var": "toggle"}, "a"]},
+            actions=[
+                {
+                    "uuid": "981ebf58-4d2e-4b1f-bf6b-709a52104714",
+                    "component": "kaart",
+                    "formStepUuid": None,
+                    "action": {
+                        "type": "property",
+                        "property": {"value": "hidden", "type": "bool"},
+                        "state": True,
+                    },
+                }
+            ],
+        )
+        step = submission.form.formstep_set.get()
+        self._add_submission_to_session(submission)
+
+        endpoint = reverse(
+            "api:submission-steps-validate",
+            kwargs={
+                "submission_uuid": submission.uuid,
+                "step_uuid": step.uuid,
+            },
+        )
+
+        response = self.client.post(
+            endpoint,
+            {
+                "data": {
+                    "radio": "a",
+                    "map": None,
+                }
+            },
         )
 
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -241,7 +241,7 @@ drf-jsonschema-serializer==3.0.0
     # via -r requirements/base.in
 drf-nested-routers==0.93.4
     # via -r requirements/base.in
-drf-polymorphic==1.0.0
+drf-polymorphic==2.0.0
     # via -r requirements/base.in
 drf-spectacular==0.27.0
     # via -r requirements/base.in
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -415,7 +415,7 @@ drf-nested-routers==0.93.4
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
-drf-polymorphic==1.0.0
+drf-polymorphic==2.0.0
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -447,7 +447,7 @@ drf-nested-routers==0.93.4
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
-drf-polymorphic==1.0.0
+drf-polymorphic==2.0.0
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
--- a/requirements/extensions.txt
+++ b/requirements/extensions.txt
@@ -395,7 +395,7 @@ drf-nested-routers==0.93.4
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
-drf-polymorphic==1.0.0
+drf-polymorphic==2.0.0
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
--- a/requirements/type-checking.txt
+++ b/requirements/type-checking.txt
@@ -444,7 +444,7 @@ drf-nested-routers==0.93.4
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
-drf-polymorphic==1.0.0
+drf-polymorphic==2.0.0
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt


src/openforms/submissions/tests/test_submission_step_validate.py
```diff
@@ -683,3 +684,72 @@ def test_validate_step_with_nested_files_in_fieldset(self):
         )
 
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    @tag("gh-5191")
+    def test_validate_map_component_hidden_and_clear_on_hide(self):
+        submission = SubmissionFactory.create(
+            form__generate_minimal_setup=True,
+            form__formstep__form_definition__configuration={
```
viktorvanwijk, line 692: I think I generally use `SubmissionFactory.from_components` to generate submissions in tests. Is there a specific reason you initialize it this way?

viktorvanwijk (approved):

src/openforms/submissions/tests/test_submission_step_validate.py
sergei-maertens: `from_components` does more stuff than necessary for this test, while the `.create` using Traits is much closer to core factory boy API.

`from_components` pretty much only has value for me to pass a `bsn`/`kvk` and immediately provide values for the components being set up. I'd love to convert it to the `Params` pattern, but I think it's too complex and too ingrained in the existing test suite.

PR #5210, opened 2025-03-27
Fix non-existing variables being included in the required property of JSON schema

### LINKED ISSUE #5205: JSON Dump tests 3.1.0

3.1.0

### Customer reference

_No response_

### Bugs:
- I did a schema validation, and some issues popped up with values having different shapes then the schema claims:
Image
Using `anyOf` we can explain that values can have multiple shapes (for example; `null` vs `object`, or an empty string vs a pattern). For `auth` this would be something like: 
```
"auth": {
  "title": "Authentication summary",
  "anyOf": [
    {
      "type": "null"
    },
    {
      "type": "object",
      "additionalProperties": true
    }
  ]
}
```

- Hidden form fields are not send as part of the data, but the schema still has them as `required` property. Fixed with #5211 
- Removed form fields, whose variables were part of the JSON Dump data collection, are still part of the `required` list. Fixed with #5210 
    Steps to reproduce:
    1. Create form with 2 textfields
    2. Add JSON Dump registration, and add both fields
    3. Remove 1 textfield component (from formbuilder) and save the form
    4. Test form and check JSON Dump data

### Improvements:
- In the overview it isn't clear if the selectboxes component will be transformed to a list. Maybe good to add a `Transform to list` indication, along the `Values` and `Metadata`. (I think this was added for another ticket/feature, but adding it for completeness)
Image

- There are a couple components that follow some sort of pattern (like `postcode`, `iban`, `phone`) that don't have a pattern description in the schema. Maybe good to investigate which patterns we can add. (I think `postcode` can be added quite easily with `\d{4} [a-zA-Z]{2}`, but the others might be a bit more difficult/less predictable)

### Note:

I constantly forgot to add new form field variables to the Dump data. It could very well be a "me" thing, but maybe we can do something to make it clear that new form fields aren't added automatically? Just food for thought :)

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

viktorvanwijk:
Partly closes #5205 

**Changes**

The required property now accurately reflects the variables that are in the schema

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/forms/json_schema.py
+++ b/src/openforms/forms/json_schema.py
@@ -59,7 +59,7 @@ def generate_json_schema(
         "$schema": "https://json-schema.org/draft/2020-12/schema",
         "type": "object",
         "properties": requested_variables_schema,
-        "required": limit_to_variables,
+        "required": list(requested_variables_schema.keys()),
         "additionalProperties": False,
     }
 
--- a/src/openforms/forms/tests/test_json_schema.py
+++ b/src/openforms/forms/tests/test_json_schema.py
@@ -188,17 +188,17 @@ class GenerateJsonSchemaTests(TestCase):
                 },
                 "foo": {"type": "array", "title": "Foo"},
             },
-            "required": (
+            "required": [
+                "today",
+                "auth_bsn",
                 "firstName",
                 "lastName",
                 "select",
                 "selectboxes",
                 "radio",
                 "file",
-                "auth_bsn",
-                "today",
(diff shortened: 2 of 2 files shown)


sergei-maertens (commented):
Can you fix the failing test?

sergei-maertens (approved):

PR #5219, opened 2025-03-31
[#5207] Do not show valid items if table is invalid (Referentielijsten)

### LINKED ISSUE #5207: Referentielijsten tests 3.1.0
## General notes

* [x] The 'Reference list service' and 'Reference lists table code' are bold and have an asterisk, but no error is shown when leaving them empty and saving the modal
   
   Image

* [x] Is there documentation on how to set this up? I wasn't able to find anything #5215 

## Expired tables/items
- [x] Not sure if we should take this edge case into account, but valid items in a non-valid table still show up in the options of the fields #5219 

Image

Image

Image

## Email digest
- [x] Looks like there is some weird nesting going on in the email digest template: #5220 

Image

## Integration with Objects API
- [x] When a selectboxes component is mapped to the geometry filed - not sure if it makes sense to do this, though - it is not converted to a list: #5222 

Image
```bash
Outgoing request
---------------- request ----------------
POST http://localhost:8002/api/v2/objects
User-Agent: python-requests/2.32.2
Accept-Encoding: gzip, deflate, br
Accept: */*
Connection: keep-alive
Content-Crs: EPSG:4326
Content-Length: 211
Content-Type: application/json
Authorization: Token 171be5abaf41e7856b423ad513df1ef8f867ff48 
Request body:
b'{"type": "http://objecttypes-web:8000/api/v2/objecttypes/8faed0fa-7864-4409-aa6d-533a37616a9e", "record": {"typeVersion": 1, "data": {}, "startAt": "2025-03-27", "geometry": {"option1": false, "option2": true}}}'
```

## Integration with JSON Dump plugin
* [x] The schema is not updated properly when component options are prefilled from reference lists. Most likely because `rewrite_formio_components`, which is called in the plugin, does not perform a reference-list fetch.
#5244 

```json
{
  "values": {
    "select": ["option2"],
    "selectBoxes": {"option1": false, "option2": true},
    "radio1": "option2"
  },
  "values_schema": {
    "type": "object",
    "properties": {
      "select": {
        "type": "array",
        "items": {
          "type": "string",
          *"enum": ["",  ""]*
        },
        "title": "Select"
      },
      "selectBoxes": {
        "title": "Select Boxes",
        "type": "object",
        "properties": {
          *"": {"type": "boolean"}*
        },
        *"required": [""]*,
        "additionalProperties": false
      },
      "radio1": {
        "title": "Radio",
        "type": "string",
        *"enum": ["", ""]*
      }
    },
    "required": [
      "selectBoxes",
      "radio1",
      "select"
    ],
    "additionalProperties": false
  },
```
* [x] When the values of a selectboxes component are transformed to a list, the schema is not updated properly. The type is correctly set to 'array', but 'properties', 'required', and 'additionalProperties' should be removed.
#5244 
```json
{
  "selectBoxes": {
    "title": "Select Boxes",
    "type": "array",
    *"properties": {"": {"type": "boolean"}},*
    *"required": [""]*,
    *"additionalProperties": false*
  }
}
```
Also might be good to add the 'items' property:
```json
"items": {"type": "string", "enum": ["option1", "option2", etc...]
```

vaszig:
Closes #5207 partly

[skip: e2e]

**Changes**

- Added regression test for an invalid table and a valid option
- Check if table is valid, otherwise no need to retrieve and show options
- Converted Dutch code to English 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/contrib/referentielijsten/helpers.py
```diff
@@ -3,7 +3,10 @@
 from django.utils import timezone
 
 
-def check_expired_or_near_expiry(einddatum: str) -> bool:
+def check_expired_or_near_expiry(einddatum: str, near_expiry: bool = False) -> bool:
```
sergei-maertens, line 6: these kind of boolean flags are not intuitive, especially combined with the function name. I think it's cleaner to add an additional function `check_expired`. `near_expiry` being a boolean doesn't not make it easier for me to understand if that excludes or includes items that are near expiry.

* `check_expired`: it's expired or not - maps well to a boolean return value
* `check_expired_or_near_expiry` it is expired or will soon expire - also maps well

*if* you were to use one function, I'd make it `check_expiry` that returns `Literal["not_expired", "nearly_expired", "expired"]` but that gets too close to magic strings so I prefer separate functions.

And I think it'd be even better to implement these things as properties on the client - you already have `class TabelItem` and `class Tabel` - if you change the return data into dataclasses instead of typeddicts, you can add properties to them to check:

```py
@dataclass
class Tabel:
    code: str
    naam: str
    beheerder: Beheerder
    einddatumGeldigheid: str | None  # ISO 8601 datetime string

    @cached_property
    def parsed_end_date(self) -> datetime | None:
        if self.einddatumGeldigheid is None:
            return None
        return datetime.from_isoformat(self.einddatumGeldigheid)

    @cached_property
    def is_expired(self) -> bool:
        if self.parsed_end_date:
            return self.parsed_end_date <= timezone.now()
        return False

    @cached_property
    def is_nearly_expired(self) -> bool:
        if self.parsed_end_date:
            return self.parsed_end_date <= (timezone.now() + timedelta(days=7))
        return False
```
this properly encapsulates the domain logic

src/openforms/emails/digest.py
```diff
@@ -552,31 +551,30 @@ def collect_expired_or_near_expiry_referentielijsten_data():
...
                     ) as client:
                         try:
-                            response = client.get(
-                                "tabellen",
-                                params={"code": component["openForms"]["code"]},
-                            )
+                            table = client.get_tabel(component["openForms"]["code"])
                         except RequestException:
-                            response = []
+                            table = None
```
sergei-maertens, line 556: this doesn't make sense to me - if you get a crash when retrieving the table, surely you'll want to report that in the digest email instead of continuing the rest of the checks?

sergei-maertens (changes requested):

src/openforms/formio/dynamic_config/referentielijsten.py
```diff
@@ -55,7 +58,15 @@ def fetch_options_from_referentielijsten(
 
     try:
         with build_client(service, client_factory=ReferentielijstenClient) as client:
-            result = client.get_items_for_tabel_cached(code)
+            # check if the table is valid (we don't want to show the possible valid
+            # options of an invalid table)
+            table_data = client.get_tabel(code)
+            if table_data:
+                table = Tabel.from_dict(table_data)
```
sergei-maertens, line 65: Why not let the client return the data classes directly? You now have a data class and a typed dict describing the data, and they're practically the same. It makes much more sense to only use the dataclasses and not expose the typed dicts at all.

Make `client.get_table` return the `Table` data class instance, instead of converting the dict data outside of the client into such an instance. This will make it so much easier to work with the data and you'll have proper strong type checking.

And please, replace `tabel` with `table` in code.

sergei-maertens (changes requested):

src/openforms/contrib/reference_lists/client.py
```diff
@@ -0,0 +1,125 @@
...
+        if self.parsed_end_date:
+            return self.parsed_end_date <= (timezone.now() + timedelta(days=7))
+        return False
+
+
+@dataclass
+class TableItem(EndDateMixin):
+    code: str
+    name: str
+    expires_on: Optional[datetime] = None  # ISO 8601 datetime string
```
sergei-maertens, line 45:
```suggestion
    expires_on: datetime | None = None  # ISO 8601 datetime string
```
please use the modern syntax :) it's more explicit/obvious what it means

src/openforms/emails/digest.py
```diff
@@ -541,80 +540,84 @@ def collect_expired_or_near_expiry_referentielijsten_data():
...
                                 ].append(
-                                    ExpiredOrNearExpiryReferentielijstenTabel(
-                                        naam=table["naam"],
-                                        eindatum=datetime.fromisoformat(tabel_eindatum),
-                                        expired=datetime.fromisoformat(tabel_eindatum)
-                                        <= timezone.now(),
+                                    ExpiredOrNearExpiryReferenceListsTable(
+                                        name=table.name,
+                                        expires_on=table.expires_on,
+                                        expired=table.is_expired,
```
sergei-maertens, line 578: do we need to copy this information in another dataclass? Can we not just pass the `Table` and `TableItem` instances?

sergei-maertens (approved):

PR #5222, opened 2025-03-31
[#5207] Remove geometry field from variables configuration (`selectboxes`)

### LINKED ISSUE #5207: Referentielijsten tests 3.1.0

vaszig:
Closes #5207 partly

**Changes**

- Removed geometry field from variables configuration when component is selectboxes

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
@@ -28,7 +28,6 @@ export const AddressNlEditor = ({
   variable,
   components,
   namePrefix,
-  isGeometry,
   index,
   mappedVariable,
   objecttype,
@@ -147,7 +146,7 @@ export const AddressNlEditor = ({
             name={`${namePrefix}.targetPath`}
             index={index}
             mappedVariable={mappedVariable}
-            isDisabled={isGeometry || specificTargetPaths}
+            isDisabled={specificTargetPaths}
             isLoading={loading}
             targetPaths={objectTypeTargetPaths}
           />
--- a/src/openforms/js/components/admin/form_design/registrations/objectsapi/SelectboxesObjectsApiVariableConfigurationEditor.js
+++ b/src/openforms/js/components/admin/form_design/registrations/objectsapi/SelectboxesObjectsApiVariableConfigurationEditor.js
@@ -1,6 +1,5 @@
-import {FieldArray, useFormikContext} from 'formik';
+import {useFormikContext} from 'formik';
 import isEqual from 'lodash/isEqual';
-import PropTypes from 'prop-types';
 import React, {useContext} from 'react';
 import {FormattedMessage} from 'react-intl';
 import {useAsync, useToggle} from 'react-use';
@@ -8,75 +7,14 @@ import {useAsync, useToggle} from 'react-use';
 import {APIContext} from 'components/admin/form_design/Context';
 import Field from 'components/admin/forms/Field';
 import FormRow from 'components/admin/forms/FormRow';
-import {TargetPathSelect} from 'components/admin/forms/objects_api';
+import {Checkbox} from 'components/admin/forms/Inputs';
 import ErrorMessage from 'components/errors/ErrorMessage';
 
+import {MappedVariableTargetPathSelect} from './GenericObjectsApiVariableConfigurationEditor';
 import {asJsonSchema} from './utils';
 import {fetchTargetPaths} from './utils';
 
-/**
- * Hack-ish way to manage the variablesMapping state for one particular entry.
- *
- * We ensure that an item is added to `variablesMapping` by using the `FieldArray`
- * helper component if it doesn't exist yet, otherwise we update it.
- */
-export const MappedVariableTargetPathSelect = ({
-  name,
-  index,
-  mappedVariable,
-  isLoading = false,
-  targetPaths = [],
-  isDisabled = false,
-}) => {
-  const {
-    values: {variablesMapping = []},
-    setFieldValue,
-  } = useFormikContext();
-  const isNew = variablesMapping.length === index;
-  return (
-    <FieldArray
-      name="variablesMapping"
-      render={arrayHelpers => (
-        <TargetPathSelect
-          name={name}
-          isLoading={isLoading}
-          targetPaths={targetPaths}
-          isDisabled={isDisabled}
-          onChange={newValue => {
-            // Clearing the select means we need to remove the record from the mapping,
-            // otherwise it's not a valid item for the backend.
-            if (newValue === null) {
-              arrayHelpers.remove(index);
-              return;
-            }
-
-            // otherwise, either add a new item, or update the existing
-            if (isNew) {
-              const newMapping = {...mappedVariable, targetPath: newValue.targetPath};
-              arrayHelpers.push(newMapping);
-            } else {
-              setFieldValue(name, newValue.targetPath);
-            }
-          }}
-        />
-      )}
-    />
-  );
-};
-
-MappedVariableTargetPathSelect.propTypes = {
-  name: PropTypes.string.isRequired,
-  index: PropTypes.number.isRequired,
-  mappedVariable: PropTypes.shape({
-    variableKey: PropTypes.string.isRequired,
-    targetPath: PropTypes.arrayOf(PropTypes.string),
-    options: PropTypes.object,
-  }).isRequired,
-  isLoading: PropTypes.bool,
-  isDisabled: PropTypes.bool,
-};
-
-export const GenericEditor = ({
+export const SelectboxesEditor = ({
   variable,
   components,
   namePrefix,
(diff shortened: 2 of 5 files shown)


vaszig:
I thought that it is better to adjust the existing condition of rendering the `transformToList` field in combination with the `geometryVariableKey` instead of creating a new component (like we do with `addressNL`). Let me know if you do not agree.

sergei-maertens (changes requested):
No, this will definitely not scale to more component types and you lose all the aspects of "generic variable configuration editor" by polluting it with component specific code/conditions. If I accept this, we allow technical debt to creep in.

Let's do it proper instead and set up the proper configuration options form for selectboxes component.

vaszig:
> No, this will definitely not scale to more component types and you lose all the aspects of "generic variable configuration editor" by polluting it with component specific code/conditions. If I accept this, we allow technical debt to creep in.
> 
> Let's do it proper instead and set up the proper configuration options form for selectboxes component.

Ok,I will add another configuration/ component. I just didn't get the difference of what already tied to the generic editor (`transformToList`) with what we are doing now. Basically show/hide fields according to a specific component type.

sergei-maertens:
> > No, this will definitely not scale to more component types and you lose all the aspects of "generic variable configuration editor" by polluting it with component specific code/conditions. If I accept this, we allow technical debt to creep in.
> > Let's do it proper instead and set up the proper configuration options form for selectboxes component.
> 
> Ok,I will add another configuration/ component. I just didn't get the difference of what already tied to the generic editor (`transformToList`) with what we are doing now. Basically show/hide fields according to a specific component type.

Perhaps this geometry option should only be shown if it's a map component?

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
```diff
@@ -28,12 +28,12 @@ export const AddressNlEditor = ({
   variable,
   components,
   namePrefix,
-  isGeometry,
   index,
   mappedVariable,
   objecttype,
   objectsApiGroup,
   objecttypeVersion,
+  backendOptions,
```
sergei-maertens, line 36: is this used? If it is, then older versions are broken, no?

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
```diff
@@ -81,20 +80,17 @@ export const GenericEditor = ({
...
   components,
   namePrefix,
-  isGeometry,
-  transformToList,
   index,
   mappedVariable,
   objecttype,
   objectsApiGroup,
   objecttypeVersion,
+  backendOptions,
```
sergei-maertens, line 88: I don't think this is used?

src/openforms/js/components/admin/form_design/registrations/objectsapi/SelectboxesObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,132 @@
...
+  objectsApiGroup,
+  objecttypeVersion,
+  backendOptions,
+}) => {
+  const {csrftoken} = useContext(APIContext);
+  const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
+  const {setFieldValue} = useFormikContext();
+  const {transformToList} = backendOptions;
+
+  const transformationNeeded = transformToList?.includes(variable.key);
```
sergei-maertens, line 33:
```suggestion
  const {transformToList = []} = backendOptions;

  const transformationNeeded = transformToList.includes(variable.key);
```

sergei-maertens (changes requested):
Minor tweaks and then it'll be good to merge

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
vaszig: No, `backendOptions` was added by me only for using it in the editors that I need to access properties like the `transformToList` in the `SelectBoxesVariableConfigurationEditor`  (didn't want to pass along all the editors specific properties that may confuse us).

src/openforms/js/components/admin/form_design/registrations/objectsapi/GenericObjectsApiVariableConfigurationEditor.js
vaszig: Same as above

src/openforms/js/components/admin/form_design/registrations/objectsapi/AddressNlObjectsApiVariableConfigurationEditor.js
sergei-maertens: yeah but you don't need to specify the props in the signature that aren't used. it's okay if more props are passed, but now we have an unused variable that will trip up linters/typescript in the future :)

src/openforms/js/components/admin/form_design/registrations/objectsapi/SelectboxesObjectsApiVariableConfigurationEditor.js
```diff
@@ -0,0 +1,132 @@
...
+  objectsApiGroup,
+  objecttypeVersion,
+  backendOptions,
+}) => {
+  const {csrftoken} = useContext(APIContext);
+  const [jsonSchemaVisible, toggleJsonSchemaVisible] = useToggle(false);
+  const {setFieldValue} = useFormikContext();
+  const {transformToList = []} = backendOptions;
+
+  const transformationNeeded = transformToList?.includes(variable.key);
```
sergei-maertens, line 33:
```suggestion
  const transformationNeeded = transformToList.includes(variable.key);
```
the nullish-coalesce `?.` is not necessary anymore because we default `tranformToList` to an empty array on L33

vaszig: And this is exactly what we discussed yesterday about the feedback I receive..apologies, I will update it

sergei-maertens (approved):

PR #5228, opened 2025-04-02
Replace `DataMapping type hints with `FormioData`

### LINKED ISSUE #5221: Replace `DataMapping` with `FormioData`
Wherever `DataMapping` is used as a type hint, replace it with `FormioData` and ensure the corresponding function can deal with it.

This will reduce the amount of places that initialize a `FormioData` instance, and remove any remaining raw `glom` usages.

viktorvanwijk:
Closes #5221

**Changes**

Replaced `DataMapping type hints with `FormioData`

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/formio/rendering/tests/test_utils.py
```diff
@@ -216,8 +216,8 @@ def test_columns(self):
             {
                 "fd0": {
                     "column1": {
-                        0: {"input1": "Foo"},
-                        1: {
+                        "0": {"input1": "Foo"},
+                        "1": {
```
viktorvanwijk, line 220: Not completely sure if this matters

src/openforms/formio/rendering/nodes.py
```diff
@@ -42,11 +40,11 @@ class RenderConfiguration:
 @dataclass
 class ComponentNode(Node):
     component: Component
-    step_data: DataMapping  # XXX refactor to FormioData
+    step_data: FormioData
     depth: int = 0
     is_layout = False
-    path: Path | None = None  # Path in the data (#TODO rename to data_path?)
-    json_renderer_path: Path | None = (
+    path: str | None = None  # Path in the data (#TODO rename to data_path?)
```
sergei-maertens, line 46: can we further simplify this to 
```suggestion
    path: str = ""  # Path in the data (#TODO rename to data_path?)
```
or is that going to cause issues if someone uses a key like `foo..bar`?

src/openforms/formio/rendering/nodes.py
```diff
@@ -56,11 +54,11 @@ class ComponentNode(Node):
...
     @staticmethod
     def build_node(
-        step_data: DataMapping,
+        step_data: FormioData,
         component: Component,
         renderer: "Renderer",
-        path: Path | None = None,  # Path in the data
-        json_renderer_path: Path | None = None,
+        path: str | None = None,  # Path in the data
+        json_renderer_path: str | None = None,
```
sergei-maertens, line 61: and again :P

src/openforms/formio/rendering/nodes.py
```diff
@@ -161,14 +162,6 @@ def key(self):
         assert "key" in self.component
         return self.component["key"]
 
-    @property
-    def key_as_path(self) -> Path:
-        """
-        See https://glom.readthedocs.io/en/latest/api.html?highlight=Path#glom.Path
-        Using Path("a.b") in glom will not use the nested path, but will look for a key "a.b"
-        """
-        return Path.from_text(self.key)
```
sergei-maertens, line 170: good riddance :coffin:

src/openforms/formio/rendering/tests/test_utils.py
sergei-maertens: for the serialized JSON not, because only strings can be used as key in JSON. I'm a little bit surprised that `json.dumps` doesn't choke on this.

```py
>>> import json
>>> json.dumps({0: "foo"})
'{"0": "foo"}'
```

src/openforms/formio/service.py
```diff
@@ -63,8 +66,8 @@ def normalize_value_for_component(component: Component, value: Any) -> Any:
 @elasticapm.capture_span(span_type="app.formio")
 def get_dynamic_configuration(
     config_wrapper: FormioConfigurationWrapper,
-    submission: Submission,
-    data: DataMapping | None = None,
+    submission: "Submission",
```
sergei-maertens, line 69: could benefit from `from __future__ import annotations` too

src/openforms/registrations/contrib/json_dump/plugin.py
```diff
@@ -52,10 +52,8 @@ def register_submission(
         # Update static values with registration variables
         static_values.update(state.get_static_data(other_registry=variables_registry))
 
-        all_values: JSONObject = {
-            **static_values,
-            **state.get_data(),  # dynamic values from user input
-        }
+        all_values = state.get_data()
+        all_values.update(static_values)
```
sergei-maertens, line 56: oooh this one has a semantic difference!

If there's a key collission, e.g. a component with the key `now`, then now the static variable value is used whereas the original had the component override the static variable.

I'm not sure if we do this override-behaviour consistent in the code base, but it's worth checking.

sergei-maertens (commented):
:bow:

src/openforms/registrations/contrib/objects_api/submission_registration.py
```diff
@@ -506,15 +506,15 @@ def get_record_data(
...
         # static values: values not driven by user input
         static_values = state.get_static_data()
         # update with the registration static values - a special subtype of static
         # variables. They're possibly derived from user input.
         static_values.update(state.get_static_data(other_registry=variables_registry))
 
-        # merge everything in one container where we can easily do dotted key lookups
-        all_values = FormioData({**dynamic_values, **static_values})
+        # add static values
+        all_values.update(static_values)
```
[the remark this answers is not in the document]
viktorvanwijk: OK, will keep the same behaviour. Isn't this v2, though?

src/openforms/registrations/contrib/json_dump/plugin.py
viktorvanwijk: It seems there are only two other occurrences: Objects API and `SubmissionValueVariablesState.to_python`, both of which override the dynamic with the static values.

Should I keep it consistent with the others, or revert the change?

src/openforms/formio/rendering/nodes.py
sergei-maertens: I'd prefer that, since always the same type is easier to reason about than even a simple union of types with `None` :)

src/openforms/registrations/contrib/json_dump/plugin.py
sergei-maertens: Let's keep it consistent then, but make sure to add a placeholder entry in the changelog that possibly forms may be affected. I think our validation now guards against using a key that's already present in static variables, but that doesn't cover old forms and newly added static variables of course, so let's at least inform users in case their forms start working differently.

src/openforms/registrations/contrib/objects_api/submission_registration.py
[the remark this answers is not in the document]
sergei-maertens: I was wrong, this is v2 indeed and not v1. Let's keep your change then with the above remark in mind from the generic json registration plugin.

If someone complains, we can always revert if really needed and then deprecate it to be changed in v4, but I hope it doesn't come to that.

Could also be a good email digest check - checking if form variable keys collide with static/registration variables :thinking:

viktorvanwijk:
I'm not sure why I thought I didn't need to re-record the cassettes when fixing #5263... Clearly we are sending different data :see_no_evil:

sergei-maertens (approved):
Let's go! :zap:

PR #5241, opened 2025-04-07
Support broader range of geojson in schema generation

### LINKED ISSUE #5027: Support broader range of GeoJSON in schema generation
Robin added support for drawing multiple kinds of shapes on maps, which is a subset of the GeoJSON spec. This is more extensive than the current latitude/longitude array that is support for a single marker - see #2177 and #2178

Keep in mind that the component settings also determine the available shapes - this will be a polymorphic structure too (using `anyOf`).

viktorvanwijk:
Closes #5027

[skip: e2e]

**Changes**
Update the JSON schema of the map component to include line and polygon shapes

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -13,13 +13,16 @@ from glom import glom
 from rest_framework import ISO_8601, serializers
 from rest_framework.request import Request
 
-from openforms.api.geojson import GeoJsonGeometryPolymorphicSerializer
+from openforms.api.geojson import (
+    GeoJsonGeometryPolymorphicSerializer,
+    GeoJsonGeometryTypes,
+)
 from openforms.authentication.service import AuthAttribute
 from openforms.config.models import GlobalConfiguration, MapTileLayer
 from openforms.submissions.models import Submission
 from openforms.typing import DataMapping, JSONObject
 from openforms.utils.date import TIMEZONE_AMS, datetime_in_amsterdam, format_date_value
-from openforms.utils.json_schema import to_multiple
+from openforms.utils.json_schema import GEO_JSON_COORDINATE_SCHEMAS, to_multiple
 from openforms.utils.validators import BSNValidator, IBANValidator
 from openforms.validations.service import PluginValidator
 
@@ -48,6 +51,13 @@ from .np_family_members.stuf_bg import get_np_family_members_stuf_bg
 from .utils import _normalize_pattern, salt_location_message
 
 logger = logging.getLogger(__name__)
+
+
+GEO_JSON_TYPE_TO_INTERACTION = {
+    GeoJsonGeometryTypes.point: "marker",
+    GeoJsonGeometryTypes.polygon: "polygon",
+    GeoJsonGeometryTypes.line_string: "polyline",
+}
 
 
 class FormioDateField(serializers.DateField):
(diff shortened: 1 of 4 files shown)


src/openforms/utils/json_schema.py
```diff
@@ -1,5 +1,41 @@
+from copy import deepcopy
+
+from openforms.api.geojson import GeoJsonGeometryTypes
```
viktorvanwijk, line 3: Not entirely sure about this import, but I saw that it was also used in `formio.components.custom`

src/openforms/formio/components/custom.py
```diff
@@ -243,21 +254,33 @@ def build_serializer_field(
...
+            {
+                "properties": {
+                    "type": {"type": "string", "const": geometry_type},
+                    "coordinates": deepcopy(GEO_JSON_COORDINATE_SCHEMAS[geometry_type]),
+                },
+                "additionalProperties": False,
+            }
+            for geometry_type in GeoJsonGeometryTypes.values
+            # Only include the schema of types that are allowed
+            if interactions.get(GEO_JSON_TYPE_TO_INTERACTION[geometry_type], True)
```
sergei-maertens, line 270:
```suggestion
            if interactions.get(GEO_JSON_TYPE_TO_INTERACTION[geometry_type], False)
```
If the interaction key is not defined, it'll be disabled in the frontend, so the fallback value should also be falsy.

src/openforms/formio/components/custom.py
```diff
@@ -243,21 +254,33 @@ def build_serializer_field(
...
-        label = component.get("label", "Map coordinate")
+        label = component.get("label", "Map")
+        interactions = component["interactions"]
 
-        base = {
+        properties = [
+            {
+                "properties": {
+                    "type": {"type": "string", "const": geometry_type},
+                    "coordinates": deepcopy(GEO_JSON_COORDINATE_SCHEMAS[geometry_type]),
```
sergei-maertens, line 264: do we need the deepcopy here? If someone is mutating these constants, I think they are a prime candidate to receive a practical example of physical violence :laughing:

src/openforms/formio/tests/test_component_json_schemas.py
```diff
@@ -396,3 +387,189 @@ def test_data_source_is_another_form_variable(self):
...
+            "properties": {
+                "type": {"type": "string", "const": "Point"},
+                "coordinates": {
+                    "type": "array",
+                    "prefixItems": [
+                        {"title": "Longitude", "type": "number"},
+                        {"title": "Latitude", "type": "number"},
+                    ],
+                    "items": False,
+                    "minItems": 2,
```
sergei-maertens, line 416: should we also have `maxItems` here? Or is this covered by `prefixItems`?

src/openforms/utils/json_schema.py
sergei-maertens: yeah it's okay :) `openforms.api` is public/core API

sergei-maertens (commented):
Some questions - not sure whether to comment or approve :D

src/openforms/formio/components/custom.py
viktorvanwijk: You can never be too sure, though

src/openforms/formio/tests/test_component_json_schemas.py
viktorvanwijk: In combination with `"items": false`, yes :)

sergei-maertens (approved):

PR #5294, opened 2025-04-22
[#5254] Impement family members prefill plugin

### LINKED ISSUE #5254: Implement new prefill plugin for (BRP) family members
**Acceptance criteria**

- [x] I can add a user defined variable and prefill it with the partners or children of the authenticated user
- [x] The data is retrieved from Haal Centraal BRP personen bevragen or StUF-BG
- [x] The retrieved data can optionally be filtered by some parameters and only the filtered result is stored on our end
- [x] The plugin and its configuration is documented
- [x] There is an example form with working/representative configuration in the docs

**Wireframes**

<details>
<summary>Partners</summary>

```
+------------------------------------------------------+
| 'Partners' prefill                                 X |
|                                                      |
|         +---------------------+                      |
| Plugin: | Family members    v |                      |
|         +---------------------+                      |
|                                                      |
|         +---------------------+                      |
| Type:   | Partners          v |                      |
|         +---------------------+                      |
|                                                      |
|                                           +------+   |
|                                           | Save |   |
|                                           +------+   |
+------------------------------------------------------+
```
</details>

<details>
<summary>Children</summary>

```
+------------------------------------------------------+
| 'Children' prefill                                 X |
|                                                      |
|         +---------------------+                      |
| Plugin: | Family members    v |                      |
|         +---------------------+                      |
|                                                      |
|         +---------------------+                      |
| Type:   | Children          v |                      |
|         +---------------------+                      |
|                                                      |
| **Filters**                                          |
|                                                      |
| Minimum age: _3__                                    |
|                                                      |
| Maximum age: _18_                                    |
|                                                      |
| ☑ Include deceased                                   |
|                                                      |
| ☑ Exclude children on different address              |
|                                                      |
|                                           +------+   |
|                                           | Save |   |
|                                           +------+   |
+------------------------------------------------------+

```
</details>

**Tasks**

- [x] Add one (or two, if that makes more sense, adapt accordingly) prefill plugins: key `family-members` and name "Family members"
- [x] The plugin uses the `options` mechanism rather than the `attributes` of the legacy pattern (so it's more similar to objects API prefill)
- [x] Define a (polymorphic?) serializer for the plugin options, with fields:
    - `type`, available choices are `partners` or `children`
    - for `partners`, currently no additional fields are required
    - for `children`, the following extra configuration options become relevant:
        - [x] Minimum age, inclusive
        - [x] Maximum age, inclusive
        - [x] Include deceased boolean, default False
        - [ ] (to validate with BRP behaviour) Include only children living on the same address as the logged in parent.
- [x] Define which attributes to lookup/store for partners and children each. Initially we hardcode this, this may become dynamic in the future
- [x] Use the existing configuration for whether HC BRP bevragen or StUF-BG must be used for family members component. If it's not configured, this must be communicated through a validation error.
- [x] Implement looking up and storing the partners/children data
- [x] Implement the frontend to configure the prefill plugin
    - [x] In the user defined variables, this new plugin needs to be available for the prefill options
    - [x] The user defined variable must be of `array` type for both partners and children, as the real world returns 0, 1 or more records
    - [x] The additional filters/options are displayed conditionally based on the selected type (partners vs children)
    - [x] Ensure that validation errors are displayed correctly (in particular the non-field-error about missing plugin configuration)
    - [x] The components must be documented in Storybook

The expected schema/types for the data. Note that we'll calculate the age dynamically, as this depends on when
the form is filled out or resumed (may be on another day!).

```py
class NaturalPersonDetails(TypedDict):
    bsn: str
    initials: str
    affixes: str
    lastName: str
    dateOfBirth: str (ISO-8601)
    dateOfBirthPrecision: Literal['day', 'month', 'year']

type Partners = list[NaturalPersonDetails]
type Children = list[NaturalPersonDetails]
```

vaszig:
Closes #5254

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/prefill/contrib/family_members/plugin.py
```diff
@@ -0,0 +1,78 @@
...
+from .typing import FamilyMembersChildOptions, FamilyMembersPartnerOptions
+
+logger = logging.getLogger(__name__)
+
+
+PLUGIN_IDENTIFIER = "family_members"
+
+
+@register(PLUGIN_IDENTIFIER)
+class FamilyMembersPrefill(
```
vaszig, line 30: As mentioned during standup as well, I chose to go for one prefill plugin because to me it has to do with one thing (retrieving person's information) that can be achieved by using two different sources. Let me know if you agree with my thoughts.

docker/haal_centraal_brp/data/test-data-changelog.json
```diff
@@ -0,0 +1,17 @@
+{
```
vaszig, line 1: This is the way I thought about it but I am wondering if there is a more straight forward/simpler way to keep track of these changes, since we cannot add comments to the json file.

src/openforms/prefill/contrib/family_members/filters.py
```diff
@@ -0,0 +1,133 @@
...
+
+from openforms.contrib.haal_centraal.clients import get_brp_client
+from openforms.contrib.haal_centraal.clients.brp import NaturalPersonDetails
+from openforms.contrib.haal_centraal.constants import BRPVersions
+from openforms.utils.date import get_today
+
+from .constants import HC_DECEASED_ATTRIBUTES
+
+
+def exclude_deceased_for_haal_centraal(
```
sergei-maertens, line 12: the function name creates the impression that you'd be filtering on already retrieved results, but this function is actually making the API calls. That's a bit unexpected.

src/openforms/prefill/contrib/family_members/filters.py
```diff
@@ -0,0 +1,133 @@
...
+        eligible_members = [
+            result for result in results if result["bsn"] not in deceased_bsns
+        ]
+        return eligible_members
+
+
+def is_eligible_based_on_age(
+    provided_age: int | None = None,
+    date_of_birth: str | None = None,
+    date_of_birth_precision: Literal["date", "year_month", "year"] | None = None,
```
sergei-maertens, line 51: I'd really define a helper `type` for this `Literal` so that you don't have to repeat it everywhere :)

src/openforms/prefill/contrib/family_members/filters.py
```diff
@@ -0,0 +1,133 @@
...
+    max_age: int | None = None,
+) -> bool:
+    # Haal Centraal v1
+    if provided_age:
+        age = provided_age
+
+    # Haal Centraal v2 and StUF-BG
+    elif date_of_birth:
+        date_of_birth_obj = None
+        today_obj = datetime.strptime(get_today(), "%Y-%m-%d")
```
sergei-maertens, line 62: I'd use `datetime_in_amsterdam(timezone.now()).date()` instead to avoid the extra roundtrip of parsing. Perhaps you can add `get_today_date()` in the utils that does that?

src/openforms/prefill/contrib/family_members/filters.py
```diff
@@ -0,0 +1,133 @@
...
+            for member in members
+            if isinstance(member, dict) and "overlijden" in member
+        ]
+        eligible_members = [
+            result for result in results if result["bsn"] not in deceased_bsns
+        ]
+        return eligible_members
+
+
+def is_eligible_based_on_age(
```
sergei-maertens, line 48: Probably a (calculated) (cached_)property on a natural person pydantic model/dataclass would be easier so you can just check on `person.age`:

```py
age_filtered = [
    person for person in persons
    if min_age <= person.age <= max_age
]
```
and then you can also set `min_age` to `min_age or 0` (in case it's None, just go with the lowest possible value) and do the same for `max_age` with `500` or so - I think it'll be a while before someone actually lives for 5 centuries :)

src/openforms/prefill/contrib/family_members/plugin.py
```diff
@@ -0,0 +1,78 @@
...
+from ...registry import register
+from .api.serializers import FamilyMembersOptionsSerializer
+from .haal_centraal import get_data_from_haal_centraal
+from .stuf_bg import get_data_from_stuf_bg
+from .typing import FamilyMembersChildOptions, FamilyMembersPartnerOptions
+
+logger = logging.getLogger(__name__)
+
+
+PLUGIN_IDENTIFIER = "family_members"
```
sergei-maertens, line 26:
```suggestion
PLUGIN_IDENTIFIER = "family-members"
```
I think this is our stylistic approach?

src/openforms/prefill/contrib/family_members/typing.py
```diff
@@ -0,0 +1,16 @@
+from typing import Literal, TypedDict
+
+from .constants import FamilyMembersTypeChoices
+
+
+class FamilyMembersPartnerOptions(TypedDict):
+    variable_key: str
+    type: Literal[FamilyMembersTypeChoices.partners, FamilyMembersTypeChoices.children]
```
sergei-maertens, line 8: you can't do `Literal` with dynamic code, it has to be statically defined :(

I'd go for `type: FamilyMembersTypeChoices` here

src/openforms/prefill/contrib/family_members/utils.py
```diff
@@ -0,0 +1,18 @@
+# StUF-BG config checks specific
```
sergei-maertens, line 1: is this something that belongs in `src/stuf/stuf_bg`?

src/stuf/stuf_bg/checks.py
```diff
@@ -0,0 +1,59 @@
...
+        return True
+
+
+def is_empty_wrapped_response(xml):
+    meta = xml.xpath("//*[local-name()='Body']/*/*[local-name()='stuurgegevens']")
+    response = xml.xpath("//*[local-name()='Body']/*/*[local-name()='antwoord']")
+    if meta and not response:
+        return True
+    else:
+        return False
```
sergei-maertens, line 25: duplicated?

src/stuf/stuf_bg/checks.py
```diff
@@ -0,0 +1,59 @@
+from django.utils.translation import gettext_lazy as _
+
+import requests
+from lxml import etree
+
+from openforms.plugins.exceptions import InvalidPluginConfiguration
+from openforms.utils.xml import fromstring
```
sergei-maertens, line 7: the `stuf` package cannot know about the `openforms` package

`src/stuf` is a candidate to be published on PyPI when other projects also do things with stuf, here especially you have to respect the allowed public APIs

src/stuf/stuf_bg/client.py
```diff
@@ -1,11 +1,16 @@
 import logging
 from collections.abc import Mapping
 from functools import partial
-from typing import TypeVar
+from typing import Literal, TypeVar, cast
 
 import xmltodict
 from glom import glom
 
+from openforms.contrib.haal_centraal.typing import (
```
sergei-maertens, line 9: forbidden `openforms` import :)

src/stuf/stuf_bg/tests/test_checks.py
```diff
@@ -0,0 +1,87 @@
+from contextlib import contextmanager
+from unittest.mock import patch
+
+from django.test import TestCase
+
+import requests_mock
+from requests import RequestException
+
+from openforms.plugins.exceptions import InvalidPluginConfiguration
+from openforms.prefill.contrib.stufbg.plugin import StufBgPrefill
```
sergei-maertens, line 10: more forbidden imports

src/stuf/stuf_bg/typing.py
```diff
@@ -0,0 +1,9 @@
+from typing import TypedDict
+
+StufBgIncompleteDateType = TypedDict(
+    "StufBgIncompleteDateType",
+    {
+        "@indOnvolledigeDatum": str,
+        "#text": str,
+    },
+)
```
sergei-maertens, line 9: ...I didn't even know you could do that

sergei-maertens (commented):
Can you also update the pyright config file to type check the new code in CI? Let's make sure that the type annotations are correct in new code :)

The review is a rough general-pattern review - haven't looked into the details too much yet!

src/openforms/contrib/family_members/models.py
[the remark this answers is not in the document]
vaszig: Yes, I don't see a specific reason for having a whole separate model for this. The only thing is that we keep adding fields in the global configuration model, which I guess cannot be avoided.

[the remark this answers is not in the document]
sergei-maertens: yeah it's not very elegant, I think if/when we redesign the settings we can take a critical look at a better approach to manage settings :)

src/openforms/contrib/haal_centraal/typing.py
```diff
@@ -0,0 +1,49 @@
...
+
+class NaturalPersonDetails(TypedDict):
+    bsn: str | None
+    firstNames: str | None
+    initials: str | None
+    affixes: str | None
+    lastname: str | None
+    date_of_birth: str | None  # ISO - 8601
+    # Haal Centraal v2 specific
+    date_of_birth_precision: Literal["date", "year_month", "year"] | None
```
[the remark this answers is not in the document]
vaszig: No, they are part of HaalCentraal where the date of birth has an extra attribute `type` which declares which format is available.

src/openforms/prefill/contrib/family_members/plugin.py
vaszig: No, I have done a research and I saw `objects_api` for example.

src/openforms/prefill/contrib/family_members/filters.py
vaszig: I thought `get_today` does that already, perhaps I am missing something here..
https://github.com/open-formulieren/open-forms/blob/master/src/openforms/utils/date.py#L84

src/openforms/prefill/contrib/family_members/filters.py
```diff
@@ -0,0 +1,133 @@
...
+        age = provided_age
+
+    # Haal Centraal v2 and StUF-BG
+    elif date_of_birth:
+        date_of_birth_obj = None
+        today_obj = datetime.strptime(get_today(), "%Y-%m-%d")
+
+        # we need the complete date of birth in order to be able to filter a person's age.
+        # Haal Centraal already makes it clear which type of date we have (date_of_birth_precision)
+        # for StUF-BG we already have only the complete dates.
```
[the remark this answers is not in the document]
vaszig: I am normalizing the date according to `indOnvolledigeDatum` in the StUF-BG utils before I pass it here, that's why I am saying that's already the correct date.

src/openforms/prefill/contrib/family_members/tests/test_filters.py
```diff
@@ -0,0 +1,308 @@
...
+        ]
+
+        self.STUF_BG_DATA = [
+            StUFBG_NaturalPersonDetails(
+                bsn="123456788",
+                first_names="Doe",
+                lastname="Pirsen",
+                date_of_birth="2007-01-19",
+            ),
+        ]
```
sergei-maertens, line 35: setting up test data in `setUp` that's modified inside each tests or not used/relevant for a test at all is a complete anti-pattern. Set up the necessary test data at the beginning of the test, and then assert on it instead.

For the irrelevant fields, you can use factory boy - it's not just for Django models.

src/openforms/prefill/contrib/family_members/tests/test_plugin.py
```diff
@@ -0,0 +1,694 @@
...
+        submission = SubmissionFactory.from_components(
+            auth_info__value="999970124",
+            auth_info__attribute=AuthAttribute.bsn,
+            components_list=[],
+        )
+        FormVariableFactory.create(
+            key="hc_prefill_partners_mutable",
+            form=submission.form,
+            data_type=FormVariableDataTypes.array,
+        )
```
sergei-maertens, line 78:
```suggestion
        FormVariableFactory.create(
            key="hc_prefill_partners_mutable",
            form=submission.form,
            user_defined=True,
            data_type=FormVariableDataTypes.array,
        )
```
please make this representative, we have traits for the common use cases

src/openforms/prefill/contrib/stufbg/tests/test_checks.py
```diff
@@ -0,0 +1,87 @@
...
+    content = get_mock_response_content(template)
+    with requests_mock.Mocker() as m:
+        m.register_uri(requests_mock.ANY, requests_mock.ANY, content=content)
+        yield
+
+
+class StufBgCheckTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.stuf_bg_service = StufServiceFactory.create()
```
sergei-maertens, line 28: missing `super` call

src/openforms/submissions/models/submission.py
```diff
@@ -541,6 +541,7 @@ def remove_sensitive_data(self):
     def load_submission_value_variables_state(
         self, refresh: bool = False
     ) -> SubmissionValueVariablesState:
+        # breakpoint()
```
sergei-maertens, line 544:
```suggestion
```

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -98,6 +98,7 @@ def get_variables_in_submission_step(
     def collect_variables(self) -> dict[str, SubmissionValueVariable]:
         # leverage the (already populated) submission state to get access to form
         # steps and form definitions
+        # breakpoint()
```
sergei-maertens, line 101:
```suggestion
```

src/stuf/stuf_bg/client.py
```diff
@@ -45,15 +51,17 @@ class Client(BaseClient):
     sector_alias = "bg"
     soap_security_expires_minutes = STUF_BG_EXPIRY_MINUTES
 
-    def get_values_for_attributes(self, bsn: str, attributes: list[str]) -> bytes:
+    def get_values_for_attributes(
+        self, bsn: str, attributes: list[str], template: str = ""
```
sergei-maertens, line 55:
```suggestion
        self, bsn: str, attributes: list[str], template: str = "stuf_bg/StufBgRequest.xml"
```

src/stuf/stuf_bg/client.py
```diff
@@ -45,15 +51,17 @@ class Client(BaseClient):
...
+    ) -> bytes:
         context = {
             # replace . with _ to circumvent dot notation in template
             **{attr.replace(".", "_"): True for attr in attributes},
             "bsn": bsn,
         }
         response = self.templated_request(
             "npsLv01",
-            template="stuf_bg/StufBgRequest.xml",
+            template=template if template else "stuf_bg/StufBgRequest.xml",
```
sergei-maertens, line 64:
```suggestion
            template=template,
```

src/stuf/stuf_bg/client.py
```diff
@@ -93,6 +101,86 @@ def get_values(self, bsn: str, attributes: list[str]) -> dict:
...
+
+        # no fault, but also empty antwoord data -> treat this as empty response
+        if fault is None:
+            return []
+
+        # we have a fault -> log it appropriately and raise an exception
+        logger.error(
+            "Response data has an unexpected shape",
+            extra={"response": dict_response, "fault": fault},
+        )
```
sergei-maertens, line 181: wrong logging format

src/stuf/stuf_bg/constants.py
```diff
@@ -17,6 +24,37 @@
 STUF_BG_EXPIRY_MINUTES = 5
 
 
+type DateOfBirthPrecisionType = Literal["date", "year_month", "year"]
+
+
+class NaturalPersonDetails(BaseModel):
```
sergei-maertens, line 30: this is not a constant, it's a data container. Probably better to put this in a `data.py`. (I'd normally go for `models.py`, but that's too confusing with django models)

src/stuf/stuf_bg/constants.py
```diff
@@ -17,6 +24,37 @@
...
+    bsn: str | None = None
```
sergei-maertens, line 31: why/how can these string fields be `None`? Shouldn't those be normalized to empty strings?

src/stuf/stuf_bg/constants.py
```diff
@@ -17,6 +24,37 @@
...
+    first_names: str | None = None
+    initials: str | None = None
+    affixes: str | None = None
+    lastname: str | None = None
+    date_of_birth: str | None = None  # ISO - 8601
```
sergei-maertens, line 36: is this really an ISO-8601 date string? How does that work with partial dates?

sergei-maertens (changes requested):

src/openforms/contrib/haal_centraal/constants.py
```diff
@@ -1,7 +1,77 @@
...
+        try:
+            birth_date = datetime.strptime(self.date_of_birth, "%Y-%m-%d")
+            today = datetime.strptime(get_today(), "%Y-%m-%d")
+            age = today.year - birth_date.year
+
+            if (today.month, today.day) < (birth_date.month, birth_date.day):
+                age -= 1
+
+            return age
+        except ValueError:
```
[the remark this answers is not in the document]
vaszig: Yeah, it's only needed in the `StUF-BG` implementation where we do not know if the date (str) will be suitable. I will remove it from here.

src/openforms/prefill/contrib/family_members/api/serializers.py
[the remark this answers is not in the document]
vaszig: I will change it, I saw we used the same for the `ObjectsAPIOptionsSerializer` and I wanted to keep it consistent.

src/openforms/prefill/contrib/family_members/haal_centraal.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if options["type"] == FamilyMembersTypeChoices.partners:
+            partners_list = client.get_family_members(
+                bsn, include_partner=True, include_children=False
+            )
+
+            results[form_variable_to_update][options["type"]] = partners_list
+            results[initial_data_form_variable][options["type"]] = partners_list
+        elif options["type"] == FamilyMembersTypeChoices.children:
+            include_deceased = options.get("include_deceased")
```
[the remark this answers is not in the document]
sergei-maertens: yeah this looks like a type narrowing problem - it's unfortunate if you have to handle something that should never happen because of polymorphism, but I think that you can point the type checker in the right direction by using a `match` statement so that it can properly narrow the `options` type.

src/openforms/prefill/contrib/family_members/tests/test_plugin.py
vaszig: The factory creates a `user-defined` by default that's why I did it like this

sergei-maertens: Ah right, the factory itself is a bit confusing then :P I'll allow it :wink:

src/stuf/stuf_bg/constants.py
vaszig: When it's complete yes, but I guess I have to remove it since it's not always the case.

sergei-maertens (approved):

PR #5295, opened 2025-04-22
Extend authentication module plugin architecture

### LINKED ISSUE #5140: Extend authentication module capability
### Thema / Theme

Frontend

Before we can add support for eIDAS and Yivi, we have to update our authentication module.

- Update/rework the architecture a bit of the authentication module. In particular, we need to be able to receive form/plugin-specific options in the start_login and handle_return hooks of the BasePlugin. The DigiD plugin currently uses the check_requirements hook for this, this could/should be adopted.
- Extend the openforms.authentication models so that we can capture arbitrary key/value pairs from the obtained OIDC claims (e.g. first/last name, date of birth, ...). This includes being able to configure claim mappings on a per-form level for the OIDC-based plugins (eIDAS, Yivi)

TODO:
- [X] Provide plugin-specific options to `start_login`
- [X] Provide plugin-specific options to `handle_return`
- [ ] Update `check_requirements` of DigiD (and eHerkenning) plugin
- [x] Extend openforms.authentication models with OIDC claims
- [x] Add custom configurable scopes to form

_No response_

_No response_

robinmolen:
Partly closes #5140

**Changes**

Introducing the new `FormAuthenticationBackend` class. This new class works similar to the `FormRegistrationBackend` class, as a standalone object oriented setup for form specific plugin configuration.

`FormAuthenticationBackend` will replace the Form object field `authentication_backend_options`. This field served a similar purpose, only using JSON format to store the configuration information. With the new class based solution, we can perform actual validation on the data and provide a structured way to define configuration options per auth plugin.

Currently the `authentication_backend_options` was only used for loa overriding (only possible when using the DigiD auth plugin). With the upcoming Yivi plugin, the need for structure and proper validation has grown.

This PR contains the following changes:
- Adding a new FormAuthenticationBackend class (with migrations)
- Replacing all instances of `authentication_backend_options` with `FormAuthenticationBackend`
- Removing `authentication_backend_options` from the Form object
- Moved the loa configuration in the admin to a plugin specific modal (as preparation for the Yivi plugin)

**Checklist**

- Impact on features

  - [x] Checked copying a form
  - [X] ~Remove old `authentication_backend_options` field~ Marked `authentication_backend_options` as deprecated and to be removed in 4.0.0

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/authentication/utils.py
```diff
@@ -53,11 +53,17 @@ def is_authenticated_with_an_allowed_plugin(
...
+    request: Request, config: list[FormAuthenticationBackend]
+) -> bool:
     # called after is_authenticated_with_plugin so this is correct
     plugin_id = request.session[FORM_AUTH_SESSION_KEY]["plugin"]
     plugin = auth_register[plugin_id]
-    return plugin.check_requirements(request, config.get(plugin_id, {}))
+    plugin_config = [item for item in config if item.backend == plugin_id]
+
+    return plugin.check_requirements(
+        request, plugin_config[0] if len(plugin_config) > 0 else None
```
vaszig, line 65:
```suggestion
        request, plugin_config[0] if plugin_config else None
```

vaszig (changes requested):

src/openforms/forms/migrations/0105_transform_form_authentication_backend_options.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 4.2.20 on 2025-04-23 10:01
+
+from django.db import migrations
+from django.db.migrations.state import StateApps
+from django.utils.module_loading import import_string
+
+from openforms.upgrades.upgrade_paths import setup_scripts_env
+
+
+class FixScript:
```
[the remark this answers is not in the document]
robinmolen: Ah yeah i wasn't sure if this would be the right way to do this...

A data migration does indeed sound a lot better. I'll change it 👍

src/openforms/forms/models/form.py
```diff
@@ -111,6 +111,8 @@ class Form(models.Model):
     auto_login_authentication_backend = models.CharField(
         _("automatic login"), max_length=UNIQUE_ID_MAX_LENGTH, blank=True
     )
+    # @TODO deprecated, remove in 4.0.0
```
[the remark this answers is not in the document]
robinmolen: Yeah fair point.. I kept it in for "just in case" reasons, but it can/should be removed..

Will do!

src/openforms/js/components/admin/form_design/AuthPluginField.js
```diff
@@ -1,40 +1,106 @@
...
+          <ActionButton
+            className="authentication-plugin__configuration-button"
+            text={intl.formatMessage({
+              description: 'Link label to open authentication options modal',
+              defaultMessage: 'Configure options',
+            })}
+            type="button"
+            onClick={() => setIsModalOpen(true)}
+          />
+          <FormModal
```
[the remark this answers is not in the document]
robinmolen: Mhh yeah fair point...

Currently all changes are immediately "saved", which is (i think) okay for now, seeing as there is only one field for LoA. With the other changes (i.e. adding the Yivi plugin) this modal will contain more fields, and more reasons to save/confirm manually.

I'll add a save button which will work as a confirmation of the changes. (i.e. if you close the modal, without clicking the "save" button, the changes will be forgotten)

src/openforms/authentication/contrib/digid/tests/test_auth_procedure.py
```diff
@@ -167,12 +168,14 @@ def test_authn_request(self, mock_id):
     def test_authn_request_uses_minimal_loa_from_form(self, mock_id):
         form = FormFactory.create(
             authentication_backends=["digid"],
```
sergei-maertens, line 170: isn't this obsoleted by the `FormAuthenticationBackendFactory` line?

DX-wise, it'd probably be nice to be able to do:

```py
FormFactory.create(
    authentication_backend__backend="digid",
    authentication_backend__options={"loa": ...},
)
```
where the `authentication_backend` kwarg is a post-generation that creates the underlying `FormAuthenticationBackend` instance:

```py
@factory.post_generation
def authentication_backend(obj: Form, extracted, create, **kwargs):
    if (extracted or kwargs) and not create:
        raise ValueError("You must use a create strategy")

    if extracted:
        extracted.form = obj
        extracted.save()
    else:
        FormAuthenticationBackendFactory.create(form=obj, **kwargs)
```
something along those lines

src/openforms/authentication/contrib/digid/views.py
```diff
@@ -43,7 +43,12 @@ def get_level_of_assurance(self):
         _, _, kwargs = resolve(return_path)
         form = get_object_or_404(Form, slug=kwargs.get("slug"))
 
-        loa = form.authentication_backend_options.get(PLUGIN_ID, {}).get("loa")
+        loa = None
+        plugin_configuration = form.authentication_backends_configuration.filter(
+            backend=PLUGIN_ID
+        ).first()
```
sergei-maertens, line 49: I'd use a `.get(backend=PLUGIN_ID)` instead to hint that at most one record is expected (and then you'll need a try catch of course)

src/openforms/authentication/utils.py
```diff
@@ -53,11 +53,17 @@ def is_authenticated_with_an_allowed_plugin(
         return False
 
 
-def meets_plugin_requirements(request: Request, config: dict) -> bool:
+def meets_plugin_requirements(
+    request: Request, config: list[FormAuthenticationBackend]
+) -> bool:
```
sergei-maertens, line 58: looking at the places where this is called, we have a `Form` instance in scope, so I think we can update the signature to:

```py
def meets_plugin_requirements(form: Form, plugin: str):
    ...
```
which will simplify the implementation here, since we can now look up the form authentication backend from the form + plugin combination.

The first call-site is straight forward to provide those arguments, the second is a bit more complicated, but I think we should refactor `is_authenticated_with_an_allowed_plugin` to something like `get_authentication_plugin` which returns the plugin ID used for authentication (and raise an exception if a disallowed plugin is used).

We then don't have to pass the request along and poke into the session data all the time, but only have to do it once. (I'm not a fan of the implied race conditions from storing this in the session, so we should reduce our reliance on the session as much as possible).

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,45 @@ class FormLiteralsSerializer(serializers.Serializer):
     confirm_text = ButtonTextSerializer(raw_field="confirm_text", required=False)
 
 
+class FormAuthenticationBackendSerializer(serializers.ModelSerializer):
+    options = serializers.DictField(label=_("authentication backend options"))
```
sergei-maertens, line 59: I'd go with a `serializers.JSONField` so that we have full freedom about the shape of plugin-specific options.

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,45 @@ class FormLiteralsSerializer(serializers.Serializer):
...
+        # set backend choices at runtime
+        fields = super().get_fields()
+        fields["backend"].choices = auth_register.get_choices()
+        return fields
+
+    def validate(self, attrs):
+        # validate config options with selected plugin
+        if "backend" not in attrs:
+            # performing nested PATCH
+            return attrs
```
sergei-maertens, line 78: where does this happend? do we need to support PATCH in this way? This code path prevents us from validating the shape of `options` against the selected `backend`.

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,45 @@ class FormLiteralsSerializer(serializers.Serializer):
...
+
+        serializer = plugin.configuration_options(
+            data=attrs["options"], context=self.context
+        )
+        try:
+            serializer.is_valid(raise_exception=True)
+        except serializers.ValidationError as e:
+            detail = {"options": e.detail}
+            raise serializers.ValidationError(detail) from e
```
sergei-maertens, line 91:
```suggestion
        if not serializer.is_valid():
            raise serializer.ValidationError({"options": serializer.errors})
```
for a simpler variant?

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,45 @@ class FormLiteralsSerializer(serializers.Serializer):
...
+        # serializer does some normalization, so make sure to update the data
+        attrs["options"] = serializer.data
```
sergei-maertens, line 93: this looks wrong - when consuming validated data, you should be using `serializer.validated_data` as that handles the conversion from primitives to Django concepts, but I can see how that would cause problems when trying to save this again into a JSONField :thinking:

src/openforms/forms/api/serializers/form.py
```diff
@@ -121,6 +160,8 @@ class FormSerializer(PublicFieldsSerializerMixin, serializers.ModelSerializer):
         required=False,
         default=list,
     )
+    # @TODO deprecated, remove in 4.0.0
```
sergei-maertens, line 163: please include the literal `DeprecationWarning` string in these TODOs so we can easily CTRL + F them

src/openforms/forms/api/serializers/form.py
```diff
@@ -220,6 +261,9 @@ class FormSerializer(PublicFieldsSerializerMixin, serializers.ModelSerializer):
     translations = ModelTranslationsSerializer()
 
     registration_backends = FormRegistrationBackendSerializer(many=True, required=False)
+    authentication_backends_configuration = FormAuthenticationBackendSerializer(
```
sergei-maertens, line 264: let's go with a shorter name:

```suggestion
    authentication_backends = FormAuthenticationBackendSerializer(
```

src/openforms/forms/migrations/0105_transform_form_authentication_backend_options.py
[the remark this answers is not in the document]
sergei-maertens: data migration is indeed better + you need to write some tests for that too :)

src/openforms/forms/models/form.py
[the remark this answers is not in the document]
sergei-maertens: +1 - after the data migration the field can be removed. w/r to import/export, we should handle that at the serializer level and not the model layer

src/openforms/forms/models/form_authentication_backend.py
```diff
@@ -0,0 +1,40 @@
...
+    BackendChoiceField as AuthenticationBackendChoiceField,
+)
+from . import Form
+
+
+class FormAuthenticationBackend(models.Model):
+    form = models.ForeignKey(
+        Form,
+        on_delete=models.CASCADE,
+        related_name="authentication_backends_configuration",
```
sergei-maertens, line 16: related name is the accessor to go from the form to the backends, so I'd use `auth_backends` here instead so that you can do `form.auth_backends.all()`

src/openforms/forms/models/form_authentication_backend.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+    backend = AuthenticationBackendChoiceField(_("authentication backend"))
+    options = models.JSONField(
+        _("authentication backend options"),
+        default=dict,
+        blank=True,
+    )
+
+    class Meta:
+        unique_together = ("form", "backend")
```
sergei-maertens, line 26: these days `models.UniqueConstraint` is preferred over `unique_together` :)

src/openforms/forms/models/form_authentication_backend.py
```diff
@@ -0,0 +1,40 @@
...
+        return _("{backend} on {form}").format(backend=self.backend, form=self.form)
+
+    def get_backend_display(self):
+        choices = dict(auth_register.get_choices())
+        return choices.get(
+            self.backend,
+            "-",
+        )
+
+    get_backend_display.short_description = _("authentication backend")
```
sergei-maertens, line 40: can we remove this and if we still need it, move it into the `admin.py` instead? The type checker does not like this and it's really only useful for the admin anyway.

src/openforms/forms/tests/factories.py
```diff
@@ -115,6 +123,14 @@ class Meta:
         model = "forms.FormRegistrationBackend"
 
 
+class FormAuthenticationBackendFactory(factory.django.DjangoModelFactory):
+    backend = factory.LazyFunction(random_authentication_plugin)
```
sergei-maertens, line 127: can you check if this can be replaced with `factory.fuzzy.FuzzyChoice` somehow? The `random` module produces non-reproducible results, while factory boy ensures that a seed is set up to feed the RNG for `factory.fuzzy`, meaning that failures due to randomness can be reproduced exactly if you know the random seed (which we print in CI).

src/openforms/forms/utils.py
```diff
@@ -233,6 +233,29 @@ def import_form_data(
                 # possible at this time, so we reset the theme and admins need to update
                 # the imported form.
                 entry["theme"] = None
+                # For backwards capability, transform the `authentication_backend_options`
+                # dict to a `authentication_backends_configuration` list. The form
+                # serializer will handle the rest. This is needed for forms created in
+                # versions before 3.2.0.
+                if "authentication_backend_options" in entry:
```
sergei-maertens, line 240: can we do this in the serializer layer by checking `self.context["is_import"]` instead? That's the way we did this for registration backends too. You can check the 2.8 code for inspiration.

src/openforms/js/components/admin/form_design/AuthPluginField.js
```diff
@@ -1,40 +1,106 @@
...
+          >
+            <Fieldset>
+              {plugin.supportsLoaOverride && (
+                <LoAOverride
+                  prefix={authenticationBackend.prefix}
+                  plugin={plugin}
+                  authenticationBackend={authenticationBackend}
+                  onChange={onChange}
+                />
+              )}
```
sergei-maertens, line 79: let's do this right from the start and allow specifying the form options body from a registry like construct similar to registration options. We can (and should) also refactor those modals to use Formik for sane form management.

src/openforms/js/components/admin/form_design/AuthPluginField.js
```diff
@@ -1,40 +1,106 @@
...
+  errors,
+}) => {
+  const authenticationBackendsWithPrefix = useMemo(
+    () =>
+      authenticationBackends.map((authBackend, index) => ({
+        ...authBackend,
+        prefix: `form.authenticationBackendsConfiguration.${index}.options`,
+      })),
+    [authenticationBackends]
+  );
```
sergei-maertens, line 103: do we really need the `useMemo`? this does not look like an expensive calculation

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -121,13 +122,15 @@ const initialFormState = {
       brpPersonenPurposeLimitationHeaderValue: '',
       brpPersonenProcessingHeaderValue: '',
     },
+    authenticationBackendsConfiguration: [],
```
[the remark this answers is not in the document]
sergei-maertens: I don't see why we need to deprecate this instead of replace it :thinking:

src/openforms/authentication/api/views.py
```diff
@@ -39,6 +39,33 @@ def get_objects(self):
...
+@extend_schema_view(
+    get=extend_schema(summary=_("List available authentication backends")),
+)
+class BackendListView(ListMixin, APIView):
+    """
+    List all authentication plugins that have been registered and can be registered as
+    backends with additional configuration.
+
+    Currently only plugins that support LOA override are returned.
+    """
```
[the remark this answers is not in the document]
robinmolen: I thought it would be a bit cleaner if the plugins that allow additional configuration would be return from a separate endpoint.

I can extend the `PluginListView`, and add a `allows_additional_config` property to the plugins, or something like that

src/openforms/scss/components/admin/_authentication-plugin.scss
```diff
@@ -0,0 +1,11 @@
+@use 'microscope-sass/lib/bem';
+
+.authentication-plugin {
+  padding: 0;
```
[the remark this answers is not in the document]
robinmolen: The `li` html tag has some default styling that adds 1 px to the top- and bottom-padding, which i wanted to removed.

I took a look around and noticed that the radio buttons list for "Authentication automatic login" has the same issue. To keep things consisted, i've removed my new styling rule

bin/transform_form_authentication_backend_options.py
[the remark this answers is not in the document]
robinmolen: Check, yeah i wasn't sure how to do this any other way... But i've changed it now to a regular data migration

src/openforms/authentication/api/tests/test_endpoints.py
```diff
@@ -36,37 +44,58 @@ class DemoAuthPlugin(BasePlugin):
...
         other_user = UserFactory.create(is_staff=True)
 
         self.client.force_authenticate(user=other_user)
 
-        response = self.client.get(self.endpoint)
+        response = self.client.get(reverse_lazy("api:authentication-plugin-list"))
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_backend_list_forbidden_not_authenticated(self):
```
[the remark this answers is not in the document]
robinmolen: Check.

Regarding another comment, the `AuthBackendList` endpoint has been merged into the `AuthPluginList`, so these tests are mostly reverted back to the previous setup (the additional plugin remains, to cover plugins with `configuration_options` configured)

src/openforms/authentication/api/views.py
```diff
@@ -39,6 +39,33 @@ def get_objects(self):
         return list(register.iter_enabled_plugins())
 
 
+@extend_schema_view(
+    get=extend_schema(summary=_("List available authentication backends")),
+)
+class BackendListView(ListMixin, APIView):
```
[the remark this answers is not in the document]
robinmolen: After a discussion with Sergei the `BackendListView` class was removed. The plugins in `PluginListView` now return additional information, making the separate `BackendListView` obsolete

src/openforms/forms/models/form_authentication_backend.py
robinmolen: Yeah thats what i was thinking as well, but form already has a `authentication_backends` for the authentication backends that can be used for authenticating...

So either we rename `authentication_backends` to something like `available_authentication_backends` (because having `authentication_backends` and `auth_backends` on a form seems really error prone), with a big migration/import handling/rewriting a ton of tests.
Or we keep the names a bit more separated, like `auth_backends_config` for the `FormAuthenticationBackend` relation

src/openforms/forms/utils.py
robinmolen: Ah ofcourse, thats also an option.. Man the import really needs a re-write 😉

src/openforms/forms/tests/factories.py
robinmolen: Check, yeah using `factory.fuzzy.FuzzyChoice` seems to work fine 👍
Reproducible randomness sounds quit nice

src/openforms/authentication/api/tests/test_endpoints.py
```diff
@@ -144,5 +193,50 @@ def test_disabled_plugins_are_not_returned(self):
...
+                    {"label": "Stare into the Sun", "value": "∞"},
+                ],
+            },
+        ]
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json(), expected)
+
+    def test_backend_list_only_loa_overridable_plugins_are_returned(self):
+        endpoint = reverse("api:authentication-backend-list")
```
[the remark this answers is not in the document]
robinmolen: Yeah this was part of my initial train of thought, about a separate endpoint that would only return the plugins that would be configurable. Seeing as, at this time, LoA is the only possible configuration, validating that the endpoint would indeed only return LoA overridable plugins seemed correct.

Like we discussed/decided, the "configurability" is now shown/suggested via the plugin schema (the plugin schema describes the shape of configurable options). With this, a separate endpoint doesn't really make sense anymore (or, even less then it might have before).

TLDR; I thought we wanted/needed to separate normal plugins and configurable plugins, so this test was to ensure this functionality. After recent discussions the separate endpoint has been removed, and this test no-longer applies.

src/openforms/authentication/contrib/digid/tests/test_auth_procedure.py
robinmolen: Oh yeah, thats a really good addition. If you don't mind, I'll keep it similar to the registration backend naming: `form__authentication_backend` and `form__authentication_backend_options` (to minimize confusion)

src/openforms/js/components/admin/form_design/AuthPluginField.js
robinmolen: Like discussed, for now we keep the existing options location (without modals, just below the authentication selection).

The plugin options, if supported by the plugin, can be configured in `AuthPluginOptions`. `AuthPluginOptions` will render a simplified form with the available plugin options fields. For now the authenticationBackend data and onchange handlers will be passed to this form, Formik isn't used just yet.


With the changes that will come with the Yivi plugin, we can take another look how this should work (from a UI standpoint, and also technical about Formik usage)

src/openforms/js/components/admin/form_design/AuthPluginField.js
robinmolen: With the change to remove the modal, this isn't needed anymore. But indeed, `useMemo` was a bit expensive/over-kill

src/openforms/forms/api/serializers/form.py
robinmolen: Ah okay, i was looking for a proper deprecation marker, good to know.

Like you suggested, these "deprecated" parts have been removed, as they are now obsolete.

src/openforms/js/components/admin/form_design/form-creation-form.js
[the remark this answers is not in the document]
robinmolen: I removed the `authenticationBackendOptions` entirely (frontend and backend)

src/openforms/authentication/utils.py
robinmolen: Ah cool, i didn't know that that would work.. 👍

src/openforms/js/components/admin/form_design/AuthPluginField.js
```diff
@@ -1,40 +1,106 @@
...
+              <FormattedMessage
+                description="Authentication options modal title"
+                defaultMessage="Plugin configuration: {plugin}"
+                values={{plugin: plugin.label}}
+              />
+            }
+            closeModal={() => setIsModalOpen(false)}
+            extraModifiers={['small']}
+          >
+            <Fieldset>
```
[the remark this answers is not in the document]
robinmolen: Yeah exactly, at the moment only LoA override was a possible configurable part of auth plugins (only for the DigiD plugin). Yivi will add more configuration options/possibilities.

In response to a comment from Sergei, the plugin configuration fields will be defined per plugin. So this semi-hard coded `LoAOverride` field will no-longer be placed for plugins that don't support LoA overriding

src/openforms/forms/api/serializers/form.py
robinmolen: Mhh, yeah the `serializer.data` was an oversight. `serializer.validated_data` seems alright for the simple JSON object for the `loa` data.

I'll change it to `serializer.validated_data` for now, and with the Yivi plugin we can validate this again with bigger/(slightly) more complex JSON objects

src/openforms/forms/api/serializers/form.py
robinmolen: Ah check, yeah `backend` should always be present

src/openforms/authentication/contrib/digid/views.py
```diff
@@ -43,8 +43,12 @@ def get_level_of_assurance(self):
...
         form = get_object_or_404(Form, slug=kwargs.get("slug"))
 
-        loa = form.authentication_backend_options.get(PLUGIN_ID, {}).get("loa")
-        return loa if loa else DIGID_DEFAULT_LOA
+        try:
+            plugin_configuration = form.auth_backends.get(backend=PLUGIN_ID)
+        except FormAuthenticationBackend.DoesNotExist:
+            return DIGID_DEFAULT_LOA
+
+        return plugin_configuration.options.get("loa", DIGID_DEFAULT_LOA)
```
sergei-maertens, line 51: see earlier remark about the `.get` behaviour

src/openforms/authentication/fields.py
```diff
@@ -9,6 +9,15 @@
...
 
 class AuthenticationBackendMultiSelectField(ArrayField):
+    system_check_deprecated_details = {
+        "msg": (
+            "openforms.authentication.fields.AuthenticationBackendMultiSelectField has"
+            "been deprecated. Support for it (except in historical migrations) will be"
+            "removed in Open Forms 3.2.0."
+        ),
+        "id": "authentication.fields.W001",
+    }
```
sergei-maertens, line 19: I think you can do the removed instead of deprecated immediately, we don't need to go through a deprecation period for this.

src/openforms/authentication/registry.py
```diff
@@ -16,8 +16,8 @@
 
 
 def _iter_plugin_ids(form: Form | None, registry: Registry) -> Iterator[str]:
-    if form is not None:
-        yield from form.authentication_backends
+    if form is not None and hasattr(form, "auth_backends"):
+        yield from list(form.auth_backends.values_list("backend", flat=True))
```
sergei-maertens, line 20: no need to call `list`, the queryset itself is an iterator already.

src/openforms/authentication/registry.py
```diff
@@ -16,8 +16,8 @@
 
 
 def _iter_plugin_ids(form: Form | None, registry: Registry) -> Iterator[str]:
-    if form is not None:
-        yield from form.authentication_backends
+    if form is not None and hasattr(form, "auth_backends"):
```
sergei-maertens, line 19: `form` always has `auth_backends` because it's the reverse relation you set up. I think you can add a type annotation on the `Form` model like:

```py
auth_backends: models.Manager[FormAuthenticationBackend]
```
and then the type checker will probably be satisfied :)

src/openforms/authentication/utils.py
```diff
@@ -37,27 +37,41 @@ def store_registrator_details(
...
 
+def get_authentication_plugin(request: Request, form: Form) -> str:
+    try:
+        plugin = request.session[FORM_AUTH_SESSION_KEY]["plugin"]
+        if not form.auth_backends.filter(backend=plugin).exists():
+            raise ValueError(f"Unexpected plugin {plugin} used")
+
+        return plugin
+    except KeyError:
+        raise KeyError("Plugin cannot be retrieved from session")
```
sergei-maertens, line 48:
```suggestion
    except KeyError as exc:
        raise KeyError("Plugin cannot be retrieved from session") from exc
```

src/openforms/authentication/utils.py
```diff
@@ -37,27 +37,41 @@ def store_registrator_details(
...
 
 
-def is_authenticated_with_an_allowed_plugin(
-    request: Request, allowed_plugins: list[str]
-) -> bool:
+def is_authenticated_with_an_allowed_plugin(request: Request, form: Form) -> bool:
     try:
-        return request.session[FORM_AUTH_SESSION_KEY]["plugin"] in allowed_plugins
+        plugin_id = request.session[FORM_AUTH_SESSION_KEY]["plugin"]
+        return form.auth_backends.filter(backend=plugin_id).exists()
```
sergei-maertens, line 61: can't you re-use `get_authentication_plugin` here? The same plugin access/checking code is now repeated in a number of utilities, this is bound to go wrong when some business logic changes and not all places are updated at the same time.

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
sergei-maertens, line 1: please do not update the backend translations except for in release preparation. they cause annoying merge conflicts

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,33 @@ class FormLiteralsSerializer(serializers.Serializer):
...
+    options = serializers.JSONField(
+        label=_("authentication backend options"), allow_null=True, required=False
+    )
+
+    class Meta:
+        model = FormAuthenticationBackend
+        fields = [
+            "backend",
+            "options",
+        ]
```
sergei-maertens, line 68: can you stick to the pattern of using tuples for these sort of declarations? tuples are immutable and convey the nature of this configuration better

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,33 @@ class FormLiteralsSerializer(serializers.Serializer):
     confirm_text = ButtonTextSerializer(raw_field="confirm_text", required=False)
 
 
+class FormAuthenticationBackendSerializer(serializers.ModelSerializer):
+    options = serializers.JSONField(
+        label=_("authentication backend options"), allow_null=True, required=False
+    )
```
sergei-maertens, line 61: I notice this accepts `null`, but the model field has a default of `dict`. What's the intention here?

src/openforms/forms/api/serializers/form.py
```diff
@@ -55,6 +55,33 @@ class FormLiteralsSerializer(serializers.Serializer):
...
+        model = FormAuthenticationBackend
+        fields = [
+            "backend",
+            "options",
+        ]
+
+    def validate(self, attrs):
+        plugin = auth_register[attrs["backend"]]
+        if not hasattr(plugin, "configuration_options") or "options" not in attrs:
+            return attrs
```
sergei-maertens, line 73: I think bailing out if options are not provided (PATCH calls) is an edge case that can be wrong - if there are existing options but you change the plugin, then the options won't validate for that plugin.

During partial updates, you probably want to look up the existing instance to grab the unspecified fields to be able to run the full validation. See https://github.com/open-formulieren/open-forms/blob/09a42eced25c49f2623bd62b2ce46c32f32360bf/src/openforms/api/utils.py#L26

src/openforms/forms/migrations/0105_data_migrate_form_authentication_backend_options_to_form_authentication_backend.py
```diff
@@ -0,0 +1,97 @@
...
+        for plugin in plugins:
+            form_auths_to_create[plugin] = {}
+        for plugin, options in config.items():
+            form_auths_to_create[plugin] = options
+
+        for plugin, options in form_auths_to_create.items():
+            # Only create a FormAuthenticationBackend if it doesn't exist already
+            if not FormAuthenticationBackend.objects.filter(
+                form=form, backend=plugin
+            ).exists():
```
sergei-maertens, line 34: why would these exist? the preceding migration just created the new tables so there cannot possibly be any existing data, right?

src/openforms/forms/migrations/0105_data_migrate_form_authentication_backend_options_to_form_authentication_backend.py
```diff
@@ -0,0 +1,97 @@
...
+                form_authentication_backends_to_create.append(
+                    FormAuthenticationBackend(
+                        form=form, backend=plugin, options=options
+                    )
+                )
+
+        form.authentication_backends = []
+        form.authentication_backend_options = {}
```
sergei-maertens, line 42: I wouldn't reset these in the data migration - if something goes wrong, there's a snapshot to recover the configuration and the columns will be dropped in a future migration anyway.

src/openforms/forms/migrations/0105_data_migrate_form_authentication_backend_options_to_form_authentication_backend.py
```diff
@@ -0,0 +1,97 @@
...
+
+    forms_to_update_map = {}
+    form_authentication_backends = FormAuthenticationBackend.objects.iterator()
+
+    for form_authentication_backend in form_authentication_backends:
+        if form_authentication_backend.form.id not in forms_to_update_map:
+            # Store the form in a dict, so it can easily be found/updated
+            forms_to_update_map[form_authentication_backend.form.id] = (
+                form_authentication_backend.form
+            )
```
sergei-maertens, line 68: you could also loop over the form instances to make it easier to handle each form, or use `itertools.groupby` to handle one form at a time, which will simplify the migration code

src/openforms/forms/migrations/0105_data_migrate_form_authentication_backend_options_to_form_authentication_backend.py
```diff
@@ -0,0 +1,97 @@
...
+        form = forms_to_update_map[form_authentication_backend.form.id]
+        config = form.authentication_backend_options
+
+        form.authentication_backends.append(form_authentication_backend.backend)
+        config[form_authentication_backend.backend] = (
+            form_authentication_backend.options
+        )
+
+    # All FormAuthenticationBackend should have been reversed, so we can delete them all
+    FormAuthenticationBackend.objects.all().delete()
```
sergei-maertens, line 80: no need, reversing to the previous migration will drop the table

src/openforms/forms/models/form.py
```diff
@@ -455,16 +448,16 @@ def get_payment_backend_display(self):
     get_payment_backend_display.short_description = _("payment backend")
 
     def get_authentication_backends_display(self):
-        if not self.authentication_backends:
+        if not self.auth_backends:
             return "-"
```
sergei-maertens, line 452: `self.auth_backends` is always truthy

you'll also want to `prefetch_related` the queryset you're using for the admin, otherwise you end up with the class `n + 1` queries problem

src/openforms/forms/models/form_authentication_backend.py
```diff
@@ -0,0 +1,43 @@
...
+        Form,
+        on_delete=models.CASCADE,
+        related_name="auth_backends",
+    )
+    backend = AuthenticationBackendChoiceField(_("authentication backend"))
+    options = models.JSONField(
+        _("authentication backend options"),
+        blank=True,
+        null=True,
+    )
```
sergei-maertens, line 23: my earlier remark about `default=dict` is wrong, not sure where I saw that, maybe I'm confusing it with another PR!

src/openforms/forms/models/form_authentication_backend.py
```diff
@@ -0,0 +1,43 @@
...
+        verbose_name_plural = _("authentication backends")
+        constraints = [
+            models.UniqueConstraint(
+                fields=["form", "backend"],
+                name="form_backend_unique_together",
+            ),
+        ]
+
+    def __str__(self):
+        return _("{backend} on {form}").format(backend=self.backend, form=self.form)
```
sergei-maertens, line 36: I think we want to use the verbose name of the plugin rather than the identifier here?

Also - note that this will crash for unsaved instances that don't have a form set yet:

```py
str(FormAuthenticationBackend())
```
usually in the __str__` you want to avoid looking up relations/foreign keys

src/openforms/forms/models/form_authentication_backend.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def get_backend_display(self):
+        choices = dict(auth_register.get_choices())
```
sergei-maertens, line 39: can we use `auth_register[self.backend]` for some nicer syntactic sugar?

src/openforms/forms/tests/admin/test_form.py
```diff
@@ -60,7 +60,15 @@ def test_form_admin_export(self):
         forms = json.loads(zf.read("forms.json"))
         self.assertEqual(len(forms), 1)
         self.assertEqual(forms[0]["uuid"], str(form.uuid))
-        self.assertEqual(forms[0]["authentication_backends"], ["digid"])
+        self.assertEqual(
+            forms[0]["auth_backends"],
+            [
+                {
+                    "backend": "digid",
+                    "options": {},
```
sergei-maertens, line 68: I'd expect `loa` to be in here

src/openforms/forms/tests/admin/test_form.py
```diff
@@ -128,7 +136,8 @@ def test_form_admin_import(self):
         self.assertNotEqual(form.uuid, "b8315e1d-3134-476f-8786-7661d8237c51")
         self.assertEqual(form.name, "Form 000")
         self.assertEqual(form.internal_name, "Form internal")
-        self.assertEqual(form.authentication_backends, ["digid"])
+        self.assertEqual(form.auth_backends.count(), 1)
+        self.assertEqual(form.auth_backends.first().backend, "digid")
```
sergei-maertens, line 140:
```suggestion
        self.assertEqual(form.auth_backends.get().backend, "digid")
```
avoid the type checker error that `NoneType` doesn't have the attribute `backend`

src/openforms/forms/tests/factories.py
```diff
@@ -115,6 +135,14 @@ class Meta:
         model = "forms.FormRegistrationBackend"
 
 
+class FormAuthenticationBackendFactory(factory.django.DjangoModelFactory):
+    backend = factory.fuzzy.FuzzyChoice(authentication_plugins())
```
sergei-maertens, line 139: do we need to *call* `authentication_plugins` or can you also provide it as a callback? Generally you want to avoid calling functions at the module level.

src/openforms/forms/tests/factories.py
```diff
@@ -115,6 +135,14 @@ class Meta:
...
+    form = factory.SubFactory(FormFactory)
+
+    class Meta:
+        model = "forms.FormAuthenticationBackend"
```
sergei-maertens, line 143: you'll want to specify `django_get_or_create` too because there's a unique constraint on `form + backend` and otherwise this will introduce flaky tests

src/openforms/js/components/admin/form_design/AuthPluginField.js
```diff
@@ -24,15 +24,16 @@ const AuthPluginField = ({availableAuthPlugins, selectedAuthPlugins, onChange, e
...
-        label={label}
-        onChange={onChange}
-        checked={selectedAuthPlugins.includes(plugin.id)}
-        noVCheckbox
-      />
+      <li key={plugin.id}>
+        <Checkbox
+          name={plugin.label}
+          value={plugin.id}
+          label={label}
```
sergei-maertens, line 31: can we inline that variable from above? the nested fragments are making my head fuzzy

src/openforms/js/components/admin/form_design/authentication/LoAOverride.js
```diff
@@ -0,0 +1,54 @@
...
+        value={loa}
+        onChange={onChange}
+        allowBlank={true}
+        choices={plugin.assuranceLevels.map(loa => [loa.value, loa.label])}
+      />
+    </Field>
+  </FormRow>
+);
+
+export default LoAOverride;
```
sergei-maertens, line 36: can you move that to the bottom? it's a bit weird to see more code after the (default) export(s)

sergei-maertens (changes requested):

src/openforms/forms/tests/factories.py
robinmolen: Ah check, will do

src/openforms/authentication/api/serializers.py
```diff
@@ -35,6 +35,12 @@ class AuthPluginSerializer(PluginBaseSerializer):
         label=_("Levels of assurance"),
         help_text=_("The levels of assurance this plugin defines."),
     )
+    schema = serializers.DictField(
```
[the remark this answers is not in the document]
sergei-maertens: I think it's even more concrete that you just implement the configuration forms correctly for each plugin - the generic nature of these option forms is gone :)

src/openforms/authentication/api/tests/test_endpoints.py
```diff
@@ -36,10 +55,11 @@ class DemoAuthPlugin(BasePlugin):
 register = Registry()
 register("plugin1")(SingleAuthPlugin)
 register("plugin2")(DemoAuthPlugin)
+register("plugin3")(AdditionalConfigAuthPlugin)
 
 
 class AuthTests(APITestCase):
-    endpoint = reverse_lazy("api:authentication-plugin-list")
+    endpoint = reverse("api:authentication-plugin-list")
```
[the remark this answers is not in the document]
robinmolen: I've changed this to non-lazy due to your previous comment https://github.com/open-formulieren/open-forms/pull/5295#discussion_r2063428608.

Is it a "test instance" vs "test class" case? Where non-lazy should be used in a test, but the lazy variant should be used in the test class?

src/openforms/authentication/api/serializers.py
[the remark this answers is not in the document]
robinmolen: Yeah, it might be the best move to go towards a true generic setup.

I think i've replaced/removed all `supports_loa_override` and `assurance_levels` usage

src/openforms/authentication/api/tests/test_endpoints.py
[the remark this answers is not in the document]
robinmolen: Check.. okay get it

src/openforms/authentication/contrib/digid/plugin.py
```diff
@@ -94,10 +96,10 @@ def handle_return(self, request, form):
 
         return HttpResponseRedirect(form_url)
 
-    def check_requirements(self, request: AnyRequest, config: dict) -> bool:
+    def check_requirements(self, request: AnyRequest, options: DigidOptions) -> bool:
         # check LoA requirements
         authenticated_loa = request.session[FORM_AUTH_SESSION_KEY]["loa"]
-        required = config.get("loa") or DIGID_DEFAULT_LOA
+        required = options.get("loa", DIGID_DEFAULT_LOA)
```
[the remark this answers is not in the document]
robinmolen: But this would crash and burn if `options.loa` isn't defined, right? That's why i opted to `dict.get` (although, it should indeed have been `dict.get(key) or default`)

Maybe i missed something with the digid options serializer setup, but if i don't define loa, it isn't set by the serializer `default` and causing this to crash. I'll take a look if this is indeed something i missed, or if crashing is "supposed to happen" in this case.

src/openforms/authentication/contrib/digid/plugin.py
```diff
@@ -29,11 +30,12 @@ def loa_order(loa: str) -> int:
 
 
 @register(PLUGIN_ID)
-class DigidAuthentication(BasePlugin):
+class DigidAuthentication(BasePlugin[DigidOptions]):
     verbose_name = _("DigiD")
     provides_auth = AuthAttribute.bsn
     supports_loa_override = True
```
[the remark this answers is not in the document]
robinmolen: You mean the `supports_loa_override`? With your previous comment i've removed all `supports_loa_override` and `assurance_levels` definitions, so this will be removed in the next commit/rebase/version of this PR

src/openforms/authentication/contrib/digid/views.py
robinmolen: Check, i've changed it to a `options["loa"] or DIGID_DEFAULT_LOA`

src/openforms/authentication/contrib/digid/plugin.py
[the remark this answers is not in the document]
sergei-maertens: the serializer defines a `default`, so the key will always be defined when you deserialize the options again. This is similar with the registrations: https://github.com/open-formulieren/open-forms/blob/3bf1c17d12962a57d6abce59d51ef4901cd50f8c/src/openforms/registrations/tasks.py#L96

When the serializer has a default, you can drop the `NotRequired` in the `OptionsT`, so you can do direct access without needing to fall back because we have code that ensures our options are in a consistent state.

src/openforms/authentication/contrib/digid/views.py
```diff
@@ -43,8 +43,12 @@ def get_level_of_assurance(self):
         _, _, kwargs = resolve(return_path)
         form = get_object_or_404(Form, slug=kwargs.get("slug"))
 
-        loa = form.authentication_backend_options.get(PLUGIN_ID, {}).get("loa")
-        return loa if loa else DIGID_DEFAULT_LOA
+        try:
+            plugin_configuration = form.auth_backends.get(backend=PLUGIN_ID)
+        except FormAuthenticationBackend.DoesNotExist:
+            return DIGID_DEFAULT_LOA
```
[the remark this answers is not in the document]
robinmolen: Fair this should indeed not be possible, but i think try/except would still be the correct implementation. And indeed, instead of returning the `DIGID_DEFAULT_LOA`, we should raise an error.

src/openforms/authentication/contrib/digid_eherkenning_oidc/admin.py
```diff
@@ -54,7 +54,7 @@ def clean_enabled(self):
         # deteermine which plugin ID we need to query for
         plugin = get_config_to_plugin()[self._meta.model]
         forms_with_backend = Form.objects.live().filter(
-            authentication_backends__contains=[plugin.identifier]
+            auth_backends__backend__contains=plugin.identifier,
```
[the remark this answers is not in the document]
robinmolen: Ah chips, changed it!

src/openforms/authentication/fields.py
robinmolen: I've added this deprecation notice because the field `AuthenticationBackendMultiSelectField` is used in migration `openforms.forms.migrations.0001_initial_to_v250.py`. As we discussed last week, this can/should be removed with a migration squashing round, but i don't think that this should happen right now.

src/openforms/authentication/registry.py
robinmolen: Ahh okay check

src/openforms/authentication/utils.py
robinmolen: Oh damn, yeah thats real risky.. Changing it

src/openforms/forms/api/serializers/form.py
robinmolen: I think i can ignore this https://github.com/open-formulieren/open-forms/pull/5295#discussion_r2075195029

src/openforms/authentication/registry.py
robinmolen: I have a setup but its not perfect. Started with circular imports issues, which i fixed by importing the `FormAuthenticationBackend` only when `TYPE_CHECKING`. But this results into incorrect interpretation of `form.auth_backends.get(...)`.

I've spend quite some time on this, but can't get it to work nicely. Lets discuss it tomorrow

src/openforms/forms/api/serializers/form.py
robinmolen: Ah yeah, i took a look at this this morning, and started with some implementation of correctly handling options when undefined. The `get_from_serializer_data_or_instance` seems like a good fit

src/openforms/forms/migrations/0105_data_migrate_form_authentication_backend_options_to_form_authentication_backend.py
robinmolen: Mhh check... Yeah i added this condition due to existing test/development data. But in a real situation this is indeed not needed.

src/openforms/authentication/fields.py
sergei-maertens: I thought the link I shared had an example of removal notice instead of deprecation too, but maybe I misremember

src/openforms/forms/migrations/0105_data_migrate_form_authentication_backend_options_to_form_authentication_backend.py
robinmolen: I've opted for `itertools.groupby`, which indeed cleans it up nicely 👌

src/openforms/forms/models/form_authentication_backend.py
robinmolen: Ah check, yeah for this i used the `FormRegistrationBackend` as guideline (to keep the output and code similar).

I think we should keep form in the stringified version, as this is some crucial information about the FormAuthenticationBackend. I've added some checks and fallbacks in case form isn't available (or when backend is undefined, which would be the case in your example. Maybe a bit overkill, but it simultaneously makes sure that the provided plugin is correct/defined in the registry)

src/openforms/forms/tests/admin/test_form.py
robinmolen: This is currently only done in the api serializer. This would, i think, require a custom validator that validates and updates the options based on the chosen backend before saving the `FormAuthenticationBackend` object.

sergei-maertens: you need to provide accurate test data in tests :) in the registration tests, we use the type hints for this: https://github.com/open-formulieren/open-forms/blob/3bf1c17d12962a57d6abce59d51ef4901cd50f8c/src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py#L84

src/openforms/forms/models/form.py
robinmolen: I've updated the condition to check if any auth backends have been selected.

Im not sure which queryset you are referring to? The `self.auth_backends.values_list("backend", flat=True)` is a list of plugin keys, and i'm not sure what to prefetch from the authentication register.

src/openforms/authentication/fields.py
robinmolen: Ah yeah i can change it to `system_check_removed_details`, if thats what you mean?

sergei-maertens (commented):
Nice work - some more finishing touches and then we can merge it!

src/openforms/authentication/utils.py
```diff
@@ -37,27 +37,41 @@ def store_registrator_details(
...
-def meets_plugin_requirements(request: Request, config: dict) -> bool:
+def meets_plugin_requirements(request: Request, form: Form, plugin_id: str) -> bool:
     # called after is_authenticated_with_plugin so this is correct
-    plugin_id = request.session[FORM_AUTH_SESSION_KEY]["plugin"]
     plugin = auth_register[plugin_id]
-    return plugin.check_requirements(request, config.get(plugin_id, {}))
+    try:
+        plugin_configuration = form.auth_backends.get(backend=plugin_id)
+    except FormAuthenticationBackend.DoesNotExist:
+        return False
```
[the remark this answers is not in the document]
robinmolen: I took a look if this could be covered, but this scenario should not be possible. This function is always called after `get_authentication_plugin`, which also checks if `FormAuthenticationBackend` for this plugin exists.

I've removed the try/except block and added a comment in the code describing why try/except isn't needed.

src/openforms/authentication/contrib/digid/views.py
```diff
@@ -42,8 +43,11 @@ def get_level_of_assurance(self):
         _, _, kwargs = resolve(return_path)
         form = get_object_or_404(Form, slug=kwargs.get("slug"))
 
-        loa = form.authentication_backend_options.get(PLUGIN_ID, {}).get("loa")
-        return loa if loa else DIGID_DEFAULT_LOA
+        try:
+            auth_backend = form.auth_backends.get(backend=PLUGIN_ID)
+            return auth_backend.options["loa"] or DIGID_DEFAULT_LOA
+        except FormAuthenticationBackend.DoesNotExist:
+            raise PluginNotEnabled(_("DigiD is not enabled for this form."))
```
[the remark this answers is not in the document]
robinmolen: Like the `authentication/utils.py`, this scenario is also unpossible. `get_level_of_assurance` gets executed/called after the generic authentication login (`AuthenticationStartView.get()`), which also checks if the plugin is registered for the form.

Just like the `authentication/utils.py`, I've removed the try/except block and added a comment describing the situation. I've also added a test that validates that DigiD login, when the digid plugin isn't active for the form, results into a 400 "Bad Request" error.

src/openforms/authentication/contrib/digid/views.py
```diff
@@ -42,8 +42,10 @@ def get_level_of_assurance(self):
         _, _, kwargs = resolve(return_path)
         form = get_object_or_404(Form, slug=kwargs.get("slug"))
 
-        loa = form.authentication_backend_options.get(PLUGIN_ID, {}).get("loa")
-        return loa if loa else DIGID_DEFAULT_LOA
+        # called after AuthenticationStartView.get(), which already checks if there is an
+        # FormAuthenticationBackend object for the plugin
+        auth_backend = form.auth_backends.get(backend=PLUGIN_ID)
+        return auth_backend.options["loa"] or DIGID_DEFAULT_LOA
```
sergei-maertens, line 48: ah but you can still trigger errors this way :) Try navigating directly to `http://localhost:8000/digid/login/?next=http://localhost:8000/some-test-form-slug/` in your browser!

sergei-maertens (approved):
As discussed in the office - handle the `FormAuthenticationBackend.DoesNotExist` so that it returns an HTTP 400 and then it's good to merge :)

PR #5298, opened 2025-04-24
:bug: Fix HTML sanitization of design tokens

sergei-maertens:
The previous code assumed values were always strings, but they can technically also be numbers, null or booleans, as those are the valid JSON primitives.

Was reported via Sentry for DH.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/ui/templatetags/style_dictionary.py
+++ b/src/openforms/ui/templatetags/style_dictionary.py
@@ -1,7 +1,7 @@
 from django import template
 from django.utils.safestring import mark_safe
 
-from openforms.typing import JSONPrimitive
+from openforms.typing import JSONPrimitive, JSONValue
 
 register = template.Library()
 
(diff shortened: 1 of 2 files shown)


vaszig (approved):
I cannot reproduce the failing test locally and don't see what it has to do with your fix. I guess it's flakiness.

Concerning the backport I think we do not need it (agree with your comment on slack).

sergei-maertens:
> I cannot reproduce the failing test locally and don't see what it has to do with your fix. I guess it's flakiness.
> 
> Concerning the backport I think we do not need it (agree with your comment on slack).

To make this public on GitHub, possible workarounds are:

* Use a string value, eg 0 can be replaced with '0px'
* Define the design tokens in a stylesheet instead of the JSON in the admin (preferred)

PR #5306, opened 2025-05-02
Enable most of the bugbear linter rules

sergei-maertens:
See https://docs.astral.sh/ruff/rules/#flake8-bugbear-b

[skip: e2e]


pyproject.toml
```diff
@@ -86,9 +86,13 @@ extend-select = [
...
-    "B006",# flake8-bugbear
-    "B010",
-    # "B904",
+    "B",     # flake8-bugbear
+]
+ignore = [
+    "B905",
+    "B017",
+    "B018",
+    "B904",
```
sergei-maertens, line 95: we'll have to address these at a later time :)

viktorvanwijk: Because we can't right now, or do they require significant more work? B905 doesn't seem super big, or am I missing something?

sergei-maertens: there's a lot of violations and very little time :P 

Right now I'd rather "plug the hole" as much as possible and then drain the boat later :)

sergei-maertens: ah right, the auto-fix just makes it `strict=False`, lol. I'll enable that one too :)

viktorvanwijk (approved):

PR #5317, opened 2025-05-13
Produce more useful logs in registrations process

### LINKED ISSUE #5285: Adapt and switch to structlog
Structlog provides a better DX and has the options to emit stdlib logging records as JSON (to stdout/stderr). JSON output makes log aggregation in a cloud-native environment better.

**Tasks**

- [x] Set up and configure structlog
- [x] Refactor existing logging to provide necessary context
- [x] How does it interact with Sentry log captures? (check https://github.com/kiwicom/structlog-sentry)
- [x] Celery?
- [x] uwsgi logs as JSON
- [x] Clean up flaky test logging

sergei-maertens:
Closes #5285 [skip: e2e]

**Changes**

* Added more context and log entries
* Fixed non-VCR tests
* Added logs specifically for the Objects API registration

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/plugin.py
+++ b/src/openforms/registrations/contrib/objects_api/plugin.py
@@ -88,10 +88,15 @@ class ObjectsAPIRegistration(BasePlugin[RegistrationOptions]):
         """
 
         self.set_defaults(options)
+        log = logger.bind(
+            handler_version=options["version"],
+            public_reference=submission.public_registration_reference,
+        )
 
         handler = HANDLER_MAPPING[options["version"]]
 
         handler.save_registration_data(submission, options)
+        log.debug("registration_data_saved")
 
         record_data = handler.get_record_data(
             submission=submission,
@@ -101,6 +106,7 @@ class ObjectsAPIRegistration(BasePlugin[RegistrationOptions]):
         with get_objecttypes_client(options["objects_api_group"]) as objecttypes_client:
             objecttype = objecttypes_client.get_objecttype(options["objecttype"])
             objecttype_url = objecttype["url"]
+            log.info("objecttype_resolved", objecttype_url=objecttype_url)
 
         with get_objects_client(options["objects_api_group"]) as objects_client:
             # update or create the object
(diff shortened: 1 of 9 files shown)


src/openforms/registrations/contrib/objects_api/plugin.py
```diff
@@ -88,10 +88,15 @@ def register_submission(
...
         self.set_defaults(options)
+        log = logger.bind(
+            handler_version=options["version"],
+            public_reference=submission.public_registration_reference,
+        )
 
         handler = HANDLER_MAPPING[options["version"]]
 
         handler.save_registration_data(submission, options)
+        log.debug("registration_data_saved")
```
vaszig, line 99: I do not see an exception handler in the `save_registration_data`, are we sure that `registration_data_saved` or should we have something like "trying" to save the data?

src/openforms/registrations/contrib/objects_api/tests/test_template.py
```diff
@@ -16,17 +15,42 @@
...
 from ..models import ObjectsAPIConfig
 from ..plugin import PLUGIN_IDENTIFIER, ObjectsAPIRegistration
+from ..typing import RegistrationOptionsV1
 
+TEST_FILES = Path(__file__).parent / "files"
 
-class JSONTemplatingTests(TestCase):
+
+class JSONTemplatingTests(OFVCRMixin, TestCase):
     maxDiff = None
```
vaszig, line 28: I guess we do not need that but it's ok if we do not have to change anything else.

vaszig (commented):
I guess Objects API is the only one which was missing some logs, or it's just the first one that you handled?

src/openforms/registrations/contrib/objects_api/plugin.py
sergei-maertens: yeah so if it crashes, the log will never be emitted :)

src/openforms/registrations/contrib/objects_api/tests/test_template.py
sergei-maertens: oh I definitely needed it while adapting the tests to see the full diff! there was quite a bit wrong in the original assertions

vaszig (approved):

PR #5321, opened 2025-05-14
Fix type checking

sergei-maertens:
[skip: e2e]


src/openforms/prefill/contrib/family_members/filters.py
```diff
@@ -1,19 +1,21 @@
...
 class HasAge(Protocol):
     @property
     def age(self) -> int | None: ...
 
 
-T = TypeVar("T", bound=HasAge)
-
-
-def filter_members_by_age(
+def filter_members_by_age[T: HasAge](
```
vaszig, line 10: I keep forgetting that we use python 3.12 and the new syntax..

src/openforms/prefill/contrib/family_members/haal_centraal.py
```diff
@@ -9,58 +9,44 @@
...
 
 logger = structlog.stdlib.get_logger(__name__)
 
 
 def get_data_from_haal_centraal(
-    bsn: str,
-    options: FamilyMembersPartnerOptions | FamilyMembersChildOptions,
-) -> dict[str, list[NaturalPersonDetails]]:
+    bsn: str, options: FamilyMemberOptions
+) -> list[NaturalPersonDetails]:
```
vaszig, line 19: I had the dict in order to define specifically if it's `partners` or `children` for the frontend, but I guess we will know with this way as well.

src/openforms/prefill/contrib/family_members/haal_centraal.py
```diff
@@ -9,58 +9,44 @@
...
-
-                children_list = client.get_family_members(
+                children = client.get_family_members(
                     bsn,
                     include_partner=False,
                     include_children=True,
-                    include_deceased=(
-                        include_deceased if include_deceased is not None else True
-                    ),
+                    include_deceased=options["include_deceased"],
```
vaszig, line 41: So happy that if statements are removed now!

src/openforms/prefill/contrib/family_members/tests/test_filters.py
```diff
@@ -19,29 +19,25 @@ def test_age_filter_with_empty_input(self):
...
-                bsn="123456788",
-                first_names="Doe",
-                lastname="Pirsen",
-                date_of_birth="",
-            ),
-        ]
-
-        for data in (HC_DATA, STUF_BG_DATA):
-            with self.subTest():
+        hc_item = HC_NaturalPersonDetails(
```
vaszig, line 22: I guess I will update the other tests too with this naming

vaszig (approved):

src/openforms/prefill/contrib/family_members/haal_centraal.py
sergei-maertens: Discussed on Slack - this is still the same shape except the assignment is moved into the plugin code rather than the handler.

PR #5331, opened 2025-05-21
Enable extra type checking

sergei-maertens:
Hobby work - low prio!

This addresses some more type errors and increases the scope of files being type checked. Whenever we now add a new registration plugin or make changes, they shall by default be type checked.

[skip: e2e]


src/openforms/registrations/contrib/objects_api/handlers/v1.py
```diff
@@ -0,0 +1,109 @@
...
+from openforms.authentication.service import AuthAttribute
+from openforms.contrib.objects_api.rendering import render_to_json
+from openforms.submissions.models import Submission
+from openforms.typing import JSONObject
+from openforms.variables.utils import get_variables_for_context
+
+from ..registration_variables import get_cosign_value
+
+
+class PaymentContextData(TypedDict):
```
vaszig, line 19: Wouldn't be better if we had these TypeDicts in a separate file?

src/openforms/registrations/contrib/objects_api/templatetags/registrations/contrib/objects_api/json_tags.py
```diff
@@ -8,6 +8,8 @@
 from openforms.registrations.contrib.objects_api.utils import html_escape_json
 from openforms.typing import JSONEncodable
 
+from .....handlers.v1 import SubmissionContext
```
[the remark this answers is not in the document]
vaszig: Yeah..I can understand!Maybe if we had the separate files that I mentioned above it would be prettier!

src/openforms/registrations/contrib/objects_api/tests/test_api_group_model.py
```diff
@@ -6,10 +6,10 @@
 from django.test import TestCase, override_settings
 from django.utils.translation import gettext as _
 
+from openforms.contrib.objects_api.models import ObjectsAPIGroupConfig
```
vaszig, line 9: I see all these places that I was importing the `ObjectsAPIGroupConfig` from the registrations module and I am trying to remember why I had done it in this way..maybe beacuse of the migrations problems?

src/openforms/typing.py
```diff
@@ -3,7 +3,8 @@
 import datetime
 import decimal
 import uuid
-from typing import TYPE_CHECKING, Any, NewType, Protocol, Sequence
+from collections.abc import Sequence
```
vaszig, line 6: I thought it was the same if we import it from typing

vaszig (commented):
More types!Nice

src/openforms/registrations/contrib/objects_api/tests/test_api_group_model.py
sergei-maertens: yeah I also think it was migrations related, I know it was deliberate!

src/openforms/typing.py
sergei-maertens: they're aliases, but the canonical place is `collections.abc`.

In the future, this will be addressed with the Ruff `UP` rules too I think: https://docs.astral.sh/ruff/rules/#pyupgrade-up

src/openforms/registrations/contrib/objects_api/handlers/v1.py
vaszig: I didn't think the possibility that someone will make use of them, let's keep it as is then, I agree.

vaszig (approved):

PR #5339, opened 2025-05-23
Amend documentation of HC BRP docker compose

viktorvanwijk:
[skip: e2e]

**Changes**

Clarified documentation of the Haal Centraal BRP docker compose

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


vaszig (commented):
Nice update!Thanks Victor!

The only thing I have to mention is that I would create a temp directory and not messing with the history of Open Forms repo (even locally). This is indeed something done for this repository but I am not sure what is the right approach. @sergei-maertens what do you think?

docker/haal_centraal_brp/README.md
```diff
@@ -21,38 +21,54 @@ This brings up the service and you can now make API calls by using
...
-- create a patch file (_some-change-bsnNumber.patch_) and save it to open-forms/patches directory by
-  using the last commit (with the changes ou made)
+- Commit the file to ensure changes are being tracked (this is only a temporary commit).
+- Modify the data of the local file according to your needs.
+- Commit the file again with the changes (this is only a temporary commit).
+- Create a patch file (_some-change-bsnNumber.patch_) and save it to _open-forms/patches_ directory by
+  using the last commit (with the changes you made).
 
-  `git format-patch -1`
+  `git format-patch -1 --stdout > /path/to/open-forms/patches/some-change-bsnNumber.patch`
```
sergei-maertens, line 48: I don't think the second commit is necessary, some for of

```bash
git diff > change.patch
```
should be sufficient to record the changes into a patch file

A helper to apply the existing patches also wouldn't be wrong, you can use https://github.com/open-formulieren/open-forms/blob/archive/2.4.x/patches/apply.sh for reference :-)

In general, rather than doing these manual instructions I'd consider writing a bash script to automate this, it could live in the `bin` folder or in the `docker/haal_centraal_brp` folder

vaszig: Ok! @viktorvanwijk you can assign this to me if you prefer since this was my task.

docker/haal_centraal_brp/README.md
```diff
@@ -21,40 +21,58 @@ This brings up the service and you can now make API calls by using
...
 
-- restart the container and you can test your new test cases
+- Commit and push the patch file for others to use (if applicable).
 
-#### In case you want to apply a new patch you follow these steps:
+#### Applying an existing patch:
 
-- apply the patch to your local json file
+- If not done already, copy the _/app/Data/test-data.json_ file into your local machine. This must
+  be the initial file, without any changes/patches already applied and inside a directory tracked by
```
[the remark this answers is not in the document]
viktorvanwijk: OK, thanks :). I usually manage my containers with the Docker Desktop app, which does keep the changes, but I can confirm that it doesn't when using the command line (as is described in the readme :))

PR #5352, opened 2025-05-27
Refactor OIDC models

SilviaAmAm:
Closes https://github.com/maykinmedia/django-digid-eherkenning/issues/96

[skip: e2e]

**Changes**

Rebase code in progress: feature/rebase-in-progress
continue fixing tests: src/openforms/authentication/contrib/yivi_oidc/tests/test_plugin.py `YiviPluginCheckRequirementsTest`

Still work in progress. I still need to:

- [x] ~Figure out infinite redirect because of 2FA~ Problem solved itself ¯\\_(ツ)_/¯.
- [x] Add backward migration and migration to delete the old models
- [x] ~Check that oidc_plugin does not import anything from OF (see if you can use ruff lint to check)~ Moved inside OF
- [x] Write the data migrations to convert from the old models to the new ones
- [x] Migrations to delete the old models
- [x] Test the refactored code: Move these tests to OF https://github.com/maykinmedia/django-digid-eherkenning/tree/master/tests/oidc
- [x] Test the migrations
- [x] Re-record cassettes
- [x] Docs

**Checklist**

- Impact on features

  - [x] Checked copying a form
  - [x] Problem detection in the admin email digest is handled (I think that this is irrelevant?)

- Release management

- Dockerfile/scripts

- Commit hygiene

**Testing**
- [x] We don't have a good way to test the cancel flow. This should be checked manually.
- [x] Test migrations with dumps from real DBs

**Other**
- [x] Remember to remove Git from the packages installed in the Dockerfile


src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
```diff
@@ -0,0 +1,395 @@
...
+                    "legacy": "",
+                },
+                {
+                    "path": config.options["identity_settings"]["kvk_claim_path"],
+                    "legacy": "",
+                },
+                {
+                    "path": config.options["identity_settings"]["pseudo_claim_path"],
+                    "legacy": "",
+                },
```
robinmolen, line 221: Should the additional attributes from `get_additional_attributes()` also be added to the `optional_claims`?

robinmolen (changes requested):
Some small tweaking comments

src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
```diff
@@ -0,0 +1,346 @@
...
+from openforms.forms.models.form_authentication_backend import FormAuthenticationBackend
+
+from .....contrib.auth_oidc.views import anon_user_callback_view
+from ...digid_eherkenning_oidc.oidc_plugins.types import (
+    ClaimProcessingInstructions,
+)
+from ...digid_eherkenning_oidc.oidc_plugins.utils import (
+    get_of_auth_plugin,
+    process_claims,
+)
```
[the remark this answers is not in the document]
SilviaAmAm: I think that the last 2 should remain with the dots because the path is actually `openforms.authentication.contrib.digid_eherkenning_oidc...`. I changed the first one!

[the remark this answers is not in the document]
SilviaAmAm: Discussed: use absolute

src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+@register(OIDC_YIVI_IDENTIFIER)
+class YiviPlugin(BaseOIDCPlugin, AnonymousUserOIDCPluginProtocol):
+    def get_schema(self) -> JSONObject:
+        return YIVI_SCHEMA
+
+    def verify_claims(self, claims: JSONObject) -> bool:
+        # Not used
+        return False
```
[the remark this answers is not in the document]
SilviaAmAm: I struggled a bit with this,  `digid_eherkenning_oidc.oidc_plugins.plugins.BaseDigiDeHerkenningPlugin.verify_claims`, takes the claims, tries to process them and then returns `True` or `False` depending on whether there was an error during processing. 
Afterwards, the claims are processed again to give the "processed claims". So the same operation (processing the claims) is done twice in a row.

`verify_claims` is called in mozilla-django-oidc once in the [get_or_create_user](https://github.com/mozilla/mozilla-django-oidc/blob/2c2334fdc9b2fc72a492b5f0e990b4c30de68363/mozilla_django_oidc/auth.py#L349). Maybe it would be neater to move `verify_claims` to the `AbstractUserOIDCPluginProtocol` in mozilla-django-oidc-db, so that it is not a required abstract method.

Since for Yivi  processing the claims requires more work (figuring out the form, retrieving the auth backend options, getting the AttributeGroup...) it felt like too many queries to justify doing it twice in a row.

src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
SilviaAmAm: It would indeed be nice if the ClaimProcessingInstructions could deal also with the additional attributes :thinking: 

I reorganised quite a bit to move it there.

src/openforms/authentication/contrib/digid_eherkenning_oidc/oidc_plugins/schemas.py
```diff
@@ -0,0 +1,246 @@
...
+
+EIDAS_SCHEMA = {
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Options",
+    "description": _("OIDC eIDAS Configuration options."),
+    "type": "object",
+    "required": ["identity_settings"],
+    "properties": {
+        "loa_settings": {
+            "description": _("Level of Assurance related settings."),
```
robinmolen, line 15: We should also add the titles here (like what was done in https://github.com/maykinmedia/django-digid-eherkenning/pull/97)

sergei-maertens (changes requested):
There's a lot of type checking stuff going on, which makes me happy, but I'd also like to see the types being checked in CI, so can you add as much as possible to the `pyright.pyproject.toml` config file? Especially the *new* code should be added there so that it's type checked in CI from the moment it gets added.

src/openforms/utils/tests/keycloak.py
```diff
@@ -97,3 +123,237 @@ def mock_oidc_db_config(app_label: str, model: str, **overrides):
...
+        ),
+    ):
+        yield
+
+
+@contextmanager
+def mock_get_random_string():
+    """Mock the state & nonce random value generation
+
+    Needed so that we get predictable URLs to match with VCR.
```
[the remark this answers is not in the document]
SilviaAmAm: Yes, but then the cassettes don't seem to match anymore if the state query param is changed.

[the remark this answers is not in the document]
sergei-maertens: https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-information-from-http-querystring I think we can filter out this parameter from the URLs?

SilviaAmAm:
Mentioned by @robinmolen: https://github.com/open-formulieren/open-forms/blob/7f1349d4dacdf8780d0484ae18cb3970b1124bd4/src/openforms/authentication/contrib/digid_eherkenning_oidc/tests/oidc_plugins/test_plugins.py#L16

We are missing a test for the obfuscation of the claims of eIDAS.

src/openforms/utils/tests/keycloak.py
```diff
@@ -97,3 +123,237 @@ def mock_oidc_db_config(app_label: str, model: str, **overrides):
...
+    with (
+        patch(
+            "mozilla_django_oidc_db.models.OIDCClient.objects.get", return_value=client
+        ),
+        patch(
+            "mozilla_django_oidc_db.models.OIDCClient.objects.resolve",
+            return_value=client,
+        ),
+    ):
+        yield
```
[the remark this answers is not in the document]
SilviaAmAm: As discussed: this is now changed so that there is a factory to create the provider/client and then test plugins are registered within the test, for example:

```python
@mock_auth_and_oidc_registers()
def test_record_auth_context(self):
        oidc_client = OFOIDCClientFactory.create(
            with_keycloak_provider=True, with_digid=True
        )
        oidc_register(oidc_client.identifier)(OIDCDigidPlugin)

        @auth_register("digid_oidc")
        class OFTestAuthPlugin(DigiDOIDCAuthentication):
            oidc_plugin_identifier = oidc_client.identifier
```

I decided to go with mocking the registers as follows:

```python
@contextmanager
def mock_auth_and_oidc_registers():
    with (
        patch.object(oidc_register, "_registry", new={}),
        patch.object(auth_register, "_registry", new={}),
    ):
        yield
```
instead of mocking the whole register, because otherwise every place in the code where the register is imported needs to be patched, and i feel that this would be difficult to manage? Since there are many more places where these registers are imported than for https://github.com/open-formulieren/open-forms/blob/042dae84ed3a3b2f7ca34d11f47ae044d69cf899/src/openforms/appointments/tests/test_admin.py#L38.

sergei-maertens (approved):

src/openforms/authentication/contrib/org_oidc/oidc_plugins/plugins.py
```diff
@@ -23,9 +23,6 @@ def get_schema(self) -> JSONObject:
     def handle_callback(self, request: HttpRequest) -> HttpResponse:
         return callback_view(request)  # pyright: ignore[reportReturnType] # .as_view() returns HttpResponseBase
 
-    def _get_legacy_callback(self) -> str:
```
SilviaAmAm: Since this class implements this protocol: https://github.com/open-formulieren/open-forms/blob/8dc49136586929e646349b267f8125dea75d72d0/src/openforms/contrib/auth_oidc/plugin.py#L140
Shouldn't this method be implemented?

src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
```diff
@@ -388,8 +387,10 @@ def _get_signicat_yivi_condiscon_scope(self, options: YiviOptions) -> str:
         base64_bytes = base64.b64encode(condiscon_string.encode("ascii"))
         return f"signicat:param:condiscon_base64:{base64_bytes.decode('ascii')}"
 
-    # TODO: parent has django.http.HttpRequest, but it should be rest_framework.request.Request (maykinmedia/mozilla-django-oidc-db/issues/151)
-    def get_extra_params(self, request: Request, extra_params: dict) -> dict:  # pyright: ignore[reportIncompatibleMethodOverride]
+    def get_extra_params(
+        self, request: HttpRequest, extra_params: dict
+    ) -> dict[str, str | bytes]:
+        assert isinstance(request, Request)  # stronger guarantee
```
SilviaAmAm, line 393: Ahh I see, i was struggling because HttpRequest doesn't have the query_params. But so this is the way to do it. :nerd_face:

src/openforms/contrib/auth_oidc/admin.py
```diff
@@ -0,0 +1,62 @@
...
+                    "for one or more forms, please remove this backend from these "
+                    "forms before disabling this authentication backend."
+                ).format(plugin_name=matching_plugin.verbose_name)
+            )
+        return enabled
+
+
+@admin.register(OIDCClient)
+class OFOIDCClientAdmin(OIDCClientAdmin):
+    form = OIDCClientForm
```
SilviaAmAm, line 62: Doesn't this mean that we loose the fieldsets?

SilviaAmAm (commented):
Some questions. Thank you for cleaning it up further :slightly_smiling_face:

src/openforms/contrib/auth_oidc/admin.py
sergei-maertens: No, those are defined on the model admin. Django forms don't have a concept of fieldsets.

src/openforms/authentication/contrib/org_oidc/oidc_plugins/plugins.py
sergei-maertens: IMO this whole protocol inheritance chain is messy. It's hard to follow what comes from where, it breaks type checking w/r to the `__init__` method, the order of base classes goes from generic -> specific instead of the other way around and I think it needs to be addressed in the library first.

I also have the impression that you think protocols act like abstract base classes, but that's not the case. They essentially only exist for static code analysis (from a practical point of view).

A child class will not complain about a method not being implemented/defined, because the protocol already defines it. A protocol basically says to the type checker "trust me, this thing exists". If it doesn't and you don't call it, nothing will fall over.

SilviaAmAm: I think that what i am confused about is how to make it clear that a class implements a particular protocol, so that then you know which methods you can expect in a class :thinking: So I went with the inheritance thing, but are there better ways to achieve this?

I know it won't complain if you don't call the method, but I thought that since it is "expected" because you are implementing a protocol, then if in the future things changes and it does get called then there is a problem. In this case it doesn't really matter since this method is deprecated, but more for my understanding :sweat_smile:

src/openforms/contrib/auth_oidc/admin.py
SilviaAmAm: Strange, when I had tried this approach I had lost all fieldsets. I wonder what I did wrong. Oh well :shrug:

PR #5354, opened 2025-05-28
JSON schema generation polishing

### LINKED ISSUE #5312: Polishing of JSON schema generation
#5305 implements the core behaviour of generating a JSON schema from a form, but still requires some more _polishing_.

### Additional requirements

Generating a schema can currently be done using buttons in the form list and form detail page, which link to a simple API endpoint that returns a schema of all variables (component and user-defined) in a form. From this [comment](https://github.com/open-formulieren/open-forms/issues/4923#issuecomment-2789905021), I understand that it would be desired to be able to specify variables for which the schema should be generated. 

Furthermore, there are differences in the data format between submitted data of a form and data sent through registration backends. For example, an attachment will be present in the submission data as an array of objects with information about name, size, storage, etc. (see `File.as_json_schema`). However, in the Objects API and Generic JSON registration backends, attachments will be present as a list of urls and base64-encoded content, respectively. Also, the selectboxes data can be transformed from an object to a list here.

### Suggestion

Taking these two things into account, the suggestion is to create a (simple) view where a schema can be generated. ~Here, specific variables can be selected in a similar way that already happens in the Generic JSON registration options.~ Also, the data that the schema should represent can be selected here: 

* Objects API registration data
* Generic JSON registration data

The resulting schema can be shown in a JSON field, which makes it available for copying

Tasks

 - [x] Process the schema to represent the data that is output by Objects API and Generic JSON registration
 - [x] View/modal where ~options can be set and~ generated schema is shown
 - [ ] Add documentation

viktorvanwijk:
Closes #5312

**Still TODO**
- [ ] Add documentation

**Changes**

- Added support for plugin selection
- Added button in admin to generate schema
- Refactored schema processing of Generic JSON backend

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


docs/manual/forms/json_schema.rst
viktorvanwijk, line 1: Will translate this once reviewed. Seems like all files in the `manual` folder are Dutch by default

robinmolen (changes requested):

src/openforms/forms/tests/test_json_schema.py
```diff
@@ -676,3 +684,556 @@ def test_schema_assigned_manually(self):
...
+                ]
+            }
+        )
+        FormStepFactory.create(form=form, form_definition=form_def)
+
+        var = form.formvariable_set.get(key="file")
+        component = form_def.configuration_wrapper["file"]
+        schema = var.as_json_schema()
+
+        process_variable_schema_objects_api(component, schema, self.options)
```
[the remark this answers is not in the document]
viktorvanwijk: I wrote some targeted tests for these because I was thinking about moving the code to their respective registrations. I agree that using `generate_variable_schema` tests a more complete flow, will change it :)

I do test the _entire_ flow in `FormJsonSchemaAPITests`, so adding them here as well seems a bit redundant to me

src/openforms/forms/tests/test_api_form_json_schema.py
```diff
@@ -23,52 +27,130 @@ def setUp(self):
...
+        response = self.client.get(url, options)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        schema = response.json()
+
+        # Note: convert the required list to a set to counteract test flakiness. The
+        # order of this list can change because we iterate over the variables from the
+        # database when generating the schema.
+        schema["required"] = set(schema["required"])
+        expected_schema["required"] = set(expected_schema["required"])
```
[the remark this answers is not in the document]
viktorvanwijk: No, because these are not fetched from the database. This is a specific problem for form variables because we iterate over them in `openforms.forms.json_schema._iter_form_variables`

src/openforms/forms/tests/test_api_form_json_schema.py
```diff
@@ -111,20 +216,52 @@ def test_happy_flow(self):
...
```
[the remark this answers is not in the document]
viktorvanwijk: Same answer as above :)

src/openforms/registrations/contrib/generic_json/tests/test_backend.py
```diff
@@ -481,43 +481,6 @@ def test_path_traversal_attack(self):
                 options["path"] = path
                 json_plugin.register_submission(submission, options)
 
-    def test_select_boxes_schema_required_is_empty_when_no_data_is_submitted(self):
```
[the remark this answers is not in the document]
viktorvanwijk: I removed it because this was a very specific case that *shouldn't* occur anymore with the fix in 5eebff6e.

A hidden selectboxes component with another form variable or reference lists as a data source would be present in the submission data as an empty dict. This was because of a difference in the empty value of this component between the frontend and the backend. An unfilled selectboxes component should have all of its options in a dict set to false, e.g. `{"option_1": False, "option_2": False, "option_3": False}`. In this specific case, `get_component_empty_value` would return an empty dict instead, because `component["defaultValue"]` value is an empty dict, which happens when we have another form variable or reference lists as a data source.

src/openforms/js/components/admin/forms/ModalJsonSchemaGeneration.js
```diff
@@ -0,0 +1,70 @@
...
+  const [schema, setSchema] = useState(null);
+  const [theme] = useGlobalState(currentTheme);
+  const {form} = useContext(FormContext);
+
+  async function onClick() {
+    setModalOpen(true);
+    const params = new URLSearchParams({registration_backend_key: backendKey});
+    const response = await get(`${FORM_ENDPOINT}/${form.uuid}/json_schema?${params}`);
+    if (!response.ok) {
+      setErrorState(true);
```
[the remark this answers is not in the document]
viktorvanwijk: Good point :)

src/openforms/js/components/admin/forms/ModalJsonSchemaGeneration.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+    setSchema(response.data);
+  }
+
+  return (
+    <>
+      <a onClick={onClick} className="button-link" style={{paddingBlock: '10px'}}>
```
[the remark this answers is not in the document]
viktorvanwijk: The idea was to have a link instead of a button, to ensure it's not such a prominent feature of the registration backend. The options already have a big button, and we don't have a secondary button as far as I understood. So I took a link and style it to work like a button. Perhaps this is the wrong way around, though, and would it be better to use a `button` that is styled like a link?

src/openforms/forms/json_schema.py
```diff
@@ -35,12 +51,14 @@ def _iter_form_variables(
...
     # Handle form variables holding dynamic data (component and user defined)
-    yield from form.formvariable_set.all()
+    yield from form.formvariable_set.all()  # pyright: ignore[reportAttributeAccessIssue]
 
 
 def generate_json_schema(
     form: Form,
     limit_to_variables: Sequence[str],
+    backend_id: str = "",
+    backend_options: BackendOptions = None,
```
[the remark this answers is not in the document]
viktorvanwijk: Yeah I had that at first as well, but the routine is used in the generic json plugin too, where we only have the id and options. As I'm writing this, though, I realize we could get the `FormRegistrationBackend` from the form through the submission there. OR, get the backend inside `generate_json_schema`, and pass only the id to it. I guess this last option is the safest, but not sure what's "better" tbh.

To be fair, if someone uses `generate_json_schema` and passes a generic json plugin id with objects api options, they deserve to be spoken to sternly :upside_down_face:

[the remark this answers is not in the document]
robinmolen: Fair point..

src/openforms/forms/tests/test_api_form_json_schema.py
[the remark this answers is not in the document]
robinmolen: Ah okay, then everything should be fine 👍

docs/manual/forms/json_schema.rst
```diff
@@ -0,0 +1,99 @@
+.. _manual_json_schema:
+
+JSON-schema genereren
+=====================
+Het is mogelijk om een `JSON-schema`_ van een formulier te genereren. Momenteel kan dit voor de Objecten API- of de
```
robinmolen, line 5: 'JSON-schema'_ causes build issues due to typo:
```suggestion
Het is mogelijk om een `JSON schema`_ van een formulier te genereren. Momenteel kan dit voor de Objecten API- of de
```

robinmolen (approved):
Small typo. Besides that, looks good to me!

docs/manual/forms/json_schema.rst
```diff
@@ -0,0 +1,99 @@
...
+configuratieopties (indien gespecificeerd) opgenomen in het schema:
+
+ - Label
+ - Beschrijving
+ - Validatieregels, bijvoorbeeld maximale lengte of een regex-patroon
+
+Hoe werkt het
+-------------
+Om een JSON-schema van een formulier te genereren, ga naar het tabblad **Registratie** en configureer een
+:doc:`Objecten API- </configuration/registration/objects>` of
```
sergei-maertens, line 18: please use `:ref:` instead where you point to the label: https://github.com/open-formulieren/open-forms/blob/34c8023497c0bc49dab35a5e6729d4a6aff60e11/docs/configuration/registration/objects.rst?plain=1#L1

```suggestion
:ref:`Objecten API- <configuration_registration_objects>` of
```

docs/manual/forms/json_schema.rst
```diff
@@ -0,0 +1,99 @@
...
+:doc:`Generieke JSON-registratie </configuration/registration/generic-json>`. De fieldset van de plugin bevat een knop
+waarmee een venster kan worden geopend. Hierin is het gegenereerde schema beschikbaar is in een JSON-editor.
```
sergei-maertens, line 20:
```suggestion
waarmee een venster kan worden geopend. Hierin is het gegenereerde schema beschikbaar in een JSON-editor.
```

pyright.pyproject.toml
```diff
@@ -61,6 +61,8 @@ include = [
     "src/openforms/template",
     # generic typing helpers
     "src/openforms/typing.py",
+    # JSON schema generation
+    "src/openforms/forms/json_schema.py"
```
sergei-maertens, line 65: :tada: :label:

src/openforms/forms/api/viewsets.py
```diff
@@ -613,18 +625,42 @@ def logic_rules_list(self, request, *args, **kwargs):
         summary=_("JSON schema"),
         description=_("Generate the JSON schema for a form."),
         parameters=[UUID_OR_SLUG_PARAMETER],
-        request=None,
-        responses={status.HTTP_200_OK: OpenApiTypes.OBJECT},
+        request=FormJsonSchemaOptionsSerializer,
+        responses={
+            status.HTTP_200_OK: OpenApiTypes.OBJECT,
+            status.HTTP_400_BAD_REQUEST: ExceptionSerializer,
```
sergei-maertens, line 631: shouldn't this be the validation errors serializer?

src/openforms/forms/api/viewsets.py
```diff
@@ -389,7 +401,7 @@ def retrieve(self, request, *args, **kwargs):
         summary=_("Export form"),
         tags=["forms"],
         parameters=[UUID_OR_SLUG_PARAMETER],
-        request=None,
+        request=FormJsonSchemaOptionsSerializer,
```
sergei-maertens, line 404: the export endpoint doesn't take a request body, does it?

src/openforms/forms/api/viewsets.py
```diff
@@ -613,18 +625,42 @@ def logic_rules_list(self, request, *args, **kwargs):
         summary=_("JSON schema"),
         description=_("Generate the JSON schema for a form."),
         parameters=[UUID_OR_SLUG_PARAMETER],
-        request=None,
-        responses={status.HTTP_200_OK: OpenApiTypes.OBJECT},
+        request=FormJsonSchemaOptionsSerializer,
```
sergei-maertens, line 628: this endpoint is called via a `GET` operation, which typically doesn't have a request body. Instead, you want to document the GET params via the `parameters` kwarg, see here for example: https://github.com/open-formulieren/open-forms/blob/34c8023497c0bc49dab35a5e6729d4a6aff60e11/src/openforms/prefill/api/views.py#L22

This serializer approach for filter parameters is not great, but it's the best we have for the time being.

src/openforms/forms/api/viewsets.py
```diff
@@ -613,18 +625,42 @@ def logic_rules_list(self, request, *args, **kwargs):
...
+    )
+    @action(
+        detail=True,
+        methods=["get"],
+        permission_classes=(permissions.IsAdminUser,),
+        renderer_classes=(FormJSONSchemaRenderer,),
     )
-    @action(detail=True, methods=["get"], permission_classes=(permissions.IsAdminUser,))
     def json_schema(self, request, *args, **kwargs):
+        serializer = FormJsonSchemaOptionsSerializer(data=request.GET)
```
sergei-maertens, line 641: in API endoints, you can use `request.query_params` with DRF, which works for both GET and POST requests :)

```suggestion
        serializer = FormJsonSchemaOptionsSerializer(data=request.query_params)
```

src/openforms/forms/api/viewsets.py
```diff
@@ -613,18 +625,42 @@ def logic_rules_list(self, request, *args, **kwargs):
...
         form = self.get_object()
 
-        schema = generate_json_schema(
-            form,
-            limit_to_variables=form.formvariable_set.values_list("key", flat=True),
+        # Note that a backend is unique by its key and corresponding form
+        backend = get_object_or_404(
+            FormRegistrationBackend,
+            key=serializer.validated_data["registration_backend_key"],
+            form=form.id,
```
sergei-maertens, line 650:
```suggestion
            form=form,
```
should also work and is a bit more idiomatic - you could also pass `form.formregistrationbackend_set` as first argument instead

src/openforms/forms/api/viewsets.py
```diff
@@ -613,18 +625,42 @@ def logic_rules_list(self, request, *args, **kwargs):
...
+        try:
+            schema = generate_json_schema(
+                form=form,
+                limit_to_variables=form.formvariable_set.values_list("key", flat=True),
+                backend_id=backend.backend,
+                backend_options=backend.options,
+            )
+        except InvalidBackendIdError as exc:
+            logger.error("invalid_backend_id", exc_info=exc)
+            return HttpResponseBadRequest("Invalid backend id")
```
sergei-maertens, line 662: is this something that can be validated in the `FormJsonSchemaOptionsSerializer`?

You could look up the form before instantiating the serializer and then pass it in the context:

```py
FormJsonSchemaOptionsSerializer(data=..., context={"form": form})
```

src/openforms/forms/json_schema.py
```diff
@@ -179,18 +269,180 @@ def __setitem__(self, key: str, value: JSONValue):
...
         """
+        assert isinstance(key, str)
         try:
             self[key]
         except KeyError:
             return False
         return True
+
+
+def process_variable_schema_objects_api(
```
sergei-maertens, line 292: I would expect this function to live in `openforms.registrations.contrib.objects_api` instead of in the core module, and for it to be looked up/resolved via the registration plugin registry similar to how `plugin.register_submission` is resolved.

src/openforms/forms/json_schema.py
```diff
@@ -179,18 +269,180 @@ def __setitem__(self, key: str, value: JSONValue):
...
+                )
+        case _:
+            pass
+
+
+def process_variable_schema_generic_json(
+    component: Component,
+    schema: JSONObject,
+    backend_options: GenericJSONOptions,
+):
```
sergei-maertens, line 372: same remark here, but for another plugin of course

src/openforms/js/components/admin/forms/ModalJsonSchemaGeneration.js
```diff
@@ -0,0 +1,76 @@
...
+  const [modalOpen, setModalOpen] = useState(false);
+  const [hasError, setErrorState] = useState(false);
+  const [schema, setSchema] = useState(null);
+  const [theme] = useGlobalState(currentTheme);
+  const {form} = useContext(FormContext);
+
+  async function onShowModal() {
+    setModalOpen(true);
+    const params = new URLSearchParams({registration_backend_key: backendKey});
+    const response = await get(`${FORM_ENDPOINT}/${form.uuid}/json_schema?${params}`);
```
sergei-maertens, line 27: https://github.com/open-formulieren/open-forms/blob/34c8023497c0bc49dab35a5e6729d4a6aff60e11/src/openforms/js/utils/fetch.js#L58 params support is baked in with an API similar to python:

```suggestion
    const response = await get(`${FORM_ENDPOINT}/${form.uuid}/json_schema`, {registration_backend_key: backendKey});
```

src/openforms/js/components/admin/forms/ModalJsonSchemaGeneration.js
```diff
@@ -0,0 +1,76 @@
...
+import {FORM_ENDPOINT} from 'components/admin/form_design/constants';
+import Modal from 'components/admin/modals/Modal';
+import ErrorMessage from 'components/errors/ErrorMessage';
+import {get} from 'utils/fetch';
+import {currentTheme} from 'utils/theme';
+
+/**
+ * A container for JSON schema generation of a form.
+ */
+const ModalJsonSchemaGeneration = ({modalTitle, backendKey, modalSize = 'large'}) => {
```
sergei-maertens, line 17: is there are reason you support multiple modal sizes?

src/openforms/forms/api/viewsets.py
viktorvanwijk: I guess it should, yeah. I wasn't aware of its existence

src/openforms/forms/api/viewsets.py
```diff
@@ -612,17 +622,36 @@ def logic_rules_list(self, request, *args, **kwargs):
     @extend_schema(
         summary=_("JSON schema"),
         description=_("Generate the JSON schema for a form."),
-        parameters=[UUID_OR_SLUG_PARAMETER],
-        request=None,
-        responses={status.HTTP_200_OK: OpenApiTypes.OBJECT},
+        request=FormJsonSchemaOptionsSerializer,
```
sergei-maertens, line 625:
```suggestion
        request=None,
```

src/openforms/forms/api/serializers/form.py
```diff
@@ -613,3 +619,59 @@ class FormImportSerializer(serializers.Serializer):
...
+        except FormRegistrationBackend.DoesNotExist as exc:
+            raise ValidationError(
+                f"Backend with key '{key}' does not exist for form '{form}'"
+            ) from exc
+
+        if not plugin_allows_json_schema_generation(backend.backend, backend.options):
+            raise ValidationError(
+                f"Backend with id '{backend.backend}' does not allow JSON schema "
+                "generation"
+            )
```
sergei-maertens, line 660: please use translatable strings for user-facing validation error messages :)

```py
_("Backend with key '{key}' ...").format(key=...)
```
additionally, I think you can associate both errors with with the `registration_backend_key` field, so perhaps this validation should be implemented in `def validate_registration_backend_key` instead, so that the error message ends up under the right key instead of `nonFieldErrors`.

src/openforms/registrations/contrib/generic_json/plugin.py
```diff
@@ -311,42 +387,18 @@ def get_attachments_and_base_schema(
...
     :return encoded_attachments: List of encoded attachments for this file component.
-    :return base_schema: JSON schema describing the entries of ``encoded_attachments``.
     """
     key = f"{key_prefix}.{component['key']}" if key_prefix else component["key"]
 
-    encoded_attachments: list[JSONObject] = [
+    def encode_attachment(attachment: SubmissionFileAttachment) -> str:
+        with attachment.content.open("rb") as f:
+            f.seek(0)
+            return base64.b64encode(f.read()).decode()
```
sergei-maertens, line 396: this can be moved outside of the function right? I don't see what would make it a closure.

sergei-maertens (changes requested):
Couple more final tweaks, looks nice though!

src/openforms/registrations/contrib/generic_json/plugin.py
viktorvanwijk: Yeah, not sure why I did that

src/openforms/forms/api/serializers/form.py
viktorvanwijk: > please use translatable strings for user-facing validation error messages

Probably good to actually show these messages then :upside_down_face: 

> so perhaps this validation should be implemented in `def validate_registration_backend_key` instead

I did not know you could do this

sergei-maertens (approved):

Issue #5132, opened 2025-03-03
As municipality I want to have Yivi / Wallet support


joeribekker:
### Thema / Theme

Frontend

### Omschrijving / Description

In order to do this, we need to make OIDC more generic.

### Added value / Toegevoegde waarde

Add Yivi-support. This encompases scopes, claims configurable at the form level.

WietekeNijmegen:
For testing purposes of one of the options. This is a demo of Yivi conditional with test attributes.
It shows how to load them and how they are used.

WietekeNijmegen:
https://github.com/GemeenteNijmegen/open-forms/issues/186
Here you can find a demo showing the use of yivi conditional.
- How to load test attributes
- How a user sees the conditional attributes in the Yivi app as a choice

WietekeNijmegen:
@robinmolen heb je komende tijd nog behoefte aan testen/afstemmen over bepaalde zaken in de feature?

robinmolen:
At this time we unfortunately cannot get access to a Signicat Yivi test environment. Signicat puts all Yivi support requests on hold, due to discussions and uncertainties at Yivi itself. To test and evaluate the correctness of our implementation, we will need a Yivi test environment.

Besides an unavailable test environment, provides Signicat also very limited technical Yivi information and documentation.

For now, we can continue development on some remaining points surrounding this plugin, but we cannot test or confidently support Yivi at this time.

PR #5363, opened 2025-06-04
Supporting authentication with Yivi

robinmolen:
Closes partly #5132

**Changes**

Adding support for Yivi authentication via Signicat OIDC.

Allowing end-users to authenticate with great flexibility; providing options for natural person authentication (BSN), company authentication (KVK) and anonymous user (pseudo). Per form the admin can specify which authentication options are available for the end-user. The end-user decides which authentication option they use. If any authentication option is specified by the admin, that the user MUST choose one (if non are specified, the authentication automatically uses pseudonyms for identification)

In addition to form specific authentication options, the Yivi plugin also support form specific scopes/claims. Meaning that one form could request the users e-mail address, while another form could request the users residential address. These form specific scopes are optional, meaning that the users can decide whether they provide that information or not.

**Checklist**

- [ ] Update docs

- Impact on features

  - [ ] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/authentication/contrib/generic_oidc/plugin.py
```diff
@@ -0,0 +1,17 @@
+from ...registry import register
+from ..digid_eherkenning_oidc.plugin import (
+    get_config_to_plugin as digid_eherkenning_config_to_plugin,
+)
+from ..yivi_oidc.plugin import YiviOIDCAuthentication
+
+
+def get_config_to_plugin() -> dict:
```
viktorvanwijk, line 8: Perhaps this can be combined with the original, given that it's just a mapping from a class to identifier?

viktorvanwijk (commented):
Some thoughts and questions about the backend :). I tried to have a look at the structure as well, but I guess Sergei will have a better overview for that, as this is the first time I'm looking at the code for authentication

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+20. Laat de overige endpoints leeg - deze worden automatisch aangevuld.
+
+Vervolgens kan je de **Attribuut groepen** instellen. Deze zijn niet strikt nodig om de
+Yivi plugin te kunnen gebruiken, deze bieden wel meer mogelijkheden voor gebruikers en
+formulierbouwers.
+
+21. Indien gewenst, kan je instellen welke attributen op formulier niveau geconfigureerd
+    kunnen worden, bijvoorbeeld:
+
+    * Type in het **Groep naam** tekstveld de naam van de groep in. Deze waarde wordt
```
viktorvanwijk, line 243:
```suggestion
    * Typ in het **Groepsnaam**-tekstveld de naam van de groep in. Deze waarde wordt
```

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+      enkel gebruikt om de attributen verzameling herkenbaar te maken voor
```
viktorvanwijk, line 244:
```suggestion
      enkel gebruikt om de attributenverzameling herkenbaar te maken voor
```

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+      formulierbouwers.
+    * In **Groep beschrijving** kan je de groep een begrijpelijke omschrijving geven,
```
viktorvanwijk, line 246:
```suggestion
    * In **Groepsbeschrijving** kan je de groep een begrijpelijke omschrijving geven,
```

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+      hiermee kunnen formulierbouwers gemakkelijk herkennen waar de groep voor dient en
+      welke attributen ermee gebruikt worden.
+    * Type in het **Attributen** tekstveld het attribuut die voor deze groep gebruikt zal
```
viktorvanwijk, line 249:
```suggestion
    * Typ in het **Attributen**-tekstveld het attribuut dat voor deze groep gebruikt zal
```

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+      worden. Om meerdere attributen in dezelfde groep te plaatsen, klik op "Add item"
+      en herhaal totdat je alle gewenste attributen hebt gedefinieerd.
+    * Voor een volgende attributen groep, klik op "Add another Attribute group" en
```
viktorvanwijk, line 252:
```suggestion
    * Voor een volgende attributengroep, klik op "Add another Attribute group" en
```
Should "Add another Atrribute group" also be translated to Dutch?

src/openforms/js/components/admin/form_design/authentication/yivi/AdditionalAttributesGroupsField.js
```diff
@@ -0,0 +1,83 @@
...
+import ReactSelect from 'components/admin/forms/ReactSelect';
+import {getReactSelectOptionsFromSchema} from 'utils/json-schema';
+
+const AdditionalGroupSelectorOption = props => {
+  const {value, label} = props.data;
+  return (
+    <components.Option {...props}>
+      <span className="form-additional-attributes-groups-dropdown-option">
+        <b className="form-additional-attributes-groups-dropdown-option__key">{value}</b>
+        <span className="form-additional-attributes-groups-dropdown-option__label">{label}</span>
```
viktorvanwijk, line 17: Do we really need the description of the group here? If the descriptions are long it might get quite convoluted. Perhaps a smaller font could also work?

src/openforms/js/components/admin/form_design/authentication/yivi/YiviOptionsFormBsnFields.js
```diff
@@ -0,0 +1,61 @@
...
+
+const YiviOptionsFormBsnFields = ({plugin}) => {
+  const {
+    values: {authenticationOptions = []},
+  } = useFormikContext();
+
+  // Only show the bsn options fields if bsn is selected for authentication
+  if (!authenticationOptions.includes('bsn')) {
+    return null;
+  }
```
viktorvanwijk, line 19: Out of curiosity, I usually do this check around the whole component, so in `YiviOptionsFormFields` in this case. Is there a good practice for this?

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -346,7 +345,7 @@ function reducer(draft, action) {
         const plugin = draft.availableAuthPlugins.find(backend => backend.id === pluginId);
         draft.form.authBackends.push({
           backend: pluginId,
-          options: getInitialPluginOptions(plugin),
+          options: {},
```
viktorvanwijk, line 348: For my understanding, why is this not necessary anymore?

src/openforms/js/components/admin/forms/ReactSelect.js
```diff
@@ -68,6 +68,9 @@ const getValue = (options, value) => {
   return null;
 };
 
+const getValues = (options, values) =>
+  !!values ? values.map(value => getValue(options, value)) : [];
```
viktorvanwijk, line 72: Is it good practice to add `!!`? `values` can just be interpreted as falsy as well, right?

For the `SelectWithoutFormik` component, we decided to always expect an array for the `isMulti` case, btw

viktorvanwijk (commented):
And finally, few questions about the frontend :)

src/openforms/authentication/contrib/generic_oidc/backends.py
```diff
@@ -0,0 +1,187 @@
...
+            ):
+                case str(), None:
+                    self._config.loa_claim = self._config.bsn_loa_claim
+                    self._config.default_loa = self._config.bsn_default_loa
+                    self._config.loa_value_mapping = self._config.bsn_loa_value_mapping
+                case None, str():
+                    self._config.loa_claim = self._config.kvk_loa_claim
+                    self._config.default_loa = self._config.kvk_default_loa
+                    self._config.loa_value_mapping = self._config.kvk_loa_value_mapping
+                case _:
```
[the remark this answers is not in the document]
robinmolen: Check, yeah makes sense

src/openforms/authentication/contrib/generic_oidc/backends.py
```diff
@@ -0,0 +1,187 @@
...
+
+        return True
+
+    def _extract_additional_claims_for_auth_backend(
+        self, auth_backend_id: str, claims: JSONObject
+    ) -> JSONObject:
+        try:
+            auth_backend = FormAuthenticationBackend.objects.get(id=auth_backend_id)
+        except FormAuthenticationBackend.DoesNotExist:
+            raise ValidationError(
```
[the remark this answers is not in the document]
robinmolen: Fair.. I've added a custom exception for this case 👍

src/openforms/authentication/contrib/yivi_oidc/admin.py
```diff
@@ -0,0 +1,112 @@
...
+from solo.admin import SingletonModelAdmin
+
+from openforms.forms.models import Form
+
+from .constants import PLUGIN_ID
+from .models import AttributeGroup, YiviOpenIDConnectConfig
+from .plugin import YiviOIDCAuthentication
+
+
+class OIDCConfigForm(OpenIDConnectConfigForm):
```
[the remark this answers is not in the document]
robinmolen: Good point, i'll move it to the generic_oidc package

src/openforms/authentication/contrib/yivi_oidc/models.py
```diff
@@ -0,0 +1,225 @@
...
+from mozilla_django_oidc_db.models import OpenIDConnectConfigBase
+
+
+def get_callback_view(self):
+    from .views import callback_view
+
+    return callback_view
+
+
+def attach_loa_choices():
```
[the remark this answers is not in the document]
robinmolen: I thought to keep it close to the existing `default_loa_choices` usage, but this can indeed be greatly simplified.
I also couldn't find why the django-digid-eherkenning project uses those decorators (maybe just for ease of use), so I don't think we are missing anything important with this more simplistic setup.

So I'll remove the `attach_loa_choices` decorator, in favor of direct assignment and a `get_loa_schema` method.

src/openforms/authentication/contrib/yivi_oidc/models.py
```diff
@@ -0,0 +1,225 @@
...
+        ),
+    )
+
+    pseudo_claim = ClaimField(
+        verbose_name=_("pseudo identifier claim"),
+        default=ClaimFieldDefault("sub"),
+        help_text=_(
+            "Name of the claim holding the (opaque) identifier of the user. "
+            "This claim will be used when the plugin is set to anonymous authentication "
+            "(when neither bsn or kvk are selected as authentication options)."
```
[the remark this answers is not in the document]
robinmolen: Thats kinda an issue right now. Yes, that should be possible (allow pseudo, bsn or kvk login on the same form via the same Yivi auth button), but there currently is an issues with Signicat/Yivi that makes this impossible.

This is something I have been discussing with Signicat for the past few weeks (with still no clear answer, issue or solution). The problem is: if we have optional attributes, and the user provides those attributes, the website doesn't receive them. So attributes have to be required, otherwise they are never returned.

Pseudo id is a piece of information we receive from Signicat, not from Yivi. So for Yivi, logging in with Pseudo id means logging in without giving authentication information. If we say: "login with BSN or Pseudo" to Yivi, we actually say: "give us BSN or don't". Like I said before, if we do this we can never receive BSN, even if the user did provide it.

TLDR; Yes a form should allow BSN, KVK or Pseudo login, but we can't due to some bug/mis-configuration/something else?

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+``attributen``
+    De attributen die aan de inloggende gebruiker gevraagd worden. Wanneer de gebruiker
+    deze gegevens aanbiedt, kunnen deze gebruikt worden bij het
+    :ref:`voorinvullen (prefill) <example_prefill>`.
+    De attribuut groepen zijn collectief optioneel, de gebruiker kan zelf bepalen welke
+    attribuut groepen wel en niet aan te leveren.
+
+.. _configuration_oidc_yivi_appgroup:
+
+Configureren van OIDC-provider
```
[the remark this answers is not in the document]
robinmolen: There is already a general OpenID Connect documentation that covers the OIDC-provider, lets use that one 👍

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+4. Laat bij **OpenID Connect scopes** de standaardwaarden staan, of stel deze in volgens
+   de instructies van je OpenID Connect provider.
+5. Vul bij **OpenID sign algorithm** ``RS256`` in.
+6. Laat **Sign key** leeg.
+
+Stel dan de claims in:
+
+Beginnend bij de claims gebruikt voor BSN authenticatie.
+
+7. Laat bij **BSN-scope** de standaardwaarde staan, tenzij de scope voor BSN
```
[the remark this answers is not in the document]
robinmolen: Yeah it was, in my previous setup. Removing it 👍 (same goes for KVK-scope)

docs/configuration/authentication/oidc_yivi.rst
```diff
@@ -0,0 +1,258 @@
...
+    * klik op "Add item"
+    * Kies "Tekstuele waarde" in de **From** dropdown en voer de waarde ``10`` op in het
+      tekstveld
+    * Selecteer "DigiD Basis" in de **To** dropdown
+    * Herhaal voor andere waarden en niveaus
+
+Vervolgens richt je de claims voor KVK authenticatie in.
+
+11. Laat bij **KVK-scope** de standaardwaarde staan, tenzij de scope voor KVK
+    authenticatie in de OpenID Connect provider anders is dan ``kvk``.
```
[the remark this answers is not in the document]
robinmolen: Is indeed no-longer relevant

src/openforms/js/components/admin/form_design/authentication/yivi/AdditionalAttributesGroupsField.js
robinmolen: Mhh yeah.. Im not entirely sure. I think it would be beneficial to give some context about the attribute group (so you don't have to check the Yivi config every time you want to add an attribute group), but it might indeed cause some distracting UI "noise".

I'll check if a smaller font size works better.

src/openforms/js/components/admin/form_design/form-creation-form.js
robinmolen: Because we now use Formik to manage the plugin options fields. Previously the options where managed by manually setting/updating/checking the data, so we needed a base to work from (just making everything a bit easier). But now, we don't have to manually check/correct the data anymore, Formik does it for us.

So this just lowers the amount of stuff that we have to manage/maintain.

src/openforms/authentication/contrib/yivi_oidc/tests/test_config.py
```diff
@@ -0,0 +1,177 @@
...
+                "additional_attributes_groups": [],
+            },
+        )
+
+        self.assertFalse(serializer.is_valid())
+
+        self.assertTrue("authentication_options" in serializer.errors)
+        self.assertEqual(
+            serializer.errors["authentication_options"][0][0].code, "invalid_choice"
+        )
```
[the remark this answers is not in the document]
robinmolen: Mhh fair... I thought, better save then sorry?

I'll squash the bsn and kvk tests together (they do almost the exact same thing), but I think it would be "good" to keep them.

sergei-maertens (changes requested):
Initial review based on outdated diff, submitting it now to make sure they're not lost!

src/openforms/authentication/contrib/generic_oidc/backends.py
```diff
@@ -0,0 +1,194 @@
...
+                    self._config.default_loa = self._config.bsn_default_loa
+                    self._config.loa_value_mapping = self._config.bsn_loa_value_mapping
+                case None, str():
+                    self._config.loa_claim = self._config.kvk_loa_claim
+                    self._config.default_loa = self._config.kvk_default_loa
+                    self._config.loa_value_mapping = self._config.kvk_loa_value_mapping
+                case None, None:
+                    self._config.loa_claim = [""]
+                    self._config.default_loa = None
+                    self._config.loa_value_mapping = None
```
sergei-maertens, line 105: I don't see definitions for the `loa_claim`, `default_loa` and/or `loa_value_mapping` properties/attributes on the `YiviOpenIDConnectConfig` model. My type checker is not happy about this :grimacing: 

image

src/openforms/authentication/contrib/generic_oidc/backends.py
```diff
@@ -0,0 +1,194 @@
...
+        # see if we can use a cached config instance from the settings configuration
+        assert hasattr(self, "_config") and (
+            isinstance(self._config, BaseConfig)
+            or isinstance(self._config, YiviOpenIDConnectConfig)
+        )
+
+        # Because Yivi has loa config for BSN and KVK, we need to set the "global" loa
+        # config dynamically based on which data is presented.
+        # This will be moved after oidc-db changes.
+        if issubclass(self.config_class, YiviOpenIDConnectConfig):
```
sergei-maertens, line 85: Is there a way we can push this code to the plugin? This is a big OOP violation - suddenly your base class/module/package needs to know about more concrete subclasses/derivates.

src/openforms/authentication/contrib/digid_eherkenning_oidc/views.py
```diff
@@ -25,103 +13,23 @@
...
-eherkenning_init = OIDCInit.as_view(
+eherkenning_init = OIDCAuthenticationInitView[OptionsT].as_view(
     config_class=OFEHerkenningConfig,
     allow_next_from_query=False,
 )
-eherkenning_bewindvoering_init = OIDCInit.as_view(
+eherkenning_bewindvoering_init = OIDCAuthenticationInitView[OptionsT].as_view(
     config_class=OFEHerkenningBewindvoeringConfig,
     allow_next_from_query=False,
 )
```
sergei-maertens, line 35: passsing a generic typevar in a concrete instantiation doesn't make a lot of sense - you would typically pass the actual plugin-specific option shapes here. But either way there's little point in passing the type var because we don't do anything with the init views ourselves anymore that may consume the `options`, it's all down to django at this point to call the view.

src/openforms/authentication/contrib/generic_oidc/plugin.py
```diff
@@ -0,0 +1,125 @@
...
+    return {
+        plugin.config_class: plugin
+        for plugin in register
+        if isinstance(plugin, OIDCAuthentication)
+    }
+
+
+class AuthInit(Protocol):
+    def __call__(
+        self, request: HttpRequest, return_url: str, *args, **kwargs
```
sergei-maertens, line 41: let's make this `options` kwarg explicit! Took me a long time to figure out where the `options` were passed along :sweat_smile:

src/openforms/authentication/contrib/yivi_oidc/plugin.py
```diff
@@ -0,0 +1,139 @@
...
+
+@register(PLUGIN_ID)
+class YiviOIDCAuthentication(OIDCAuthentication[YiviClaims, YiviOptions]):
+    verbose_name = _("Yivi via OpenID Connect")
+    # Yivi can provide a range of auth attributes, based on form config and user choice
+    provides_auth = (AuthAttribute.bsn, AuthAttribute.kvk, AuthAttribute.pseudo)
+    session_key = "yivi_oidc"
+    config_class = YiviOpenIDConnectConfig
+    configuration_options = YiviOptionsSerializer
+    init_view = staticmethod(yivi_init)
```
sergei-maertens, line 71: I think you can get rid of the whole `options` via `def get(...)` kwarg if you make this a computed property:

```py
@property
def init_view(self):
    def view(self, request, options, *args, **kwargs):
        yivi_init = OIDCAuthenticationInitView.as_view(
            config_class=YiviOpenIDConnectConfig,
            allow_next_from_query=False,
            options=options,
        )
        return yivi_init(request, *args, **kwargs)
    return view
```

This avoids the risk that you can accidentally set the options via URL parameters, as the URL params end up in the `*args` and `**kwargs` of view functions.

src/openforms/authentication/contrib/generic_oidc/views.py
```diff
@@ -0,0 +1,139 @@
...
+        request: HttpRequest,
+        return_url: str = "",
+        options: OptionsT = None,
+        *args,
+        **kwargs,
+    ) -> HttpResponseRedirect:
+        if options is None:
+            options = {}
+        self.options = options
+        return super().get(request, return_url, *args, **kwargs)
```
sergei-maertens, line 45: See my comment at https://github.com/open-formulieren/open-forms/pull/5363/files#diff-722534ade5c574db4a55aa7b240672888840d3d20ab27425816fd886df38d01fR71 for how you can avoid having to do this override entirely. Check the base class `OIDCInit` implementation to figure out what's needed to make this work via `.as_view(...)` and the django docs of course: https://docs.djangoproject.com/en/5.2/topics/class-based-views/#usage-in-your-urlconf

src/openforms/authentication/contrib/org_oidc/plugin.py
```diff
@@ -31,7 +31,9 @@ class OIDCAuthentication(BasePlugin):
     verbose_name = _("Organization via OpenID Connect")
     provides_auth = AuthAttribute.employee_id
 
-    def start_login(self, request: HttpRequest, form: Form, form_url: str):
+    def start_login(
+        self, request: HttpRequest, form: Form, form_url: str, options: dict
```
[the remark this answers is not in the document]
robinmolen: Ah okay, check. So its best to change it to:

```
self, request: HttpRequest, form: Form, form_url: str, options
```

instead of this?:

```
self, request: HttpRequest, form: Form, form_url: str, options: OptionsT
```

docs/developers/plugins/authentication_plugins.rst
```diff
@@ -0,0 +1,157 @@
...
+.. warning:: This notation for the optional disjunction **doesn't** work:
+
+   .. code:: json
+
+      [
+        [
+          [],
+          ["pbdf.pbdf.diploma.degree"]
+        ]
+      ]
```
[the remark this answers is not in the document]
robinmolen: Good point.. No not yet, will add it

src/openforms/authentication/contrib/yivi_oidc/config.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class YiviOptions(TypedDict):
+    """
+    Shape of the DigiD authentication plugin options.
+    This describes the shape of :attr:`YiviOptionsSerializer.validated_data`, after
+    the input data has been cleaned/validated.
+    """
+
+    authentication_options: List[YiviAuthenticationAttributes]
```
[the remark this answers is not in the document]
robinmolen: Check, will replace it. I just saw that `typing.Type` and some others are also deprecated, will replace those as well (https://medium.com/@anuraagkhare_39064/typing-module-deprecated-in-python-what-you-need-to-know-d4348b694141).

src/openforms/authentication/contrib/generic_oidc/__init__.py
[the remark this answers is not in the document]
sergei-maertens: Yeah, perhaps `openforms.contrib.auth_oidc` is a good place for it - we have a similar pattern there for `openforms.contrib.objects_api` where code that is shared by the registration and prefill plugins lives. So it's still clearly marked as "plugin-related" because it's inside a `contrib`, but we keep the `openforms.authentication.contrib.*` packages pure so that you know every package there is an actual plugin provider.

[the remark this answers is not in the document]
robinmolen: This also highlights another issue.. Yivi plugin specific code in the generic setup.. Will change that aswell

src/openforms/authentication/contrib/yivi_oidc/admin.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    A custom Yivi implementation for applying the shared fieldsets configuration with the
+    model-specific overrides.
+    This Yivi specific implementation replaces the ATTRIBUTES_MAPPING_TITLE with separate
+    BSN, KVK and pseudo mappings. Adding more overview and structure to the Yivi
+    configuration view.
+    """
+
+    _fieldsets = {}
+    for key, value in COMMON_FIELDSETS.items():
```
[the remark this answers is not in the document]
sergei-maertens: Let's be pragmatic and add a `# TODO -> define proper json schema once the mozilla_django_oidc_db rework has landed`

src/openforms/authentication/contrib/generic_oidc/constants.py
```diff
@@ -0,0 +1 @@
+OIDC_ID_TOKEN_SESSION_KEY = "oidc_id_token"
```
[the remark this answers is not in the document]
robinmolen: I thought it would be best to keep constants and modular functionality separate.

Also, I've no idea what the impact of import stuff is, but to point the import in the test files to the constants file seemed better..

src/openforms/authentication/contrib/yivi_oidc/tests/test_views.py
```diff
@@ -0,0 +1,251 @@
...
+    def test_create_condiscon_scope_with_only_bsn_attribute(self):
+        plugin_options: YiviOptions = {
+            "authentication_options": [YiviAuthenticationAttributes.bsn],
+            "additional_attributes_groups": [],
+            "bsn_loa": "",
+            "kvk_loa": "",
+        }
+        condiscon_scope = OIDCAuthenticationInitView._yivi_condiscon_scope(
+            plugin_options
+        )
```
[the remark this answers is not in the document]
robinmolen: The reason why i've opted to test the individual functions/methods, instead of the init flow, is because testing against Signicat (or at least, as far as i could see) is fickle:

- The need of user input with the QR code, even when we use VCR testing this is needed for recording
- The need to test from a https environment AND with a static domain/port/host (this needs to be explicitly defined in Signicat)

The init before Signicat interaction could maybe work, so this test i could probably replace... (will take a look). But the entire auth flow, might be difficult

src/openforms/authentication/tests/factories.py
```diff
@@ -90,3 +97,18 @@ class RegistratorInfoFactory(factory.django.DjangoModelFactory):
...
     class Meta:
         model = RegistratorInfo
+
+
+class YiviOpenIDConnectConfigFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = YiviOpenIDConnectConfig
+        django_get_or_create = ["id"]
+
+    id = factory.Faker("id")
```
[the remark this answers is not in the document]
robinmolen: Yeah, with the removal of the FK field this is no-longer needed. So i've removed it 👍

src/openforms/authentication/models.py
```diff
@@ -15,13 +16,15 @@
     AuthAttribute,
     LegalSubjectIdentifierType,
 )
+from .contrib.yivi_oidc.constants import PLUGIN_ID as YIVI_PLUGIN_ID
```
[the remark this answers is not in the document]
robinmolen: Mhh, then we need the plugins themselves to handle the "to auth context" proces (with Yivi we cannot rely on available attributes, as it supports kvk, bsn, pseudo, whatever..)

src/openforms/authentication/contrib/yivi_oidc/plugin.py
robinmolen: Ah okay, cool! Yeah that would massively simplify things

src/openforms/authentication/models.py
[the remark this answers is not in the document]
sergei-maertens: that could be viable - have the fallback approach defer it to the plugin itself. In tests we perform the json-schema validation, that should make it sufficiently robust.

src/openforms/authentication/contrib/yivi_oidc/views.py
```diff
@@ -0,0 +1,131 @@
...
+                ]
+            )
+
+        # Turn condiscon list into a string, and base64 encode it
+        condiscon_string = json.dumps(condiscon_items)
+
+        # Create a signicat param scope with the base64 condiscon string.
+        # https://developer.signicat.com/broker/signicat-identity-broker/authentication-providers/yivi.html#example-of-adding-condiscon-parameter-in-your-oidc-request
+        base64_bytes = base64.b64encode(condiscon_string.encode("ascii"))
+        return f"signicat:param:condiscon_base64:{base64_bytes.decode('ascii')}"
```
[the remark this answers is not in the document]
sergei-maertens: It's okay, just make sure to outline the plugin requirements in the configuration documentation and mention that we currently only support Signicat.

src/openforms/authentication/contrib/yivi_oidc/tests/test_plugin.py
```diff
@@ -0,0 +1,227 @@
...
+            ),
+            ({"kvk_claim": "12345678"}, YiviAuthenticationAttributes.kvk),
+            ({"pseudo_claim": "hash"}, YiviAuthenticationAttributes.pseudo),
+            ({}, YiviAuthenticationAttributes.pseudo),
+        )
+        for claims, expected_result in claims_expected_result_map:
+            with self.subTest(data=claims, expected_result=expected_result):
+                result = plugin._get_user_chosen_authentication_attribute(
+                    YiviAuthenticationAttributes.values, claims
+                )
```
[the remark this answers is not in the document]
robinmolen: I've replaced the `_get_user_chosen_authentication_attribute` specific tests for `transform_claims` tests. A part of the `transform_claims` proces is checking the chosen auth attributes, so this is still covered 👍

src/openforms/authentication/contrib/yivi_oidc/plugin.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # Mapping for additionally fetched claims
+    additional_claims: dict[str, str]
+
+
+@register(PLUGIN_ID)
+class YiviOIDCAuthentication(OIDCAuthentication[YiviClaims, YiviOptions]):
+    verbose_name = _("Yivi via OpenID Connect")
+    # Yivi can provide a range of auth attributes, based on form config and user choice
+    provides_auth = (AuthAttribute.bsn, AuthAttribute.kvk, AuthAttribute.pseudo)
```
sergei-maertens, line 69: this doesn't type check and will probably cause issues with the prefill plugins, as they check up on `provides_auth` afaik.

src/openforms/authentication/contrib/yivi_oidc/plugin.py
```diff
@@ -0,0 +1,225 @@
...
+            "authorizee": {
+                "legalSubject": {
+                    "identifierType": auth_attribute.value,
+                    "identifier": auth_info.value,
+                    "additionalInformation": cast(dict, auth_info.additional_claims),
+                }
+            },
+        }
+        if auth_attribute is not AuthAttribute.pseudo:
+            yivi_context["levelOfAssurance"] = auth_info.loa
```
sergei-maertens, line 171: we need to set a default of `unknown`, right?

src/openforms/authentication/contrib/yivi_oidc/tests/test_auth_context_data.py
```diff
@@ -0,0 +1,114 @@
...
+
+from openforms.authentication.constants import AuthAttribute
+from openforms.authentication.models import AuthInfo
+from openforms.authentication.tests.factories import AuthInfoFactory
+from openforms.authentication.tests.utils import AuthContextAssertMixin
+
+from ..constants import PLUGIN_ID
+
+
+class YiviAuthContextUnitTests(AuthContextAssertMixin, TestCase):
```
sergei-maertens, line 13: please assert that it passes the json-schema validation too!

src/openforms/authentication/contrib/yivi_oidc/tests/test_auth_flow_init.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class YiviInitTests(IntegrationTestsBase):
+    """
+    Test the outbound part of OIDC-based Yivi authentication.
+    """
+
+    CALLBACK_URL = f"http://testserver{reverse_lazy('oidc_authentication_callback')}"
+
+    def _assert_condiscon_scope(self, scope: str, expected: list[list[list[str]]]):
```
sergei-maertens, line 26:
```suggestion
    def assertCondisconScope(self, scope: str, expected: list[list[list[str]]]):
```
within this test case, it's just public API - you're free to define your own assertion helpers (and we use the same existing unittest.TestCase assert naming style)

src/openforms/authentication/contrib/yivi_oidc/tests/test_auth_flow_init.py
```diff
@@ -0,0 +1,234 @@
...
+        self.assertEqual(redirect_target.host, "localhost")
+        self.assertEqual(redirect_target.port, 8080)
+        self.assertEqual(
+            redirect_target.path,
+            "/realms/test/protocol/openid-connect/auth",
+        )
+        self.assertEqual(
+            query_params["scope"],
+            ("openid signicat:param:condiscon_base64:W1tbImF0dHJpYnV0ZS5wc2V1ZG8iXV1d"),
+        )
```
sergei-maertens, line 68: why don't you assert the condiscon scope instead of this magic base64 thing? Semantically, I find the `assertCondisconScope` much more clear.

src/openforms/authentication/contrib/yivi_oidc/tests/test_auth_flow_init.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self.assertEqual(
+            query_params["scope"],
+            (
+                "openid "
+                "signicat:param:condiscon_base64:"
+                "W1tbImF0dHJpYnV0ZS5ic24iXV0sIFtbImZpcnN0X25hbWUiLCAibGFzdF9u"
+                "YW1lIl0sIFtdXSwgW1siZW1haWxfYWRkcmVzcyJdLCBbXV1d"
+            ),
+        )
```
sergei-maertens, line 101: probably you rather want to assert that `openid` is in `scope.split(" ")` - the order is not relevant afaik and you have the condiscon assert below anyway. I don't find the asserts on the raw base64 strings very meaningful.

src/openforms/authentication/tests/factories.py
```diff
@@ -90,3 +97,16 @@ class RegistratorInfoFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = RegistratorInfo
+
+
+class YiviOpenIDConnectConfigFactory(factory.django.DjangoModelFactory):
```
sergei-maertens, line 102: we shouldn't need this factory at all

src/openforms/authentication/tests/factories.py
```diff
@@ -90,3 +97,16 @@ class RegistratorInfoFactory(factory.django.DjangoModelFactory):
...
+    class Meta:
+        model = YiviOpenIDConnectConfig
+        django_get_or_create = ["id"]
+
+    id = factory.Faker("id")
+
+
+class AttributeGroupFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = AttributeGroup
```
sergei-maertens, line 112: please define a `factory.Sequence` for the `name`. It's a required and unique field.

src/openforms/authentication/contrib/yivi_oidc/models.py
```diff
@@ -0,0 +1,187 @@
...
+    name = models.CharField(
+        _("group name"),
+        max_length=100,
+        unique=True,
+        help_text=_(
+            "A human-readable name for the group of attributes, used in the form "
+            "configuration."
+        ),
+    )
+    description = models.CharField(
```
sergei-maertens, line 160: let's make this `blank=True`

src/openforms/authentication/types.py
```diff
@@ -96,3 +96,33 @@ class EmployeeContext(TypedDict):
...
+
+class YiviAuthorizee(TypedDict):
+    legalSubject: YiviLegalSubject
+
+
+# This type definition is highly experimental, and will most definitely change
+class YiviContext(TypedDict):
+    source: Literal["yivi"]
+    # The levelOfAssurance changes based on the used configured auth attribute
+    levelOfAssurance: NotRequired[
```
sergei-maertens, line 115: should be `Required` and add the `Literal["unknown"]`, no?

src/openforms/contrib/auth_oidc/plugin.py
```diff
@@ -0,0 +1,148 @@
...
+class AuthInit(Protocol):
+    def __call__(
+        self, request: HttpRequest, return_url: str, options: OptionsT, *args, **kwargs
+    ) -> HttpResponseBase: ...
+
+
+# can't bind T to JSONObject because TypedDict and dict[str, ...] are not considered
+# assignable... :(
+class OIDCAuthentication[T, OptionsT](BasePlugin[OptionsT]):
+    verbose_name: StrOrPromise = ""
```
sergei-maertens, line 56: I don't understand why this override is needed? The base class sets a message already.

src/openforms/contrib/auth_oidc/plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    provides_auth: AuthAttribute
```
sergei-maertens, line 57: why declare this again, the base class already has it?

sergei-maertens (changes requested):

src/openforms/authentication/contrib/yivi_oidc/plugin.py
robinmolen: Prefill seems to work with the `auth_info.attribute`, which for Yivi will be set to the authentication option chosen by the user. (If the user provides bsn data, that the auth_info `attribute` will be set to "bsn").

The prefill `CoSignValidator` does use the `provides_auth`, which will have to be updated

src/openforms/authentication/contrib/yivi_oidc/tests/test_auth_flow_init.py
robinmolen: Mhh yeah I thought it would be best to make it explicit which `condiscon_base64` we expect (thats why I added it to the other tests as well).

For this test I thought the condiscon assert wasn't very interesting, but to be concise/precise it should indeed be added..

I'll remove the precise `condiscon_base64` scope assertion, and add the `assertCondisconScope`

src/openforms/authentication/contrib/yivi_oidc/plugin.py
sergei-maertens: #5412 addresses all this :)

> Prefill seems to work with the auth_info.attribute, which for Yivi will be set to the authentication option chosen by the user.

I was confusing it with `prefill_plugin.requires_auth` which is the counterpart of `auth_plugin.provides_auth`.

> The prefill CoSignValidator does use the provides_auth, which will have to be updated (if we want to support Yivi with multiple provides_auth values)

Adapted - decided to not support Yivi in the legacy cosign. For the "modern" cosign, it's fine as is.

PR #5375, opened 2025-06-10
Align clearOnHide default value with formio.js SDK

sergei-maertens:
Part of open-formulieren/formio-renderer#59

The Formio.js SDK uses a default value of `true` for `clearOnHide`.

In practice this shouldn't matter much because the default value should have been baked in the JSON already, but in the future we will use more minimal schemas (probably), and then the behaviour should be properly aligned.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


viktorvanwijk:
> but in the future we will use more minimal schemas

Do you mean components might be defined by a subset of the properties? For example in the form builder you might not be able to see a 'clearOnHide' in the JSON if the corresponding checkbox was not activated?

viktorvanwijk (approved):

sergei-maertens:
> > but in the future we will use more minimal schemas
> 
> Do you mean components might be defined by a subset of the properties? For example in the form builder you might not be able to see a 'clearOnHide' in the JSON if the corresponding checkbox was not activated?

Yes - now Formio dumps all the JSON, even for properties we don't support, and our types only define a minimal set of properties as required.

In practice, we already set defaults in the builder for this, but relying on that is a bit fragile.

The DX is much better with minimal schemas, for testing purposes, it introduces less friction.

PR #5380, opened 2025-06-16
[#5268] Allow mapping partners components to StUF-ZDS registration options

### LINKED ISSUE #5268: Allow mapping partners/children components to registration options

- [x] StUF-ZDS

vaszig:
Closes #5268 partly

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -3465,3 +3472,249 @@ def test_register_submission_with_payment(self):
...
+            "openforms.config.models.GlobalConfiguration.get_solo",
+            return_value=global_config,
+        )
+
+        hc_config = hc_config_patcher.start()
+        global_config = global_config_patcher.start()
+        cls.addClassCleanup(hc_config_patcher.stop)
+        cls.addClassCleanup(global_config_patcher.stop)
+
+    def test_create_zaak_with_partners_as_betrokkene(self):
```
sergei-maertens, line 3524: can you also assert that it's not pesent in extra elementen, both the `partnersKey` and `partners_immutable`?

src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -3465,3 +3472,249 @@ def test_register_submission_with_payment(self):
...
+            "//zkn:object/zkn:heeftAlsOverigBetrokkene[@stuf:entiteittype='ZAKBTROVR']/zkn:gerelateerde/zkn:natuurlijkPersoon/bg:voornamen": 2,
+            "//zkn:object/zkn:heeftAlsOverigBetrokkene[@stuf:entiteittype='ZAKBTROVR']/zkn:gerelateerde/zkn:natuurlijkPersoon/bg:voorletters": 2,
+            "//zkn:object/zkn:heeftAlsOverigBetrokkene[@stuf:entiteittype='ZAKBTROVR']/zkn:gerelateerde/zkn:natuurlijkPersoon/bg:geslachtsnaam": 2,
+            "//zkn:object/zkn:heeftAlsOverigBetrokkene[@stuf:entiteittype='ZAKBTROVR']/zkn:gerelateerde/zkn:natuurlijkPersoon/bg:geboortedatum": 2,
+        }
+
+        for path, count in partners_paths.items():
+            self.assertXPathCount(xml_doc, path, count)
+
+    def test_create_zaak_with_partners_as_extraElementen(self):
```
sergei-maertens, line 3574: can you also assert that it's not present in zaakbetrokkenen?

src/stuf/stuf_zds/client.py
```diff
@@ -218,7 +219,7 @@ def create_zaak_identificatie(self) -> str:
     def create_zaak(
         self,
         zaak_identificatie: str,
-        zaak_data: dict,
+        zaak_data: MutableMapping[str, Any],
```
sergei-maertens, line 222:
```suggestion
        zaak_data: MutableMapping[str, object],
```
recently Chris pointed out that `object` is a better alternative than `Any` - it will still type check because everything is an object in python, but it doesn't taint the types in the rest of the codebase.

@CharString any chance you can provide an example where the `Any` messes up things?

src/stuf/stuf_zds/templates/stuf_zds/soap/includes/partners.xml
```diff
@@ -0,0 +1,34 @@
...
+        <ZKN:heeftAlsOverigBetrokkene StUF:entiteittype="ZAKBTROVR" StUF:verwerkingssoort="T">
+            <ZKN:gerelateerde>
+                <ZKN:natuurlijkPersoon StUF:entiteittype="NPS" StUF:verwerkingssoort="T">
+                {% if partner.bsn %}
+                    <BG:inp.bsn>{{ partner.bsn }}</BG:inp.bsn>
+                    {% if partner.prefilled %}
+                        <BG:authentiek StUF:metagegeven="true">J</BG:authentiek>
+                    {% else %}
+                        <BG:authentiek StUF:metagegeven="true">J</BG:authentiek>
+                    {% endif %}
```
sergei-maertens, line 12:
```suggestion
                    {% if partner.prefilled %}
                        <BG:authentiek StUF:metagegeven="true">J</BG:authentiek>
                    {% else %}
                        <BG:authentiek StUF:metagegeven="true">N</BG:authentiek>
                    {% endif %}
```
I think it's supposed to be `N`, but do check with the XSD's!

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -219,6 +227,76 @@ def get_extra_data(
...
+                # xml file generation depends on whether the variable was prefilled or not
+                # (authentiek)
+                state = submission.load_submission_value_variables_state()
+                submission_variable = state.get_variable(fm_immutable_variable.key)
+
+                # update the zaak data with the information needed for the xml request
+                for partner in zaak_data[RegistrationAttribute.partners]:
+                    partner.update(
+                        {
+                            **partner,
```
sergei-maertens, line 272:
```suggestion
```
`.update` with keys that already exist doesn't do anything, right?

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -219,6 +227,76 @@ def get_extra_data(
...
+                )
+                value = extra_data[from_key]
+                for item in value:
+                    item["description"] = description
+                    del item["dateOfBirthPrecision"]
+
+                extra_data[to_key] = value
+                del extra_data[from_key]
+
+        return zaak_data, extra_data
```
sergei-maertens, line 298: because we are doing mutation, I would not return anything from this function and update the type hint for the return value to `-> None`. That makes it crystal clear that we're mutating the dicts.

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -259,6 +337,13 @@ def register_submission(
                 zaak_data["locatie"]["key"] = component["key"]
 
             extra_data = self.get_extra_data(submission, options)
+
+            # mutate the zaak_data and the extra data for the partners component according
+            # to the type of the registration we want to send (zaakbetrokkene or extraElementen)
+            zaak_data, extra_data = self.process_partners(
+                submission, zaak_data, extra_data
+            )
```
sergei-maertens, line 345:
```suggestion
            self.process_partners(submission, zaak_data, extra_data)
```

sergei-maertens (changes requested):

src/stuf/stuf_zds/client.py
vaszig: I used `Any` because it's more flexible (for example when I do `for partner in zaak_data[RegistrationAttribute.partners]` the type checker complains that `object` is not iterable).

sergei-maertens: okay for now then, hopefully we'll be able to clean that up later :)

sergei-maertens (approved):

PR #5392, opened 2025-06-27
:memo: Document why we don't perform XSD validation in stuf-zds

sergei-maertens:


sergei-maertens:
Merging this despite the failing webkit e2e test, that has a different cause and is being fixed in #5393

PR #5400, opened 2025-07-01
Enable pyupgrade checks

sergei-maertens:
General QOL

[skip: e2e]


src/openforms/tests/runner.py
```diff
@@ -40,7 +40,7 @@ def _setup_random_state():
         encoded_state = base64.b64encode(
             pickle.dumps(factory.random.get_random_state())
         )
-        print("Current random state: %s" % encoded_state.decode("ascii"))
+        print("Current random state: {}".format(encoded_state.decode("ascii")))
```
viktorvanwijk, line 43: These were all changed using an auto-fix, right? Interesting that it doesn't use an f-string here :thinking:

viktorvanwijk (approved):

src/openforms/tests/runner.py
sergei-maertens: Indeed, automatic fix here.

PR #5408, opened 2025-07-04
[#5268] Add Partners custom component node for the email registration

### LINKED ISSUE #5268: Allow mapping partners/children components to registration options

vaszig:
Closes #5268 partly

**Changes**

- Added the custom component node for the Partners component. 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


vaszig:
I haven't touched the SDK yet. Let's see that this is the correct approach first and then I guess we will have to render only the `<EmptyValue>` here?

https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/FormStepSummary/ComponentValueDisplay.jsx#L177

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +350,60 @@ def is_visible(self) -> bool:
...
+                depth=self.depth + 1,
+                path=f"{self.key}.{node_index}",
+                parent_node=self,
+            )
+
+            for key, label in fields:
+                yield ComponentNode.build_node(
+                    step_data=self.step_data,
+                    component={
+                        "type": key,
```
sergei-maertens, line 401: careful here - this must be a valid/supported component type, so I think you'll have to add that to `fields`. Though I'd recommend just doing something like:

```py
components: list[Component] = [
    {
        "type": "bsn",
        "key": "bsn",
        "label": _("BSN"),
    },
    {
        "type": "textfield",
        "key": "initials",
        "label": _("Initials"),
    },
    ...
]
```
which gives a bit more type-safety too and keeps with the existing pattern of defining the formio structure in memory.

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +350,60 @@ def is_visible(self) -> bool:
...
+
+@register("partners")
+class PartnersNode(ComponentNode):
+    @property
+    def value(self) -> Any:
+        value = glom(self.step_data.data, self.path or self.key)
+        return value
+
+    @property
+    def display_value(self) -> str | Any:
```
sergei-maertens, line 363: can we avoid the `Any` and use a more specific type instead?

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +350,60 @@ def is_visible(self) -> bool:
         registration data...
         """
         return False
+
+
+@register("partners")
+class PartnersNode(ComponentNode):
+    @property
+    def value(self) -> Any:
+        value = glom(self.step_data.data, self.path or self.key)
```
sergei-maertens, line 359: `self.step_data` is of type `FormioData` and `glom` has serious performance issues, instead you should be able to simply do:

```py
value = self.step_data[self.path or self.key]
```
thanks to @viktorvanwijk 's efforts :)

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +350,60 @@ def is_visible(self) -> bool:
...
+            ("dateOfBirth", _("Date of birth")),
+        ]
+
+        for node_index in range(repeats):
+            yield ComponentNode.build_node(
+                step_data=self.step_data,
+                component={
+                    "type": "component_label",
+                    "key": "component_label",
+                    "label": f"Partner {node_index + 1}",
```
sergei-maertens, line 389: this label should be marked translatable

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +350,60 @@ def is_visible(self) -> bool:
...
+    @property
+    def display_value(self) -> str | Any:
+        # in export mode, expose the raw datatype
+        if self.mode == RenderModes.export:
+            return self.value
+
+        if isinstance(self.value, list):
+            return ""
+
+        return format_value(self.component, self.value, as_html=self.renderer.as_html)
```
sergei-maertens, line 371: I think we have some kitchensink tests somewhere that should lead to this codepath being covered too

src/openforms/formio/rendering/tests/test_vanilla_formio_components.py
```diff
@@ -1236,3 +1236,51 @@ def test_fieldset_with_logic_depending_on_selectboxes(self):
         # Check that the fieldset is present
         # Nodes: Form, SubmissionStep, Fieldset, Component (textfield), Component (radio)
         self.assertEqual(len(nodes), 5)
+
+    def test_partners(self):
```
sergei-maertens, line 1240: this is the wrong test file - it's meant for built-in formio component types (vanilla in the name + see the docstring at the beginning of the test file). please put the tests for custom components in a different file

src/openforms/formio/rendering/tests/test_vanilla_formio_components.py
```diff
@@ -1236,3 +1236,51 @@ def test_fieldset_with_logic_depending_on_selectboxes(self):
...
+            },
+        )
+
+        renderer = Renderer(submission, mode=RenderModes.registration, as_html=True)
+
+        nodes = list(renderer)
+
+        # Nodes: Form, SubmissionStep, Partners, Partner index x2, Partner limited fields x2
+        # (bsn, initials, affixes, lastname, date of birth)
+        self.assertEqual(len(nodes), 15)
```
sergei-maertens, line 1286: this will need some more assertions on the actual content of the nodes!

sergei-maertens (changes requested):
The general approach/structure are great, you can proceed with the SDK changes

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +349,90 @@ def is_visible(self) -> bool:
...
+    @property
+    def display_value(self) -> str | list[PartnerValue]:
+        # in export mode, expose the raw datatype
+        if self.mode == RenderModes.export:
+            return self.value
+
+        if isinstance(self.value, list):
+            return ""
+
+        return format_value(self.component, self.value, as_html=self.renderer.as_html)
```
vaszig, line 380: I have added a test for this, I don't get it why is still failing

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +349,90 @@ def is_visible(self) -> bool:
...
+        return value
+
+    @property
+    def display_value(self) -> str | list[PartnerValue]:
+        # in export mode, expose the raw datatype
+        if self.mode == RenderModes.export:
+            return self.value
+
+        if isinstance(self.value, list):
+            return ""
```
viktorvanwijk, line 378: For my understanding, why return an empty string here?

src/openforms/submissions/tests/test_submission_report.py
```diff
@@ -406,12 +431,23 @@ def test_report_is_generated_in_same_language_as_submission(self):
...
 
+        # Partners component tests
+        if component_type == "partners":
+            del localised_input[0]["firstNames"]
+            del localised_input[0]["dateOfBirthPrecision"]
+            for partner_label, partner_value in localised_input[0].items():
+                self.assertIn(partner_label, html_report)
+                self.assertIn(partner_value, html_report)
+                self.assertIn(localised_input, html_report)
+
```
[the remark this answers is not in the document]
vaszig: Oops, of course you are right. Thanks for catching it.

src/openforms/formio/rendering/default.py
vaszig: Yes, this is what I understand, I guess this should be removed.

src/openforms/formio/rendering/default.py
```diff
@@ -345,3 +346,91 @@ def is_visible(self) -> bool:
...
+    first_names: NotRequired[str]
+    last_name: NotRequired[str]
+    date_of_birth: NotRequired[str]
+    date_of_birth_precision: Literal["date", "year_month", "year"] | None
+
+
+@register("partners")
+class PartnersNode(ComponentNode):
+    @property
+    def value(self) -> Any:
```
sergei-maertens, line 364: this can probably be the proper return type?


```suggestion
    def value(self) -> Sequence[PartnerValue]:
```

sergei-maertens (commented):
Some tweaks that I didn't want to block the PR on - feel free to pick those up if you want.

PR #5409, opened 2025-07-04
Add confirmation email(s) to zaak

### LINKED ISSUE #4877: Add a copy of the confirmation "e-mail" as a attachment to a zaak
### Thema / Theme

Form designer

As a form designer i want to be able to include the confirmation that was sent to the initiator together with all other attachments/informatieobjecten to a zaak. 
Including this copy of the confirmation and it's content will be "proof" in regards of the requirements for the wmebv.

_No response_

_No response_

viktorvanwijk:
Closes #4877

**Changes**

* Added celery task that executes after sending confirmation email
* Added possibility to attach confirmation emails to a zaak in the ZGW API and StUF ZDS registrations

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/emails/service.py
```diff
@@ -0,0 +1,56 @@
...
+    :return: A tuple of the HTML content and message ID of the last created confirmation
+      email for the specified submission. ``None`` if there weren't any.
+    """
+    # We only care about messages that were created after the submission
+    messages = Message.objects.filter(date_created__gte=submission.created_on).order_by(
+        "-date_created"
+    )
+
+    for message in messages:
+        email = message.get_email_message()
```
sergei-maertens, line 30: we can narrow down the queryset more before checking in python code by filtering on the `_message_data` text field which contains the raw e-mail message.

```suggestion
    messages = Message.objects.filter(
        date_created__gte=submission.created_on,
        _message_data__contains=X_OF_CONTENT_TYPE_HEADER,
        _message_data__contains=EmailContentTypeChoices.submission,
        _message_data__contains=str(submission.uuid),
    ).order_by(
        "-date_created"
    )

    for message in messages:
        email = message.get_email_message()
```
this should substantially narrow down the result set.

It's probably a good idea to test this on a big environment with many email logs to see the performance of the query.

src/openforms/emails/service.py
```diff
@@ -0,0 +1,56 @@
...
+
+        has_submission = (
+            email.extra_headers.get(X_OF_CONTENT_TYPE_HEADER)
+            == EmailContentTypeChoices.submission
+        )
+        if not has_submission:
+            continue
```
sergei-maertens, line 37: you can trigger this by using `@override_settings` to pin the email backend to django-yubin in tests (django automatically patches this to the in memory backend) and use a plain `django.core.email.send_mail` for some bogus email that should end up in the table as well. Then just make sure to include our "magic markers" in the email body to get false positives and verify they're filtered out.

Orrr... this check may even not be relevant at all because of the checks we perform later anyway to correlate the message and submission :thinking:

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -477,6 +478,58 @@ def get_config_actions(self):
             ),
         ]
 
+    def update_registration_with_confirmation_email(
+        self, submission: Submission, options: RegistrationOptions
+    ) -> dict:
+        logger.info("stuf_zds_finalise_registration_started", options=options)
```
sergei-maertens, line 484: Probably best to update the event here?

```suggestion
        logger.info("update_registration_with_confirmation_email_started", options=options)
```
the structlog context should already have bound the plugin in the outer/generic code, so I think this is sufficient to identify which plugin ran.

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -477,6 +478,58 @@ def get_config_actions(self):
...
+        res = get_last_confirmation_email(submission)
+        if res is None:
+            logger.warning(
```
sergei-maertens, line 487: I think we can lower this to `info` instead of `warning`, I'd expect confirmation email failures to be logged themselves. This may also fail because there was no recipient e-mail address configured/provided.

src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -477,6 +478,58 @@ def get_config_actions(self):
...
+                "stuf_zds_finalise_registration_aborted",
+                reason="confirmation_email_not_sent",
+            )
+            raise RegistrationFailed("Confirmation email was not sent")
```
sergei-maertens, line 491: probably we should identify if a confirmation e-mail is supposed to have been sent in the first place. It'd be nice if we can rely on `submission.confirmation_email_sent` for this, but then we probably need to update `send_confirmation_email` to set this flag if there are no `to_emails`.

src/openforms/submissions/tasks/emails.py
```diff
@@ -29,6 +29,7 @@
     "send_email_cosigner",
 ]
 
+from ...registrations.tasks import update_registration_with_confirmation_email
```
sergei-maertens, line 32:
```suggestion
from openforms.registrations.tasks import update_registration_with_confirmation_email
```

src/openforms/utils/pdf.py
```diff
@@ -109,3 +109,16 @@ def render_to_pdf(template_name: str, context: dict) -> tuple[str, bytes]:
...
+    """Convert a string with HTML to a PDF."""
+    import weasyprint  # heavy import
+
+    html_object = weasyprint.HTML(
+        string=html,
+        url_fetcher=UrlFetcher(),
+        base_url=settings.BASE_URL,
+    )
+    pdf: bytes = html_object.write_pdf(pdf_variant="pdf/ua-1")
+    return pdf
```
sergei-maertens, line 124: good to duplicate the code from `render_to_pdf` - that will be replaced by `maykin-common` PDF utilities.

src/openforms/emails/service.py
viktorvanwijk: Would there be a way to do that before releasing it first?

src/openforms/registrations/contrib/stuf_zds/plugin.py
viktorvanwijk: Forgot to change these when renaming

src/openforms/submissions/tasks/emails.py
viktorvanwijk: Stupid auto-import :vomiting_face:

src/openforms/emails/tests/test_service.py
```diff
@@ -0,0 +1,60 @@
+from django.test import TestCase, override_settings
+
+from ...submissions.tests.factories import SubmissionFactory
```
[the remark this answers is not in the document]
viktorvanwijk: Stupid auto-import :vomiting_face:

src/openforms/emails/tests/test_service.py
```diff
@@ -0,0 +1,60 @@
...
+    def test_multiple_confirmation_emails(self):
+        submission = SubmissionFactory.create()
+
+        self.send_email(submission, "Email 1", EmailEventChoices.confirmation)
+        self.send_email(submission, "Email 2", EmailEventChoices.confirmation)
+        self.send_email(submission, "Email 3", EmailEventChoices.confirmation)
+
+        content, _ = get_last_confirmation_email(submission)
+
+        self.assertIn("Email 3", content)
```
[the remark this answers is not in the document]
viktorvanwijk: Isn't that already implied by what the function returns? If someone modifies `get_last_confirmation_email` to combine content of multiple emails, they deserve a slap on the wrist no?

src/openforms/emails/service.py
sergei-maertens: Let me see if I can get my hands on a dataset tomorrow!

sergei-maertens: So - the conclusion here for me would be:

* Add a migration in our own `emails` app to add an index to django yubins `Message` model. You can use `SeparateDatabaseAndState` for this.
* Add the submission UUID query/filter to the queryset, but not more than that. The other fluff seems to make query performance worse
* The high specificity should result in very few messages being loaded into memory to perform the remaining header checks. I'd use a `.iterator()` on the queryset though to be sure we don't load too much data in memory, since those e-mail messages can contain attachments as well.

viktorvanwijk: Thank you :bow:

I will have a look!

src/openforms/utils/pdf.py
sergei-maertens: yes

sergei-maertens (approved):

src/openforms/emails/migrations/0004_add_index_to_django_yubin_message_date_created.py
```diff
@@ -0,0 +1,26 @@
...
+                    sql=(
+                        "CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_date_created_desc "
+                        "ON django_yubin_message (date_created DESC);"
+                    ),
+                    reverse_sql="DROP INDEX CONCURRENTLY IF EXISTS idx_date_created_desc;",
+                )
+            ],
+            state_operations=[],
+        )
+    ]
```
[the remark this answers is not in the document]
viktorvanwijk: To be honest, I did not come across `AddIndex` in my quest of making this migration. But yeah, looks like it's not possible to use `CONCURRENTLY` with that.

Will remove the `SeparateDatabaseAndState` then :)

PR #5431, opened 2025-07-14
Obfuscate Yivi and eIDAS claims for logging

### LINKED ISSUE #5430: eIDAS and Yivi claims logging
### Thema / Theme

Plugin

Obfuscate any Yivi and eIDAS claims that could be sensitive, before logging them.

* eIDAS for person: `legal_subject_identifier_claim`, `legal_subject_first_name_claim`, `legal_subject_family_name_claim`.
* eIDAS for company: `legal_subject_identifier_claim`, `acting_subject_identifier_claim`, `acting_subject_first_name_claim`, `acting_subject_family_name_claim`.
* Yivi: `bsn_claim`, `kvk_claim` and any additional claim (we don't know/can't know what the additional claims could be, so lets handle them all as sensitive)

_No response_

_No response_

robinmolen:
Closes #5430 

**Changes**

For the claims logging, we need to obfuscate all sensitive Yivi and eIDAS claims.

Yivi:
Because municipalities control which additional attributes are requested during login (which results into the claims), we cannot know which claims are and aren't sensitive. To be safe, we should handle all Yivi additional claims as sensitive. Besides that, we also handle the bsn, kvk and pseudo claims as sensitive.

eIDAS for person:
For personal eIDAS authentication, we regard the legal identifier and additional personal data (firstname and lastname) as sensitive.

eIDAS for companies:
For eIDAS authentication as a company, we regard the legal and acting identifiers as sensitive. In addition, the additional acting personal data (firstname and lastname) are also handled as sensitive.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/contrib/auth_oidc/backends.py
+++ b/src/openforms/contrib/auth_oidc/backends.py
@@ -37,6 +37,12 @@ class GenericOIDCBackend(OIDCAuthenticationBackend):
             return super()._check_candidate_backend()
         return False
 
+    @property
+    def auth_plugin(self):
+        config_to_plugin = get_config_to_plugin()
+        assert self.config_class and self.config_class in config_to_plugin
+        return config_to_plugin[self.config_class]
+
     def get_or_create_user(
         self, access_token: str, id_token: str, payload: JSONObject
     ) -> AnonymousUser:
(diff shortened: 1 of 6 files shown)


viktorvanwijk (commented):
Some minor things :)

General thought: you could add an option to the attribute groups which indicate whether they are sensitive or not. Though, this might be a little overkill if this is only relevant for the logging

src/openforms/contrib/auth_oidc/backends.py
```diff
@@ -65,6 +65,45 @@ def get_or_create_user(
...
+        )
+        authentication_backend_options = {}
+        if authentication_backend:
+            authentication_backend_options = authentication_backend.options or {}
+
+        # Obfuscate claims using static and dynamic lists of sensitive plugin claims
+        return obfuscate_claims(
+            claims,
+            self.OIDCDB_SENSITIVE_CLAIMS
+            + plugin.sensitive_claims(authentication_backend_options, claims),
```
sergei-maertens, line 104: please use `get_sensitive_claims` for methods that take arguments.

if it's a method that doesn't require arguments, you want to define a property instead so you can do `plugin.sensitive_claims`:

```py
@property
def sensitive_claims(self) -> Sequence[...]: ...
```

src/openforms/contrib/auth_oidc/backends.py
```diff
@@ -65,6 +65,45 @@ def get_or_create_user(
...
+        except (Form.DoesNotExist, FormAuthenticationBackend.DoesNotExist):
+            return None
+
+        return auth_backend
+
+    def _obfuscate_claims(self, claims: JSONObject):
+        # Get plugin
+        config_to_plugin = get_config_to_plugin()
+        assert self.config_class and self.config_class in config_to_plugin
+        plugin = config_to_plugin[self.config_class]
```
sergei-maertens, line 90: if this is repeated multiple times, you might want to consider putting it in a property:

```py
@property
def auth_plugin(self):
    config_to_plugin = get_config_to_plugin()
    assert self.config_class and self.config_class in config_to_plugin
    return config_to_plugin[self.config_class]
```

this has the potential to be upgraded to `@cached_property` in the future if it's an expensive call

You can then replace your code with:


```suggestion
        plugin = self.auth_plugin
```

src/openforms/contrib/auth_oidc/plugin.py
```diff
@@ -104,6 +104,13 @@ def before_process_claims(self, config: BaseConfig, claims: JSONObject):
     def strict_mode(self, request: HttpRequest) -> bool:
         return False
 
+    def sensitive_claims(self, options: OptionsT, claims: JSONObject) -> list[str]:
```
sergei-maertens, line 107:
```suggestion
    def sensitive_claims(self, options: OptionsT, claims: JSONObject) -> Sequence[str]:
```
if possible, specifying the interface rather than the implementation is a bit more robust/flexible/future proof

in this case, I don't think the ordering is even relevant, so you may even want to use `Collection` instead

src/openforms/contrib/auth_oidc/backends.py
```diff
@@ -65,6 +65,45 @@ def get_or_create_user(
...
+
+        # Get authentication_backend options and make sure that they are a dict
+        authentication_backend = self._get_form_authentication_backend(
+            plugin.identifier
+        )
+        authentication_backend_options = {}
+        if authentication_backend:
+            authentication_backend_options = authentication_backend.options or {}
```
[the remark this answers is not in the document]
sergei-maertens: it is needed to transform the raw JSON data into appropriate Python/Django types. E.g. a `serializers.RelatedField` stores the PK or slug in the JSON, but resolves back to an actual model instance. Similarly, when you have a `serializers.DateField`, you want to receive `options["my_date"]` as a date instance, not a string representing a date.

It can also mean that over time it no longer is valid, an example of that is in the ZGW API registration options: selecting a case type that has a version/exists at the time of configuration can become invalid if the case type expires after a certain date and no new version exists for it anymore.

robinmolen:
> General thought: you could add an option to the attribute groups which indicate whether they are sensitive or not. Though, this might be a little overkill if this is only relevant for the logging

It does indeed sound nice.. I'll take a look at this, to see if it can easily be added

robinmolen:
The obfuscating of the claims has been included in the bigger OIDC refactor/solo model changes https://github.com/open-formulieren/open-forms/pull/5352

Closing this PR

PR #5438, opened 2025-07-18
[OF-SDK#825] Add custom children component

vaszig:
Closes open-formulieren/open-forms-sdk#825 partly

### Requires:

- [x] new version of formio-builder
- [x] open-formulieren/open-forms-sdk#847 to be merged

**Changes**

- Added the custom children component

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene


src/openforms/formio/components/custom.py
```diff
@@ -769,6 +770,135 @@ def as_json_schema(component: Component) -> JSONObject:
...
+        label=_("bsn"),
+        max_length=9,
+        help_text=_("The BSN of the child"),
+        validators=[BSNValidator()],
+    )
+    firstNames = serializers.CharField(
+        label=_("firstNames"),
+        help_text=_("The first names of the child"),
+        required=False,
+        allow_blank=True,
```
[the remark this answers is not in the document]
vaszig: Yes, it's required when the user adds a child manually (like the partners, only the BSN is required).

src/openforms/formio/components/custom.py
```diff
@@ -769,6 +770,135 @@ def as_json_schema(component: Component) -> JSONObject:
...
+    )
+    dateOfBirth = serializers.DateField(
+        label=_("date of birth"),
+        help_text=_("The date of birth of the child"),
+        required=False,
```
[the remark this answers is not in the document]
vaszig: Same as above.

src/openforms/submissions/tests/test_submission_step_validate.py
```diff
@@ -936,3 +936,143 @@ def test_partners_component_fails_when_data_is_tampered(self):
...
+                            "lastName": "Paassen",
+                            "firstNames": "Pero",
+                            "dateOfBirth": "2023-02-01",
+                            "dateOfBirthPrecision": "date",
+                            "selected": False,
+                            "__addedManually": False,
+                        },
+                    ]
+                }
+            },
```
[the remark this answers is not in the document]
vaszig: This is the data coming from the frontend. I will add a comment.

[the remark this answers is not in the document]
robinmolen: But then it should only contain `bsn`, `firstNames`, `dateOfBirth`, `selected` and `__addedManually` (and perhaps `__id`?) right?

```
"data": {
  "children": [
    {
     "bsn": "999970409",
      "firstNames": "Pero",
      "dateOfBirth": "2023-02-01",
      "selected": False,
      "__addedManually": False,
    },
  ]
}
```

vaszig:
@robinmolen FYI I have added the comments we discussed and besides that I added one more missing test for the valid data (without all the keys that the frontend returns) with only the required keys

https://github.com/open-formulieren/open-forms/pull/5438/files#diff-38bbaf27753ad6f2523d8168cf3546248409e1b4be515ef1f9effdfafef124b1R10

vaszig:
@robinmolen While I was trying to implement the second step of the children component flow, I found a couple of problems and things that I would like to improve so that's why I pushed one more commit. You can see in the description of the commit the bullet points.

src/openforms/formio/components/custom.py
```diff
@@ -770,6 +771,156 @@ def as_json_schema(component: Component) -> JSONObject:
...
+    def validate_list(self, children):
+        component_key = self.component["key"]
+        submission = self.context["submission"]
+        prefill_data = submission.get_prefilled_data()
+
+        fm_immutable_variable = FormVariable.objects.filter(
+            source=FormVariableSources.user_defined,
+            prefill_plugin=FM_PLUGIN_IDENTIFIER,
+            prefill_options__mutable_data_form_variable=component_key,
+            form=submission.form.id,
```
sergei-maertens, line 837:
```suggestion
            form=submission.form,
```
it's more pythonic to reason with the model instances than their underlying database ID

src/openforms/formio/components/custom.py
```diff
@@ -770,6 +771,156 @@ def as_json_schema(component: Component) -> JSONObject:
...
+            ]
+
+            if not self.validate_selected_children(children):
+                raise serializers.ValidationError(_("Invalid data."))
+
+            # We also receive the boolean `selected` which is set on the frontend and
+            # therefore should not be part of the data for validation (in combination
+            # with the retrieved data from the external source)
+            for child in children:
+                child.pop("selected", None)
```
sergei-maertens, line 869: I think it's a bit smelly that a `validate` function performs side effects by mutating its argument. To keep a pure function, it'd probably be best to `copy.deepcopy(children)`, otherwise weird things will happen if we will use the validation serializer output as data to save in variables (something we probably want to do in the long term)

src/openforms/formio/components/custom.py
```diff
@@ -770,6 +771,156 @@ def as_json_schema(component: Component) -> JSONObject:
...
+                        "lastName",
+                        "affixes",
+                        "initials",
+                    )
+                }
+                for child in prefill_data[fm_immutable_variable.key]
+            ]
+
+            if not self.validate_selected_children(children):
+                raise serializers.ValidationError(_("Invalid data."))
```
sergei-maertens, line 863: I would probably call `self.validate_selected_children(...)` and put the `raise` in that method, for consistency with other validation functions. Or, raise here, but rename the method to `def check_selected_key_tampering(...)` or something like that that makes it more clear that the return value is relevant

src/openforms/formio/components/custom.py
```diff
@@ -770,6 +771,156 @@ def as_json_schema(component: Component) -> JSONObject:
...
+                "properties": {
+                    "bsn": {
+                        "type": "string",
+                        "pattern": r"^\d{9}$",
+                        "format": "nl-bsn",
+                    },
+                    "firstNames": {"type": "string"},
+                    "dateOfBirth": {"type": "string", "format": "date"},
+                },
+                "additionalProperties": False,
```
sergei-maertens, line 917: is that true? will we not send extra properties like `affixes` and `lastName` etc.?

src/openforms/formio/components/custom.py
vaszig: Apologies, it was there before our last discussion and I forgot about it.

sergei-maertens (approved):

PR #5467, opened 2025-07-24
Data types

### LINKED ISSUE #2324: Normalize (variable) data types

### LINKED ISSUE #4883: template filters not working for date/time in confirmation

2.7.x - 2.8.x

### Customer reference

_No response_

it's impossible to use a date with a filter to present in the confirmation page and mail: 

```
today: {{ today|date:"d/m/Y" }}
datefromform: {{ datum|date:"d/m/Y" }}
```
results in: 
image
image

i expect the same result in both cases since both "today" and "datum" are dates.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

viktorvanwijk:
Closes #2324, closes #4883

Note that this [comment](https://github.com/open-formulieren/open-forms/issues/2324#issuecomment-2821603808) contains more issues that need to be solved, but I think it's better to close #2324 with this PR and open new issues for those.

**Changes**

* Cleaned-up methods to get data from the state
* Fixed values of date/time/datetime components configured as multiple not being converted to Python types
* Fixed values of partners component not being converted to Python types
* Removed on-the-fly conversions to Python types
* Added `data_type` and `data_subtype` to `SubmissionValueVariable`

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/handlers/v2.py
+++ b/src/openforms/registrations/contrib/objects_api/handlers/v2.py
@@ -4,7 +4,7 @@ Implementation details for the v2 registration handler.
 
 from collections.abc import Sequence
 from dataclasses import dataclass
-from datetime import date, datetime
+from datetime import date, datetime, time
 from typing import assert_never, cast
 
 from glom import Assign, Path, glom
@@ -72,8 +72,9 @@ def process_mapped_variable(
     :arg value: The raw value of the form variable for the submission being processed.
       The type/shape of the value depends on the variable/component data type being
       processed and even the component configuration (such as multiple True/False).
-    :arg transform_to_list: transform_to_list: Component keys in this list will be sent as an array of values rather than the default
-      object-shape for selectboxes components.
+    :arg transform_to_list: transform_to_list: Component keys in this list will be sent
+      as an array of values rather than the default object-shape for selectboxes
+      components.
     :arg component: If the variable corresponds to a Formio component, the component
       definition is provided, otherwise ``None``.
     :arg attachment_urls: The registration plugin uploads attachments to a Documents API
@@ -90,19 +91,21 @@ def process_mapped_variable(
     variable_key = mapping["variable_key"]
     target_path = Path(*bits) if (bits := mapping.get("target_path")) else None
 
-    # normalize non-primitive date/datetime values so that they're ready for JSON
-    # serialization in the proper format
-    if isinstance(value, date | datetime):
+    # normalize non-primitive date/datetime values for non components so that they're
+    # ready for JSON serialization in the proper format
+    if isinstance(value, date | datetime) and not component:
         value = value.isoformat()
-
-    # Date and datetime objects are now converted to strings, so we can safely cast to
-    # JSONValue
-    value = cast(JSONValue, value)
 
     # transform the value within the context of the component
     # TODO: convert this in a proper registry in due time so we can use better type
     # annotations
     match component:
+        case {"type": "date" | "datetime" | "time", "multiple": True}:
+            assert isinstance(value, list)
+            value = [v.isoformat() for v in value]  # pyright: ignore[reportAttributeAccessIssue, reportOptionalMemberAccess]
+        case {"type": "date" | "datetime" | "time"}:
+            assert isinstance(value, date | time | datetime)
+            value = value.isoformat()
         case {"type": "addressNL"}:
             assert isinstance(value, dict)
             value = value.copy()
@@ -114,7 +117,7 @@ def process_mapped_variable(
                     AssignmentSpec(
                         # the typeddict union of keys/values is lost when looping over them
                         destination=Path(*target_path_bits),  # pyright: ignore[reportGeneralTypeIssues]
-                        value=_value,
+                        value=_value,  # pyright: ignore[reportArgumentType]
                     )
                     for key, target_path_bits in detailed_mappings.items()
                     if target_path_bits
@@ -132,7 +135,7 @@ def process_mapped_variable(
             # map the address NL values as a whole
             else:
                 assert target_path is not None
-                return AssignmentSpec(destination=target_path, value=value)
+                return AssignmentSpec(destination=target_path, value=value)  # pyright: ignore[reportArgumentType]
 
         case {"type": "file"}:
             assert attachment_urls is not None
@@ -160,6 +163,7 @@ def process_mapped_variable(
 
             for partner in value:
                 assert isinstance(partner, dict)
+                assert isinstance(partner["dateOfBirth"], date)
 
                 partner["dateOfBirth"] = partner["dateOfBirth"].isoformat()
 
@@ -173,6 +177,9 @@ def process_mapped_variable(
             pass
 
     assert target_path is not None
+
+    # We can "safely" cast to a JSONValue now that the value has been processed
+    value = cast(JSONValue, value)
     return AssignmentSpec(destination=target_path, value=value)
 
 
--- a/src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
+++ b/src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
@@ -1606,3 +1606,47 @@ class V2HandlerTests(TestCase):
         data = record_data["data"]
 
         self.assertEqual(data, {"path1": ["option1"], "path2": {"option2": True}})
+
+    def test_date_related_components(self):
+        submission = SubmissionFactory.from_components(
+            [
+                {"key": "date", "type": "date"},
+                {"key": "date_multiple", "type": "date", "multiple": True},
+            ],
+            completed=True,
+            submitted_data={
+                "date": "2000-01-01",
+                "date_multiple": ["2000-01-01", "2025-08-11"],
+            },
+        )
+
+        ObjectsAPIRegistrationData.objects.create(submission=submission)
+        v2_options: RegistrationOptionsV2 = {
+            "objects_api_group": self.group,
+            "version": 2,
+            "objecttype": UUID("f3f1b370-97ed-4730-bc7e-ebb20c230377"),
+            "objecttype_version": 1,
+            "update_existing_object": False,
+            "auth_attribute_path": [],
+            "variables_mapping": [
+                {"variable_key": "date", "target_path": ["path1"]},
+                {"variable_key": "date_multiple", "target_path": ["path2"]},
+            ],
+            "iot_attachment": "",
+            "iot_submission_csv": "",
+            "iot_submission_report": "",
+            "transform_to_list": [],
+        }
+        handler = ObjectsAPIV2Handler()
+
+        record_data = handler.get_record_data(submission=submission, options=v2_options)
+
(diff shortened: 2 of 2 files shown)


docs/developers/backend/core/variables.rst
```diff
@@ -133,3 +133,21 @@ User defined ``SubmissionValueVariables`` are rendered when the renderer is in m
 ``registration`` (for the data sent to the registration backends). They are **NOT** included in the summary page of the
 form, the confirmation email or the PDF of the submission report.
 
+
+Working with submission data
+============================
+
+Accessing the values stored in the ``SubmissionValueVariables`` should done through the
```
sergei-maertens, line 140:
```suggestion
Accessing the values stored in the ``SubmissionValueVariables`` should be done through the
```

docs/developers/backend/core/variables.rst
```diff
@@ -133,3 +133,21 @@ User defined ``SubmissionValueVariables`` are rendered when the renderer is in m
...
+``SubmissionValueVariablesState``. There is a single method (``SubmissionValueVariablesState.get_data()``, see below)
+which collects all submission value variables of the corresponding submission, and converts them to native Python types.
+This means that submitted data for date, time, and datetime components will be converted to ``date``, ``time``, and
+``datetime`` objects respectively. It allows us to easily perform operations (comparison, relative deltas, etc.) on the
+values of these components, without having to deal with on-the-fly conversions from (ISO) strings to time-related
+objects.
+
+Note that the data is still stored as JSON in the database, and also exists our Python-type domain once we serialize
```
sergei-maertens, line 148:
```suggestion
Note that the data is still stored as JSON in the database, and also exits our Python-type domain once we serialize
```

src/openforms/formio/utils.py
```diff
@@ -189,6 +189,23 @@ def get_component_datatype(component):
     return COMPONENT_DATATYPES.get(component_type, FormVariableDataTypes.string)
 
 
+def get_component_data_subtype(component) -> str:
```
sergei-maertens, line 192:
```suggestion
def get_component_data_subtype(component: Component) -> str:
```

src/openforms/forms/models/form_variable.py
```diff
@@ -292,6 +294,16 @@ class FormVariable(models.Model):
         choices=FormVariableDataTypes.choices,
         max_length=50,
     )
+    data_subtype = models.CharField(
```
sergei-maertens, line 297: Should be have a data migration or `bin` script to set the subtype for form variables that currently have the type `array`?

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -327,6 +337,22 @@ class SubmissionValueVariable(models.Model):
...
+    data_subtype = models.CharField(
+        verbose_name=_("data subtype"),
+        help_text=_(
+            "This field represents the data type of the values inside the container "
+            "for components that are configured as 'multiple'."
+        ),
+        choices=FormVariableDataTypes.choices,
+        max_length=50,
+        blank=True,
+    )
```
sergei-maertens, line 355: same question about data migration here :)

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -359,23 +385,13 @@ def to_python(self, value: VariableValue | object = empty) -> VariableValue:
...
-                action="submissions.convert_value_to_python",
-                submission_id=self.submission_id,
-                key=self.key,
-                submission_value_id=self.pk,
-            )
-            return value
+        if not self.data_subtype:
+            return self._value_to_python(value, self.data_type)
+        else:
+            assert self.data_type == FormVariableDataTypes.array
```
sergei-maertens, line 391: we *can* express this in a database constraint I think:

```py
condition=(
    (models.Q(data_type=FormVariableDataTypes.array) & ~models.Q(data_subtype=""))
    | (~models.Q(data_type=FormVariableDataTypes.array) & models.Q(data_subtype=""))
)
```

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -419,4 +435,16 @@ def to_python(self, value: VariableValue | object = empty) -> VariableValue:
                 return value
             return parse_time(value)
 
+        if value and data_type == FormVariableDataTypes.partners:
+            # This is a work-around to convert the date of birth string into a date
+            # object. For now, we always expect the ``dateOfBrithPrecision`` to be
```
sergei-maertens, line 440:
```suggestion
            # object. For now, we always expect the ``dateOfBirthPrecision`` to be
```

src/openforms/variables/constants.py
```diff
@@ -21,6 +21,9 @@ class FormVariableDataTypes(models.TextChoices):
     datetime = "datetime", _("Datetime")
     time = "time", _("Time")
     date = "date", _("Date")
+    # This should only be used as a data subtype, and is a work-around to enable
+    # converting the partners submission data to native Python types
+    partners = "partners", _("Partners")
```
sergei-maertens, line 26: can also be expressed as check constraint!

src/openforms/variables/utils.py
```diff
@@ -34,26 +38,18 @@ def check_initial_value(initial_value: JSONObject, data_type: str) -> JSONObject
...
     if settings.ESCAPE_REGISTRATION_OUTPUT:
-        data = html_escape_json(data)
-
-    formio_data = FormioData(
-        **{
-            variable.key: variable.initial_value
-            for variable in get_static_variables(submission=submission)
-        },
-        **data,
-    )
```
sergei-maertens, line 58: we have a behaviour change here - the static variables are no longer overwritten by submission data. I'm okay with it, but we need to document this like [here](https://open-forms.readthedocs.io/en/3.2.0/changelog.html#upgrade-procedure). Can you add a notice already in the changelog so we don't forget?

sergei-maertens (commented):
Really, really nice work and I'm pleasantly surprised by the other small cleanups that can be done because `evaluate_form_logic` is now centralizing all that control!

src/openforms/forms/models/form_variable.py
viktorvanwijk: I was about to suggest that we only need to do this for components configured as multiple, because files and editgrids also have a type array. But this reminds me of something. Our lord and savior editgrid is currently not processed, so any date/time/datetime components that are inside will not be converted :(. I will have to do some more work I think...

sergei-maertens: I :pray: for you

src/openforms/forms/migrations/0113_add_missing_data_subtype_to_existing_form_variable.py
```diff
@@ -0,0 +1,51 @@
...
+from openforms.variables.constants import FormVariableDataTypes
+
+
+def add_data_subtype_to_existing_form_variables(apps: StateApps, _):
+    FormVariable = apps.get_model("forms", "FormVariable")
+    FormDefinition = apps.get_model("forms", "FormDefinition")
+
+    form_definition_iterator = FormDefinition.objects.prefetch_related(
+        "formvariable_set"
+    ).iterator()
```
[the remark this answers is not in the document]
viktorvanwijk: Ah I see, so basically I'm only loading extra stuff that is not used now...

viktorvanwijk:
I combined the data (sub)type migrations where possible, and also fixed some type checking errors that did not occur previously (not entirely sure why). Because there are a lot of changed files, here is a list of files that I changed after the last review
- openforms.registrations.contrib.objects_api.handlers.v2.py
- openforms.registrations.contrib.objects_api.submission_registration.py
- openforms.registrations.contrib.generic_json.plugin.py

sergei-maertens (approved):

PR #5470, opened 2025-07-28
Delete obsoleted Ansible deployment sample

sergei-maertens:
Very few people actually deploy using Ansible, and those who do are assumed to have sufficient knowledge. Instead, we now point to the Helm charts that people can use for production deployments on K8s.

[skip: e2e]


annashamray (approved):
:+1:

Issue #5384, opened 2025-06-18
Importing form with objects_api_group gives error


SilviaAmAm:
### Product versie / Product version

3.0.7

### Customer reference

Taiga dimpact-enschede-ssc-ict-1/issue/295

### Omschrijf het probleem / Describe the bug

When exporting a form from one environment to another where the objects api group are present, if there is a mistmatch between the PK of the objects api groups, this can lead to this error (or the usage of the wrong objects_api_group):

```
Er is iets foutgegaan bij het importeren van formulier: {'registration_backends': [{'options': {'objects_api_group': [ErrorDetail(string='Ongeldige pk "3" - object bestaat niet.', code='does_not_exist')]}}]}
```

### Stappen om te reproduceren / Steps to reproduce

- Export a form with a registration backend with the objects api
- Import the form in an environment where the PK of the object_api_group does not correspond to that in the export

### Verwacht gedrag / Expected behavior

I would expect the form to use the slug to reconcile the objects api group instead of the PK?

### Screen resolution

None

### Device

None

### OS

None

joeribekker:
Refinement: Agree with silvia to use the natural key, being the slug.

PR #5474, opened 2025-07-29
[#4879] worldline payment redirect flow

### LINKED ISSUE #4879: Migratie van Ogone-classic naar Ingenico/Worldline
Email d.d. 2 december 2024:

> Worldline heeft sinds enige tijd een nieuwere API conform de nieuwste hedendaagse normen en technieken: de Direct-API.
> Deze vervanger van de Ogone POST-methode dient vóór eind van 2025 ingebouwd te zijn, zodat we onze gezamenlijke klanten kunnen blijven bedienen.

SonnyBA:
Partially closes #4879 

**Changes**

Adds the parts required for the redirection flow to make payments.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

Requires https://github.com/open-formulieren/open-forms-sdk/pull/852 to function properly.

[skip: e2e]

### DIFF
--- a/src/openforms/payments/contrib/worldline/plugin.py
+++ b/src/openforms/payments/contrib/worldline/plugin.py
@@ -138,6 +138,7 @@ class WorldlinePaymentPlugin(BasePlugin[PaymentOptions]):
         assert valid_options, "Incorrect payment options encountered"
 
         payment.plugin_options = option_serializer.data
+        payment.provider_payment_id = checkout_response.merchant_reference or ""
         payment.save(update_fields=("plugin_options",))
 
         return PaymentInfo(
@@ -238,9 +239,16 @@ class WorldlinePaymentPlugin(BasePlugin[PaymentOptions]):
                 )
 
             external_payment_id = (
-                payment_data.payment.id
-                if payment_data and payment_data.payment and payment_data.payment.id
-                else ""
+                payment_data.payment.payment_output.references.merchant_reference
+                if payment_data
+                and payment_data.payment
+                and payment_data.payment.payment_output
+                and payment_data.payment.payment_output.references
+                else None
+            )
+
+            assert external_payment_id, (
+                "No merchant reference found in checkout status response"
             )
 
             self.apply_status(payment, status, external_payment_id)


src/openforms/payments/contrib/worldline/tests/test_plugin.py
```diff
@@ -0,0 +1,435 @@
...
+
+        self.assertEqual(data["type"], "get")
+        self.assertEqual(data["data"], {})
+        self.assertTrue(data["url"].startswith(self.merchant.endpoint))
+
+        payment = submission.payments.get()
+        self.assertEqual(payment.plugin_id, "worldline")
+        self.assertEqual(payment.status, PaymentStatus.started)
+
+        # parse the payment URL from within the initial Worldline page
```
sergei-maertens, line 97: sounds like a `with self.subTest("worldline payment page"):` to make the test blocks more clear

src/openforms/payments/contrib/worldline/tests/test_plugin.py
```diff
@@ -0,0 +1,435 @@
...
+            "",
+            "12345",
+        )
+
+        self.assertEqual(payment.status, "")
+
+    def test_config_check(self):
+        configuration_entries = WorldlinePaymentPlugin.iter_config_checks()  # pyright: ignore[reportAttributeAccessIssue]
+
+        self.assertTrue(all(entry.status for entry in configuration_entries))
```
sergei-maertens, line 416: I'm not seeing any merchant setup in this test, which means that if one of these tests fails, I need to context switch and check if something happens in `def setUpTestData`. That's undesired - make sure you control the exact state of the system in the test itself and avoid using `setUp` and `setUpTestData`. You should only put stuff that's a dependency in those methods, like creating a superuser for admin tests.

Additionally, `assertTrue` on a collection as a whole is not informative if there's a test failure. Instead, considering looping over each item and putting `self.subTest` for each iteration, and then make the assertion on the item itself so that when there is a failure, it's immediately clear which case failed.

Here you also have an opportunity to combine the correct/incorrect tests into a single test with two sub tests instead, and that's a more logical unit of tests.

src/openforms/payments/contrib/worldline/plugin.py
sergei-maertens, line 1: Still need to review this, and the frontend code -> tomorrow!

sergei-maertens (changes requested):

docs/configuration/payment/worldline.rst
```diff
@@ -0,0 +1,44 @@
...
+=========
+Worldline
+=========
+
+Open Forms supports the **Worldline** payment backend (using a ``PSPID``).
+
+In order to make use of this module, administrators must create an *Worldline merchant* in
+the admin interface.
+
+1. Navigate to **Configuration** > **Configuration Overview**. In the **Payment Provider Plugin** group, click on **Configuration** for the **Worldline: Test merchant** line.
```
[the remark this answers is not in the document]
SonnyBA: image
The above is shown with atleast 1 merchant configured. Whenever no merchants are configured the entry is not there.

src/openforms/payments/contrib/worldline/tests/test_plugin.py
```diff
@@ -0,0 +1,435 @@
...
+        kwargs = super()._get_vcr_kwargs(**kwargs)
+        kwargs["filter_headers"] = ["Authorization"]
+        return kwargs
+
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        cls.merchant = WorldlineMerchantFactory.create(
+            pspid=os.getenv("WORLDLINE_PSPID", "maykinmedia"),
```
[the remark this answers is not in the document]
SonnyBA: [This](https://github.com/open-formulieren/open-forms/blob/82fbbb882ccfe3b3c4d183fef517a80e40e6ce93/src/openforms/payments/contrib/worldline/tests/test_plugin.py#L35) seems to work (also with recreating the cassettes)

src/openforms/payments/contrib/worldline/tests/test_plugin.py
```diff
@@ -0,0 +1,435 @@
...
+        # we need an arbitrary request
+        request = factory.get("/foo")
+
+        # start url
+        url = plugin.get_start_url(request, submission)
+
+        # good
+        response = self.app.post(
+            url,
+            json.dumps({"merchant": self.merchant.id}),
```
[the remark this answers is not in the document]
SonnyBA: I'm not sure if I understand the whole comment correctly but the `merchant` being in the test request payload was not used at all so I removed that from the tests.

src/openforms/payments/contrib/worldline/admin.py
```diff
@@ -0,0 +1,20 @@
...
+@admin.register(WorldlineMerchant)
+class WorldlneMerchantAdmin(admin.ModelAdmin):
+    fields = (
+        "label",
+        "pspid",
+        "api_key",
+        "api_secret",
+        "endpoint",
+    )
+    readonly_fields = ("endpoint",)
```
sergei-maertens: should this still be readonly now that it's a dropdown?

src/openforms/payments/contrib/worldline/models.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def get_merchant_client(self) -> IMerchantClient:
+        configuration = CommunicatorConfiguration(
+            api_endpoint=self.endpoint,
+            api_key_id=self.api_key,
+            secret_api_key=self.api_secret,
+            authorization_type="v1HMAC",
+            integrator="openforms",  #  TODO: is this the correct value?
+            connect_timeout=5000,
+            socket_timeout=10000,
```
sergei-maertens, line 53: we have `DEFAULT_TIMEOUT_REQUESTS` in the settings which you can customize with an envvar, it'd probably be nice to use that instead of some hardcoded magic numbers.

It's applied here: https://github.com/open-formulieren/open-forms/blob/cfe21393ebd22e3cd30673ea2c1e6fcd8d2b0afe/src/openforms/setup.py#L134C47-L134C71

So perhaps if these are not required options, we can just rely on that implicitly?

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+                },
+            )
+            return HttpResponseRedirect(redirect_url)
+
+    # TODO
+    def handle_webhook(self, request: Request) -> SubmissionPayment:
+        raise NotImplementedError()
+
+    @classmethod
+    def iter_config_checks(cls) -> Generator[Entry, Any, Any]:
```
sergei-maertens, line 227:
```suggestion
    def iter_config_checks(cls) -> Iterator[Entry]:
```
usually this is the more ergonomic variant, see https://docs.python.org/3/library/typing.html#annotating-generators-and-coroutines

> Simple generators that only ever yield values can also be annotated as having a return type of either [Iterable[YieldType]](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) or [Iterator[YieldType]](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterator):

Finally, if you *must* use `Any`, consider using `object` instead as bottom type, as it doesn't leak into all dependent types, which `Any` does.

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+    Order,
+    PaymentOptions,
+)
+from .utils import get_payment_status
+
+logger = structlog.stdlib.get_logger(__name__)
+
+
+class WorldlineOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
+    merchant = PrimaryKeyRelatedAsChoicesField(
```
sergei-maertens, line 51: since this is new code, can we make it export/setup-configuration friendly by not relying on PK as identifier, but instead a unique `identifier` slug field? See also the ticket Anna is working on: #5384 - it's the same root problem.

sergei-maertens: @annashamray FYI - make sure to talk to Sonny if/when you make a new generic API serializer field or modify the existing one.

Probably it's a good idea to deprecate `PrimaryKeyRelatedAsChoicesField` in favour of the alternative.

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+    HttpResponse,
+    HttpResponseBadRequest,
+    HttpResponseRedirect,
+    HttpResponseServerError,
+)
+from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
+
+import structlog
+from onlinepayments.sdk.api_exception import ApiException
```
sergei-maertens, line 15:
```suggestion
from onlinepayments.sdk.api_exception import ApiException as WorldlineApiException
```
the name almost collides with `rest_framework.exceptions.APIException` and it confused me, so I think we should make the name more clear

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+            raise APIException(
+                detail="Failed to retrieve redirect URL from payment provider"
+            ) from e
+
+        plugin_options = payment.plugin_options
+        payment.plugin_options = {
+            **plugin_options,
+            "returnmac": checkout_response.returnmac,
+            "checkout_id": checkout_response.hosted_checkout_id,
+        }
```
sergei-maertens, line 116: I think I would namespace this a bit more to make the private/runtime nature more obvious:


```suggestion
        assert "_checkoutDetails" not in payment.plugin_options
        payment.plugin_options["_checkoutDetails"] = {
            "returnmac": checkout_response.returnmac,
            "checkout_id": checkout_response.hosted_checkout_id,
        }
```

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+
+        payment.save(update_fields=("plugin_options",))
+
+        return PaymentInfo(
+            type=PaymentRequestType.get,
+            url=checkout_response.redirect_url or "",  # valid for three hours
```
sergei-maertens, line 122: an empty string as redirect URL doesn't seem very useful to me, what's the use case here?

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+            data={},
+        )
+
+    def apply_status(
+        self,
+        payment: SubmissionPayment,
+        worldline_status: str,
+        checkout_status: str,
```
sergei-maertens, line 130: shouldn't this be annotated with the constants/enums?

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+        payment_id: str,
+    ) -> None:
+        if payment.status in PAYMENT_STATUS_FINAL:
+            # shouldn't happen or race-condition
+            return
```
sergei-maertens, line 135: can we do a `logger.warning` here?

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+
+        status = get_payment_status(worldline_status, checkout_status=checkout_status)
+
+        # run this query as atomic update()
+        qs = SubmissionPayment.objects.filter(id=payment.id)
+        qs = qs.exclude(status__in=PAYMENT_STATUS_FINAL)
+        qs = qs.exclude(status=status)
+        res = qs.update(status=status, provider_payment_id=payment_id)
```
sergei-maertens, line 143:
```suggestion
        qs = (
            SubmissionPayment.objects
            .filter(id=payment.id)
            .exclude(status__in=PAYMENT_STATUS_FINAL)
            .exclude(status=status)
           )
        res = qs.update(status=status, provider_payment_id=payment_id)
```

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+        options: PaymentOptions,
+    ) -> HttpResponse:
+        with structlog.contextvars.bound_contextvars(
+            submission_uuid=str(payment.submission.uuid),
+            payment_uuid=str(payment.uuid),
+            plugin=self,
+            entrypoint="browser",
+        ):
+            log = logger.bind()
+            log.info("process_payment_status")
```
sergei-maertens, line 161:
```suggestion
            logger.info("process_payment_status")
```
the `bind` doesn't do anything

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+
+            returnmac = payment.plugin_options.get("returnmac", "")
+            checkout_id = payment.plugin_options.get("checkout_id", "")
```
sergei-maertens, line 164: are these properties not present in the `options` argument from `handle_return`? The idea with the plugins is that you never have to deal with the raw JSON primitives, but always get back the validated/processed serializer data from the primitives, so that you don't need to access `payment.plugin_options`.

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+
+            if (
+                not request.query_params.get("RETURNMAC") == returnmac
```
sergei-maertens, line 167: I'd consider using `django.utils.crypto.constant_time_compare` here to guard against timing attacks

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+                    kwargs={"uuid": payment.submission.uuid, "token": token},
+                )
+            )
+
+            client = options["merchant"].get_checkout_client()
+
+            try:
+                response = client.get_hosted_checkout(checkout_id)
+            except ApiException:
+                return HttpResponseServerError(b"Unable to retrieve checkout status")
```
sergei-maertens, line 185: we definitely want to `logger.warning` this :)

src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -0,0 +1,255 @@
...
+
+            try:
+                self.apply_status(
+                    payment,
+                    status,
+                    checkout_status,
+                    payment_provider_id,
+                )
+            except KeyError:
+                return HttpResponseServerError(b"Unable to retrieve payment status")
```
sergei-maertens, line 209: I think the `KeyError` is a leaky abstraction and it wasn't obvious were it's coming from. Perhaps a custom exception would make this more obvious:

```py
class StatusMappingError(Exception):
    pass
```

docs/configuration/payment/worldline.rst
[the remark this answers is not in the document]
SonnyBA: Now shows the following whenever no merchants are configured:
image

src/openforms/payments/contrib/worldline/plugin.py
SonnyBA: This method (currently) gets the plain statuses retrieved with the SDK, they are [converted](https://github.com/open-formulieren/open-forms/blob/0c4446d8e022d29949c2fab263fc048ebd2b9215/src/openforms/payments/contrib/worldline/plugin.py#L146) though inside the method itself.

src/openforms/payments/contrib/worldline/plugin.py
SonnyBA: Changed this through a9d30e3e7e28c4a12c890ba919cca7d3c7fe9b02

src/openforms/payments/contrib/worldline/plugin.py
SonnyBA: This was also updated in https://github.com/open-formulieren/open-forms/commit/890e7b97eefdd0cd478114ab6613460587f1f5ca (see line 221-224).

sergei-maertens (approved):
Let's get this merged, I can definitely nitpick more but that's not very productive 😆 Looks good overall!

PR #5482, opened 2025-07-30
[#5481] Fix old unmatched partners variables

### LINKED ISSUE #5481: Old/unmatched partners variables can cause `KeyError`

3.2

### Customer reference

_No response_

In order to set up the partners component prefill plugin, we need the `partners` component and a user defined variable pointing to that component key. If somehow the component has been removed the user defined variable stays orphan and can cause KeyError.

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #5481

**Changes**

- The user defined variables in the partners validation were not based on the specific form (submission)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Release management

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -712,6 +712,7 @@ class PartnerListField(serializers.Field):
             source=FormVariableSources.user_defined,
             prefill_plugin=FM_PLUGIN_IDENTIFIER,
             prefill_options__mutable_data_form_variable=component_key,
+            form=submission.form,
         ).first()
         if fm_immutable_variable:
             # we do not receive these fields from the frontend (since they are not used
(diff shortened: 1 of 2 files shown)


src/openforms/formio/components/custom.py
```diff
@@ -712,6 +712,7 @@ def validate_list(self, partners):
             source=FormVariableSources.user_defined,
             prefill_plugin=FM_PLUGIN_IDENTIFIER,
             prefill_options__mutable_data_form_variable=component_key,
+            form=submission.form.id,
```
sergei-maertens, line 715:
```suggestion
            form=submission.form,
```
it will already be loaded, so there's no performance gain here, and even if you want to optimize for performance, then you'd use `submission.form_id` because the `.id` makes django do a query for the form instance in this form.

sergei-maertens (approved):

PR #5509, opened 2025-08-05
Convert some heavily used table PK fields to bigint

sergei-maertens:
Note that these naive Django migrations take exclusive locks on the involved tables. The impact on operations is being researched.

Given some of the observed workloads, there is a risk that the submission value variable PK will run out of 32 bit integers in a couple of months to years, so it's better to be prepared now.

[skip: e2e]


viktorvanwijk (commented):
Did you include authentication, payments, and objects_api because they are related to the submission, or are they also 'at risk' in general?

sergei-maertens:
> Did you include authentication, payments, and objects_api because they are related to the submission, or are they also 'at risk' in general?

* authentication is included because it's reasonable to assume "one submission, one auth info instance"
* payments included for a similar reason, the `SubmissionPayment` model has a foreign key to submission so zero or more instances may be related to one submission
* objects_api is included because for every submissions registered with the objects API, there's some additional registration data saved in the database

viktorvanwijk:
Alright, that's seems nice :). Probably good to run it on the DH env anyway, right?

Issue #3999, opened 2024-03-13
Als gemeente wil ik metrics van alle containers ontvangen via open telemetry


joeribekker:
**Reference:** [PF-126](https://dimpact.atlassian.net/browse/PF-126)

### Thema / Theme

Other

### Omschrijving / Description

Ref. Utrecht 182

Om goed inzicht te krijgen de beschikbaarheid en snelheid van services moeten deze metrics aanbieden middels een exporter. Prometheus is een optie, maar mooier is OTLP exporter. (open telemetry, daar kan sentry ook mee omgaan)

zouden jullie dat kunnen toevoegen voor Open Forms om mee te beginnen en  
Nginx, Redis, RabbitMQ.

Hieronder linkjes ter inspiratie.  
https://opentelemetry.io/docs/languages/python/getting-started/  
https://prometheus.io/docs/instrumenting/exporters/

**Metrics to expose**

- [x] Generic HTTP metrics
- [x] Generic "active requests"
- [x] Auth: number of users, logins, logouts, failed logins, axes lockouts
- [x] Submissions: starts, completed, paused, steps submitted
- [x] Submissions: total amount in database by form / lifecycle stage
- [ ] Attachments: (histogram) of file sizes + amount
- [ ] Forms: total amount (not in trash), amount active
- [ ] Formio components: how often is each component used
- [ ] Payments: amount & status
- [ ] Plugins used: authentication, prefill, registrations, validations, appointments, dmn

joeribekker:
Refinement: This needs to be investigated to make this work for Open Forms, but we can make 2 tasks: 1) the service container observability and 2) the application

For (1) we know Utrecht already uses infra-stuff for this so also check if we need to do this.

joeribekker:
Refinement met Utrecht, begin bij de basics: https://wikitech.wikimedia.org/wiki/SLO/template_instructions#Organizational

joeribekker:
Pauline wil graag een endpoint waar metrics op worden gepublished, hoeveel 500 errors per X tijd, en latency.

sergei-maertens:
IMO those kind of things are best solved via the service mesh solution in kubernetes, for example with Istio: https://istio.io/latest/docs/ops/integrations/prometheus/

sdegroot:
@sergei-maertens your application still at least has to expose the metrics (or push them). The application also decides which metrics are useful.

joeribekker:
@sdegroot We're tackling this topic as part of #4966 

[...]

PR #5529, opened 2025-08-14
Tweak otel collector config

sergei-maertens:
Part of #3999 

[skip: e2e]

**Changes**

Iterating based on a live deployment.

* The labels for client/target were missing, these are now there
* De-duplication of global metrics can be performed at the prometheus level, unburdening dashboard builders
* Added some more sample configuration for real-ish deployments

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -352,7 +352,7 @@ maykin-2fa==1.0.0
     # via
     #   -r requirements/base.in
     #   maykin-common
-maykin-common==0.7.0
+maykin-common==0.8.0
     # via -r requirements/base.in
 maykin-json-logic-py==0.13.0
     # via -r requirements/base.in
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -613,7 +613,7 @@ maykin-2fa==1.0.0
     #   -c requirements/base.txt
     #   -r requirements/base.txt
     #   maykin-common
-maykin-common==0.7.0
+maykin-common==0.8.0
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -678,7 +678,7 @@ maykin-2fa==1.0.0
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
-maykin-common==0.7.0
+maykin-common==0.8.0
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
--- a/requirements/extensions.txt
+++ b/requirements/extensions.txt
@@ -574,7 +574,7 @@ maykin-2fa==1.0.0
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
-maykin-common==0.7.0
+maykin-common==0.8.0
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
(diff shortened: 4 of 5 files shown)


docker-compose.yml
```diff
@@ -53,12 +53,13 @@ x-app-env: &x-app-env
...
   # to be up and running.
   - OTEL_SDK_DISABLED=${OTEL_SDK_DISABLED:-true}
+  - OTEL_RESOURCE_ATTRIBUTES=of.client=of,of.target=docker-compose
   - OTEL_METRIC_EXPORT_INTERVAL=${OTEL_METRIC_EXPORT_INTERVAL:-60000}
   - OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_EXPORTER_OTLP_ENDPOINT:-http://otel-collector:4317}  # gRPC
   # otel:supersecret, escape spaces and = with percent encoding
   - OTEL_EXPORTER_OTLP_HEADERS=Authorization=Basic b3RlbDpzdXBlcnNlY3JldA==
   - OTEL_EXPORTER_OTLP_METRICS_INSECURE=${OTEL_EXPORTER_OTLP_METRICS_INSECURE:-true}
-  - OF_OTEL_ENABLE_CONTAINER_RESOURCE_DETECTOR=true
+  - _OTEL_ENABLE_CONTAINER_RESOURCE_DETECTOR=true
```
annashamray, line 62: Could you also rename it in the documentation?

docker/observability/otel/otel-collector-config.yml
```diff
@@ -25,6 +25,12 @@ processors:
   memory_limiter:
     check_interval: 1s
     limit_mib: 1024
+  transform:
+    metric_statements:
+      - context: datapoint
+        statements:
+        - set(attributes["client"], resource.attributes["of.client"])
+        - set(attributes["target"], resource.attributes["of.target"])
```
annashamray, line 33: Could you please explain why this is needed? I though "client" and "target" attributes are already set during deployment? Doesn't prometheus see it the same way as loki?

docker/observability/otel/otel-collector-config.yml
```diff
@@ -35,7 +41,9 @@ exporters:
     endpoint: 0.0.0.0:8889
     namespace: otel  # this becomes a prefix in the prometheus metrics: `otel_*`
     const_labels:
-      environment: docker
+      environment: local_otel_collector
+    resource_to_telemetry_conversion:
```
annashamray, line 45: isn't it duplicating of what you do in "metric_statements"? Does it mean that the metric datapoints would have both "of.client" and "client" labels?

sergei-maertens: these are fixed labels that are added to the metrics. they can help point back to the origin, because prometheus scrapes metrics from many sources:

* the node-exporter that runs on servers, for each server
* the container metrics (`cAdvisor`) for every container on every server
* itself
* the open telemetry collector running on the same host as prometheus (localhost:8889), which exposes the metrics from all open forms services/instances and in the future also other projects like OIP

sergei-maertens: > isn't it duplicating of what you do in "metric_statements"? Does it mean that the metric datapoints would have both "of.client" and "client" labels?

misread the question - this `resource_to_telemetry_conversion` is necessary to get the labels in the first place! The prometheus config then needs some tweaking to drop irrelevant labels. In this sample set up it will indeed have both the `client` and `of_client` labels, and the same for `target`.

docker/observability/otel/otel-collector-config.yml
annashamray: Thank you for the explanation!

annashamray (approved):

PR #5535, opened 2025-08-20
Set up mechanism to serve static files via nginx instead of uwsgi

sergei-maertens:
Sparked by a production incident before my vacation.

Copying the static files to a volume shared with the nginx container makes it possible for nginx to serve the static files directly.

The mechanism is opt-in so that we can keep things backwards compatible. Possibly in 4.0 we remove the uwsgi --static-map option as it could mask potential issues, but that's just a thought at this time.

The recursive copy should be sufficiently safe even with race conditions between replicas:

* we use the manifest static file storage, which appends a unique hash to each static file
* the replicas deploy the same container version and the same set of static files is included in each, so even if one overwrites the other the end-result will be the same
* we don't perform file removal at all, so old versions of static files are still present and won't cause issues if they're still being downloaded while an upgrade is taking place

So for now, any complexity with temporary directories and atomic copies is deferred until it turns out to actually be needed.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

[skip: e2e]

### DIFF
--- a/bin/docker_start.sh
+++ b/bin/docker_start.sh
@@ -14,6 +14,11 @@ mountpoint=${SUBPATH:-/}
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
 
+# Copy static root to volume, if required
+if [ -n "$STATIC_ROOT_VOLUME" ]; then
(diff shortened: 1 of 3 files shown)


sergei-maertens:
Discussed in the office with Joeri & Alex and we proceed with this

PR #5536, opened 2025-08-20
Add changelog entries for latest patch releases

SonnyBA:
Adds the changelog entries for the latest patch releases

[skip: e2e]

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -56,10 +56,7 @@ Detailed changes
 3.2.1 (2025-08-19)
 ==================
 
-Upgrade procedure
------------------
-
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -52,6 +52,64 @@ Detailed changes
...
 ...
 
+
+3.2.1 (2025-08-19)
+==================
+
+Upgrade procedure
+-----------------
+
+**💰 Ingenico Ogone migration**
```
SonnyBA, line 62: Not sure if we want to include this note here (as it is also mentioned in the 3.1.6 release)

PR #5541, opened 2025-08-21
Using UUID as identifier for attribute groups

### LINKED ISSUE #5515: Simplify Yivi attributes group
When configuring the Yivi attributes group in form authentication options, the attributes group name is used as identifier. This should be changed for the attributes group `identifier`, to make this configuration more solid

Task:
- [ ] Update the Yivi authentication backend options, to use id as value for `additionalAttributesGroups`
- [ ] Add data migration, to migrate existing Yivi authentication backends

robinmolen:
Closes #5515

**Changes**

* Introducing a new UUID-field for AttributeGroup objects
* Migrating existing FormAuthenticationBackend's to swap the AttributeGroup `name` configuration with the new `uuid` value
* Updating Yivi authentication backend configuration in Form-builder
* Added new api endpoint for fetching AttributeGroups data (used by the Yivi authentication backend configuration in Form-builder)
* Adding import and export functionalities to the AttributeGroup using django-import-export

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

[skip: e2e]


src/openforms/authentication/api/serializers.py
```diff
@@ -30,6 +30,22 @@ class AuthPluginSerializer(PluginBaseSerializer):
     )
 
 
+class AttributeGroupSerializer(serializers.Serializer):
```
sergei-maertens, line 33: this does not belong in the `openforms.authentication` namespace, but in `openforms.authentication.contrib.yivi`. Currently this is violating the self-containment idea - we organize code in django apps/packages to maintain a predictable structure. See https://open-forms.readthedocs.io/en/stable/developers/backend/index.html#general-principles: "Keep Django apps contained".

The pattern is similar to plugin-specific endpoints in registrations: https://github.com/open-formulieren/open-forms/blob/b04a5c3d8e7178e318700f12b094ae6a61fa7b6d/src/openforms/registrations/api/urls.py#L14

Additionally, for the actual code - why can't you use `ModelSerializer`?

src/openforms/authentication/api/urls.py
```diff
@@ -1,9 +1,14 @@
...
-from .views import PluginListView, SubmissionLogoutView
+from .views import AttributeGroupListView, PluginListView, SubmissionLogoutView
 
 urlpatterns = [
     path("plugins", PluginListView.as_view(), name="authentication-plugin-list"),
+    path(
+        "plugins/yivi/attribute-groups",
+        AttributeGroupListView.as_view(),
+        name="yivi-authentication-plugin-attribute-group-list",
+    ),
```
sergei-maertens, line 11: please break this out into a separate block of code to clearly separate generic vs. plugin-specific endpoints

src/openforms/authentication/api/views.py
```diff
@@ -39,6 +40,25 @@ def get_objects(self):
         return list(register.iter_enabled_plugins())
 
 
+@extend_schema_view(
+    get=extend_schema(summary=_("List available Yivi attribute groups")),
+)
+class AttributeGroupListView(ListMixin, APIView):
```
sergei-maertens, line 46: this looks like it could/should be a plain `rest_framework.generics.ListApiView` instead of using our own `ListMixin` alternative. The latter is inspired by the former do deal with non-database data (this is described in the `ListMixin` docstring 😉 )

sergei-maertens, line 46: Additionally, this views belongs in the yivi app, not the generic app that's plugin-agnostic.

src/openforms/authentication/contrib/yivi_oidc/migrations/0004_attributegroup_uuid.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ("yivi_oidc", "0003_delete_yiviopenidconnectconfig"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="attributegroup",
+            name="uuid",
+            field=models.UUIDField(
+                default=uuid.uuid4, unique=True, verbose_name="UUID"
```
sergei-maertens, line 18: I think this `default` will/may fail, because it's calculated only once when the migration executes, so existing records will end up with duplicates, which violate the unique constraint.

typically in these kind of migrations you need to:

* add the field, make it nullable, don't specify the unique constraint
* add a `RunPython` operation to backfill with unique UUIDs
* modify the field: remove `null=True` and add the unique constraint

Try this in your environment with more than one attribute group :)

src/openforms/authentication/contrib/yivi_oidc/migrations/0005_data_migrate_set_attributegroup_uuid.py
```diff
@@ -0,0 +1,59 @@
...
+from django.db.models import Q
+
+
+def transform_set_attributegroup_uuid(apps: StateApps, _):
+    AttributeGroup = apps.get_model("yivi_oidc", "Attributegroup")
+
+    # Get a set of uuid that already exist
+    occupied_uuids: set = set(
+        AttributeGroup.objects.filter(~Q(uuid=None)).values_list("uuid", flat=True)
+    )
```
sergei-maertens, line 15: there cannot be any if the field is being added in the previous migration

src/openforms/authentication/contrib/yivi_oidc/migrations/0005_data_migrate_set_attributegroup_uuid.py
```diff
@@ -0,0 +1,59 @@
...
+    for attribute_group in attribute_groups:
+        attribute_group_uuid = uuid.uuid4()
+
+        # Check if the uuid is unique, if not generate a new one and check again
+        while attribute_group_uuid in occupied_uuids:
+            attribute_group_uuid = uuid.uuid4()
+
+        # Add the new uuid to the `occupied` set and set the attribute group uuid
+        occupied_uuids.add(attribute_group_uuid)
+        attribute_group.uuid = attribute_group_uuid
```
sergei-maertens, line 28: can we simplified entirely to:

```py
for group in AttributeGroup.objects.all():
    group.uuid = uuid4()
    group.save()
```

* no need to bother with checking whether it exists or not, see my comments on the previous migration
* no need to `bulk_update`, we expect data sets < 100 records so the migration will be fast, even with a naive (but very readable) loop + save

src/openforms/authentication/contrib/yivi_oidc/migrations/0005_data_migrate_set_attributegroup_uuid.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Update the attribute groups
+    AttributeGroup.objects.bulk_update(attribute_groups, fields=["uuid"])
+
+
+def reverse_set_attributegroup_uuid(apps: StateApps, _):
```
sergei-maertens, line 34: no point implementing this, you can just use `migrations.RunPython.noop` instead.

When you reverse migrations, the reverse from the previous migration (adding a field) is to drop the UUID field again, so the data will be gone anyway.

src/openforms/authentication/contrib/yivi_oidc/tests/oidc_plugins/test_plugins.py
```diff
@@ -44,14 +44,15 @@ def test_yivi_process_claims_with_dots_in_path(self):
         plugin = oidc_register[oidc_client.identifier]
         AttributeGroupFactory.create(
             name="know_attributes",
+            uuid="00000000-0000-0000-0000-12345678abcd",
```
sergei-maertens, line 47: Use https://www.uuidgenerator.net/ or a plain python shell to generate actually valid uuid4's

These examples aren't.

src/openforms/authentication/contrib/yivi_oidc/tests/oidc_plugins/test_plugins.py
```diff
@@ -127,7 +134,9 @@ def test_extract_additional_claims_with_unknown_attributes(self):
         request = self._setup_form(
             options={
                 "authentication_options": [],
-                "additional_attributes_groups": ["unknow_attributes"],
+                "additional_attributes_groups": [
+                    "00000000-0000-0000-0000-12345678abcd"
```
sergei-maertens, line 138: a comment saying that this one is unknown/non-existent would be nice for the person that has to debug failures in this test

src/openforms/authentication/contrib/yivi_oidc/tests/test_admin.py
```diff
@@ -0,0 +1,268 @@
...
+
+        return json_option[0]
+
+    def test_successful_export_attribute_groups(self):
+        user = UserFactory.create(
+            is_staff=True,
+        )
+
+        # Create test data
+        AttributeGroupFactory(
```
sergei-maertens, line 31: please refrain from using bare factory instantation, always explicitly ues `Factory.create` or `Factory.build` so that it's obvious if we're hitting the DB or not.

src/openforms/authentication/contrib/yivi_oidc/tests/test_admin.py
```diff
@@ -0,0 +1,268 @@
...
+        )
+
+        # Go to export page
+        export_page = self.app.get(self.admin_export_url, user=user)
+        form = export_page.forms["export-attributegroups"]
+
+        # Select json file format, and export
+        form["file_format"] = self._get_value_for_select_field(
+            form["file_format"], "json"
+        )
```
sergei-maertens, line 51: no need to complicate things so much, you can use the built-in webtest form handling:

https://github.com/modelbrouwers/modelbrouwers/blob/6cb8a2fef8df311c690fe0df8a1e2de75cd241d2/src/brouwers/shop/tests/admin/test_products_admin.py#L29

src/openforms/authentication/contrib/yivi_oidc/tests/test_admin.py
```diff
@@ -0,0 +1,268 @@
...
+        response = form.submit()
+
+        # Expect that the export was successful
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content_type, "application/json")
+        self.assertGreater(int(response["Content-Length"]), 0)
+
+        # Expect that the export contains all attribute groups
+        decoded_response = json.loads(response.body.decode())
```
sergei-maertens, line 60:
```suggestion
        decoded_response = response.json()
```
(django-)webtest has this covered :)

src/openforms/authentication/contrib/yivi_oidc/tests/test_admin.py
```diff
@@ -0,0 +1,268 @@
...
+            self.assertEqual(group1.attributes, ["foo", "bar", "foo.bar"])
+
+            group2 = AttributeGroup.objects.get(
+                uuid="12345678-abcd-1234-abcd-1234567890cd"
+            )
+            self.assertEqual(group2.name, "group2")
+            self.assertEqual(group2.description, "For retrieving group 2 data")
+            self.assertEqual(group2.attributes, ["zoo", "zar"])
+
+    def test_successful_update_attribute_group_using_import(self):
```
sergei-maertens, line 171: unless there's a lot of custom code going on in the import-export code, these tests aren't really necessary since they're behaviour should be covered by the django-import-export testsuite itself.

src/openforms/authentication/contrib/yivi_oidc/tests/test_auth_flow_init.py
```diff
@@ -86,14 +86,29 @@ def test_signicat_condiscon_contains_only_the_chosen_authentication_and_addition
             with_yivi=True,
             options__identity_settings__bsn_claim_path=["attribute.bsn"],
         )
-        AttributeGroupFactory(name="personal", attributes=["first_name", "last_name"])
-        AttributeGroupFactory(name="mail", attributes=["email_address"])
-        AttributeGroupFactory(name="phone", attributes=["phone_number"])
+        AttributeGroupFactory(
```
sergei-maertens, line 89:
```suggestion
        AttributeGroupFactory.create(
```

src/openforms/authentication/contrib/yivi_oidc/tests/test_config.py
```diff
@@ -28,14 +28,22 @@ def test_valid_empty_options(self):
...
             data={
                 "authentication_options": [
                     AuthAttribute.bsn,
                     AuthAttribute.kvk,
                 ],
-                "additional_attributes_groups": ["some_group"],
+                "additional_attributes_groups": [
+                    "00000000-0000-0000-0000-12345678abcd",
+                    "11111111-0000-0000-0000-12345678abcd",
+                ],
```
sergei-maertens, line 46: the previous version of this test only added one group, now there are two?

src/openforms/authentication/contrib/yivi_oidc/tests/test_plugin.py
```diff
@@ -92,10 +92,14 @@ def test_transform_claims_without_authentication_or_additional_attributes(self):
         )
 
     def test_transform_claims_with_additional_attributes_claims(self):
-        AttributeGroupFactory(name="foo_attribute", attributes=["attribute_name"])
+        AttributeGroupFactory(
```
sergei-maertens, line 95:
```suggestion
        AttributeGroupFactory.create(
```

src/openforms/authentication/contrib/yivi_oidc/admin.py
```diff
@@ -1,17 +1,46 @@
 from django.contrib import admin
 
+from import_export import resources
+from import_export.admin import ImportExportModelAdmin
+
 from .models import AttributeGroup
 
 
+class AttributeGroupResource(resources.ModelResource):
```
sergei-maertens, line 9: we typically put this in a `resources.py` file

src/openforms/authentication/contrib/yivi_oidc/admin.py
```diff
@@ -1,17 +1,46 @@
...
+    def dehydrate_description(self, obj: AttributeGroup) -> str:
+        # Trimming whitespace
+        return obj.description.strip()
+
+
 @admin.register(AttributeGroup)
-class AttributeGroupAdmin(admin.ModelAdmin):
+class AttributeGroupAdmin(ImportExportModelAdmin):
+    import_template_name = "admin/forms/attributegroup/import_form.html"
+    export_template_name = "admin/forms/attributegroup/export_form.html"
```
sergei-maertens, line 28: why are these template overrides necessary? If it's for the django-admin-index stuff, aren't there some base templates to override instead?

src/openforms/authentication/contrib/yivi_oidc/admin.py
```diff
@@ -1,17 +1,46 @@
...
+
     fields = [
         "name",
+        "uuid",
         "description",
         "attributes",
     ]
+    readonly_fields = ["uuid"]
```
sergei-maertens, line 36:
```suggestion
    readonly_fields = ("uuid",)
```
I prefer tuples at the class/module level and especially in model admins because they're immutable and better convey the intent that this is not supposed to be modified at runtime.

src/openforms/authentication/contrib/yivi_oidc/admin.py
```diff
@@ -1,17 +1,46 @@
...
     list_display = [
         "name",
+        "uuid",
         "description",
         "attributes",
     ]
+    search_fields = ("uuid", "name")
+
+    # Import and export options:
+    resource_classes = [AttributeGroupResource]
```
sergei-maertens, line 46:
```suggestion
    resource_classes = (AttributeGroupResource,)
```

src/openforms/authentication/contrib/yivi_oidc/config.py
```diff
@@ -44,8 +45,12 @@ class YiviOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
         required=False,
         allow_empty=True,
     )
-    additional_attributes_groups = serializers.ListField(
-        child=serializers.ChoiceField(choices=[]),  # Choices are dynamically defined
+    additional_attributes_groups = serializers.SlugRelatedField(
+        slug_field="uuid",
+        queryset=AttributeGroup.objects.annotate(
+            str_uuid=Cast("uuid", output_field=CharField())
+        ).values_list("str_uuid", flat=True),
```
sergei-maertens, line 52: I think if you drop this `values_list`, you don't need the `_handle_before_import` anymore. DRF will properly resolve it to the model instance and do the lookup query.

src/openforms/authentication/tests/factories.py
```diff
@@ -101,6 +101,7 @@ class Meta:
 
 class AttributeGroupFactory(factory.django.DjangoModelFactory):
     name = factory.Sequence(lambda n: f"attribute-group-{n}")
+    uuid = factory.Faker("uuid4")
```
sergei-maertens, line 104: you don't need to specify a factory attribute when the model has `default=...` set - the model default will kick in.

src/openforms/forms/templates/admin/forms/attributegroup/export_form.html
sergei-maertens, line 1: This seems like way too much code to override from the library.

src/openforms/forms/templates/admin/forms/attributegroup/import_form.html
sergei-maertens, line 1: This is also too much code override, will be a nightmare to keep up to date whenever django-import-export or the django admin make changes in future versions.

sergei-maertens (changes requested):
Part 1 review

src/openforms/authentication/contrib/yivi_oidc/migrations/0005_data_migrate_set_attributegroup_uuid.py
robinmolen: Isn't `bulk_create`/`bulk_update` the preferred way to update data in migrations? Because less DB actions => faster & faster == better?
Or is a over-simplification?

src/openforms/authentication/api/urls.py
robinmolen: With your previous comment, i've moved to the plugin specific endpoints further below (like the registrations setup). The actual Yivi endpoint definition is moved to the Yivi plugin

src/openforms/authentication/contrib/yivi_oidc/migrations/0004_attributegroup_uuid.py
robinmolen: Ah, i was hoping to avoid that.. 👍

src/openforms/authentication/contrib/yivi_oidc/tests/test_config.py
robinmolen: Yeah.. multiple items doesn't add much, if anything. Refactoring

src/openforms/authentication/contrib/yivi_oidc/tests/test_admin.py
robinmolen: Check!

src/openforms/authentication/contrib/yivi_oidc/tests/oidc_plugins/test_plugins.py
robinmolen: Yeah good point, adding it

src/openforms/forms/templates/admin/forms/attributegroup/import_form.html
robinmolen: As mentioned above, i've removed the attributegroup custom templates

src/openforms/authentication/contrib/yivi_oidc/config.py
robinmolen: I've tried to simplify the `SlugRelatedField`, but this caused problems with regular imports (the export data is a list of UUID's, and the import process seemed to expect AttributeGroup objects). We could solve this by extending the `_handle_before_import`, but that doesn't sound right..?

Anyway, the `_handle_before_import` now tries to fix old form imports. During import it checks the used values of `additional_attributes_groups` and, if the used value is a known attributeGroup name, tries to correct it to use UUID's

src/openforms/authentication/contrib/yivi_oidc/api/views.py
```diff
@@ -0,0 +1,30 @@
...
+from rest_framework.mixins import ListModelMixin
+
+from ..models import AttributeGroup
+from .serializers import AttributeGroupSerializer
+
+
+@extend_schema_view(
+    get=extend_schema(summary=_("List available Yivi attribute groups")),
+)
+class AttributeGroupListView(ListModelMixin, GenericAPIView):
```
sergei-maertens, line 15:
```suggestion
class AttributeGroupListView(generics.ListAPIView):
```
see https://www.django-rest-framework.org/api-guide/generic-views/#listapiview

src/openforms/authentication/contrib/yivi_oidc/api/views.py
```diff
@@ -0,0 +1,30 @@
+from django.utils.translation import gettext_lazy as _
+
+from drf_spectacular.utils import extend_schema, extend_schema_view
+from rest_framework import authentication, permissions
+from rest_framework.generics import GenericAPIView
+from rest_framework.mixins import ListModelMixin
```
sergei-maertens, line 6:
```suggestion
from rest_framework import generics
```

src/openforms/authentication/contrib/yivi_oidc/api/views.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+    serializer_class = AttributeGroupSerializer
+    queryset = AttributeGroup.objects.all()
+    lookup_field = "uuid"
+
+    def get(self, request, *args, **kwargs):
+        return self.list(request, *args, **kwargs)
```
sergei-maertens, line 30:
```suggestion
```
not necessary, that's what the generic view already does :)

src/openforms/forms/tests/test_import_export.py
```diff
@@ -1684,10 +1696,55 @@ def test_import_form_with_yivi_auth_backend_with_unknown_additional_attributes_g
...
         self.assertEqual(
-            auth_backend.options["additional_attributes_groups"], ["some_known_scope"]
+            auth_backend.options["additional_attributes_groups"],
+            ["a1c9df69-927c-4390-be26-f39daf107691"],
+        )
+
+    def test_import_form_with_yivi_auth_backend_with_unknown_additional_attributes_groups_raises_exception(
+        self,
+    ):
+        AttributeGroupFactory(
```
sergei-maertens, line 1708:
```suggestion
        AttributeGroupFactory.create(
```

sergei-maertens (changes requested):

Issue #5475, opened 2025-07-29
It's impossible to retrieve additional claim values and use them in logic (probably because of dots in the names)


robinmolen:
The problem is that the yivi attributes are saved with their original key; for example "irma-demo.gemeente.personalData.over18". JsonLogic, or at least the Python implementation, cannot target attributes with dots (".") in their key.

PR #5543, opened 2025-08-22
Implement Yivi prefill plugin

### LINKED ISSUE #5419: Prefill plugin for Yivi and eIDAS
### Thema / Theme

Plugins

Add a prefill plugins for the new Yivi and eIDAS authentication backends.

_No response_

_No response_

robinmolen:
Partly closes #5419

[skip: e2e]

~Requires #5515 (wait with merging until #5515 is in master)~

**Changes**

Added a prefill plugin specifically for Yivi authentication. It grants access to the configured Yivi additional attributes and the regular authentication identifier and LoA values.

As the Yivi prefill attributes change depending on the used form configuration, which prefill attributes are available is completely determined in the frontend. By adding a Yivi prefill plugin API endpoint we provide the frontend (i.e. form-builder) with the prefill attributes and conditions when the prefill attributes can be used.

Most of the Yivi prefill attributes should be used in conjunction with specific Yivi attribute groups. Using the `attribute_group_uuid` this relationship is defined. To also provide access to static prefill attributes (the prefill attributes for the regular authentication identifier and LoA values), a `is_static` flag is used. `is_static` is used to convey that the attributes aren't related to any attributegroups and are always available for Yivi prefill.

The changes include:

* The new prefill plugin
* New Yivi prefill API endpoint
* Custom React components for the Yivi prefill config (with a "use the Yivi auth plugin" warning)
* Updated Yivi authentication claims processing (see Important to note for further information)

**Important to note:**

The Yivi prefill plugin only works in conjunction with the Yivi authentication plugin. This is made clear in the Yivi prefill plugin documentation, and in the Yivi prefill configuration in the form-builder.

The way-of-working of the Yivi authentication plugin was also touched-up.
The Yivi attributes use periods (".") in their keys, which messes with glom and jsonLogic. The jsonLogic issue was resolved in https://github.com/open-formulieren/open-forms/issues/5475, by re-formatting the Yivi static variables. To resolve the issue with glom, and to prevent any future issues, the Yivi attributes will now be re-formatting during the processing of the authentication claims.

**Checklist**

- Impact on features

  - [X] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
+++ b/src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
@@ -268,7 +268,12 @@ class YiviPlugin(BaseOIDCPlugin, AnonymousUserOIDCPluginProtocol):
             [
                 {
                     "path_in_claim": [attribute],
-                    "processed_path": ["additional_claims", attribute],
+                    # Jsonlogic, which we use to access the variables in the json logic and components,
+                    # cannot retrieve data from a key that contains periods (aka dots ".").
+                    # To allow these values to be targeted, we replace the period with an underscore.
+                    # In case of nested attributes, this replacing is done recursively.
+                    # https://github.com/open-formulieren/open-forms/issues/5475
+                    "processed_path": ["additional_claims", attribute.replace(".", "_")],
                 }
                 for attribute_group in additional_attributes
                 for attribute in attribute_group
--- a/docs/configuration/prefill/index.rst
+++ b/docs/configuration/prefill/index.rst
@@ -16,3 +16,4 @@ Prefill plugins
    family_members
    eidas
    eidas_company
+   yivi
--- a/docs/configuration/prefill/yivi.rst
+++ b/docs/configuration/prefill/yivi.rst
@@ -1,0 +1,40 @@
+.. _configuration_prefill_yivi:
+
+====
+Yivi
+====
+
+`Yivi`_ is a identity wallet, allowing greater control for users about which personal
+data they share with a website. The information available after Yivi authentication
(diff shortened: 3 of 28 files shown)


src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
```diff
@@ -268,7 +268,12 @@ def get_claim_processing_instructions(
             [
                 {
                     "path_in_claim": [attribute],
-                    "processed_path": ["additional_claims", attribute],
+                    # Jsonlogic, which we use to access the variables in the json logic and components,
+                    # cannot retrieve data from a key that contains periods (aka dots ".").
+                    # To allow these values to be targeted, we replace the period with an underscore.
+                    # In case of nested attributes, this replacing is done recursively.
+                    # https://github.com/open-formulieren/open-forms/issues/5475
+                    "processed_path": ["additional_claims", attribute.replace(".", "_")],
```
robinmolen, line 276: As discussed in the office, if we want to "explode" the attributes as they enter the website, this is the place to do it. (By "exploding the attributes" I mean turning `foo.bar: "hello"` into `foo: {bar: "hello"}`, instead of replacing the period with an underscore)

Exploding the attributes might also be better for form-builders, as accessing the attributes in template tags would still follow the Yivi notation, instead of a "replace periods with underscores" mutation.

When trying to retrieve the attributes using glom, having multi level attributes also simplifies things (this might happen in the "get_attribute_value" function of the Yivi prefill plugin). This way we can just use the Yivi attributes (like `irma-demo.gemeente.personalData.firstnames`) as prefill attributes

PR #5554, opened 2025-08-25
:page_facing_up: Reflect the stakeholders in the license and readme.

joeribekker:
**Changes**

The creation of Open Forms was commissioned by Dimpact. However, Dimpact was the primary point of contact but the original tender was by a collective of municipalities. In addition, after the 1.0 release, additional stakeholders - including Maykin itself - made significant contributions to Open Forms of which the intellectual property was not transfered.

Although this change does not change the license, it does broaden the copyright holders to a group of parties, namely, the Stakeholders. This reflects the principle of a community more accurately, where a group within that community (again, the Stakeholders) is responsible for the stewardship ("broncode beheer" in Dutch) of this codebase.

[skip: e2e]

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 # Licence
 
-Copyright © Dimpact, 2021
+Copyright © "the Stakeholders" (see: STAKEHOLDERS.md), 2025
 
 Licensed under the EUPL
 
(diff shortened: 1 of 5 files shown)


joeribekker:
> This probably requires an update to `publiccode.yaml` too?

Done. I read the specification of publiccode. They suggest to use another attribute (`authorsFile`) if there are multiple copyright owners. This attribute referes to "authors" which I think has a different meaning AND the attribute they refer to is deprecated... So, I removed the attribute since its optional.

PR #5566, opened 2025-09-01
[#4879] add data migration for Worldline migration

### LINKED ISSUE #5503: Add migration script

SonnyBA:
Closes #5503

**Changes**

Adds a data migration for the Ogone Merchants and the OgoneWebhookConfiguration.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/payments/contrib/worldline/tests/test_migrations.py
```diff
@@ -0,0 +1,268 @@
...
+        WorldlineMerchant.objects.create(
+            label="Merchant 2",
+            pspid="merchant-2",
+            api_key="merchant-2-api-key",
+            api_secret="merchant-2-api-secret",
+            endpoint=WorldlineEndpoints.live,
+        )
+
+    def test_duplicate_ogone_merchants(self):
+        executor = MigrationExecutor(connection)
```
sergei-maertens, line 147: this seems too low-level, the whole purpose of `TestMigrations` is that you don't have to manually use these django internals.

This test also has very little added value, because the upgrade check with the check script that you're adding will block migrations from starting in the first place, exactly so that you don't have to worry about duplicates in the actual data migration.

This entire test case can be removed 😃

src/openforms/payments/contrib/worldline/migrations/0007_auto_20250829_1446.py
sergei-maertens, line 1: can you give this migration a meaningful name instead of the auto generated name?

```bash
src/manage.py makemigrations worldline --empty --name migrate_ogone_merchants_to_worldline
```
is usually what I do

src/openforms/payments/contrib/worldline/migrations/0007_auto_20250829_1446.py
```diff
@@ -0,0 +1,101 @@
...
+def migrate_merchants(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    OgoneMerchant = apps.get_model("payments_ogone", "OgoneMerchant")
+    WorldlineMerchant = apps.get_model("payments_worldline", "WorldlineMerchant")
+    known_pspids = WorldlineMerchant.objects.values_list("pspid", flat=True)
+    ogone_merchants = OgoneMerchant.objects.all()
+    new_merchants = {}
+
+    for merchant in ogone_merchants:
+        if merchant.pspid in known_pspids:
+            logger.warning("skipped_merchant_migration", merchant=merchant)
```
sergei-maertens, line 30:
```suggestion
            logger.warning("skipped_merchant_migration", merchant=merchant.pspid)
```
avoid logging rich objects with structlog, as they will eventually be logged using `repr(...)` I believe. You should only pass JSON-serializable primitives as extra kwargs.

sergei-maertens (changes requested):

PR #5577, opened 2025-09-04
Improve expiry messages for expired certificates

sergei-maertens:
The old messages said they would expire soon, which is confusing if they're already expired.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


vaszig (approved):
I guess there is  a specific reason for this PR to have the translations updated (having in mind that we do that during the release preparation).

PR #5595, opened 2025-09-08
[#4879] generate merchant reference

### LINKED ISSUE #5581: Generate merchant reference through OF
Currently the `merchant_reference` gets generated by Worldline whereas for Ogone this (previously called `PAYID`) was generated by Open Forms. Open Forms should apply the previous behavior and generate the `merchant_reference` by itself.

Current situation is this:

Image

Even though for the top entry, the payment reference in OF is still similar to the bottom one, while in the Merchant portal, it gives a UUID. The bottom one is preferred, also in the new situation.

SonnyBA:
Closes #5581 

**Changes**

Generates the `merchantReference` field through Open Forms (instead of having it get generated by Worldline). Also saves the payment ID field in the `provider_payment_id` field so that we can perform requests to the Worldline API with the saved payments. 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/payments/contrib/worldline/plugin.py
+++ b/src/openforms/payments/contrib/worldline/plugin.py
@@ -50,11 +50,9 @@ from .constants import (
 )
 from .models import WorldlineMerchant, WorldlineWebhookConfiguration
 from .typing import (
-    AmountOfMoney,
     CheckoutInput,
     Order,
     PaymentOptions,
-    References,
 )
 from .utils import get_webhook_helper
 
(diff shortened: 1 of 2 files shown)


src/openforms/payments/contrib/worldline/plugin.py
```diff
@@ -246,30 +242,22 @@ def handle_return(
...
                     if payment_data and payment_data.payment
                     else None
                 )
-            except ValueError:
+
+                payment.provider_payment_id = payment_data.payment.id
+                payment.save(update_fields=("provider_payment_id",))
+            else:
                 status = WorldlineHostedCheckoutStatus.to_payment_status(
                     checkout_status
```
SonnyBA, line 257: If we want to adhere to the minimum amount of coverage we could chose to remove these lines (as the status also could be updated at a later point through the webhooks) or we could create a test for this with mocks 🤔

PR #5607, opened 2025-09-10
Reshape Yivi attributes when storing them in the auth_info

### LINKED ISSUE #5419: Prefill plugin for Yivi and eIDAS

sergei-maertens:
Closes #5419 (partly)

**Changes**

Instead of processing the attributes *after* obtaining them from Yivi by replacing <kbd>.</kbd> with <kbd>_</kbd>, we transform the attributes into a nested object where each period indicates a level.

This makes them naturally available in the DTL via `{{ foo.bar }}` and in json logic with `{"var": "foo.bar"}`.

Historical data is not being updated as the Yivi feature released in 3.2 was still quite experimental.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

[skip: e2e]


robinmolen (approved):
Flipping nice! I was afraid that the destructing would be more of a hassle, but that is really clean!

PR #5617, opened 2025-09-12
[#4879] admin migrate form action

### LINKED ISSUE #5579: Add admin form action to migrate forms
This allows administrators to migrate forms using the Ogone payment provider to easily migrate form configurations to the Worldline Payment provider. The previous `COM` attribute should also be migrated whenever the action is performed. The `TITLE` query parameter can be ignored as this is handled different through Worldline.

SonnyBA:
Closes #5579 

**Changes**

Adds a form action which allows administrators to migrate forms with the Ogone payment backend to the Worldine payment backend.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

[skip: e2e]

### DIFF
--- a/src/openforms/forms/admin/views.py
+++ b/src/openforms/forms/admin/views.py
@@ -9,6 +9,7 @@ from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.contrib.messages.views import SuccessMessageMixin
 from django.contrib.postgres.forms import SimpleArrayField
+from django.db import transaction
 from django.http import FileResponse, HttpResponse
 from django.shortcuts import get_object_or_404
 from django.urls import reverse_lazy
@@ -23,6 +24,7 @@ from privates.storages import private_media_storage
 from rest_framework.exceptions import ValidationError
 
 from openforms.logging import logevent
+from openforms.payments.contrib.ogone.models import OgoneMerchant
 
 from ..forms import ExportStatisticsForm
 from ..forms.form import FormImportForm
@@ -166,3 +168,72 @@ class ExportSubmissionStatisticsView(
             }
         )
         return context
+
+
+class PaymentMigrationForm(forms.Form):
+    forms_to_migrate = forms.ModelMultipleChoiceField(
+        queryset=Form.objects.filter(payment_backend="ogone-legacy"),
+        error_messages={
+            "invalid_choice": _(
+                "The selected form(s) cannot be migrated to the Worldline payment backend."
+                " Please verify the selected forms have the ogone-legacy payment backend"
+                " configured."
+            ),
+        },
+        label=_("Forms"),
+    )
+
+
+def _migrate_form(form: Form) -> bool:
+    ogone_merchant_id = form.payment_backend_options["merchant_id"]
+
+    try:
+        ogone_merchant = OgoneMerchant.objects.get(pk=ogone_merchant_id)
+    except OgoneMerchant.DoesNotExist:
+        return False
+
+    form.payment_backend = "worldline"
+    if "com_template" in form.payment_backend_options:
+        form.payment_backend_options = {
+            "merchant": ogone_merchant.pspid,
+            "descriptor_template": form.payment_backend_options["com_template"],
+        }
+    else:
+        form.payment_backend_options = {"merchant": ogone_merchant.pspid}
+
+    form.save()
+
+    return True
+
+
+class PaymentMigrationView(PermissionRequiredMixin, FormView):
+    form_class = PaymentMigrationForm
+    template_name = "admin/forms/form/migrate-payment-backend.html"
+    permission_required = "forms.change_form"
+    success_url = reverse_lazy("admin:forms_form_changelist")
+
+    def form_valid(self, form):
+        unmigrated_forms = []
+
+        with transaction.atomic():
+            for form_to_migrate in form.cleaned_data["forms_to_migrate"]:
+                if not _migrate_form(form_to_migrate):
+                    unmigrated_forms.append(form_to_migrate)
+
+        if unmigrated_forms:
+            message = _(
+                "The following forms reference a merchant that is non-existent: {form_ids}"
+                "Please select a new merchant for these forms."
+            ).format(form_ids=",".join(str(form.name) for form in unmigrated_forms))
+
+            messages.warning(self.request, message)
+
+        success_count = len(form.cleaned_data["forms_to_migrate"]) - len(
+            unmigrated_forms
+        )
+        if success_count > 0:
+            message = _("{succes_count} forms were migrated.").format(
+                succes_count=success_count
+            )
+            messages.success(self.request, message)
+        return super().form_valid(form)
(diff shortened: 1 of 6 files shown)


src/openforms/forms/admin/views.py
```diff
@@ -166,3 +168,92 @@ def form_fields():
...
+                "The following forms reference a merchant that is non-existent: %s"
+                "Please select a new merchant for these forms."
+            ) % ",".join(str(form.name) for form in form.outdated_forms)
+
+            message_level = messages.WARNING
+        else:
+            message = _("%s forms were migrated.") % len(
+                form.cleaned_data["forms_to_migrate"]
+            )
+            message_level = messages.SUCCESS
```
sergei-maertens, line 256: this is weird - the branching logic here implies one or the other happens, but both can happen at the same time. Part of the forms can be outdated, part of the forms can be migrated.

src/openforms/forms/admin/views.py
```diff
@@ -166,3 +168,92 @@ def form_fields():
...
+
+
+class PaymentMigrationView(PermissionRequiredMixin, FormView):
+    form_class = PaymentMigrationForm
+    template_name = "admin/forms/form/migrate-payment-backend.html"
+    permission_required = "forms.change_form"
+    success_url = reverse_lazy("admin:forms_form_changelist")
+
+    def form_valid(self, form):
+        form.save()
```
sergei-maertens, line 243: I think the code here is doing some awkward things. The presence of the `.save` method on the form confused me at first, because it makes it look like a `ModelForm`, which it isn't.

Additionally, you're carying over data from validation to processing via `cleaned_data`, which does not map to fields in the form, again confusing things because that's not what you expect with typical Django form - `cleaned_data` usually just contains the validated form fields.

Finally, you're setting/inventing custom attributes on the form class like `oudated_forms` (initialized with a mutable default, which is quite an anti-pattern as well).

This leads to some gynmnastics to be able to produce the user-facing messages in the view layer.

Personally I'd be lot more strict w/r to the typing and technical approach to avoid mutation side-effects, and instead rely more on pure functions. I think you can even set up this form/view combination to make the form purely validation related ("are the selected forms valid for migration?") and do the actual migration in the view in `form_valid` so you have all the information at hand - it's the only scope where you need it anyway.

I'm conflicted about the amount of feedback on this one-off migration thing, but I do want to address the anti-patterns that I see before they are applied more in the codebase.

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
             "Form successfully imported!",
             messages[0].message,
         )
+
+
+@disable_admin_mfa()
+class TestMigrationForm(WebTest):
+    csrf_checks = False
```
sergei-maertens, line 315: why do CSRF checks need to be disabled?

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
...
+
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        cls.admin_user = StaffUserFactory(user_permissions=["forms.change_form"])
```
sergei-maertens, line 321:
```suggestion
        cls.admin_user = StaffUserFactory.create(user_permissions=["forms.change_form"])
```
*always* use `Factory.build` or `Factory.create` explicitly and don't rely on the implicit create behaviour.

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
...
+
+    def test_not_superuser_cant_access(self):
+        user = UserFactory(is_staff=False, is_superuser=False)
```
sergei-maertens, line 324:
```suggestion
        user = UserFactory.create(is_staff=False, is_superuser=False)
```

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
...
+        )
+
+        form_to_migrate.refresh_from_db()
+
+        self.assertEqual(form_to_migrate.payment_backend, "ogone-legacy")
+        self.assertEqual(
+            form_to_migrate.payment_backend_options, {"merchant_id": ogone_merchant.pk}
+        )
+
+    def test_form_permissions(self):
```
sergei-maertens, line 355: please take care in naming your tests appropriately and be verbose.

This test name only tells me something with permissions is being tested, but it's not clear if permissions should allow or block the operation.


```suggestion
    def test_staff_user_without_change_form_permissions_has_no_access(self):
```

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
...
+        self.assertEqual(response.status_code, 403)
+
+        form_to_migrate.refresh_from_db()
+
+        self.assertEqual(form_to_migrate.payment_backend, "ogone-legacy")
+        self.assertEqual(
+            form_to_migrate.payment_backend_options, {"merchant_id": ogone_merchant.pk}
+        )
+
+    def test_confirmation_page(self):
```
sergei-maertens, line 380: what confirmation page? what is being tested here?

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
...
+        response = self.app.post(
+            reverse("admin:forms_form_changelist"),
+            {
+                "action": "migrate_to_worldline",
+                "select_across": "0",
+                "index": "0",
+                "_selected_action": "1",
+            },
+            user=self.admin_user,
+        )
```
sergei-maertens, line 399: when you're using web-test, you don't want to be bothering with these kind of payloads. Instead, use webtest to:

1. Navigate to the form change list
2. Locate and select an option in the dropdown using `admin_page.forms[...]["action"].select(...)`
3. Similarly check the checkbox of the form to migrate
4. Use webtest to submit the form `admin_page.forms[...].submit()`

and then verify the result.

Webtest/admin tests should closely mimick the user actions in the browser without being full-blown playwright end-to-end tests

src/openforms/forms/tests/admin/test_views.py
```diff
@@ -305,3 +308,271 @@ def test_single_import(self, m_import):
...
+
+        self.assertRedirects(response, reverse("admin:forms_form_changelist"))
+
+        form_to_migrate.refresh_from_db()
+
+        self.assertEqual(form_to_migrate.payment_backend, "worldline")
+        self.assertEqual(
+            form_to_migrate.payment_backend_options, {"merchant": "merchant-y"}
+        )
```
sergei-maertens, line 426:
```suggestion
        self.assertRedirects(response, reverse("admin:forms_form_changelist"))
        form_to_migrate.refresh_from_db()
        self.assertEqual(form_to_migrate.payment_backend, "worldline")
        self.assertEqual(
            form_to_migrate.payment_backend_options, {"merchant": "merchant-y"}
        )
```
please maintain the visual structure of

```
arrange

act

assert
```
extra blank lines make it harder to scan for what the test setup, execution and assertions are

sergei-maertens (changes requested):

src/openforms/forms/tests/admin/test_views.py
SonnyBA: It tests the confirmation page is rendered properly:
image

PR #5621, opened 2025-09-15
[#5268] Configure Objects API registration for the children component

### LINKED ISSUE #5268: Allow mapping partners/children components to registration options

vaszig:
Closes #5268 partly

**Changes**

- Updated Objects/Objecttypes API with a new object type (children)
- Configured registration to handle children component's data

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
+++ b/src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
@@ -2,7 +2,7 @@ from datetime import UTC, datetime
 from decimal import Decimal
 from pathlib import Path
 from unittest.mock import patch
-from uuid import UUID
+from uuid import UUID, uuid4
 
 from django.test import TestCase, tag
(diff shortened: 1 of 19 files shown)


src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
SonnyBA, line 1: Seeing the amount of lines added to this file might be worth it to split the tests into a separate test case/file 🤔

SonnyBA (approved):
Looks good 👍

src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py
vaszig: I agree and I split that into two separate files, one for the handler and one for the backend (VCR tests).

PR #5627, opened 2025-09-18
Update release template

SonnyBA:
**Changes**

Updates the list of VCR cassettes.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


SonnyBA:
> I was a bit confused with `yivi` plugin, I see that we have `IntegrationTestsBase` class which inherits from the `OFVCRMixin`, but I do not see any data from the tests, so I guess we do not use VCR there?

Maybe @robinmolen can chime in here but I'm also not seeing any cassettes for the tests here. The usage of the `OFVCRMixin` (inherited through the testcase's base class) does not seem relevant for the tests. It is also the only testcase that uses the base class so it's not being used for other testcases inside the app.

PR #5641, opened 2025-09-24
[#5507] prefer detected mimetype for .msg files

### LINKED ISSUE #5507: ".msg" Content type incorect

3.1.x + 3.2.x

### Customer reference

_No response_

The mimetype of .msg files seems to be incorrect in OF causing issues processing the registration in the zaaksystemen. 

This sample .msg file used

```
file --mime-type sample.msg 
sample.msg: application/vnd.ms-outlook
```

But from the moment of uploading to OF this becomes: octet-stream.

Upload to OF:
------WebKitFormBoundaryA5fc2uZ2qDL2oX1x
Content-Disposition: form-data; name="file"; filename="sample.msg"
Content-Type: application/octet-stream

Admin:
Image

voegZaakdocumentToe_Lk01:
```xml
<ZKN:inhoud xmime:contentType="application/octet-stream" StUF:bestandsnaam="sample.msg">
```

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

SonnyBA:
Closes #5507

**Changes**

Overrides the content type for outlook data files from `application/octet-stream` to `application/vnd.ms-outlook`.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

[skip: e2e]


SonnyBA:
Some extra context; while trying to reproduce the current behavior I noticed formio sends the incorrect content type before any backend processing is done:

image

I'm not sure whether this detection can be fixed in formio, since it seems to use the `FileReader` API  (see [second note about mime type detection](https://developer.mozilla.org/en-US/docs/Web/API/Blob/type)), so I opted to fix this through the backend.

src/openforms/formio/tests/test_api_fileupload.py
```diff
@@ -184,6 +184,24 @@ def test_content_inconsistent_with_mime_type(self):
...
         )
 
+    @tag("gh-5507")
+    def test_msg_mimetype_override(self):
+        self._add_submission_to_session(self.submission)
+        url = reverse("api:formio:temporary-file-upload")
+        with open(Path(TEST_FILES, "sample.msg"), "rb") as f:
+            file = SimpleUploadedFile(
+                "sample.msg", f.read(), content_type="application/octet-stream"
+            )
```
sergei-maertens, line 194: I believe you can simplify this to:


```suggestion
        with (TEST_FILES / "sample.msg").open("rb") as infile:
            file = SimpleUploadedFile(
                "sample.msg", infile, content_type="application/octet-stream"
            )
```

src/openforms/formio/api/views.py
```diff
@@ -64,12 +63,22 @@ def post(self, request, *args, **kwargs):
...
+        mime_type = file.content_type
+        _file = file.open()  # reset the file descriptor
+        detected_mime_type = magic.from_buffer(_file.read(), mime=True)
+
+        if (
+            detected_mime_type == "application/vnd.ms-outlook"
+            and _file.content_type == "application/octet-stream"
+            and ext == ".msg"
+        ):
+            mime_type = detected_mime_type
```
sergei-maertens, line 75: we don't need to perform the mime type validation with `magic` here any more, as that has been checked by the validation in the serializer


```suggestion
        match (mime_type := file.content_type):
            # mime type + extension validation was performed in the serializer
            case "application/octet-stream" if ext == ".msg":
                mime_type = "application/vnd.ms-outlook"
            case _:
                pass
```

sergei-maertens (changes requested):

PR #5649, opened 2025-09-25
Squash migrations

sergei-maertens:
Squashed a bunch of migrations as part of the release cycle - once 3.3.x is released, the squashed migrations can be converted to normal migrations.

[skip: e2e]


SonnyBA (approved):
Ran the migrations locally with a up-to-date database and gave no errors. The same was the case for a new database and CI is also happy.

PR #5654, opened 2025-09-26
Optimize slow migration

sergei-maertens:
Completed in 1.5s on my machine for the DH dataset, which ran in 7h (!) on their cluster.

[skip: e2e]


SilviaAmAm (approved):

viktorvanwijk:
> Completed in 1.5s on my machine for the DH dataset, which ran in 7h (!) on their cluster.

Wow 😬

sergei-maertens:
> > Completed in 1.5s on my machine for the DH dataset, which ran in 7h (!) on their cluster.
> 
> Wow 😬

It's bogus - I ran the same benchmark without the optimized code and it completes on my machine in 3.5s. Still a nice speedup, but nowhere close the 7h.

PR #5657, opened 2025-09-29
[#5646] Family members styling enhancements

### LINKED ISSUE #5646: Childeren & Partners enhancements

3.3.0

### Customer reference

_No response_

- [ ] Children labels should work like a repeating group, where you can set the plural label, and the items as a singular item: Kinderen > Kind 1, Kind 2

- [ ] Children in the overview step are hard to distinguish:

Image

- [ ] the pdf contains empty fields for both children and partners:

Image

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #5646

**Changes**

- Removed message for "no information provided" when it's not needed

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/formio/rendering/default.py
+++ b/src/openforms/formio/rendering/default.py
@@ -1,5 +1,4 @@
 from collections.abc import Callable, Iterator
-from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Literal, NotRequired, TypedDict
 
@@ -7,12 +6,12 @@ from django.conf import settings
 from django.urls import reverse
 from django.utils.html import format_html_join
 from django.utils.safestring import SafeString, mark_safe
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext, gettext_lazy as _
 
 from furl import furl
 
 from openforms.emails.utils import strip_tags_plus  # TODO: put somewhere else
-from openforms.formio.typing import ChildrenComponent, Component
+from openforms.formio.typing import Component
 from openforms.submissions.rendering.constants import RenderModes
 from openforms.utils.urls import build_absolute_uri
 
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/templates/report/submission_report.html
```diff
@@ -86,7 +86,10 @@ <h2 class="subtitle">{{ submission_step_node.render }}</h2>
                                                 <div aria-labelledby="{{ node_path }}" class="submission-step-row__value{% if not component_node.display_value %} submission-step-row__value--empty{% endif %}">
                                                     {% if component_node.display_value %}
                                                         {{ component_node.display_value }}
-                                                    {% elif not component_node.layout_modifier %}
+                                                        {% comment %} 
+                                                            the component_label is the custom name that we have for the family members components
+                                                        {% endcomment %}
+                                                    {% elif not component_node.layout_modifier and component_node.key != "component_label" %}
```
sergei-maertens, line 92: I don't understand what's going on here - is this still relevant now that a `layout_modifier` is set?

vaszig: Yes, but I understood that this is only relevant for the parent and not for each item like it is for the `editgrid` for example. The way that the ` editgrid` is implemented is different so that's why I had to add this statement here.

sergei-maertens: Hrm, I don't like how much specificity is leaking into the templates that are meant to be generic node visitors. Is there no other way we can express this? E.g. for the edit grid we have `EditGridNode` which yields `EditGridGroupNode` from its children - can't we do something similar there?

We would then be able to piggy-back on the `layout_modifier: str = "editgrid-group"` which gives us the right styling "for free".

This applies for both the children and partners components I think.

PR #5663, opened 2025-10-01
[#5391] Fix wrong logs and exceptions in the prefill plugins

### LINKED ISSUE #5391: Empty response from prefill plugins when no authentication is present.

3.1.x

### Customer reference

venray 162

Fixed in #4959 where DigiD or KvK would report a empty response if the other plugin was used to start a submission which added a lot of false entries in the daily digest.

Another variant has emerged when there is no authentication plugin used(anonymous). In this case all prefill plugins will log empty responses:

Image
when a form is started with:

Image

When starting with DigiD or eHerkenning the logs show the correct entries. 

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #5391

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/prefill/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class PrefillSkippedException(Exception):
```
sergei-maertens, line 1:
```suggestion
class PrefillSkipped(Exception):
```
the editor/type checker can already tell it's an exception :) I think a shorter name is nicer.

src/openforms/prefill/exceptions.py
```diff
@@ -0,0 +1,10 @@
...
+    """
+    Custom exception for prefill plugins
+
+    This is used in cases where the plugin, for some reason, doesn't manage to complete
+    the operation. This is happening before even making an external call, which can cause
+    another Exception.
```
sergei-maertens, line 7:
```suggestion
    Raise this exception in a plugin if the preconditions or business logic lead to the
    prefill not being necessary, e.g. a certain authentication attribute is not available or
    configuration of the plugin results in no operation needing to be performed. If
    the implementation decides to not make any external calls to perform the prefill,
    then you should raise this exception.
    
    Do not raise this exception in case of prefill failure - those errors are handled 
    separately.
```

sergei-maertens (changes requested):

src/openforms/prefill/sources.py
```diff
@@ -70,6 +71,9 @@ def invoke_plugin(
         log.debug("prefill.plugin.lookup_attributes", attributes=attributes)
         try:
             values = plugin.get_prefill_values(submission, attributes, identifier_role)
+        except PrefillSkipped:
+            log.info("prefill.plugin.skipped")
+            values = {}
```
sergei-maertens, line 76: this is not covered by tests yet :)

src/openforms/prefill/sources.py
```diff
@@ -168,6 +172,9 @@ def fetch_prefill_values_from_options(
             new_values = plugin.get_prefill_values_from_options(
                 submission, plugin_options, variable
             )
+        except PrefillSkipped:
+            log.info("prefill.plugin.skipped")
+            values = {}
```
sergei-maertens, line 177: this is not covered by tests yet :)

sergei-maertens (changes requested):

PR #5665, opened 2025-10-02
Fix empty values of date related objects

### LINKED ISSUE #5134: Move handling of clearOnHide property from `evaluate_form_logic` to `PropertyAction`
Part of #5117, see [comment](https://github.com/open-formulieren/open-forms/issues/5117#issuecomment-2696634735) for more context. 

Suggestion in that comment:
> Extend `PropertyAction` with an `eval()` method so we can address the `clearOnHide` property directly while evaluating all the logic rules. (Something to take into account #4324?)

This will enable tracking of variable changes during instead of after logic evaluation.

viktorvanwijk:
Partly closes #5134

[skip: e2e]

**Changes**

* The initial value of a submission value variable is now the empty value instead of `None`, if the related form value had no initial value
* Submission value variables and `SubmissionStep.data` are now serialized to JSON manually when saving to the DB and serializing for API response (this was done exclusively by the `DjangoJSONEncoder` before)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -271,10 +271,25 @@ class SubmissionStepSerializer(NestedHyperlinkedModelSerializer):
         new_configuration = evaluate_form_logic(instance.submission, instance)
         # update the config for serialization
         instance.form_step.form_definition.configuration = new_configuration
+
         representation = super().to_representation(instance)
 
-        if self.context.get("unsaved_data_only", False):
-            representation["data"] = instance.unsaved_data
+        # Serialize the data to JSON. We can't use a generic JSON encoder here, because
+        # we need context from the variable
+        data = (
+            instance.unsaved_data
+            if self.context.get("unsaved_data_only", False)
+            else instance.data
+        )
+        serialized_data = FormioData()
+        state = instance.submission.load_submission_value_variables_state()
+        variables = state.get_variables_in_submission_step(instance)
+        for key, variable in variables.items():
+            if key not in data:
+                continue
+            serialized_data[key] = variable.to_json(data[key])
+
+        representation["data"] = serialized_data.data
 
         return representation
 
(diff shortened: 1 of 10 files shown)


src/openforms/submissions/api/serializers.py
```diff
@@ -271,10 +271,25 @@ def to_representation(self, instance):
...
+        )
+        serialized_data = FormioData()
+        state = instance.submission.load_submission_value_variables_state()
+        variables = state.get_variables_in_submission_step(instance)
+        for key, variable in variables.items():
+            if key not in data:
+                continue
+            serialized_data[key] = variable.to_json(data[key])
+
+        representation["data"] = serialized_data.data
```
sergei-maertens, line 292: do you have some performance benchmarks for the before/after of this change? I'm curious if there's any difference.

src/openforms/submissions/tests/form_logic/test_submission_logic.py
```diff
@@ -745,47 +745,6 @@ def test_with_default_values(self):
             )
 
     @tag("gh-2056")
```
sergei-maertens, line 747: this tag probably also needs to be removed?

src/openforms/submissions/api/serializers.py
viktorvanwijk: I did some testing using silk on the submission step detail endpoint for a large DH form step with lots of fields (including editgrids).

**Response times**
Before changes (10 requests, times in ms):
1050, 380, 231, 201, 204, 218, 212, 216, 309, 210 (mean 323, std 248)

After changes (10 requests, times in ms):
214, 445, 365, 633, 390, 398, 211, 206, 212, 215 (mean 328, std 136)

The mean before changes is slightly lower, but the variation much higher. The query times don't seem to be consistent here.

**Cprofiling**
A typical request before changes:
image
0.064 sec inside `to_representation` of which 0.061 sec spent on `evaluate_form_logic`

A typical request after changes:
image
0.066 s inside `to_representation` of which 0.062 sec spent on `evaluate_form_logic`

It doesn't seem like a lot of extra time is spent in `to_representation`, so should be OK I think. I didn't check the logic check endpoint, as the additional time will be even lower because we only return the data difference

sergei-maertens (approved):

PR #5666, opened 2025-10-02
[#3535] Fix pausing form allows navigating to next step

### LINKED ISSUE #3535: Pausing a form allows user to click next step, even if progress is blocked

2.2.0 - latest

A user can pause a form, resume and just click the next step in the progress-indicator to skip the step they are supposed to resume. 

Start form
Pause form
Continue form
Click next available step

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #3535

**Changes**

- Do not always mark a submission step as completed (now even if a form is paused the submission step is saved so the backend treats this as a completed step). This allows the user to navigate to the next step (which in a regular submission is not allowed). OF though can prevent the user from "forgetting" to provide data for a required field in a previous step (we show validation errors).

UPDATE

A new submission step field is added to track the date and time that a submission step was really completed. This is based on the SDK too, which is sending a flag for this (whether the step was saved via the regular submit or a suspended form).

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/submissions/tests/test_submission_suspension.py
```diff
@@ -314,3 +314,25 @@ def test_submission_cannot_be_suspended_when_already_completed(self):
         response = self.client.post(endpoint, {"email": "hello@open-forms.nl"})
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    @freeze_time("2020-12-11T10:53:19+01:00")
+    def test_resuming_submission_and_completed_steps(self):
```
vaszig: Reminder!

Prefer `generate_minimal_setup` and maybe `from_components`?

vaszig:
~Self-reminder: add an e2e test too.~

src/openforms/submissions/api/viewsets.py
```diff
@@ -621,6 +622,12 @@ def update(self, request, *args, **kwargs):
         serializer.is_valid(raise_exception=True)
 
         create = instance.pk is None
+
+        # regular submission step should get a timestamp upon first save
+        saving_from_suspension = serializer.validated_data["from_suspension"]
+        if create and not saving_from_suspension:
+            instance.completed_on = timezone.now()
```
sergei-maertens, line 629:
```suggestion
        # regular submission step should get a timestamp upon first save
        saving_from_suspension = serializer.validated_data["from_suspension"]
        if not saving_from_suspension:
            instance.completed_on = timezone.now()
```
Even with an update (when the user is editing data from the summary page for example) I'd update the timestamp so that we capture when the step was last submitted rather than first.

src/openforms/submissions/models/submission_step.py
```diff
@@ -101,6 +101,15 @@ class SubmissionStep(models.Model):  # noqa: DJ008
...
     modified = models.DateTimeField(_("modified on"), auto_now=True)
+    completed_on = models.DateTimeField(
+        _("completed on"),
+        blank=True,
+        null=True,
+        help_text=_(
+            "This is indicating the date and time that a submission step was for the "
+            "first time completed (all the future are considered as modifications)."
+        ),
+    )
```
sergei-maertens, line 112:
```suggestion
    completed_on = models.DateTimeField(
        _("completed on"),
        blank=True,
        null=True,
        help_text=_(
            "Marks when the step was last submitted. When a submission is being paused, "
            "the field will be empty to signal it wasn't completed yet."
        ),
    )
```

src/openforms/submissions/tests/factories.py
```diff
@@ -292,10 +292,18 @@ class SubmissionStepFactory(factory.django.DjangoModelFactory):
     form_step = factory.SubFactory(
         FormStepFactory, form=factory.SelfAttribute("..submission.form")
     )
+    completed_on = factory.Maybe(
+        "from_suspension",
+        yes_declaration=None,
+        no_declaration=factory.Faker("date_time_this_month", tzinfo=UTC),
```
sergei-maertens: you probably want to make sure this can't be before `timezone.now` :)

src/openforms/submissions/tests/factories.py
```diff
@@ -292,10 +292,18 @@ class SubmissionStepFactory(factory.django.DjangoModelFactory):
...
+    )
 
     class Meta:
         model = SubmissionStep
 
+    class Params:
+        from_suspension = False
```
sergei-maertens: You can define `completed_on` here directly as a `factory.Trait` instead of using `factory.Maybe`

src/openforms/submissions/tests/test_submission_step_validate.py
```diff
@@ -547,6 +547,36 @@ def test_data_validated(self):
         expected_names = {"data.textfield", "data.repeatingGroup.1.number"}
         self.assertEqual(names, expected_names)
 
+    def test_timestamp_is_set_for_regular_submitted_step(self):
```
sergei-maertens: this can/should be moved into the `test_submit_form_step` module - this test has nothing to do with validation.

Additionally, I'd find an existing test to add the assertion to instead of defining a completely new test. It's an additional piece of logic for when a step is created/updated, so find those existing tests and extend them.

sergei-maertens (changes requested):

src/openforms/tests/e2e/test_resume_form.py
```diff
@@ -73,5 +85,11 @@ def get_submission_uuid_and_resume_url(form, request, live_server):
                 )
 
                 header = page.get_by_role("heading", name="First step")
+                second_form_step = page.get_by_role("link", name="Second step")
 
                 await expect(header).to_be_visible()
+                # the second step should not be enabled since the first one has not yet
+                # been completed (suspended form)
+                await expect(second_form_step).to_have_attribute(
```
vaszig, line 96: This is failing only in CI, trying to solve that.

src/openforms/submissions/tests/factories.py
```diff
@@ -292,10 +292,14 @@ class SubmissionStepFactory(factory.django.DjangoModelFactory):
     form_step = factory.SubFactory(
         FormStepFactory, form=factory.SelfAttribute("..submission.form")
     )
+    completed_on = factory.LazyFunction(lambda: timezone.now() + timedelta(minutes=1))
```
sergei-maertens:
```suggestion
    completed_on = factory.LazyFunction(lambda: timezone.now() - timedelta(minutes=1))
```
this can't be in the future, can it? 😛

src/openforms/tests/e2e/test_resume_form.py
sergei-maertens: FYI - for failing e2e tests there's an artifact uploaded to github actions that you can use to the playwright trace viewer (https://playwright.dev/docs/trace-viewer)

sergei-maertens (approved):

PR #5674, opened 2025-10-03
Fix clear on hide behaviour

### LINKED ISSUE #5134: Move handling of clearOnHide property from `evaluate_form_logic` to `PropertyAction`

viktorvanwijk:
Closes #5134

Depends on #5665 

**Changes**

- Added explicit evaluation of conditional logic
- Moved clear-on-hide behaviour to the property action
- Updated bunch of tests

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/formio/registry.py
+++ b/src/openforms/formio/registry.py
@@ -189,6 +189,73 @@ class ComponentRegistry(BaseRegistry[BasePlugin]):
         formatter = component_plugin.formatter(as_html=as_html)
         return formatter(component, value)
 
+    def test_conditional(
+        self, component: Component, value: VariableValue, compare_value: VariableValue
+    ) -> bool:
+        """
+        Perform a component-specific comparison whether a conditional is triggered.
+
+        :param component: Component to evaluate.
+        :param value: Value to evaluate.
+        :param compare_value: Value to compare - should be fetched from the conditional
+          in the component configuration.
+        :return: Whether the conditional was triggered.
+        """
+        if (component_type := component["type"]) not in self:
+            component_type = "default"
+
+        plugin = self[component_type]
+
+        # First, see if the component has a custom test conditional
+        if (
+            triggered := plugin.test_conditional(component, value, compare_value)
+        ) is not None:
+            return triggered
+
+        # If it does not have one, default to a membership test if the component is
+        # configured as multiple, or a direct comparison otherwise
+        if component.get("multiple", False):
+            assert isinstance(value, list)
+            return compare_value in value
+        else:
+            return value == compare_value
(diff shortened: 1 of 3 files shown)


viktorvanwijk:
Profiling results
-------------------
Tested by filling out a large DH form (>300 components and 65 logic rules)

**Check logic response times**
Before: average 282 ms, std 92 ms
After: average 310 ms, std 127 ms

Increase of approximately 10%, but also std has increased significantly indicating more outliers. The query times don’t seem to be really consistent.

Looking at one of those outliers (after the changes) reveals that most of the time was spent getting the rules from the DB:
image


**cPython profiling**
When looking at the time spent inside the `evaluate_form_logic` function using the data obtained from the profiling:
Before: mean 67 ms, std 27 ms
After: mean 71, std 60 ms

Increase of approximately 5%, and again increased std

Looking at `evaluate_conditional_logic` specifically, it only shows up in the table ~60% of the time with the following numbers: mean 4.5 ms, std 5.7 ms. There are a few outliers of around 30-40 ms for steps with an editgrid. Here, most of the time is spent on getting the evaluation data for editgrid items (`get_evaluation_data`):
image

After implementing an optimization for getting the evaluation data, `evaluate_conditional_logic` has the following numbers: mean 3.1 ms, std 0.6 ms. The significantly lower std indicates that the outliers of 30-40 ms have disappeared.

**Summary**
Check logic response times:
Before: mean 282 ms, std 92 ms
After: mean 310 ms, std 127 ms
Optimization: 308 ms, std 110 ms

Difference seems largely due to inconsistent DB query times.

`evaluate_form_logic`:
Before: mean 67 ms, std 27 ms
After: mean 71 ms, std 60 ms
Optimization: mean 72 ms, std 50 ms

Note that this function also performs interactions with the DB.

`evaluate_conditional_logic`:
Before: -
After:  mean 4.5 ms, std 5.7 ms
Optimization: 3.1 ms, std 0.6 ms

This is a clear indication that the outliers (when processing editgrid data) have disappeared.

viktorvanwijk:
I have noticed some flakiness on (seemingly) random tests. Not sure if it was introduces with my changes. Will keep an eye on it

PR #5680, opened 2025-10-07
[#3544] Do not delete submissions related to valid appointments

### LINKED ISSUE #3544: Submissions are possibly deleted before the related appointment date has passed

master / cc8fe709fcb40280c8d32bdab45f55b57304a7a0

Open Forms has a feature where submissions (matching certain conditions) are automatically deleted/stripped of sensitive data.

If a submission is for an appointment form and configured to delete completed submissions after two weeks for example, then any cancel links sent via confirmation email likely break if the appointment is further in the future than two weeks. I don't think that the submission pruning takes into account related appointment dates.

The logic should be that if a form is an appointment form, only submissions where the appointment date has passed may be deleted.

1. Create an appointment form and configure submissions to be deleted after one day
2. Create an appointment for a week later, for example
3. Copy the cancel link from the confirmation email
4. (one day later) run the submission pruning job
5. confirm that the submission was deleted
6. use the cancel link from step 3. and assert that it's not possible anymore to cancel the appointment

Appointments can be cancelled up until the scheduled date/time of the appointment.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #3544

**Changes**

- Added extra filter for appointments. When a valid appointment (date in the future) exists, its related submission cannot be deleted. This makes sure that the "cancel" endpoint is working as expected.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/appointments/tests/test_api_cancel_appointment.py
+++ b/src/openforms/appointments/tests/test_api_cancel_appointment.py
@@ -1,12 +1,22 @@
+from datetime import datetime
+
+import pytz
+from freezegun import freeze_time
 from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APITestCase
 
 from openforms.appointments.tests.factories import AppointmentInfoFactory
+from openforms.data_removal.constants import RemovalMethods
+from openforms.data_removal.tasks import delete_submissions
+from openforms.forms.tests.factories import FormFactory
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):
Nice fix - I'd just rename a variable.

PR #5700, opened 2025-10-15
Rename metrics with proper naming conventions

sergei-maertens:
OTel prescribes naming conventions: https://opentelemetry.io/docs/specs/semconv/general/naming/ and we were not following those yet.

Now that we're still in the early stages, we can still get away with this kind of breaking change.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/accounts/metrics.py
+++ b/src/openforms/accounts/metrics.py
@@ -32,32 +32,32 @@ def count_users(options: metrics.CallbackOptions) -> Collection[metrics.Observat
 
 
 meter.create_observable_gauge(
-    name="user_count",
+    name="openforms.auth.user_count",
     description="The number of application users in the database.",
-    unit="",  # no unit so that the _ratio suffix is not added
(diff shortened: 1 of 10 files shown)


sergei-maertens:
@swrichards I'd appreciate if you can take a look at this!

annashamray (approved):
I confirm that I see the metrics with the changed names in the local setup.

src/openforms/accounts/metrics.py
swrichards: Do we need to look into scoping these attributes?

sergei-maertens: I considered it, but `type` is really a very common attribute and the values are determined by the context, similarly `scope` is a pattern that transcends across multiple unrelated metrics so there too it makes more sense to keep them in the global scope IMO. The alternative might be `maykin.scope` or something like that, which I don't find more elegant.

swrichards: Clear, and agree, thanks.

PR #5703, opened 2025-10-17
[#5696] Add JCC api calls (2)

### LINKED ISSUE #5696: Implement JCC API calls
Note: to properly test these, we need #5688 to be ready. Though the functionality and tests can already be written, and tested manually with Swagger (or a quick and dirty local OAuth implementation)

- [x] `get_available_products` @viktorvanwijk 
- [x] `get_locations` @viktorvanwijk 
- [x] `get_dates` @viktorvanwijk 
- [x] `get_times` @viktorvanwijk 
- [x] `get_required_customer_fields` @vaszig 
- [x] `create_appointment` @vaszig 
- [x] `delete_appointment` @vaszig 
- [x] `get_appointment_details` @vaszig

vaszig:
Closes #5696 partly

**Changes**

- [x] `get_required_customer_fields`
- [x] `create_appointment`
- [x] `delete_appointment`
- [x] `get_appointment_details`

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/appointments/contrib/jcc_rest/client.py
```diff
@@ -0,0 +1,155 @@
...
+            response = self.get(endpoint)
+            response.raise_for_status()
+
+        return response.json()
+
+    def get_required_customer_fields(
+        self, product_uuids: Sequence[str]
+    ) -> CustomerFields:
+        """Retrieve all the required customer field names based on the products UUIDs."""
+        params = {"selectedActivityId": product_uuids}
```
viktorvanwijk: TIL you can just pass it as a list and it will add a parameter for each item. I did it manually for my API calls :sweat_smile:

src/openforms/appointments/contrib/jcc_rest/constants.py
```diff
@@ -1,5 +1,295 @@
...
+    """
+
+    country = "country", _("Country")
+    """
+    Name of the country where customer lives in, max length 80.
+    """
+
+
+def get_component(component_type: CustomerFields, required: bool):
+    FIELD_TO_FORMIO_COMPONENT: dict[CustomerFields, Component] = {
```
viktorvanwijk: Perhaps define `FIELD_TO_FORMIO_COMPONENT` separately and use `get_component` to set the required field
```python
def get_component(component_type: CustomerFields, required: bool):
    component = deepcopy(FIELD_TO_FORMIO_COMPONENT[component_type])
    component["validate"]["required"] = required
    return component
```
Or use a match statement? Seems a bit wasteful to define the whole dict every time a single component is requested

src/openforms/appointments/contrib/jcc_rest/constants.py
```diff
@@ -1,5 +1,295 @@
...
+            "type": "radio",
+            "key": CustomerFields.gender.value,
+            "label": CustomerFields.gender.label,
+            "validate": {
+                "required": required,
+            },
+            "values": [  # pyright: ignore[reportAssignmentType]
+                {"value": "M", "label": _("Male")},
+                {"value": "F", "label": _("Female")},
+                {"value": "N", "label": _("Not specified")},
```
viktorvanwijk: Shouldn't the values here represent the options we need to send to the endpoint (0, 1, or 2)? Or is there a custom conversion somewhere else?

Also, I think "Other" or "Prefer not to say" is a better label, as it should be an answer to the question for the user "What is your gender?"

src/openforms/appointments/contrib/jcc_rest/plugin.py
```diff
@@ -97,7 +96,74 @@ def get_required_customer_fields(
...
     ) -> list[Component]:
-        return []
+        product_ids = [product.identifier for product in products]
+        with JccRestClient() as client:
+            # Each field from the API has a value which determines if it's visible, hidden
+            # or required. We need only the visible (optional fields in the form) and the
+            # required ones (required fields in the form).
+            # 0 = visible
+            # 1 = hidden
+            # 2 = required
```
viktorvanwijk: I think adding an enum/constants for these values would improve readability of the statements below :)

src/openforms/appointments/contrib/jcc_rest/plugin.py
```diff
@@ -97,7 +96,74 @@ def get_required_customer_fields(
...
+            required_fields = client.get_required_customer_fields(product_ids)
+
+            # If 'areFirstNameOrInitialsRequired' and 'isAnyPhoneNumberRequired' are set
+            # to 'true' in the configuration and both fields are set to 'required' by the
+            # municipality, then both must still be completed. The municipality can specify
+            # in the application whether a field should only be 'visible' or whether a
+            # field should be 'required.'
+            default_number_field = []
+            default_name_or_initials = []
+            if required_fields.get("isAnyPhoneNumberRequired"):
```
viktorvanwijk: What if `isAnyPhoneNumberRequired` is true, and both fields are set to visible? We should mark one of them as required then right?

src/openforms/appointments/contrib/jcc_rest/plugin.py
```diff
@@ -97,7 +96,74 @@ def get_required_customer_fields(
...
+                    not mobile_number or mobile_number == 1
+                ):
+                    default_number_field = [
+                        get_component(CustomerFields.phone_number, True)
+                    ]
+            if required_fields.get("areFirstNameOrInitialsRequired"):
+                initials = required_fields.get("initials")
+                first_name = required_fields.get("firstName")
+                if (not initials or initials == 1) and (
+                    not first_name or first_name == 1
```
viktorvanwijk: Could you add a test what happens when both fields are marked as visible (0), but `areFirstNameOrInitialsRequired` is set to `True`? It seems like we do set a `default_name_or_initials` because of the `not initials` and `not first_name`, but I'm afraid it gets added twice at the end as all fields from `field_names` are added to the list as well.

I think being explicit here in checking the value of the fields might help with readability as well: `initials in (0, 1)`, or using an enum `initials in (state.visible, state.hidden)`

src/openforms/appointments/contrib/jcc_rest/plugin.py
```diff
@@ -97,7 +96,74 @@ def get_required_customer_fields(
...
+
+            return (
+                default_number_field
+                + default_name_or_initials
+                + last_name
+                + main_phone_number
+                + [
+                    get_component(CustomerFields(field), required)
+                    for field, required in field_names.items()
+                ]
```
viktorvanwijk: Is this also the order in which the fields are shown in the frontend? If so, perhaps move the default number field to below the default and last name? I know we can't really control the order returned by JCC anyway, but I guess it's the best we can do :shrug:

src/openforms/appointments/contrib/jcc_rest/types.py
```diff
@@ -0,0 +1,195 @@
...
+# 0: visible (it can be submitted without a value)
+# 1: hidden (it's not shown and it's not required)
+# 2: required (should be shown and should have a value)
+FIELD_TYPE = Literal[0, 1, 2]
+
+
+class Revision(TypedDict):
+    id: str
+    creationDateTime: str  # ISO 8601 format
+    creatorAuthenticationUserId: str | None
```
viktorvanwijk: Whenever a field has "[type] or null" in the documentation, I think we should treat it as "[type] or missing", which was confirmed by JCC when we asked. Also in my own VCR casssettes, I didn't find a single instance where a field was `null`

src/openforms/appointments/contrib/jcc_rest/types.py
```diff
@@ -0,0 +1,195 @@
...
+    appointmentDuration: int
+    maxCountForEvent: int | None
+    activityGroupId: str | None
+    synonyms: list[str] | None
+
+
+class Customer(TypedDict):
+    id: str | None
+    isMainCustomer: bool
+    gender: FIELD_TYPE
```
viktorvanwijk: I know it's the same values, but could you use a separate literal for this? It has a different meaning than the one for the customer fields

viktorvanwijk (commented):
Some thoughts and suggestions :)

src/openforms/appointments/contrib/jcc_rest/client.py
```diff
@@ -0,0 +1,155 @@
...
+        activity_ids: Sequence[str],
+        amounts: Sequence[int],
+    ) -> int:
+        """Retrieve the duration (in minutes) for an appointment."""
+        assert len(activity_ids) == len(amounts), (
+            "Number of activity IDs and amounts must be equal"
+        )
+
+        params = []
+        for id_, amount in zip(activity_ids, amounts, strict=False):
```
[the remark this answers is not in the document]
viktorvanwijk: OK, tested it on my endpoint, and it does work. Though maybe better to keep it like this to match the example in the documentation?

src/openforms/appointments/api/serializers.py
```diff
@@ -282,7 +285,39 @@ def validate(self, attrs: dict) -> dict:
...
+
+        # Plugins may require combined fields validation so we handle these situations
+        # based on specific rules (for now only JCC_REST uses it so we use just an if statement)
+        specific_errors = {}
+        if required_group_fields:
+            for group in required_group_fields:
+                if group["type"] != "require_one_of":
+                    continue
+
+                contact_details_meta = deepcopy(contact_details_meta)
```
viktorvanwijk: This copy is not necessary when you set it to the same variable, right?

src/openforms/appointments/base.py
```diff
@@ -73,6 +74,22 @@ def __str__(self):
         return self.identifier
 
 
+@dataclass
+class RequireOneOfRule:
+    enabled: bool
+    fields: tuple[str, ...]
+    labels: tuple[str, ...]
```
viktorvanwijk: I feel like this should be part of the JCC plugin specifically, because it's only used as a helper dataclass there. Nothing in the api for `get_required_customer_fields` dictates that this should be used

src/openforms/appointments/contrib/jcc_rest/client.py
```diff
@@ -162,6 +171,93 @@ def get_available_times_for_appointment(
...
+    def get_duration_for_appointment(
+        self,
+        date: date,
+        activity_ids: Sequence[str],
+        amounts: Sequence[int],
+    ) -> int:
+        """Retrieve the duration (in minutes) for an appointment."""
+        assert len(activity_ids) == len(amounts), (
+            "Number of activity IDs and amounts must be equal"
+        )
```
viktorvanwijk: You didn't want to combine them like I did in `get_available_times_for_appointment`?

src/openforms/appointments/contrib/jcc_rest/constants.py
```diff
@@ -1,5 +1,296 @@
...
+FIELD_TO_FORMIO_COMPONENT: dict[CustomerFields, Component] = {
+    CustomerFields.gender: {
+        "type": "radio",
+        "key": CustomerFields.gender.value,
+        "label": CustomerFields.gender.label,
+        "validate": {},
+        "values": [  # pyright: ignore[reportAssignmentType]
+            {"value": GenderType.other, "label": _("Other")},
+            {"value": GenderType.male, "label": _("Male")},
+            {"value": GenderType.female, "label": _("Female")},
```
viktorvanwijk:
```suggestion
            {"value": GenderType.male.value, "label": GenderType.male.label},
            {"value": GenderType.female.value, "label": GenderType.female.label},
            {"value": GenderType.other.value, "label": GenderType.other.label},
```
Put other at the end maybe?

src/openforms/appointments/contrib/jcc_rest/plugin.py
```diff
@@ -157,12 +177,144 @@ def get_times(
...
+            return [component_factory]
+
+        return []
+
+    def _extract_rendered_fields(
+        self, required_fields: RawCustomerFields
+    ) -> dict[str, bool]:
+        """
+        Based on the retrieved fields returns a dict with the fields and whether they are
+        required or not.
```
viktorvanwijk:
```suggestion
        Based on the retrieved fields returns a dict with the fields that should be rendered and whether they are
        required or not.
```

src/openforms/appointments/contrib/jcc_rest/plugin.py
```diff
@@ -157,12 +177,144 @@ def get_times(
...
+
+        rendered_fields = self._extract_rendered_fields(required_fields)
+
+        components = []
+        # Name / initials defaults
+        components.extend(
+            self._get_default_component(
+                required_fields.get("areFirstNameOrInitialsRequired"),
+                required_fields.get("initials"),
+                required_fields.get("firstName"),
```
viktorvanwijk: These values do not match the type hints of `self._get_default_components` as they can be `None` as well. Suggestion to update the type hints or add default values

src/openforms/appointments/contrib/jcc_rest/tests/test_plugin.py
```diff
@@ -29,7 +32,39 @@ def _scrub_access_token(response):
...
+
+    To do so:
+
+    #. Define the required environmental variables and their values (`JCC_REST_CLIENT_ID`,`JCC_REST_SECRET`)
+    #. Ensure the config is still valid:
+       - `token` needs to be valid
+       - endpoints must work as expected
+       - data is still valid
+    #. Delete the VCR cassettes
+    #. Run the tests
```
viktorvanwijk:
```suggestion
    #. Delete the VCR cassettes
    #. Update the ``RECORDING_DATETIME`` class attribute to the date of recording
    #. Run the tests
```

viktorvanwijk (commented):
General flow looks OK, just a few minor things :)

One thought I had is to put all of the mocked requested into a separate test class, which will avoid a lot of VCR cassettes for the token requests. But we can leave that for later. Regarding the mocks themselves, I think you've covered all the relevant cases now. I would be nice of course if we could get this data from JCC, but I'm really not sure if we should spend the time on it to be honest. If you want to send the email, feel free.

Could you do me a favor and process the feedback in a separate commit :innocent:

vaszig:
> General flow looks OK, just a few minor things :)
> 
> One thought I had is to put all of the mocked requested into a separate test class, which will avoid a lot of VCR cassettes for the token requests. But we can leave that for later.
> 
> Could you do me a favor and process the feedback in a separate commit 😇

My apologies, I am used to other people reviewing the processed feedback by clicking the "compare" button. I will do that in a separate commit though!

viktorvanwijk (approved):

src/openforms/appointments/contrib/jcc_rest/types.py
```diff
@@ -107,25 +96,25 @@ class Location(BaseLocation):
 class CustomerFields(TypedDict):
     auditId: str
     id: NotRequired[str]
-    gender: GENDER_TYPE
-    firstName: FIELD_STATE
-    initials: FIELD_STATE
+    gender: GenderType
```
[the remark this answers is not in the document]
vaszig: Stupid..I had it correct and last minute I saw it and made it wrong again. I am fixing this.

PR #5719, opened 2025-10-24
Enable type checking on entire `src/openforms/authentication` package

sergei-maertens:
Finally :weary: 

[skip: e2e]

### DIFF
--- a/src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
+++ b/src/openforms/authentication/contrib/yivi_oidc/oidc_plugins/plugins.py
@@ -6,7 +6,7 @@ from typing import cast, override
 
 from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import PermissionDenied
-from django.http import HttpRequest, HttpResponse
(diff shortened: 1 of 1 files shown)


src/openforms/authentication/forms.py
```diff
@@ -45,10 +45,9 @@ class RegistratorSubjectInfoForm(forms.Form):
     )
 
     def clean(self):
-        cleaned_data = super().clean()
-        bsn = cleaned_data.get("bsn")
-        kvk = cleaned_data.get("kvk")
-        skip_subject = cleaned_data.get("skip_subject")
+        bsn = self.cleaned_data.get("bsn")
```
vaszig: Why did we remove the `super()` call?Isn't it more safe concerning the values?I had in mind that we first call the `super().clean()` and then do what we need with these values.

sergei-maertens: `super().clean()` can return `None`, there's no guarantee that it will actually return `self.cleaned_data`. This is something that changed/got clarified around Django 1.11-2.2 if I remember correctly.

I do need to add the `super` call back though

PR #5732, opened 2025-11-05
:memo: Add documentation on how to develop UI components

sergei-maertens:
Let's make some implicit knowledge explicit.

[skip: e2e]

### DIFF
--- a/docs/developers/ui-components.rst
+++ b/docs/developers/ui-components.rst
@@ -1,0 +1,337 @@
+.. _developers_ui_components:
+
+======================
+Creating UI components
+======================
+
+UI components are (reusable) logical building blocks to build user interfaces. Some
+obvious examples are buttons and links. More complex examples are footers with navigation
+elements or a form field input with label, input and validation errors.
+
+Development of (new) UI components in Open Forms is subject to some guidelines and
+rules, with the goal of:
+
+* contribute to good accessibility
+* make it possible to *theme* them, so appearance can be tailored to the organization
+* being reusable in similar but different contexts
+* make it easy to maintain the code
+
+The next sections describe how to approach creating or modifying UI components.
+
+Check for existing community components
+=======================================
+
+The `NL Design System components <https://www.nldesignsystem.nl/componenten/>`_
+catalogue already lists a large number of components that may fit your requirements.
+
+You should always check first if there's anything suitable. This can lead to an easily
+re-used community component, or at a minimum provide resources about how to implement
+such a component.
+
+Make sure to read up on how other organizations use components, which design constraints
+they have taken into consideration and where the variation lies if different community
+implementations exist.
+
+If nothing exists that is readily usable, you can proceed with implementing it yourself,
+but you should still keep in mind the NL DS research so that future implementations can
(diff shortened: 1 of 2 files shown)


docs/developers/ui-components.rst
```diff
@@ -0,0 +1,301 @@
...
+
+You should always check first if there's anything suitable. This can lead to an easily
+re-used community component, or at a minimum provide resources about how to implement
+such a component.
+
+Make sure to read up on how other organizations use components, which design constraints
+they have taken into consideration and where the variation lies if different community
+implementations exist.
+
+If nothing exists that is readily usable, you can proceed with implementing it yourself,
```
vaszig: Should we mention to consider contributing this to NL DS?

docs/developers/ui-components.rst
```diff
@@ -0,0 +1,301 @@
...
+.. tip::
+
+    * You should try to limit the number of different blocks that exist in a codebase.
+    * Avoid using extremely specific blocks that cannot be reused in other places.
+
+**Elements**
+
+The element can contain zero or more child elements, which get a ``.my-block__element-1``
+or ``.my-block__element-2`` class name. They can exist anywhere in the HTML tree, they
+are not limited to direct children.
```
vaszig: Could you please explain a bit this one?

docs/developers/ui-components.rst
```diff
@@ -0,0 +1,301 @@
...
+
+.. warning::
+    A modifier is always relative to its base block or element, meaning you must include
+    that base class name too. ``<a href="#" class="utrecht-link--html-a">...</a>`` is
+    incorrect usage, instead it must be:
+    ``<a href="#" class="utrecht-link utrecht-link--html-a">...</a>``
+
+**SCSS helpers**
+
+The library ``microscope-sass`` has a helper to build the correct BEM class names.
```
vaszig: I am again confused with this library, I thought we wanted to gradually get rid of it.

docs/developers/ui-components.rst
sergei-maertens: I have some ideas about this - I don't think there will always be a case to put it in the NL DS organization/maintainer scope. We might instead at some point create a UI components package for OF components, which is then used by the SDK/renderer and could offer re-usable components in the community.

For now, we stick to serving our own needs though.

docs/developers/ui-components.rst
sergei-maertens: Pointed out in the docs that there's nesting happening here, even though in the BEM they seem siblings.

docs/developers/ui-components.rst
```diff
@@ -0,0 +1,301 @@
...
+
+.. code-block:: css
+
+    .openforms-theme {
+        --of-footer-background-color: transparent;
+        --of-footer-color: #000000;
+    }
+
+Each component can only consume design tokens that it "owns" - you may not consume
+other tokens that happen or are likely to have the same value:
```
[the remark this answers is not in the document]
sergei-maertens: Added documentation about the token structure and an example of tokens for an element!

PR #5750, opened 2025-11-14
Fix bad instances of assert usage

sergei-maertens:
Detected via Sentry and some vague bug reports.

[skip: e2e]


sergei-maertens:
Backports:

* stable/3.3.x: a545fba1d29c5d6262a1b5b9f400008d5fb6791e
* older versions are not affected

PR #5755, opened 2025-11-18
Ban typing.cast from the codebase

sergei-maertens:
typing.cast always overrules the type checker and allows you to write lies way too easily. The code being cast may also change, and the CI/type checker will never report if the cast is no longer necessary or changed in nature making it more or less wrong. Additionally, it's hard to keep track of why the cast is/was necessary in this context.

Instead, it's usually preferred to narrow types for the type checker using asserts, which are stripped out in production so the perf hit is irrelevant. If that is not possible, using a specific type ignore comment is better because it at least points at the kind of problem is being resolved, and we can (in the future) check if the type ignores go down rather than up when the majority of the codebase is checked in CI *and* tooling in the ecosystem improves.

The rule suppressions in the registrations app are necessary because TypedDicts aren't easy/pleasant to work with, especially when correlating classes with behaviour and value types with them. The types in the openforms.formio package are far from ideal, but this is too much to address now.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

[skip: e2e]

### DIFF
--- a/pyright.pyproject.toml
+++ b/pyright.pyproject.toml
@@ -106,3 +106,5 @@ ignore = []
 
 pythonVersion = "3.12"
 pythonPlatform = "Linux"
+
+reportUnnecessaryTypeIgnoreComment = "error"
--- a/src/openforms/authentication/contrib/digid_eherkenning_oidc/migrations/0002_ofeidasconfig.py
+++ b/src/openforms/authentication/contrib/digid_eherkenning_oidc/migrations/0002_ofeidasconfig.py
@@ -210,7 +210,7 @@ class Migration(migrations.Migration):
                 ),
                 (
                     "person_identifier_claim",
-                    mozilla_django_oidc_db.fields.ClaimField(  # type: ignore
+                    mozilla_django_oidc_db.fields.ClaimField(
                         base_field=models.CharField(
                             max_length=50, verbose_name="claim path segment"
                         ),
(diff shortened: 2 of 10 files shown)


src/openforms/api/validators.py
```diff
@@ -73,15 +70,17 @@ def __init__(self, validator_func: Callable[[MT], None]):
...
         model_fields = [f.name for f in serializer.Meta.model._meta.get_fields()]
 
         # Assign the serializer data to the model instance as the model serializer would
         # do for a create or update when all the data is valid - this allows validator
         # functions to operate on model instances and shared logic between serializers
         # and Django model.clean() methods.
-        instance = cast(MT, get_model_serializer_instance(serializer))
+
+        instance = get_model_serializer_instance(serializer)
+        assert isinstance(instance, models.Model)
```
vaszig: A bit confused with that..I would expect this to be of type `MT` .

src/openforms/validations/api/serializers.py
```diff
@@ -35,18 +33,21 @@ def as_openapi_params(cls) -> list[OpenApiParameter]:
         # FIXME: this should be solved as an extension instead, but at least it's now
         # kept together
         instance = cls()
-        ct_field = cast(serializers.CharField, instance.fields["component_type"])
+        ct_field = instance.fields["component_type"]
+        # assert isinstance(ct_field, serializers.CharField)
```
vaszig: Is this on purpose here?

PR #5789, opened 2025-12-02
Communication preferences prefill configuration constraints

### LINKED ISSUE #5762: Restrict varialbe selection only for profile component vars
### Thema / Theme

Frontend

We only want variables of profile components, right? So can you extend this filter to also check the component type? (I suspect `VariableSelection` will need to be updated to make this possible)

_Originally posted by @robinmolen in https://github.com/open-formulieren/open-forms/pull/5712#discussion_r2541376816_
            

_No response_

_No response_

robinmolen:
Partly closes #5762

**Changes**

Each communication preferences prefill configuration needs a unique combination of form and component. Meaning: each communication preferences prefill plugin, on the same form, needs to target a different profile component.

The uniqueness of the communication preferences prefill plugin is covered in the form variable serializers, and on database level. The database level validation ensures a situation where multiple form variables implement the communication preferences prefill using the same configuration can never happen. The serializer validation provides a nice early validation, which also provides the frontend with a descriptive error message (without the serializer validation, we would fallback to the generic "something went wrong" error message)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/forms/api/serializers/form_variable.py
+++ b/src/openforms/forms/api/serializers/form_variable.py
@@ -8,6 +8,9 @@ from rest_framework.exceptions import ValidationError
 
 from openforms.api.fields import RelatedFieldFromContext
 from openforms.api.serializers import ListWithChildSerializer
+from openforms.prefill.contrib.customer_interactions.constants import (
+    PLUGIN_IDENTIFIER as COMMUNICATION_PREFERENCES_PLUGIN_IDENTIFIER,
+)
 from openforms.prefill.registry import register
 from openforms.variables.api.serializers import ServiceFetchConfigurationSerializer
 from openforms.variables.constants import FormVariableSources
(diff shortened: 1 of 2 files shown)


src/openforms/forms/api/serializers/form_variable.py
```diff
@@ -92,11 +96,70 @@ def validate(self, attrs):
...
+                if not valid:
+                    errors[f"{index}.prefill_options.profile_form_variable"].append(
+                        serializers.ErrorDetail(
+                            _(
+                                "This profile form variable is already used in another communication preferences prefill plugin."
+                            ),
+                            code="invalid",
+                        )
+                    )
+
```
annashamray: I think this check can be done easier.
And now you don't need to have another static method

```suggestion

            if prefill_plugin != COMMUNICATION_PREFERENCES_PLUGIN_IDENTIFIER:
                continue

            profile_form_variable = prefill_options.get("profile_form_variable")
            if profile_form_variable in existing_profile_form_vars:
                errors[f"{index}.prefill_options.profile_form_variable"].append(
                    serializers.ErrorDetail(
                        _(
                            "This profile form variable is already used in another communication preferences prefill plugin."
                        ),
                        code="invalid",
                    )
                )

```

src/openforms/forms/tests/variables/test_viewset.py
```diff
@@ -1320,3 +1326,272 @@ def test_api_call_returns_all_variables(self):
...
+                        "name": "Second profile component",
+                        "digitalAddressTypes": ["phoneNumber"],
+                        "shouldUpdateCustomerData": True,
+                    },
+                ]
+            },
+        )
+        form_definition = form_step.form_definition
+        customer_interactions_api = CustomerInteractionsAPIGroupConfigFactory.create(
+            for_test_docker_compose=True
```
annashamray: this trait is used only when you need to request the Open Klant API (which would be accessible via docker compose), here it's not necessary.

src/openforms/prefill/contrib/customer_interactions/tests/test_plugin.py
```diff
@@ -283,3 +288,149 @@ def test_prefill_values_no_channel(self):
         state = submission.load_submission_value_variables_state()
 
         self.assertEqual(state.variables["empty-preferences"].value, {})
+
+    def test_create_multiple_prefill_variables_for_same_component(self):
```
annashamray: All these tests below are testing the new DB constraint for `FormVariable` model, right? I think they belong to `src/openforms/forms/test/variables/test_model.py`

Issue #5735, opened 2025-11-07
Logic rules are breaking submissions


LaurensBurger:
### Product versie / Product version

3.2.x

### Customer reference

Utrecht 413

### Omschrijf het probleem / Describe the bug

It seems we introduced a scenario where logic rules that involve variables that are "skipped" (step not applicable, or hidden) are breaking pdf generation, thus also registration.
example sentry: 453476, 453603

[2025-11-04 11:30:32 CET] (Inzending 9968): PDF generatietaak mislukt.

In this scenario the user skipped a step, but a logic rule is triggered involving variables from this step. 
~~The logic rules involve floats that are all `null` when the step is skipped.~~
tested for both float, and string - both cause the same issue.

Also setting the step "non applicable" state by default and making it applicable with logic also resolves the issue

### Stappen om te reproduceren / Steps to reproduce

example of breakage in form: 796 on test
reverse applicable example: 909 on test

### Screen resolution

None

### Device

None

### OS

None

viktorvanwijk:
There are two things going wrong here:
1. This particular crash happens only if there are user-defined variables present in the form. At submission completion, we persist the user-defined variables using the `persist_user_defined_variables` function, which calls `SubmissionVariableValueManager.bulk_create_or_update_from_data`. This has the (unintended) side-effect that submission value variables are created in the database for fields that are part of a non-applicable step (probably also hidden fields). 

2. The `StepNotApplicableAction` also uses `.bulk_create_or_update_from_data` to remove existing submission value variables from the database. This also removes the variables from the state completely (against our policy to always have a submission value variable available), and in turn causes a key error when trying to convert a value to the Python domain during logic evaluation.

Note that if there is no user-defined variables, `.bulk_create_or_update_from_data` is never called, and the variables related to fields of non-applicable steps are never saved to the DB. This means the `StepNotApplicableAction` will never delete them. 

I think it might be worth taking a good look at `.bulk_create_or_update_from_data`, because it seems like there is some funny business going on

That being said, it would be best to skip _any_ logic triggers/actions that involve variables that are part of non-applicable steps. We need #2409 for that, though, because other triggers/actions might depend on these variables.

PR #5799, opened 2025-12-08
Refactor bulk creating submission value variables

### LINKED ISSUE #5747: Side-effects of creating/updating variables using the submission value variable manager
Follow up from #5735, where the following chain of events caused a crash:

> 1. This particular crash happens only if there are user-defined variables present in the form. At submission completion, we persist the user-defined variables using the `persist_user_defined_variables` function, which calls `SubmissionVariableValueManager.bulk_create_or_update_from_data`. This has the (unintended) side-effect that submission value variables are created in the database for fields that are part of a non-applicable step (probably also hidden fields).
>
> 2. The `StepNotApplicableAction` also uses `.bulk_create_or_update_from_data` to remove existing submission value variables from the database. This also removes the variables from the state completely (against our policy to always have a submission value variable available), and in turn causes a key error when trying to convert a value to the Python domain during logic evaluation.
>
> Note that if there is no user-defined variables, `.bulk_create_or_update_from_data` is never called, and the variables related to fields of non-applicable steps are never saved to the DB. This means the `StepNotApplicableAction` will never delete them.

I think `.bulk_create_or_update_from_data` needs a revision, as it is used in multiple places with different goals it seems, which results in unwanted side-effects. 

Also maybe worth considering not removing the variables from the state, but set their primary keys to `None` for example, to ensure we always have all variables available (either from DB, or just in-memory for hidden fields or fields in non-applicable steps etc.)

viktorvanwijk:
Closes #5747

**Changes**

* Ensured variables that are not in `data` are not created/updated unexpectedly
* Ensured variables that are missing from `data` are not removed from the state, but rather reset to their unsaved version
* Reverted implemented bandaid fix for #5735 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/submissions/logic/rules.py
+++ b/src/openforms/submissions/logic/rules.py
@@ -155,7 +155,6 @@ def iter_evaluate_rules(
                     mutations_python = {
                         key: state.variables[key].to_python(value)
                         for key, value in mutations.items()
-                        if key in state.variables
                     }
                     data.update(mutations_python)
 


src/openforms/submissions/tests/form_logic/test_modify_variables.py
```diff
@@ -94,9 +94,8 @@ def test_modify_variable_related_to_step_being_edited(self):
 
         evaluate_form_logic(submission, submission_step2)
 
-        variables_state = submission.load_submission_value_variables_state()
-        variable = variables_state.variables["nTotalBoxes"]
-        self.assertEqual(7, variable.value)
+        state = submission.load_submission_value_variables_state()
+        self.assertEqual(state.get_data(include_unsaved=True)["nTotalBoxes"], 7)
```
annashamray: Why is this change needed? I thought that `variables_state.variables` includes unsaved vars, so the test should still work without the change, right?

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -215,10 +206,34 @@ def collect_variables(self) -> dict[str, SubmissionValueVariable]:
 
         return all_submission_variables
 
-    def remove_variables(self, keys: list) -> None:
+    def reset_variables(self, keys: list) -> None:
```
sergei-maertens:
```suggestion
    def reset_variables(self, keys: Collection[str]) -> None:
```

PR #5800, opened 2025-12-09
Communication preferences configurable portal url

### LINKED ISSUE #5795: Add `portalUrl` field to global configuration
### Thema / Theme

Frontend

The `portalUrl` will be used in the profile component, to direct users to the only portal where they can manage their digital address preferences

_No response_

_No response_

robinmolen:
Closes #5795

**Changes**

Adding a new configurable url field to the global configuration. This is used to guide users to an external portal where they can manage their communication preferences. This is used in the profile component, in conjunction with the klantinteracties API.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/forms/tests/test_api_forms.py
+++ b/src/openforms/forms/tests/test_api_forms.py
@@ -607,6 +607,29 @@ class FormsAPITests(APITestCase):
         self.assertEqual(step_literals_data["nextText"]["resolved"], "Next")
         self.assertEqual(step_literals_data["nextText"]["value"], "")
 
+    def test_global_config_communication_preferences_portal_url_value_returned_through_api(
+        self,
+    ):
+        config = GlobalConfiguration.get_solo()
+        config.communication_preferences_portal_url = "https://example-portal-url.com"
+        config.save()
+
+        form = FormFactory.create()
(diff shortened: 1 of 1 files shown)


src/openforms/config/models/config.py
```diff
@@ -576,6 +576,14 @@ class GlobalConfiguration(SingletonModel):
         max_length=100,
         blank=True,
     )
+    communication_preferences_portal_url = models.URLField(
+        _("communication preferences portal url"),
+        blank=True,
+        help_text=_(
+            "Url to the online portal where users can update their communication preferences. "
+            "This is used in conjunction with the communication preferences (customer interactions API) prefill in the profile component."
```
sergei-maertens:
```suggestion
            "This is used by the customer profile component."
```
- simpler language for accessibility 
- it's not used in/during prefill, only exposed in the form detail endpoint for the customer profile component

robinmolen: I meant it more as "this is only used when there are prefilled/fetched digital addresses", which happens using the communication preferences prefill.

But it does complicate things.. Changing the message

PR #5805, opened 2025-12-11
[#5598] prevent cosigning by initial form submitter

### LINKED ISSUE #5598: Never allow auth_bsn and co_signer.value to be identical
### Thema / Theme

Admin

Addition to #4459  

As a admin i want to prevent submissions to contain identical values for the submitter (auth_bsn) and the co_signer. 
OF should reject these attempts, because they are either made by mistake, which the user can then "fix" by using the right account. Or they might try to submit something for someone else, which should not be allowed/possible.

_No response_

_No response_

SonnyBA:
Closes #5598 

**Changes**

Prevents the original form submitter from cosigning the submission.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/authentication/utils.py
+++ b/src/openforms/authentication/utils.py
@@ -1,3 +1,6 @@
+from django.contrib.auth.hashers import check_password as check_salted_hash
+from django.http import HttpRequest
+
 import structlog
 from furl import furl
 from rest_framework.request import Request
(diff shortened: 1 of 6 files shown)


src/openforms/authentication/utils.py
```diff
@@ -106,6 +106,30 @@ def get_cosign_login_url(request: Request, form: Form, plugin_id: str) -> str:
...
+
+    if is_same_login_value(submission, request) and is_same_login_attribute(
+        submission, request
+    ):
+        logger.info(
+            "cosign_not_allowed",
+            reason="cannot_sign_own_cosign_request",
+            submission_uuid=submission.uuid,
+        )
+        return True
```
SonnyBA: This currently only verifies that the original form submitter is not the same person trying to cosign the submission by checking if the login method used (DigiD/eHerkenning) and the saved value does not match with what is used in the session during cosigning.

This does not (at the moment):
- Verify that whenever another login method is used (for example eHerkenning), compared to the original submission (for example DigiD), that both values match
- Verify that the user trying to cosign, has the same emailadres as was given during the form submission (see second last point from https://github.com/open-formulieren/open-forms/issues/4459#issuecomment-2612168297)

src/openforms/submissions/api/viewsets.py
```diff
@@ -246,6 +247,11 @@ def cosign(self, request, *args, **kwargs):
                 )
             )
 
+        if same_cosigner(request, submission):
+            raise PermissionDenied(
+                _("Submission cannot be co-signed by the initial submitter.")
+            )
+
```
SonnyBA: Discussed today, this validation can be moved to the `SearchSubmissionForCosignFormView` so that the validation takes place at an earlier point. 

Using an assert statement here will be sufficient so that it can be picked up during testing (but will be stripped in live environments).

src/openforms/submissions/forms.py
```diff
@@ -35,4 +39,10 @@ def clean(self):
                 )
             )
         self.cleaned_data["submission"] = submission
+
+        if same_cosigner(self.auth_attribute, self.auth_value, submission):
+            raise ValidationError(
+                _("Submission cannot be co-signed by the initial submitter.")
```
sergei-maertens:
```suggestion
                _("The submission cannot be co-signed by the original submitter.")
```

src/openforms/submissions/views.py
```diff
@@ -110,29 +114,11 @@ def is_auth_data_correct(self, submission: Submission) -> bool:
...
-            # timing attacks are not a concern here, as you need to go through a valid
-            # authentication flow first which sets the value in the session. Additionally,
-            # enumeration would still be possible and you don't necessarily leak passwords
-            # that may be used on other sites - end-users cannot change their BSN/KVK etc.
-            # anyway.
-            is_auth_data_correct = submission_auth_value == current_auth_value
+        auth_attribute = self.request.session[FORM_AUTH_SESSION_KEY]["attribute"]
+        auth_value = self.request.session[FORM_AUTH_SESSION_KEY]["value"]
+        is_auth_attribute_correct = is_same_login_attribute(submission, auth_attribute)
+        is_auth_data_correct = is_same_login_value(submission, auth_value)
```
sergei-maertens: I don't think these need to be separate functions, ultimately what we want to know is:

```py
is_same_user = check_user_matches_submission(self.request, submission)
```
with `check_user_matches_submission` imported from `openforms.authentication.service`

which turns this methods return statement into:

```py
return is_auth_plugin_correct and is_same_user
```

src/openforms/submissions/utils.py
sergei-maertens: all the changes here can be reverted if you implement `check_user_matches_submission` as suggested :)

src/openforms/authentication/utils.py
```diff
@@ -106,6 +106,30 @@ def get_cosign_login_url(request: Request, form: Form, plugin_id: str) -> str:
...
 
+def same_cosigner(auth_attribute: str, auth_value: str, submission: Submission) -> bool:
+    from openforms.submissions.utils import is_same_login_attribute, is_same_login_value
+
+    if not submission.is_authenticated:
+        logger.info(
+            "cosign_not_allowed",
+            reason="missing_auth_info",
+            submission_uuid=submission.uuid,
+        )
```
sergei-maertens: This logging line can move to the form validation.

Also, make sure you stringify the UUID otherwise you get the `repr` in logs:

```suggestion
        logger.info(
            "cosign_start_blocked",
            reason="cosigner_same_as_submitter",
            submission_uuid=str(submission.uuid),
        )
```

I've also changed the event to `cosign_start_blocked`, which is typically the form you want to use for the event name - describe the observable effect as the end-user experiences it, if possible.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #5810, opened 2025-12-12
[#5776] Extend "mapping" of variables in Stuf-ZDS registration

### LINKED ISSUE #5776: Extend "mapping" of variables in Stuf-ZDS registration
### Thema / Theme

Form designer

(Alternative to #5775)

As form designer i want to map variables to certain specific "attributes" for context in de registration system. 
While it is possible to set these key's in regular components, we cannot set these in the variables tab, and these specific keys: "casespecific.code_variable" also cause some unwanted behavior in the forms. (especially with groups of these fields and backend logic)

Currently there is already a "mapping" option for payment updates. 
This seems like a suitable place to also being able to use this this more "generic" for the creeerzaak. 
This will allow a form builder to build a form with no need for specific key's, simply add the specific attributes to this "mapping" and match the components/variables from the form.

_No response_

_No response_

vaszig:
Closes #5776

**Changes**

Refactored dropdown of the variables in the extraElementen table. The user could only select the static variables for changing/defining their names. Now they can choose the user defined variables too.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/forms/migrations/0117_stuf_zds_variables_mapping.py
```diff
@@ -0,0 +1,43 @@
...
+        (
+            "forms",
+            "0116_formvariable_unique_form_id_and_profile_form_variable_for_prefill_plugin_communication_preferences",
+        ),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            copy_payment_variables_to_variables_mapping,
+            reverse_code=migrations.RunPython.noop,
```
sergei-maertens: the reverse operation shouldn't be hard to implement :)

PR #5811, opened 2025-12-12
5711 write to open klant (other flows)

### LINKED ISSUE #5711: Updating data to klanten service
### Thema / Theme

Frontend

Update customer, and digital address data to "Klantinteracties" service after form submission.

**Enabled/disabled**

The "profile" component contains a "Should Open Forms update user preferences in 'Klantinteracties' system" configuration field. Only when this is set to `true`, should the following functionality be performed.

**Where in the backend happens**

As part of the `on_post_submission_event` flow in Open Forms, the new/updated data should be persisted to the connected "Klantinteracties" service.

The task should be parallel to registration tasks, so if it fails the registration tasks could still be finished.
If the task failed the retry task should be triggered. The logging event should be also issued.
In case the task failed, the status and the number of affected submissions should be visible in sentry. 

**Implementation**

Discussed today (6th of November) during standup. The technical implementation of processing the submitted data to an external service should NOT be klantinteracties API specific. In the future there might be more plugins/functionalities that would process data in a similar way.

We should:
1. Design a generic implementation, that leaves room for future expansions/re-use
2. Discuss the implementation design with Sergei, and possible polish it
3. Implement the generic setup
4. Implement data processing for klantinteracties API

**Upd - registration charts**

As is:

Image

Possible implementation:
* add new module into OF (Side effect plugins?), which allow to write into external API after the form is submitted, in parallel to registration process. 
* all of the plugins would be run in parallel to each other and independent of each other

Image

Update after discussion:

* not making a separate plugin registry, use formio registry
* run group of tasks before pre_registration as much as possible
* if needed run update/patch after pre-registration

**Upd. Status and processing errors:**

What needs to be refined:
1.  how to see the status and results of the tasks
2. show we retry failed tasks

**Possible options**
1. use `Submission` model:
   * add `component_pre_registration_completed` boolean which will show if all component tasks are completed successfully
   * reuse `registration_result` variable to store possible exceptions (like with payments) and/or results
   * use `submission.needs_on_completion_retry` if any of component task fail

2. use `SubmissionValueVariable` model, because tasks are used per component:
   * add `pre_registration_hook_status` (enum)
   * add `pre_registration_hook_result` (JSON) to store possible exceptions or results
   * if any of component task fail we can still update `submission.needs_on_completion_retry = True` to reuse the general retry process

3. create new model `ComponentHook` model with FK to `SubmissionValueVariable` model, which will contain all the hook result and statuses variables, so that SubmissionValueVariable` will be left unchanged 

Discussed:
we'll use 2nd option but take a look how `Submission.registration_result` is used and should we write there 

### What to write

**Not depend on flow:**
* we always create new ``betrokkene``, ``klantcontact`` and ``OnderwerpObject `` in Open Klant (POST `/api/v1/maak-klantcontact`), 
* save `submission.public_reference` in the ``OnderwerpObject``

**Depends on flow:**
* interaction with ``partij``resource
* creating digital addresses
* linking digital addresses to ``partij``

There are several available flows:

1. _Anonymous user provides email address and/or phone number. They verify it, and continue the form._  

   We create the digital address (POST `/api/v1/digitaleadressen`), and use the uuid of the previously created ``betrokkene`` for the `verstrektDoorBetrokkene` property.

2. _Authenticated user isn't known in Open Klant, so no "digitaal adres" can be fetched. The user provides a new email address and/or phone number. They verify it,_
  
   We create a new ``Partij``  (POST `/api/v1/partijen`). and use its uuid for `betrokkene.wasPartij.uuid` while creating betrokkene 

   2a.  _decide to "use it once" and continue_
        Next we create the digital address (POST `/api/v1/digitaleadressen`) and **DON'T** use `Partij` uuid for the `verstrektDoorPartij` property. We only use `verstrektDoorBetrokkene` here

   2b.  _decide to "as new preferred" and continue_
         Next we create the digital address (POST `/api/v1/digitaleadressen`) and use `Partij` uuid for the `verstrektDoorPartij` property.

3. _Authenticated user is known in Open Klant, they have one known "digitaal adres", which is fetched. They use their defined "digitaal adres" and continue the form._ 
  
     No extra requests after (POST `/api/v1/maak-klantcontact`)

4. _Authenticated user is known in Open Klant, they have one known "digitaal adres", which is fetched. They define another email address and/or phone number. They verify it,_ 

   4a.  _decide to "use it once" and continue_
       We create the digital address (POST `/api/v1/digitaleadressen`) and **DON'T** use `Partij` uuid for the `verstrektDoorPartij` property. We only use `verstrektDoorBetrokkene` here

   4b.  _decide to "as new preferred" and continue_
         We create the digital address (POST `/api/v1/digitaleadressen`) and use `Partij` uuid for the `verstrektDoorPartij` property

5. _Authenticated user is known in Open Klant, multiple "digitale adressen" are fetched. They use their preferred "digitaal adres", or select another verified "digitaal adres", and continue the form._

   same as **3.** flow

6. Authenticated user is known in Open Klant, multiple "digitale adressen" are fetched. They define a new email address and/or phone number, verify it, decide to "use it once, or as new preferred" and then continue.
   
   Same as **4.** flow

7. Authenticated user is known in Open Klant, multiple "digitale adressen" are fetched. They define a new email address and/or phone number, verify it, decide to "use it once, or as new preferred" and then continue. In the backend validation 
we notice that the new digital address is already known as a valid digital address of this user.

   We don't create a new digital address, because we don't want to pollute the communication preferences data with duplicate records. Instead, we use the information of the already existing digital address, as if this was selected by the user.
   If the preference of this "new" digital address was set to "use as new preferred", we update the preferences. If the user decided that this digital address should be used for future forms, we should listen to that request.

   It looks like subtype of **6.** flow. I think OK validation should be able to handle it. We can add a separate test for it? 

**Comments**
For authenticated users we need to fetch their "Partij" data, using the GET `/partijen` and the `partijIdentificator__*` parameters. The "Partij" data can be used when creating the "klantinteractie" instance.

To create the "klantinteractie" instance, the `/maak-klantcontact` endpoint can be used. For adding/updating the digital addresses, the POST `/digitaleadressen` and PATCH `/digitaleadressen/{uuid}` endpoints can be used.

This sequence should be followed: https://platformvoordienstverlening.notion.site/Klantinteractie-patronen-1f378b5f4db480b19384e535925f1724#1f378b5f4db480b19384e535925f1724

_No response_

_No response_

annashamray:
Closes #5711 (other flows)

**Changes**
Follow-up of https://github.com/open-formulieren/open-forms/pull/5796 (should be merged after the #5796 )

Adds writing to Open Klant when:
* user is authenticated with BSN in the form, but is unknown in Open Klant
* user is authenticated with BSN in the form and is found in Open Klant and uses addresses from prefill
* user is authenticated with BSN in the form and is found in Open Klant and adds new addresses

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/contrib/customer_interactions/client.py
+++ b/src/openforms/contrib/customer_interactions/client.py
@@ -4,6 +4,8 @@ from collections.abc import Iterator
 from typing import assert_never
 
 from openklant_client.client import OpenKlantClient
+from openklant_client.types.common import ForeignKeyRef
+from openklant_client.types.iso_639_2 import LanguageCode
 from openklant_client.types.methods.maak_klant_contact import (
     MaakKlantContactCreateData,
     MaakKlantContactResponse,
@@ -66,19 +68,20 @@ class CustomerInteractionsClient(LoggingMixin, OpenKlantClient):
         }
         response = self.digitaal_adres.list_iter(params=params)
 
-        address = next(response)
-        return address
+        digitaal_adres = next(response)
+        return digitaal_adres
 
     def create_customer_contact(
         self,
         submission: Submission,
         party_uuid: str = "",
     ) -> MaakKlantContactResponse:
+        language_code: LanguageCode = to_iso639_2b(submission.language_code)  # pyright: ignore[reportAssignmentType]
         data: MaakKlantContactCreateData = {
             "klantcontact": {
                 "kanaal": "Webformulier",
                 "onderwerp": submission.form.name,
-                "taal": to_iso639_2b(submission.language_code),
+                "taal": language_code,
                 "vertrouwelijk": True,
             },
             "betrokkene": {
@@ -108,13 +111,14 @@ class CustomerInteractionsClient(LoggingMixin, OpenKlantClient):
         betrokkene_uuid: str,
         party_uuid="",
     ) -> DigitaalAdres:
-        party_data = {"uuid": party_uuid} if party_uuid else None
+        party_data: ForeignKeyRef | None = {"uuid": party_uuid} if party_uuid else None
         data = DigitaalAdresCreateData(
             adres=address,
             soortDigitaalAdres=address_type,
             isStandaardAdres=is_preferred,
             verstrektDoorBetrokkene={"uuid": betrokkene_uuid},
             verstrektDoorPartij=party_data,
+            omschrijving="",
         )
         return self.digitaal_adres.create(data=data)
 
@@ -200,6 +204,13 @@ class CustomerInteractionsClient(LoggingMixin, OpenKlantClient):
                     created = True
 
                 return party, created
+
+            case (
+                AuthAttribute.kvk | AuthAttribute.pseudo | AuthAttribute.employee_id
+            ):  # pragma: no cover
+                raise NotImplementedError(
+                    "Only bsn authentication is supported for Customer Interactions API"
+                )
 
             case _:  # pragma: no cover
                 assert_never(auth_attribute)
--- a/src/openforms/contrib/customer_interactions/update.py
+++ b/src/openforms/contrib/customer_interactions/update.py
@@ -1,12 +1,17 @@
+from collections.abc import Mapping
 from typing import TypedDict
 
 import structlog
 from openklant_client.types.methods.maak_klant_contact import MaakKlantContactResponse
 from openklant_client.types.resources.betrokkene import Betrokkene
-from openklant_client.types.resources.digitaal_adres import DigitaalAdres
+from openklant_client.types.resources.digitaal_adres import (
+    DigitaalAdres,
+    SoortDigitaalAdres,
+)
 from openklant_client.types.resources.klant_contact import KlantContact
 from openklant_client.types.resources.onderwerp_object import OnderwerpObject
 
+from openforms.authentication.constants import AuthAttribute
 from openforms.formio.typing.custom import DigitalAddress, SupportedChannels
 from openforms.prefill.contrib.customer_interactions.typing import CommunicationChannel
 from openforms.prefill.contrib.customer_interactions.variables import (
@@ -83,7 +88,7 @@ def update_customer_interaction_data(
 
     # submission profile data
     state = submission.load_submission_value_variables_state()
-    profile_submission_data: list[DigitalAddress] = state.get_data()[profile_key]
+    profile_submission_data: list[DigitalAddress] = state.get_data()[profile_key]  # pyright: ignore[reportAssignmentType]
 
     # prefill config
     prefill_form_variable = fetch_user_variable_from_profile_component(
@@ -93,7 +98,9 @@ def update_customer_interaction_data(
         logger.info("missing_prefill_variable", component=profile_key)
         return
 
-    prefill_value = state.get_data(include_unsaved=True)[prefill_form_variable.key]
+    prefill_value: list[CommunicationChannel] = state.get_data(include_unsaved=True)[
+        prefill_form_variable.key
+    ]  # pyright: ignore[reportAssignmentType]
 
     plugin = prefill_registry[prefill_form_variable.prefill_plugin]
     options_serializer = plugin.options(data=prefill_form_variable.prefill_options)
@@ -101,12 +108,14 @@ def update_customer_interaction_data(
     plugin_options = options_serializer.validated_data
     api_group = plugin_options["customer_interactions_api_group"]
 
-    channels_to_address_types = {v: k for k, v in ADDRESS_TYPES_TO_CHANNELS.items()}
+    channels_to_address_types: Mapping[SupportedChannels, SoortDigitaalAdres] = {
+        v: k for k, v in ADDRESS_TYPES_TO_CHANNELS.items()
+    }
 
     with get_customer_interactions_client(api_group) as client:
         if submission.is_authenticated:
             auth_value = submission.auth_info.value
-            auth_attribute = submission.auth_info.attribute
+            auth_attribute: AuthAttribute = submission.auth_info.attribute
 
             # link authenticated user to the party
             party, created = client.get_or_create_party(auth_attribute, auth_value)
@@ -117,6 +126,8 @@ def update_customer_interaction_data(
         customer_contact: MaakKlantContactResponse = client.create_customer_contact(
             submission, party_uuid
         )
+        assert customer_contact["betrokkene"] is not None
+        assert customer_contact["onderwerpobject"] is not None
 
         created_addresses: list[DigitaalAdres] = []
         updated_addresses: list[DigitaalAdres] = []
@@ -156,7 +167,7 @@ def update_customer_interaction_data(
                     address=address_value,
                     address_type=channels_to_address_types[address_channel],
                     betrokkene_uuid=customer_contact["betrokkene"]["uuid"],
-                    party_uuid=party_uuid if is_address_new_preferred else None,
+                    party_uuid=party_uuid if is_address_new_preferred else "",
                     is_preferred=is_address_new_preferred,
                 )
                 created_addresses.append(created_address)
--- a/pyright.pyproject.toml
+++ b/pyright.pyproject.toml
@@ -89,6 +89,7 @@ exclude = [
     "src/openforms/authentication/contrib/digid/tests/test_auth_procedure.py",
     "src/openforms/authentication/contrib/digid/tests/test_signicat_integration.py",
     "src/openforms/authentication/contrib/eherkenning/tests/test_signicat_integration.py",
+    "src/openforms/contrib/customer_interactions/tests/",
     "src/openforms/contrib/objects_api/tests/",
     "src/openforms/contrib/objects_api/json_schema.py",
     "src/openforms/formio/formatters/tests/",


src/openforms/contrib/customer_interactions/utils.py
```diff
@@ -0,0 +1,165 @@
...
+    result = {}
+    with get_customer_interactions_client(api_group) as client:
+        if bsn:
+            # link authenticated user to the party
+            party = client.find_party_for_bsn(bsn)
+            if not party:
+                party = client.create_party_for_bsn(bsn)
+                result["partij"] = party
+
+            party_uuid = party["uuid"]
```
robinmolen: Should `partij` not be included if it already exists?

```suggestion
        if bsn:
            # link authenticated user to the party
            party = client.find_party_for_bsn(bsn)
            if not party:
                party = client.create_party_for_bsn(bsn)
            
            result["partij"] = party
            party_uuid = party["uuid"]
```

Or is `result` only the newly created/updated data?

src/openforms/contrib/customer_interactions/utils.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            party_uuid = None
+
+        maak_klant_contact: MaakKlantContactResponse = client.create_customer_contact(
```
robinmolen:
```suggestion
        create_customer_contact_response: MaakKlantContactResponse = client.create_customer_contact(
```
Or simply `customer_contact`

src/openforms/contrib/customer_interactions/utils.py
```diff
@@ -0,0 +1,165 @@
...
+                else None,
+            )
+            created_addresses.append(created_address)
+
+        result.update(
+            {
+                "klantcontact": maak_klant_contact["klantcontact"],
+                "betrokkene": maak_klant_contact["betrokkene"],
+                "onderwerpobject": maak_klant_contact["onderwerpobject"],
+                "digital_addresses": created_addresses,
```
robinmolen: If we only list the digital_addresses that were created, than maybe a name like `created_digital_addresses` would be better?

src/openforms/contrib/customer_interactions/client.py
```diff
@@ -70,17 +79,74 @@ def create_customer_contact(
...
     ):
+        party_data = {"uuid": party_uuid} if party_uuid else None
         data = DigitaalAdresCreateData(
             adres=address,
             soortDigitaalAdres=address_type,
             omschrijving="Open Forms profile",
             verstrektDoorBetrokkene={"uuid": betrokkene_uuid},
-            verstrektDoorPartij=None,
-            # isStandaardAdres=is_preferred,
+            verstrektDoorPartij=party_data,
```
robinmolen: Shouldn't `verstrektDoorPartij` be included in this PR?

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
```diff
@@ -0,0 +1,415 @@
...
+                "verstrektDoorBetrokkene": {
+                    "url": betrokkene["url"],
+                    "uuid": betrokkene["uuid"],
+                },
+                "verstrektDoorPartij": None,
+            },
+            {
+                "adres": "0612345678",
+                "soortDigitaalAdres": "telefoonnummer",
+                "isStandaardAdres": False,
```
robinmolen: The profile data contains a phoneNumber with `"preferenceUpdate": "isNewPreferred"`. The `isStandaardAdres` should be `True`

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
```diff
@@ -0,0 +1,415 @@
...
+                "verstrektDoorBetrokkene": {
+                    "url": betrokkene["url"],
+                    "uuid": betrokkene["uuid"],
+                },
+                "verstrektDoorPartij": None,
+            },
+            {
+                "adres": "0611111111",
+                "soortDigitaalAdres": "telefoonnummer",
+                "isStandaardAdres": False,
```
robinmolen: Should be `True` as the profile data phoneNumber states `"preferenceUpdate": "isNewPreferred"`

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
```diff
@@ -0,0 +1,415 @@
...
+            },
+        ]
+        for expected_address in expected_addresses:
+            with self.subTest(expected_address):
+                self.assertAddressPresent(digital_addresses, expected_address)
+
+    def test_auth_user_known_in_openklant_known_addresses(self):
+        profile_channels: list[SupportedChannels] = ["email", "phoneNumber"]
+        profile_data: list[DigitalAddress] = [
+            {
```
robinmolen: Could you add a comment which digital address is already known. Or are they both already known/registered?

src/openforms/contrib/customer_interactions/utils.py
annashamray: I thought to include only new data there. We can include the existing partij, but maybe using another name (like "existing_partij") ? 
Should I include it?

src/openforms/contrib/customer_interactions/utils.py
```diff
@@ -0,0 +1,165 @@
...
+            party_uuid = None
+
+        maak_klant_contact: MaakKlantContactResponse = client.create_customer_contact(
+            submission, party_uuid
+        )
+
+        created_addresses = []
+        for digital_address in profile_submission_data:
+            address_channel: SupportedChannels = digital_address["type"]
+            prefill_address: CommunicationChannel | None = next(
```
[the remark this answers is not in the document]
annashamray: You are right, it's confusing. Renamed to `prefill_communication_channel`

src/openforms/contrib/customer_interactions/utils.py
```diff
@@ -0,0 +1,165 @@
...
+                (value for value in prefill_value if value["type"] == address_channel),
+                None,
+            )
+            prefill_options = prefill_address["options"] if prefill_address else []
+            if digital_address["address"] in prefill_options:
+                # used value from prefill, so it already exists in Open Klant
+                continue
```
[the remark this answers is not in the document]
annashamray: Let's discuss it! 

I thought that we don't change the preference for the address in the Open Klant: isn't it the reason, why we show the user link to Open Klant with the text "You can edit your preferences in the online portal"? 
This is why in the flow 5 we don't do any requests to digital addresses. 
This is how I understood the requirements. 

Do you think that if the user uses the existing address and marks it "isNewPreferred" than we need to update preferences in the Open Klant, i.e. add PATCH request for this address and make it preferred (`isStandaardAdres = True`)

src/openforms/contrib/customer_interactions/utils.py
annashamray: renamed to `customer_contact`

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
annashamray: Yes, they are both are in the Open Klant fixture. I'll add a comment

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
```diff
@@ -0,0 +1,415 @@
...
+                    "label": "Profile",
+                    "digitalAddressTypes": profile_channels,
+                    "shouldUpdateCustomerData": True,
+                }
+            ],
+            submitted_data={
+                "profile": profile_data,
+            },
+        )
+        result = update_customer_interaction_data(submission, "profile")
```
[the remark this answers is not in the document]
annashamray: I think we discussed it [here](https://github.com/open-formulieren/open-forms/pull/5796#discussion_r2620020350) 
For now we just add logging, and I've also created an issue for adding a check in email digest for it (https://github.com/open-formulieren/open-forms/issues/5831)
In this case it would be visible.  Do you think we need also to raise en exception?

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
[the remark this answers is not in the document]
annashamray: Test case is added, but the preferences are not updated

src/openforms/contrib/customer_interactions/tests/test_update_customer_interaction_data.py
```diff
@@ -69,25 +72,407 @@ def test_non_auth_user(self):
...
+
+        # no new address is added
+        self.assertEqual(digital_addresses, [])
+
+    def test_auth_user_known_in_openklant_known_addresses_with_different_preference(
+        self,
+    ):
+        """
+        even if we change the preference for the known address - we don't update it
+        """
```
sergei-maertens: as discussed before - this seems an incorrect interpretation of the requirements, but feel free to double check with Joeri.

src/openforms/contrib/customer_interactions/client.py
```diff
@@ -72,15 +81,92 @@ def create_customer_contact(
...
-    def create_digital_address_for_betrokkene(
-        self, address: str, address_type: SoortDigitaalAdres, betrokkene_uuid: str
-    ):
+    def create_digital_address(
+        self,
+        address: str,
+        address_type: SoortDigitaalAdres,
+        is_preferred: bool,
+        betrokkene_uuid: str,
+        party_uuid: str | None = None,
```
sergei-maertens: I'm not a fan of `str | None` unions, unless the empty string doesn't mean that processing can ignored:


```suggestion
        party_uuid: str = "",
```
simplicity in types usually means simpler code :D

src/openforms/contrib/customer_interactions/update.py
```diff
@@ -72,30 +89,61 @@ def update_customer_interaction_data(
 
     channels_to_address_types = {v: k for k, v in ADDRESS_TYPES_TO_CHANNELS.items()}
 
+    result: UpdateCustomerInteractionsResult = {}
```
sergei-maertens: How does this type check if the keys are `Required` by default 🤔 ?

sergei-maertens (changes requested):

src/openforms/contrib/customer_interactions/client.py
annashamray: changed

src/openforms/contrib/customer_interactions/update.py
```diff
@@ -72,30 +89,61 @@ def update_customer_interaction_data(
...
-                created_addresses.append(created_address)
-
-            result = {
-                "klantcontact": maak_klant_contact["klantcontact"],
-                "betrokkene": maak_klant_contact["betrokkene"],
-                "onderwerpobject": maak_klant_contact["onderwerpobject"],
+            if digital_address["address"] in prefill_channel_options:
+                # used value from prefill, so it already exists in Open Klant
+                # we don't update existing addresses
+                continue
```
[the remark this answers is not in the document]
annashamray: PATCH is added

src/openforms/contrib/customer_interactions/client.py
```diff
@@ -72,15 +81,92 @@ def create_customer_contact(
...
+            raise StandardViolation(
+                "Combination of 'codeObjecttype', 'codeSoortObjectId', 'objectId' and 'codeRegister' "
+                "must be unique according to the standards."
+            )
+        if num_results == 0:
+            return None
+
+        return response["results"][0]
+
+    def find_party(self, auth_value: str, auth_attribute: str) -> Partij | None:
```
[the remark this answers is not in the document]
annashamray: added `AuthAttribute` and `assert_never` to `get_or_create_party` method

src/openforms/contrib/customer_interactions/update.py
```diff
@@ -72,30 +89,61 @@ def update_customer_interaction_data(
...
-        if not submission.is_authenticated:
-            # 1. User is not authenticated
-            maak_klant_contact: MaakKlantContactResponse = (
-                client.create_customer_contact(submission)
+        if submission.is_authenticated:
+            auth_value = submission.auth_info.value
+            auth_attribute = submission.auth_info.attribute
+
+            # link authenticated user to the party
+            party = client.find_party(auth_value, auth_attribute)
```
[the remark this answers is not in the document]
annashamray: `find_party` is replaced with `get_or_create_party`, where I changed the order of arguments

src/openforms/contrib/customer_interactions/update.py
annashamray: included it into `pyright.pyproject.toml`

PR #5863, opened 2025-12-24
Logic rules dependency graph

### LINKED ISSUE #2409: The logic evaluation should be staggered and based on a dependency graph
From the analysis of #2340 it's clear that data mutates the formio configuration and vice versa. This leads to potentially incorrect states when data needs to be mutated depending on certain formio component properties and additional rules to be evaluated can be affected by this.

Ideally, the logic rules should be (statically) analysed for dependencies and the mutations/operations should be executed following these dependencies. E.g. the result of an operation/action can be input for another operation, so that second operation needs to be applied _after_ the first.

The dependency must be an directed acyclic graph to guarantee that execution completes/is finite.

This rework has a number of benefits:

* it allows us to represent this dependency tree and order of exection in an UML-like presentation, which is highly requested by form designers
* we should be able to detect cycles and mark those combinations of rules/actions as invalid
* we don't need infinite loops to reach a resolved (data + configuration) state, but instead a finite loop over each layer of depth in the graph
* the amount of queries remains a fixed number, so performance should not suffer

viktorvanwijk:
Partly closes #2409

~~Dependent on #5936~~

[skip: e2e]

**Changes**

* Added routines to build and analyze a dependency graph
* Added analysis of logic rules when saving them (behind a form feature flag)
* Added fetching which logic rules to execute from the step (behind a form feature flag)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/forms/models/form.py
```diff
@@ -397,6 +397,16 @@ class Form(models.Model):
...
+    logic_rule_analysis_enabled = models.BooleanField(
+        _("enable logic rule analysis"),
+        default=False,
+        help_text=_(
+            "Enabling this will analyze logic rules and re-order them according to "
+            "their dependency on other logic rules (happens when the form is saved). "
+            "Each rule will be automatically assigned to one or more steps on which "
+            "it will be executed."
+        ),
+    )
```
sergei-maertens: should we widen the scope of this feature flag to opt into the new logic evaluation entirely? Or, do we:

* expect that we cannot gate the client-side logic evaluation behind a feature flag (initially) or
* expect that we need to decouple rule analysis from client-side logic evaluation?

If possible, I'd like to avoid a situation where two different feature flags need to be setup to use the new logic evaluation, or misconfigurations are possible where the feature flags are mutually exclusive and require validation to prevent that. If we can hang up the entire "new logic evaluation for performance" thing on a single feature flag, that'd be ideal, but then I'd probably suggest renaming this feature flag to something that covers the scope better.

Of course, this all depends on how far we can go with the feature flags, which I can't really tell yet :)

src/openforms/forms/api/serializers/logic/form_logic.py
```diff
@@ -21,6 +30,72 @@ class FormLogicListSerializer(ListWithChildSerializer):
...
+        # Note: model instances without a pk are not hashable, which is a requirement to
+        # use them in the graph, so we assign it manually. These rules will just live in
+        # memory, so it will not result in conflicts with existing rules.
+        rules = [FormLogic(**rule_data, pk=i) for i, rule_data in enumerate(data)]
+
+        graph = create_graph(rules)
+        cycles = detect_cycles(graph)
+        if cycles:
+            msg = _("Rule contains cycles through variable(s): {variables}.")
+            errors = defaultdict(list)
```
sergei-maertens:
```suggestion
            errors: defaultdict[str, list[ErrorDetail]] = defaultdict(list)
```

src/openforms/forms/api/serializers/logic/form_logic.py
```diff
@@ -21,6 +30,72 @@ class FormLogicListSerializer(ListWithChildSerializer):
...
+
+        # Form steps are added to the context dictionary from the form-step set, which
+        # is an ordered model queryset, so we can just select the first value.
+        first_step = next(iter(self.context["form_steps"].values()))
+        assert first_step.order == 0
+        add_missing_steps(graph, first_step)
+        new_rule_order = resolve_order(graph)
+
+        # Reorder the incoming data according to the determined order.
+        data_new, steps = [], []
```
sergei-maertens: type hints please 😇

src/openforms/forms/tests/test_api_form_logic_bulk.py
```diff
@@ -1519,3 +1519,227 @@ def test_create_rule_with_evaluate_dmn_action_empty_config(self):
...
+            for description, rule in zip(expected_order, form_logic_qs, strict=True):
+                self.assertEqual(rule.description, description)
+
+        with self.subTest("Rules of step 1"):
+            self.assertEqual(
+                list(step_1.logic_rules.all()), [form_logic_qs[0], form_logic_qs[1]]
+            )
+
+        with self.subTest("Rules of step 2"):
+            self.assertEqual(list(step_2.logic_rules.all()), [form_logic_qs[2]])
```
sergei-maertens: here too I think `self.assertQuerySetEqual` would be nice

sergei-maertens (changes requested):
Nice work on the analysis and graph theory application!

The bulk of the suggestions aim to improve the readability of the code.

Questions and edge cases:

* logic rule analysis triggers on form save. What happens when someone edits a logic rule or form definition in the admin outside of the regular form-creation-form page?
    * form definitions can remove components -> they remove variables that may be used in logic rules
    * logic rule edits would result in an out-of-sync analysis result compared to the actual rule and lead to weird evaluation bugs

src/openforms/forms/models/form.py
viktorvanwijk: Yeah my intention was definitely a single feature flag for all the changes. I will rename it :)

src/openforms/forms/tests/test_logic_analysis.py
```diff
@@ -0,0 +1,627 @@
...
+                list(graph.edges),
+                [(rule_1, rule_2), (rule_3, rule_1), (rule_2, rule_3)],
+            )
+
+        with self.subTest("Cycles"):
+            cycles = detect_cycles(graph)
+            assert cycles is not None
+            self.assertEqual(len(cycles), 1)
+            cycle = list(cycles)[0]
+            self.assertCountEqual(set(cycle["rules"]), {rule_1, rule_2, rule_3})
```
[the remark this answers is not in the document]
viktorvanwijk: Ah, yeah, I think I had a list comparison before

src/openforms/forms/logic_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+from typing import TypedDict
+
+from networkx import DiGraph
+from networkx.algorithms import simple_cycles, topological_sort
+from networkx.algorithms.dag import lexicographical_topological_sort
+
+from .models import FormLogic, FormStep
+
+
+def create_graph(rules: Iterable[FormLogic]) -> DiGraph:
```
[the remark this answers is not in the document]
viktorvanwijk: Yeah this is what I wanted to do as well, but it raises a `TypeError: type 'DiGraph' is not subscriptable`

src/openforms/forms/logic_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+                # Add an edge from the dependency rule to the rule we are currently
+                # evaluating.
+                graph.add_edge(dep_rule, rule, variables={key})
+
+    return graph
+
+
+class Cycle(TypedDict):
+    rules: Collection[FormLogic]
+    variables: Collection[str]
```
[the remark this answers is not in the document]
viktorvanwijk: I think I've seen too many type hints for the components :sweat_smile:

src/openforms/forms/models/logic.py
```diff
@@ -25,6 +25,13 @@ class FormLogic(OrderedModel):
         on_delete=models.CASCADE,
         help_text=_("Form to which the JSON logic applies."),
     )
+    form_steps = models.ManyToManyField(
+        FormStep,
+        verbose_name=_("form steps"),
+        help_text=_("List of form steps on which this logic rule should be executed."),
```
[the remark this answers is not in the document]
viktorvanwijk: No for the form steps it doesn't really matter, was just wondering :). Will change it to "Collection"

sergei-maertens (approved):

PR #5904, opened 2026-01-20
Add step to "disable next" logic action

### LINKED ISSUE #5861: Change the "disable next" logic action to "disable step"
### Context
For #2409, we need a way of assigning logic rules to a specific step. The problem with the "disable next" action is that there is no context about when to execute it, unlike making a step (not) applicable for example.

### Suggested change
Suggestion is to change it into a "disable specific step" action, where the form builders can assign a step manually.

### Problems with existing forms
The difficulty with this change is determining which steps to assign to the existing "disable next" actions.

One option is to determine a step from the "input" variables used in the JSON logic trigger of the rule. Analysis of production environments has shown that it will be accurate (doesn't result in behavior changes w.r.t. the current situation) in 95%[^1] of the cases. Here, the decision whether to disable continuing to the next step is only dependent on component variables of a single step.

A more difficult situation is when the trigger includes variables that have been prefilled, which means the values will be available upon submission creation. Take for example the rule with the trigger 
```
{
    ">": [
        {"date": {"var": "geboorte"}},
        {"date": {"-": [{"var": "today"}, {"duration": "P24Y"}]}},
    ]
}
```
where "geboorte" is a prefilled date field on the second step. Trying to resolve a step from the input variables will result in step 2. In this case, the "trigger_from_step" is set to `None`, which means the rule will be executed on the first step, and will have an effect on whether the user can continue to step 2. Meaning a change in behavior.

The inverse of above described situation also requires special attention. Consider another prefilled "geboorte" date field, but on step 1 in this case, and the "trigger_from_step" is set to step 2. Resolving for the input variables will result in step 1, even though the current situation dictates the rule should be executed from step 2. We probably need to look at the "trigger_from_step" first, and the input variables second. 

The most complex case is when a user-defined variable is used as an input, and the "trigger_from_step" is not specified. In this case, we can only assign a step if we were to analyze all logic rules into a DAG, and look at the step(s) of the predecessors of this rule. Seems a bit much to implement this in a single script, though. Might just assign the first step as a best guess, and create a warning that this form needs to be checked manually.

[^1]: Not the exact number, just an educated guess 😅 

### Suggested flow for converting existing forms
1. Check whether "trigger_from_step" is set -> if yes, assign the step after that one
2. Resolve steps of input variables
3. If variables have not been prefilled -> assign the step after the resolved one. Create more "disable step" actions on the rule if the input variables are from multiple steps.
4. Assign the first step in all other cases and notify that they need a manual check

viktorvanwijk:
Closes #5861 

**Changes**

* Added step selection field to the "disable next" action, which indicates on which step the corresponding rule should be executed
* Added data migration to populate existing actions with the _correct_ step
* Added detection script that indicates which rules need a manual check by the form designers/maintainers to ensure the right step was assigned during the data migration

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


sergei-maertens (approved):
I'm worried that I didn't find any remarks 😛

PR #5905, opened 2026-01-21
[#5820] Add translations metadata model and admin actions

### LINKED ISSUE #5820: Provide "power user" tooling to replace the static translations used by Open Forms
### Thema / Theme

Other

- modify those files
- upload the modified files

**Drawbacks**

vaszig:
Closes #5820 partly

**Changes**

- Added `TranslationsMetaData` model
- Updated the admin and the relevant actions (download the default messages)
- Updated groups (only super user can deal with custom translations)
- Added tests

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/translations/models.py
```diff
@@ -0,0 +1,93 @@
...
+            "The output from the processing phase, useful for debugging purposes."
+        ),
+    )
+    last_updated = models.DateTimeField(
+        _("last updated"),
+        blank=True,
+        null=True,
+        help_text=_(
+            "Keeps track of the last successful processing of the uploaded file. "
+            "Basically this helps to understand since when the custom translations are "
```
viktorvanwijk:
```suggestion
            "Useful for understanding since when the custom translations are "
```

viktorvanwijk (commented):
Few minor comments :)

viktorvanwijk (approved):

sergei-maertens (changes requested):
Please enable type checking in `pyright.pyproject.toml` for the new code :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #5907, opened 2026-01-21
#5319 make OIDC plugin conditionally visible

### LINKED ISSUE #5319: Dynamic display of login options
### Thema / Theme

Frontend

In Open Formulieren, users can log in using pre-configured authentication methods, such as eHerkenning via OpenID Connect or DigiD. For Rotterdam, we would like to use OpenID Connect for municipal employees as well. 

The use case involves a front desk employee, for example, who can open a form, enter customer data, and submit the form on behalf of the customer. 

Image

However, there is a challenge where the OpenID Connect authentication method for employees becomes visible to all users, including external customers, on the login screens of the forms. This leads to confusion, as the login method intended for employees should only be available within the municipality’s internal network.

This RFC seeks to identify the best solution for restricting the visibility of the OpenID Connect method to only internal municipal users, providing a cleaner, less confusing user experience. 

By ensuring that only relevant authentication methods are displayed for external users, we aim to enhance usability for customers.

In the previous system (e-Suite), this issue was somewhat mitigated by allowing the URL to include an option specifying which login method would be visible. While this workaround helped, it is far from ideal, as the URL be manipulated. 

We are seeking feedback on the feasibility of this approach and if a more secure or elegant solution exists.

annashamray:
Closes #5319

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


sergei-maertens (approved):
This requires an SDK counterpart to relay the `authVisible` query param, no?

sergei-maertens (approved):

PR #5912, opened 2026-01-22
Experiment: use msgspec to parse Formio (component) definitions

sergei-maertens:
[skip: e2e]

**Plan to get this in main**

The impact is huge and big bang change is pretty much the only way to get this through, as it affects all the component-specific processing which happens in all layers of the application. Though, this PR will get split up to do this in a responsible way.

1. Implement all the changes in this PR and get the test suite to still pass.
2. Once that is done (or gradually), the fixes/improvements to test data setup can be extracted and merged into main. That's a net benefit either way.
3. The Formio type definitions and msgspec-specific parsing code can be extracted and merged, without wiring things up. At this point, the type definitions are fully validated because of 1, and few changes will be expected in that area.
4. Our own test factories can be updated to run the msgspec validation. This prevents further issues from seeping into (new) tests - relates to 2.
5. Once the type conversion/parsing code is present in a release, we can define a check script that loads all existing form definitions on (SAAS) instances. This allows us to find gaps between the type definitions and the reality.
6. With the findings from 6, we can write fix scripts/data migrations/converters that bring existing data properly in line.
7. We can add a sanity check in the form definition serializer that validates that whatever the API submits, is understood and correctly shaped. This stops new issues seeping into data and relates to 5/6.
8. Merge the refactor that performs the actual processing on msgspec structs & get that released.

**Changes**

[Describe the changes here]

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


bin/report_component_problems.py
```diff
@@ -193,6 +193,57 @@ def check_component(component: Component) -> Iterator[str]:
...
+        case {
+            "type": "textfield"
+            | "email"
+            | "phoneNumber"
+            | "postcode"
+            | "textarea"
+            | "select"
+            | "date"
+            | "datetime"
+            | "time",
```
sergei-maertens: remove these

PR #5914, opened 2026-01-23
[#5553] organisation branch number

### LINKED ISSUE #5553: Support for vestigingsnummer in Employee Login within Open Formulieren
### Thema / Theme

Frontend

Currently, Open Formulieren supports the scenario where an employee logs in and submits a form on behalf of a customer. When entering the customer data, only a Chamber of Commerce (KvK) number can be provided. There is no option to enter a specific establishment number. As a result, the main company is always registered, even when the data of a particular establishment would be more relevant to the context of the request.

We propose extending Open Formulieren so that, in addition to the KvK number, an establishment number can also be entered. This makes it possible to link the form to the correct establishment’s data and forward it to the back-end systems.

Image

By supporting the establishment number, we can record more accurately on whose behalf a request has been submitted. This is important for both the Verzoeken API and for registration in Open Zaak, where the distinction between headquarters and a sub-establishment can carry legal and administrative weight. It contributes to higher data quality and prevents confusion or errors in the processing of requests.

Additionally, it improves the usability of the forms for employees and customers, as the data capture aligns more closely with the actual situation. Recording the correct establishment data also supports better traceability and reporting.

Many organizations have multiple establishments, with different contact details, addresses, and responsible parties per location. Think of healthcare institutions, educational organizations, or companies with multiple branches. If only the main KvK number can be registered, this leads to information loss or requires manual corrections.

By supporting the use of establishment numbers, we better align with the practices of the Chamber of Commerce and data exchange with government organizations. This also supports the overall objective of ensuring reliable and consistent data exchange within the municipal application landscape.

SonnyBA:
Closes #5553

**Changes**

Adds the ability to provide a branch number whenever a user chooses the "Company" option  for the "Continue as..." field. The branch number will be used after submission for several registration plugins (either as variable or corresponding field).

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/authentication/signals.py
```diff
@@ -113,7 +113,15 @@ def set_auth_attribute_on_session(
...
-            registrator_save: BaseAuth = {
+
+            if "branch_number" in registrator_subject:
+                auth_save.update(
+                    legal_subject_service_restriction=registrator_subject[
+                        "branch_number"
+                    ]
+                )
+
+            registrator_save: FormAuth = {
```
sergei-maertens: why is this changed to `FormAuth`? `FormAuth` has many more properties than `BaseAuth`, and the `store_registrator_details` signature only takes `BaseAuth`, which is passed to `RegistratorInfo`, which does not support all the additional properties of `FormAuth` compared to `BaseAuth`.

src/openforms/contrib/customer_interactions/client.py
```diff
@@ -77,14 +77,40 @@ def _get_digital_addresses_for_kvk(self, kvk: str) -> Iterator[DigitaalAdres]:
...
+        Search digital addresses by branch number of the party.
+
+        The selection of query params and their values are based on the OAS for the
+        "partij identificator":
+        https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/maykinmedia/open-klant/master/src/openklant/components/klantinteracties/openapi.yaml#tag/partij-identificatoren
+        """
+        params: ListDigitaalAdresParams = {
+            "verstrektDoorPartij__partijIdentificator__codeSoortObjectId": "vestigingsnummer",
+            "verstrektDoorPartij__partijIdentificator__codeRegister": "hr",
+            "verstrektDoorPartij__partijIdentificator__codeObjecttype": "vestiging",
```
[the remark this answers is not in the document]
sergei-maertens: I don't think this is  complete? Searching purely on vestigingsnummer does not uniquely identify the party - it's the combination of `KVK` + `vestigingsnummer` iirc, right @swrichards?

src/openforms/contrib/kvk/client.py
```diff
@@ -78,6 +95,30 @@ def get_profile(self, kvk_nummer: str) -> BasisProfiel:
         return response.json()
 
 
+class KVKBranchProfileClient(HALClient):
+    @tracer.start_as_current_span(
+        name="get-profile", attributes={"span.type": "app", "span.subtype": "kvk"}
+    )
+    @elasticapm.capture_span("app.kvk")
```
sergei-maertens: this can be removed, elastic APM integration is deprecated

src/openforms/prefill/contrib/kvk/plugin.py
```diff
@@ -86,13 +95,16 @@ def get_prefill_values(
...
+    def modify_result(result: BasisProfiel | VestigingsProfiel):
+        if "_embedded" in result:  # BasisProfiel
+            # first try getting the addresses from the embedded 'hoofdvestiging'. Note that
+            # this may be absent or empty depending on the type of company (see #1299).
+            # If there are no addresses found, we try to get them from 'eigenaar' instead.
+            addresses = glom(result, "_embedded.hoofdvestiging.adressen", default=None)
+            if addresses is None:
+                addresses = glom(result, "_embedded.eigenaar.adressen", default=None)
+        else:  # VestigingsProfiel
+            addresses = glom(result, "adressen", default=None)
```
sergei-maertens: can we use some form of pattern matching or instance checking to determine when the result is what?

Imagine another flavour coming in in the future and then the association of `else` with `VestigingsProfiel` is pretty weird and may just be plain wrong.

You could pass in the `client` argument so that you can pattern match that:

```py
match client:
    case KVKProfileClient():
        ...
    case KVKBranchProfileClient():
        ...
    case _:
        assert_never(client)
```

src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -1577,6 +1577,182 @@ def test_plugin_vestiging_initiator_kvk_and_vestigingsnummer(self, m, mock_task)
             },
         )
 
+    @patch("celery.app.task.Task.request")
```
sergei-maertens: please limit the assertions to just what is related to the branch number stuff instead of copy pasting the bloat from other tests

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -325,6 +326,11 @@ def register_submission(
         kvk = betrokkene_identificatie.get("innNnpId")
         vestigingsnummer = betrokkene_identificatie.get("vestigingsNummer")
 
+        if not vestigingsnummer and submission.is_authenticated:
```
sergei-maertens: I think here too we want a `DeprecationWarning` so that we can assign the vestigingsnummer by default from the auth context as default behaviour in 4.0

sergei-maertens (changes requested):

sergei-maertens (changes requested):

sergei-maertens (changes requested):
For the Open Klant stuff - please create a follow up ticket to handle one of both cases, so that we can proceed with this merging this PR :)

sergei-maertens (approved):

PR #5927, opened 2026-01-28
[#5820] Process custom translations asset in the backend

### LINKED ISSUE #5820: Provide "power user" tooling to replace the static translations used by Open Forms

vaszig:
Closes #5820 partly

**Changes**

- Celery task triggered by the admin page for processing the custom (uploaded) file messages
- Updated frontend dependencies

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/translations/subprocesses.py
```diff
@@ -0,0 +1,47 @@
+import subprocess
+import tempfile
+from pathlib import Path
+
+
+def compile_messages_file(input_path: str) -> tuple[bool, str | None]:
```
viktorvanwijk: When is the second return value `None`? Can only be a string, no?

src/openforms/translations/subprocesses.py
```diff
@@ -0,0 +1,47 @@
...
+            compiled_json: str = f.read()
+
+        return True, compiled_json
+    except subprocess.CalledProcessError as exc:
+        error_msg: str = exc.stderr.decode("utf-8").strip()
+        return False, error_msg
+
+    finally:
+        # Clean up the temporary file if it exists
+        if output_path and (temp_output_path := Path(output_path)).exists():
```
viktorvanwijk: Will it ever not exist?

src/openforms/translations/tests/test_admin.py
```diff
@@ -139,16 +139,93 @@ def test_default_messages_download_link_with_instance_saved(self):
...
+            "test.json",
+            b'{"s":"test"}',
+            "application/json",
+        )
+        form.submit().follow()
+
+        obj = TranslationsMetaData.objects.get()
+
+        self.assertEqual(obj.language_code, "nl")
+        self.assertEqual(obj.messages_file.file.read(), b'{"s":"test"}')
```
viktorvanwijk: Perhaps add some assertions on `compiled_assets`? Seems to me like this only ensures the Django admin  works properly, not that the messages are compiled on saving

viktorvanwijk (commented):
Few minor questions/suggestions :)

I have noticed that if you try to delete the uploaded messages file, the admin crashes with a `ValueError: The 'messages_file' attribute has no file associated with it.`

Another small thing is that clicking the compiled translations link will actually save the file locally, but clicking the "Download" link for the default messages will only open it in a new tab. Should we make it consistent to always save the file locally directly?

src/openforms/translations/subprocesses.py
vaszig: Just for making sure, I can't think of a bad situation but I can remove it.

sergei-maertens (approved):
Some details/fine-tuning, but functionally this looks solid.

src/openforms/translations/subprocesses.py
```diff
@@ -0,0 +1,46 @@
...
+        (True, compiled_json) on success
+        (False, error_message) on failure
+    """
+    with tempfile.NamedTemporaryFile(
+        mode="r",
+        suffix=".json",
+        encoding="utf-8",
+        delete=False,
+    ) as output_tmp:
+        output_path = output_tmp.name
```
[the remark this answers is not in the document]
vaszig: Shouldn't I use the context manager for closing the file?This is what I thought.

src/openforms/translations/tasks.py
```diff
@@ -0,0 +1,86 @@
...
+    success, result = compile_messages_file(instance.messages_file.path)
+
+    if success:
+        log.info("custom_translations_file_compiling_succeeded")
+
+        instance.processing_status = StatusChoices.done
+        instance.debug_output = ""
+        instance.last_updated = timezone.now().astimezone(ZoneInfo("Europe/Amsterdam"))
+
+        assert result is not None
```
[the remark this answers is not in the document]
vaszig: Yeah missed that part when I removed `None` from the subprocess function type hints.

Dockerfile
```diff
@@ -56,6 +56,10 @@ COPY ./*.json ./*.js /app/
 # install WITH dev tooling
 RUN npm ci --legacy-peer-deps
 
+# install necessary global dependencies (for example subprocess.run(["formatjs", ...]) 
+# looks for a global CLI on PATH)
+RUN npm install -g @formatjs/cli@6.6.1
```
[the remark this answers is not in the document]
sergei-maertens: The Dockerfile has multiple stages:

1. L8 - install the backend dependencies etc.
2. L48 - install the frontend dependencies and build the frontend code
3. L66 (70 in your PR) - prepare the production image, which gets pushed to Docker Hub

The production image copies the binaries & assets from earlier stages into itself, see L112-L119.

In its current form, the `@formatjs/cli` is not present in the production image. You only globally install the package in the second stage where frontend dependencies are installed and built.

You can test things out in the container by building the image and running it locally:

```bash
docker compose build web
```

and then run the container with bash, removing it on exit:

```bash
docker run -it --rm openformulieren/open-forms:latest bash
```

you can then explore in the container what is and isn't there :-)

sergei-maertens (approved):
I suspect this will crash on our test env for the actual compilation, but we can follow that up in a separate PR :-)

PR #5929, opened 2026-01-28
[#5820] Add endpoint for retrieving the customized translations

### LINKED ISSUE #5820: Provide "power user" tooling to replace the static translations used by Open Forms

vaszig:
Closes #5820 partly

**Changes**

- Add endpoint for getting the customized translations from the backend based on the language code.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/translations/api/views.py
```diff
@@ -126,3 +132,84 @@ def get(self, request, language: str):
...
+
+    def _get_custom_translation_data(self, language_code: str) -> JSONObject | None:
+        # filter the objects by the received language code and the most recent and
+        # successfully processed upload
+        translations_metadata = (
+            TranslationsMetaData.objects.filter(
+                language_code=language_code,
+                last_updated__isnull=False,
+            )
+            .order_by("-last_updated")
```
vaszig: We do not have a constraint in the model concerning how many rows can exist with the same language code so I thought the last successful update can be our choice. Not sure about this, but I believe it's better from preventing the user to add multiple rows with the same language code.

src/openforms/translations/api/views.py
```diff
@@ -78,7 +84,7 @@ def put(self, request: Request) -> Response:
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
 
-        language_code = serializer.data["code"]
+        language_code = serializer.data["code"]  # pyright: ignore[reportCallIssue,reportArgumentType]
```
vaszig: This and the similar ones below were added because the types in the library do not help. I avoided using `cast` as we try to get rid of it in the project.

src/openforms/translations/api/views.py
```diff
@@ -131,11 +142,94 @@ def get(self, request, language: str):
...
+        "200": {
+            "oneOf": [
+                build_object_type(additionalProperties=False),
+                build_object_type(
+                    additionalProperties=build_array_type(
+                        build_object_type(additionalProperties=True)
+                    )
+                ),
+            ],
+        },
```
sergei-maertens:
```suggestion
        "200": build_object_type(),
```
I think I'd keep things simple like this. We don't control the compilation output format anyway, and it's not very relevant either because it just needs to feed into `react-intl`.

I'd only textually document in the description that if there's no custom translations, an empty object `{}` is returned.

src/openforms/translations/api/views.py
```diff
@@ -131,11 +142,94 @@ def get(self, request, language: str):
...
+            )
+
+        compiled_asset_path = self._get_custom_translations_file_path(language_code)
+
+        if not compiled_asset_path:
+            return JsonResponse(data={})
+
+        return sendfile(
+            request, compiled_asset_path, attachment=True, mimetype="application/json"
+        )
```
sergei-maertens: I don't think you want `attachment=True`, as that will make the browser download the file.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #5931, opened 2026-01-28
Replace/refactor logevent module to structlog adapters

### LINKED ISSUE #5351: Refactor `openforms.logging.logevent` to use structlog
We're now manually creating database records for relevant log events, which should instead be driven by logging calls from structlog. In many places, we're now doing double logging work while the database logs can be inferred from the structlog calls instead.

This also gives the opportunity to make sure the logs are persisted in the DB in a thread-safe way and don't vanish when the DB transaction gets rolled back.

To be fleshed out more in terms of marking records for DB storage and setting up *multiple* handlers for them + ensuring the log level doesn't cause them to be dropped.

sergei-maertens:
Closes #5351

[skip: e2e]

**Changes**

* Defined an audit logger that doesn't drop any log records because of the log level (in django-timeline-logger)
* Replaced all the `openforms.logging.logevent.*` calls with their equivalent audit logger (structlog) call
* Removed the StUF request/response audit logs to the database after checking with support that they never use this.
* Fixed up the type checking in `openforms.logging` and added the entire package to the Pyright config\
* Added documentation about audit logging

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/logging/adapter.py
```diff
@@ -0,0 +1,765 @@
...
+                event=_event,
+                instance=submission,
+                plugin=_plugin,
+                tags=[TimelineLogTags.submission_lifecycle],
+                extra_data={
+                    "object_reference": event_dict.get("object_reference"),
+                },
+            )
+
+        case _:  # pragma: no cover
```
annashamray: I don't know how I feel about 600+ lines of `match case`. But using atomic functions was also the same amount of effort. But you don't want to have some generic adapter (or adding it here in the wildcard block) ?

src/openforms/appointments/utils.py
```diff
@@ -69,13 +70,11 @@ def delete_appointment_for_submission(submission: Submission, plugin=None) -> No
     try:
         plugin.delete_appointment(appointment_info.appointment_id)
         appointment_info.cancel()
-    except AppointmentDeleteFailed as e:
-        log.warning("appointment_cancel_failure")
-        logevent.appointment_cancel_failure(appointment_info, plugin, e)
+    except AppointmentDeleteFailed as exc:
+        audit_log.warning("appointment_cancel_failure", exc_info=exc)
```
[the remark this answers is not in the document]
sergei-maertens: yeah when I was converting everything from stdlib logging to structlog, those stuck out like a sore thumb, so I'm quite happy with the impact of the actually useful application code.

src/openforms/logging/adapter.py
annashamray: Thank you for the explanation of the reasoning!

annashamray (approved):

Issue #5882, opened 2026-01-06
Develop new API endpoint(s) for the new admin UI


sergei-maertens:
The current admin interface to edit and save forms uses the RESTful API endpoints for the resources that define "a form", such as:

* form definitions
* form steps
* form variables
* form logic rules
* ...

This has some serious drawbacks for the validation, because of race conditions in network requests some aspects cannot be validated reliably on the backend. It also makes the orchestration on the frontend complex to make sure the network calls happen in the right order and validation error messages are extracted properly so they can be displayed.

For the new admin UI, we want to simplify this with a complex `Form` resource that contains all the necessary information. The end-goal is that the frontend code does a single `PUT /api/v3/forms/:uuid` call to update (or create?) a form, which gets processed and saved in a single database transaction and where all validation of the integrity is possible.

**Tasks**

- [x] Define v3 top level API routes (we currently use v2)
- [x] Ensure API documentation is available on `/api/v3/docs/`. It's okay if the endpoints are not listed in `src/openapi.yaml`, it's also okay if both are.
- [x] Implement the new `/api/v3/forms/:uuid` detail endpoint
    - [x] Accept `PUT` calls. The frontend can generate a uuid through `window.crypto.randomUUID()` for new forms
    - [x] (Session) Authentication is required, and you must have the `FormAPIPermissions` permission class.
    - [x] You can specify the top-level properties (name, internal name, internal remarks, activation date... etc.)
    - [x] You can provide the nested `steps`
        - The array of steps replaces any previous version entirely
        - It contains the `FormStep` data, and contains the nested `FormDefinition`
        - Ideally there would be matching on UUID, in case step re-orders are submitted that records do not get recreated, but that's open for discussion depending on complexity
        - form definitions can be single-use or re-usable
    - [x] Keep track of `login required` at the form level rather than the step. A transitional period where both exist is acceptable.
    - [x] You can provide all the confirmation templates/options, such as the custom confirmation email template
    - [x] You can submit the registration backends inline
    - [x] Product reference is a related resource reference, but payment configuration is provided inline
    - [x] You can provide the logic rules inline
    - [x] You can provide the (user defined) variables inline
- [x] Managing variables
    - The form variables are derived/generated from the provided forms steps
    - The user defined variables are taken from the user input
-  Implement missing validations
    - [x] Validate logic rules: input parameters and outputs must point to valid references (variables, steps...)
[...]

PR #5935, opened 2026-01-30
Show form steps in logic rule overview

### LINKED ISSUE #2409: The logic evaluation should be staggered and based on a dependency graph

viktorvanwijk:
Closes #2409

~~Dependent on #5863~~

[skip: e2e]

**Changes**

* Added form steps to the form logic serializer
* Ensured we show form steps on logic rule overview if the logic rule analysis feature flag is enabled
* Updated and add tests

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -8584,9 +8584,20 @@ components:
           type: boolean
           description: Is this an advanced rule (the admin user manually wrote the
             trigger as JSON)?
+        formSteps:
+          type: array
+          items:
+            type: string
+            format: uri
+            title: form steps
+          readOnly: true
+          description: Form steps on which the rule will be executed, determined by
+            logic rule analysis. Note that this is only relevant when the `new_logic_evaluation_enabled`
+            feature flag on the form is set to `True`.
       required:
       - actions
       - form
+      - formSteps
       - jsonLogicTrigger
       - order
       - url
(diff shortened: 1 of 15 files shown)


sergei-maertens:
> As can be seen, only a single extra query is performed when retrieving and updating the logic rules. It seems like we could do some more optimizations when updating the logic rules, as 62 queries is still quite a lot. I quickly looked if there was some low-hanging fruit, but wasn't able to find where these queries come from. Need to do some more investigation.

Write operations being slower is fine and an acceptable tradeoff. There's also a chance that Sonny's rework in #5882 already improves the performance or at least makes it more within reach to optimize, so let's not burn our energy on this. The read aspect is important for the load times of a form in the admin, and that's great now with the prefetch query.

sergei-maertens (approved):

PR #5940, opened 2026-02-03
[#5952] add initial form endpoint

### LINKED ISSUE #5952: Create endpoint with top-level properties
Should include name, internal name, internal remarks, activation date...

SonnyBA:
Closes #5952 

[skip: e2e]

**Changes**

Adds a stripped down version (top level attributes only) of what will be required for #5882

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/forms/api/v3/viewsets.py
sergei-maertens: Type check the new modules in `pyright.pyproject.toml` please, and add type hints where they're missing.

src/openforms/api/urls.py
```diff
@@ -46,6 +47,10 @@
 forms_router.register(r"steps", FormStepViewSet, basename="form-steps")
 forms_router.register(r"versions", FormVersionViewSet, basename="form-versions")
 
+# v3 router
+router_v3 = routers.DefaultRouter(trailing_slash=False)
+router_v3.register(r"forms", FormViewSetV3, basename="form-v3")
```
annashamray: I think "v3"  belongs to the namespace here. And I'd suggest some file reorganization, because now v3 urls are hidden among different v2 urls despite being for another admin UI version. 
maybe it's better to create `v2` and `v3` url files, the namespace of `v2` urls can be left the same, not to rewrite all `reverse` calls, but I think the namespace of new urls should include `v3` 

Also it will help with generating different OAS files/pages for v2 and v3 urls

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -0,0 +1,290 @@
...
+from openforms.forms.models.form import Form
+from openforms.forms.models.form_authentication_backend import FormAuthenticationBackend
+from openforms.forms.models.form_registration_backend import FormRegistrationBackend
+from openforms.products.models.product import Product
+from openforms.translations.api.serializers import ModelTranslationsSerializer
+
+
+@extend_schema_serializer(component_name="FormV3Serializer")
+class FormSerializer(serializers.ModelSerializer):
+    login_options = LoginOptionsReadOnlyField()
```
annashamray: This is a little confusing - you have a lot of readonly fields, but the endpoint supported only PUT method. Are you going to add GET support in the next PR?

src/openapi.yaml
```diff
@@ -6162,6 +6162,53 @@ paths:
               $ref: '#/components/headers/X-Is-Form-Designer'
             Content-Language:
               $ref: '#/components/headers/Content-Language'
+  /api/v3/forms/{uuid}:
+    put:
+      operationId: api_v3_forms_update
```
annashamray: In the web UI it's shown in the "api" group which is confusing
Also "v3" endpoint is shown at "/api/v2/docs/" path, which is unexpected

annashamray (changes requested):
1. Could you check all tasks that are covered by this PR in the issue? 
2. Please add new files to `include` in `pyright.pyproject.toml`

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: The readonly fields not only show up for GET requests but also in response bodies for other methods (the fields are not writable though). A lot of these fields might be familiar because I took the existing `FormSerializer` as a starting point but I agree that we might need to clean up some fields here.

@robinmolen I think it would be nice if we could discuss the fields that you will require for the front-end so that we can remove the fields here that aren't necessary. I'm not sure though if you are at a stage with the front-end work where you have a clear picture of that.

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -0,0 +1,429 @@
...
+            "slug": "update-form",
+        }
+        response = self.client.put(url, data=data)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(Form.objects.count(), 1)
+        form = Form.objects.get()
+
+        self.assertEqual(form.name, "Update form")
+        self.assertEqual(form.slug, "update-form")
```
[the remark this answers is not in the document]
SonnyBA: See [comment thread](https://github.com/open-formulieren/open-forms/pull/5940/changes/BASE..2de24eeaa823e0094223de8c1594c02c1f4cfb32#r2799872920) earlier.

src/openforms/forms/api/v3/serializers/form.py
robinmolen: I'll create an overview of the required data based on the designs and the current admin. We can discuss it this week 👍

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -0,0 +1,429 @@
...
```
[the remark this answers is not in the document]
annashamray: Actually I don't know what is the expected behavior here with PUT as create. 
Should we ask @sergei-maertens to help us out here?

annashamray (changes requested):

src/openforms/forms/api/v3/serializers/category.py
```diff
@@ -0,0 +1,30 @@
...
+            "depth",
+        )
+
+    class NameUUID(TypedDict):
+        name: str
+        uuid: str
+
+    def get_ancestors(self, obj) -> list[NameUUID]:
+        ancestors = obj.get_ancestors()
+        return [{"uuid": str(n.uuid), "name": n.name} for n in ancestors]
```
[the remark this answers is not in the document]
sergei-maertens: Afaik we don't yet have designs for the form list page, with forms grouped by category. So let's wait for that until we decide what we need for the endpoints for that page.

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -0,0 +1,497 @@
...
+        self.assertTrue(form._is_deleted)
+
+    def test_update_form_incorrect_request(self):
```
sergei-maertens: see earlier remark - where are the error field + code assertions?

sergei-maertens (changes requested):

sergei-maertens (approved):
Some small touchups and then it can be merged I think.

src/openforms/forms/api/v3/viewsets.py
```diff
@@ -0,0 +1,63 @@
...
+
+from ...api.parsers import FormCamelCaseJSONParser
+from ...api.permissions import FormAPIPermissions
+from ...api.renderers import FormCamelCaseJSONRenderer
+from ...models.form import Form
+from .serializers import FormSerializer
+
+
+class KwargsType(TypedDict):
+    uuid: str
```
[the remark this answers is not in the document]
SonnyBA: It is converted like you mentioned, it also is [mentioned as a default path converter in the docs](https://docs.djangoproject.com/en/5.2/topics/http/urls/#path-converters).

annashamray (approved):

sergei-maertens (approved):

PR #5947, opened 2026-02-06
[#5820] Add components configuration for contact details to the Appointments

### LINKED ISSUE #5820: Provide "power user" tooling to replace the static translations used by Open Forms

vaszig:
Closes #5820 partly

**Changes**

- Added components configuration to the JccRestConfig model
- Introduce `formMode` for the type of a form in the form builder ('default', 'appointment')

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene


src/openforms/appointments/contrib/jcc_rest/admin.py
```diff
@@ -1,10 +1,36 @@
...
+        request,
+        context: dict[str, Any],
+        add=False,
+        change=False,
+        form_url="",
+        obj=None,
+    ):
+        context.update(
+            {
+                "available_components": ", ".join(CustomerFields.values),
```
viktorvanwijk: Where/how is this used?

src/openforms/appointments/contrib/jcc_rest/forms.py
```diff
@@ -0,0 +1,58 @@
...
+        user_set = set([component["key"] for component in components])
+        modified = user_set - initial_set
+
+        if modified:
+            raise ValidationError(
+                {
+                    "configuration": _(
+                        "Unknown component key for fields: {keys}. "
+                        "Adding a component or modifying a component's key is not allowed."
+                    ).format(keys=",".join(modified))
```
viktorvanwijk:
```suggestion
                    ).format(keys=", ".join(modified))
```

src/openforms/appointments/contrib/jcc_rest/forms.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    def clean(self) -> dict[str, Any]:
+        super().clean()
+
+        components = self.cleaned_data["configuration"]["components"]
+
+        initial_set = set(CustomerFields.values)
+        user_set = set([component["key"] for component in components])
+        modified = user_set - initial_set
```
viktorvanwijk: Should we also show a warning when a component is missing?

viktorvanwijk (commented):
I was a bit scared by the amount of lines and files updated, but it was mostly because of the VCR cassettes 😅 

General structure seems OK to me, just a few questions :)

src/openforms/appointments/contrib/jcc_rest/admin.py
vaszig: Helpful in development when was testing things, I forgot it, will remove it.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #5965, opened 2026-02-11
Bump to ClamAV 1.5 in docker compose

sergei-maertens:
The 1.0.0 tag is gone on Docker Hub - 1.0 still exists, but there is already a newer version available that should be backwards compatible.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -109,7 +109,7 @@ services:
       - open-forms-dev
 
   clamav:
-    image: clamav/clamav:1.0.0
+    image: clamav/clamav:1.5
     healthcheck:
       test: ["CMD", "clamdcheck.sh"]
       interval: 5s


viktorvanwijk (approved):

viktorvanwijk (commented):
If it turns out to be not backwards compatible, where/how would we notice this?

sergei-maertens:
> If it turns out to be not backwards compatible, where/how would we notice this?

I'll test it out with docker-compose after lunch!

sergei-maertens:
Backports:

* stable/3.4.x: d1b21b122bc0bb5b082ab8def89f451e66cdd739
* stable/3.3.x: 3ff1ae2d8aedbfacd4d28d7cd04d6f7780ead4c7

PR #5979, opened 2026-02-17
[#5820] Update Dockerfile to support formatjs during runtime

### LINKED ISSUE #5820: Provide "power user" tooling to replace the static translations used by Open Forms

vaszig:
Closes #5820 partly

**Changes**

Due to the subprocess we have for the custom translations in the backend, we have to use formatjs during runtime. This was not available in the last stage of the Dockerfile (no need for such frontend realated stuff until now).

Instead of installing everything in the last stage we only copy what we need from the frontend-build stage and in this way we can execute the binary code of the library without having to do network calls (everything needed is found in the already bin directory except for the node, which is a must).

Last needed update is the path in the subprocess because this changes depending on the environment we use.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

### DIFF
--- a/src/openforms/translations/subprocesses.py
+++ b/src/openforms/translations/subprocesses.py
@@ -2,6 +2,11 @@ import subprocess
 import tempfile
 from pathlib import Path
 
+# make sure the right path is used for the subporcess (different paths are used based on
+# the environment)
+PROJECT_ROOT = Path(__file__).parent.parent.parent.parent
+FORMATJS_BIN = PROJECT_ROOT / "node_modules/.bin/formatjs"
+
 
 def compile_messages_file(input_path: str) -> tuple[bool, str]:
     """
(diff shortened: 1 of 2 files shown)


src/openforms/translations/subprocesses.py
```diff
@@ -2,6 +2,11 @@
 import tempfile
 from pathlib import Path
 
+# make sure the right path is used for the subporcess (different paths are used based on
+# the environment)
+PROJECT_ROOT = Path(__file__).parent.parent.parent.parent
```
sergei-maertens: you should be able to grab that from `django.conf.settings`: https://github.com/open-formulieren/open-forms/blob/5b956b957cae583b1118884854703df92d96dd9b/src/openforms/conf/base.py#L34

sergei-maertens (changes requested):

sergei-maertens (approved):
Let's go and get this behind us!

PR #5995, opened 2026-02-24
Optimize flower memory usage

sergei-maertens:
(no ticket, coming from our Maykin's SAAS monitoring)

**Changes**

* Using the --broker option instead of pointing to the app avoids having to load all the application code in memory - saves about 240+ MiB on my local machine
* Use 100x lower defaults for max number of tasks and workers to keep in memory, cutting down on memory growth over time, especially on k8s where worker names can be unstable

The envvars for tasks/workers are deliberately undocumented. I suspect these new defaults are fine, but if not, then an escape hatch is available.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/bin/celery_flower.sh
+++ b/bin/celery_flower.sh
@@ -2,4 +2,10 @@
 # Set defaults for OTEL
 export OTEL_SERVICE_NAME="${OTEL_SERVICE_NAME:-openforms-flower}"
 
-exec celery --app openforms --workdir src flower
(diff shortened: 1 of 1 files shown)


vaszig (approved):

sergei-maertens:
Result on our test instance :slightly_smiling_face: 

image

PR #5999, opened 2026-02-25
[#5639] Clean up StUF-BG tests

### LINKED ISSUE #5639: Make StUF-BG testing easier
Testing the StUF-BG calls is a bit hard without a proper instance (which would allow having VCR tests) and we have to add some mocked responses (as we already do). 

This seems to be a bit hard in our tests, so it would be wise to refactor a bit the code and add a helper function or a mixin which is going to be used for each testcase. This can assure that we have the proper xsd validation against the requests we create (and the mocked responses?).

vaszig:
Closes #5639

[skip: e2e]

**Changes**

Add class mixin for all the necessary helpers and the most used assertion (xml file validity based on the XSDs).

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


sergei-maertens (commented):
Yeah nice stuff!

Regarding your questions/remarks - I would aim to incorporate this as much as possible and fix/clean up what needs to be cleaned up. Tests that are based on (strictly speaking) invalid setup are unreliable and they encourage future workarounds/shortcuts, which we want to avoid.

If we write code based on (mock) responses that end up not being compliant according to the XSD, we can't trust our implementation.

sergei-maertens (approved):

PR #6026, opened 2026-03-02
[#6016] Sanitize string values for StUF XML file

### LINKED ISSUE #6016: Illegal characters in XML file crash StUF-ZDS registration

3.3.12

### Customer reference

Sentry: 374865

When an illegal character is present in the XML file this is not properly parsed (removed).

We can try to reproduce this by writing a regression test with an illegal character in the submission (do not have access to an environment in order to properly reproduce this).

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #6016

**Changes**

Some characters are considered as illegal in XML content so we make sure that these are removed before we construct the XML files.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


sergei-maertens (changes requested):

sergei-maertens (changes requested):
Some more tweaks with regards to the logging and then this looks ready!

sergei-maertens (approved):

PR #6047, opened 2026-03-05
Fix variable action not applied when component also has hidden property action

### LINKED ISSUE #6046: Variable action not applied when component also has hidden property action
This a regression from the fix for #6005

When a component has a logic rule with a variable action, and another logic rule with a hidden property action, the variable action will not be applied even if the rule with the property action wasn't triggered.

Steps to reproduce:
1. Create form with checkbox and textfield component
2. Create first logic rule with trigger `true` that sets the value of textfield to "foo"
3. Create second logic rule that triggers when the checkbox is checked, and hides the textfield
4. Fill out the textfield and observe it will not be set to "foo" after logic evaluation, even if the checkbox was not checked.

viktorvanwijk:
Closes #6046

[skip: e2e]

**Changes**

* Added regression test
* Ensured `original_input_data` is updated with mutations applied by other logic rules

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/logic/rules.py
+++ b/src/openforms/submissions/logic/rules.py
@@ -246,6 +246,7 @@ def iter_evaluate_rules(
                         for key, value in mutations.items()
                     }
                     data.update(mutations_python)
+                    original_input_data.update(mutations_python)
 
                 yield operation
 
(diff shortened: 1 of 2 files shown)


sergei-maertens (approved):
Nice find, and it highlights the complexity again. It's clear to me that after the breaking changes will be made and we support only one logic evaluation mode (frontend), that we must do more efforts to bring down the complexity of our logic engine because this will not be sustainable.

(cc @bregje-hub )

PR #6059, opened 2026-03-10
Enable the new renderer by default

sergei-maertens:
**Changes**

Changed the model-level default for the new-renderer feature flag so that new forms have it enabled by default.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


vaszig (commented):
I guess we are going to use the django command (`set_forms_new_renderer_enabled`) for existing rows (instead of data migration)?Or do we only care for the new ones (and if so, why)?

vaszig (approved):
Concerning the failing tests, have you seen them?Is it flakiness or something else?

PR #6061, opened 2026-03-10
Add data type information to JSON logic expressions

### LINKED ISSUE #6038: Add routine in backend which can wrap all variable operations in JSON logic with the relevant operators
Relevant operators are:  "date", "time", and "datetime"

For example, a rule with trigger `{"==": [{"var": "dateField"}, {"date": "2026-01-01"}]}` should be converted to `{"==": [{"date": {"var": "dateField"}}, {"date": "2026-01-01"}]}`. 

Possible difficulties:
* Components with children (editgrid, partners, children): we would need to also wrap `{"var": "editGrid.0.dateField"}` with a date operator, which is not directly clear by looking at the form variables.
* Possibly variables inside "map" and "reduce" operations, e.g. "current.[something]", might also need to be converted.
* Date-related components with `multiple: true`. We need to recognize that `{"date": ["2026-01-01", "2026-01-02"]}` contains a list of date strings that need to be converted individually. Ideally, each individual entry needs to get the relevant operator: `[{"date": "2026-01-01"}, {"date": "2026-01-02"}]`. Though note that this won't always be possible. Fields from the current step will not get substituted for a value during partial evaluation, because the value depends on user input. The best we can do in that case is `{"date": {"var": "dateMultiple"}}`. OR maybe we could rewrite `{"var": "dateMultiple"}` into `{"map": [{"var": "dateMultiple"}, {"date": {"var": ""}}]}`?

viktorvanwijk:
Closes #6038

[skip: e2e]

**Changes**

* Added routine to include data-type information in JSON logic expressions
* Added `static_variables` property to `SubmissionValueVariableState`

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/json_logic.py
+++ b/src/openforms/submissions/json_logic.py
@@ -1,0 +1,156 @@
+from collections.abc import Mapping
+
+from json_logic.meta.expressions import (
+    JSONLogicExpression,
+    destructure,
+)
+from json_logic.typing import JSON
+
+from openforms.variables.constants import FormVariableDataTypes
+from openforms.variables.service import resolve_key
+
+from .models import SubmissionValueVariablesState
+
+DATA_TYPE_TO_DATA_OPERATOR_MAP: Mapping[str, str] = {
+    FormVariableDataTypes.date: "date",
+    FormVariableDataTypes.datetime: "datetime",
+}
+
+
+def _get_data_type_operator(
+    var_key: str, state: SubmissionValueVariablesState
+) -> str | None:
+    """Get a data-type operator from a data-access key."""
+    variables = {**state.variables, **state.static_variables}
+
+    form_variable_key = resolve_key(var_key, variables)
+    if form_variable_key is None:
+        return None
+
+    variable = variables[form_variable_key]
+    if (data_type := variable.get_data_type(var_key)) is not None:
+        return DATA_TYPE_TO_DATA_OPERATOR_MAP.get(data_type, None)
+
+    return None
+
+
+def add_data_type_information(
+    expression: JSON,
+    state: SubmissionValueVariablesState,
+    parent_key: str | None = None,
+    in_reduce: bool = False,
+) -> JSON:
+    """
+    Add data-type information to a JSON logic expression.
+
+    Searches the expression for variable operations, and wraps them with the relevant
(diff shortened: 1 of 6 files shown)


src/openforms/submissions/tests/test_json_logic.py
```diff
@@ -0,0 +1,431 @@
+from django.test import TestCase
+
+from openforms.forms.tests.factories import FormFactory
+
+from ..json_logic import add_data_type_information
+from ..tests.factories import SubmissionFactory
+
+
+class AddDataTypeInformation(TestCase):
```
sergei-maertens:
```suggestion
class AddDataTypeInformationTests(TestCase):
```
Convention is to use a `Tests` suffix in test cases

src/openforms/submissions/json_logic.py
```diff
@@ -0,0 +1,155 @@
...
+
+from json_logic.meta.expressions import (
+    JSONLogicExpression,
+    destructure,
+)
+from json_logic.typing import JSON
+
+from ..variables.constants import FormVariableDataTypes
+from ..variables.service import resolve_key
+from .models.submission_value_variable import SubmissionValueVariablesState
```
sergei-maertens:
```suggestion
from .models import SubmissionValueVariablesState
```

src/openforms/submissions/json_logic.py
```diff
@@ -0,0 +1,154 @@
...
+    operator, argument = destructure(normalized)
+    assert isinstance(argument, list)
+
+    match operator:
+        case "var":
+            var_name = argument[0]
+            # We can't add any data type information if the argument is another
+            # (variable) expression, because we have no data available here.
+            if not isinstance(var_name, str):
+                return normalized  # pyright: ignore[reportReturnType]
```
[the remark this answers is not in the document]
sergei-maertens: Doesn't seem to be that since it's all the same internal types, but whatever, leave it for now.

sergei-maertens (changes requested):
I'm surprised with how readable this turns out to be!

Mostly some cosmetic changes not related to the core feature of the ticket.

sergei-maertens (approved):

PR #6065, opened 2026-03-11
Unchecking deceased children by default

### LINKED ISSUE #5857: Default to unchecked "include_deceased" for Childeren
### Thema / Theme

Frontend

The inclusion of a deceased person is a submission is much less likely, compared to excluding them. 

_No response_

_No response_

robinmolen:
Closes #5857

**Changes**

Changing the default value of the Family members prefill plugin option `include_deceased` to `False`. With that, the Family member prefill plugin tests have also been updated, to reflect the new default situation.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/.github/ISSUE_TEMPLATE/prepare-release.md
+++ b/.github/ISSUE_TEMPLATE/prepare-release.md
@@ -36,7 +36,7 @@ assignees: sergei-maertens
 
     - [ ] `openforms.contrib.brk`
     - [ ] `openforms.contrib.customer_interactions`
-    - [ ] `openforms.contrib.haal_centraal.tests.test_integration`
+    - [ ] `openforms.contrib.haal_centraal`
     - [ ] `openforms.contrib.kvk`
     - [ ] `openforms.contrib.objects_api`
     - [ ] `openforms.contrib.reference_lists`
--- a/src/openforms/contrib/haal_centraal/clients/brp.py
+++ b/src/openforms/contrib/haal_centraal/clients/brp.py
@@ -208,7 +208,7 @@ class V2Client(BRPClient):
         bsn: str,
         include_children: bool,
         include_partner: bool,
-        include_deceased: bool = True,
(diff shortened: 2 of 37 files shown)


vaszig (commented):
- Looking at the ticket I am a bit confused. I thought that according to this https://github.com/open-formulieren/open-forms/issues/5857#issuecomment-3709667098 we would expect an answer from NL DS (maybe you are aware of something else though).

- Maybe we should change the default values for these too?
https://github.com/open-formulieren/open-forms/blob/main/src/openforms/contrib/haal_centraal/clients/brp.py#L211
https://github.com/open-formulieren/open-forms/blob/main/src/stuf/stuf_bg/client.py#L105

robinmolen:
>Looking at the ticket I am a bit confused. I thought that according to this https://github.com/open-formulieren/open-forms/issues/5857#issuecomment-3709667098 we would expect an answer from NL DS (maybe you are aware of something else though).

During the previous refinement (9th of March) we decided that we can safely pick this up, and if municipalities expected it differently they would let us know.

> Maybe we should change the default values for these too?
https://github.com/open-formulieren/open-forms/blob/main/src/openforms/contrib/haal_centraal/clients/brp.py#L211
https://github.com/open-formulieren/open-forms/blob/main/src/stuf/stuf_bg/client.py#L105

Good point! I've updated the default `include_deceased` values of these functions 👍

vaszig (commented):
I have some questions/remarks again!

Apart from the ones below I have one more concerning the mocked files in the `openforms/contrib/haal_centraal/tests/files`, I would expect that the ones for `v2` are not necessary any more, no?

sergei-maertens (commented):
Does the github release issue template need updates with the new cassette paths?

sergei-maertens (approved):

PR #6069, opened 2026-03-12
[#5701] Fix `selectboxes` in admin step handler (Objects API v1)

### LINKED ISSUE #5701: Object API v1 registration (template method) + removal of "selectboxes" from form causes crash

3.2.x

### Customer reference

Dimpact 349

When a form uses the (v1) Object API registration, removing a selectboxes component will crash the form. 

introduced in #5243 

```
stepEditHandlers.js:102 Uncaught TypeError: Cannot read properties of undefined (reading 'findIndex')
    at onStepEdit (stepEditHandlers.js:102:78)
    at form-creation-form.js:538:32
    at Proxy.map (<anonymous>)
    at LN (form-creation-form.js:533:73)
    at immer.esm.mjs:1:16170
    at e.produce (immer.esm.mjs:1:16318)
    at immer.esm.mjs:1:16124
    at Object.Ea [as useReducer] (react-dom.production.min.js:170:451)
    at t.useReducer (react.production.min.js:25:274)
    at Hi (use-immer.module.js:1:334)

```

```js    
const matchingMappingIndex = registrationBackendOptions.variablesMapping.findIndex(
      mapping => mapping.variableKey === componentSchema.key
    );
```

This does not occur when using the current (variable) version of the Objects API. 

Form 796 on test
remove first checkboxes component

vaszig:
Closes #5701

**Changes**

When a step is edited we have a handler for the Objects API in order to update the `transformToList` array with the existing component keys or the modified ones. This is not the case in v1 since we do not use it there.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


SonnyBA (approved):
Can confirm this fixes the crash in the admin. Should this also be back-ported (since the issue mentions version `3.2.x`?

sergei-maertens (approved):

PR #6078, opened 2026-03-16
Implement require backend analysis

### LINKED ISSUE #6072: Implement "require backend" analysis for logic rules
Each rule should have a `requireBackend` flag, which is based on the actions a rule must perform (not all actions can be executed in the frontend).

List of actions, and  when the `requireBackend` flag can (possibly) be updated from `True -> False`

* Property action → never (always in frontend)
* Disable next → never (always in frontend)
* Variable action:
  * When the output variable is user defined and the input variables are static (authentication) variables, and we have executed it once in the backend → input (and in turn output) will not change after this. Note that the output variables being components, means we can always execute it in the frontend
* Step (not) applicable → never (always in frontend. Though note that the "not applicable" action requires resetting already submitted steps, which happens during the step PUT call `SubmissionSetViewSet.update`)
* Synchronize variables:
    * When the input variable is on the step before the output variable, and we have executed it once in the backend → input (and in turn output) will not change after this
* Service fetch:
    * When the input variables are static (authentication) variables, and we have executed it once in the backend → input (and in turn output) will not change after this
* Evaluate DMN:
  * When the input variables are static (authentication) variables, and we have executed it once in the backend → input (and in turn output) will not change after this
  * When the input variables are on the step before the output variable, and we have executed it once in the backend → input (and in turn output) will not change after this
* Set registration backend -> never (should be always in the backend. Though, this is never persisted, so we might be able to get away with excluding it from the analysis)

**Conclusion**
It seems like the flag can only be updated after executing logic once, and does not change during filling out a step

**Notes**
Note that whether input variable steps are before output variable steps, and input variables being static (authentication) variables can be detected during saving of logic rules. I'm afraid that doing this dynamically during serialization of the logic rules will take too long?

Besides the rules itself, we also need to analyze the formio configuration on usage of template expression. Variables can be used inside the configuration, for example to have dynamic options of a radio component. In the backend, this is done with `get_dynamic_configuration` and `inject_variables`. It probably means the form step/definition itself needs a `requireBackend` flag as well, because determining this dynamically might take too long as well?

viktorvanwijk:
Closes #6072

[skip: e2e]

**Changes**
* Added "require backend" analysis
* Updated some return types
* Removed unsupported template expression properties

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/forms/models/form_step.py
+++ b/src/openforms/forms/models/form_step.py
@@ -7,6 +7,9 @@ from django.utils.translation import gettext_lazy as _
 from autoslug import AutoSlugField
 from ordered_model.models import OrderedModel
 
+from openforms.formio.constants import DataSrcOptions
+from openforms.formio.variables import extract_variables_from_template_properties
+
 from .utils import literal_getter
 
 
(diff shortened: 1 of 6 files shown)


src/openforms/forms/models/form_step.py
```diff
@@ -116,3 +119,54 @@ def clean(self) -> None:
...
+         2. Date and datetime components having "minDate" or "maxDate" mode set to
+            "relativeToVariable".
+         3. Template expressions are used in the relevant properties. Note that the
+            value could be a list of templates because of `multiple: True`
+
+        .. note:: this assumes logic is executed before submission step serialization
+            once anyway. For example, "minDate" and "maxDate" also have mode "future",
+            which is resolved by the backend as well. This doesn't have to block us from
+            executing in the frontend, though, because the value will not change with
+            subsequent logic evaluations.
```
viktorvanwijk: I was doubting a bit whether to make this explicit by introducing a `is_logic_evaluated_once` argument to this method. Though I'm not expecting to be calling it with `is_logic_evaluated_once=False`. 

Without that argument, I guess these methods can be converted to properties

sergei-maertens (approved):
Looks good to me - some cosmetic changes.

One thing I'm wondering about if we can create some visibility into the forms that aren't easily optimized (anywhere we need to fall back to backend logic). Can you create a follow-up ticket to add some telemetry for those? I think I'd like to know which form steps and form logic rules are problematic - probably we can use an async otel metric for that but implementation details are for later.

PR #6102, opened 2026-03-20
[#6098] Introduce a single page form

### LINKED ISSUE #6098: Introduce a "single page form" variant in the form designer
Refactor the `is_appointment` field into a type field with three options: `regular`, `single_page` and `appointment`

vaszig:
Closes #6098

**Changes**

- New form `type` field (regular, appointment and single page)
- Migrations for the new field and the removed one (`is_appointment`)
- Data migration for the existing forms (update the form type)
- Necessary changes (remove old checkbox, added the new radio select for the form type) in the admin UI, the real changes will be in a different PR

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


SonnyBA (commented):
I'm not seeing any changes in this PR related to the different admin tabs that should be loaded if the new `single_page` form type is selected, as is mentioned in the parent ticket:
>Adapt the form designer UI based on the form type 

 but I'm guessing those will come in a separate PR.

The admin type filter for the admin list view also seems to be missing but this might also be added in later.

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -91,6 +91,7 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
             "literals",
             "product",
             "slug",
+            "type",
```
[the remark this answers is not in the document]
vaszig: Yes, you are rightbut https://github.com/open-formulieren/open-forms/blob/main/src/openforms/forms/api/v3/serializers/form.py#L63 doesn't help me!Basically I don't understand why the formsteps should have at least one step?In an appointment form we do not expect any steps for example.

vaszig:
> The feedback has been processed by considering #6124 is correct. I have added this code here too as we won't merge this until 3.5 is released. In case something changes I will update this PR accordingly.

#6124 has been merged and this one contains the changes too. This can be re-reviewed concerning the feedback.

vaszig:
@SonnyBA I am not sure if you have already reviewed this after processing your last feedback. So I guess now can be reviewed one more time and probably @sergei-maertens can take a look as well?

I changed it to "ready for review" since 3.5 is out now!

src/openforms/forms/migrations/0128_form_is_appointment.py
SonnyBA, line 1: This migration and the migrations after this are as far as I can tell performing the same operations as the files in `0126` and `0127` (the previous migrations). You can probably remove the last three.

SonnyBA (changes requested):

src/openforms/forms/migrations/0128_form_is_appointment.py
vaszig: Yeah probably messed up with the several rebase tries.

vaszig:
@SonnyBA @sergei-maertens my PR is back (reset that hard locally to the last healthy commit and reopened it).
FYI, the last commit has the changes regarding last Sonny's feedback about the wrong migrations.

src/openforms/forms/tests/test_migrations.py
```diff
@@ -515,3 +517,57 @@ def test_migration(self):
...
+
+
+class AddFormTypeFieldReverseMigrationTests(MigratorTestCase):
+    migrate_from = (
+        "forms",
+        "0128_remove_form_is_appointment",
+    )
+    migrate_to = (
+        "forms",
+        "0124_alter_form_new_renderer_enabled",
```
SonnyBA:
```suggestion
        "0125_alter_form_new_logic_evaluation_enabled",
```

SonnyBA (approved):
Other than using the latest migration in the migration test things are looking good 👍

src/openforms/appointments/api/serializers.py
```diff
@@ -44,7 +38,7 @@ def _appointment_plugin(self) -> BasePlugin | None:
 
     def get_supports_multiple_products(self, obj: Form) -> bool | None:
         # not an appointment -> don't bother looking up plugin configuration
-        if not obj.is_appointment:
+        if obj.type != FormTypeChoices.appointment:
             return None
         plugin = self._appointment_plugin
         return plugin.supports_multiple_products if plugin else None
```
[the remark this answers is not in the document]
sergei-maertens: tri-state logic! True/False/Undetermined.

It's the right approach here :-) the calling code is sort-of expected to `assert result is not None` to make sure they're aware of the conditions before they may call this.

src/openforms/appointments/tests/test_api_appointment_create.py
```diff
@@ -53,7 +53,9 @@ class AppointmentCreateSuccessTests(ConfigPatchMixin, SubmissionsMixin, APITestC
     def setUpTestData(cls):
         super().setUpTestData()
 
-        cls.submission = SubmissionFactory.create(form__is_appointment_form=True)
+        cls.submission = SubmissionFactory.create(
+            form__type=FormTypeChoices.appointment
+        )
```
sergei-maertens, line 58: Probably too late now, but the existing factory invocations could be kept backwards compatible with a trait on the form factory:

```py
class FormFactory(factory.django.DjangoModelFactory):
    ...

    class Params:
        is_appointment_form = factory.Trait(type=FormTypeChoices.appointment)
```

src/openforms/forms/api/serializers/form.py
```diff
@@ -520,8 +523,29 @@ def validate(self, attrs):
...
+
+        try:
+            get_plugin()
+        except ValueError:
+            raise serializers.ValidationError(
+                {
+                    "type": ErrorDetail(
+                        _(
+                            "This type of form requires an appointment plugin to be "
+                            "configured"
```
sergei-maertens:
```suggestion
                            "Appointment forms require an appointment plugin to be configured."
```

src/openforms/forms/api/v3/typing.py
```diff
@@ -138,7 +139,3 @@ class NewFormValidatedData(TypedDict):
     new_logic_evaluation_enabled: NotRequired[bool]
 
     translations: NotRequired[FormTranslationsData]
-
-
-class FormValidatedData(NewFormValidatedData):
-    uuid: UUID
```
sergei-maertens, line 144: What's the reason for this removal?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -307,29 +310,55 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
         # Fixing/updating ModelTranslationsSerializer can be tricky (it's used a lot in
         # the project), so that's why we do the check here.
         steps = attrs.get("formstep_set")
         if steps is None:
-            return attrs
-
-        is_appointment = attrs.get("is_appointment")
+            return
 
+        form_type = attrs.get("type")
```
sergei-maertens, line 321:
```suggestion
        form_type = attrs["type"]
```
or is there a reason why this key would not be required? I can see it being `None` for the same reason that `formstep_set` would be none, but then it's probably better to check for none in line 320 as well?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -307,29 +310,55 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
-        if is_appointment and len(steps) > 0:
+        if form_type == FormTypeChoices.appointment and len(steps) > 0:
             raise serializers.ValidationError(
                 _("Form steps are not allowed in an appointment form.")
             )
+        # single page form should have exactly one step
+        if form_type == FormTypeChoices.single_page and len(steps) != 1:
+            raise serializers.ValidationError(
+                _("Exactly one form step is required in a single page form.")
+            )
```
sergei-maertens: I'd write this as a match statement because the cases are mutually exclusive:

```py
num_steps = len(steps)
match form_type:
    case FormTypeChoices.regular if num_steps == 0:
        raise ...
    case FormTypeChoices.appointment if num_steps > 0:
        raise ...
    case FormTypeChoices.single_page if num_steps != 1:
        raise ...
```

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -307,29 +310,55 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+            get_plugin()
+        except ValueError:
+            raise serializers.ValidationError(
+                {
+                    "type": _(
+                        "This type of form requires an appointment plugin to be "
+                        "configured"
+                    ),
+                }
+            )
```
sergei-maertens: Seeing as this is used in two places, I think it makes sense to implement it as a DRF validator and wire that up in the two serializer classes:

```py
class RequireAppointmentsPlugin:
    ...


class FormSerializerV3(...):
    class Meta:
        ...
        extra_kwargs = {
            "type": {
                "validators": [RequireAppointmentsPlugin()],
            },
        }
```
something like that, off the top of my head

src/openforms/forms/migrations/0127_update_form_type_in_existing_forms.py
```diff
@@ -0,0 +1,53 @@
...
+    Form = apps.get_model("forms", "Form")
+
+    available_forms = Form.objects.all()
+    for form in available_forms:
+        form.type = (
+            FormTypeChoices.appointment
+            if form.is_appointment
+            else FormTypeChoices.regular
+        )
+        form.save(update_fields=["type"])
```
sergei-maertens:
```suggestion
    available_forms = Form.objects.all()
    available_forms.update(
        type=Case(
            When(is_appointment=True, then=Value(FormTypeChoices.appointment.value)),
            default=Value(FormTypeChoices.regular.value)
        ),
    )
```
bulk update query :-)

I'm not sure if the `.value` on the enum is necessary or not

src/openforms/forms/migrations/0127_update_form_type_in_existing_forms.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    The reverse data migration.
+    """
+    Form = apps.get_model("forms", "Form")
+
+    available_forms = Form.objects.all()
+    for form in available_forms:
+        if form.type == FormTypeChoices.appointment:
+            form.is_appointment = True
+            form.save(update_fields=["is_appointment"])
```
sergei-maertens: can also be a bulk update query, but even simpler:

```py
available_forms.filter(type=FormTypeChoices.appointment).update(is_appointment=True)
```

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -1722,11 +1724,16 @@ def setUp(self):
     async def test_appointment_form_nukes_irrelevant_configuration(self):
         @sync_to_async
         def setUpTestData():
+            # set up an appointment plugin
+            appointments_config = AppointmentsConfig.get_solo()
+            appointments_config.plugin = "demo"
+            appointments_config.save()
```
sergei-maertens, line 1730:
```suggestion
            appointments_config.save()
            
            self.addCleanup(AppointmentsConfig.clear_cache)
```

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -1749,8 +1756,10 @@ def setUpTestData():
             await self._admin_login(page)
             await page.goto(str(admin_url))
 
-            await page.get_by_label("Appointment enabled").click()
-            await expect(page.get_by_label("Appointment enabled")).to_be_checked()
+            await page.locator("input[name='form.type'][value='appointment']").check()
```
sergei-maertens: Surely we can keep using an accessible query with `page.get_by_label`?

src/openforms/forms/tests/factories.py
sergei-maertens, line 1: it would probably be a good idea to define a trait `is_single_step_form` that encodes all the implicit requirements/expected state (so no introduction page etc., force the `generate_minimal_setup__form_step__form_configuration__login_required=False` etc.

src/openforms/forms/tests/test_api_forms.py
```diff
@@ -1412,6 +1413,56 @@ def test_create_form_with_unknown_authentication_backend_attribute_succeeds_with
         self.assertEqual(data["backend"], "digid")
         self.assertEqual(data["options"], {"loa": DigiDAssuranceLevels.middle})
 
+    def test_appointment_type_can_be_assigned_when_plugin_configured(self):
+        config = AppointmentsConfig.get_solo()
+        config.plugin = "demo"
+        config.save()
```
sergei-maertens, line 1419: please don't forget the `self.addCleanup` to avoid future broken test isolation because caches are not rolled back between tests

src/openforms/forms/constants.py
```diff
@@ -61,3 +61,9 @@ class StatementCheckboxChoices(models.TextChoices):
     global_setting = "global_setting", _("Global setting")
     required = "required", _("Required")
     disabled = "disabled", _("Disabled")
+
+
+class FormTypeChoices(models.TextChoices):
+    regular = "regular", _("Regular")
+    appointment = "appointment", _("Appointment")
+    single_page = "single_page", _("Single page")
```
sergei-maertens: can we fix the language from "single page" to "single step"? The single page is already moot - there's at least the form fields page *and* the confirmation page

The core functionality here is that we only have the one form step that you can start filling out and submit immediately without needing additional steps to click through. Our core limitation/validation is that there's exactly one step in the form.

src/openforms/js/components/admin/form_design/constants.js
```diff
@@ -23,3 +25,9 @@ export const DMN_DECISION_DEFINITIONS_LIST = '/api/v2/dmn/decision-definitions';
 export const DMN_DECISION_DEFINITIONS_VERSIONS_LIST = '/api/v2/dmn/decision-definitions/versions';
 export const DMN_DECISION_DEFINITIONS_PARAMS_LIST =
   '/api/v2/dmn/decision-definitions/inputs-outputs';
+
+export const FORM_TYPES = [
+  ['regular', 'Regular'],
+  ['appointment', 'Appointment'],
+  ['single_page', 'Single page'],
```
sergei-maertens: same language remark here :)

And these labels need to be defined as translatable strings too!

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -651,6 +635,39 @@ const FormConfigurationFields = ({
...
+                <Radio
+                  key={`id_${value}_${index}`}
+                  name="form.type"
+                  idFor={`id_${value}_${index}`}
+                  label={
+                    <FormattedMessage
+                      defaultMessage="{formType}"
+                      description="Form type label"
+                      values={{formType: label}}
+                    />
```
sergei-maertens: yeah that won't work for translations, you'll need to `defineMessage` construct in the constants and pass that through.

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -811,6 +828,7 @@ FormConfigurationFields.propTypes = {
     internalName: PropTypes.string.isRequired,
     internalRemarks: PropTypes.string.isRequired,
     slug: PropTypes.string.isRequired,
+    type: PropTypes.string.isRequired,
```
sergei-maertens:
```suggestion
    type: PropTypes.oneOf(['regular', 'appointment', 'single_step']).isRequired,
```

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -822,7 +840,7 @@ FormConfigurationFields.propTypes = {
     askPrivacyConsent: statementChoices.isRequired,
     askStatementOfTruth: statementChoices.isRequired,
     appointmentOptions: PropTypes.shape({
-      isAppointment: PropTypes.bool.isRequired,
+      supportsMultipleProducts: PropTypes.oneOf([PropTypes.bool, null]),
```
sergei-maertens:
```suggestion
      supportsMultipleProducts: PropTypes.oneOfType([PropTypes.bool, null]),
```
I believe.

Though `PropTypes.bool` (without `isRequired`) already allows `null` too

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
vaszig: So, according to https://github.com/open-formulieren/open-forms/pull/6102#discussion_r3117888722 I thought we only want to change the language, but this makes me understand we want to change both the value and the label, is that correct?

src/openforms/appointments/tests/test_api_appointment_create.py
vaszig: I have kept the trait in the factory, just changed the code where it's used and made explicit the form type since we have the new field. Do you want me to change them?

sergei-maertens: No - both are equally valid :) You can leave it with the explicit type fields.

These trait strategies just often can help with semantic transformation for tests and reduce the impact/size of a feature/PR.

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
sergei-maertens: ah, with "the language" I mean "the terms/concepts we use", not just the translations!

I think for our collective reasoning it's best that all code references (system facing or user facing) are updated from `single page` to `single step`

sergei-maertens (approved):
Let's go!

PR #6104, opened 2026-03-20
Capture TraceID and SpanID from OTel for Sentry events

sergei-maertens:
Allowing correlation between Sentry events and traces/logs in Grafana.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/conf/utils.py
+++ b/src/openforms/conf/utils.py
@@ -3,8 +3,10 @@ import re
 from collections.abc import Callable, MutableMapping
 from typing import ClassVar
 
+from opentelemetry import trace
 from sentry_sdk.integrations import DidNotEnable, django, redis
 from sentry_sdk.integrations.logging import LoggingIntegration
+from sentry_sdk.types import Event, Hint
 
 type ConverterMapping = MutableMapping[str, Callable[[int], int]]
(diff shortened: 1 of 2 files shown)


src/openforms/conf/utils.py
```diff
@@ -102,6 +104,21 @@ def get_sentry_integrations() -> list:
...
+def sentry_before_send(event: Event, hint: Hint) -> Event | None:
+    span = trace.get_current_span()
+    if span.is_recording():
+        ctx = span.get_span_context()
+        if "extra" not in event:  # type checker doesn't like setdefault
+            event["extra"] = {}
+        event["extra"].update(
+            {
+                "otel.span_id": format(ctx.span_id, "016x"),
+                "otel.trace_id": format(ctx.trace_id, "032x"),
```
viktorvanwijk: Why the difference in format here?

viktorvanwijk (approved):
So how should these IDs be used? Is the idea is to pass the IDs from OTel to the sentry event, so that when you look at sentry, you can find the corresponding event in grafana?

PR #6117, opened 2026-03-25
[#6106] Adapt admin form designer to the new type form field

### LINKED ISSUE #6106: Adapt the form designer UI based on the new form type field
The `Form` now has a new field (`type`) which defines the type of the form, regular, appointment or a single page. With this change the admin UI has to be updated too (see the parent issue and the technical approach).

- [x] reset and hide login options
- [x] ensure only one step can be added/configured
- [x] prevent login requires on the single step
- [x] allow registration
- [x] allow logic
- [x] allow confirmation templates
- [x] disable prefill
- [x] disable/hide progress indicator display
- [x] disable submit options (always submit without overview)

vaszig:
Closes #6106

**Changes**

- [x] disable introduction page (see comment https://github.com/open-formulieren/open-forms-sdk/pull/962#discussion_r3117415271)
- [x] reset and hide login options
- [x] ensure only one step can be added/configured
- [x] prevent login requires on the single step
- [x] allow registration
- [x] allow logic (only the registration backends action for now)
- [x] allow confirmation templates
- [x] disable prefill (backend from submission and frontend in the variables tab, it's also disabled in the form builder)
- [x] disable/hide progress indicator display
- [x] disable submit options (always submit without overview)
- [x] disable new form renderer (default to true)
- [x] disable advanced configuration 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


vaszig:
~Do NOT merge until 3.5 is released.~
~Requires #6102~

src/openforms/forms/admin/mixins.py
```diff
@@ -54,7 +56,7 @@ def render_change_form(
                     {"label": label, "value": value}
                     for value, label in VertrouwelijkheidsAanduidingen.choices
                 ],
-                "form_type": "regular",
+                "form_type": obj.type if isinstance(obj, Form) else "regular",
```
vaszig: This mixin is also used by the form definition so I don't know, I would separate this even if it will be the same code. I don't know if there is a more elegant solution though because I don't really like this code.

sergei-maertens: I'll allow it, I agree with your feeling.

I'd only use the constant/enum instead of a hardcoded string for `regular`.

src/openforms/js/components/admin/form_design/data/complete-form.js
```diff
@@ -90,6 +90,23 @@ const handleAppointmentForm = draft => {
   draft.formVariables = [];
 };
 
+/**
+ * Mutate the draft state in case the form is a single page(step) form.
+ */
+const handleSinglePageForm = draft => {
```
sergei-maertens: page -> step

src/openforms/js/components/admin/form_design/logic/actions/Action.js
```diff
@@ -1,21 +1,28 @@
...
 import ToggleCodeIcon from 'components/admin/form_design/logic/ToggleCodeIcon';
-import {ACTION_TYPES} from 'components/admin/form_design/logic/constants';
+import {
+  REGULAR_FORM_ACTION_TYPES,
+  SINGLE_PAGE_FORM_ACTION_TYPES,
+} from 'components/admin/form_design/logic/constants';
 import Select from 'components/admin/forms/Select';
 import {DeleteIcon, WarningIcon} from 'components/admin/icons';
 
+import {FormContext} from '../../Context';
```
sergei-maertens: can you make this an "absolute" import? We only use relative imports for same-folder imports, as the dots are so hard to read to figure out which module is actually being loaded IMO.

src/openforms/js/components/admin/form_design/logic/constants.js
```diff
@@ -120,6 +120,15 @@ const ACTION_TYPES = [
     }),
   ],
 ];
+const SINGLE_PAGE_FORM_ACTION_TYPES = [
```
sergei-maertens:
```suggestion
const SINGLE_STEP_FORM_ACTION_TYPES = [
```

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1171,7 +1171,10 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl, outgoingRequestsUr
   // dev/debug helper
   const activeTab = new URLSearchParams(window.location.search).get('tab');
 
+  const isRegular = state.form.type === 'regular';
   const isAppointment = state.form.type === 'appointment';
+  const isSinglePage = state.form.type === 'single_step';
```
sergei-maertens:
```suggestion
  const isSingleStep = state.form.type === 'single_step';
```

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -263,7 +264,10 @@ const PresentationFields = ({
   showProgressIndicator,
   showSummaryProgress,
   onChange,
+  formType,
 }) => {
+  const isSinglePage = formType === 'single_step';
```
sergei-maertens:
```suggestion
  const isSingleStep = formType === 'single_step';
```

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -312,12 +316,13 @@ const PresentationFields = ({
           }
           helpText={
             <FormattedMessage
-              defaultMessage="Whether the step progression should be displayed in the UI or not."
+              defaultMessage="Whether the step progression should be displayed in the UI or not. This is not relevant for the single page form."
```
sergei-maertens:
```suggestion
              defaultMessage="Whether the step progression should be displayed in the UI or not. This is not relevant for the single step form."
```

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -380,15 +388,16 @@ const SubmissionFields = ({
             }
             helpText={
               <FormattedMessage
-                defaultMessage="Whether the user is allowed to submit this form or not, and whether the overview page should be shown if they are not."
+                defaultMessage="Whether the user is allowed to submit this form or not, and whether the overview page should be shown if they are not. Submission is always allowed for the single page forms."
```
sergei-maertens:
```suggestion
                defaultMessage="Whether the user is allowed to submit this form or not, and whether the overview page should be shown if they are not. Submission is always allowed for the single step forms."
```

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -550,7 +561,13 @@ const FormConfigurationFields = ({
     onChange({target: {name, value: !currentValue}});
   };
 
-  const isAppointment = type == 'appointment';
+  const isRegular = type === 'regular';
+  const isSinglePage = type === 'single_step';
```
sergei-maertens:
```suggestion
  const isSingleStep = type === 'single_step';
```

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -550,7 +561,13 @@ const FormConfigurationFields = ({
...
+  const singlePageWarningText = intl.formatMessage({
+    description: 'Single page warning icon text',
+    defaultMessage:
+      'Single page can be enabled only when there is exactly one form step configured.',
```
sergei-maertens:
```suggestion
    description: 'Single step warning icon text',
    defaultMessage:
      'Single step can be enabled only when there is exactly one form step configured.',
```

sergei-maertens (changes requested):
Here too this quite some phrasing to be updated from "single page" to "single step", also the story titles.

src/openforms/submissions/api/viewsets.py
```diff
@@ -191,7 +191,10 @@ def perform_create(self, serializer: SubmissionSerializer):  # pyright: ignore[r
 
         audit_logger.info("submission_start")
 
-        prefill_variables(submission)
+        # single step forms do not use prefill functionality
+        if form.type != FormTypeChoices.single_step:
+            prefill_variables(submission)
```
sergei-maertens, line 196: This is missing test coverage - I think that a test in the submissions app (`test_single_step_form.py`) that performs the (expected) API call sequence from the SDK would be sufficient.

Alternatively, I was thinking that an end-to-end test here would be great for this flow, but that doesn't produce coverage (by design) either, so let's do the former. Once the SDK is implemented and merged, we can also add the end-to-end test.

sergei-maertens (changes requested):

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -0,0 +1,177 @@
...
+
+        response = self.client.post(self.endpoint, body, HTTP_HOST="testserver.com")
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        m.assert_not_called()
+
+    @patch("openforms.submissions.api.viewsets.prefill_variables")
+    def test_submission_calls_prefill_when_not_single_step(self, m: MagicMock):
+        single_step_form = FormFactory.create()
```
sergei-maertens: I"m confused about the test data setup here vs the test name? Nothing makes it seems here like there are multiple steps.

I'm also wondering if the branch where prefill is called isn't covered already by the rest of the test suite? So this test may not be necessary and can probably be deleted.

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -0,0 +1,177 @@
...
+                    },
+                ]
+            },
+        )
+
+        step = submission.form.formstep_set.get()
+        self._add_submission_to_session(submission)
+
+        endpoint = reverse(
+            "api:submission-steps-validate",
```
sergei-maertens, line 47: Oh this is a left-over cleanup for the SDK - this `validate` usage should not be necessary any more as the `PUT` call itself can and should do it.

(test is okay, let's handle the above in a separate PR)

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -0,0 +1,177 @@
...
+
+        response = self.client.post(endpoint, {"data": {"foo": "d"}})
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    @patch("openforms.submissions.api.mixins.on_post_submission_event")
+    @freeze_time("2026-04-29T10:53:19+01:00")
+    def test_complete_submission_in_single_step(self, mock_on_post_submission_event):
+        form = FormFactory.create(is_single_step_form=True)
+        step = FormStepFactory.create(form=form)
```
sergei-maertens: doesn't this create a second step? I'd expect `is_single_step_form=True` to already set up the step.

src/openforms/submissions/tests/test_single_step_form.py
sergei-maertens, line 1: You've now broken out everything per API call/endpoint, but I think that basically a single test with multiple calls should be sufficient, as that's what the frontend will be doing too:

1. Create submission
2. Call validate endpoint with step data
3. Submit step data (PUT)
4. Complete form

This way you don't have to manually manage the submission factory & add it to the session - we know that those mechanisms already work from other tests. It's particular sequence of API calls to make (and the assertion that prefill is not invoked, even though an initial data reference is passed) that's important.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #6124, opened 2026-03-26
[#6119] Fix min length of steps in FormSerializer v3

### LINKED ISSUE #6119: Form serializer in API v3 does not allow zero steps

latest

### Customer reference

_No response_

The steps in the `FormSerializer` have `min_length=1` which does not help with appointments and single page form. Appointments do not have steps and single page needs exactly one step. 

https://github.com/open-formulieren/open-forms/blob/main/src/openforms/forms/api/v3/serializers/form.py#L63

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #6119

**Changes**

The previous min_length=1 was ok for regular forms but does not work for the appointment forms where no steps are defined. We explicitly remove all the steps in the frontend. This is now done in the validate method where we can access both the is_appointment and the formstep_set.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/forms/api/v3/serializers/form.py
+++ b/src/openforms/forms/api/v3/serializers/form.py
@@ -61,7 +61,7 @@ class FormSerializer(serializers.ModelSerializer):
         slug_field="uuid",
     )
 
-    steps = FormStepSerializer(many=True, source="formstep_set", min_length=1)  # pyright: ignore[reportCallIssue]
+    steps = FormStepSerializer(many=True, source="formstep_set")
 
     payment = FormPaymentSerializer(required=False, source="*")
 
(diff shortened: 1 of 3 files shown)


src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -303,6 +303,31 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
 
         return value
 
+    def validate(self, attrs):
```
vaszig: Of course this is going to be updated in #6102 PR in favor of all the types of form (regular, appointment and single page).

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -303,6 +303,31 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+        steps = attrs.get("formstep_set")
+
+        # validate is called multiple times because of the nested serializer fileds and
+        # this means that steps will not exist at every call
+
+        if steps is None:
```
SonnyBA:
```suggestion
        # Validate is called multiple times because of the nested serializer fields,
        # therefore the `step` field will not be present every call.
        if steps is None:
```

sergei-maertens (approved):

PR #6154, opened 2026-04-01
Replace `is_layout_component` with `holds_submission_data`

### LINKED ISSUE #6018: Tidy up code after the hotfixes
Via #5980, #6001, #6005, #6007 and #6014 we did some hotfixes that are probably not the most elegant.

Collection of topics (non-exhaustive):

- [x] https://github.com/open-formulieren/open-forms/pull/6017#discussion_r2864470214 (#6048)
- [x] https://github.com/open-formulieren/open-forms/pull/6017#discussion_r2864489975 (#6048)
- [x] handling of `initial_data` vs `original_input_data` (#6048)
- [x] Replace calls to `Submission.data` (#6129) and `SubmissionStep.data` (#6127) with `SubmissionValueVariablesState.get_data(...)`
- [x] See if we can replace `is_layout_component(component)` with `register.holds_submission_data(component)` (#6154)

It's probably wise to wrap up the logic-to-frontend epic first though.

viktorvanwijk:
Partly closes #6018, #6164

[skip: e2e]

**Changes**

* Added some extra regression tests
* Removed compatibility shim of `build_serializer_fields` for non-registered components
* Replaced `is_layout_component` with `holds_submission_data`
* Removed `is_layout_component` 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -411,6 +411,9 @@ class NPFamilyMembers(BasePlugin):
     # not actually relevant, as we transform the component into a different type
     formatter = DefaultFormatter
 
+    def build_serializer_field(self, component: Component) -> serializers.Field:
+        raise NotImplementedError()
+
     @staticmethod
     def _get_handler() -> FamilyMembersHandler:
         handlers = {
@@ -1095,6 +1098,9 @@ class CustomerProfile(BasePlugin):
 class SoftRequiredErrors(BasePlugin):
     holds_submission_data = False
 
+    def build_serializer_field(self, component: Component) -> serializers.Field:
+        raise NotImplementedError()
+
     @staticmethod
     def as_json_schema(component: Component) -> None:
         return None
@@ -1103,6 +1109,9 @@ class SoftRequiredErrors(BasePlugin):
 @register("coSign")
 class CoSign(BasePlugin):
     holds_submission_data = False
+
+    def build_serializer_field(self, component: Component) -> serializers.Field:
+        raise NotImplementedError()
 
     @staticmethod
     def as_json_schema(component: Component) -> None:
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -1071,6 +1071,12 @@ class LicensePlate(BasePlugin):
 class CustomerProfile(BasePlugin):
     formatter = CustomerProfileFormatter
 
+    def build_serializer_field(
+        self, component: CustomerProfileComponent
+    ) -> serializers.JSONField:
+        required = component.get("validate", {}).get("required", False)
+        return serializers.JSONField(required=required, allow_null=True)
+
     @staticmethod
     def pre_registration_hook(
         component: CustomerProfileComponent, submission: Submission
--- a/src/openforms/formio/registry.py
+++ b/src/openforms/formio/registry.py
@@ -12,7 +12,7 @@ smeared out across the codebase in similar but different implementations, while 
 the public API better defined and smaller.
 """
 
-import warnings
+import abc
 from collections.abc import Callable
 from typing import (
     TYPE_CHECKING,
@@ -37,7 +37,6 @@ from openforms.typing import JSONObject, JSONValue, VariableValue
 
 from .datastructures import FormioConfigurationWrapper, FormioData
 from .typing import Component
-from .utils import is_layout_component
 
 if TYPE_CHECKING:
     from openforms.submissions.models import Submission
@@ -69,7 +68,7 @@ class PreRegistrationHookProtocol(Protocol[ComponentT]):
     ) -> ComponentPreRegistrationResult: ...
 
 
-class BasePlugin(Generic[ComponentT], AbstractBasePlugin):  # noqa: UP046
+class BasePlugin(Generic[ComponentT], AbstractBasePlugin, abc.ABC):  # noqa: UP046
     """
     Base class for Formio component plugins.
     """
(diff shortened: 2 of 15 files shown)


src/openforms/formio/registry.py
```diff
@@ -114,31 +112,8 @@ def localize(self, component: ComponentT, language_code: str, enabled: bool):
...
-                validate.get("required", False)
-                if (validate := component.get("validate"))
-                else False
-            )
-
-        # Allow anything that is valid JSON, taking into account the 'required'
-        # validation which is common for most components.
-        return serializers.JSONField(required=required, allow_null=True)
+        """Translate a component configuration into a serializer."""
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
viktorvanwijk: 3.5.0 is out now - let's yolo this

src/openforms/formio/components/custom.py
```diff
@@ -1068,6 +1068,17 @@ def as_json_schema(component: Component) -> JSONObject:
...
     formatter = CustomerProfileFormatter
 
+    def build_serializer_field(
+        self, component: CustomerProfileComponent
+    ) -> serializers.JSONField:
+        required = (
+            validate.get("required", False)
+            if (validate := component.get("validate"))
+            else False
+        )
```
annashamray:
```suggestion
        required = component.get("validate", {}).get("required", False)
```

src/openforms/formio/tests/validation/test_generic.py
```diff
@@ -1,26 +1,10 @@
 from django.test import SimpleTestCase
 
-from openforms.typing import JSONValue
-
 from ...typing import Component
 from .helpers import extract_error, validate_formio_data
 
 
-class FallbackBehaviourTests(SimpleTestCase):
-    def test_unknown_component_passthrough(self):
```
annashamray, line 10: Why do you remove the test? It should now raise NotImplemented exception, right?

src/openforms/formio/serializers.py
```diff
@@ -31,7 +31,10 @@
 
 class StepDataSerializer(serializers.Serializer):
     def apply_hidden_state(
-        self, configuration: FormioConfiguration, fields: dict[str, FieldOrNestedFields]
+        self,
+        configuration: FormioConfiguration,
+        fields: dict[str, FieldOrNestedFields],
+        register: ComponentRegistry,
```
annashamray, line 37: Do you add this parameter only to for `register.holds_submission_data` check? 
But why not to use `formio.service.holds_submission_data` function? Or is it a different register?

src/openforms/forms/tests/variables/test_model.py
```diff
@@ -935,3 +935,56 @@ def test_synchronize_for_correctness(self):
                     self.assertEqual(textfield_variable.data_format, "")
                     self.assertFalse(textfield_variable.is_sensitive_data)
                     self.assertEqual(textfield_variable.initial_value, [])
+
+    def test_with_component_which_do_not_hold_submission_data(self):
```
annashamray: Could you please add a test with layout component + nested regular component, like fieldset + nested textfield? Just to make sure that the form variable for textfield would be created in this case

src/openforms/submissions/tests/test_submission_step_validate.py
```diff
@@ -1470,3 +1471,35 @@ def test_radio_with_dynamic_options_from_another_component_form_variable(self):
             },
         )
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    @tag("gh-6155")
+    @expectedFailure
```
annashamray, line 1476: Nice one! We check prefill variable config in `FormVariableListSerializer` for profile component, but not the component itself

annashamray (changes requested):

src/openforms/formio/serializers.py
viktorvanwijk: I added this here because it is possible to pass a custom registry to `build_serializer_field`. Using `formio.service.holds_submission_data` will use the default registry, which _could_ contain differences

src/openforms/formio/tests/validation/test_generic.py
viktorvanwijk: Yeah wasn't sure if it was worth testing tbh - will revise instead of remove it

src/openforms/formio/components/custom.py
viktorvanwijk: I just copied this from the original implementation and didn't bother checking if it could be more efficient, but I can update it. It's temporary anyway ;)

src/openforms/forms/tests/variables/test_model.py
viktorvanwijk: Will do! I'm surprised that doesn't exist yet tbh.

EDIT: to be fair though, I suspect a LOT of tests will fail if this weren't the case :p

viktorvanwijk: Very useful actually, as it uncovered an incorrect test setup for this one as well :). Needed to create a step to ensure the form is created too

src/openforms/formio/serializers.py
annashamray: Thanks for the explanation

annashamray (approved):

src/openforms/formio/serializers.py
sergei-maertens: having the register as argument works really nice in tests for dependency injection and test isolation :-)

src/openforms/formio/registry.py
```diff
@@ -114,31 +112,8 @@ def localize(self, component: ComponentT, language_code: str, enabled: bool):
         pass  # noop by default, specific component types can extend the base behaviour
 
     def build_serializer_field(self, component: ComponentT) -> serializers.Field:
```
sergei-maertens, line 116: you can also decorate the method with `@abstractmethod` from `abc` to force subclasses to implement the method.

sergei-maertens (approved):

src/openforms/formio/registry.py
viktorvanwijk: That completely slipped my mind :sweat_smile: It will definitely help prevent issues like #6155

PR #6158, opened 2026-04-02
Script logic rule cycle detection

### LINKED ISSUE #6150: Add helper/detection script for logic analysis
We soon have `bin/report_invalid_form_logic.py` that checks for invalid configurations. This should be extended to call `form.apply_logic_analysis` for each form and check if there are any logic cycles reported, as those would prevent migrating to the new logic evaluation.

Note that this by default saves the result in the database, so:

* it should run in a transaction and roll back the transaction savepoint
* catch the `CyclesDetected` exception and report for which forms those happened

For now, it's in diagnostics mode, but for 4.0 this will be a hard requirement that existing forms need to be fixed before we can migrate.

The script needs to be documented in the changelog (cc @annashamray )

sergei-maertens:
Closes #6150

[skip: e2e]

**Changes**

* Test all forms with legacy logic evaluation for cycles
* Report problematic forms

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,18 @@ To upgrade to 3.5, please:
 * ⚠️ If you haven't done so yet, please review the manual intervention mentioned in the
   3.4.2 release notes.
 
+* Inventarize the forms that are not compatible yet with the new logic evaluation -
+  we've included a utility script to report the problematic forms. You can run it with:
+
+  .. code-block:: bash
+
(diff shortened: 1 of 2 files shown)


viktorvanwijk (approved):
I have tested this on my local env, and only found two cases with (self) cycles. Both had a variable action that used a single variable as input and output, so the fix in #6160 is not relevant for my env

PR #6186, opened 2026-04-13
[#5959] add support for variables to v3 form endpoint

### LINKED ISSUE #5959: Add support for variables

SonnyBA:
Closes #5959

[skip: e2e]

**Changes**

Adds the `variables` field to the `v3` form endpoint.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/forms/api/v3/serializers/form.py
+++ b/src/openforms/forms/api/v3/serializers/form.py
@@ -1,4 +1,4 @@
-from collections import Counter
+from collections import Counter, defaultdict
 from uuid import UUID
 
 from django.db import transaction
@@ -7,7 +7,7 @@ from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.utils import extend_schema_serializer
 from rest_framework import serializers
-from rest_framework.exceptions import ValidationError
+from rest_framework.exceptions import ErrorDetail, ValidationError
 
 from openforms.appointments.api.serializers import AppointmentOptionsSerializer
 from openforms.config.models import Theme
@@ -18,9 +18,16 @@ from openforms.formio.service import (
     FormioConfigurationWrapper,
     get_readable_path_from_configuration_path,
 )
+from openforms.formio.typing.base import Component
+from openforms.prefill.contrib.customer_interactions.constants import (
+    PLUGIN_IDENTIFIER as COMMUNICATION_PREFERENCES_PLUGIN_IDENTIFIER,
+)
 from openforms.products.models import Product
 from openforms.translations.api.serializers import ModelTranslationsSerializer
 from openforms.typing import StrOrPromise
+from openforms.variables.constants import FormVariableSources
+from openforms.variables.models import ServiceFetchConfiguration
+from openforms.variables.service import get_static_variables
 
 from ....api.serializers.form import (
     FormLiteralsSerializer,
@@ -37,9 +44,10 @@ from ....models import (
     FormVariable,
 )
 from ...validators import RequireAppointmentsPlugin
-from ..typing import FormStepData, FormValidatedData
+from ..typing import FormStepData, FormValidatedData, FormVariableData
 from .form_step import FormStepSerializer
 from .payment import FormPaymentSerializer
+from .variables import FormVariableSerializer
 
 
 @extend_schema_serializer(component_name="FormV3Serializer")
@@ -64,6 +72,9 @@ class FormSerializer(serializers.ModelSerializer):
     )
 
     steps = FormStepSerializer(many=True, required=True, source="formstep_set")
+    variables = FormVariableSerializer(
+        many=True, source="formvariable_set", required=False
+    )
 
     payment = FormPaymentSerializer(required=False, source="*")
 
@@ -91,8 +102,11 @@ class FormSerializer(serializers.ModelSerializer):
     _nested_fields = (
         "confirmation_email_template",
         "formstep_set",
+        "formvariable_set",
         "registration_backends",
     )
+
+    form_definition_configurations: dict[UUID, FormioConfigurationWrapper]
 
     class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
         model = Form
@@ -104,6 +118,7 @@ class FormSerializer(serializers.ModelSerializer):
             "login_required",
             "translation_enabled",
             "registration_backends",
+            "variables",
             "payment",
             "appointment_options",
             "literals",
@@ -165,12 +180,14 @@ class FormSerializer(serializers.ModelSerializer):
         )
         # fmt:on
         len(form_definitions)  # Evaluate the queryset to acquire the locks.
+        form_definitions_created: dict[UUID, FormDefinition] = {}
         for index, step_data in enumerate(form_step_data):
             form_definition_data = step_data["form_definition"]
             form_definition, _ = form_definitions.update_or_create(
                 uuid=form_definition_data["uuid"],
                 defaults={k: v for k, v in form_definition_data.items() if k != "uuid"},
             )
+            form_definitions_created[form_definition_data["uuid"]] = form_definition
 
             FormStep.objects.create(
                 **{**step_data, "form_definition": form_definition},
@@ -178,11 +195,43 @@ class FormSerializer(serializers.ModelSerializer):
                 form=instance,
             )
 
+        # These calls are required to create the corresponding component form variables
+        # from the form definitions.
+        for form_definition in form_definitions_created.values():
+            FormVariable.objects.synchronize_for(form_definition)
+
         registration_backends = validated_data.get("registration_backends", [])
         FormRegistrationBackend.objects.bulk_create(
             FormRegistrationBackend(form=instance, **backend)
             for backend in registration_backends
         )
+
+        form_variables_data = validated_data.get("formvariable_set", [])
+        form_variables: list[FormVariable] = []
+        for variable_data in form_variables_data:
+            service_fetch_configuration = None
+            if service_configuration_data := variable_data.pop(
+                "service_fetch_configuration", None
+            ):
+                service_fetch_configuration_id = service_configuration_data.pop(
+                    "id", None
+                )
+                service_fetch_configuration, __ = (
+                    ServiceFetchConfiguration.objects.update_or_create(
+                        id=service_fetch_configuration_id,
+                        defaults=service_configuration_data,
+                    )
+                )
+
+            variable_kwargs = dict(variable_data)
+            form_variable = FormVariable(
+                form=instance,
+                service_fetch_configuration=service_fetch_configuration,
+                **variable_kwargs,
+            )
+            form_variable.check_data_type_and_initial_value()
+            form_variables.append(form_variable)
+        FormVariable.objects.bulk_create(form_variables)
 
         return instance
 
@@ -210,12 +259,14 @@ class FormSerializer(serializers.ModelSerializer):
         )
         # fmt:on
         len(form_definitions)  # Evaluate the queryset to acquire the locks.
+        form_definitions_created: dict[UUID, FormDefinition] = {}
         for index, step_data in enumerate(form_step_data):
             form_definition_data = step_data["form_definition"]
             form_definition, _ = form_definitions.update_or_create(
                 uuid=form_definition_data["uuid"],
                 defaults={k: v for k, v in form_definition_data.items() if k != "uuid"},
             )
+            form_definitions_created[form_definition.uuid] = form_definition
 
             step = FormStep(  # TODO: use update_or_create (nice to have)
                 **{**step_data, "form_definition": form_definition},
@@ -224,16 +275,30 @@ class FormSerializer(serializers.ModelSerializer):
             )
             assigned_steps.append(step)
 
+        # Remove the form steps not that are not part of the request along with their
+        # form definitions when applicable.
         steps_to_delete = instance.formstep_set.exclude(
             pk__in=(step.pk for step in assigned_steps)
         )
-        for step in steps_to_delete:
-            step.delete()  # Removes the form definition when applicable by calling `.delete`.
+        form_definitions_to_delete = [
+            step.form_definition.pk
+            for step in steps_to_delete
+            if not step.form_definition.is_reusable
+            and step.form_definition.uuid not in form_definitions_created
+        ]
+        # Use query manager delete methods to bypass model defined `delete` methods.
+        steps_to_delete.delete()
+        FormDefinition.objects.filter(pk__in=form_definitions_to_delete).delete()
 
         # Generate form steps after deleting existings steps, to allow correct calculation of
         # the `order` field.
         for step in sorted(assigned_steps, key=lambda step: step.order):
             step.save()
+
+        # These calls are required to create the corresponding component form variables
+        # from the form definitions.
+        for form_definition in form_definitions_created.values():
+            FormVariable.objects.synchronize_for(form_definition)
 
         registration_backends = validated_data.get("registration_backends", None)
         if registration_backends is not None:
@@ -242,6 +307,35 @@ class FormSerializer(serializers.ModelSerializer):
                 FormRegistrationBackend(form=instance, **backend)
                 for backend in registration_backends
             )
+
+        form_variables_data = validated_data.get("formvariable_set", [])
+        form_variables: list[FormVariable] = []
+        for variable_data in form_variables_data:
+            service_fetch_configuration = None
+            if service_configuration_data := variable_data.pop(
+                "service_fetch_configuration", None
+            ):
+                service_fetch_configuration_id = service_configuration_data.pop(
+                    "id", None
+                )
+                service_fetch_configuration, __ = (
+                    ServiceFetchConfiguration.objects.update_or_create(
+                        id=service_fetch_configuration_id,
+                        defaults=service_configuration_data,
+                    )
+                )
+
+            variable_kwargs = dict(variable_data)
+            form_variable = FormVariable(
+                form=instance,
+                service_fetch_configuration=service_fetch_configuration,
+                **variable_kwargs,
+            )
+            form_variable.check_data_type_and_initial_value()
+            form_variables.append(form_variable)
+        # Remove the stale variables that were not part of the request.
+        instance.formvariable_set.exclude(source=FormVariableSources.component).delete()
+        FormVariable.objects.bulk_create(form_variables)
 
         return instance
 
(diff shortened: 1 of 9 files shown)


src/openforms/forms/models/form.py
```diff
@@ -422,6 +423,7 @@ class Form(models.Model):
         FormManager
     ] = FormManager()
     formstep_set: models.Manager[FormStep]
+    formvariable_set: models.Manager[FormVariable]
```
annashamray, line 426: Nice one!

src/openforms/forms/api/v3/serializers/variables.py
```diff
@@ -0,0 +1,39 @@
...
+from rest_framework import serializers
+
+from openforms.forms.models import FormVariable
+
+from .service_fetch_config import ServiceFetchConfigurationSerializer
+
+
+@extend_schema_serializer(component_name="FormVariableV3Serializer")
+class FormVariableSerializer(serializers.ModelSerializer):
+    form_definition = serializers.UUIDField(required=False, allow_null=True)
```
annashamray: Can `RelatedField` be used here? 

```suggestion
    form_definition = serializers.SlugRelatedField(queryset=FormDefinition.objects.all(), slug_field="uuid")
```

In this case you wont' need to use `to_representation` here, and it would be more clear, that these fields are related

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -311,6 +420,153 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+        index: int,
+        variable_data: FormVariableData,
+    ) -> None:
+        prefill_options = variable_data.get("prefill_options", {})
+        if not (
+            profile_form_variable_key := prefill_options.get("profile_form_variable")
+        ):
+            return
+
+        if not (profile_form_definition_uuid := variable_data.get("form_definition")):
```
annashamray: Hmmm I think that user_defined variables never have `form_definition`, it can only be applicable for component variables. 
Is it going to be changed in v3 ?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -165,25 +177,69 @@ def create(self, validated_data: FormValidatedData) -> Form:
...
             FormRegistrationBackend(form=instance, **backend)
             for backend in registration_backends
         )
 
+        form_variables_data = validated_data.get("formvariable_set", [])
+        form_variables: list[FormVariable] = []
+        for variable_data in form_variables_data:
+            form_definition = None
+            form_variable_kwargs = {
+                k: v for k, v in variable_data.items() if k != "form_definition"
```
annashamray: Minus one dict comprehension :) 

```suggestion
            form_definition = variable_data.pop("form_definition", None)
```

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -228,13 +286,23 @@ def update(self, instance: Form, validated_data: FormValidatedData) -> Form:
             pk__in=(step.pk for step in assigned_steps)
         )
         for step in steps_to_delete:
-            step.delete()  # Removes the form definition when applicable by calling `.delete`.
+            delete_form_definition = not bool(
+                step.form_definition.uuid in form_definitions_created
+            )
```
annashamray:
```suggestion
            delete_form_definition = step.form_definition.uuid not in form_definitions_created
```

src/openforms/forms/models/form_step.py
```diff
@@ -157,9 +157,14 @@ def is_backend_logic_evaluation_required(self) -> bool:
         return False
 
     def delete(self, *args, **kwargs):
+        delete_form_definition = kwargs.pop("delete_form_definition", True)
```
annashamray: Tbh I'm not sure about reusing `delete` method here, the added behaviour is quite different from the generic one. 
Maybe just move this code to FormSerializer?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -311,6 +420,153 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+            for configuration in self.form_definition_configurations.values()
+            for component in configuration
+        ]
+        existing_profile_form_vars: list[str] = []
+        for index, variable_data in enumerate(variables_data):
+            # To have a smooth transition in the front-end from v2 to v3, only user-defined
+            # variables will be processed/saved from the request. Component variable
+            # are automagically created/updated when the form gets saved.
+            if variable_data["source"] == FormVariableSources.component:
+                continue
```
annashamray, line 508: Won't it be easier to remove component variables from attrs here, in this case you won't need to do the same in `create` and `update` methods?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -243,6 +311,45 @@ def update(self, instance: Form, validated_data: FormValidatedData) -> Form:
...
+            form_variable = FormVariable(
+                form=instance,
+                service_fetch_configuration=service_fetch_configuration,
+                **{
+                    k: v
+                    for k, v in form_variable_kwargs.items()
+                    if k != "service_fetch_configuration"
+                },
+            )
+            form_variable.check_data_type_and_initial_value()
```
annashamray, line 334: is it possible to move this check to the validation step?

annashamray (changes requested):

src/openforms/forms/api/v3/serializers/form.py
sergei-maertens: correct, there should even be a database constraint for that if I'm not mistaken

src/openforms/forms/api/v3/serializers/variables.py
SonnyBA: I don't think we'll be able to use that here, at that point (during validation) the form definition does not exist yet.

src/openforms/forms/api/v3/serializers/variables.py
```diff
@@ -0,0 +1,39 @@
...
+    service_fetch_configuration = ServiceFetchConfigurationSerializer(
+        required=False, allow_null=True
+    )
+    prefill_options = serializers.JSONField(required=False)
+
+    def to_representation(self, instance: FormVariable) -> dict:
```
sergei-maertens: can't you use the `source` argument on the serializer field instead of overriding `to_representation`?

```py
form_definition = serializers.UUIDField(required=False, allow_null=True, source="form_definition.uuid")
```

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: I removed the form definition validation for the user defined variables but now that the `form_definition` field does not really serve any purpose anymore I suggest we remove it (this will be a breaking change though for the front-end).

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: I had to cast the typed dict into a regular dict here, before instantiating the `FormVariable` instance, to make Pyright happy.

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: I'm not sure what the benefit(s) of that would be as the method sets values on a `FormVariable` instance and does not perform any validation :thinking:

src/openforms/forms/api/v3/serializers/variables.py
SonnyBA: Seems like you can't use it when the name matches that of the field:
```
It is redundant to specify `source='form_definition'` on field 'UUIDField' in serializer 'FormVariableSerializer', because it is the same as the field name. Remove the `source` keyword argument.
```

sergei-maertens: did you miss the `.uuid` suffix? :smile:

SonnyBA: > did you miss the `.uuid` suffix? 😄

That fixed it :+1: and also outputs to the expected uuid:
```
....
 'variables': [{'dataFormat': '',
                'dataType': 'array',
                'formDefinition': 'f8ee3ca4-9a20-4ce0-a02b-8b31d8af5fc0',
                'initialValue': None,
                'isSensitiveData': False,
                'key': 'profile',
                'name': 'profile',
                'prefillAttribute': '',
                'prefillIdentifierRole': 'main',
                'prefillOptions': {},
                'prefillPlugin': '',
                'serviceFetchConfiguration': None,
                'source': 'component'},
.....
```

src/openforms/forms/api/v3/serializers/form.py
annashamray: Right, I though it checks initial value, but I see that it doesn't raise any ValidationError
Please ignore the comment above

src/openforms/forms/api/v3/serializers/form.py
annashamray: I don't know how frontend will use the response from this endpoint, but perhaps `form_definition` should be left as read-only field? In this case frontend will know what form-definition component variables belong to. 

@SonnyBA @sergei-maertens what do you think?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -165,25 +178,53 @@ def create(self, validated_data: FormValidatedData) -> Form:
...
+        for variable_data in form_variables_data:
+            service_fetch_configuration = None
+            if service_configuration_data := variable_data.pop(
+                "service_fetch_configuration", None
+            ):
+                service_fetch_configuration = ServiceFetchConfiguration.objects.create(
+                    **service_configuration_data
+                )
+
+            variable_kwargs = dict(variable_data)
```
annashamray, line 226: I don't like it but I can't think of more elegant solution to keep pyright happy, so please ignore my rambling

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: I removed the field in 2c120327e but I think leaving it in as a read-only field might make things easier for the front-end like you mentioned :+1:

src/openforms/forms/api/v3/serializers/form.py
annashamray: I'm not sure about removing list items during the loop through the same list. Maybe copying a list and iterating on the copy would be a safer approach?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -243,6 +300,29 @@ def update(self, instance: Form, validated_data: FormValidatedData) -> Form:
...
+        form_variables_data = validated_data.get("formvariable_set", [])
+        form_variables: list[FormVariable] = []
+        for variable_data in form_variables_data:
+            service_fetch_configuration = None
+            if service_configuration_data := variable_data.pop(
+                "service_fetch_configuration", None
+            ):
+                service_fetch_configuration = ServiceFetchConfiguration.objects.create(
+                    **service_configuration_data
+                )
```
sergei-maertens, line 326: shouldn't this be update or create? what happens if you update a form that already has a service fetch configuration wired up? How does that compare to the current situation?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -243,6 +300,29 @@ def update(self, instance: Form, validated_data: FormValidatedData) -> Form:
...
+            form_variable = FormVariable(
+                form=instance,
+                service_fetch_configuration=service_fetch_configuration,
+                **variable_kwargs,
+            )
+            form_variable.check_data_type_and_initial_value()
+            form_variables.append(form_variable)
+        # Remove the stale variables that were not part of the request.
+        instance.formvariable_set.exclude(source=FormVariableSources.component).delete()
+        FormVariable.objects.bulk_create(form_variables)
```
sergei-maertens, line 338: right, that answers my question, we delete whatever is there and re-insert 😄 

We'll probably have to check at some point if we can clean up unused service fetches 😬

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -260,6 +340,8 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
             )
             for step in value
         }
+        # Required for the variable validation.
+        self.form_definition_configurations = configurations
```
sergei-maertens, line 358: this is missing a type annotation in the class, right? Though it suprises me a bit that the type checker does not complain about it since the whole `src/openforms/forms/api/v3/` path is in the pyright config 🤔

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -311,6 +393,139 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+                    {
+                        f"variables.{index}.key": [
+                            (
+                                "The variable key cannot be equal to any of the "
+                                "following component variable keys: {component_keys}."
+                            ).format(component_keys=", ".join(component_keys)),
+                        ]
+                    },
+                    code="unique",
+                )
```
sergei-maertens, line 526: For better UX, I'd suggest collecting all the errors that you can and then raise them all at once. Now you're forcing the user to correct the first mistake, re-submit, then fix the second, re-submit etc... whereas they could receive all the error information in one go.

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -311,6 +393,139 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+        prefill_options = variable_data.get("prefill_options", {})
+        if not (
+            profile_form_variable_key := prefill_options.get("profile_form_variable")
+        ):
+            return
+
+        component: Component | None = None
+        for form_definition in self.form_definition_configurations.values():
+            if profile_form_variable_key not in form_definition:
+                continue
```
sergei-maertens: I'd replace `form_definition` with `form_configuration_wrapper` or `form_configuration` - reading the code in Github gave me the impression that it was a `FormDefinition` model instance.

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -1770,6 +1825,1164 @@ def test_unsupported_post(self):
         self.assertEqual(Form.objects.count(), 0)
 
 
+class FormEndpointVariableTests(APITestCase):
+    @classmethod
+    def setUpTestData(cls) -> None:
+        cls.admin_user = UserFactory.create(
+            is_staff=True, user_permissions=("forms.change_form",)
+        )
```
sergei-maertens, line 1836: don't forget to call the `super()` method

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: Added the `form_definition` field back in as a read-only field in 2474d2792f80b8ee69cd09f642ca7a4056bdecd7.

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -310,6 +393,139 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+        index: int,
+        variable_data: FormVariableData,
+    ) -> None:
+        prefill_options = variable_data.get("prefill_options", {})
+        if not (
+            profile_form_variable_key := prefill_options.get("profile_form_variable")
+        ):
+            return
+
+        component: Component | None = None
```
annashamray, line 423: I think this line is not used

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -310,6 +393,139 @@ def validate_steps(self, value: list[FormStepData]) -> list[FormStepData]:
...
+            if profile_form_variable in existing_profile_form_vars:
+                raise ValidationError(
+                    {
+                        f"variables.{index}.prefillOptions.profileFormVariable": _(
+                            "This profile form variable is already used in another "
+                            "communication preferences prefill plugin."
+                        )
+                    },
+                    code="unique",
+                )
```
annashamray: I think you can also move this validation to `validate_variable_profile_options`, because it also checks profile options

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -1767,6 +1825,1164 @@ def test_unsupported_post(self):
         self.assertEqual(Form.objects.count(), 0)
 
 
+class FormEndpointVariableTests(APITestCase):
```
annashamray, line 1829: Nice tests

annashamray (commented):
Looks good, just a couple of small remarks

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: [Changed this](https://github.com/open-formulieren/open-forms/compare/2474d2792f80b8ee69cd09f642ca7a4056bdecd7..d60af47db7db2eaeb8ef50f395f20bd345058891#diff-6be5244bfd1f414f9c309480c540a4e04dd1b56b8d936235a6157e5323754fabL475) by reassigning the field with a new filtered list of components.

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: I added the same behavior `v2` uses [in the most recent pushed commit](https://github.com/open-formulieren/open-forms/compare/2474d2792f80b8ee69cd09f642ca7a4056bdecd7..d60af47db7db2eaeb8ef50f395f20bd345058891#diff-6be5244bfd1f414f9c309480c540a4e04dd1b56b8d936235a6157e5323754fabR216). Should we use/add an slug field for this :thinking:?

src/openforms/forms/api/v3/serializers/form.py
SonnyBA: [It is used to check the component type of the specified profile component key](https://github.com/open-formulieren/open-forms/compare/2474d2792f80b8ee69cd09f642ca7a4056bdecd7..d60af47db7db2eaeb8ef50f395f20bd345058891#diff-6be5244bfd1f414f9c309480c540a4e04dd1b56b8d936235a6157e5323754fabR435) (it should be of the type `profile`).

src/openforms/forms/api/v3/serializers/form.py
sergei-maertens: The frontend doesn't use it afaik, but that's not easy to see :) Either way, we can leave it as is per Sonny's last commit, future cleanups once we overhaul some frontend code can be done later.

src/openforms/forms/api/v3/serializers/form.py
sergei-maertens: Let's leave it at that for now - this whole service-fetch-via-logic was never the original intention. If we move things around, I think it belongs in the logic action definition itself, but there's no hurry for that now and would only break form exports/imports more.

sergei-maertens (approved):

annashamray (approved):

PR #6189, opened 2026-04-15
:fire: remove camunda registration backend

### LINKED ISSUE #5223: Remove Camunda registration backend
The Camunda registration backend does not appear to actually be used by anyone. Utrecht was the organization that initially requested it as part of the Open Forms 1.0 feature set, but they have since settled on the "productaanvraag" pattern via the Objects API, combined with CommUtr/GZAC which has their own BPMN engine inside.

The code for this registration backend is outdated and we can't justify putting in resources for something that doesn't appear to be used without financial backing. It's potentially blocking library upgrades and lags behind in UX improvements. We can't simply move it to an extension either, as it relies on custom JS for the registration options configuration for which no pluggable mechanism exists (and building one for an unused feature is not a good use of our resources).

Therefore, we have decided to remove this in Open Forms 4.0 (unknown planned release date), and will deprecate it now.

annashamray:
Closes #5223

**Changes**
* remove camunda registration backend from code and docs

**Checklist**

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form - **When the form with camunda registration backend is imported, the validation error is raised now, but it's expected**

- Dockerfile/scripts

- Commit hygiene

- Documentation


docs/configuration/dmn/camunda.rst
```diff
@@ -1,11 +1,65 @@
...
+**Security**
+IP address                    The IP address of the Open Forms server (optional, for whitelisting).
+============================  =======================================================================================
+
+
+Plugin configuration
+====================
+
+1. In the Open Forms admin, navigate to: **Configuration** > **Camunda configuration**
+2. In the **Camunda root** field, fill out the protocol and domein (and optional port)
```
SonnyBA:
```suggestion
2. In the **Camunda root** field, fill out the protocol and domain (and optional port)
```

docs/configuration/dmn/camunda.rst
```diff
@@ -1,11 +1,65 @@
...
+   of the API root.
+3. In the **REST api path** field, fill out the path of the REST API root, relative to
+   the Camunda root URL.
+4. Under **Auth**, fill out the username and password. Note that the password is never
+   displayed, even if filled out correctly.
+
+The configuration page tests the connection, if any mistakes are made, you will receive
+validation errors.
```
SonnyBA:
```suggestion
The configuration page tests the connection, if any mistakes are made, validation errors will be shown.
```

src/openforms/js/compiled-lang/en.json
SonnyBA, line 1: I think we should only update these files (the dutch compiled version below too) whenever a new release is being prepared (to prevent excessive diffs in between PRs)?

SonnyBA (commented):
This might not have been merged yet whenever the changes for this PR were made but I saw [one of the tests ](https://github.com/open-formulieren/open-forms/blob/700b12d7e09bbf7ad10e09523e01cd30d3dcd1a1/src/openforms/forms/tests/v3/test_endpoints.py#L517)from the `v3` form endpoint is still using the Camunda registration backend.

src/openforms/js/compiled-lang/en.json
annashamray: Unfortunately, if they are not updated the CI is failing 
[Here](https://github.com/open-formulieren/open-forms/actions/runs/24461526770/job/71476803742) is the CI result in this case

SonnyBA: It does make sense to update them in that case :sweat_smile:. I might have been confused with one of the other repositories that does not update them per PR.

annashamray: No, you are right, we have two checks for translations:
1. check message extraction to see if frontend translatable messages are being extracted without checking if they are translated, and this check is always used
2. check missing translation, to see if the extracted messages are translated, this happens only for release branches and tags

annashamray:
@SonnyBA Thanks for catching this test, I've changed it

src/openforms/js/compiled-lang/en.json
SonnyBA: That clears things up, so because the removed code contained translations before, the check (which you mentioned in the first point) failed (when no changes were made to the translations) as it generated different output.

SonnyBA (approved):

docs/configuration/dmn/camunda.rst
```diff
@@ -1,11 +1,64 @@
 .. _configuration_dmn_camunda:
 
-
+=========
 Camunda 7
----------
+=========
+
+Camunda_ is a process engine using BPM models. Open Forms supports Camunda 7.x
```
sergei-maertens, line 7:
```suggestion
Camunda_ is a process orchestrator with a DMN engine. Open Forms supports Camunda 7.x
```

src/openforms/js/compiled-lang/en.json
sergei-maertens: Compiling frontend translations is indeed necessary. For the backend translations, we don't do it because it creates too many merge conflicts - the frontend translations don't suffer that problem.

sergei-maertens (approved):

docs/configuration/dmn/camunda.rst
annashamray: Oops I merged before I saw your remark, I'll make a separate PR for it

Issue #6155, opened 2026-04-02
Profile component has no custom data serializer for validation


viktorvanwijk:
Specifically: `CustomerProfile.build_serializer_field`. It currently relies (or after #6154, "relied") on `BasePlugin.build_serializer_field` which has been deprecated for a while. That serializer only validates whether the data is valid JSON - there is no validation on the structure of profile component data.

PR #6193, opened 2026-04-16
Fix fieldset in editgrid breaking json schema generation

### LINKED ISSUE #6148: Fieldset in repeating group breaks JSON Schema generation

3.5.0 (but probably also 3.4)

### Customer reference

_No response_

Having a fieldset in a repeating group crashes the schema generation. 
KeyError
Level: Error
'veldengroep'

sentry: 467273 

_No response_

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

viktorvanwijk:
Closes #6148

[skip: e2e]

**Changes**

* Ensured children of editgrids are processed properly (NOTE: only need to backport this!)
* Moved all JSON schema generation to the component registry (cleanup)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/components/custom.py
+++ b/src/openforms/formio/components/custom.py
@@ -1086,12 +1086,24 @@ class CustomerProfile(BasePlugin):
         result["data"] = data
         return result
 
+    @staticmethod
+    def as_json_schema(component: CustomerProfileComponent) -> None:
+        raise NotImplementedError()
+
 
 @register("softRequiredErrors")
 class SoftRequiredErrors(BasePlugin):
     holds_submission_data = False
 
+    @staticmethod
+    def as_json_schema(component: Component) -> None:
+        return None
+
 
 @register("coSign")
 class CoSign(BasePlugin):
     holds_submission_data = False
+
+    @staticmethod
+    def as_json_schema(component: Component) -> None:
+        return None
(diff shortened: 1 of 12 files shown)


src/openforms/formio/components/custom.py
```diff
@@ -1086,12 +1086,24 @@ def pre_registration_hook(
         result["data"] = data
         return result
 
+    @staticmethod
+    def as_json_schema(component: CustomerProfileComponent) -> None:
+        raise NotImplementedError()
```
vaszig, line 1091: A bit confused here, shouldn't we have  a `return None` here instead of an exception?

viktorvanwijk: No, because I think it should exist :). See #6155. `None` is meant for components for which it does not make sense to generate a JSON schema

vaszig: Yeah, didn't explain it correctly. I guess it will be properly defined in the mentioned issue right?Because the return type is misleading.

viktorvanwijk: > I guess it will be properly defined in the mentioned issue right?

It should! Will likely be `JSONObject` I think

vaszig (approved):
I didn't find something weird apart from my question.

sergei-maertens (approved):
I trust in the regression tests

viktorvanwijk:
Backports: 
- 3.4.x: de3d2290
- 3.5.x: 4241a7de

PR #6201, opened 2026-04-21
[#6128] hide empty label and value pairs for radio fields

### LINKED ISSUE #6128: Never display items from a list (dynamic in radio/choices/checkboxes) where "label" and "value" are empty.
### Thema / Theme

Form builder

If a logic rule, or service fetch call adds a empty item `""` to a list
```
["foo", "bar", ""] 
```
The radio/choices/checkboxes component should not include this in the displayed options.

This should prevent logic/services that do not always provide "correct" items to not mess up the answers.  

_No response_

1. Create a form with two input fields, with the second being optional.
2. Set up a logic rule that assigns the value of a user defined variable `radioOptions` to `{"merge": ["field1", "field2"]}`. The JSON logic should cast those fields to arrays to produce a single array. This is expected to result in `radioOptions = ["foo", ""]`. The logic trigger can be `true` (use advanced logic).
3. Add a radio field (on the second step) that sets the options from a variable, using the `radioOptions` user defined variable.
4. Save the form and start a submission.
5. Enter a value for the first field, leave the second field blank.
6. Observe that:
    * on 3.3.x only one radio option is rendered
    * on main the radio option + a blank option is rendered (this is not behaving correctly)

The form has:
* new logic evaluation disabled
* new renderer disabled

90% sure this is to do with some of the logic rework. Ask @viktorvanwijk if you get stuck in the code.

SonnyBA:
Closes #6128 

**Changes**

Removes radio selection options where both the label and value are an empty string.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

Tested locally using the form from this export:
[forms-export_0b55bcd5-a299-42ab-8282-cd8ad85a5627.zip](https://github.com/user-attachments/files/26928757/forms-export_0b55bcd5-a299-42ab-8282-cd8ad85a5627.zip)

### DIFF
--- a/src/openforms/formio/dynamic_config/dynamic_options.py
+++ b/src/openforms/formio/dynamic_config/dynamic_options.py
@@ -127,6 +127,7 @@ def add_options_to_config(
         [
             {"label": escaped_label, "value": escaped_key}
             for escaped_key, escaped_label in items_array
+            if escaped_label
         ],
         missing=dict,
     )
--- a/src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
+++ b/src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
@@ -764,6 +764,70 @@ class TestDynamicConfigAddingOptions(TestCase):
             ],
         )
 
+    def test_empty_label_value_pairs(self):
+        submission = SubmissionFactory.create()
+
+        with self.subTest("With empty label & value"):
+            configuration = {
+                "display": "form",
+                "components": [
+                    {
+                        "key": "radio",
+                        "type": "radio",
+                        "values": [{"label": "", "value": ""}],
+                        "openForms": {
+                            "dataSrc": "variable",
+                            "translations": {},
+                            "itemsExpression": {"var": "radioOptions"},
+                        },
+                    }
+                ],
+            }
+
+            rewrite_formio_components(
+                FormioConfigurationWrapper(configuration),
+                submission,
+                FormioData({"radioOptions": [["bar", "Foo"], ["", ""]]}),
+            )
+
+            self.assertEqual(
+                configuration["components"][0]["values"],
+                [
+                    {"label": "Foo", "value": "bar"},
+                ],
+            )
+
+        with self.subTest("With non-empty label & empty value"):
+            configuration = {
+                "display": "form",
+                "components": [
+                    {
+                        "key": "radio",
+                        "type": "radio",
+                        "values": [{"label": "", "value": ""}],
+                        "openForms": {
+                            "dataSrc": "variable",
+                            "translations": {},
+                            "itemsExpression": {"var": "radioOptions"},
+                        },
+                    }
+                ],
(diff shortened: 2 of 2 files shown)


src/openforms/formio/dynamic_config/dynamic_options.py
```diff
@@ -127,6 +127,7 @@ def add_options_to_config(
         [
             {"label": escaped_label, "value": escaped_key}
             for escaped_key, escaped_label in items_array
+            if escaped_key and escaped_label
```
sergei-maertens: Can you point me to the commit that created this regression?

I don't think this fix is correct because an option with an empty value and not-empty label is valid, as remarked in one of the tickets.

I have a feeling this patch is treating symptoms rather than the root cause

SonnyBA: Discussed this with @viktorvanwijk, and correct me if I'm wrong @viktorvanwijk, but the options we have for this at the moment are:

- Revert 45b4692ee28e88aefce9a5c33902978453b86cb9 for the `3.4.x` and `3.5.x` branches, this commit [introduced the behavior that assigns empty values for fields that have no value assigned to them](https://github.com/open-formulieren/open-forms/blob/45b4692ee28e88aefce9a5c33902978453b86cb9/src/openforms/submissions/models/submission_value_variable.py#L722). This does reintroduce the infinite loops that the commit fixed for forms that use the new renderer. Because `4.0.x` allows us to introduce breaking changes we could keep the commit in for that version.
- Apply the changes above, but in the way it was intended; that empty label **and** empty value pairs are filtered out for `radio` components. This could be seen as a band-aid fix and could then also only be applied for the `3.4.x` and `3.5.x` branches.
- Investigate other possible solutions for this problem, this probably would take more time that there is for this ticket

viktorvanwijk: A minor addition to the first point: we will be reworking clear-on-hide behavior anyway in 4.0, so I'm actually expecting to revert the commit in that version as well.

sergei-maertens: Thanks for the analysis. Understanding the problem is key. Let's go with option 2.

src/openforms/formio/dynamic_config/tests/test_dynamic_options.py
```diff
@@ -764,6 +764,70 @@ def test_different_label_key_options(self):
             ],
         )
 
+    def test_empty_label_value_pairs(self):
+        configuration = {
```
viktorvanwijk: Minor thing, but could you move this configuration to the first subtest?

viktorvanwijk (approved):

sergei-maertens (approved):

SonnyBA:
Backports:

- 3.4.x: 303aef6ec4c50c5f19582af47cd3ed23262be102

PR #6220, opened 2026-04-24
Bandaid fix for odd conditional.eq comparison with file components

### LINKED ISSUE #6181: Use correct comparison value types in `component.conditional` when comparing against file components

3.4.8, most likely current & older too

### Customer reference

Taiga Leiden 51

When conditionally displaying a component only if one or more files are added to a file component, this conditional display doesn't work anymore in the new renderer while it did in the old one.

The condition is defined as:

```json
"conditional": {
  "show": false,
  "when": "bestandsupload",
  "eq": ""
}
```

1. Add a form with two fields:
2. File component
3. Content component & on the "Advanced" tab configure a simple conditional:
  * should display: False
  * when: file component from 2.
  * has the value: (leave blank)
4. Save the form
5. Start a form submission
6. Upload a file
7. Observe that the content component is not displayed

The content component is displayed as soon as one file is uploaded.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Partly closes #6181

The builder is misconfigured and allows string values to be provided as comparison value, and the proper fix for it takes a lot more work so we apply a workaround to match the legacy renderer behaviour.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/rendering/nodes.py
+++ b/src/openforms/formio/rendering/nodes.py
@@ -11,7 +11,7 @@ from openforms.submissions.models import SubmissionStep
 from openforms.submissions.rendering.base import Node
 from openforms.submissions.rendering.constants import RenderModes
 
-from ..datastructures import FormioData
+from ..datastructures import FormioConfigurationWrapper, FormioData
 from ..service import format_value, holds_submission_data
 from ..typing import Component
 from ..utils import (
@@ -136,10 +136,18 @@ class ComponentNode(Node):
                 current_iteration_data
             )
             if not is_visible_in_frontend(
-                self.component, artificial_repeating_group_data
+                self.component,
+                artificial_repeating_group_data,
+                configuration_wrapper=FormioConfigurationWrapper(
+                    configuration=self.parent_node.component
+                ),
             ):
                 return False
-        elif not is_visible_in_frontend(self.component, self.step_data):
+        elif not is_visible_in_frontend(
+            self.component,
+            self.step_data,
+            configuration_wrapper=self.renderer.submission.total_configuration_wrapper,
+        ):
             return False
 
         render_configuration = RENDER_CONFIGURATION[self.mode]
--- a/src/openforms/formio/utils.py
+++ b/src/openforms/formio/utils.py
@@ -400,7 +400,8 @@ def is_visible_in_frontend(
         return conditional_show
 
     if (
-        isinstance(trigger_component_value, dict)
+        trigger_component["type"] == "selectboxes"
+        and isinstance(trigger_component_value, dict)
         and compare_value in trigger_component_value
     ):
         return (
--- a/src/openforms/formio/utils.py
+++ b/src/openforms/formio/utils.py
@@ -17,7 +17,7 @@ from .constants import COMPONENT_DATA_SUBTYPES, COMPONENT_DATATYPES
 from .typing import Column, ColumnsComponent, Component, FormioConfiguration
 
 if TYPE_CHECKING:
-    from .datastructures import FormioData
+    from .datastructures import FormioConfigurationWrapper, FormioData
 
 logger = structlog.stdlib.get_logger(__name__)
 tracer = trace.get_tracer("openforms.formio.utils")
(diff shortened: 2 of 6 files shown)


src/openforms/formio/utils.py
```diff
@@ -344,7 +344,11 @@ def conform_to_mask(value: str, mask: str) -> str:
     return "".join(result)
 
 
-def is_visible_in_frontend(component: Component, data: FormioData) -> bool:
+def is_visible_in_frontend(
+    component: Component,
+    data: FormioData,
+    configuration_wrapper: FormioConfigurationWrapper,
+) -> bool:
```
sergei-maertens, line 351: I think there's still a case missing here for the membership test in case the component has `multiple: true`, which is handled in formio.js and in our renderer. I find it a bit sus that afaik there haven't been bugs reported about that 🤔 

Decided to leave that out of scope for now, but I think we should have the necessary building blocks now.

src/openforms/formio/rendering/nodes.py
```diff
@@ -136,10 +136,18 @@ def is_visible(self) -> bool:
                 current_iteration_data
             )
             if not is_visible_in_frontend(
-                self.component, artificial_repeating_group_data
+                self.component,
+                artificial_repeating_group_data,
+                configuration_wrapper=FormioConfigurationWrapper(
+                    configuration=self.parent_node.component
```
viktorvanwijk, line 142: Could you write a test for this?

I think it should be
```suggestion
                    configuration={"components": [self.parent_node.component]}
```
to get the magic editgrid behavior, because `FormioConfigurationWrapper.component_map` uses `iter_components` on the configuration. With the configuration being the editgrid component itself, it will just loop over the children

viktorvanwijk (changes requested):
Probably also need to add this check to the file component in the registry with a `test_conditional` override, to avoid weird things happening during logic evaluation

src/openforms/formio/rendering/nodes.py
sergei-maertens: yeah it was my intention to loop over the children indeed. I'll add a test!

sergei-maertens: There is an existing test that passes: `openforms.submissions.tests.renderer.test_submission_summary.SubmissionSummaryRendererTests.test_namespaced_editgrid_summary_with_conditions_in_iteration`

sergei-maertens: I've amended it with a more complex case by pointing to a selectboxes outside of the editgrid itself and it passes 🤷

sergei-maertens:
Backports:

* stable/3.5.x: 112a3339315f474692bd8f61c863a5fd8e2c3ea3
* stable/3.4.x: 28d2584d9a4e2c695d9d08babf1e09302e236a58

src/openforms/formio/rendering/nodes.py
viktorvanwijk: That test passes because there is a fallback in `is_visible_in_frontend` to the unknown component type, and a radio component doesn't have a special override, but the actual component is never found. Without that fallback it would just crash.

But you make a good point about the root selectboxes component. That actually means we should pass the complete configuration, not just the editgrid-specific one. Note that the selectboxes component you added is actually a radio component, so the test still passes.

Why merge without approval :(

sergei-maertens: > Why merge without approval :(

I dislike it, but certain clients are breathing down my neck for the patch release and stressing me out ☹️

Thanks for the additional information, I'll tinker with it a bit more over the weekend.

sergei-maertens: I tried out both cases with the outside side selectboxes component that does have special cased treatment, and it did remove the second conditional text fields, so something is still making that behave as it did before without the full configuration. I'll get to the bottom of it.

sergei-maertens: oh ffs, I read `radio` as `selectboxes` :facepalm: which gave me false confidence that this was working correctly

PR #6224, opened 2026-04-27
Clean up the (S)CSS for the public UI & admin

### LINKED ISSUE #2705: Untangling CSS in the backend
### Thema / Theme

Frontend

We need to reorganise the CSS that is used in the frontend of the backend, because it can interfere with the CSS of the SDK.

This was noticed with issue https://github.com/open-formulieren/open-forms/issues/2686, where the dropdown part of the select component was partly overlapping with the component under it. This was not happening in the SDK on its own, but only when a form was embedded. 

This is because this style in the backend was overwriting the style of the select component: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/scss/components/builder/_builder.scss#L78

We need to have a better structure so that we prevent more conflicting CSS in the future.

sergei-maertens:
Closes #2705

**Changes**

* Replaced a number of custom CSS usages with NL DS components
* Removed dead code (hopefully I correctly identified it as dead)
* Moved public UI CSS code to more logical location in project structure
* De-cluttered the build/webpack config
* Reduce our reliance/dependance on microscope-sass to the bare minimum
* Deleted dead code
* Converted `@import` rules to `@use` rules where possible

This puts us in a great position to almost rip out microscope-sass in the future.

I tried adding stories for those components, but the styles are completely messed up due to the global styles of the admin being loaded :(

[skip: e2e]

### Before & after screenshots

The before screenshots are taken from the `latest` build based on `main`, on our test environment. After is taken on my local environment with browser cache disabled.

**Homepage / form list page**

Notes: I don't have a logo configured locally.

Before

image

After

image

**Cookie notice**

Before

image

After

image

**Back to top & footer **

Before

image

After

image

**403 error**

Before

image

After

image

**Registrator info page**

Before

image

After

image

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


docs/installation/upgrade-400.rst
```diff
@@ -110,3 +110,73 @@ with the ``open-forms-sdk-wrapper.mjs`` example code:
...
+* ``--utrecht-paragraph-line-height: 1.5;`` for the spacing between text lines
+* ``--utrecht-button-group-inline-gap: 20px;``
+
+We do not provide defaults/fallbacks for these design tokens, as they are considered
+too invasive.
+
+Cookie notice
+-------------
+
+.. todo:: Check if we can provide a shim for this.
```
sergei-maertens: TODO

docs/installation/upgrade-400.rst
```diff
@@ -110,3 +110,73 @@ with the ``open-forms-sdk-wrapper.mjs`` example code:
...
+and we don't have any guarantees that the relevant design tokens are set.
+
+Below you find a summary of components that were moved from custom CSS to existing
+NL DS components that may require visual inspection/additional definitions in your
+custom theme stylesheet(s).
+
+Cookie group
+------------
+
+.. todo:: Check if we can provide a shim for this.
```
sergei-maertens: TODO

src/openforms/scss/public-ui/openforms-theme/_page.scss
```diff
@@ -1,14 +1,17 @@
...
+ *
+ * @TODO Move these to a separate theme-specific style sheet.
  */
 .openforms-theme {
   .utrecht-page {
     @include desktop {
       --utrecht-page-max-inline-size: 1200px;
     }
 
+    // TODO: deprecate and explicitly define design token?
```
sergei-maertens, line 9: TODO: define own design token

src/openforms/scss/public-ui/openforms-theme/_page.scss
```diff
@@ -1,14 +1,12 @@
...
-.openforms-theme {
+@mixin theme-styles {
   .utrecht-page {
-    @include desktop {
+    @include responsive.desktop {
       --utrecht-page-max-inline-size: 1200px;
     }
 
+    // TODO: deprecate and explicitly define design token?
     --utrecht-page-background-color: var(--of-view-background, #e6e6e6);
```
sergei-maertens, line 10: TODO

src/openforms/scss/public-ui/openforms-theme/_page.scss
sergei-maertens: Leaving this as is right now - it looks like the `utrecht-page` component is not documented yet/anymore, and utrecht-page-body exists now, but that migration is too big to pull into scope.

src/openforms/scss/pdfs/_typography.scss
```diff
@@ -1,9 +1,63 @@
...
+  margin: 0;
+
+  &:not(:last-child) {
+    margin-block-end: 0.5rem;
+  }
+}
+
+@font-face {
+  font-family: 'Fira Sans';
+  src: url('../fonts/fira-sans-v11-latin-regular.eot');
```
viktorvanwijk, line 31: These should be updated no?
```suggestion
  src: url('../../static/fonts/fira-sans-v11-latin-regular.eot');
```

viktorvanwijk (commented):
Nice refactor! Really crazy how much the file sizes have decreased.

I went through the commits one-by-one instead of looking at the complete list of files changed on Github, and that really helped :). Only managed to find one thing

In `ui-components.rst` there is a section **SCSS helpers**. I think that list could be updated with the backend now, right?

src/openforms/scss/pdfs/_typography.scss
sergei-maertens: no, they're correct I believe - the font asset loading is relative to browser paths from the CSS file that references it.

So, the PDF CSS file is served at `/static/bundles/pdf-css.css`, meaning that `../` points to `/static/` already, resolving to `/static/fonts/...`.

The file system path is confusing here, but it's not what's used when a browser (or PDF renderer pretending to be a browser) loads the assets.

sergei-maertens:
> In `ui-components.rst` there is a section **SCSS helpers**. I think that list could be updated with the backend now, right?

Yes - I've added an entry!

src/openforms/scss/pdfs/_typography.scss
viktorvanwijk: Ah I see, thanks :)

viktorvanwijk (approved):

PR #6233, opened 2026-04-29
Onegov/Ilionx-specific StUF-ZDS tweaks

### LINKED ISSUE #6230: Expose `completed_on` static variable for StUF-ZDS extraElementen + document OneGov specific workarounds
OneGov uses StUF-ZDS, but have some vendor-specific limitations/workarounds. These can usually be addressed with some guidelines/remarks in the documentation, except for the zaak-startdatum issue.

**Zaak startdatum**

Long story short: on their end, they need a datetime, we send a date, filling out the time portion is not correct because the StUF-ZDS standard/XSD is very clear.

However, there is a workaround: if an extra element is sent with the name `zaak.pv_startdatum` and the value of the element is an ISO-8601 datetime string, then it will override the zaak startdatum element.

We need a static variable that exposes `submission.completed_on` so that it can be mapped in the extra elementen.

sergei-maertens:
Closes #6230

**Changes**

* Added a new registration variable for StUF-ZDS
* Documented Onegov-specific configuration instructions and pitfalls

While this is technically a new feature, we plan to backport it as the changes are simple and will greatly benefit the municipality that's currently blocked by these limitations.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/docs/configuration/registration/stufzds.rst
+++ b/docs/configuration/registration/stufzds.rst
@@ -136,8 +136,11 @@ These SOAP-operations are used by this plugin:
 * ``creeerZaak_Lk01``
 
    * ``heeftAlsInitiator`` (the initiator can be excluded if needed)
-   * ``heeftAlsOverigBetrokkene`` (will only be set if an employee logs in on behalf of a client)
+   * ``heeftAlsOverigBetrokkene`` (set if: an employee logs in on behalf of
+     a client, the submission has been cosigned or there are family member relations)
    * ``heeft`` (the status can be excluded if needed)
+   * ``extraElementen``, all remaining unmapped submission data elements + explicitly
+     mapped (registration) variables through the configuration
 
 * ``updateZaak_Lk01`` (only used when delayed payments are enabled)
 * ``genereerDocumentIdentificatie_Di02``
--- a/docs/configuration/registration/stufzds.rst
+++ b/docs/configuration/registration/stufzds.rst
@@ -15,6 +15,10 @@ register form submissions.
    This service contains sensitive data and requires a connection to an
    external system, offered or maintained by a service provider.
 
+.. contents:: Jump to
+   :depth: 2
+   :local:
+   :backlinks: none
 
 What does the Open Forms administator need?
 ===========================================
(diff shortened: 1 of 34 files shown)


docs/configuration/registration/stufzds.rst
```diff
@@ -136,9 +140,72 @@ These SOAP-operations are used by this plugin:
...
+* ``zaak.pv_orderid``: the registration variable "Payment public order IDs"
+  (``payment_public_order_ids``) exists, however emitting it will result in a ``.0``,
+  ``.1``, ... suffix being added to the name of the extra element, as multiple order
+  IDs may exist. This will effectively produce XML looking like:
+
+  .. code-block:: xml
+
+    <StUF:extraElement naam="zaak.pv_orderid.0">12334</StUF:extraElement>
+
+  .. warning:: It's currently unknown if Onegov can handle this or not.
```
sergei-maertens, line 192: I will share this with Ilionx and check in how to deal with this, to be continued, but no need to block this PR because of it!

vaszig (approved):
I am not familiar with all this specification (outside the StUF-ZDS) but it seems correct to me.

annashamray (changes requested):
I've tested it locally and it doesn't work for me: the variable is None

here is config

Image

And here is a part of request which was sent:
```
<StUF:extraElement naam="zaak.pv_startdatum">None</StUF:extraElement>
```

annashamray:
@vaszig Does it work locally for you? For me some other extra elements are filled, but this one is `None`

vaszig:
> @vaszig Does it work locally for you? For me some other extra elements are filled, but this one is `None`

I trusted the test and the code I saw, for me it's `None` too (did a quick check in the `extra_data` during the registration). I may miss something, will have a proper look tomorrow

sergei-maertens:
Huh, odd, I was testing it through the `register_submission` management command with an existing submission in my database that was completed, and inspected the XML produced in the request.

The `completed_on` model field is supposed to be populated for a submission that enters the registration pipeline :thinking:

vaszig:
> Huh, odd, I was testing it through the `register_submission` management command with an existing submission in my database that was completed, and inspected the XML produced in the request.
> 
> The `completed_on` model field is supposed to be populated for a submission that enters the registration pipeline 🤔

I guess this is the problem, the missing new variable?Or maybe not, looking at this now
https://github.com/open-formulieren/open-forms/blob/main/src/openforms/registrations/contrib/stuf_zds/plugin.py#L154

sergei-maertens:
> > Huh, odd, I was testing it through the `register_submission` management command with an existing submission in my database that was completed, and inspected the XML produced in the request.
> > The `completed_on` model field is supposed to be populated for a submission that enters the registration pipeline 🤔
> 
> I guess this is the problem, the missing new variable?Or maybe not, looking at this now https://github.com/open-formulieren/open-forms/blob/main/src/openforms/registrations/contrib/stuf_zds/plugin.py#L154

Yes, looks like it!

And that explains why things were working, I had an initial test setup with a static variable and only afterwards added the completed_on variable :facepalm: I was mixing up things

annashamray (approved):
Works as expected on my machine:
```
<StUF:extraElement naam="zaak.pv_startdatum">2026-04-30T12:35:23.103310+00:00</StUF:extraElement>
```

sergei-maertens:
Backports:

* stable/3.5.x: 046539d7183be76e4eb29b30e447ff2f2fe5c3b9
* stable/3.4.x: 7c6150df22cbd3f5ca6e6ea872f8119bcbb0b72f

Issue #6164, opened 2026-04-03
:boom: 4.0 - overview of deprecated functionalities removal


sergei-maertens:
Scavenged from grepping on `DeprecationWarning` in the codebases, milestone 4.0 issues and post-its.

TODO: create sub-issues for the tasks listed here

* [x] Remove formio.js from the SDK & support for old renderer / associated feature flag https://github.com/open-formulieren/open-forms-sdk/pull/963
* ~~Remove (through forced new renderer) support for `textfield` address autofill/derivation -> sort out a migration path with address NL~~ (See #6239)
* [ ] Remove legacy logic evaluation & associated feature flag 
    - SDK: https://github.com/open-formulieren/open-forms-sdk/pull/981
    - backend: #6261 
* [x] Remove JCC SOAP appointment plugin (@vaszig)
* [x] Remove Elastic-APM support (#6257)
* [x] Simplify registration workflow (ZGW APIs -> change default to use OF-generated public reference) (https://github.com/open-formulieren/open-forms/issues/6291)
* [ ] Check what we can do with prefill data validation to mark as authentic/prevent tampering without relying on the form field being marked as read-only
* [x] Delete `authentication_success` signal - it's unused afaik (#6238)
* [x] Switch the default/legacy SDK bundles being offered - there are now compatibility URLs
    - SDK part: https://github.com/open-formulieren/open-forms-sdk/pull/975
    - Backend part: #6215
* [x] Remove formio registry `build_serializer_field` fallback (#6154)
* [ ] StUF-ZDS: prefer branch number from auth data rather than form variable (@vaszig)
* [x] `CosignV2Data` -> make `cosign_date` required (#6238)
* [x] Remove formio translations endpoints in the backend (if not used by our own backend JS!) https://github.com/open-formulieren/open-forms/pull/6237
* [x] Remove fallbacks in backtotop-link CSS (done in f2e2121a162234d8453cbfbc4a136cf86aec7f52)

Explicitly *NOT* yet scheduled for removal because it's still used too often:

* cosign v1
* Objects API registration options v1 (template based)

PR #6237, opened 2026-04-30
Remove formio translations endpoints

sergei-maertens:
Related to #6164

**Changes**

* Deleted the endpoints for the formio translations JSON

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/translations/api/views.py
```diff
@@ -96,49 +93,6 @@ def put(self, request: Request) -> Response:
...
-    ],
-    responses={
-        ("200", "application/json"): build_object_type(
-            additionalProperties={"type": "string"}
-        ),
-        "404": ExceptionSerializer,
-        "5XX": ExceptionSerializer,
-    },
-)
-class FormioTranslationsView(APIView):
```
vaszig, line 120: What about https://github.com/open-formulieren/open-forms/blob/main/src/openforms/translations/views.py#L15 ?I do not remember the difference.

sergei-maertens: uuhhh... did I remove the wrong endpoint? :sweat_smile: let me check

sergei-maertens: okay, didn't delete the wrong one and it turns out that the one you link could've been removed ages ago! thanks for catching that

vaszig (approved):

PR #6251, opened 2026-05-04
Minor corrections in translation

magdalena-source:
**Changes**

Corrected some minor issues in the translation doc

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/conf/locale/nl/LC_MESSAGES/django.po
+++ b/src/openforms/conf/locale/nl/LC_MESSAGES/django.po
@@ -3549,7 +3549,7 @@ msgstr "Geeft aan hoe sessies worden opgeschoond na de bewaartermijn."
 #: openforms/config/models/config.py:522 openforms/forms/models/form.py:363
 #: openforms/forms/models/form.py:373
 msgid "errored submission removal limit"
-msgstr "bewaartermijn voor niet voltooide inzendingen inzendingen"
+msgstr "bewaartermijn voor niet voltooide inzendingen"
 
 #: openforms/config/models/config.py:526
 msgid "Amount of days errored submissions will remain before being removed"


src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -3514,42 +3514,42 @@ msgstr ""
 "informatie heb verzwegen."
 
 #: openforms/config/models/config.py:492 openforms/forms/models/form.py:323
-msgid "successful submission removal limit"
-msgstr "bewaartermijn voor voltooide inzendingen."
+msgid "Successful submission removal limit"
```
sergei-maertens: You can't update the source strings, these are defined in the code and extracted from the code, which makes them end up in .po files.

Additionally, the labels of model fields are lower cased on purpose, as these are also used in the middle of sentences like validation errors.

sergei-maertens (changes requested):

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
magdalena-source: Thanks for explaining - I suspected that they're lower case on purpose, but noticed some inconsistency and thought it's a mistake. Could you also clarify what source strings are - msgid, msgstr or both? And lf I want to change something in the source strings, should I look for it in the code?

sergei-maertens: `msgid` is always extracted from the code :)

`msgstr` holds the translation - and yes, the comments above point to the location in the code where they're defined

magdalena-source: Crystal clear! TY!

sergei-maertens (approved):

PR #6261, opened 2026-05-07
Ensure new logic evaluation is the only logic evaluation

### LINKED ISSUE #6164: :boom: 4.0 - overview of deprecated functionalities removal

viktorvanwijk:
Partly closes #6164

**Changes**

* Removed usages of `new_logic_evaluation_enabled` feature flag
* Removed usages of the `trigger_from_step` setting for logic rules 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


docs/installation/upgrade-400.rst
```diff
@@ -205,3 +205,31 @@ code, and if you haven't done so yet, recommend you to set up the necessary
...
+    Before upgrading, all existing forms should be converted to the new logic evaluation. This can be done on a
+    per-form basis, or in bulk using the following management command. Any forms which contain cycles in their
+    logic rules need to be resolved manually. The output of the management command will include relevant form
+    details if this is the case.
+
+    .. code-block:: bash
+
+        python /app/src/manage.py enable_new_logic_evaluation_for_all_forms
+
+Clearing of values
```
vaszig, line 262: Is this the new epic that you are working on (6121)?

src/openforms/forms/migrations/0131_remove_formlogic_trigger_from_step.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 5.2.13 on 2026-05-01 14:25
```
vaszig: I guess you weren't able to remove both fields and have one migration file?

src/openforms/js/components/admin/form_design/FormLogic.stories.js
```diff
@@ -223,16 +223,6 @@ export const FullFunctionality = {
 
     availableFormVariables: AVAILABLE_FORM_VARIABLES,
     availableFormSteps: AVAILABLE_FORM_STEPS,
-    newLogicEvaluationEnabled: false,
-  },
-
-  play: async ({canvasElement}) => {
```
vaszig, line 229: Do we want a play function here or doesn't make any sense any more?

bin/check_disable_next_logic_action.py
```diff
@@ -1,240 +0,0 @@
-#!/usr/bin/env python
```
vaszig, line 1: So, this is removed because the new logic evaluation handles this properly, correct?

vaszig (commented):
Basically removing stuff related to the flag so not to much to say. Only a couple of questions (as I haven't dived into the new logic evaluation).

- Also saw some data migrations tests related to `trigger_from_step`, I believe they can be removed in the next version
- Regarding the UI tests I see  some differences in the languages, can you take a look?

docs/installation/upgrade-400.rst
viktorvanwijk: Yes!

src/openforms/forms/migrations/0131_remove_formlogic_trigger_from_step.py
viktorvanwijk: It should be possible I think, just wanted to do it step-by-step to make it easier for myself :)

Now that everything is converted, I can extract them from the commits and combine into one, though

src/openforms/js/components/admin/form_design/FormLogic.stories.js
viktorvanwijk: I added this play function before to make sure legacy logic evaluation still worked properly, so thought it was OK to remove it again. Do you have suggestions on what to add instead?

bin/check_disable_next_logic_action.py
viktorvanwijk: This is a task form designers have to do after upgrading to 3.5, since the disable-next action has changed there (both for new and legacy logic evaluation). I figured that 3.5 will be prerequisite for 4.0, so I have removed it

viktorvanwijk:
>     Also saw some data migrations tests related to `trigger_from_step`, I believe they can be removed in the next version

You mean the `ChangeDisableLogicActionMigrationTests`? I thought as long as the migration isn't squashed, it's OK to leave it

>     * Regarding the UI tests I see  some differences in the languages, can you take a look?

Will do!

src/openforms/js/components/admin/form_design/FormLogic.stories.js
vaszig: Not particularly, just asking if you indeed wanted to show something different instead of the specific configuration.

vaszig:
> > ```
> > Also saw some data migrations tests related to `trigger_from_step`, I believe they can be removed in the next version
> > ```
> 
> You mean the `ChangeDisableLogicActionMigrationTests`? I thought as long as the migration isn't squashed, it's OK to leave it

Yes, I agree!

src/openforms/forms/migrations/0131_remove_formlogic_trigger_from_step.py
viktorvanwijk: ~Added a new commit where they are combined~

EDIT: we want `report_invalid_logic_rules.py` to be available in both 3.5 and 4.0, which means we cannot remove the `trigger_from_step` and `new_logic_evaluation_enabled` model fields yet

viktorvanwijk:
> Regarding the UI tests I see some differences in the languages, can you take a look?

Discussed on Slack: seems like it compares to an outdated chromatic snapshot? I have triggered a rebuild, but that didn't solve it unfortunately. The messages as they are in the new snapshots are the correct ones, see https://github.com/open-formulieren/open-forms/blob/main/src/openforms/js/lang/nl.json#L3493 (they are not translated yet)

src/openforms/conf/base.py
```diff
@@ -1320,6 +1321,9 @@
     "3.5.0": UpgradeCheck(
         VersionRange(minimum="3.3.1"),
     ),
+    "4.0.0": UpgradeCheck(
+        VersionRange(minimum="3.5.1"), (BinScriptCheck("report_invalid_form_logic.py"),)
```
sergei-maertens:
```suggestion
        VersionRange(minimum="3.5.1"), (BinScriptCheck("report_invalid_form_logic"),)
```
Probably the `ci.yml` workflow needs an update with the upgrade_simulation upgrade matrix :)

Dockerfile
```diff
@@ -115,9 +115,7 @@ COPY \
     ./bin/report_component_problems.py \
     ./bin/fix_submission_value_variable_missing_fields.py \
     ./bin/fix_payment_status.py \
-    ./bin/check_disable_next_logic_action.py \
     ./bin/report_completed_submissions_access.py \
-    ./bin/report_invalid_form_logic.py \
```
sergei-maertens, line 120: this will need to be added back

viktorvanwijk: I think we need a new checkbox in the PR checklist:

- [x] Updated the Dockerfile with the scripts that were re-added after a PR review

src/openforms/conf/base.py
viktorvanwijk: Will it already execute though? 4.0.0 doesn't exist yet, so there is nothing to upgrade to, right?

I have also added additional tests to `test_upgrade_paths`, and noticed an error on `transaction.set_autocommit(False)` in the report script, because it is executed in an 'atomic' block already. I'm assuming the upgrade checks are rolled back automatically after executing?

sergei-maertens: if you edit the CI YAML:

```yaml
        env:
          RELEASE: '3.5.0'
          # ensure local image gets used
          TAG: ${{ needs.setup.outputs.version }}
```
and change the 3.5.0 into 4.0, you're "tricking" Open Forms in thinking it's already a released 4.0 version :wink:

sergei-maertens: > I'm assuming the upgrade checks are rolled back automatically after executing?

Upgrade *checks* shouldn't perform any writes in the first place! So there should be nothing to roll back.

viktorvanwijk: > and change the 3.5.0 into 4.0, you're "tricking" Open Forms in thinking it's already a released 4.0 version 😉

:grimacing: 

> Upgrade checks shouldn't perform any writes in the first place! So there should be nothing to roll back.

Fair enough :) Will remove cycle detection completely actually, since all those forms will be caught by `if forms_to_check.exists()` anyway

sergei-maertens (approved):

PR #6268, opened 2026-05-08
Zaaktype/Documenttype urls migration tool

### LINKED ISSUE #6262: Implement file component migration and problem detection
- [x] The file component migration script must resolve the specified document type URL and fill out the catalogue + omschrijving details.
    - [x] `file` components
    - [x] `openforms.contrib.objects_api.models.ObjectsAPIGroupConfig` (submission report, submission cvs, attachment)
    - [x] `openforms.registrations.contrib.objects_api.config.ObjectsAPIOptionsSerializer` (submission report, submission cvs, attachment)
    - [x] `openforms.registrations.contrib.zgw_apis.options.ZaakOptionsSerializer` (zaaktype, informatieobjecttype)
- [x] The migration script must have a dry-run mode that reports problems:
    - Report mismatch in catalogues
    - Report URL resolution problems
    - Report other inconsistencies
- [x] Also check against the selected zaaktype for ZGW APIs
- ~~Add warnings in UI~~ too much work, we can run the migration tool and report back the output

A dry-run/detection only mode for the migration script is appropriate.

sergei-maertens:
Closes #6262

**Changes**

* Added management command to report/migrate forms to migrate
* Handles migration of file components, zgw apis and objects api registration backends

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/tests/test_zgw_urls_migrator.py
+++ b/src/openforms/tests/test_zgw_urls_migrator.py
@@ -1,0 +1,1552 @@
+from io import StringIO
+from uuid import UUID, uuid4
+
+from django.core.management import CommandError, call_command
+from django.test import TestCase
+
+from openforms.contrib.objects_api.tests.factories import ObjectsAPIGroupConfigFactory
+from openforms.forms.tests.factories import (
+    FormDefinitionFactory,
+    FormFactory,
+    FormRegistrationBackendFactory,
+)
+from openforms.registrations.contrib.objects_api.config import (
+    ObjectsAPIOptionsSerializer,
+)
+from openforms.registrations.contrib.objects_api.constants import (
+    PLUGIN_IDENTIFIER as OBJECTS_PLUGIN_IDENTIFIER,
+)
+from openforms.registrations.contrib.objects_api.typing import (
+    RegistrationOptionsV2 as ObjectsRegistrationOptionsV2,
+)
+from openforms.registrations.contrib.zgw_apis.options import ZaakOptionsSerializer
+from openforms.registrations.contrib.zgw_apis.tests.factories import (
+    ZGWApiGroupConfigFactory,
+)
+from openforms.registrations.contrib.zgw_apis.typing import (
+    RegistrationOptions as ZGWRegistrationOptions,
+)
+from openforms.utils.tests.vcr import OFVCRMixin
+
+from ..zgw_urls_migrator import (
+    MigrationProblem,
+    Migrator,
+    look_up_case_type,
+    look_up_document_type,
+)
+
+
+class MigratorTests(OFVCRMixin, TestCase):
+    """
+    Requires the VCR services, see ``docker/start_vcr_services.sh``.
+    """
+
+    maxDiff = None
+
+    def setUp(self):
+        super().setUp()
+
+        def clear_cache():
+            look_up_case_type.cache_clear()
+            look_up_document_type.cache_clear()
+
+        self.addCleanup(clear_cache)
+
+    def test_doesnot_crash_without_any_data(self):
+        outfile = StringIO()
+        migrator = Migrator(outfile=outfile)
+
+        result = migrator.migrate()
+
+        self.assertIsNone(result)
+
+    def test_ok_for_forms_with_nothing_to_do(self):
+        FormFactory.create_batch(
+            3,
+            generate_minimal_setup=True,
+            formstep__form_definition__configuration={
+                "components": [
+                    {
+                        "type": "textfield",
+                        "key": "ignoreMe",
+                        "label": "Ignore me",
+                    },
+                ]
+            },
+        )
+        FormFactory.create(is_appointment_form=True)
+        outfile = StringIO()
+        migrator = Migrator(outfile=outfile)
+
+        result = migrator.migrate()
+
+        self.assertIsNone(result)
+
+    def test_happy_flow_migrate_objects_api_groups(self):
+        objects_api_group_with_modern_config = ObjectsAPIGroupConfigFactory.create(
+            for_test_docker_compose=True,
+            catalogue_domain="DNT",
+            catalogue_rsin="DNTDNTDNT",
+            iot_submission_report="do not touch",
+            iot_submission_csv="do not touch",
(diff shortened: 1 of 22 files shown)


sergei-maertens:
> Checked import/export of a form

I don't think there's anything to do here - it's a lot of work to support this and there are no real benefits:

* Samen Delen forms between different organisations -> probably a bad idea anyway to keep registration backend configuration in there, it will already have broken references to the API groups anyway
* Within the same organization -> we'll remove the legacy functionality in 4.0, so probably best to just create a new export on a migrated environment

src/openforms/zgw_urls_migrator.py
```diff
@@ -0,0 +1,669 @@
...
+        database.
+
+        :return: Component keys that were updated.
+        """
+        # key: component key, value: detailed reporting information for the component
+        component_details: dict[str, str] = {}
+        config_wrapper = form_definition.configuration_wrapper
+        # loop over all components in the form definition and process the file
+        # components
+        needs_save = False
```
annashamray, line 202:
```suggestion
        needs_save = False
        changed = False
```

otherwise it can be referenced without assignment on line 217

src/openforms/zgw_urls_migrator.py
```diff
@@ -0,0 +1,669 @@
...
+        for component in config_wrapper:
+            match component["type"]:
+                case "file":
+                    try:
+                        changed = self.process_file_component(component)  # pyright: ignore[reportArgumentType]
```
annashamray, line 207:
```suggestion
                        changed = self.process_file_component(component) 
```

I thought the whole file is ignored :)

src/openforms/zgw_urls_migrator.py
```diff
@@ -0,0 +1,669 @@
...
+            if changed:
+                needs_save = True
+                component_details[component["key"]] = "Configuration update ok."
+
+        if needs_save:
+            form_definition.save()
+
+        return component_details
+
+    def process_file_component(self, component: FileComponent) -> bool:
```
annashamray: it can be static or moved to a separate function

src/openforms/zgw_urls_migrator.py
```diff
@@ -0,0 +1,669 @@
...
+        if any(
+            migrator.has_errors
+            for migrator in (formio_config_migrator, registration_backend_migrator)
+        ):
+            raise MigrationProblem(
+                "There are automatic migration problems, please analyze the output."
+            )
+
+
+def look_up_document_type(url: str) -> tuple[tuple[str, str], str]:
```
annashamray, line 72:
```suggestion
@lru_cache
def look_up_document_type(url: str) -> tuple[tuple[str, str], str]:
```

If we expect a lot of file components with legacy documenttype url, they probably have the limited number of document types, so maybe we can cache them?

src/openforms/tests/test_zgw_urls_migrator.py
```diff
@@ -0,0 +1,1538 @@
...
+            migrator.migrate()
+
+        output = outfile.getvalue()
+
+        self.assertIn(
+            "Resolving the case/document types didn't converge to a single catalogue.",
+            output,
+        )
+
+    def test_error_flow_zgw_document_type_does_not_belong_to_case_type(self):
```
annashamray, line 1390: I don't think you can add such documenttype in the admin, but better safe than sorry!

annashamray (approved):
I've tested locally such cases:
* file component:
  * correct legacy url
  * incorrect legacy url
* registration backend
  * zgw - correct legacy  url

All work as expected

I've reviewed code and looked at test cases and just have a couple minor remarks, which are nice to have

src/openforms/zgw_urls_migrator.py
sergei-maertens: hrm that's odd, I'd expect my type checker to complain about that. But I also don't see the problem when looking at the code:

`changed` is in scope for each `component` being iterated over from line 200. The match statement assigns the `changed` variable if it's a file component, and in all other cases the loop iteration is skipped over (the `continue` statements).

the `changed` variable is being read in the scope of the loop iteration, so it's guaranteed to be defined, no?

src/openforms/zgw_urls_migrator.py
sergei-maertens: in CI yes, in my editor locally no :-) I use pyright via LSP-pyright in my editor and don't like the red squiggly lines when it's a false positive

src/openforms/zgw_urls_migrator.py
sergei-maertens: I'll make it a static method

src/openforms/tests/test_zgw_urls_migrator.py
sergei-maertens: I don't think so either, but given it's a JSONField under the hood weird shit can happen for people messing with the JSON directly :grimacing:

src/openforms/zgw_urls_migrator.py
sergei-maertens: yeah, if I remove the `case _:`, then my type checker complains about possibly unbound variable :)

src/openforms/zgw_urls_migrator.py
sergei-maertens: > I thought the whole file is ignored :)

The opposite actually after double checking, it's explicitly added to be type checked in CI!

src/openforms/zgw_urls_migrator.py
annashamray: right, I didn't take into account that `continue` for all other component types, so you are right! 
My typechecker also didn't complain, but I didn't trust him here -_-

src/openforms/tests/test_zgw_urls_migrator.py
annashamray: True! Never trust the user!

sergei-maertens:
Backport commit 0f6af59c8725a97498a57116f17a3f4acc2f8dd8 to stable/3.5.x in PR #6266

PR #6270, opened 2026-05-08
Add management command for bulk converting forms to new logic evaluation

### LINKED ISSUE #6164: :boom: 4.0 - overview of deprecated functionalities removal

viktorvanwijk:
Partly closes #6164

Note that this merges into `stable/3.5.x`

[skip: e2e]

**Changes**

* Added management command
* Added placeholder in the changelog 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,31 @@ Changelog
         `latest <https://open-forms.readthedocs.io/en/latest/changelog.html>`_ docs
         version.
 
+3.5.2 (2026-05-xx)
+==================
+
+Regular bugfix release.
+
+.. note::
+
+    Because the legacy logic evaluation will be removed in Open Forms 4.0, we have added some
+    additional tooling to convert all forms to the new logic evaluation.
+
+    For each form, the logic rules will be reordered and any 'trigger from step' setting will be
+    removed. Note that this action cannot be easily reverted, so for large forms it is recommended to
+    create a copy of the form first, and enable the new logic evaluation feature flag manually
+    before running this command.
+
+    Any forms which contain rules with cycles will not be converted, and need to be resolved manually.
+    If such forms exist, the output of this command will contain a list of relevant form details.
+
+    In an app container, execute:
+
+    .. code-block:: bash
+
+        python /app/src/manage.py enable_new_logic_evaluation_for_all_forms
+
+
 3.5.1 (2026-05-06)
 ==================
 
@@ -130,7 +155,7 @@ removing the old renderer and Textfield address autofill support.
 
 **Legacy logic evaluation**
 
-The old logic evaluation will be removed in Open Forms 4.0. We will prioritize bugfixes in the new 
+The old logic evaluation will be removed in Open Forms 4.0. We will prioritize bugfixes in the new
 logic evaluation to make sure this migration will be as smooth as possible.
 
 **File Registration tab**
@@ -151,7 +176,7 @@ We plan to change the default for "Use generated zaaknummer" to unchecked in an 
 
 **StUF-ZDS registration**
 
-The usage of the branch number retrieved from the form's submission data will be replaced in 
+The usage of the branch number retrieved from the form's submission data will be replaced in
 Open Forms 4.0 by favouring the branch number obtained through the eHerkenning login.
 
 **DigiD/eHerkenning OIDC Strict mode**
--- a/src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
+++ b/src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
@@ -1,0 +1,93 @@
+from traceback import format_exc
+
+from django.core.management import BaseCommand
+from django.db import transaction
+
+from tabulate import tabulate
+
+from ...logic_analysis import CyclesDetected
+from ...models import Form
+
+
+class Command(BaseCommand):
+    help = """
+    Enable new logic evaluation for all forms.
+
+    Note that this will re-order the logic rules and remove any 'trigger from step'
+    setting, which cannot be easily reverted. So, for large forms, it is recommended
+    to create a copy of the form first, and enable the new logic evaluation feature flag
+    manually before running this command.
+
+    Any forms which contain rules with cycles will not be converted, and need to be
+    resolved manually as well. If such forms exist, the output will contain a list of
+    relevant form details.
+    """
+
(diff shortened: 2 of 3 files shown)


src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
```diff
@@ -0,0 +1,80 @@
...
+    relevant form details.
+    """
+
+    def handle(self, **options):
+        forms_to_convert = Form.objects.filter(
+            _is_deleted=False, new_logic_evaluation_enabled=False
+        )
+
+        if not forms_to_convert.exists():
+            self.stdout.write("All forms are already converted.")
```
vaszig, line 32:
```suggestion
            self.stdout.write("All active forms are already converted.")
```
Maybe something like this?Because the filter does two things.

vaszig (approved):

src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
```diff
@@ -0,0 +1,80 @@
...
+            return
+
+        forms_with_cycles_information = []
+        forms_to_update = []
+        for form in forms_to_convert:
```
sergei-maertens:
```suggestion
        for form in forms_to_convert.iterator():
```
these kind of scripts can unexpectedly load large amounts of data in memory :)

src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
```diff
@@ -0,0 +1,80 @@
...
+    setting, which cannot be easily reverted. So, for large forms, it is recommended
+    to create a copy of the form first, and enable the new logic evaluation feature flag
+    manually before running this command.
+
+    Any forms which contain rules with cycles will not be converted, and need to be
+    resolved manually as well. If such forms exist, the output will contain a list of
+    relevant form details.
+    """
+
+    def handle(self, **options):
```
sergei-maertens, line 26: you'll probably want an explicit `@transaction.atomic` here to prevent forms being stuck in some in-between state if something crashes!

CHANGELOG.rst
```diff
@@ -14,6 +14,31 @@ Changelog
...
+
+Regular bugfix release.
+
+.. note::
+
+    Because the legacy logic evaluation will be removed in Open Forms 4.0, we have added some
+    additional tooling to convert all forms to the new logic evaluation.
+
+    For each form, the logic rules will be reordered and any 'trigger from step' setting will be
+    removed. Note that this action cannot be easily reverted, So for large forms it is recommended to
```
sergei-maertens:
```suggestion
    removed. Note that this action cannot be easily reverted, so for large forms it is recommended to
```

src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
viktorvanwijk: Hmm, I think "active" could be confusing with the "Active" setting of a form. I could add "non-deleted", but not sure if it's relevant tbh :shrug:

src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
viktorvanwijk: Ah yeah, I meant to put that there :)

src/openforms/forms/management/commands/enable_new_logic_evaluation_for_all_forms.py
viktorvanwijk: I didn't want to use the decorator to avoid having to reconvert forms which didn't crash, but I realize now that I don't take any other exceptions into account when analyzing form logic. What if I add the transaction around `form.apply_logic_analysis` and catch a base `Exception` there?

sergei-maertens: seems reasonable to me yeah, though you'll want to be careful with the `bulk_update` that comes after too, otherwise we can still end up in a corrupt state

viktorvanwijk: I have included updating the `new_logic_evaluation_enabled` field in the transaction. The `bulk_update` only covers forms which have no steps and/or are of type appointment. If that fails, it's "OK" because nothing else was changed in those forms

sergei-maertens (approved):

PR #6272, opened 2026-05-11
[#6087] Drop support for HaalCentraal v13

### LINKED ISSUE #6087: Drop support for Haal Centraal BRP Personen bevragen < 2.0
As far as I can tell, HC BRP Personen bevragen 1.3 was never in production, and we haven't supported older versions, and there exist no newer versions.

Testing/developing/maintaining against this version causes overhead and challenges w/r to mock data, limiting our options to use VCR for tests. IMO it would be best to just drop support for this unused/outdated version.

vaszig:
Closes #6087

**Changes**

- Removed version 1.3
- Updated the tests
- Removed old JSON based tests and updated them to VCR

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


docs/installation/upgrade-400.rst
```diff
@@ -9,6 +9,16 @@ Open Forms 4.0 is a major version release that contains breaking changes.
    :local:
    :backlinks: none
 
+Removal of the unused HaalCentraal version 1.3
+==============================================
+
+HaalCentraal BRP Personen bevragen 1.3 was never in production so this is not available
+any more in Open Forms. The only supported version is v2 by default. All the related tests
+are now using VCR cassettes and are based on the docker container provided by BRP
+``https://brp-api.github.io/Haal-Centraal-BRP-bevragen/v2/getting-started#probeer-en-test-de-api-lokaal``.
```
sergei-maertens: The remark about VCR is not relevant here in this document. Users of Open Forms need not worry about VCR/mocks.

docs/installation/upgrade-400.rst
```diff
@@ -9,6 +9,16 @@ Open Forms 4.0 is a major version release that contains breaking changes.
...
+
```
sergei-maertens, line 16: I also think this should have a note that if you used 1.3, that the prefill configuration of components may require updating (fix plugin + attribute), as we cannot do that automatically I think (and it's not worth the effort given that probably nobody used this). But DH may be affected.

sergei-maertens (approved):
Nice cleanup - some remarks about the documentation but once those are processed, you can merge this.

PR #6280, opened 2026-05-11
Create digital address for betrokkene in Customer Interactions API

### LINKED ISSUE #6082: Use existing, non-default, email address in Partij in Open Klant and store this
### Thema / Theme

Form designer

"Als inwoner wil ik een bestaand (niet voorkeur) emailadres uit mijn profiel gebruiken voor een aanvraag"

In the client profile plugin, there is one scenario missing: As authenticated submitter, I have 2 emailaddresses (A and B), where A is my default. When I choose B in the dropdown, this is not stored anywhere.

This might be because 1 scenario was missing in the sequence diagram: https://platformvoordienstverlening.notion.site/Standaardiseren-klantinteractie-1f378b5f4db480b19384e535925f1724?pvs=74

Image

The change however is very simple: We store a **new** DIGITAALADRES for the BETROKKENE (even though it's an existing one). This is shown in scenario `1.a.ii.2` and mimics `1.a.i.2` *The reason we create a new one is because the stored list of emails (in PARTIJ) can change at any time, so we don't want to rely on that list. A case-specific email address however, can only be changed for that case.*

Image

Fixes missing scenario for client profiles.

_No response_

annashamray:
Closes #6082

**Changes**

* create a new digital address linked to "betrokkene" in Open Klant when
    * user selects existing address from dropdown
    * this is not a default address
    * they don't change the default preference (don't select "Sla mijn gegevens op
      voor de volgende keer.")

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/docs/configuration/prefill/communication_preferences.rst
+++ b/docs/configuration/prefill/communication_preferences.rst
@@ -94,14 +94,20 @@ The communication preferences prefill configuration is also used to update commu
 Customer Interactions API. To enable it you need to check "Open Forms should update customer data after submission"
 in the "Profile" component configuration.
 
-When the form is submitted Open Forms records following data in the Customer Interactions API:
+When the form is submitted, Open Forms records the following data in the Customer Interactions API:
 
-* klantcontact
-* betrokkene
-* onderwerpobject, where a public reference of the submission is stored
-* digitaleadressen, where new communication preferences are stored if user has chosen "Sla mijn gegevens op
-  voor de volgende keer." while filling the form.
-* partij, if user is authenticated in the form, but unknown in the Customer Interactions API.
+* ``klantcontact``
+* ``betrokkene``
+* ``onderwerpobject``, where a public reference of the submission is stored
+* ``digitaleadressen``:
+
+    The new and updated digital addresses for this interaction. The digital addresses are either linked to:
+
+    * ``betrokkene`` if the user is not authenticated or the selected address already exists for this ``partij``
+    * ``partij`` if the user is authenticated and they use the new address and choose "Sla mijn gegevens op
+      voor de volgende keer." while filling the form.
+
+* ``partij``, if user is authenticated in the form, but unknown in the Customer Interactions API.
 
 Technical
 =========
--- a/src/openforms/contrib/customer_interactions/update.py
+++ b/src/openforms/contrib/customer_interactions/update.py
@@ -139,9 +139,10 @@ def update_customer_interaction_data(
                 continue
 
             address_channel: SupportedChannels = digital_address["type"]
-            is_address_new_preferred: bool = bool(
+            is_address_new_preferred = (
                 digital_address.get("preferenceUpdate") == "isNewPreferred"
             )
+
             # prefill values
             prefill_communication_channel: CommunicationChannel | None = next(
                 (value for value in prefill_value if value["type"] == address_channel),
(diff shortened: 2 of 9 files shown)


src/openforms/contrib/customer_interactions/update.py
```diff
@@ -158,14 +159,28 @@ def update_customer_interaction_data(
...
+                # flow 5. we update it only if it's marked as "isNewPreferred"
                 if is_address_new_preferred and address_value != prefill_preferred:
                     updated_address = client.update_digital_address_for_party(
                         address=address_value, party_uuid=party_uuid, is_preferred=True
                     )
                     updated_addresses.append(updated_address)
 
+                # flow 3. we create a new address and link it to betrokkene
+                elif (
+                    not is_address_new_preferred and address_value != prefill_preferred
```
robinmolen: I had some difficulty trying to understand this condition (mostly because i wasn't sure what `prefill_preferred` was, as i forgot what the property `preferred` on the `CommunicationChannel` represented).

Because we have the same check for flow 5, it might be nice to define a `is_preferred` variable.

So:
```python
is_preferred_address: bool = bool(address_value == prefill_preferred)

...

# flow 3. we create a new address and link it to betrokkene
elif not is_address_new_preferred and not is_preferred_address:
```

Bonus: in the type definition of `CommunicationChannel` it might be nice to add a comment what the property `preferred` contains. Something like:
```python
class CommunicationChannel(TypedDict):
    type: SupportedChannels
    options: Sequence[str]
    preferred: str | None # The preferred address in this channel
```

docs/configuration/prefill/communication_preferences.rst
```diff
@@ -96,12 +96,16 @@ in the "Profile" component configuration.
 
 When the form is submitted Open Forms records following data in the Customer Interactions API:
```
robinmolen:
```suggestion
When the form is submitted, Open Forms records the following data in the Customer Interactions API:
```

docs/configuration/prefill/communication_preferences.rst
```diff
@@ -96,12 +96,16 @@ in the "Profile" component configuration.
...
-  voor de volgende keer." while filling the form.
-* partij, if user is authenticated in the form, but unknown in the Customer Interactions API.
+* ``klantcontact``
+* ``betrokkene``
+* ``onderwerpobject``, where a public reference of the submission is stored
+* ``digitaleadressen``:
+
+    * linked to ``betrokkene`` if the user is not authenticated or the selected address already exists for this ``partij``
+    * linked to ``partij`` if the user is authenticated and they use the new address and choose chosen "Sla mijn gegevens op
+      voor de volgende keer." while filling the form.
```
robinmolen, line 108: I think we need some more information here, about what `digitaleadressen` contains.

```suggestion
* ``digitaleadressen``:

    The new and updated digital addresses for this interaction. The digital addresses are either linked to:

    * ``betrokkene`` if the user is not authenticated or the selected address already exists for this ``partij``
    * ``partij`` if the user is authenticated and they use the new address and choose chosen "Sla mijn gegevens op
      voor de volgende keer." while filling the form.
```

robinmolen (dismissed):
I have a few small suggestions, but otherwise this looks good!

src/openforms/contrib/customer_interactions/update.py
annashamray: Yes, I think with these changes the code is more readable. I also changed `if` statements, I couldn't get rid of nested "ifs", but I think now is clearer what happens in each flow

src/openforms/contrib/customer_interactions/update.py
```diff
@@ -156,16 +157,33 @@ def update_customer_interaction_data(
                 if prefill_communication_channel
                 else None
             )
+            is_preferred_address: bool = bool(address_value == prefill_preferred)
```
sergei-maertens, line 161:
```suggestion
            is_preferred_address = address_value == prefill_preferred
```
the cast to `bool` is not necesary for a simple equality test

sergei-maertens (approved):

src/openforms/contrib/customer_interactions/update.py
annashamray: changed

PR #6296, opened 2026-05-15
Upgrade to django-privates 4.0+

sergei-maertens:
Periodic maintenance stuff and all that :)

Turns out this hits quite a bit of dependencies and test code. Rough overview of the changes:

* added the missing `@temp_private_root` decorators and/or updated them to not clear the files created in the test case data setup
* made (test) code agnostic about the storage implementation details - not everything is always backend by a local filesystem!
* updated related dependencies: zgw-consumers, django-simple-certmanager and django-digid-eherkenning that all make use of django-privates too
* ran makemigrations
* added CI step to fail the build if the effect of missing `@temp_private_root` decorators is detected (local private_media files are created)
* Updated a factory method to *not* create the submission PDF by default


sergei-maertens:
Merged like this - the changes were discussed with the team on Monday and no further input/checks were deemed necessary :)

PR #6299, opened 2026-05-18
[#6217] allow choosing document summaries for ZGW registrations

### LINKED ISSUE #6217: Allow a JSON-document to be added instead of a PDF with ZGW
### Thema / Theme

Admin

OF supports adding a PDF, with the submission details in it, as document to the Zaak. This is done with a specific DocumentType. Can we instead, or in addition to, add a JSON-file with the submission details as document to the Zaak.

I guess this can be determined by the mimetype in documenttype, or be made a switch in OF.. or whatever. When you select a documenttype in the dropdown, maybe show this mimetype as well 1) Bijlage (PDF), 2) Inzendingsdetails (JSON).

Adding a JSON document is a variant on extraElementen in StUF (I guess) that allows programmatic access to all the data by third party components, like ZAC or GZAC.

_No response_

SonnyBA:
Closes #6217 

**Changes**

Allows form builders to specify which registration summary documents they want to include for requests to the configured ZGW API:

image

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

[skip: e2e]


src/openforms/js/compiled-lang/nl.json
```diff
@@ -1237,6 +1237,12 @@
       "value": "Kies het soort logica:"
     }
   ],
+  "9EaH83": [
+    {
+      "type": 0,
+      "value": "Samenvattingsdocumenten"
```
viktorvanwijk: Can we leave adding translations for the release? 😇

src/openforms/contrib/zgw/service.py
```diff
@@ -104,3 +100,27 @@ def create_attachment_document(
                 "doc_vertrouwelijkheidaanduiding", ""
             ),
         )
+
+
+def create_json_document(
+    client: DocumentenClient,
+    name: str,
+    document_data: dict[str, dict[str, VariableValue] | JSONObject],
```
viktorvanwijk: Can be FormioData instance instead of `dict[str, VariableValue]` I think

src/openforms/contrib/zgw/service.py
```diff
@@ -104,3 +100,27 @@ def create_attachment_document(
...
+    options: DocumentOptions,
+    language: SupportedLanguage,
+) -> dict:
+    content = BytesIO(json.dumps(document_data, cls=DjangoJSONEncoder).encode("utf-8"))
+    return client.create_document(
+        informatieobjecttype=options["informatieobjecttype"],
+        bronorganisatie=options["organisatie_rsin"],
+        title=options.get("titel") or name,
```
viktorvanwijk: Why first check the "titel" and fallback to `name`, instead of using `name` directly?

src/openforms/js/components/admin/form_design/registrations/zgw/fields/SummaryDocuments.js
```diff
@@ -0,0 +1,60 @@
...
+        label={
+          <FormattedMessage
+            description="ZGW APIs registration options 'summaryDocuments' label"
+            defaultMessage="Summary documents"
+          />
+        }
+        helpText={
+          <FormattedMessage
+            description="ZGW APIs registration options 'summaryDocuments' help text"
+            defaultMessage="Wether to include summary documents in the request or not."
```
viktorvanwijk:
```suggestion
            defaultMessage="Which summary documents to include during registration."
```
Like this maybe?

src/openforms/js/components/admin/form_design/registrations/zgw/fields/SummaryDocuments.js
```diff
@@ -0,0 +1,60 @@
...
+      >
+        <ul>
+          {summaryDocumentChoices.map(([value, label], index) => (
+            <li key={index}>
+              <Checkbox
+                key={index}
+                label={label}
+                value={value}
+                name={fieldName}
+                defaultChecked={initialValue ? initialValue.find(v => v === value) : false}
```
viktorvanwijk:
```suggestion
                defaultChecked={initialValue ? initialValue.includes(value) : false
```
`find` will return the value instead of a boolean, no? Also, shouldn't the argument be `checked` instead of `defaultChecked`?

src/openforms/js/components/admin/form_design/registrations/zgw/fields/SummaryDocuments.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+SummaryDocuments.propTypes = {
+  summaryDocumentChoices: PropTypes.arrayOf(
+    PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.string, // value
+        PropTypes.string, // label
+      ])
+    )
```
viktorvanwijk:
```suggestion
    PropTypes.arrayOf(PropTypes.string)
```
Because they are both string right?

src/openforms/registrations/contrib/zgw_apis/tests/test_backend_partial_failure.py
```diff
@@ -10,6 +10,7 @@
 
 from openforms.contrib.objects_api.tests.factories import ObjectsAPIGroupConfigFactory
 from openforms.contrib.zgw.clients.zaken import ZakenClient
+from openforms.registrations.contrib.zgw_apis.constants import SummaryDocumentChoices
```
viktorvanwijk:
```suggestion
from ..constants import SummaryDocumentChoices
```
It's a local import :)

src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -47,6 +48,13 @@ class MappedVariablePropertySerializer(serializers.Serializer):
     )
 
 
+class JSONMultipleChoiceField(serializers.MultipleChoiceField):
+    def to_representation(self, value) -> list[str]:  # pyright: ignore[reportIncompatibleMethodOverride]
+        # Return a list as MultipleChoiceField fields return a set here by default
+        # which cannot be serialized to JSON.
+        return [value for value in super().to_representation(value)]
```
viktorvanwijk:
```suggestion
        return list(super().to_representation(value))
```
Should also work, no?

Also, seems a bit weird to me that this is necessary. Not a DRF expert, but isn't the `to_representation` method supposed to be serializable?

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
sergei-maertens, line 1: Can you keep this out of the diff? It's too large to render in Github anyway

src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -102,6 +110,12 @@ class ZaakOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serializer):
         default="",
     )
 
+    summary_documents = JSONMultipleChoiceField(
+        choices=SummaryDocumentChoices.choices,
+        label=_("Send a summary PDF document"),
```
viktorvanwijk: Could you add a label which is more representative?

src/openforms/forms/migrations/0130_set_default_summary_documents.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 5.2.14 on 2026-05-18 09:53
+
+from django.db import migrations
+from django.db.migrations.state import StateApps
+
+
+def add_default_summary_document(apps: StateApps, _) -> None:
```
sergei-maertens: Do we need the data migration? Can't this be a regular `default=...` kwarg on the options serializer field?

src/openforms/registrations/contrib/zgw_apis/typing.py
```diff
@@ -57,3 +59,4 @@ class RegistrationOptions(TypedDict):
     auteur: NotRequired[str]
     zaak_omschrijving: NotRequired[str]
     zaak_toelichting: NotRequired[str]
+    summary_documents: NotRequired[list[SummaryDocumentChoices]]
```
viktorvanwijk: With the added migration, doesn't it make more sense to have this as required?

I also think `Collection` is a better type hint here, as it doesn't really matter whether it's a list, tuple, set, etc...

src/openforms/forms/migrations/0130_set_default_summary_documents.py
```diff
@@ -0,0 +1,56 @@
...
+        .exclude(options__isnull=True)
+        .filter(
+            backend="zgw-create-zaak"
+        )
+    )
+    # fmt:on
+
+    for backend in zgw_registration_backends:
+        if "summary_documents" in backend.options:
+            print(f"Not updating backend {backend}")
```
sergei-maertens: left-over print statement

src/openforms/forms/migrations/0130_set_default_summary_documents.py
```diff
@@ -0,0 +1,56 @@
...
+        FormRegistrationBackend.objects
+        .exclude(options__isnull=True)
+        .filter(
+            backend="zgw-create-zaak"
+        )
+    )
+    # fmt:on
+
+    for backend in zgw_registration_backends:
+        if "summary_documents" in backend.options:
```
sergei-maertens: you can already include that in the `.filter` query I think so that you only return backends that don't have this key at all.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -422,38 +430,85 @@ def register_submission(
...
-                    document=summary_pdf_document,
-                ),
-                submission,
-                "intermediate.documents.report.relation",
-            )
+            if SummaryDocumentChoices.json in summary_documents:
+                state = submission.variables_state
+                all_values = state.get_data(include_static_variables=True)
+                variables = [key for key in all_values]
+                values = {key: all_values[key] for key in variables}
```
viktorvanwijk: A few questions/remarks about this:
1. What exactly are you trying to achieve here? Won't `all_values` just equal `values` at the end?
2. Iterating over a `FormioData` instance is not allowed, because you will not get the full variable keys if there is nested data. For example, component with key "foo.bar" will be present in the data as `{"foo": {"bar": "baz"}}`, so iterating over it will just give you "foo".
3. Does it make sense to include static variables here? I don't think they are in the PDF summary right?

Also, I think we need some processing of the values and schema here. Looking at the generic json plugin, partners and children data need to be cleaned up. Also, I'm not sure if it makes sense to include the submission data from file components as is, because it just contains metadata about the upload to the server. Would it be possible to relate it to the created attachment documents, like in the Objects API?

Given that it's the third plugin that requires this data cleanup, perhaps this is the part that can be refactored/generalized? Specifically processing partners and children will be the same for all plugins _I think_. (still a bit unclear to me whether this is worth it tbh, because there is also plugin-specific things happening)

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -422,38 +430,85 @@ def register_submission(
...
+                document_data = {
+                    "values": values,
+                    "values_schema": values_schema,
+                }
+                summary_json_document = execute_unless_result_exists(
+                    partial(
+                        create_json_document,
+                        client=documents_client,
+                        document_data=document_data,
+                        name=submission.form.name,
```
viktorvanwijk: Should we include PDF or JSON in the name?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -422,38 +430,85 @@ def register_submission(
...
+                        options=summary_document_options,
+                        language=submission_report.submission.language_code,
+                    ),
+                    submission,
+                    "intermediate.documents.json.document",
+                )
+                result["intermediate"]["documents"]["json"]["schema"] = document_data[
+                    "values_schema"
+                ]
```
viktorvanwijk: Why the change in key names here?

viktorvanwijk (changes requested):

src/openforms/contrib/zgw/service.py
SonnyBA: Upon having another look at the `DocumentOptions` and [the plugin code](https://github.com/open-formulieren/open-forms/blob/3f1eb2815/src/openforms/registrations/contrib/zgw_apis/plugin.py#L433) I saw that the `title`  attribute was never passed through (also applies for the PDF summary) so removed the usage of `title` here.

src/openforms/registrations/contrib/zgw_apis/typing.py
sergei-maertens: With my comment above about the migrations, I think these aren't even necesary:

* define a `default` on the serializer field
* mark this property as required, so remove the `NotRequired` which is guaranteed by the default in the serializer
* migration is obsolete

The bonus is that (see checklist!) form imports from older versions will still properly work, because you don't need a converter/manual step as it will be handled by the serializer field default.

src/openforms/registrations/contrib/zgw_apis/options.py
SonnyBA: I'm guessing this is handled by DRF's json encoder when generating responses but becomes a problem when you try to save these values into a `JSONField` in the database (which uses the python's stdlib encoder by default).

src/openforms/formio/registry.py
```diff
@@ -97,6 +97,8 @@ class BasePlugin(Generic[ComponentT], AbstractBasePlugin, abc.ABC):  # noqa: UP0
     Flag to indicate whether data can be submitted for this component.
     """
 
+    internal_properties: set[str] | None = None
```
viktorvanwijk: Could you add some docs here, like the others 😇 

Also, maybe it can be a `frozenset` to be more explicit that it shouldn't be changed?

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -930,3 +998,200 @@ def update_registration_with_confirmation_email(
...
+                assert isinstance(schema["items"], dict)
+                _properties = schema["items"]["properties"]
+                assert isinstance(_properties, dict)
+
+                _properties["affixes"] = {"type": "string"}
+                _properties["initials"] = {"type": "string"}
+                _properties["lastName"] = {"type": "string"}
+            case {"type": "file"}:
+                assert isinstance(schema["items"], dict)
+                schema["items"] = {"type": "string", "format": "uri"}
```
viktorvanwijk: I think it would be good to keep this component consistent with the Generic JSON and Objects API, where the value (and schema) depends on `multiple`. This gives the following four cases:

1. No multiple, no attachment -> empty string
2. No multiple, one attachment -> uri
3. Multiple, no attachment -> empty list
4. Multiple, one or more attachments -> list of uris

This keeps it consistent with the other components, of which the values will be converted to a list only when `multiple: True`. 

(Admittedly, the fact that submission data for a file component is always a list is confusing here imo)

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -930,3 +998,200 @@ def update_registration_with_confirmation_email(
...
+                    continue
+
+                component_plugin = as_component_plugin(component)
+                assert component_plugin.internal_properties is not None
+                for internal_property in component_plugin.internal_properties:
+                    child.pop(internal_property, None)
+
+                updated.append(child)
+
+            values[key] = updated
```
viktorvanwijk: These still share a lot of similarities with how Generic JSON and Objects API process the value, but there is also plugin-specific things happening in these routines, so we cannot use one routine directly for all of them.

Perhaps a general registration helper `process_children_and_partners_data`, and then here:

```python
match component:
    case {"type": "children" | "partners"}:
         process_children_and_partners_data(...)
```

@sergei-maertens what are your thoughts on this?

(Same comment about the process_variables_schema method of the plugin btw. Although the schema of children and partners components do not seem to be processed in the Objects API. Not sure if this is intentional or an oversight 🤔 )

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -686,9 +700,63 @@ def register_submission(
...
+                    ),
+                    submission,
+                    "intermediate.documents.json.document",
+                )
+                result["intermediate"]["documents"]["json"]["values_schema"] = (
+                    document_data["values_schema"]
+                )
+                result["intermediate"]["documents"]["json"]["values"] = document_data[
+                    "values"
+                ]
```
viktorvanwijk:
```suggestion
                result["intermediate"]["documents"]["json"].update(document_data)
```

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -686,9 +700,63 @@ def register_submission(
                     f"intermediate.documents.{attachment.id}.relation",  # type: ignore
                 )
 
+            if SummaryDocumentChoices.json in summary_documents:
+                state = submission.variables_state
+                all_values = state.get_data(include_static_variables=True)
```
viktorvanwijk: `state.variables` does not include static variables, so they are not added, but perhaps use `include_static_variables=False` as well to not create any ambiguity

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -930,3 +998,200 @@ def update_registration_with_confirmation_email(
...
+
+        )
+        .filter(
+            submission_step__submission=submission,
+            submission_variable__isnull=False
+        )
+    )
+    # fmt: on
+
+    registration_result = submission.registration_result
```
viktorvanwijk: Minor thing, but perhaps just pass the result directly to this function, instead of doing it through the submission? This flow might be more difficult to follow, and it's not a generic helper function anyway

src/openforms/submissions/migrations/0013_merge_20260526_1606.py
viktorvanwijk, line 1: Instead of adding this merge migration, could you just change the added migration to be dependent on the new 0012? These merge files are noise that can be avoided imho

src/openforms/formio/service.py
```diff
@@ -163,6 +163,13 @@ def build_serializer(
     return _build_serializer(components, register=_register or register, **kwargs)
 
 
+def as_component_plugin(
+    component: Component, _register: ComponentRegistry | None = None
+) -> BasePlugin:
+    registry = _register or register
+    return registry[component["type"]]
```
sergei-maertens: This seems to expose the internal `openforms.formio` API into the rest of the application, I don't think that's a good idea. I'm already feeling the pain from all the violations in the msgspec refactor branch 😩

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -413,38 +427,38 @@ def register_submission(
...
-            execute_unless_result_exists(
-                partial(
-                    zaken_client.relate_document,
-                    zaak=zaak,
-                    document=summary_pdf_document,
-                ),
-                submission,
-                "intermediate.documents.report.relation",
-            )
+            summary_documents = options.get("summary_documents", []) or []
```
sergei-maertens:
```suggestion
            summary_documents = options["summary_documents"]
```
the serializer having a default implies that this key is `Required` and direct key access can be used

src/openforms/formio/registry.py
sergei-maertens: I'll take a closer look in light of the refactor potential as well, I'm not excited by this pattern and how it's not used at all inside `openforms.formio` but only in other apps :thinking:

sergei-maertens: Refactored this out

src/openforms/registrations/contrib/zgw_apis/plugin.py
sergei-maertens: Refactored this out - moved into the formio comonent registry :tada:

src/openforms/registrations/contrib/zgw_apis/plugin.py
sergei-maertens: values/value_schema is excluded from the intermediate results. it's bloat and not encrypted or hashed. Instead the tests read the JSON document from the Documents API

src/openforms/registrations/contrib/zgw_apis/plugin.py
sergei-maertens: Incorporated!

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -718,9 +736,57 @@ def register_submission(
...
+            # needs to be done *after* the attachments have been uploaded to the
+            # Documenten API
+            if SummaryDocumentChoices.json in summary_documents:
+                state = submission.variables_state
+                variable_keys = list(state.variables.keys())
+                all_values = state.get_data(
+                    include_static_variables=False,
+                    # TODO: check if we want to be dense or sparse, IMO it should be
+                    # dense for consistency
+                    include_unsaved=False,
```
sergei-maertens: @viktorvanwijk opinions welcome :D

src/openforms/registrations/json_output.py
sergei-maertens, line 1: Looking at the size of the PR/diff, I've opted to *not* chase merging `as_json_data` and `as_json_schema` into a single thing 😬

viktorvanwijk: Yeeaahh :grimacing:

src/openforms/formio/components/custom.py
```diff
@@ -696,6 +696,17 @@ def build_serializer_field(
             **extra,
         )
 
+    @staticmethod
+    def as_json_data(component: Component, value: VariableValue) -> VariableValue:
+        """
+        Filter out deselected children and drop internal keys from the data.
```
viktorvanwijk: Children in addressNL? 🤔

src/openforms/formio/components/custom.py
```diff
@@ -830,6 +842,18 @@ class Partners(BasePlugin[Component]):
     def build_serializer_field(self, component: Component) -> PartnerListField:
         return PartnerListField(component=component)
 
+    @staticmethod
+    def as_json_data(component: Component, value: VariableValue) -> VariableValue:
+        """
+        Drop the internal data keys from the partner objects.
+        """
+        assert isinstance(value, Sequence)
+        for partner in value:
```
viktorvanwijk, line 852: I see that the value for addressNL is first copied and then mutated. Should we do the same here?

src/openforms/formio/components/custom.py
```diff
@@ -990,6 +1016,26 @@ class Children(BasePlugin[ChildrenComponent]):
...
         return ChildListField(component=component)
 
+    @staticmethod
+    def as_json_data(component: Component, value: VariableValue) -> VariableValue:
+        """
+        Filter out deselected children and drop internal keys from the data.
+        """
+        assert isinstance(value, Sequence)
+        selected_children = []
+        for child in value:
```
viktorvanwijk, line 1028: Same question about copying here

src/openforms/formio/registry.py
```diff
@@ -140,6 +140,24 @@ def localize(self, component: ComponentT, language_code: str, enabled: bool):
     def build_serializer_field(self, component: ComponentT) -> serializers.Field:
         """Translate a component configuration into a serializer."""
 
+    @staticmethod
+    def as_json_data(component: Component, value: VariableValue) -> VariableValue:
```
viktorvanwijk:
```suggestion
    def as_json_data(component: ComponentT, value: VariableValue) -> VariableValue:
```

src/openforms/registrations/json_output.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def get_json_data(
+    submission: Submission,
+    *,
+    values: FormioData,
+    limit_to_variables: Sequence[str],
+    backend_id: str,
+    backend_options: AnyOptionsDict,
+    post_process_component: ComponentPostProcessHook | None = None,
```
viktorvanwijk, line 101: Could you add this parameter to the docstring as well?

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
```diff
@@ -3410,3 +3568,388 @@ def test_addressnl_empty_or_broken_mapping(self, submitted_value):
...
+
+        submission.refresh_from_db()
+        assert submission.registration_result
+
+        document_results = submission.registration_result["intermediate"]["documents"]
+        pdf_details = document_results["report"]["document"]
+        self.assertEqual(pdf_details["titel"], "Public form name (PDF)")
+        self.assertTrue("json" not in document_results)
+
+    def test_no_summary_documents(self):
```
viktorvanwijk: Is it worth keeping this test? The fact that "documents" is not in the registration result doesn't really prove anything it seems like

viktorvanwijk (commented):
Few minor questions/comments :)

I'm trusting the existing tests to account for any behavior changes hehe

src/openforms/registrations/contrib/zgw_apis/plugin.py
viktorvanwijk: I agree, though at the moment it still seems to be a mix :grimacing: Generic JSON does not include them for example

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
sergei-maertens: agree!

src/openforms/formio/components/custom.py
sergei-maertens: caught in 4K copy pasting

src/openforms/formio/components/custom.py
sergei-maertens: yeah I think a copy if safer - if future code relies on the non-processed values, we don't want those side-effects to have been applied.

src/openforms/registrations/json_output.py
sergei-maertens: there's always one I miss :facepalm:

viktorvanwijk (approved):

PR #6301, opened 2026-05-19
Add detection script for breaking changes regarding clear on hide

### LINKED ISSUE #6274: Add detection script for breaking change regarding clearOnHide
The breaking change:
JSON logic expressions which compare to the empty value of a component, can never trigger if that component is hidden, because it is removed from the data rather than set to empty/default value.

This script needs to be able to detect rules (JSON logic triggers) that compare the value of a variable to the related component empty value. Possibly this can be narrowed down by also checking whether the visibility of that component is affected by another rule and/or conditional logic. Though perhaps this will take too much time, and we can get away with just the first check. This would also mean checking whether the (possible) parent components are affected by (conditional) logic.

Can be backported to allow for easier preparation for municipalities

viktorvanwijk:
Closes #6274

[skip: e2e]

**Changes**

* Added detection script
* Added changelog placeholder

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/bin/report_logic_with_deprecated_clear_on_hide_behavior.py
+++ b/bin/report_logic_with_deprecated_clear_on_hide_behavior.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python
+# This script detects rules (JSON logic triggers) that compare the value of a variable
+# to the related component empty value.
+from __future__ import annotations
+
+import sys
+from collections import defaultdict
+from pathlib import Path
+
+import django
+
+import click
+from json_logic.meta import JSONLogicExpression
+from json_logic.meta.expressions import destructure
+from json_logic.typing import JSON
+from tabulate import tabulate
+
+SRC_DIR = Path(__file__).parent.parent / "src"
+sys.path.insert(0, str(SRC_DIR.resolve()))
+
+
+def iter_variables_with_compare_value(expression: JSON):
+    if isinstance(expression, list):
+        for expr in expression:
+            yield from iter_variables_with_compare_value(expr)
+
+    if not isinstance(expression, dict):
+        # We have reached a primitive, so there's nothing to do
+        return
+
+    normalized = JSONLogicExpression.normalize(expression)
+    assert isinstance(normalized, dict)
+    operator, argument = destructure(normalized)
+
+    # I think using `>`, `>=`, `<`, and `<=` to perform comparisons with empty
+    # containers and strings seems like a SUPER RARE, ULTRA DIRTY edge case that we can
+    # leave out here :sweat_smile: Let those rules break and force form designers to
+    # rewrite them
+    if operator in ["==", "!=", "===", "!=="]:
+        a, b = argument
+        match a, b:
+            case {"var": _}, {"var": _}:
+                # Cannot determine anything if we have two variable expressions
+                return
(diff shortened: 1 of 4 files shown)


bin/report_logic_with_deprecated_clear_on_hide_behavior.py
```diff
@@ -0,0 +1,238 @@
...
+    from openforms.forms.models import Form, FormLogic
+    from openforms.variables.service import resolve_key
+
+    forms_to_check = Form.objects.filter(_is_deleted=False).prefetch_related(
+        "formlogic_set", "formstep_set"
+    )
+    form_with_logic_with_risk: defaultdict[tuple[Form, FormLogic], set[str]] = (
+        defaultdict(set)
+    )
+    for form in forms_to_check.iterator(chunk_size=10):
```
vaszig, line 101: I didn't know that iterator is used now in django along with the prefetch_related, nice!

bin/report_logic_with_deprecated_clear_on_hide_behavior.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Mapping from component to step for quick access
+        form_steps = form.formstep_set.select_related("form_definition")
+        component_map: dict[str, Component] = {}
+        for form_step in form_steps:
+            for component in iter_components(
+                form_step.form_definition.configuration,
+                recursive=True,
+                recurse_into_editgrid=False,
+            ):
```
vaszig, line 112: I guess you know better with all these refactors and the code you touched but why don't we use the `form.iter_components` which does the same thing afaik?Is it because of the editgrids flag?

vaszig (commented):
I am not the most suitable person to review this but I do not see any weird things. I went through the code and have some questions for my better understanding.Maybe we can have a quick call too in order to get a bit more familiar with some things.

bin/report_logic_with_deprecated_clear_on_hide_behavior.py
```diff
@@ -0,0 +1,238 @@
...
+
+        for rule in form.formlogic_set.iterator():
+            for var_name, comp_value in iter_variables_with_compare_value(
+                rule.json_logic_trigger
+            ):
+                if (resolved_key := resolve_key(var_name, component_map)) is None:
+                    # Variable cannot be resolved, so we cannot determine anything
+                    continue
+
+                component = component_map[resolved_key]
```
sergei-maertens, line 147: ~~wouldn't this raise a `KeyError` when trying to look up a key like `editgrid.7.nested`?~~

ok because `resolve_key` handles that :-)

sergei-maertens (approved):

bin/report_logic_with_deprecated_clear_on_hide_behavior.py
sergei-maertens: indeed, by default it recurses into editgrids which has turned out to have been A Mistake :tm: 

I was wondering if using the formio configuration wrapper datastructure would be better, but for a check script this is plenty readable!

sergei-maertens:
Backported to 3.5: ba03ce26bd0d2fea7981b937b3cf7879edae6f32

PR #6305, opened 2026-05-20
Migrate API groups configuration to backend options

### LINKED ISSUE #4939: Remove legacy zaaktype/documenttype URL fields in registration backends
Currently we support he indirect references by using omschrijving/identificatie and the legacy explicit version URLs for documenttypes/zaaktypes in the Objects API and ZGW API's registration. The latter can be removed as a cleanup.

**NOTE**

This requires the file upload component URL references to be replaced with documenttype descriptions too, and made context-aware of the registration backend that they're used by (moving out of formio builder into the variables tab).

**Tasks**

- [x] Add detection script/upgrade check blocking upgrades if legacy fields are still used somewhere (#6309)
- [ ] Delete the fields `ZaakOptionsSerializer.zaaktype`, `ZaakOptionsSerializer.informatieobjecttype`, `ZaakOptionsSerializer.objecttype`
- [ ] Delete the fields `ObjectsAPIOptionsSerializer.informatieobjecttype_submission_report`, `ObjectsAPIOptionsSerializer.informatieobjecttype_submission_csv`, `ObjectsAPIOptionsSerializer.informatieobjecttype_attachment`
- [ ] Delete the fields `ObjectsAPIGroupConfig.informatieobjecttype_submission_report`, `ObjectsAPIGroupConfig.informatieobjecttype_submission_csv` and `ObjectsAPIGroupConfig.informatieobjecttype_attachment`

sergei-maertens:
Partly closes #4939 - best to get #6309 merged first

**Changes**

- [x] Added data migration for Objects API group options
- [x] Added data migration for ZGW API group options
- [x] Remove/hide legacy/obsoleted configuration fields
- [x] Fix migrator code - this needs to be run on 3.5.x but we also need it for our latest env until 4.0 releases.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


sergei-maertens:
Import/export of (old) forms exports can't be addressed here, because the API group is not present in the export and we're removing these fields from the existing groups. Users will need to re-generate the exports on 4.0.

sergei-maertens:
@SonnyBA this follows up on #6268, though before merging this I want to run the migrator tool on our test env and implement the upgrade check :wink:

src/openforms/forms/migrations/0131_move_api_group_configuration_to_backends.py
```diff
@@ -0,0 +1,118 @@
...
+
+def move_objects_group_config_to_backends(apps, _):
+    ObjectsAPIGroupConfig = apps.get_model("objects_api", "ObjectsAPIGroupConfig")
+    FormRegistrationBackend = apps.get_model("forms", "FormRegistrationBackend")
+
+    # there is a DB constraint that validates both domain and RSIN are set at the
+    # same time, or none are set
+    api_groups_by_identifier = ObjectsAPIGroupConfig.objects.exclude(
+        catalogue_domain=""
+    ).in_bulk(field_name="identifier")
```
annashamray, line 18: TIL! Nice method!

docs/installation/upgrade-400.rst
```diff
@@ -261,4 +261,50 @@ For more information about the new logic evaluation, please refer to the
...
+* users of setup-configuration may need to update their YAML assets that would populate
+  the catalogue domain/RSIN.
+* form exports pre 4.0 will not import cleanly into 4.0 environments, it's recommended
+  to upgrade the source instance to 4.0 and re-generate the exports. Most notably, the
+  registration backend options must contain:
+
+  - an API group reference.
+  - a catalogue reference (domain + RSIN) combination.
+
+Objects API registration
```
annashamray, line 293: I think this documentation still have configuration of removed fields for Objects API group: 
* https://open-forms.readthedocs.io/en/latest/configuration/registration/objects.html (step 7.)

src/openforms/forms/tests/test_migrations.py
```diff
@@ -571,3 +574,260 @@ def test_migration(self):
...
+        )
+        FormRegistrationBackend.objects.create(
+            form=form,
+            key="group-to-update",
+            name="Should be updated",
+            backend="zgw-create-zaak",
+            # incomplete according to the types/serializer, but shouldn't break the
+            # migration!
+            options={
+                "zgw_api_group": group_2.pk,
```
annashamray, line 771: I don't think it's possible to configure ZGW API registration backend without specifying a catalog:
* if you leave it empty, zaaktype dropdown is also empty, which will lead to the validation error when you try to save it
* If you add a legacy zaaktype url, there still be a validation error, because such zaaktype doesn't exist in the selected catalog (which is empty)

I understand that the JSON field can be edited, but I think it's more an edge case now.

src/openforms/registrations/contrib/zgw_apis/admin.py
```diff
@@ -44,16 +44,6 @@ class ZGWApiGroupConfigAdmin(admin.ModelAdmin):
...
             },
         ),
-        (
-            _("Catalogue"),
-            {
-                "description": _(
-                    "Specify the catalogue in the selected catalogi API service where "
-                    "the case and document types are defined."
-                ),
-                "fields": ("catalogue_domain", "catalogue_rsin"),
```
annashamray, line 54: Are you going to remove these fields in the next PR? (same question is for Objects API groups)

annashamray:
Love the amount of removed lines :fire:

src/openforms/forms/tests/test_migrations.py
sergei-maertens: But on 3.5 you can still have the legacy URLs without catalogue - so the starting situation for the migration where `catalogue` key is missing is realistic, right?

src/openforms/registrations/contrib/zgw_apis/admin.py
sergei-maertens: In the next release (4.1) is the soonest we can remove them, because these fields are used in the upgrade check script.

src/openforms/forms/tests/test_migrations.py
annashamray: I haven't managed to reproduce it on `stable/3.5.x` branch:

image


I believe it's this [line of code](https://github.com/open-formulieren/open-forms/blob/98e32d8a9fff6d07497580db65f073527a952d43/src/openforms/registrations/contrib/zgw_apis/options.py#L486)

But I got that the configuration can be inherited from older versions, just I hope that with this validation check no a lot of forms are kept without catalog registration option

sergei-maertens: Hrm odd, that line should only trigger if there's a case type identification specified, but I'm not seeing that in the UI?

annashamray (approved):

PR #6308, opened 2026-05-22
support contactpersoonRol attributes for ZGW backend

### LINKED ISSUE #5984: ZGW plugin contactpersoonRol email, phone and name
### Thema / Theme

Plugins

**Current situation**
- Admin is able to select registration "Initiator > contactpersoonNaam", "Initiator > emailadres" and "Initiator > telefoonnummer"

Image

- The ZGW api is able to receive a contactpersoonRol with required Naam and optional emailadres, telefoonnummer and functie properties within a POST rol. 

Image

- ZGW standards 1.5.1 have this contactpersoonRol in the Rol, but concept 1.6 as well. ZGW 2.0 status unknown.
- The ZGW goal is to add contactinformation through the klanten standard. However, thiis has not been provided as a standard as of now and probably will take some time. https://github.com/VNG-Realisatie/gemma-zaken/blob/96ca2c0f77f3930efb4c128bf51c0f3059854a61/api-specificatie/zrc/1.6.x/1.6.0/openapi.yaml#L10937

- In the plugin code these properties have been disabled https://github.com/open-formulieren/open-forms/blob/5b956b957cae583b1118884854703df92d96dd9b/src/openforms/registrations/contrib/zgw_apis/plugin.py#L196
- But the properties have been added everywhere: https://github.com/open-formulieren/open-forms/blob/5b956b957cae583b1118884854703df92d96dd9b/src/openforms/registrations/constants.py#L38

- Right now when these properties have been selected for a text field registration, the outgoing Rol call does not contain the Initiator ContactpersoonNaam, Emailadres and Telefoonnummer. Even when they have been connected in the registration tab to the text field.

- The main issue it the "what if" situation. What happens if we remove contactpersoonRol from the ZGW standard. However, when and if this will happen is unknown

**New situation**
- When a Initiator Contactpersoonnaam has been added to the ZGW registration, it will add a contactpersoonrol to the rol
- When Initiator emaildres and or telefoonnummer have been added to text fields, they will be added to the contactpersoonrol
- Admins are warned in the documentation that this situation might change due to the changing ZGW standard
- Admins are warned in the documentation that contact information is best set through the Klantapi
- Suppliers who use contactpersoonRol in their domain specific applications with Zaken are warned to not rely on contactpersoonRol and base strategies on kalntenapi in the future

- Municipalities (Gemeenten) who do not have a OpenKlant instances are able to see the contact information in the Rol of the Zaak until they do get a Klantinstance
- RxMission ZGW cases from Open Forms can show contactinformation until a Klantconnection van be made

What am I missing when it comes to the risks?
- ContactpersoonRol being removed from ZGW standard after 1.6
- Municipalities and some suppliers who might rely too much on a "possibly temporary" Contactpersoonrol
- Other zaak systems that cannot deal with ContactpersoonRol
- Stuf issues?

Since you already built most of the technical implementation, I suppose this is mote an issue of risks and standards?

annashamray:
[skip: e2e]
Closes #5984

**Changes**

* add 3 contactpersoonRol attribute mappings
* document how to configure it 

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/docs/manual/forms/form_fields.rst
+++ b/docs/manual/forms/form_fields.rst
@@ -183,7 +183,9 @@ Deze worden naar de ZGW API's verstuurd als een ``contactpersoonRol``-object:
 
   ``naam`` is een verplicht attribuut volgens de ZGW API's-specificatie. Geef dit altijd
   mee wanneer je een ander ``contactpersoonRol``-attribuut gebruikt (``telefoonnummer``
-  en/of ``emailadres``), anders retourneren de ZGW API's een foutmelding.
+  en/of ``emailadres``), anders retourneren de ZGW API's een foutmelding. Bovendien wordt
+  een lege waarde voor de ``naam`` niet ondersteund in de ZGW-API's, dus zorg ervoor dat
+  het bijbehorende formulierveld verplicht is.
 
 .. warning::
 
--- a/docs/manual/forms/form_fields.rst
+++ b/docs/manual/forms/form_fields.rst
@@ -131,6 +131,68 @@ Registratie
   naar het achterliggende registratie systeem, dan kun je hier een attribuut
   kiezen dat beschikbaar is in het achterliggende registratiesysteem.
 
+De registratiesystemen die registratieattributen ondersteunen:
+
+* StUF-ZDS
+* :ref:`manual_registration_zgw_apis`
+
+Tijdens het registratieproces wordt de ingediende data gekoppeld aan de geselecteerde
+attributen en verstuurd naar de registratieservice.
+
+.. note::
+
+  Verschillende registratiesystemen kunnen dezelfde registratieattributen op een
+  andere manier verwerken, afhankelijk van de mogelijkheden en beperkingen van
+  externe API's.
+
+ZGW API's initiator
+~~~~~~~~~~~~~~~~~~~
+
+In de ZGW API's wordt de aanvrager als initiator geregistreerd bij de zaak.
+
+Je kunt de formuliervelden koppelen aan verschillende initiatorattributen, bijvoorbeeld:
+
+* Initiator > Voorletters
+* Initiator > Geslachtsnaam
+* etc.
+
+**Contactpersoon**
+
+Bij het configureren van registratieattributen moet je rekening houden met hoe ze
+worden verwerkt in het registratiesysteem.
+Er zijn 3 beschikbare attributen voor contactpersoon:
+
+* Initiator > ContactpersoonNaam
+* Initiator > Telefoonnummer
+* Initiator > Emailadres
+
+Deze worden naar de ZGW API's verstuurd als een ``contactpersoonRol``-object:
+
+.. code-block:: json
+
+    {
+        // andere rollenattributen
+        "contactpersoonRol": {
+            "naam": "<waarde uit het veld 'Initiator > ContactpersoonNaam'>",
+            "telefoonnummer": "<waarde uit het veld 'Initiator > Telefoonnummer'>",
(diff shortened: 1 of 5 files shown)


docs/manual/forms/form_fields.rst
```diff
@@ -131,6 +131,68 @@ Registratie
...
+ZGW API's initiator
+~~~~~~~~~~~~~~~~~~~
+
+In de ZGW API's wordt de aanvrager als initiator geregistreerd bij de zaak.
+
+Je kunt de formuliervelden koppelen aan verschillende initiatorattributen, bijvoorbeeld:
+
+* Initiator > Voorletters
+* Initiator > Geslachtsnaam
+* en anderen…
```
viktorvanwijk:
```suggestion
* en andere...
```
Adding the 'n' is only necessary when referring to people :)

Using `etc.` is also OK here

docs/manual/forms/form_fields.rst
```diff
@@ -131,6 +131,68 @@ Registratie
...
+* Initiator > ContactpersoonNaam
+* Initiator > Telefoonnummer
+* Initiator > Emailadres
+
+Deze worden naar de ZGW API's verstuurd als een ``contactpersoonRol``-object:
+
+.. code-block:: json
+
+    {
+        // other rol attributes
```
viktorvanwijk: Maybe remove this commented line, or change to Dutch?

src/openforms/contrib/zgw/clients/zaken.py
```diff
@@ -127,6 +127,9 @@ def create_rol(
             "betrokkeneIdentificatie": betrokkene.get("betrokkeneIdentificatie", {}),
         }
 
+        if contactperoon_rol := betrokkene.get("contactpersoonRol"):
```
viktorvanwijk, line 130: Should we perhaps just specify a default empty string if `naam` is not present, instead of leaving it out and crashing hard? I'm seeing a lot of defaults above, but I'm not entirely sure if there is a general pattern for this 🤔

viktorvanwijk (commented):
Nice Dutch addition to the manual!

docs/manual/forms/form_fields.rst
annashamray: Changed to 'etc.'

src/openforms/contrib/zgw/clients/zaken.py
annashamray: It's a good question! With the empty string it will still crash (I've tried it), we can make some text to be default like for `roltoelichting` attribute, but I don't know if it's a good idea. I do make a note in the documentation that `initiator_contactpersoonNaam` is required

docs/manual/forms/form_fields.rst
annashamray: Forgot to translate it! I'll translate and keep it, because I want to show that there are other attributes there,

src/openforms/contrib/zgw/clients/zaken.py
viktorvanwijk: > With the empty string it will still crash (I've tried it)

Hmm interesting. So that means that, in addition to specifying the `Initiator > ContactpersoonNaam` registration attribute on a "name" component, it needs to be marked as required as well. Otherwise it could still end up crashing on the empty string. Maybe good to mention that in the documentation note too?

annashamray: Added to the documentation note!

viktorvanwijk (approved):

PR #6323, opened 2026-05-27
Fix translations not being "activated" for forms that use client-side logic evaluation

### LINKED ISSUE #6320: AddressNL - translations missing

3.5.2

### Customer reference

Venlo 140

config:
Image

in "en" form:
Image

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #6320

**Changes**

* Added missing test
* Added regression test
* Ensure formio-level translations are applied for the `defaultConfiguration` which is used by the SDK if backend logic evaluation is not necessary

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/api/serializers.py
+++ b/src/openforms/submissions/api/serializers.py
@@ -28,6 +28,7 @@ from openforms.formio.api.fields import FormioDataField
 from openforms.formio.service import (
     FormioData,
     build_serializer,
+    get_dynamic_configuration,
     rewrite_formio_components_for_request,
 )
 from openforms.formio.utils import iter_components
@@ -400,6 +401,21 @@ class SubmissionStepSerializer(NestedHyperlinkedModelSerializer):
             wrapper = deepcopy(instance.form_step.form_definition.configuration_wrapper)
             rewrite_formio_components_for_request(
                 wrapper, request=self.context["request"]
+            )
+            variables_state = instance.submission.variables_state
+            wrapper = get_dynamic_configuration(
+                wrapper,
+                submission=instance.submission,
+                # dynamic formio component configuration may depend on filled out data,
+                # but those cases should land us in the 'require backend' mode for the
+                # logic check and then this whole code branch is irrelevant. However,
+                # static variables can be used as input as well for some dynamic
+                # evaluation, most notably the future/past modes of date/datetime
+                # fields, so we pass a minimal view of the submission data.
+                data=variables_state.get_data(
+                    include_static_variables=True,
+                    include_unsaved=False,
+                ),
             )
             default_configuration = wrapper.configuration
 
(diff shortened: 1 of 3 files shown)


src/openforms/formio/service.py
```diff
@@ -59,6 +59,10 @@
     "get_component_empty_value",
     "get_readable_path_from_configuration_path",
     "holds_submission_data",
+    # TODO: refactor so this is all contained in itself, possibly merge with more
+    # dynamic config. The msgspec experiment branch might introduce this soon-ish.
```
viktorvanwijk: 👀

src/openforms/submissions/api/serializers.py
```diff
@@ -401,6 +403,15 @@ def to_representation(self, instance):
...
             )
+            # Add to each component the custom errors in the current locale
+            get_translated_custom_error_messages(
+                wrapper, instance.submission.language_code
+            )
+            localize_components(
+                wrapper,
+                instance.submission.language_code,
+                enabled=instance.submission.form.translation_enabled,
+            )
```
viktorvanwijk: I see these are from the `get_dynamic_configuration` hook, which made me think: it also rewrites the formio configuration and injects the prefill variables...

I verified manually that min/max date validation with the mode `future` or `past` is broken now too, because the default configuration is not updated.

The solution is probably to call `get_dynamic_configuration` here instead, no?

sergei-maertens: yeah probably :thinking: I thought there was a deliberate reason why that functionality was not called, so opted to do the bare minimum, but apparently that's not enough :D

viktorvanwijk: No I think I just forgot to add it :sweat_smile:

sergei-maertens: The test suite seems to be okay with it 🥳

src/openforms/submissions/api/serializers.py
```diff
@@ -401,6 +402,16 @@ def to_representation(self, instance):
...
             )
+            wrapper = get_dynamic_configuration(
+                wrapper,
+                submission=instance.submission,
+                # dynamic formio component configuration may depend on filled out data
+                # XXX: I think we can use `data=None` here because we currently don't
+                # have cases where the step-data itself modifies the component
+                # definition afaik, and *if* that happens, it should kick the logic
+                # evaluation into backend-required mode anyway
+                data=None,
```
viktorvanwijk: I agree that configuration being dependent on step data should require backend logic evaluation, but I think passing `None` here still breaks the future / past modes. It seems to convert them into a relativeToVariable mode, and using the `now` static variable for the calculations 🤔

sergei-maertens: aaargh I got too distracted and forgot to add a test for that case :weary:

sergei-maertens:
@viktorvanwijk I've added a test for the past/future modes

For the data, I'm deliberately not using `SubmissionStep._data` because that doesn't include the static variables and that's the bit we're most interested in.

viktorvanwijk (approved):

sergei-maertens:
Backport to 3.5.x done in 90a3f51191603e2959e19604804832ed24513957

PR #6325, opened 2026-05-27
[#6294] Prefer branch number from auth for StUF-ZDS and ZGW APIs

### LINKED ISSUE #6294: StUF-ZDS and ZGW APIs: prefer branch number from auth data rather than form variable

vaszig:
Closes #6294

**Changes**

- Until now the branch number was taken from the form variable (higher prio) and then from the auth context. Now we revert this and we take the value from the auth context as a priority.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/registrations/contrib/stuf_zds/plugin.py
```diff
@@ -433,10 +432,7 @@ def process_vestiging(
             None,
         )
 
-        # prioritize the zaak_mapping as currently is
-        if not vestigings_nummer_key or (
-            zaak_data.get(vestigings_nummer_key) or not submission.is_authenticated
-        ):
+        if not vestigings_nummer_key or not submission.is_authenticated:
```
sergei-maertens: I'm not sure if this approach is the correct one, but it also seems a bit more complicated than I anticipated.

I think what we want is:

* if the submission is authenticated and we have a branch number, it needs to be assigned to `zaak_data["initiator.vestigingsNummer"]`
* if we don't have it, *then* look for `vestigings_nummer_key` and assign it to/keep it in the zaak data

Looking at the code, we can't really modify the `"initiator.vestigingsNummer": RegistrationAttribute.initiator_vestigingsnummer,` line as that would break the registration mapping that people did (initially I had hoped something like `FieldConf(...)` could work to point to the submission context, but that doesn't support the default/fallback.

However, the code updates I think that are necessary are:

1. Keep L471 `zaak_data = apply_data_mapping(`, which processes the mapping.
    1. If a KVK number was provided in the form field, it will be assigned to `zaak_data["initator.vestigingsNummer"]` by this apply mapping
    2. If nothing was provided, it will be absent or empty
2. Move/clean up `process_vestiging` to *overwrite* `zaak_data["initator.vestigingsNummer"]` _if_ a branch number is available in the auth context, do nothing otherwise.

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -359,21 +359,13 @@ def register_submission(
...
-        )
-        if not vestigingsnummer and submission.is_authenticated:
+        if submission.is_authenticated:
             auth_context = submission.auth_info.to_auth_context_data()
             vestigingsnummer = get_branch_number(auth_context)
             betrokkene_identificatie["vestigingsNummer"] = vestigingsnummer
+        else:
+            vestigingsnummer = betrokkene_identificatie.get("vestigingsNummer")
 
         if kvk and vestigingsnummer:
```
sergei-maertens:
```suggestion
        if kvk and betrokkene_identificatie.get("vestigingsNummer"):
```
and then you can remove the `else` branch I think

sergei-maertens (approved):

PR #6329, opened 2026-05-28
Fix missing translated validation errors for `addressNL` component

### LINKED ISSUE #256: AddressNL: interaction bugs noticed during 3.4.0 testing
See form export that I used for testing: [addressnl-nieuwe-renderer.zip](https://github.com/user-attachments/files/24215735/addressnl-nieuwe-renderer.zip)

* When a postcode custom pattern is specified with custom error message, the custom error message is not displayed during client side validation
* When a custom city pattern is specified, this may not ever run during client side activation - it can't be properly tested/verified because the input field is `disabled`
* When entering a postcode and house number (`1111 ZZ`, `99999`) that doesn't resolve to any address, the city and street are required and invalid, blocking the step submit button. When the inputs are corrected, the step submit button is still disabled where you'd expect it to be enabled again. I suspect this has to do with the field not being valid and the disabled nature making it impossible to retrigger validation.

sergei-maertens:
Closes open-formulieren/formio-renderer#256 (partially)

**Changes**

The `translatedErrors` for the nested component configuration was not properly processed for the `addressNL` component, while the frontend code looks at the resulting `errors` property.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/dynamic_config/__init__.py
+++ b/src/openforms/formio/dynamic_config/__init__.py
@@ -56,13 +56,46 @@ def get_translated_custom_error_messages(
     config_wrapper: FormioConfigurationWrapper, language_code: str
 ) -> FormioConfigurationWrapper:
     for component in config_wrapper:
-        if (
-            not (custom_error_messages := component.get("translatedErrors"))
-            or "errors" in component
-        ):
-            continue
+        # generic mechanism
+        match component:
+            # if there are already errors in the component definition, do not overwrite
+            case {"errors": dict()}:
+                pass
 
-        component["errors"] = custom_error_messages[language_code]
+            # grab the language-specific error messages if they're available and store
+            # them in the resolved errors property
+            case {"translatedErrors": dict(errors_by_language_code)} if (
+                custom_error_messages := errors_by_language_code.get(language_code)
+            ):
+                component["errors"] = custom_error_messages
+
+        # TODO: the msgspec branch is probably the right opportunity to rectify the
+        # component-specific processing. For this hook, there's no matching registry/
+        # plugin hook.
+
+        match component:
+            case {
+                "type": "addressNL",
+                "openForms": {"components": dict(addressnl_components)},
+            } if addressnl_components:
+                for key in ("postcode", "city"):
+                    if not (component_config := addressnl_components.get(key)):
+                        continue
+                    if component_config.get("errors"):
+                        continue
+                    if not (
+                        errors_by_language_code := component_config.get(
+                            "translatedErrors"
+                        )
+                    ):
+                        continue
+                    if not (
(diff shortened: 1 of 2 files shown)


src/openforms/submissions/api/serializers.py
```diff
@@ -401,6 +402,16 @@ def to_representation(self, instance):
...
             )
+            wrapper = get_dynamic_configuration(
+                wrapper,
+                submission=instance.submission,
+                # dynamic formio component configuration may depend on filled out data
+                # XXX: I think we can use `data=None` here because we currently don't
+                # have cases where the step-data itself modifies the component
+                # definition afaik, and *if* that happens, it should kick the logic
+                # evaluation into backend-required mode anyway
+                data=None,
```
vaszig: Is this the same with https://github.com/open-formulieren/open-forms/pull/6323/changes/BASE..77c6212df2203711120c00cf631e92299af6097c#r3318794871 ?

src/openforms/formio/dynamic_config/__init__.py
```diff
@@ -56,13 +56,46 @@ def get_translated_custom_error_messages(
...
+            case {"translatedErrors": dict(errors_by_language_code)} if (
+                custom_error_messages := errors_by_language_code.get(language_code)
+            ):
+                component["errors"] = custom_error_messages
+
+        # TODO: the msgspec branch is probably the right opportunity to rectify the
+        # component-specific processing. For this hook, there's no matching registry/
+        # plugin hook.
+
+        match component:
```
vaszig, line 76: Is the second match only used to mention the different context (instead of using one for all the cases)?

vaszig (commented):
I re-triggered the CI but this seems to fail again. Can you have a look at it?

sergei-maertens:
> I re-triggered the CI but this seems to fail again. Can you have a look at it?

Yeah it was on my todo list!

src/openforms/formio/dynamic_config/__init__.py
sergei-maertens: yes indeed - in a `match/case` you can only hit one branch so I'd either have to duplicate the generic behaviour, or split them in separate matches and it's obvious what I picked :D

src/openforms/submissions/api/serializers.py
sergei-maertens: yeah this branch is based on the other one, I'll put this back in draft!

vaszig (approved):

sergei-maertens:
Backported to 3.5.x in 98e32d8a9fff6d07497580db65f073527a952d43 - 3.4 is out of scope for addressNL.

PR #6337, opened 2026-06-01
Support `adres` registration attribute (for `addressNL`)

### LINKED ISSUE #6336: Support mapping entire address as registration attribute (StUF-ZDS, ZGW APIs)
Rather than mapping individual fields, we can map the whole addressNL component to the parts of address-related fields for the initiator.

- [x] Add registration attributes documentation in the manual

sergei-maertens:
Closes #6336

**Changes**

* Updated the ZGW APIs and StUF-ZDS backends to support mapping an `addressNL` to the initiator adres
* Extended the available registration attributes so that it can be selected in the addressNL registration tab

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
+++ b/src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
@@ -29,6 +29,7 @@ from hypothesis.extra.django import TestCase as HypothesisTestCase
 from lxml import etree
 from privates.test import temp_private_root
 from requests import ConnectTimeout
+from unittest_parametrize import ParametrizedTestCase, parametrize
 from zgw_consumers.constants import AuthTypes
 from zgw_consumers.test.factories import ServiceFactory
 
@@ -3384,7 +3385,7 @@ class XMLSanitizerVCRTests(OFVCRMixin, StUFAssertionsMixin, HypothesisTestCase):
 
 
 @temp_private_root()
-class StufZDSPluginVCRTests(OFVCRMixin, StUFZDSTestBase):
+class StufZDSPluginVCRTests(OFVCRMixin, ParametrizedTestCase, StUFZDSTestBase):
     VCR_TEST_FILES = TESTS_DIR / "files"
 
     @classmethod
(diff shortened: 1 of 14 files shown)


src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
```diff
@@ -3187,3 +3188,213 @@ def test_create_zaak_with_contactpersoon_rol_attrs(self):
...
+                "aoaIdentificatie": "OFWORKAROUND",
+                "aoaPostcode": "1043 GR",
+                "gorOpenbareRuimteNaam": "Kingsfordweg",
+                "inpLocatiebeschrijving": "",
+                "wplWoonplaatsNaam": "Amsterdam",
+            },
+        )
+
+    # breaks because we can't put a KVK number in an RSIN field
+    @expectedFailure
```
annashamray, line 3266: rol resource in Open Zaak has `betrokkeneIdentificatie.kvkNummer` field which was added for "machtiging" support

src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
```diff
@@ -3612,6 +3613,159 @@ def test_extra_mapped_extra_elementen_are_picked_up(self):
             },
         )
 
+    def test_addressnl_registration_attribute_initiator_address(self):
```
annashamray, line 3616: While I was testing Stuf-zds locally I noticed that the address is sent only if the user is authenticated. If the user is not authenticated no address is added inside `ZKN:natuurlijkPersoon` 
Is it an expected behaviour?

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
```diff
@@ -3187,3 +3188,213 @@ def test_create_zaak_with_contactpersoon_rol_attrs(self):
...
+                    "type": "addressNL",
+                    "label": "Adres",
+                    "deriveAddress": False,
+                    "registration": {
+                        "attribute": RegistrationAttribute.initiator_adres
+                    },
+                },
+            ],
+            kvk="12345678",
+            submitted_data={
```
annashamray, line 3289: Could you please add the test with the same submission data but with bsn (so it won't be expected to fail)
When I test ZGW APIs locally and I don't check "derive street" checkbox in the AddressNL component, then I receive 400 from Open Zaak:
```{'betrokkene_identificatie': {
  'verblijfsadres': {
    'wpl_woonplaats_naam': [ErrorDetail(string='Dit veld mag niet leeg zijn.', code='blank')], 
    'gor_openbare_ruimte_naam': [ErrorDetail(string='Dit veld mag niet leeg zijn.', code='blank')
    ]
  }
}}```

annashamray (changes requested):
Could you please add some documentation about this field? For example, that we expect only AddressNL fields to have mapping to the `("Initiator > Adres")` value. It's possible for now to assign this value to any field, so it's better have a documentation to point to.

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
sergei-maertens: This does not exist in the Zaken API 1.6.0 and in Open Zaak it's marked as experimental feature. Open Forms cannot rely on this, so it's still broken.

It's an old, known issue and you'll see there's another test already with the same expected failure decorator :grimacing:

src/openforms/registrations/contrib/stuf_zds/tests/test_backend.py
sergei-maertens: Correct, the `verblijfsadres` element is a child element of `<ZKN:natuurlijkPersoon StUF:entiteittype="NPS" StUF:verwerkingssoort="T">`, which is only included if there's a BSN or family name of the initiator.

The same thing happens if you use other fields types (textfield, postcode etc.) and map them to the individual registration attributes.

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
sergei-maertens: They're required fields in the API spec, so omitting them if the value is empty-ish is not an option.

I think this is a documentation thing to make sure that if you use ZGW APIs, the city/street need to be included through the address derivation feature. For StUF-ZDS, these fields are not required and it also works without derive address.

I'll handle documentation in a separate PR as discussed on Slack!

sergei-maertens: Did a patch up to ensure that `bezoekadres` is used rather than `verblijfadres` for NNP, as that was a mistake in the implementation. It still cannot be properly tested/supported because of the KVK number/RSIN problem in the API spec though :(

src/openforms/registrations/contrib/zgw_apis/tests/test_backend.py
annashamray: Ok! I didn't know it's broken

annashamray (approved):

PR #6338, opened 2026-06-01
Fix component empty value for complex components

### LINKED ISSUE #6125: `get_component_empty_value` returns wrong value for `addressNL`

latest, 3.4.x and earlier

### Customer reference

Leiden 51

Discovered while writing tests for the frontend equivalent - when you call `openforms.formio.service.get_component_empty_value` on an `addressNL` component, it returns `''` (empty string). I'd rather expect `null` / `None`, because the data structure is a complex nested object.

This may also be the reason that the renderer sometimes crashes when using addressNL combined with logic.

1. Add addressNL component to form
2. Determine the component empty value for the the component
3. Observe that it returns a non-sensical data type.

Returns `null`.

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #6125

**Changes**

* Fixed addressNL empty value
* Also address map, customer profile
* Refactored implementation to use registry (component data type, sub-type and empty value)

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/registry.py
+++ b/src/openforms/formio/registry.py
@@ -34,6 +34,7 @@ from rest_framework.request import Request
 from openforms.plugins.plugin import AbstractBasePlugin
 from openforms.plugins.registry import BaseRegistry
 from openforms.typing import JSONObject, JSONValue, VariableValue
+from openforms.variables.constants import FormVariableDataTypes
 
 from .datastructures import FormioConfigurationWrapper, FormioData
 from .typing import Component
@@ -95,6 +96,18 @@ class BasePlugin(Generic[ComponentT], AbstractBasePlugin, abc.ABC):  # noqa: UP0
     holds_submission_data: ClassVar[bool] = True
     """
     Flag to indicate whether data can be submitted for this component.
+    """
+    data_type: ClassVar[FormVariableDataTypes] = NotImplemented
+    """
+    The intrinsic data type for the component, for the non-multiple case.
+
+    Each component type produces values of a certain data type. :attr:`empty_value` is
+    expected to be an instance of that data type. Note that components that support the
+    ``multiple`` property typically return a list of the intrinsic data type.
+    """
+    data_subtype: ClassVar[FormVariableDataTypes | None] = None
+    """
+    The intrinsic item data type for components that have a collection as data type.
     """
     empty_value: ClassVar[JSONValue] = NotImplemented
     """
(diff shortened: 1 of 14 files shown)


src/openforms/formio/registry.py
```diff
@@ -169,6 +193,14 @@ def test_conditional(
         """
         return None
 
+    def get_data_type(self, component: ComponentT) -> FormVariableDataTypes:
+        return self.data_type
+
+    def get_empty_value(self, component: ComponentT):
```
viktorvanwijk:
```suggestion
    def get_empty_value(self, component: ComponentT) -> JSONValue:
```

src/openforms/formio/registry.py
```diff
@@ -369,6 +401,49 @@ def apply_pre_registration_hook(
...
+            return subtype
+
+        if not component.get("multiple"):
+            return ""
+
+        # get the intrinsic data type
+        _component: Component = {**component, "multiple": False}
+        return self.get_component_data_type(_component)
+
+    def get_empty_value(self, component: Component):
```
viktorvanwijk:
```suggestion
    def get_empty_value(self, component: Component) -> JSONValue:
```

viktorvanwijk (approved):
Nice!

Edit: I guess this PR only 'partly' closes #6125 because the frontend still needs to be changed as well?

sergei-maertens:
> Nice!
> 
> Edit: I guess this PR only 'partly' closes #6125 because the frontend still needs to be changed as well?

No - the initial values (that are used for the clear-on-hide empty values [here](https://github.com/open-formulieren/open-forms-sdk/blob/cd59b4a44206e448b366b9c8a94d28d849c04067/src/components/FormStep/logic.ts#L106)) already match, see the last paragraph in ee41b11b5a27be39e9406783419a7642e03058ad

So I believe this completes the ticket :-)

viktorvanwijk:
> No - the initial values (that are used for the clear-on-hide empty values [here](https://github.com/open-formulieren/open-forms-sdk/blob/cd59b4a44206e448b366b9c8a94d28d849c04067/src/components/FormStep/logic.ts#L106)) already match, see the last paragraph in [ee41b11](https://github.com/open-formulieren/open-forms/commit/ee41b11b5a27be39e9406783419a7642e03058ad)
> 
> So I believe this completes the ticket :-)

HA, the one commit message that I didn't read :grin:

Issue #6045, opened 2026-03-04
Infinite loop with new renderer when clearing children of editgrids with backend logic


viktorvanwijk:
#6040 described a situation where fields inside an editgrid were unexpectedly cleared if there is a backend logic rule action on the visibility of the editgrid, even if the rule wasn't triggered.

While testing with the new renderer, I noticed that this specific setup creates an infinite loop of check logic calls. 
Steps to reproduce:

1. Create form with checkbox, and editgrid (clearOnHide _disabled_) with a textfield inside it (clearOnHide _enabled_)
2. Create logic rule that hides the editgrid when the checkbox is checked.
3. Fill out the textfield inside an editgrid item
4. Check the checkbox to hide the editgrid
5. Observe that the backend returns data: `{"editgrid": [{"textField": ""}]}`, but the infinite subsequent calls have `{"editgrid": [{}]}` in the request body.

Seems like the renderer doesn't know the value should be an empty string when the component becomes visible again?

Note that it doesn't happen when the editgrid as a whole has clearOnHide enabled

Issue #5980, opened 2026-02-19
Components with "negative" logic included in submission


LaurensBurger:
### Product versie / Product version

3.4.1

### Customer reference

Utr 448

Maykin test form 975

### Omschrijf het probleem / Describe the bug

Components that have the following logic:
```json
  "conditional": {
    "show": false,
    "when": "radio1",
    "eq": ""
  },
```
are included in the submission, causing submissions to get stuck. 
The fields are never shown in the steps, as those conditions are never met while filling in the form. because `radio1` is hidden.

<img width="1165" height="494" alt="Image" src="https://github.com/user-attachments/assets/e0bc0175-292e-493d-9293-0ee06c0e8fd6" />

<img width="1120" height="515" alt="Image" src="https://github.com/user-attachments/assets/294e34c5-f04b-42e9-a761-99e57a70541c" />

When removing the "clear on hide" from "radio1" - this behaviour is no longer observable. 

It seems the "conditional" is somehow triggered at the overview step, perhaps because `radio1` is not equal to "" at this point?

### Screen resolution

None

### Device

None

### OS

None

Issue #4324, opened 2024-05-22
Hidden component unexpectedly set in submission data if `clearOnHide` is set and the default value differs from the empty value


Viicos:
### Product versie / Product version

latest

### Omschrijf het probleem / Describe the bug

This part of the code in `evaluate_form_logic` seems off:

https://github.com/open-formulieren/open-forms/blob/4734c038dc0c4e668da677ade5671e9867405b27/src/openforms/submissions/form_logic.py#L149-L159

### Stappen om te reproduceren / Steps to reproduce

1. Create a checkbox component (with key `myCheckbox`), hidden by default, clear on hide set to `True`, default value of `True` (checked).
2. Start a submission, and observe submission data: `{"myCheckbox": False}`

### Screen resolution

None

### Device

None

### OS

None

Viicos:
This also breaks submission data if:
- You define a fieldset with key `fieldset`, with `clearOnHide` set to `True`.
- You define a textfield inside with key `fieldset.textfield`.
- You set a simple conditional on `fieldset`, so that it becomes hidden depending on the value of some other variable.

If the condition of the simple conditional matches, the submission data will look like:

```json
{"fieldset": ""}
```

joeribekker:
Refinement: This is deemed to dangerous to touch at this time without any actual complaints (Viicos stumbled upon it by accident). The problem might go away with the current refactor in #3154. Also, once we start pushing logic to the frontend this hopefully is a thing of the past.

viktorvanwijk:
Our renderer correctly always removes hidden fields with clearOnHide enabled (this is tested extensively in the renderer itself), and hidden fields will be removed from the context completely in the backend too as of #6341 (see for example `CheckLogicEndpointTests.test_clear_on_hide_behaviour` and `CheckLogicSubmissionTest.test_clear_on_hide_behaviour_applied_during_evaluation`). Therefore, #6341 closes this ticket :)

PR #6341, opened 2026-06-02
Properly clear values during logic evaluation

### LINKED ISSUE #6264: Clean up `check_submission_logic`
Noticed this during the removal of the new logic evaluation enabled feature flag (#6261).

The argument `current_step` was added as a bugfix for #4579, to ensure the correct step is evaluated when serializing the result for a logic check (combination of a trigger from step and disable-next action). Given that trigger from step doesn't exist anymore, logic rules are assigned per step, and a disable-next action is step specific, it seems to me that it can be removed. Furthermore, I think we do not have to execute `check_submission_logic` at all when serializing for a logic check result, as we have just executed logic for that step anyway using `evaluate_form_logic`. There already is a `in_form_logic_evaluation` key present in the context of the `SubmissionStateLogicSerializer` which can be used to detect this.

Also it contains a `reset_configuration_wrapper` argument with the following comment: 

>  XXX: not sure why this is necessary - removing it entirely doesn't seem to break
>  any tests? Check with Viktor.

- [ ] Should check again whether this is still necessary after all the logic engine rework

### LINKED ISSUE #6275: Implement "undefined" clearOnHide behavior in the backend
Possible solutions:
* Add flag `is_undefined` flag to `SubmissionValueVariable` model, with a constraint that the value can only be `None` when `is_undefined` is true? → My first feeling is that we can actually avoid dragging the variables into this, and keep it confined within the logic evaluation context.
* Add custom `undefined` value that we can set when the value gets cleared? → This will also break being able to specify a default value in a JSON logic expression, as the keys will always be present with a (undefined) value
* Only pop the value from the context during logic evaluation and leave the variables untouched? → Makes the most sense for me, and is probably the easiest. Though it might create a difference between values of hidden components during logic evaluation and during registration. During evaluation, their values will be ‘undefined’, while they will be the empty/default value during registration. Is that OK?

Don't forget to add documentation on how it worked previously vs how it works now.

Possible rare edge case:
If a component was hidden with logic, and then had a variable action applied to it. Do we set the value, or leave “undefined”?

viktorvanwijk:
Closes #6275, #4324
Partly closes #6264

**Changes**

* Values are now removed from the context during logic evaluation when the corresponding component is hidden
* Unsaved variables of future steps are no longer available during logic evaluation (updated the relevant check script as well)
* Submission logic is not executed when evaluating step logic (see also #6264)
* Fixed step data not being included when getting the submission step
* Removed option to ignore hidden property in `is_hidden`
* Removed `is_visible_in_frontend`
* Reverted bug fix for #5980
* Reverted bug fix for #6045
* Updated tests

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/submissions/logic/actions.py
```diff
@@ -330,7 +327,7 @@ def apply(
         submission_step_to_modify.reset()
         if submission_step_to_modify == step:
             step.is_applicable = False
-            assert step.unsaved_data is None
+            # assert step.unsaved_data is None
```
vaszig: Can be removed I guess.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -522,6 +529,7 @@ class SubmissionValueVariable(models.Model):
     objects = SubmissionValueVariableManager()
 
     form_variable: FormVariable | None = None
+    _is_undefined: bool | None = None
```
vaszig, line 533: Why do we need the `None` option?Isn't `False` enough/clear?

src/openforms/submissions/tests/form_logic/test_conditional_logic.py
```diff
@@ -33,32 +33,24 @@ def test_existing_data_is_not_cleared(self):
                 "textfieldHidden": "I am submitted data",
             },
         )
-        step = submission.submissionstep_set.first()
+        step = submission.submissionstep_set.get()
```
vaszig, line 36: Why did we change these (below too)?Just to show that we do expect exactly one or something else?

src/openforms/formio/tests/test_utils.py
```diff
@@ -1,169 +1,12 @@
...
-    {
-        "type": "file",
-        "key": "file1",
-        "label": "File",
-    },
-]
-
-
-class FormioUtilsTest(SimpleTestCase):
-    def test_is_visible_in_frontend_with_selectboxes(self):
```
vaszig, line 39: Why did we remove these?

vaszig: Ok, they are still there, I now saw them below.

src/openforms/formio/visibility.py
```diff
@@ -47,24 +54,30 @@ def is_hidden(
...
 
     show, trigger_component_key, compare_value = conditional
 
-    trigger_component_value = data.get(trigger_component_key, None)
-    trigger_component = configuration[trigger_component_key]
+    # Be resilient on the component key lookup
+    if trigger_component_key in configuration:
+        trigger_component = configuration[trigger_component_key]
+    else:
+        trigger_component: Component = {
```
vaszig, line 70: Could you please elaborate a bit on this?

src/openforms/forms/models/form.py
```diff
@@ -805,15 +805,18 @@ def apply_logic_analysis(self) -> None:
...
         ).remote_field.through
 
+        updated_rules_and_steps = analyze_rules(form=self)
+
         # drop the old relations and create the expected ones.
+        FormLogic.objects.filter(form=self).delete()
+        FormLogicFormSteps.objects.filter(formlogic__form=self).delete()
 
-        updated_rules_and_steps = analyze_rules(form=self)
+        FormLogic.objects.bulk_create([rule for rule, _ in updated_rules_and_steps])
```
vaszig, line 814: I am wondering if the rules were not reordered how the tests were passing until this PR. Because I guess we do have this kind of tests.

src/openforms/submissions/tests/factories.py
```diff
@@ -205,6 +206,16 @@ def auth_info(obj, create, extracted, **kwargs):
         kwargs["submission"] = obj
         AuthInfoFactory.create(**kwargs)
 
+    @factory.post_generation
+    def initialize_user_defined_variables(obj, create, extracted, **kwargs):
+        # This happens upon submission creation in the SubmissionViewSet
+        if create:
+            initialise_user_defined_variables(obj)
+            # Need to reset these cached properties to makes sure we are not dealing
```
vaszig:
```suggestion
            # Need to reset these cached properties to make sure we are not dealing
```

src/openforms/submissions/tests/test_get_submission_step.py
```diff
@@ -190,7 +190,7 @@ def mutate_config_dynamically(component, submission, data):
                     },
                 ]
             },
-            "data": {},
+            "data": {"someField": "", "otherField": {}},
```
vaszig, line 193: This has to do with your comment that we pass the empty data but our renderer removes it anyway, correct?

src/openforms/submissions/form_logic.py
```diff
@@ -23,6 +23,9 @@
 tracer = trace.get_tracer("openforms.submissions.form_logic")
 
 
+empty = object()
```
vaszig, line 26: I saw that we use that in https://github.com/open-formulieren/open-forms/blob/main/src/openforms/submissions/models/submission_value_variable.py#L14 too, maybe we want to use the same?

vaszig (commented):
This would be nice to be in separate PRs, but I guess it was not possible to do so. It is really difficult for me to review it properly (a lot of major changes take place/bugs are fixed/reverted). Sergei has to take a look too for sure.

Spent quite some time on this but I don't have something to add/remark (tests are also passing) except for some questions.

src/openforms/submissions/logic/actions.py
viktorvanwijk: Oops, no it can't actually. Accidentally left this in when testing for #6345 :sweat_smile:

src/openforms/submissions/tests/form_logic/test_conditional_logic.py
viktorvanwijk: Yes :) Should be exactly one

src/openforms/formio/visibility.py
viktorvanwijk: It was originally added to the `is_visible_in_frontend` routine with a bug fix for #6181. Basically to prevent a hard crash when a component cannot be found, in case of incorrect form configuration. But also because `FormioConfigurationWrapper.is_hidden` (previously `.is_visible_in_frontend`) does not understand editgrids, so it has to default to an empty string comparison.

The `registry.apply_visiblity` does know how to deal with editgrids, so there is potential for further clean up, but that's too much work for this PR

src/openforms/forms/models/form.py
viktorvanwijk: Hmm, I checked again and currently no test fails when I revert this change, but I remember getting some weird errors because of this. Will see if I can add a separate one for this

src/openforms/submissions/tests/test_get_submission_step.py
viktorvanwijk: This was added because of a bandaid fix for interaction with the SDK (see fd968b7a). If we do not pass the step values here, the first invoked check-logic call is performed without any data. It wasn't a problem before, because we would get all the data from the state

src/openforms/submissions/form_logic.py
viktorvanwijk: Yeah I saw that too, though it didn't seem intuitive to me to make this file dependent on the django package just for an empty value

src/openforms/submissions/models/submission_value_variable.py
viktorvanwijk: Yeah defaulting to `False` is OK for now perhaps, though it might give the wrong impression as it wasn't processed/updated (yet) :thinking:

src/openforms/forms/models/form.py
viktorvanwijk: I think I discovered it while fixing `test_rules_flip_from_hidden_to_visible_state_should_not_clear_value` to deal with the undefined value, where the logic rule order should have never been possible in the first place.

Note that instead of calling `form.apply_logic_analysis()` there, I added:
```python    
# Deliberately do not use `form.apply_logic_analysis()` here, because it will
# re-order the third logic rule to execute before the second one, which makes it
# impossible to check the value after clearing.
form_step.logic_rules.set([rule_1, rule_2, rule_3])
```

viktorvanwijk: I added `test_logic_rules_are_properly_reordered_when_applying_rule_analysis` in a commit before the bugfix

src/openforms/submissions/form_logic.py
sergei-maertens: FYI, Python 3.15 will have a nice proper datatype for this that can also be used in type hints: https://docs.python.org/3.15/library/functions.html#sentinel

src/openforms/submissions/form_logic.py
```diff
@@ -96,14 +99,29 @@ def evaluate_form_logic(
...
 
     # 4. Apply the (dirty) data to the variable state.
-    # We need to get the initial data for clear on hide first, as we don't want it to
-    # include the unsaved data.
-    data_for_hidden_state = _get_initial_data_for_clear_on_hide(step)
-    if unsaved_data is not None:
-        submission_variables_state.set_values(unsaved_data)
+    # This setup avoids having to update the state with unsaved_data. That only happens
+    # after logic is evaluated now, which ensures it will not contain user input for
+    # hidden fields. TODO-6275: unless it was submitted before???
```
sergei-maertens: Self-reminder - check if this is addressed in later commits.

src/openforms/submissions/form_logic.py
```diff
@@ -96,14 +99,29 @@ def evaluate_form_logic(
...
+        )
+
+    # This includes user-defined variables, static variables, and previously saved step
+    # variables
     data_for_evaluation = submission_variables_state.get_data(
-        include_unsaved=True, include_static_variables=True
+        include_unsaved=False, include_static_variables=True
     )
+    # Update the data for evaluation with the data from the current step
+    data_for_evaluation.update(step_data)
```
sergei-maertens: w/r to the TODO above if you have saved variables from before, they would be included. Thoughts:

* we could loop again over the `step_variables` and for keys that are not in the unsaved data, pop them from this `data_for_evaluation`
* however, that still allows users to make cURL requests with these keys included, making this an unreliable mechanism to drop them

I'm curious to see how this is addressed :) I kinda hope that the logic evaluation (top to bottom) results in the clearing of values and this "repairing itself" automatically.

src/openforms/submissions/tests/factories.py
```diff
@@ -282,6 +293,7 @@ def from_components(
 
         # When the submission was initially created, the method calculate_price has already
         # loaded the submission_value_variables_state, but no submission variables existed at that point.
+        initialise_user_defined_variables(submission)
```
sergei-maertens: Isn't this already covered by the post-generation hook you define on L210, which I'd expect to be called through L242 `submission = cls.create(**kwargs)`?

src/openforms/submissions/tests/form_logic/test_submission_logic.py
```diff
@@ -1973,6 +2019,84 @@ def _build_visibility_action(key: str, make_hidden: bool):
...
+        self._add_submission_to_session(submission)
+        logic_check_endpoint = reverse(
+            "api:submission-steps-logic-check",
+            kwargs={
+                "submission_uuid": submission.uuid,
+                "step_uuid": form_step.uuid,
+            },
+        )
+
+        response = self.client.post(logic_check_endpoint, {"data": {"checkbox": True}})
```
sergei-maertens: Can we add a test that simulates spoofing data as well, by including a non-empty value for `textfield`? I'm curious to see what happens :-)

src/openforms/submissions/form_logic.py
```diff
@@ -101,22 +101,26 @@ def evaluate_form_logic(
     # 4. Apply the (dirty) data to the variable state.
     # This setup avoids having to update the state with unsaved_data. That only happens
     # after logic is evaluated now, which ensures it will not contain user input for
-    # hidden fields. TODO-6275: unless it was submitted before???
+    # hidden fields.
     step_variables = submission_variables_state.get_variables_in_submission_step(
         step, include_unsaved=True
     )
+    # This data is used to restore the value of a component when it goes from hidden ->
+    # visible. It includes
```
sergei-maertens: don't leave me hanging! what does it include? 😛

src/openforms/submissions/form_logic.py
```diff
@@ -273,9 +276,12 @@ def check_submission_logic(
     )
 
     mutation_operations: list[ActionOperation] = []
+    # Note: values should have already been resolved at this point, so we just pass
+    # `data_for_evaluation` as data for visible state.
     for operation in iter_evaluate_rules(
         rules,
         data_for_evaluation,
+        data_for_evaluation,
```
sergei-maertens: can you use named kwargs in this function call instead of positional arguments? It's really hard to tell what's being passed in a github diff when the same variable is passed for two different arguments.

docs/installation/upgrade-400.rst
```diff
@@ -261,8 +261,50 @@ For more information about the new logic evaluation, please refer to the
 
 Clearing of values
 ------------------
+Before Open Forms 4.0 during a logic check, a component which becomes hidden with "clear on hide" enabled,
```
sergei-maertens, line 265: can you add a blank line between heading and content? 😇

docs/installation/upgrade-400.rst
```diff
@@ -261,8 +261,50 @@ For more information about the new logic evaluation, please refer to the
...
+     "==": [
+       {
+         "var": ["textfield", ""]
+       },
+       ""
+     ]
+   }
+
+Already in 3.5.3, we have included a script that reports any forms which might be at risk of behaving
+differently. This script is still available in 4.0, and can be executed with using:
```
sergei-maertens: "with using" at the end reads a bit weird

docs/installation/upgrade-400.rst
```diff
@@ -261,8 +261,50 @@ For more information about the new logic evaluation, please refer to the
...
+
+.. code-block:: bash
+
+    # in the container via ``docker exec`` or ``kubectl exec``:
+    python /app/bin/report_logic_with_deprecated_clear_on_hide_behavior.py
+
+Values of components of future steps
+------------------------------------
+When evaluating logic on a step, values of components from all (unsubmitted) future steps are no
```
sergei-maertens, line 305: same blank line request here :)

src/openforms/submissions/form_logic.py
```diff
@@ -98,34 +98,35 @@ def evaluate_form_logic(
...
-    step_variables = submission_variables_state.get_variables_in_submission_step(
-        step, include_unsaved=True
-    )
+    # 4. Get relevant data structures.
+    if unsaved_data:
+        submission_variables_state.set_values(unsaved_data)
+
     # This data is used to restore the value of a component when it goes from hidden ->
-    # visible. It includes
+    # visible. It should include all fields to make sure we have a value for them.
```
sergei-maertens, line 106: Ah, there is my closure 😉

src/openforms/submissions/form_logic.py
```diff
@@ -98,34 +98,35 @@ def evaluate_form_logic(
...
+        step, include_unsaved=True
+    )
+    if unsaved_data:
+        # If unsaved data was passed, we rely on it accurately reflecting the visible
+        # state of the components, meaning hidden components with clearOnHide enabled
+        # will not be present. Though note that we do execute conditional logic again as
+        # an additional validation step.
+        for key in step_variables:
+            if key not in unsaved_data:
+                data_for_evaluation.pop(key)
```
sergei-maertens, line 129: alright, this does line up with my earlier thought process about popping missing keys

The comment seems to confirm that evil-submitted additional keys will still end up being cleared by the conditional logic being applied, am I understanding that correctly?

src/openforms/formio/rendering/default.py
```diff
@@ -46,10 +47,10 @@ def is_visible(self) -> bool:
             return True
 
         # We only pass the step data, since frontend logic only has access to the current step data.
-        if not is_visible_in_frontend(
+        if is_hidden(
             self.component,
             self.step_data,
-            configuration_wrapper=self.renderer.submission.total_configuration_wrapper,
+            self.renderer.submission.total_configuration_wrapper,
```
sergei-maertens: I do prefer the use of keyword arguments - they make reviewing in the web UI a better experience.

src/openforms/formio/tests/test_visibility.py
```diff
@@ -0,0 +1,159 @@
...
+        "values": [
+            {"value": "a", "label": "A"},
+            {"value": "b", "label": "B"},
+        ],
+    },
+    {
+        "type": "file",
+        "key": "file1",
+        "label": "File",
+    },
```
sergei-maertens, line 29: can you add:

```py
"file": {"type": []},
"filePattern": "",
```
for the experimental msgspec branch :D

src/openforms/formio/visibility.py
sergei-maertens: I'll also run into this again I presume with the msgspec branch, because the `type: unknown` will cause issues there. Fine to postpone - we should probably also explore how this is handled in the frontend - I suspect that ends up in the nullish value which falls back to empty string.

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -766,40 +752,24 @@ def _value_to_python(
...
             return value
 
-        if data_type == FormVariableDataTypes.editgrid:
-            assert isinstance(value, dict)
+        if value and data_type == FormVariableDataTypes.editgrid:
             value = FormioData(value)
-            for child_component in iter_components(
-                configuration, recurse_into_editgrid=False
-            ):
+            for child_component in iter_components(configuration):
```
sergei-maertens: can you make the `recurse_into_editgrid` kwarg explicit?

sergei-maertens (commented):
Can you explain what happens now in these situations that were mentioned in the commits?

1. There is saved from step data, and the user goes back to edit the step data, changing some field that's used as trigger which triggers a cascade of clear on hide actions - what will happen with the data that exists in the database in the context of being used as input for subsequent logic rules?
2. What happens when the frontend hides some fields and clears the values, but the user forces the user input to still be included for the hidden components because they're making Postman/Bruno/cURL calls? Will this indeed properly reset as I suspect?

I think I didn't see any explicit tests for these interactions, but feel free to point them out if I missed them. If they're not present, it might be a good idea to add these kind of integration tests.

It's also not clear anymore now to me what the data diff will return as the result of the logic check endpoint when hidden/clear-on-hide components are involved. I saw a bunch of assertions on the `submission.variables_state.get_data()` which includes `None` for variables that are marked as `undefined`. I suspect that these are not returned in the data diff - is that right?

docs/installation/upgrade-400.rst
viktorvanwijk: I was doubting between "with" and "using", so I did:
Test

src/openforms/submissions/form_logic.py
viktorvanwijk: That is the idea indeed :)

viktorvanwijk:
> 1. There is saved from step data, and the user goes back to edit the step data, changing some field that's used as trigger which triggers a cascade of clear on hide actions - what will happen with the data that exists in the database in the context of being used as input for subsequent logic rules?

This should work exactly the same as when there was no saved step data, because we fetch all data from the state with `include_unsaved=True` at first anyway. Then, if we have `unsaved_data` available, the relevant keys are popped. Finally, our dependency graph enforces rules that affect the components' visibility to be executed before rules that use these components as inputs. EDIT: added `test_clear_on_hide_behaviour_with_saved_step_data`

> 2. What happens when the frontend hides some fields and clears the values, but the user forces the user input to still be included for the hidden components because they're making Postman/Bruno/cURL calls? Will this indeed properly reset as I suspect?

That should be the case indeed. Most of the tests in `ConditionalLogicTests` already kinda confirm this, because the passed data would not be possible with the component configuration. Will add some interactions tests with the endpoint itself as well. EDIT: added `test_hidden_field_by_frontend_added_to_post_data_anyway`

> It's also not clear anymore now to me what the data diff will return as the result of the logic check endpoint when hidden/clear-on-hide components are involved. I saw a bunch of assertions on the submission.variables_state.get_data() which includes None for variables that are marked as undefined. I suspect that these are not returned in the data diff - is that right?

That is correct :). This [commit](https://github.com/open-formulieren/open-forms/commit/a58c84625ac9b8b7076bc6ac2c74eabb0f5f6557) contains the updated tests with this new behavior. For example, `CheckLogicEndpointTests.test_clear_on_hide_behaviour` makes assertions on the returned data, which only includes the fields that became visible.

src/openforms/submissions/tests/factories.py
viktorvanwijk: It is indeed :)

src/openforms/submissions/tests/form_logic/test_submission_logic.py
viktorvanwijk: Hmm, writing such a test for this specific setup is tricky, because all combinations of checkbox values and presence of textfield are actually valid logic calls:

* Checkbox True, Textfield missing: checkbox was False before meaning textfield was initially hidden
* Checkbox True, Textfield available: checkbox was True before meaning textfield was already visible
* Checkbox False, Textfield missing: checkbox was False before meaning textfield was already hidden
* Checkbox False, Textfield available: checkbox was True before meaning textfield was initially visible

It seems tests regarding spoofing data only make sense in the context of conditional logic.

sergei-maertens: I suppose this doesn't make much sense for the check logic endpoint (as you say, they're valid combinations), but maybe what's more interesting is if you can spoof "final" situations by making the `PUT` call to save the data?

I'm not sure now if you submit `checkbox: False` and `textfield: not-empty` while the textfield is supposed to be hidden and clear-on-hide applied, if the textfield value will be processed/saved into the database or not. I'd expect the logic/validation to correct/block this. Maybe that's a better test to write?

The output/result of check logic isn't too relevant here - if you spoof stuff and get back odd responses, okay, "lekker belangrijk", it's only in-memory data anyway and doesn't have an impact. But if the same mechanisms allow stuff through while persisting the step data, we have a potential issue.

viktorvanwijk: > if you can spoof "final" situations by making the PUT call to save the data?

I think this has been possible for a long time :grimacing:

viktorvanwijk: Discussed on slack: requires follow-up

sergei-maertens (commented):
* It looks like `openforms.formio.visibility.get_component_empty_value` can be deleted because it's now dead code
* there's some indirect coverage change in `SelectBoxesFormatter.format`, see codecov. I can't really pinpoint where this is coming from and if that branch is still relevant 🤔 
* `PropertyAction.eval` has a line `if self.component not in configuration:` that is no longer hit - looks like there's no test case any longer that targets a component in another step or the code path is just not relevant anymore? This may be related to the logic rule ordering?
* A similar situation happens in `openforms.submissions.logic.rules._handle_clear_on_hide_for_untriggered_rule`
* `openforms.submissions.api.viewsets.SubmissionStepViewSet.update` lost some branch coverage on: `if not subsequent_step.is_applicable and subsequent_step.completed:`

`openforms.formio.service.get_component_empty_value` unfortunately seems to still be relevant, but less so now.

viktorvanwijk:
> It looks like openforms.formio.visibility.get_component_empty_value can be deleted because it's now dead code

Removed

> there's some indirect coverage change in SelectBoxesFormatter.format, see codecov. I can't really pinpoint where this is coming from and if that branch is still relevant 🤔

It's because of reverting the bugfix for #5980, combined with an incorrect test setup :)

The other three points seem to be because of the removed `test_clear_on_hide_behaviour_with_multiple_steps`. I have added it back with a revised setup.

Let's see what codecov thinks. EDIT: hurray :partying_face:

sergei-maertens (approved):

viktorvanwijk:
Backport check script update to 3.5.x: 5b105b9c

PR #6345, opened 2026-06-03
Co-sign component in non-applicable step blocks registration

### LINKED ISSUE #6259: Do not change a form to a "cosign" form if the cosign component is part of a non-applicable step.

3.3+

### Customer reference

Venray 191

The behaviour should work like "not required" and "hidden", thus when the cosign component is part of a skipped step, the form should never become a cosign form. 

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Closes #6259

**Changes**

Ensures that co-sign components in non-applicable don't block the registration and PDF rendering of a form submission.

- [x] Registration
- [x] PDF rendering

I've tried to find a more generic/global solution, but couldn't find a good place to add this check (i tried it with the `total_configuration_wrapper` of `submissions/model/submission.py`, but couldn't get it to work). Let me know if there is a more suitable place/solution for this problem.

### The problem, as far as i understand it, is as follows:

The `CosignState._find_component()` method would search across all form components, looking of a `co-sign` component.  `_find_component` did take into account hidden and hidden parent situations, but it didn't account for non-applicable form steps. So `_find_component` would find a `co-sign` component, which didn't have any data (because the form step was non-applicable), and complain about it.

Next we have the `evaluate_form_logic` method, which corrects the form step data as part of the final "Update relevant data structures" step. This would set the form step `unsaved_data` to an empty dict.

Finally comes the submission report / PDF generation, which also looks for `co-sign` components as part of the `confirmation_page_content` code. This again goes through the same steps as before, checking the `CosignState`, calling `_find_component()` and applying form logic rules. This would then crash when applying the logic rules, due to the previous change in `evaluate_form_logic`.

When `StepNotApplicableAction.apply` is called, it expects the step `unsaved_date` to be `None`. Normally, when applying the logic action when using the form, this is the case. But as the `evaluate_form_logic` changed the `unsaved_date` to a dict, this crashes the PDF generation.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/cosigning.py
+++ b/src/openforms/submissions/cosigning.py
@@ -103,23 +103,27 @@ class CosignState:
 
         check_submission_logic(self.submission, reset_configuration_wrapper=False)
 
-        # use the complete view on the Formio component tree(s)
-        configuration_wrapper = self.submission.total_configuration_wrapper
-
         # there should only be one cosign component, so we check our flatmap to find the
         # component.
-        for component in configuration_wrapper.component_map.values():
-            if component["type"] == "cosign":
-                # we can't just blindly return the component, as it may be conditionally
-                # displayed, which is equivalent to "cosign not relevant". See
-                # https://github.com/open-formulieren/open-forms/issues/3901
-                state = self.submission.variables_state
-                visible = configuration_wrapper.is_visible_in_frontend(
-                    component["key"], values=state.get_data(include_unsaved=True)
-                )
-                if not visible:
-                    return None
-                return component
+        for step in self.submission.steps:
+            # co-sign components inside non-applicable steps are not relevant
+            if not step.is_applicable:
+                continue
+
+            assert step.form_step is not None
+            configuration_wrapper = step.form_step.form_definition.configuration_wrapper
+            for component in configuration_wrapper:
+                if component["type"] == "cosign":
+                    # we can't just blindly return the component, as it may be conditionally
+                    # displayed, which is equivalent to "cosign not relevant". See
+                    # https://github.com/open-formulieren/open-forms/issues/3901
+                    state = self.submission.variables_state
+                    visible = configuration_wrapper.is_visible_in_frontend(
+                        component["key"], values=state.get_data(include_unsaved=True)
+                    )
+                    if not visible:
+                        return None
+                    return component
         return None
 
     @property
--- a/src/openforms/submissions/rendering/renderer.py
+++ b/src/openforms/submissions/rendering/renderer.py
@@ -68,13 +68,15 @@ class Renderer:
         """
         prefetch_related_objects(self.steps, "form_step__logic_rules")
         for step in self.steps:
-            new_configuration = evaluate_form_logic(
-                submission=self.submission, step=step
-            )
-            # update the configuration for introspection - note that we are mutating
-            # an instance here without persisting it to the backend on purpose!
-            # this replicates the run-time behaviour while filling out the form
-            step.form_step.form_definition.configuration = new_configuration
+            if step.is_applicable:
+                new_configuration = evaluate_form_logic(
+                    submission=self.submission, step=step
+                )
+                # update the configuration for introspection - note that we are mutating
+                # an instance here without persisting it to the backend on purpose!
+                # this replicates the run-time behaviour while filling out the form
+                step.form_step.form_definition.configuration = new_configuration
+
             submission_step_node = SubmissionStepNode(renderer=self, step=step)
             if not submission_step_node.is_visible:
                 continue
--- a/src/openforms/submissions/tests/renderer/test_renderer.py
+++ b/src/openforms/submissions/tests/renderer/test_renderer.py
@@ -138,3 +138,36 @@ class FormNodeTests(TestCase):
         # 3. Query the form logic rules for the submission form (and this is cached)
         with self.assertNumQueries(3):
             list(renderer)
+
+    def test_renderer_export_mode_with_form_logic_and_disabled_step(self):
+        """
+        Assert that the renderer contains all steps when exporting, even if some are
+        disabled.
+        """
+        # set up logic
+        form = self.submission.form
+        FormLogicFactory.create(
+            form=form,
+            json_logic_trigger=True,
+            actions=[
+                {
+                    "form_step_uuid": f"{self.sstep2.form_step.uuid}",
+                    "action": {
+                        "name": "Step is not applicable",
+                        "type": "step-not-applicable",
+                    },
+                }
+            ],
+        )
+        form.apply_logic_analysis()
+
+        renderer = Renderer(
+            submission=self.submission, mode=RenderModes.export, as_html=True
+        )
+
+        nodes = [node for node in renderer if isinstance(node, SubmissionStepNode)]
+
+        self.assertEqual(len(nodes), 2)
+        enabled_step_node, non_applicable_step_node = nodes
+        self.assertEqual(enabled_step_node.step, self.sstep1)
+        self.assertEqual(non_applicable_step_node.step, self.sstep2)
--- a/src/openforms/submissions/tests/test_post_submission_event.py
+++ b/src/openforms/submissions/tests/test_post_submission_event.py
@@ -13,7 +13,7 @@ from openforms.authentication.service import AuthAttribute
 from openforms.config.models import GlobalConfiguration
 from openforms.emails.tests.factories import ConfirmationEmailTemplateFactory
 from openforms.forms.constants import LogicActionTypes, PropertyTypes
-from openforms.forms.tests.factories import FormLogicFactory
+from openforms.forms.tests.factories import FormLogicFactory, FormStepFactory
 from openforms.logging.models import TimelineLogProxy
 from openforms.payments.constants import PaymentStatus
 from openforms.payments.tests.factories import SubmissionPaymentFactory
@@ -26,7 +26,7 @@ from openforms.tests.utils import log_flaky
 from ..constants import PostSubmissionEvents, RegistrationStatuses
 from ..models import SubmissionReport
 from ..tasks import on_post_submission_event
-from .factories import SubmissionFactory
+from .factories import SubmissionFactory, SubmissionStepFactory
 
 
 @temp_private_root()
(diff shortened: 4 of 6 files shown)


src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -1195,6 +1195,99 @@ def test_preregister_success_and_registration_failure_records_counter_and_date_o
...
+        submission = SubmissionFactory.from_components(
+            components_list=[
+                {
+                    "key": "cosign",
+                    "type": "cosign",
+                    "label": "Cosign component",
+                    "validate": {"required": True},
+                },
+            ],
+            submitted_data={"cosign": ""},
```
viktorvanwijk:
```suggestion
            submitted_data={}
```
If the component was part of a non-applicable step, we will not have any submission data for it. (Passing `submitted_data` will create the related submission value variables)

src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -1195,6 +1195,99 @@ def test_preregister_success_and_registration_failure_records_counter_and_date_o
...
+            language_code="en",
+        )
+
+        FormLogicFactory.create(
+            form=submission.form,
+            json_logic_trigger=True,
+            actions=[
+                {
+                    "action": {"type": LogicActionTypes.step_not_applicable},
+                    "form_step_uuid": str(submission.form.formstep_set.all()[0].uuid),
```
viktorvanwijk:
```suggestion
                    "form_step_uuid": str(submission.form.formstep_set.get().uuid),
```
Minor thing, but you could use `.get()` to fetch the only available step. It shows that we expect only one to be present at this time :)

src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -1195,6 +1195,99 @@ def test_preregister_success_and_registration_failure_records_counter_and_date_o
...
+                        {
+                            "key": "cosign",
+                            "type": "cosign",
+                            "label": "Cosign component",
+                            "validate": {"required": True},
+                        },
+                    ],
+                },
+            ],
+            submitted_data={"cosign": ""},
```
viktorvanwijk: Also for hidden fields (with clearOnHide, which is the default when it's not specified in the component definition), no submission data will exist for this component

src/openforms/submissions/cosigning.py
```diff
@@ -108,18 +109,25 @@ def _find_component(self) -> Component | None:
...
+                    # we can't just blindly return the component, as it may be conditionally
+                    # displayed, which is equivalent to "cosign not relevant". See
+                    # https://github.com/open-formulieren/open-forms/issues/3901
+                    state = self.submission.variables_state
+                    visible = configuration_wrapper.is_visible_in_frontend(
+                        component["key"], values=state.get_data(include_unsaved=True)
+                    )
+                    if not visible:
+                        return None
+                    return component
```
viktorvanwijk, line 126: With my comments on the submission data in the tests, what if we just check whether the relevant variable for the cosign component exists in the submission data with `if component["key"] in state.get_data(include_unsaved=False)` 🤔 This ensures that the component was visible, and not part of an applicable step I think.

EDIT: hmm this might lead to an edge case when "clearOnHide" is not enabled, because a variable will have been created in that case. Probably not the safest route

src/openforms/submissions/cosigning.py
```diff
@@ -41,6 +41,7 @@
 from openforms.formio.typing import Component
 from openforms.typing import JSONObject
 
+from ..forms.models import FormStep
```
viktorvanwijk:
```suggestion
from openforms.forms.models import FormStep
```

src/openforms/submissions/form_logic.py
```diff
@@ -185,11 +185,15 @@ def evaluate_form_logic(
...
-    step.unsaved_data = updated_step_data
+    if step.is_applicable:
+        updated_step_data = FormioData()
+        for key, variable in relevant_variables.items():
+            if (
+                not variable.form_variable
+                or initial_data[key] != data_for_evaluation[key]
+            ):
+                updated_step_data[key] = data_for_evaluation[key]
+        step.unsaved_data = updated_step_data
```
viktorvanwijk: From commit message 8f9a83d3:
Unsaved data is only set in the context of `evaluate_form_logic`, so this can only be not `None`
if logic for that step was evaluated, and then evaluated again causing the same step to be not applicable -> weird.

The assertion was added with the assumption that we do not evaluate form logic for non-applicable steps. Apparently this does happen in `renderer.get_children` I think, but just wasn't uncovered by our existing tests. I still think it's weird to do this, because it also doesn't happen while filling out a form. My recommendation would be to check what happens if we skipped evaluating form logic for those steps, instead of adding this if statement. If it turns out to break a lot of stuff, I would remove the assertion in `StepNotApplicableAction.apply` with an explanation / link to a follow-up issue.

src/openforms/submissions/cosigning.py
robinmolen: Check..

I also think that explicitly skipping non-applicable steps makes it clearer when a co-sign field should be checked or not.

src/openforms/submissions/form_logic.py
robinmolen: I don't think we can easily **not** perform the form logic evaluation for non-applicable steps, as dynamically applicable steps are determined in the `evaluate_form_logic` step 7.1. Also, a form step that by default is not applicable, but becomes applicable through form logic, should be evaluated/checked.

So, besides leaving the if statement, removing the assertion seems to be the only viable action. Or, we change the assertion to `assert step.unsaved_data is None or step.unsaved_data == FormioData()`

viktorvanwijk: Talking specifically about `Renderer.get_children` where we iterate over the available submission steps: I'm expecting this iteration to be of the same order as what would happen when a user fills out a form. This means any (non) applicable action __should__ only affect steps __after__ the current one. The `SubmissionStep.is_applicable` flag should therefore already be accurate before we try to evaluate form logic for that step. It doesn't make sense for a logic rule which takes as input components from a certain step, to affect the applicability of that same (or previous) step(s).

> Also, a form step that by default is not applicable, but becomes applicable through form logic, should be evaluated/checked.

I agree, but the only way we get in this situation is if we don't evaluate logic **at all**. In that case, we cannot not trust the `is_applicable` flag. Evaluating logic of all steps before the one that is dynamically (not) applicable should bring us in the correct state

robinmolen: Check... I noticed I missing that part in my tests, that the non-applicable step should not be the first step.. 🫣

By updating my tests to ensure that the `set applicable` form logic is done for upcoming steps, and skipping the evaluation in `renderer.children` for non-applicable steps, it seems to work fine.

I've added another test that ensures that form steps that are made applicable through form logic still are evaluated.

src/openforms/submissions/rendering/renderer.py
```diff
@@ -68,6 +68,9 @@ def get_children(self) -> Iterator[SubmissionStepNode | VariablesNode]:
         """
         prefetch_related_objects(self.steps, "form_step__logic_rules")
         for step in self.steps:
+            if not step.is_applicable:
+                continue
```
viktorvanwijk: I'm not sure if skipping completely here is the solution though 🤔 `SubmissionStepNode.is_visible` always returns `True` when we are in export mode, even for non-applicable steps. That changes with this if-statement addition

If we only skip logic evaluation instead of continuing completely, the warning inside `is_visible` is not relevant anymore either. Or at least, only relevant when the step is applicable

robinmolen: I've updated the renderer to only perform the evaluation when the step is applicable.

To prevent flooding the logger with invalid data, the `submissionStepNode.is_visible` only logs missing evaluations for applicable steps.

I've also added a test to ensure that non-applicable steps remain in the renderer children when exporting.

viktorvanwijk (approved):

src/openforms/submissions/tests/renderer/test_renderer.py
```diff
@@ -138,3 +138,39 @@ def test_performance_num_queries(self):
...
+            submission=self.submission, mode=RenderModes.export, as_html=True
+        )
+
+        nodes = [
+            node for node in renderer if isinstance(node, FormNode | SubmissionStepNode)
+        ]
+
+        self.assertEqual(len(nodes), 3)
+        enabled_step_node = nodes[1]
+        non_applicable_step_node = nodes[2]
```
sergei-maertens:
```suggestion
        nodes = [
            node for node in renderer if isinstance(node, SubmissionStepNode)
        ]

        self.assertEqual(len(nodes), 2)
        enabled_step_node, non_applicable_step_node = nodes
```
the form node is not being asserted on, is it?

src/openforms/submissions/tests/test_post_submission_event.py
```diff
@@ -1197,6 +1197,171 @@ def test_preregister_success_and_registration_failure_records_counter_and_date_o
...
+                    {
+                        "key": "cosign",
+                        "type": "cosign",
+                        "label": "Cosign component",
+                        "validate": {"required": True},
+                    },
+                ]
+            },
+        )
+        SubmissionStepFactory.create(submission=submission, form_step=form_step2)
```
sergei-maertens: this is a bit odd, because with your current logic setup, the user would have never submitted this step and thus no `SubmissionStep` instance would exist in the database.

A more realistic scenario with a checkbox in step one that's used as logic trigger for step 2 will make the test easier to understand and reflect real-world scenario's. My first inclination here was to think the test setup is wrong.

src/openforms/submissions/cosigning.py
sergei-maertens: the total configuration wrapper is no longer needed if you're looping over individual steps anyway

src/openforms/submissions/cosigning.py
```diff
@@ -108,18 +109,25 @@ def _find_component(self) -> Component | None:
...
-                    return None
-                return component
+        for step in self.submission.steps:
+            # co-sign components inside non-applicable steps are not relevant
+            if not step.is_applicable:
+                continue
+
+            form_step = step.form_step
+            assert isinstance(form_step, FormStep)
+            for component in form_step.iter_components():
```
sergei-maertens:
```suggestion
            assert step.form_step is not None
            configuration_wrapper = step.form_step.form_definition.configuration_wrapper
            for component in configuration_wrapper:
```

this uses the local configuration wrapper. Usage of `iter_components` should be limited to `openforms.formio` internals as much as possible.

src/openforms/submissions/cosigning.py
robinmolen: The `configuration_wrapper` was needed to check the visibility of the components, but that now uses the form definition specific `configuration_wrapper`

src/openforms/submissions/tests/test_post_submission_event.py
robinmolen: I thought keeping the logic rule simple with a `True` trigger would be better, to prevent too much test clutter. But i understand your reasoning.

I've updated the tests by switching the trigger to a checkbox var comparision, to mimic a more real-world scenario.

sergei-maertens: yeah I understand that as well, however in the past months I've been cleaning up a lot of tests like that that did things very minimalistic to "declutter", yet now that stricter implementations are arriving those are the tests that are breaking :grimacing:

sergei-maertens (approved):

sergei-maertens:
Backports:

* 3.5.x: c4f0061088956a88edd1c83d56169fdbd6333f39
* 3.4.x: 9e87766eca9662f422f6e956f8cd7d95c004060c

PR #6348, opened 2026-06-04
[#6277] Update jsonLogic calls for variable action according to the library

### LINKED ISSUE #6277: Update frontend and backend JSON logic implementation
This is to properly check if a value is present in the context, instead of defaulting to `None`/`null`. In the backend, it means putting a custom object as default in the `not_found` argument of `get_var`. Need to implement something similar in the frontend, possibly need a custom override of the "var" operation.

To not break the `!` and `!!` operators, probably need to make sure the custom value is nullish, like the `undefined` value in JS. This also makes sure `{"if": [{"var": "foo"}, ...]}` will still work

vaszig:
Closes #6277

**Changes**

We use `jsonLogic` function from `json-logic-py` fork, where the `get_var` operation returns `None` in case of None/null values. This needs to change in OF so we adopt to the new flag` use_var_undefined` when we want to have 'undefined' as the return value instead of None. This is properly handled by OF where the logic rule is executed.

- [x] Update the code according to the new version of the library
- [x] Upgrade `json-logic-py` to new version

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -12,7 +12,7 @@ from django.core.cache import cache
 from django.core.serializers.json import DjangoJSONEncoder
 
 from glom import assign
-from json_logic import jsonLogic
+from json_logic import UNDEFINED_VALUE, jsonLogic
 
 from openforms.dmn.service import evaluate_dmn
 from openforms.formio.service import (
@@ -430,7 +430,17 @@ class VariableAction(ActionOperation):
         data_for_visible_state: FormioData,
     ) -> DataMapping:
         with log_errors(self.value, self.rule):
-            return {self.variable: jsonLogic(self.value, context.data)}  # pyright: ignore[reportArgumentType]
+            result = jsonLogic(
+                self.value,  # pyright: ignore[reportArgumentType]
+                context.data,  # pyright: ignore[reportArgumentType]
+                use_var_undefined=True,
+            )
+            # variables with None/null values are returned as UNDEFINED_VALUE when
+            # use_var_undefined is set to True and they should not be returned
+            if result is UNDEFINED_VALUE:
+                return {}
+
+            return {self.variable: result}
 
 
 class DataMappingsConfig(TypedDict):
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -436,7 +436,9 @@ class VariableAction(ActionOperation):
                 use_var_undefined=True,
             )
             # variables with None/null values are returned as UNDEFINED_VALUE when
-            # use_var_undefined is set to True and they should not be returned
+            # use_var_undefined is set to True and they are missing from the context
+            # entirely. If they are present in the context and None, then the output
+            # will be None.
             if result is UNDEFINED_VALUE:
                 return {}
 
@@ -646,6 +648,9 @@ class ServiceFetchAction(ActionOperation):
         var = self.rule.form.formvariable_set.get(key=self.variable)
         with log_errors({}, self.rule):  # TODO proper error handling
             result = perform_service_fetch(var, context, str(submission.uuid))
+            if result is None:
+                return {}
+
             return {var.key: result.value}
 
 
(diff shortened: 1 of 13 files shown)


vaszig:
Taking a look at the missing coverage now. this approach was working before we make the final changes in the library, I will take a look tomorrow to see what's wrong, because now it's not working.

viktorvanwijk:
> Taking a look at the missing coverage now. viktor this approach was working before we make the final changes in the library, I will take a look tomorrow to see what's wrong, because now it's not working.

Are you sure it was working before? I would not expect this to work without my PR being merged :thinking: 

Also I think we don't have to update all occurrences of `jsonLogic` actually - we can discuss in more detail tomorrow :)

(you tagged the wrong Viktor btw :grin:)

src/openforms/utils/json_logic/partial_evaluation.py
```diff
@@ -83,7 +83,7 @@ def partially_evaluate_json_logic(
                 argument[0], data
             )
             if resolved:
```
vaszig, line 85: Remove the flag `use_var_undefined` because whenever we call this all the variables are present (the same applies to the `jsonLogic` calls below).

src/openforms/submissions/api/serializers.py
```diff
@@ -477,9 +469,27 @@ def get_configuration(self, instance) -> dict:
...
+        so they have to be removed.
+        """
+        normalized_data: FormioData = FormioData()
+        for var_key in data:
+            var_value = data[var_key]
+            if var_value is UNDEFINED_VALUE:
+                continue
+
+            normalized_data.update({var_key: var_value})
+        return normalized_data or data
```
vaszig: This should not be needed, these variables should not end up here. 

** cannot iterate over the FormioData

vaszig:
This requires the functionality of https://github.com/open-formulieren/open-forms/pull/6341. So when this is approved/merged I can continue.

src/openforms/submissions/logic/actions.py
```diff
@@ -430,7 +430,17 @@ def eval(
         data_for_visible_state: FormioData,
     ) -> DataMapping:
         with log_errors(self.value, self.rule):
-            return {self.variable: jsonLogic(self.value, context.data)}  # pyright: ignore[reportArgumentType]
+            result = jsonLogic(
+                self.value,  # pyright: ignore[reportArgumentType]
+                context.data,  # pyright: ignore[reportArgumentType]
+                use_var_undefined=True,
+            )
+            # variables with None/null values are returned as UNDEFINED_VALUE when
```
viktorvanwijk: Only when they are missing from the context entirely. If they are present in the context with `None`, the output will be `None`

src/openforms/submissions/logic/service_fetching.py
```diff
@@ -81,7 +81,7 @@ def _do_fetch():
             # XXX raise warning if len(result) > 1 ?
             value = jq.compile(expression).input(raw_value).first()
         case DataMappingTypes.json_logic, expression:
-            value = jsonLogic(expression, raw_value)
+            value = jsonLogic(expression, raw_value, use_var_undefined=True)
```
viktorvanwijk, line 84: With this change, I think we need to make sure to remove it in `ServiceFetchAction.eval`, because otherwise the undefined value might leak into the data.

src/openforms/submissions/tests/form_logic/test_modify_components.py
```diff
@@ -1510,3 +1510,105 @@ def test_variable_action_with_non_triggered_and_property_action(self):
...
+                },
+            ]
+        }
+
+        state = submission.variables_state
+        data = state.get_data(include_unsaved=True)
+
+        self.assertEqual(configuration, expected)
+        self.assertEqual(
+            data, {"test": "", "textField": None, "checkbox": True, "textField1": ""}
```
viktorvanwijk, line 1616: I think a comment explaining why this data is expected will help future developers reading this test :)

src/openforms/submissions/logic/actions.py
viktorvanwijk:
```suggestion
            # variables are returned as UNDEFINED_VALUE when
```

viktorvanwijk (approved):
One minor suggestion for the comment still, but otherwise looks OK to me :)

PR #6353, opened 2026-06-05
Documentation for/about registration attributes

### LINKED ISSUE #6336: Support mapping entire address as registration attribute (StUF-ZDS, ZGW APIs)

sergei-maertens:
Closes #6336

Follow up for #6336 PR review comments.

There was *some* documentation about registration attributes, so I added some more nuances and extended remarks per registration plugin.

[skip: e2e]


annashamray (approved):
Nice documentation!

PR #6357, opened 2026-06-08
Move docs linkcheck to weekly cron

sergei-maertens:
With the throttling and general unavailability issues on Github, the docs linkcheck is failing often for PRs with false positives. This creates noise and a culture of ignoring docs build failures.

Instead, we can run them on a weekly basis and check once a week if the failure is genuine and needs to be handled, or if it's flakiness.

[skip: e2e]

### DIFF
--- a/.github/workflows/link-checks.yml
+++ b/.github/workflows/link-checks.yml
@@ -1,0 +1,63 @@
+name: Run link checks (docs)
+
+on:
+  schedule:
+    - cron: '0 7 * * 1'  # 7 AM every Monday
+
+permissions: {}
+
+jobs:
+
+  setup:
+    name: Set up the build variables
+    runs-on: ubuntu-latest
+    outputs:
+      sdk-ref: ${{ steps.sdk-vars.outputs.sdk-ref }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
+        with:
+          persist-credentials: false
+      - name: Extract version info
+        id: vars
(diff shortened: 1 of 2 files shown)


.github/workflows/link-checks.yml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      sdk-ref: ${{ steps.sdk-vars.outputs.sdk-ref }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
+        with:
+          persist-credentials: false
+      - name: Extract version info
+        id: vars
+        uses: maykinmedia/open-api-workflows/actions/extract-version@v6
```
viktorvanwijk: Could you already pin this a commit hash 😇 
`79102b911003d75203ca2fed7df01ad79d9b6bba # v6.4.0`

.github/workflows/link-checks.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  setup:
+    name: Set up the build variables
+    runs-on: ubuntu-latest
+    outputs:
+      sdk-ref: ${{ steps.sdk-vars.outputs.sdk-ref }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
+        with:
+          persist-credentials: false
```
viktorvanwijk, line 19: I see this is a new parameter that is not in original job in `ci.yml`. Is it something we need to add in more places?

sergei-maertens: yep! zizmor will report it

See https://github.com/maykinmedia/django-common/pull/75/changes for prior art :-)

viktorvanwijk (approved):

Issue #6281, opened 2026-05-11
Make it possible to set the document type of a file component through a dropdown


sergei-maertens:
Now you must type the description manually. It'd be nicer if the registration backend could be taken into account, which specifies which api group + catalogue to use and use that to populate the document upload mapping.

Note that this effectively obsoletes the "registration" tab for file components.

**Acceptance criteria**

- [x] For objects API, list available document types from querying the selected catalogue
- [x] For ZGWS APIs, list available document types from querying the selected catalogue *and* case type
- [x] The `registration.documentType` configuration is gone from the component level
- [x] ~~Support configuring document types based on the *api group* catalogue configuration - now the UI only allows when the backend-level catalogue is specified.~~ barely used, it's easier to move this configuration to the backend
- [x] There is an import converter that copies/moves the component level configuration if there's no plugin-level configuration
- [x] Data migration that copies over the configuration one-time (similar to converter) (partly done in earlier PR in c8ad586109 and b7a1e1709e)
- [x] Ensure it works with file components inside edit grids
- [x] Ensure that file component keys in `camelCase` don't get mangled

sergei-maertens:
* `registration.docVertrouwelijkheidaanduiding` is now handled generically through `src/openforms/submissions/models/submission_files.py`, seems to be done for both objects API and ZGW (makes sense)
* ensure bronorganisatie + titel are also taken into account accordingly!

PR #6361, opened 2026-06-09
Clean up Objects API v2 registration code structure

sergei-maertens:
Part of #6281

**Changes**

Removed code duplication by refactoring into re-usable components and/or hooks, and moved the code into a subfolder for the variables table edit options.

This makes the UI components for the map, selectboxes and addressNL components easier to read and maintain.

On top of that, the Formik state manipulation to manage the variable mappings is simplified - we no longer need to deal with FieldArray structures and can edit the options almost directly.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
```diff
@@ -81,6 +96,7 @@ const ObjectsApiVariableConfigurationEditor = ({variable}) => {
             label={
               <FormattedMessage description="'Variable key' label" defaultMessage="Variable key" />
             }
+            required
```
SonnyBA, line 99: Wasn't this always supposed to be required?

src/openforms/js/components/admin/form_design/registrations/objectsapi/hooks.js
```diff
@@ -0,0 +1,69 @@
...
+import {useAsync} from 'react-use';
+
+import {
+  getCatalogueOption,
+  groupAndSortCatalogueOptions,
+  useGetDocumentTypes as useGenericGetDocumentTypes,
+} from 'components/admin/forms/zgw';
+import {get} from 'utils/fetch';
+
+const CATALOGUES_ENDPOINT = '/api/v2/objects-api/catalogues';
```
SonnyBA, line 11: Not sure if it is the correct place for these endpoints but I noticed we have a [`constants.js`](https://github.com/open-formulieren/open-forms/blob/155f1869c54419d71cc0df6bb2d8e8cb34195b04/src/openforms/js/components/admin/form_design/constants.js) that lists several endpoints, can these be moved there?

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiVariableConfigurationEditor.js
sergei-maertens: Indeed :-) stumbled on it and decided to fix it up as I came along.

src/openforms/js/components/admin/form_design/registrations/objectsapi/hooks.js
sergei-maertens: I'm a strong believer in co-location of code that's used together to reduce the cognitive overhead when reading code.

This endpoint is only used in one place, inside this module and it's a constant to avoid hardcoding a URL in the function body. Moving this constant into a shared module makes it "public API" because it needs to be exported to be able to import it in this place where it's used, and requires you to open another file to see the value.

I think that organizing code by "what it is" (a constant) is a bit of an anti pattern if it's not a constant that's going to be used in multiple places. Also seen from the perspective of removing a feature - ideally removing a feature is equivalent to removing one file or one directory. If you'd be removing this feature here and the constant in in constants.js, I'd have to track down if that constant is used in other places and then also update the constants file. Or, just completely forget about it and then you end up with dead code.

SonnyBA (approved):

PR #6365, opened 2026-06-10
:bug: Fix submission lookup crash

sergei-maertens:
The patch in 4ea3072ef8fc6784355d6bed98bc7bdd306a60ec was not complete.

While it fixed the missing context propagation, the early resolution of the submission is overwritten in the event-specific handling and leads to a crash due to the transaction still being uncommitted.

I've opted for binding the submission pk in the prefill_variables function in the service, as that's the entrypoint most commonly used in tests.

[skip: e2e]


vaszig (approved):

sergei-maertens:
Backported to 3.5.x in d7b106710526224b927864ac6a1db7be31625f03

PR #6372, opened 2026-06-12
(2/5) File component registration options - store configuration options in backend options

### LINKED ISSUE #6281: Make it possible to set the document type of a file component through a dropdown

sergei-maertens:
Partly closes #6281

**Changes**

Updated the ZGW APIs & Objects APIs backend type definitions and configuration options serializers. They now accept and validate the `files` options array, so that configuration can be persisted. A future PR updates the runtime behaviour *during* registration to actually use these options.

Note that we can't yet validate the file component keys specified here, for that we need the v3 endpoints to be completed and used, as the registration backends are created *before* the form definitions and form steps are saved in the backend.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/contrib/zgw/typing.py
+++ b/src/openforms/contrib/zgw/typing.py
@@ -1,4 +1,15 @@
-from typing import NotRequired, TypedDict
+from typing import Literal, NotRequired, TypedDict
+
+type VertrouwelijkheidAanduiding = Literal[
+    "openbaar",
+    "beperkt_openbaar",
+    "intern",
+    "zaakvertrouwelijk",
+    "vertrouwelijk",
+    "confidentieel",
+    "geheim",
+    "zeer_geheim",
+]
 
 
 class DocumentOptions(TypedDict):
--- a/src/openforms/registrations/contrib/objects_api/config.py
+++ b/src/openforms/registrations/contrib/objects_api/config.py
@@ -74,7 +74,6 @@ class FileComponentOptionsSerializer(serializers.Serializer):
             "apply."
         ),
     )
-    # TODO: validate that it's in the specified catalogue
     document_type_description = serializers.CharField(
         label=_("Document type description"),
         required=False,
@@ -540,6 +539,24 @@ def _validate_catalogue_and_document_types(attrs: RegistrationOptions) -> None:
                     raise serializers.ValidationError(
                         {field: err_msg}, code="not-found"
                     )
+
+        _files_errors = {}
+        for index, file_options in enumerate(attrs.get("files", [])):
+            if not (description := file_options.get("document_type_description")):
+                continue
+            assert catalogus is not None
+            versions = client.find_informatieobjecttypen(
+                catalogus=catalogus["url"], description=description
+            )
+            if versions is None:
+                err_msg = _(
+                    "No document type with description '{description}' found."
+                ).format(description=description)
+                _files_errors[index] = {
+                    "document_type_description": ErrorDetail(err_msg, code="not-found")
+                }
+        if _files_errors:
+            raise serializers.ValidationError({"files": _files_errors})
 
 
 def _validate_objecttype_and_version(attrs: RegistrationOptions) -> None:
--- a/src/openforms/registrations/contrib/objects_api/config.py
+++ b/src/openforms/registrations/contrib/objects_api/config.py
@@ -7,6 +7,7 @@ from drf_spectacular.utils import OpenApiExample, extend_schema_serializer
 from furl import furl
 from rest_framework import serializers
 from rest_framework.exceptions import ErrorDetail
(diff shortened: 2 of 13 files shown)


src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -55,6 +55,53 @@ def to_representation(self, value) -> list[str]:  # pyright: ignore[reportIncomp
...
 
+class FileComponentOptionsSerializer(serializers.Serializer):
+    key = FormioVariableKeyField(
+        label=_("component key"),
+        help_text=_(
+            "Literal component key value of the file component for which the options "
+            "apply."
+        ),
+    )
+    # TODO: validate that it's in the specified catalogue & case type
```
sergei-maertens: TODO in scope of this PR :)

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -65,6 +66,52 @@ class ObjecttypeVariableMappingSerializer(serializers.Serializer):
...
+class FileComponentOptionsSerializer(serializers.Serializer):
+    key = FormioVariableKeyField(
+        label=_("component key"),
+        help_text=_(
+            "Literal component key value of the file component for which the options "
+            "apply."
+        ),
+    )
+    # TODO: validate that it's in the specified catalogue
+    document_type_description = serializers.CharField(
```
sergei-maertens: TODO in scope of this PR

src/openforms/contrib/zgw/typing.py
```diff
@@ -1,4 +1,15 @@
-from typing import NotRequired, TypedDict
+from typing import Literal, NotRequired, TypedDict
+
+type VertrouwelijkheidAanduiding = Literal[
```
SonnyBA, line 3: Can't we use the existing [`VertrouwelijkheidsAanduidingen`](https://github.com/maykinmedia/zgw-consumers/blob/2.0.2/zgw_consumers/api_models/constants.py#L5) choices from zgw-consumers?

src/openforms/registrations/contrib/zgw_apis/tests/test_validators.py
```diff
@@ -783,3 +783,75 @@ def test_validate_document_type_exists_when_description_is_provided(self):
...
+                        "key": "otherFileComponent",
+                        "title": "Custom title",
+                    },
+                ],
+            }
+            serializer = ZaakOptionsSerializer(data=data)
+
+            is_valid = serializer.is_valid()
+
+            self.assertTrue(is_valid)
```
SonnyBA, line 819:
```suggestion
            self.assertTrue(serializer.is_valid())
```

src/openforms/contrib/zgw/typing.py
sergei-maertens: Not really with the current situation - we get actual strings in Python and not the choices/enums that this class represents. We *can* convert that, but it would be a lie, and the type checker will actually complain in a number of places I suspect.

I do have this on my longer-term wish list to properly fix these types or at least make a union of the literals + enum types.

(note that this code already existed and I just moved it into a shared directory)

src/openforms/registrations/contrib/zgw_apis/tests/test_validators.py
sergei-maertens: I strongly disagree with that because I believe in clear visual structure of tests :-)

```
arrange

act

assert
```

with the blank lines being significant.

It's clearly assigning the result of the `act` step, so that you can `assert` on the result.

Your suggestion basically visually combines `act` and `assert` and that leads to harder to debug/read tests.

SonnyBA (approved):

PR #6385, opened 2026-06-18
Ensure legacy imports still work

### LINKED ISSUE #6254: Ensure legacy imports still work in OF 4.0+
I've been getting some reports from internal testing on our env that the `trigger_from_step` is blocking imports etc.

In short - old form exports that are available on for example Samen Delen should have the necessary converts to be able to be imported in 4.0 where certain features are removed. Many of these can be handled already.

Non-exhaustive list (please complete with additional bullets):

- [x] Import with new logic evaluation while logic rules have `trigger_from_step` set -> use the auto migration code that we had for the data migration to convert these
- [x] Disable-next logic actions without a specific form step -> use auto migration with the code from `0120_change_disable_next_logic_action.py`
- [ ] ...

viktorvanwijk:
Partly closes #6254

[skip: e2e]

**Changes**

* Added disable next import conversion
* Remove `trigger_from_step` from logic rules when new logic evaluation is enabled

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/forms/api/serializers/logic/form_logic.py
+++ b/src/openforms/forms/api/serializers/logic/form_logic.py
@@ -230,6 +230,11 @@ class FormLogicSerializer(
             attrs, self.context["form_variables"].variables, self.context["form_steps"]
         )
 
+        # Remove trigger from step for forms that have new logic evaluation enabled
+        # (which is the default when a form without this flag is imported).
+        if self.context["form"].new_logic_evaluation_enabled:
+            attrs.pop("trigger_from_step", None)
+
     def run_validation(self, data) -> None:
         # Override `run_validation` instead of `validate`, because it runs before
         # `LogicComponentActionSerializer.validate`.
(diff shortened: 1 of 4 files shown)


viktorvanwijk:
This merges into `stable/3.5.x` but the fix for disable next should also be in `main`. I did it this way because some of the added tests will not work anymore on 4.0 because of the removed `trigger_from_step` from the serializer. This means this field is not exported anymore in the tests, so importing it also doesn't include it. Note the `trigger_from_step` will still be properly picked up during disable-next action processing, because the actions are adjusted in the raw data loaded from the export :) (before `trigger_from_step` would be removed by the serializer).

The fix for the `trigger_from_step` should not be necessary in main anymore, because the new logic evaluation flag and the `trigger_from_step` validation have been removed (from the serializer).

sergei-maertens (approved):

PR #6392, opened 2026-06-19
[#6350] Support hiding form title

### LINKED ISSUE #6350: Add support for hiding the form title
We want to give the ability to hide the title of the form, as there might be two h1 elements on the page.

vaszig:
Closes #6350 partly

**Changes**

- Added a data attribute which will be picked up by the SDK and will render (or not) the form title.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/docs/developers/embedding.rst
+++ b/docs/developers/embedding.rst
@@ -62,6 +62,12 @@ Available options
     Required. The UUID or slug of the form to embed. This can be obtained via the Open
     Forms admin interface.
 
+
+``showFormTitle``:
+    Optional. When set to ``true`` it shows the form title when the form is rendered. This
+    can be set to ``false`` in case this is needed (for example useful for embedded forms
+    that already have an ``h1`` element / title in the dom).
+
 ``basePath``:
     Optional, but highly recommended. The SDK considers this as the base URL and builds all
     URLs relatively to this URL.
--- a/src/openforms/static/sdk-wrapper.mjs
+++ b/src/openforms/static/sdk-wrapper.mjs
@@ -13,6 +13,7 @@ const initializeSDK = async node => {
   const {
     sdkModule,
     formId,
+    showFormTitle,
     baseUrl,
     basePath,
     cspNonce,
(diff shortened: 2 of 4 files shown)


docs/developers/embedding.rst
```diff
@@ -62,6 +62,12 @@ Available options
     Required. The UUID or slug of the form to embed. This can be obtained via the Open
     Forms admin interface.
 
+
+``showFormTitle``:
+    Optional. When set to `true` it shows the form title when the form is rendered. This
```
sergei-maertens:
```suggestion
    Optional. When set to ``true`` it shows the form title when the form is rendered. This
```
and the rest as well - double backticks are necessary in RST for code formatting

src/openforms/static/sdk-wrapper.mjs
```diff
@@ -13,6 +13,7 @@ const initializeSDK = async node => {
   const {
     sdkModule,
     formId,
+    showFormTitle,
```
sergei-maertens, line 16: in its current form, this will be the literal string `'true'` rather than a boolean - is that intentional?

sergei-maertens (changes requested):

src/openforms/static/sdk-wrapper.mjs
vaszig: No, this is a mistake I guess because I couldn't see that working with our local form. I only saw/tested that in the SDK with a story. Could you please elaborate a bit?Do you mean to parse it here in order to have it as boolean?

sergei-maertens: `data-foo` will always be a string, when grabbing it from the dataset :-) so you need to explicitly cast it to a boolean - something like `showFormtitle: showFormTitle === 'false' ? false : true,` should do and is backwards-compatible in case the attribute is not provided.

vaszig: And of course this was the problem for not being able to see the results in the local form. I should have thought about it.

sergei-maertens (approved):

PR #6393, opened 2026-06-19
Provide easy way to add default value to var expression

### LINKED ISSUE #6391: Provide easy way to add default value in a JSON logic variable expression
With the upcoming changes regarding clearOnHide, form designers might want to add a default value to use when a variable cannot be found in the context during logic evaluation.

This means doing the following to a variable operation:
`{"==": [{"var": "textfield"}, ""]}` -> `{"==": [{"var": ["textfield", ""]}, ""]}`

For advanced logic rules, this change can be made easily, but for simple rules it's not possible. To avoid form designers having to discard their existing rules and create new ones, we can enable switching from simple to advanced rules.

It was briefly discussed to add another input field for the default value to a simple logic rule, but this was disregarded to avoid cluttering the simple trigger setup further.

viktorvanwijk:
Closes #6391

[skip: e2e]

**Changes**

* Added icon to simple logic rule which enables converting it to an advanced one
* Removed some unused imports

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,38 @@ Changelog
         `latest <https://open-forms.readthedocs.io/en/latest/changelog.html>`_ docs
         version.
 
+3.5.4 (2026-06-23)
+==================
+
+Regular bugfix release.
+
+**Migration tooling**
+
+.. note::
+
+    In Open Forms 3.5.3 we have provided some context about breaking changes regarding
+    "clear on hide" (see the relevant changelog entry).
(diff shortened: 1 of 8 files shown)


sergei-maertens (approved):

viktorvanwijk:
Backport to 3.5.x: 4a8221dd
(not relevant for 3.4.x)

PR #6395, opened 2026-06-19
Prepare patch release 3.4.10

sergei-maertens:
A nice set of bugfixes.


vaszig (approved):
Except for the translations, where I cannot help, this seems ok to me!

PR #6400, opened 2026-06-22
(Re?)-introduce the `DISABLE_2FA` envvar (for prod/staging settings)

sergei-maertens:
Necessary to be able to disable MFA for the Open Forms event instance where we need simple username/password logins for 8-10 groups of attendees.

Backport to 3.5 is sufficient, so we can deploy `stable/3.5.x` for that instance (currently on 3.5.3).

[skip: e2e]


viktorvanwijk (approved):

sergei-maertens:
Backported to 3.5.x in f0396a6820d97dabf522f5bebbeadb12546c42b2

PR #6401, opened 2026-06-23
[#6398] Add static variable for form url

### LINKED ISSUE #6398: Introduce a static form variable for the form url
- This will be an object like:
```python
{
    "domain": "www.example.com",
    "page": "/sub/example"
    "query": {
        "someParameter": "someValue",
    }
}
```

vaszig:
Closes #6398 

**Changes**

- Added new static variable for the form's url details/metadata.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/variables/static_variables/static_variables.py
+++ b/src/openforms/variables/static_variables/static_variables.py
@@ -1,4 +1,6 @@
 from datetime import date, datetime
+from typing import TypedDict
+from urllib.parse import ParseResult, parse_qs, urlparse
 
 from django.conf import settings
 from django.utils import timezone
@@ -9,6 +11,12 @@ from openforms.submissions.models import Submission
 from ..base import BaseStaticVariable
 from ..constants import FormVariableDataTypes
 from ..registry import register_static_variable
+
+
+class UrlInfo(TypedDict):
+    domain: str
+    page: str
+    query: dict[str, str]
 
 
 @register_static_variable("now")
@@ -88,3 +96,25 @@ class FormID(BaseStaticVariable):
 
     def as_json_schema(self):
         return {"title": "Form identifier", "type": "string", "format": "uuid"}
+
+
+@register_static_variable("form_url")
+class FormUrl(BaseStaticVariable):
+    name = _("Form Url")
+    data_type = FormVariableDataTypes.object
+
+    def get_initial_value(self, submission: Submission | None = None) -> UrlInfo:
+        form_url = str(submission.form_url) if submission else ""
+        parsed: ParseResult = urlparse(form_url)
+        return {
+            "domain": parsed.netloc,
+            "page": parsed.path if parsed.path.endswith("/") else parsed.path + "/",
+            "query": {k: v[0] for k, v in parse_qs(parsed.query).items()},
+        }
+
+    def as_json_schema(self):
+        return {
+            "title": "Form url metadata",
+            "description": "Details concerning the form's url",
+            "type": ["object", "null"],
+        }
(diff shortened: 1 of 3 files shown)


src/openforms/variables/static_variables/static_variables.py
```diff
@@ -11,6 +13,12 @@
 from ..registry import register_static_variable
 
 
+class UrlInfo(TypedDict):
+    domain: str
+    page: str
+    query: dict[str, str]
```
sergei-maertens:
```suggestion
    query: Mapping[str, str]
```
with `collections.abc.Mapping`

The advantage here is that the mapping is immutable by default :) a `dict` is a mapping, but not the other way around.

sergei-maertens: There's another edge case to consider though - repeating query params:

```py
>>> parse_qs("foo=bar&foo=baz")
{'foo': ['bar', 'baz']}
```

Though I'm not sure what's most ergonomic here - having to deal with lists in json logic is quite annoying. For now we could document it that multi-value query params aren't supported?

src/openforms/variables/static_variables/static_variables.py
```diff
@@ -88,3 +96,25 @@ def get_initial_value(self, submission: Submission | None = None) -> str:
...
+            "page": parsed.path if parsed.path.endswith("/") else parsed.path + "/",
+            "query": {k: v[0] for k, v in parse_qs(parsed.query).items()},
+        }
+
+    def as_json_schema(self):
+        return {
+            "title": "Form url metadata",
+            "description": "Details concerning the form's url",
+            "type": ["object", "null"],
+        }
```
sergei-maertens, line 126: You can include a bit more information about the schema by specifying the `properties` and their types :)

src/openforms/variables/static_variables/static_variables.py
```diff
@@ -88,3 +96,25 @@ def get_initial_value(self, submission: Submission | None = None) -> str:
...
+class FormUrl(BaseStaticVariable):
+    name = _("Form Url")
+    data_type = FormVariableDataTypes.object
+
+    def get_initial_value(self, submission: Submission | None = None) -> UrlInfo:
+        form_url = str(submission.form_url) if submission else ""
+        parsed: ParseResult = urlparse(form_url)
+        return {
+            "domain": parsed.netloc,
+            "page": parsed.path if parsed.path.endswith("/") else parsed.path + "/",
```
sergei-maertens: I wouldn't normalize this with a trailing slash tbh - I think you can just expose `parsed.path` and that's it?

src/openforms/variables/static_variables/static_variables.py
```diff
@@ -88,3 +96,25 @@ def get_initial_value(self, submission: Submission | None = None) -> str:
...
+            "query": {k: v[0] for k, v in parse_qs(parsed.query).items()},
```
sergei-maertens: maybe you can use Django's `QueryDict` here? https://docs.djangoproject.com/en/5.2/ref/request-response/#django.http.QueryDict.dict

sergei-maertens (changes requested):

src/openforms/variables/static_variables/static_variables.py
vaszig: Using Django's `QueryDict` complicates things if we go for one value because I will then have to declare `query: Mapping[str, str | list[str]]` but this will not be true. If we go for one value I would keep the manual dictionary build and the `Mapping[str, str]`

sergei-maertens: Manual dict build is okay, but then I'd use the last item rather than the first for consistency with Django's `QueryDict`

sergei-maertens (approved):

PR #6402, opened 2026-06-23
Add import converter for file-level registration options

### LINKED ISSUE #6281: Make it possible to set the document type of a file component through a dropdown

sergei-maertens:
Closes #6281

**Changes**

Updated the import code to process the registration backends included if there are file components in the form definitions of the form with registration options.

This only copies over the options without performing additional validation - that's up to the regular options serializer validation code. Additionally, the data is only copied when there's no explicit `files` configuration yet, ensuring we don't overwrite configuration that was done elsewhere already.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/forms/utils.py
+++ b/src/openforms/forms/utils.py
@@ -2,7 +2,8 @@ import json
 import random
 import string
 import zipfile
-from typing import Any
+from collections.abc import Collection
+from typing import Any, Required, TypedDict
 from uuid import uuid4
 
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


src/openforms/forms/utils.py
```diff
@@ -423,3 +441,84 @@ def clear_old_service_fetch_config(rule: dict) -> None:
...
+        in (
+            OBJECTS_API_PLUGIN_IDENTIFIER,
+            STUF_ZDS_PLUGIN_IDENTIFIER,
+            ZGW_APIS_PLUGIN_IDENTIFIER,
+        )
+    ]
+    if not relevant_backends:
+        return
+
+    # collect all file components, including the ones inside edit grids
```
vaszig, line 470: Fieldsets too, right?

vaszig (approved):

src/openforms/forms/utils.py
sergei-maertens: yep - though I consider that the "normal" behaviour - it's usually editgrid that needs some special treatment :)

PR #6404, opened 2026-06-24
Remove legacy catalogi config options (from the UI & runtime)

### LINKED ISSUE #4939: Remove legacy zaaktype/documenttype URL fields in registration backends

sergei-maertens:
Closes #4939 (partly)

**Changes**

* Removed all UI presence of the legacy configuration options
* Marked the case / document type dropdowns as required in the ZGW APIs plugin options
* Removed fields from the registration backend options serializers

The model fields are not removed yet, as these are required for the 4.0 data migrations that copy configuration from api group to the registration options JSON fields. In 4.1 at the earliest these can be cleaned up - I'll create a backlog item for that.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

  - [x] Added documentation which describes the changes (in a separate PR)


src/openforms/js/components/admin/forms/zgw/CatalogueSelect.js
```diff
@@ -47,7 +47,13 @@ export const groupAndSortOptions = catalogues => {
   return groups;
 };
 
-const CatalogueSelect = ({label, isDisabled = false, loading, optionGroups}) => {
+const CatalogueSelect = ({
+  label,
+  isDisabled = false,
+  loading,
+  optionGroups,
+  isRequired = false,
```
annashamray, line 55: I don't actually see where this option is used.
And if it's used please update also its propTypes

And I don't actually get what is the difference in the use of this component and  `admin/form_design/registrations/zgw/fields/CatalogueSelect.js`

src/openforms/js/components/admin/form_design/registrations/zgw/fields/CatalogueSelect.js
```diff
@@ -19,6 +19,7 @@ const CatalogueSelect = ({loading, optionGroups}) => {
         />
       }
       isDisabled={!zgwApiGroup}
+      isRequired
```
annashamray, line 22: It looks like `isRequired` is not assigned

src/openforms/registrations/contrib/objects_api/config.py
```diff
@@ -124,11 +121,8 @@ class ObjectsAPIOptionsSerializer(JsonSchemaSerializerMixin, serializers.Seriali
     catalogue = CatalogueSerializer(
         label=_("catalogue"),
         required=False,
-        help_text=_(
-            "The catalogue in the catalogi API from the selected API group. This "
-            "overrides the catalogue specified in the API group, if it's set. When "
-            "specified, the document types specified on the API group are ignored."
-        ),
+        help_text=_("The catalogue in the catalogi API from the selected API group."),
+        default=lambda: {"domain": "", "rsin": ""},
```
annashamray, line 125: Why do you use lambda function here?

pyright.pyproject.toml
```diff
@@ -122,10 +120,6 @@ exclude = [
     "src/openforms/payments/tests/",
     "src/openforms/payments/contrib/demo/tests/",
     "src/openforms/payments/contrib/ogone/tests/",
-    "src/openforms/registrations/contrib/objects_api/tests/test_backend_v1.py",
-    "src/openforms/registrations/contrib/objects_api/tests/test_backend_v2.py",
-    "src/openforms/registrations/contrib/objects_api/tests/test_update_payment_status_v1.py",
-    "src/openforms/registrations/contrib/objects_api/tests/test_update_payment_status_v2.py",
```
annashamray, line 128: 🎉

src/openforms/registrations/contrib/zgw_apis/plugin.py
```diff
@@ -442,24 +432,15 @@ def register_submission(
...
-            else:
-                warnings.warn(
-                    "Using the informatieobjecttype URL option is deprecated and will "
-                    "be removed in Open Forms 4.0.",
-                    DeprecationWarning,
-                    stacklevel=2,
-                )
-                informatieobjecttype_url = options["informatieobjecttype"]
+            catalogue = options.get("catalogue")
+            assert catalogue is not None  # enforced by validation
```
annashamray:
```suggestion
            catalogue = options["catalogue"]
```

Can it be this way?

src/openforms/registrations/contrib/zgw_apis/options.py
```diff
@@ -487,143 +439,110 @@ def _validate_catalogue_case_and_doc_type(
...
 
-    # legacy
-    case_type_url = attrs["zaaktype"]
-    document_type_url = attrs["informatieobjecttype"]
-
-    domain, rsin = catalogue_option["domain"], catalogue_option["rsin"]
-
     err_invalid_case_type = ErrorDetail(
-        _("The provided zaaktype does not exist in the specified Catalogi API."),  # type: ignore
+        gettext("The provided zaaktype does not exist in the specified Catalogi API."),
```
annashamray, line 461: Why can't it be lazy?

annashamray (changes requested):
Works for me

Just a couple comments about `required` attribute

src/openforms/js/components/admin/forms/zgw/CatalogueSelect.js
sergei-maertens: It's used in `admin/form_design/registrations/zgw/fields/CatalogueSelect.js`, which imports this component as `GenericCatalogueSelect` and passes the `isRequired` flag.

> And if it's used please update also its propTypes

Ugh I hate proptypes so badly. The DX is terrible, it's easy to forget to update them and the feedback is too late. I want typescript in the backend :sob: 

> And I don't actually get what is the difference in the use of this component...

this component here is specifically for the ZGW registration backend options, while there's another one for the Objects API backend. They have different labels/descriptions and `isRequired` values - ZGW is required while for Objects it's not.

They also use different api groups (`zgwApiGroup` vs `objectsApiGroup`), so that's why the different layers are used.

In general the code is a mess and badly needs to be re-organized, which we *were* going to do in the new admin-ui, but basically there's no budget/room/time for any cleanups like this (including moving to Typescript...) and I'm very bitter about that.

src/openforms/js/components/admin/form_design/registrations/zgw/fields/CatalogueSelect.js
sergei-maertens: it's a boolean flag/prop, this is idiomatic react and equivalent to `isRequired={true}`

src/openforms/registrations/contrib/objects_api/config.py
sergei-maertens: callable as default to avoid using mutable defaults, which a dict is :)

src/openforms/registrations/contrib/zgw_apis/plugin.py
sergei-maertens: Yep! I think my patch to the types came later and it was quite the rebase mess

src/openforms/registrations/contrib/zgw_apis/options.py
sergei-maertens: type checker doesn't like lazy strings, and it effectively doesn't matter because we are in runtime code anyway so the lazy result will immediately be evaluated

But it's primarily to get rid of the type error suppression.

src/openforms/registrations/contrib/objects_api/config.py
annashamray: Oh! I didn't know about that way to assign defaults, nice!

src/openforms/js/components/admin/forms/zgw/CatalogueSelect.js
annashamray: Thank you for the explanation! Now I have better understanding what happens in the code! 
Yeah I can see the benefits of typescript now :)

src/openforms/js/components/admin/form_design/registrations/zgw/fields/CatalogueSelect.js
annashamray: Oh! I didn't know it's a valid syntax, thanks!

annashamray (approved):

Issue #6352, opened 2026-06-05
Update documentation regarding the new  form types and specifically the `single_step` one


vaszig:
- [x] See outdated https://open-forms.readthedocs.io/en/latest/manual/forms/basics.html#manual-forms-basics for example?
- [x] Recipe for the whole procedure of the ability to capture which page is sending the submission (mention multi-value query params aren't supported too)
- [x] Mention explicitly the additional logic actions that can be used (only user defined variables)
- [x] Logic evaluation is performed after submission (only in the backend)

PR #6407, opened 2026-06-25
[#6399] Additional logic actions for single step forms

### LINKED ISSUE #6399: Make additional logic actions available for the Single step form
- [x] change the value of a variable/component
- [x] evaluate DMN (nice to have)

vaszig:
Closes #6399

**Changes**

- Make it possible to select and process *change the value of a variable/component* and *Evaluate DMN* actions in `single_step` forms. The documentation for these will be part of task #6352

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/js/components/admin/forms/VariableMapping.js
+++ b/src/openforms/js/components/admin/forms/VariableMapping.js
@@ -46,6 +46,7 @@ const VariableMappingRow = ({
   translatePropertyChoices,
   propertySelectLabel,
   onRemove,
+  filter,
   includeStaticVariables = false,
   alreadyMapped = [],
   rowCheck = undefined,
@@ -81,6 +82,7 @@ const VariableMappingRow = ({
               description: 'Accessible label for (form) variable dropdown',
               defaultMessage: 'Form variable',
             })}
+            filter={filter}
           />
         </Field>
       </td>
@@ -197,7 +199,14 @@ VariableMappingRow.propTypes = {
    * Validation errors from context
    */
   errors: PropTypes.objectOf(PropTypes.string),
+  /**
+   * Optional callback function to filter available variables.
+   *
+   */
+  filter: PropTypes.func,
 };
+
+const allowAny = () => true;
 
 /**
  * Map form variables to other properties.
@@ -219,6 +228,7 @@ const VariableMapping = ({
   propertySelectLabel,
   includeStaticVariables = false,
   rowCheck,
+  filter = allowAny,
 }) => {
   const {values, getFieldProps} = useFormikContext();
   const {value: mappings = []} = getFieldProps(name);
@@ -267,6 +277,7 @@ const VariableMapping = ({
                   alreadyMapped={alreadyMapped}
                   rowCheck={rowCheck}
                   errors={relatedErrors?.[index]}
+                  filter={filter}
                 />
               ))}
             </tbody>
@@ -359,6 +370,11 @@ VariableMapping.propTypes = {
    * of error messages, where `intl` is the context from react-intl's `useIntl()` hook.
    */
   rowCheck: PropTypes.func,
+  /**
+   * Optional callback function to filter available variables.
+   *
+   */
+  filter: PropTypes.func,
 };
 
 export default VariableMapping;
--- a/src/openforms/js/components/admin/forms/VariableMapping.js
+++ b/src/openforms/js/components/admin/forms/VariableMapping.js
@@ -206,8 +206,6 @@ VariableMappingRow.propTypes = {
   filter: PropTypes.func,
 };
 
-const allowAny = () => true;
-
 /**
  * Map form variables to other properties.
  *
@@ -228,7 +226,7 @@ const VariableMapping = ({
   propertySelectLabel,
   includeStaticVariables = false,
   rowCheck,
-  filter = allowAny,
+  filter,
 }) => {
   const {values, getFieldProps} = useFormikContext();
   const {value: mappings = []} = getFieldProps(name);
--- a/src/openforms/submissions/tests/test_single_step_form.py
+++ b/src/openforms/submissions/tests/test_single_step_form.py
@@ -2,11 +2,19 @@ from unittest.mock import MagicMock, patch
 
 from django.test import override_settings
 
+from django_camunda.models import CamundaConfig
 from rest_framework import status
 from rest_framework.reverse import reverse, reverse_lazy
 from rest_framework.test import APITestCase
 
-from openforms.forms.tests.factories import FormFactory
+from openforms.forms.constants import LogicActionTypes
+from openforms.forms.tests.factories import (
+    FormFactory,
+    FormLogicFactory,
+    FormVariableFactory,
+)
+from openforms.utils.tests.vcr import OFVCRMixin
+from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
 
 from ..models import Submission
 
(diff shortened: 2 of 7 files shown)


src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -411,14 +412,21 @@ def bulk_create_or_update_from_data(
...
         for key, variable in variables.items():
-            try:
-                variable.value = data[key]
-            except KeyError:
-                if reset_missing_variables:
-                    # Variable exists in the database or in-memory, but not in the
-                    # provided data, so we reset it.
-                    variables_keys_to_reset.append(variable.key)
-                continue
+            # Form logic is evaluated after submission for single step forms, so the
```
viktorvanwijk: I would expect this comment to be in the commit message instead - it's a 'why was this code changed' :)

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -411,14 +412,21 @@ def bulk_create_or_update_from_data(
...
+            # submited data will not contain the possible updated values of variables.
+            # By assigning the value from the data to the value of the variable means
+            # that we override the updated value from the logic actions. Even if a
+            # user submits a value for the connected variable it should be updated
+            # only by the logic rule.
```
viktorvanwijk: Could you give some context (to me but also the commit message) as to why is it necessary to overwrite user-submitted data? Seems weird to me that user data is changed without the user having seen it 🤔

src/openforms/submissions/tests/form_logic/test_modify_variables.py
```diff
@@ -1158,3 +1159,179 @@ def test_children_synchronization_with_no_destination_and_source_data(self):
 
         state = submission.variables_state
         self.assertEqual(state.get_data(include_unsaved=True)["editgrid"], [])
+
+    def test_modify_variable_value_in_single_step_form(self):
```
viktorvanwijk: Could you add these tests to a separate commit? They don't seem related to the change that you made in `submission_value_variable.py`, or do you disagree?

EDIT: are these tests even necessary with the ones added to `test_single_step_form.py` - they seem to kinda test the same, no?

src/openforms/submissions/tests/form_logic/test_modify_variables.py
```diff
@@ -1158,3 +1159,179 @@ def test_children_synchronization_with_no_destination_and_source_data(self):
...
+                        "name": "Update variable",
+                        "type": "variable",
+                        "value": {"var": "textField"},
+                    },
+                }
+            ],
+        )
+        form.apply_logic_analysis()
+
+        with self.subTest("no submitted data"):
```
viktorvanwijk: What do you mean with "no submitted data" here? Creating a submission step with `data={"textfield": "foo", "checkbox": True, "textfield1": ""}` implies this data was submitted by the user.

src/openforms/submissions/tests/form_logic/test_modify_variables.py
```diff
@@ -1158,3 +1159,179 @@ def test_children_synchronization_with_no_destination_and_source_data(self):
...
+            ],
+        )
+
+        form.apply_logic_analysis()
+
+        submission = SubmissionFactory.create(form=form)
+        submission_step = SubmissionStepFactory.create(
+            form_step=form.formstep_set.get(),
+            submission=submission,
+            data={"textField": "Misc", "invoiceAmount": 100},
```
viktorvanwijk: I would expect `textfield1` to be present here as well, because it wasn't hidden with clearOnHide enabled.

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +74,239 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        submission.form.apply_logic_analysis()
+
+        self.assertEqual(submission_response.status_code, status.HTTP_201_CREATED)
+
+        # 2. submit step data
+        submit_data_endpoint = reverse(
+            "api:submission-steps-detail",
+            kwargs={"submission_uuid": submission.uuid, "step_uuid": form_step.uuid},
+        )
+        submit_data = {"data": {"textField": "Misc", "invoiceAmount": 100}}
```
viktorvanwijk, line 276: Also here `textfield1` should be present

viktorvanwijk (commented):
Some questions :)

src/openforms/submissions/models/submission_value_variable.py
vaszig: In a single step form the logic actions will take place after the submission and only in the backend. 
So, the user cannot see the changes on the frontend (not possible at least with the current implementation). 

I couldn't think of another way of having the right data at this point. And in fact I do not think that this kind of form is meant for having such kind of behaviour. These actions are meant mostly to address backend logic rules that make sense only for the admins (I don't think the user needs to know anything about these details, for example building a specific url from the submission which will be available for internal use).

src/openforms/submissions/tests/form_logic/test_modify_variables.py
vaszig: They test the same thing but these test the real actions, the other ones test the whole flow of a single step form. If you think it's redundant I can remove them (but for me they test 2 different things, the functions that they process the logic rules and the other ones the the flow and the correct submitted and processed data).

src/openforms/submissions/tests/form_logic/test_modify_variables.py
vaszig: I should have a more clear text, I agree. This data is indeed submitted and I wanted to show the difference when a user fills in a field that is meant to grab its value from the logic (this can be considered though a bad configuration from the admin user I guess).

src/openforms/submissions/models/submission_value_variable.py
viktorvanwijk: > And in fact I do not think that this kind of form is meant for having such kind of behaviour. These actions are meant mostly to address backend logic rules that make sense only for the admins (I don't think the user needs to know anything about these details

But the current setup suggests otherwise, because this change is only relevant for component variables. User-defined variables should already follow the results from logic evaluation before saving I think. I'm trying to understand why it is not sufficient to rely on user-defined variables being updated, as opposed to user-submitted data (component variables) being (possibly) overwritten by logic?

src/openforms/submissions/tests/form_logic/test_modify_variables.py
viktorvanwijk: I don't think I would have added them, because there is no specific behaviour for single step forms in these logic actions, right? For me, the complete integration would be enough, but idk :shrug:

src/openforms/submissions/models/submission_value_variable.py
vaszig: Do you mean to give the opportunity to use the logic rules only for the user defined variables?If so, yes things would be easier and more straight forward

viktorvanwijk: You are probably more familiar with the use case of this single-step form, and whether or not they need this specific behaviour with updated components, but I would indeed think to limit it to user-defined variables. Which _should_ already work as expected

sergei-maertens: Agree with limiting this to user defined variables!

src/openforms/js/components/admin/forms/VariableMapping.js
```diff
@@ -219,6 +228,7 @@ const VariableMapping = ({
   propertySelectLabel,
   includeStaticVariables = false,
   rowCheck,
+  filter = allowAny,
```
viktorvanwijk: This default shouln't be necessary with the one already in `VariableSelection`, right, or am I missing something?

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,249 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+                    },
+                ]
+            },
+        )
+        FormVariableFactory.create(
+            form=form,
+            key="test",
+            source=FormVariableSources.user_defined,
+            data_type=FormVariableDataTypes.string,
+            form_definition=form.formstep_set.get().form_definition,
```
viktorvanwijk: A user-defined variable shouldn't have a reference to a form definition. Instead of using `source`, you could use the trait `user_defined=True`, which sets the source and form definition (to `None`) automatically :)

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,249 @@ def test_single_step_form_flow(self, m: MagicMock):
...
```
viktorvanwijk: See comment above

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,249 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        submission.form.apply_logic_analysis()
+
+        self.assertEqual(submission_response.status_code, status.HTTP_201_CREATED)
+
+        # 2. submit step data
+        submit_data_endpoint = reverse(
+            "api:submission-steps-detail",
+            kwargs={"submission_uuid": submission.uuid, "step_uuid": form_step.uuid},
+        )
+        submit_data = {"data": {"textField": "Misc", "invoiceAmount": 100, "test": ""}}
```
viktorvanwijk:
```suggestion
        submit_data = {"data": {"textField": "Misc", "invoiceAmount": 100}}
```

src/openforms/submissions/tests/test_single_step_form.py
vaszig: Oops..copy paste

viktorvanwijk (approved):

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+
+        self.assertEqual(complete_submission_response.status_code, status.HTTP_200_OK)
+
+        # 4. make sure the submission value variables are correctly updated
+        variables = submission.submissionvaluevariable_set.all()
+        expected = {"textField": "foo", "checkbox": True, "test": "foo"}
+
+        for key, expected_value in expected.items():
+            obj = next(o for o in variables if o.key == key)
+            self.assertEqual(obj.value, expected_value)
```
sergei-maertens:
```suggestion
        variables_data = state = submission.variables_state.get_data()
        expected = {"textField": "foo", "checkbox": True, "test": "foo"}
        
        self.assertEqual(variables_data, expected)
```
the for-loop is still an option, but then you can do direct key access instead of using `next`, but then you should also use `with self.subTest` so that on test failures you can quickly see which key/variable is problematic.

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+            rest_api_path="engine-rest/",
+            auth_header="Basic ZGVtbzpkZW1v",
+        )
+
+        patcher = patch(
+            "openforms.dmn.contrib.camunda.checks.CamundaConfig.get_solo",
+            return_value=config,
+        )
+        patcher.start()
+        self.addCleanup(patcher.stop)
```
sergei-maertens: The test class name says nothing about DMN/Camunda, so I don't find it logical to do the camunda setup in the `setUp` that's shared for all existing and future tests.

I'd move this into the test itself, and get rid of the mock too - you can set up a cleanup function that's much more targeted and explicit:

```py
self.addCleanup(CamundaConfig.clear_cache)
```

Alternatively, if you do have tests with a shared setup, the configuration can be done in `setUpTestData` class method, and then you can wire up:

```py
cls.addClassCleanup(CamundaConfig.clear_cache)
```
but my first suggestion is the best fit for this case.

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        )
+
+        form_url = reverse("api:form-detail", kwargs={"uuid_or_slug": form.uuid})
+
+        # 1. create the submission
+        submission_body = {
+            "form": f"http://testserver.com{form_url}",
+            "formUrl": "http://testserver.com/my-form",
+            "anonymous": True,
+            "initialDataReference": "of-or-3452fre3",
```
sergei-maertens: why do we have an initial data reference?

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        )
+
+        form_url = reverse("api:form-detail", kwargs={"uuid_or_slug": form.uuid})
+
+        # 1. create the submission
+        submission_body = {
+            "form": f"http://testserver.com{form_url}",
+            "formUrl": "http://testserver.com/my-form",
+            "anonymous": True,
+            "initialDataReference": "of-or-3452iou3",
```
sergei-maertens: why do we have an initial data reference?

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        }
+
+        submission_response = self.client.post(
+            endpoint, submission_body, HTTP_HOST="testserver.com"
```
sergei-maertens:
```suggestion
            endpoint, submission_body, headers={"Host": "testserver.com"}
```
is the modern variant and more readable :)

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        )
+        submission = Submission.objects.get()
+        form_step = form.formstep_set.get()
+
+        submission.form.apply_logic_analysis()
```
sergei-maertens: you can call `form.apply_logic_analysis()` *before* you make the submission create call, which is more realistic to what happens in reality - the form designer saves the form in the admin, and only then a submission is started.

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +79,247 @@ def test_single_step_form_flow(self, m: MagicMock):
...
+        variables = submission.submissionvaluevariable_set.all()
+        expected = {
+            "invoiceAmount": 100,
+            "textField": "Misc",
+            "test": "day-to-day expense",
+        }
+
+        for key, expected_value in expected.items():
+            obj = next(o for o in variables if o.key == key)
+            self.assertEqual(obj.value, expected_value)
```
sergei-maertens: The same remarks from the other test apply

sergei-maertens (changes requested):

vaszig:
@sergei-maertens I really don't get why the test for DMN is now failing (cannot reproduce this locally). The only thing that changed was `variables_data = submission.variables_state.get_data()` instead of `variables = submission.submissionvaluevariable_set.all()`, but I see that this is used in order to collect the data.

Do you have any idea?

sergei-maertens (approved):

PR #6412, opened 2026-06-26
Backport 3.4: Create digital address for betrokkene in Customer Interactions API

### LINKED ISSUE #6082: Use existing, non-default, email address in Partij in Open Klant and store this

annashamray:
Closes #6082 (backport 3.4)

### DIFF
--- a/docs/configuration/prefill/communication_preferences.rst
+++ b/docs/configuration/prefill/communication_preferences.rst
@@ -80,6 +80,30 @@ to add the relevant component and a **user defined** form variable:
     created in the "Configuration" section and the profile form variable.
 
 
+.. _configuration_prefill_communication_preferences_writing:
+
+Writing to the Customer Interactions API
+========================================
+
+The communication preferences prefill configuration is also used to update communication preferences in the
+Customer Interactions API. To enable it you need to check "Open Forms should update customer data after submission"
+in the "Profile" component configuration.
+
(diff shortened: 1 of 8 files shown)


sergei-maertens:
Is there also a backport for 3.5?

sergei-maertens (approved):

annashamray:
Yes, Backport for 3.5 is already merged, this one wasn't merged because CI doc build was failing, so I waited for it to resolve itself :)

PR #6413, opened 2026-06-26
Add check for file component catalogue consistency

### LINKED ISSUE #6304: Add check for `file` catalogue consistency
For the Objects API and ZGW APIs there can be inconsistencies that must be caught and rectified *before* the upgrade to 4.0 is done. This check script will thus need to be backported to 3.5 as part of our migration tools.

It's possible that file upload components end up with a different catalogue in the registration options than what's configured in a registration backend - or even the same file component may be configured with different backends that use different catalogues. We cannot guarantee that this will all still resolve.

The migration can copy the descriptions, and set up the catalogues in the backend options accordingly, but cannot solve merge conflicts. That's a manual action.

sergei-maertens:
Closes #6304 

[skip: e2e]

**Changes**

* Implemented a bin check script to detect mismatching catalogues
* Small refactor in existing tests for output capture

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/bin/report_file_component_inconsistent_catalogues.py
+++ b/bin/report_file_component_inconsistent_catalogues.py
@@ -1,0 +1,191 @@
+#!/usr/bin/env python
+#
+# Introspect file components used in forms with the ZGW or Objects API registration
+# plugin and report which file components resolved to a different catalogue than the
+# one in the registration plugin settings.
+#
+# This check is intended to run *after* the data migrations that copied all the
+# configuration to the individual registration backends. It cannot be run in a
+# preventive manner as upgrade check because these problems cannot be corrected pre-4.0.
+#
+from __future__ import annotations
+
+import sys
+from collections.abc import Iterator
+from dataclasses import dataclass
+from itertools import chain
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import django
+from django.db.models import Prefetch
+
+import click
+from tabulate import tabulate
+
+SRC_DIR = Path(__file__).parent.parent / "src"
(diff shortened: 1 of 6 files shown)


bin/report_file_component_inconsistent_catalogues.py
```diff
@@ -0,0 +1,191 @@
...
+    def _report_problems() -> Iterator[tuple[str, str, str, str, str]]:
+        form_id: int = 0
+        backend_name = ""
+
+        for problem in problems:
+            form_id_changed = problem.form_id != form_id
+            form_id = problem.form_id
+
+            backend_name_changed = problem.registration_backend_name != backend_name
+            backend_name = problem.registration_backend_name
```
viktorvanwijk, line 146: I'm not sure I understand why it's necessary to check the form id and backend name here?

viktorvanwijk (approved):

bin/report_file_component_inconsistent_catalogues.py
sergei-maertens: to visually structure the table data if there are multiple backends/components in a single form/backend! It declutters by not repeating the form/backend name :)

viktorvanwijk: Ah yeah, I see it now :)

sergei-maertens:
Backported to 3.5.x in 93f9b6bde3c199a00a1eb3fbb3e566b705e0cb13

PR #6424, opened 2026-07-01
[#6380] Update codebase regarding the `substr` json logic operator

### LINKED ISSUE #6380: Implement `substr` json logic operator
Currently not supported in the backend json logic, but it does work in the frontend.

**Tasks**

- [x] Implement operator (in our fork?)
- [x] Update shared test suite (`.github` repository/git-submodule)
- [x] Update comments/tests that remark it's not supported
- [x] Update documentation: https://open-forms.readthedocs.io/en/stable/manual/forms/logic.html#id1:~:text=of%20onwaar%20is.-,Note,substr,-Voorbeeld%20formulier

vaszig:
Closes #6380 partly

**Changes**

- [x] Added tests for the new operator (mainly for the single step form that will definitely make use of it)
- [x] Update the submodule for the shared tests
- [x] Pin maykin's `json-logic-py` fork to new version
- [x] Updated json logic introspection

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/tests/test_single_step_form.py
+++ b/src/openforms/submissions/tests/test_single_step_form.py
@@ -235,11 +235,10 @@ class SingleStepFormTests(APITestCase):
             "form": f"http://testserver.com{form_url}",
             "formUrl": "http://testserver.com/my-form",
             "anonymous": True,
-            "initialDataReference": "of-or-3452fre3",
         }
 
         submission_response = self.client.post(
-            endpoint, submission_body, HTTP_HOST="testserver.com"
+            endpoint, submission_body, headers={"Host": "testserver.com"}
         )
         submission = Submission.objects.get()
         form_step = form.formstep_set.get()
@@ -255,7 +254,7 @@ class SingleStepFormTests(APITestCase):
         )
         submit_data = {"data": {"textField": "foo"}}
         submit_data_response = self.client.put(
-            submit_data_endpoint, submit_data, HTTP_HOST="testserver.com"
+            submit_data_endpoint, submit_data, headers={"Host": "testserver.com"}
         )
 
         self.assertEqual(submit_data_response.status_code, status.HTTP_201_CREATED)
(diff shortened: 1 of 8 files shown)


src/openforms/utils/json_logic/descriptions.py
```diff
@@ -359,3 +360,42 @@ def op_rdelta(*args) -> str:
             raise ValueError(
                 "Unexpected amount of arguments. Expected 1-6 got {len(args)}: {args!r}"
             )
+
+
+@add_boilerplate()
+def op_substr(text, start, length=None):
```
viktorvanwijk, line 365: The point of the description is to describe the operation, not evaluate it, right?

This will not work anyway when `text` is a variable expression instead of a string.

I think returning 
```python
gettext("a substring of {text}").format(text=text)
```
is sufficient here tbh. This formats to "a substring of 'this is a string'" for a string (which would be a weird pattern anyway, because why not just put the string you want directly), and to "a substring of {{foo}}" for a variable expression. The start and length can then be inspected from the expression itself.

src/openforms/submissions/tests/test_single_step_form.py
```diff
@@ -71,3 +76,102 @@ def test_single_step_form_flow(self, m: MagicMock):
 
         # 4. make sure the prefill was not called
         m.assert_not_called()
+
+    def test_single_step_form_with_substr_and_variable_action(self):
```
viktorvanwijk, line 185: I think Sergei gave some feedback on this test flow in the other PR, could you also apply that here 😇 ? I'm seeing an initial data reference and `HTTP_HOST`, for example

src/openforms/utils/tests/test_api_logic_description.py
```diff
@@ -53,6 +55,8 @@ def test_valid_expressions_generate_descriptions(self):
             },
             # embedded _meta.description bits
             {"==": [1, 1], "_meta": {"description": "identity"}},
+            {"substr": ["test", 1, 1], "_meta": {"description": "identity"}},
```
sergei-maertens: Here the "identity" should be "substring", it's a human provided description text for the expression. TBH I think those two lines can also be omitted, the `_meta` behaviour is already covered by the existing test

sergei-maertens (approved):

src/openforms/utils/tests/test_api_logic_description.py
vaszig: I will update it, thanks!

PR #6267, opened 2026-05-08
Fix crash in logic rule when it contains a property action that refers to a non-existing component

viktorvanwijk:
Discovered while working on #6261 

[skip: e2e]

**Changes**

Added a question mark

**Checklist**

Check off the items that are completed or not relevant.

- Impact on features

  - [x] Checked copying a form
  - [x] Checked import/export of a form
  - [x] Config checks in the configuration overview admin page
  - [x] Checked new model fields are usable in the admin
  - [x] Problem detection in the admin email digest is handled

- Dockerfile/scripts

  - [x] Updated the Dockerfile with the necessary scripts from the `./bin` folder

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

- Documentation

  - [x] Added documentation which describes the changes

Issue #5977, opened 2026-02-16
You can set a fieldset as readonly with logic rules but this is not allowed for fieldsets


joeribekker:
### Product versie / Product version

3.4.0

### Omschrijf het probleem / Describe the bug

Fieldsets (when looking at the detail screen) don't have the option to be read only. However, with logic rules you can still do this.

Under the hood, this does not have the desired effect: Fields are not treated as read-only on the backend, allowing them to be modified.

### Screen resolution

None

### Device

None

### OS

None

PR #6428, opened 2026-07-02
[#6425] Add script for fixing disabled fieldsets via the logic rules

### LINKED ISSUE #6425: Fix script to apply "disable" logic on all items in a "layout" component (fieldset)

3.5.x

### Customer reference

_No response_

Since #5977 - the "disable/read-only" logic applied to Fieldsets is no longer working. The option is removed from OF and saving a form with this logic is no longer possible: 

Image

We would like forms with this exact logic to be "fixed" with a script. 
Where all components within the fieldset get added to the logic rule as "disable" actions.
resulting in forms that function the way they did before. But saving form builders a lot of manual "clicks" to convert this single disable action to many disable actions. 

beware, these fieldsets often contain fieldsets! 

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #6425

**Changes**

- Fieldsets cannot have the `disabled` logic so with this script we make sure all related forms will apply `disabled` logic in the components inside a fieldset.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


vaszig:
See if we want this in `stable 3.4.x`. Tested locally the version and the `disabled` property is shown for fieldsets.

bin/fix_disabled_components_inside_fieldset.py
annashamray, line 1: Could you please give +x permissions for the file, now I can't execute it

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
+# GH-5977 removed the ability to apply the "disable/read-only" logic to Fieldsets.
```
annashamray:
```suggestion
#!/usr/bin/env python
#
# GH-5977 removed the ability to apply the "disable/read-only" logic to Fieldsets.
```

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+    return children_actions
+
+
+def update_fieldset_logic_rules(dry_run: bool = False) -> bool:
+
+    from openforms.formio.service import iter_components
+    from openforms.formio.typing.vanilla import FieldsetComponent
+    from openforms.forms.models import Form, FormLogic
+
+    forms_to_check = Form.objects.filter(_is_deleted=False).prefetch_related(
```
annashamray:
```suggestion
    forms_to_check = Form.objects.filter(_is_deleted=False, formlogic__isnull=False).distinct().prefetch_related(
```

annashamray: This way you don't need to check if there are logic rules inside the loop (lines 55-57)

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+        "formlogic_set", "formstep_set"
```
annashamray: I don't think `formstep_set` prefetch is used later. because you add `select_related` to it. Perhaps it's better to include it here with Prefetch object

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+        component_map: dict[str, FieldsetComponent] = {}
+
+        form_steps = form.formstep_set.select_related("form_definition")
+        for form_step in form_steps:
+            for component in iter_components(
+                form_step.form_definition.configuration,
+                recursive=True,
+                recurse_into_editgrid=False,
+            ):
+                if component["type"] == "fieldset":
```
annashamray:
```suggestion
                if holds_submission_data(component):
```
despite the name of the ticket all layout components should be taken into account

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+
+        for rule in form.formlogic_set.iterator():
+            # the final logic actions we will have in the form
+            updated_logic_actions = []
+            for action in rule.actions:
+                # make sure we only get the fieldsets that have disabled property wrongly
+                # set
+                if (
+                    action["component"] in component_map
+                    and action["action"]["property"]["value"] == "disabled"
```
annashamray: Not all actions have "property" value, so it's better to use `get` here to avoid KeyError

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+                    component_map[component["key"]] = component
+
+        for rule in form.formlogic_set.iterator():
+            # the final logic actions we will have in the form
+            updated_logic_actions = []
+            for action in rule.actions:
+                # make sure we only get the fieldsets that have disabled property wrongly
+                # set
+                if (
+                    action["component"] in component_map
```
annashamray: Not all actions have "component" value, so it's better to use get here

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+                        )
+                    )
+                else:
+                    updated_logic_actions.append(action)
+
+            if dry_run:
+                continue
+
+            rule.actions = updated_logic_actions
+            rule.save()
```
annashamray: Now you always update actions for all the rules, I think you don't need to hit `save` when there is nothing to update

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+        form_steps = form.formstep_set.select_related("form_definition")
+        for form_step in form_steps:
+            for component in iter_components(
+                form_step.form_definition.configuration,
+                recursive=True,
+                recurse_into_editgrid=False,
+            ):
+                if component["type"] == "fieldset":
+                    component_map[component["key"]] = component
+
```
annashamray:
```suggestion
if not component_map:
    continue
```

If you don't have layout fields no need to update logic rules

bin/fix_disabled_components_inside_fieldset.py
annashamray:
```suggestion
            rule.save(update_fields=["actions"])
```

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+import click
+from tabulate import tabulate
+
+SRC_DIR = Path(__file__).parent.parent / "src"
+sys.path.insert(0, str(SRC_DIR.resolve()))
+
+
+def mark_children_as_disabled(component_configuration):
+    children_actions = []
+    if component_configuration["type"] == "fieldset":
```
annashamray: same comment = please include all components with `holds_submission_data` here

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+        for child_component in component_configuration.get("components", []):
+            children_actions.extend(mark_children_as_disabled(child_component))
+    else:
+        children_actions.append(
+            {
+                "action": {
+                    "type": "property",
+                    "state": True,
```
annashamray: I think the `state` value should be derived from the original logic rule action which you replace.

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+        # mappings of fieldset components (component_key: component_configuration)
+        component_map: dict[str, FieldsetComponent] = {}
+
+        form_steps = form.formstep_set.select_related("form_definition")
+        for form_step in form_steps:
+            for component in iter_components(
+                form_step.form_definition.configuration,
+                recursive=True,
+                recurse_into_editgrid=False,
+            ):
```
annashamray: Sorry I don't quite understand why not to use `form.iter_components` here?

bin/fix_disabled_components_inside_fieldset.py
```diff
@@ -0,0 +1,148 @@
...
+                    },
+                },
+                "component": component_configuration["key"],
+            }
+        )
+
+    return children_actions
+
+
+def update_fieldset_logic_rules(dry_run: bool = False) -> bool:
```
annashamray: I think having `transaction.atomic()` would be nice here

annashamray (changes requested):
I've tested it locally with a simple case of fieldset -> field
I'll test more with other component types in the re-review

annashamray:
@vaszig about backporting to 3.4 - actually I don't know

The original issue was not backported initially - https://github.com/open-formulieren/open-forms/issues/5977
But the bugfix for this issue was backported to 3.5 together with the original issue (https://github.com/open-formulieren/open-forms/pull/6267#issuecomment-4407509854) 

Maybe Sergei could help with this question

bin/fix_disabled_components_inside_fieldset.py
vaszig: Yes, agree!I copied this from another script and I didn't pay attention apparently.

bin/fix_disabled_components_inside_fieldset.py
vaszig: You mean `if not holds_submission_data(component):` right?
I need the layout components (that they do not have data).

bin/fix_disabled_components_inside_fieldset.py
vaszig: Yes, I will try this and then the prefetch query can be simpler since the `iter_components` does a `select_related` for the form definition.

bin/fix_disabled_components_inside_fieldset.py
annashamray: Yes you are right! layout components don't hold data, sorry for confusion

vaszig:
@annashamray I had to change the names because of your remarks (not only the fieldset). I do not though if you agree with my approach regarding which components should be tested. From the layout components I think that only Fieldsets and Columns make sense (tested both locally). Let me know if you agree.

bin/fix_disabled_components_inside_fieldset.py
annashamray: I don't think this one is fixed: 

```
                if updated_logic_actions:
                    rule.actions = updated_logic_actions
                    rule.save(update_fields=["actions"])
```
You can have `updated_logic_actions` filled in with old actions (from the `else:` block)

I'd add another variable (for example `should_update: bool`) which would be `True` only if you extend `updated_logic_actions`

bin/fix_disabled_logic_in_layout_components.py
```diff
@@ -0,0 +1,169 @@
...
+            children = component_configuration.get("components", [])
+
+        elif component_type == "columns":
+            children = [
+                child
+                for column in component_configuration["columns"]
+                for child in column.get("components")
+            ]
+
+        for child in children:
```
annashamray, line 52: `children` variable can possible be unbound and crash the code. 
I agree with you that only `columns` and `fieldset` components should be fixed, but in case there is a strange form let's make sure it's not crashed:

* could you declare the variable `children` above
* if it's not of the type that we process we can show a warning saying - please change this form manually (nice to have)

annashamray (commented):
I've tested it with
* fieldset -> field
* column -> field
* fieldset -> column -> field

It all works as expected. Just small remark and then let's merge it

annashamray (approved):
Please make CI happy
And then it's good to go!

vaszig:
> Please make CI happy And then it's good to go!

I can't make CI happier..only docs are failing now.

annashamray:
It's green now! Good old rerun helped :)

sergei-maertens (approved):

PR #6439, opened 2026-07-06
Clean up submission file attachment handling

### LINKED ISSUE #2713: Improve the code that creates submission attachments from files in submission data
### Thema / Theme

API

The bandaid fix for this issue https://github.com/open-formulieren/open-forms/issues/2699 showed that the code for creating submission attachments by looking through the submission data and extracting any file is quite fragile.

This should be refactored into a more robust solution.

_No response_

_No response_

sergei-maertens:
Closes #2713

**Changes**

* Update data model to capture the reality & remove nullability of required fields
* Inventarized and documented which metadata is actually necessary
* Stripped down upload handling code to the minimum required metadata
* Updated tests
* Updated registration backends that process the file components
* Deleted dead/obsoleted code
* Added more integrity checks (mostly targeting the tests in CI)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/submissions/migrations/0017_convert_attachment_metadata.py
sergei-maertens, line 1: TODO: need to profile the performance of this migration!

src/openforms/submissions/models/submission_files.py
```diff
@@ -231,24 +306,30 @@ def content_hash(self) -> str:
...
-        return component
-
-    @property
-    def form_key(self):
-        if self.submission_variable:
-            return self.submission_variable.key
-        raise Exception("tried to access an attachment without .submission_variable")
-
-    form_key.fget.short_description = _("form component key")
+        assert self._data_path, "TODO: replace with check constraint"
```
SonnyBA: The TODO can be removed now I guess? Now that the check constraint is in place.

src/openforms/submissions/models/submission_files.py
```diff
@@ -175,12 +230,28 @@ class SubmissionFileAttachment(DeleteFileFieldFilesMixin, models.Model):
...
-        blank=True,
+    )
+    component_key = models.TextField(
+        _("component key"),
+        help_text=_(
+            "Key of the file component for which the upload was made. Most of the time "
+            "this will be identical to the submission_variable.key, except when the "
+            "file component is inside an edit grid - the variable then points to the "
+            "edit grid."
+        ),
```
SonnyBA, line 254: Can we use a property, that uses the `submission_variable.key` unless the submission's form variable subtype is that of an editgrid,  to lookup this value instead of this field? I think it would simplify the usage of the component key, without having to take component configurations into consideration when accessing the value (e.g this would be handled through the property).

src/openforms/submissions/migrations/0016_rename__component_data_path_submissionfileattachment__data_path_and_more.py
SonnyBA, line 1: Should we squash this migration?

src/openforms/submissions/migrations/0017_convert_attachment_metadata.py
```diff
@@ -0,0 +1,107 @@
...
+            #    component has been removed or renamed in the Form, this snapshot still
+            #    has the correct component key.
+            # 2. For file components in edit grids, the variable contains the outer
+            #    editgrid component definition. We must then use the legacy
+            #    ``_component_configuration_path`` and look that up in the attached
+            #    submission step -> form step -> form definition configuration (that's
+            #    the mechanism that is now used at runtime). Looking up the
+            #    configuration path in the snapshotted editgrid definition is not
+            #    reliable, as the editgrid itself may be inside a fieldset or columns
+            #    layout component, for which no form variables are created becaues they
```
SonnyBA:
```suggestion
            #    layout component, for which no form variables are created because they
```

src/openforms/submissions/migrations/0017_convert_attachment_metadata.py
```diff
@@ -0,0 +1,107 @@
...
+            else:
+                config_path = attachment._component_configuration_path
+                formio_configuration = (
+                    attachment.submission_step.form_step.form_definition.configuration
+                )
+                configuration = glom(formio_configuration, config_path, default=None)
+                attachment.component_key = (
+                    configuration["key"] if configuration is not None else "<UNKNOWN>"
+                )
+
```
SonnyBA, line 88: Should we add some logging here when the `<UNKNOWN>` situation occurs? I'm not sure what the impact this has when the `component_key` is unknown.

src/openforms/submissions/tests/factories.py
```diff
@@ -375,12 +370,27 @@ def content_type(self) -> str:
...
+            lambda o: o.factory_parent.submission_step.submission
+        ),
+        form_variable__data_type=FormVariableDataTypes.array,
+        form_variable__data_subtype=FormVariableDataTypes.object,
+    )
+
+    # these are not 100% correct by default, especially when using edit grids, but
+    # they're required (non-blank) fields
+    _data_path = factory.LazyAttribute(lambda o: o.submission_variable.key)
+    component_key = factory.LazyAttribute(lambda o: o.submission_variable.key)
```
SonnyBA, line 392: Can we use the `create` method to make this correct?

src/openforms/submissions/attachments.py
```diff
@@ -252,69 +249,161 @@ def cleanup_submission_temporary_uploaded_files(submission: Submission):
...
+        case {"type": "file"}:
+            component = cast(FileComponent, component)  # can't assert with TypedDict
+            files: FileValue = data.get(data_path, [])
+            uploads: Sequence[TemporaryFileUpload] = [
+                upload
+                for file in files
+                if (upload := temporary_upload_from_url(file.get("url"))) is not None
+            ]
+            num_uploads = len(uploads)
+            for i, upload in enumerate(uploads, 1):
```
SonnyBA:
```suggestion
            for i, upload in enumerate(uploads, start=1):
```

src/openforms/submissions/models/submission_files.py
sergei-maertens: I considered it, but then in certain database queries that have now been cleaned up you again need to use some `Coalesce` nonsense. I favored this approach because the actual model instances are populated through the manager at all times (and the factories) and this should otherwise be low-maintenance code - it had been untouched for a long time before this cleanup already :)

src/openforms/submissions/migrations/0016_rename__component_data_path_submissionfileattachment__data_path_and_more.py
sergei-maertens: Discussed in the office - some schema migrations can be merged indeed with some manual actions, but then we lose the context of the individual commits. The data migration that's included is likely to cause "pending trigger" issues.

src/openforms/submissions/migrations/0017_convert_attachment_metadata.py
sergei-maertens: so these situations are currently already broken anyway, because the old situation would try to look up the component via the configuration path. Effectively, nothing changes.

Adding logging doesn't add any benefits here IMO - it's in migrations, nobody is going to look at those logs, and you can't act on them either - trying to update this in the admin is not possible because the fields aren't exposed, and we can't reconstruct anymore where the data is because the form has changed.

It fits in the bigger story of form version snapshotting, which is a completely different beast :)

src/openforms/submissions/tests/factories.py
sergei-maertens: I tried, it can't be done unfortunately :( I tried a lot of things using factory boy internals and everything was worse. However, it will also typically lead to inconsistent variable setups, which do fail loudly and force good test setups (that's how I found the broken tests).

So I considered this a "Good Enough" approach.

SonnyBA (approved):
I think the typo in the data migration is still there but the rest looks good 👍

sergei-maertens:
> I think the typo in the data migration is still there but the rest looks good 👍 

I missed that remark, will fix!

PR #6441, opened 2026-07-07
Fix dmn logic with unsaved form variables

sergei-maertens:
Regression following #6275
Unblocks #6407 

**Changes**

* Fix crash when DMN logic uses not-yet-existing form variables as input
* Update docs & detection script to report logic rules that use possibly hidden form variables as input

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/logic/actions.py
+++ b/src/openforms/submissions/logic/actions.py
@@ -707,6 +707,15 @@ class EvaluateDMNAction(ActionOperation):
         *,
         data_for_visible_state: FormioData,
     ) -> DataMapping | None:
+        # evaluation may be triggered/requested when not all input variables are
+        # available yet, in particular when a submission is being started. Normal
+        # (step-level) evaluation should not run into these situations, unless something
+        # is broken with the logic dependency graph building.
+        # If not all inputs are available yet, we abort early without doing anything,
+        # rather than guessing some kind of magic fallback value.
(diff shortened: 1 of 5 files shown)


vaszig (approved):

src/openforms/submissions/logic/actions.py
```diff
@@ -707,6 +707,15 @@ def eval(
         *,
         data_for_visible_state: FormioData,
     ) -> DataMapping | None:
+        # evaluation may be triggered/requested when not all input variables are
+        # available yet, in particular when a submission is being started. Normal
```
viktorvanwijk, line 711: Probably not likely, but I guess it could run into this situation when inputs are from multiple steps, and this rule is executed on multiple steps as well.

viktorvanwijk (approved):

src/openforms/submissions/logic/actions.py
sergei-maertens: yeah - I did consider it but then I'd expect them already to have a check set up in the trigger to prevent it from running too soon?

viktorvanwijk: That would make sense, yeah

viktorvanwijk:
Backport check script update to 3.5.x: f50a3c29

PR #6443, opened 2026-07-09
Add default value input for logic rule trigger

### LINKED ISSUE #6391: Provide easy way to add default value in a JSON logic variable expression

viktorvanwijk:
Closes #6391

**Changes**

See title :)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
+++ b/src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
@@ -1,0 +1,70 @@
+import {useState} from 'react';
+import {FormattedMessage, useIntl} from 'react-intl';
+
+import {FAIcon} from 'components/admin/icons';
+
+import LiteralValueInput, {CheckboxChoices} from './LiteralValueInput';
+
+// Component-specific overrides
+const COMPONENT_TYPE_TO_INPUT_TYPE = {
+  // The selectboxes is of type object, so a JSON widget would be used. However, we only need to
+  // provide the possibility to select true/false.
+  selectboxes: CheckboxChoices,
+};
+
+const DefaultValueSelection = ({name, dataType, componentType, value, onChange}) => {
+  // If the value is not undefined, it means it was already specified, so we should show the input
+  // field.
+  const [showInputField, setShowInputField] = useState(value !== undefined);
+  const intl = useIntl();
+
+  const onRemoveDefaultValue = () => {
+    onChange({
+      target: {
+        name: name,
+        value: undefined,
+      },
+    });
+    setShowInputField(false);
+  };
+
+  return showInputField ? (
+    <div className="default-value-input-container">
+      <FormattedMessage description="Default value input label" defaultMessage="Default value:" />
+      <LiteralValueInput
+        name={name}
+        type={dataType}
+        componentType={componentType}
+        inputTypeOverrides={COMPONENT_TYPE_TO_INPUT_TYPE}
+        value={value}
+        onChange={onChange}
+      />
+      <FAIcon
+        icon="xmark"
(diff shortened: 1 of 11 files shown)


src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
```diff
@@ -0,0 +1,81 @@
+import React, {useState} from 'react';
```
sergei-maertens:
```suggestion
import {useState} from 'react';
```
the import shouldn't be necessary

src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+    });
+    setShowInputField(false);
+  };
+
+  return showInputField ? (
+    <div className="default-value-input-container">
+      <FormattedMessage description="Default value input label" defaultMessage="Default value:" />
+      <LiteralValueInput
+        id="default_value_input"
```
sergei-maertens: I think you can add multiple rules and click the buttons for each of them, leading to multiple identical IDs being present in the DOM which is not allowed.

Do we need the ID in the first place? Probably just for accessibility, in which case you can use React's `useId()` hook and use `aria-labelledby` on the input element, pointing to the label element on L43

src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
```diff
@@ -0,0 +1,81 @@
...
+        name={name}
+        type={dataType}
+        componentType={componentType}
+        inputTypeOverrides={COMPONENT_TYPE_TO_INPUT_TYPE}
+        value={value}
+        onChange={onChange}
+      />
+      <FAIcon
+        icon="xmark"
+        extraClassname="icon actions__action"
```
sergei-maertens, line 54: technically invalid BEM usage with the `actions__action` without `actions` parent, but I can pretend I didn't see it?

sergei-maertens (approved):

src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
viktorvanwijk: Ah yeah, I knew that isn't allowed as well. Given that it's not very accessible at the moment anyway, will remove it

src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
viktorvanwijk: Copied from the wrong example then 😅 

> but I can pretend I didn't see it?

Can you? :)

src/openforms/js/components/admin/form_design/logic/DefaultValueSelection.js
viktorvanwijk: Auto-import doing unnecessary stuff :(

viktorvanwijk:
Backport to 3.5.x: 6f3c6abf

PR #6448, opened 2026-07-13
Fix broken conditional test inside edit grids

### LINKED ISSUE #6447: `component.conditional` evaluation during input validation is not always right

main (92e3015c87613346c141073538740d144b265df5), 3.5.x, 3.4.x

### Customer reference

_No response_

When a component with conditional is defined in an edit grid, the reference to the component in editgrid items does not resolve causing our backend code to fall back to an "unknown" component. This results in component-specific cmparisons not being called, which we need for `selectboxes` and `file` components:

https://github.com/open-formulieren/open-forms/blob/92e3015c87613346c141073538740d144b265df5/src/openforms/formio/visibility.py#L70

This happens because for edit grid items the serializer is built for every item with a locally scoped formio configuration wrapper that doesn't have the `editgrid.field` keys for lookup: https://github.com/open-formulieren/open-forms/blob/92e3015c87613346c141073538740d144b265df5/src/openforms/formio/serializers.py#L53

I stumbled on this in the msgspec branch where the fallback component is removed.

Regression test in linked PR:

```py
    @tag("gh-6447")
    def test_conditional_resolves_to_component_specific_test(self):
        component: EditGridComponent = {
            "type": "editgrid",
            "key": "editgrid",
            "label": "Edit grid with nested conditional",
            "groupLabel": "item",
            "components": [
                # uses selectboxes because the value comparison is special and requires
                # the registry to work
                {
                    "type": "selectboxes",
                    "key": "selectboxes",
                    "label": "Selectboxes 1",
                    "values": [  # type: ignore
                        {"value": "a", "label": "A"},
                        {"value": "b", "label": "B"},
                    ],
                },
                {
                    "type": "textfield",
                    "key": "textfield2",
                    "label": "text field 2",
                    "validate": {"required": True},
                    "conditional": {  # type: ignore
                        "eq": "a",
                        "show": True,
                        "when": "editgrid.selectboxes",
                    },
                },
            ],
        }

        invalid_data: JSONValue = {
            "editgrid": [{"selectboxes": {"a": True, "b": False}}]
        }

        is_valid, errors = validate_formio_data(component, invalid_data)

        self.assertFalse(is_valid)
        error = extract_error(errors["editgrid"][0], "textfield2")
        self.assertEqual(error.code, "required")
```

The component is properly resolved and the conditional test properly uses the registry:

https://github.com/open-formulieren/open-forms/blob/92e3015c87613346c141073538740d144b265df5/src/openforms/formio/visibility.py#L82

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #6447

**Changes**

The root cause is that the component reference 'editgrid.selectboxes' is not found in the formio configuration wrapper, because a local configuration wrapper is used that's missing the context of being in an editgrid.

The visibility code then falls back to an 'unknown' component, which works for the simple direct comparison cases, but falls apart for selectboxes and file components (and any component that needs a custom 'test_conditional' implementation).

- [x] Added regression test
- [x] Implemented fix

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/tests/validation/test_editgrid.py
+++ b/src/openforms/formio/tests/validation/test_editgrid.py
@@ -156,7 +156,10 @@ class EditGridValidationTests(SimpleTestCase):
         data: JSONObject = {
             "toplevel": [{"nested": "i am valid"}],
         }
-        context = {"submission": SubmissionFactory.build()}
+        context = {
+            "submission": SubmissionFactory.build(),
+            "configuration": {"components": [component]},
+        }
 
         serializer = build_serializer(
             components=[component], data=data, context=context
@@ -307,7 +310,10 @@ class EditGridValidationTests(SimpleTestCase):
                 {"naamWerkgever": "ABC", "nettoLoon": 5, "periodeNettoLoon": "maand"}
             ],
         }
-        context = {"submission": SubmissionFactory.build()}
+        context = {
+            "submission": SubmissionFactory.build(),
+            "configuration": {"components": components},
+        }
         serializer = build_serializer(components=components, data=data, context=context)
 
         is_valid = serializer.is_valid()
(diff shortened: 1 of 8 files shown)


src/openforms/formio/tests/validation/test_editgrid.py
```diff
@@ -418,25 +424,85 @@ def test_conditional_hidden_inside_editgrid_with_reference_to_outside_editgrid(
 
             self.assertTrue(is_valid)
 
+    @tag("gh-6447")
+    def test_conditional_resolves_to_component_specific_test(self):
```
viktorvanwijk, line 428: Perhaps it's better to test the bug fix with the submission validate call to hit the complete code path, instead of testing it with the custom `validate_formio_data` helper that is used here?

viktorvanwijk (approved):
Looks OK to me, just a minor suggestion

src/openforms/formio/tests/validation/test_editgrid.py
sergei-maertens: Discussed in the office - the `assert` statements were added to unify the calls between real endpoint code vs. testing helpers, and they revealed a couple of places where this was missed :) leaving this as-is

sergei-maertens:
Backports:

* stable/3.5.x: 2fd12bea157e9c07113fdb4597c66d5ff768d3a8
* stable/3.4.x: 39d05720f9b379264290f73f14b3b5cc2aeb2f01

PR #6461, opened 2026-07-15
Fix type error when session data contains `None`

sergei-maertens:
Looks like there's a possibility that the session data actually contains null/None rather than not exist at all for a given key, which causes runtime `TypeError` when trying to modify the session key.

Seen via Sentry error monitoring, easy fix with a different default fallback mechanism.

[skip: e2e]


vaszig (approved):

sergei-maertens:
Backports:

* stable/3.5.x: d275ba0655ad778c2cd8291ad16aa9303df9d655
* stable/3.4.x: 83ea5042e452e431f1dff79327387e9b95067ca9

PR #6481, opened 2026-07-23
Implementing form export options

### LINKED ISSUE #6429: Export options
### Thema / Theme

Admin

To allow formbuilders/users to control how form data is exported, we are introducing export options.

When creating a form export, a modal menu is shown containing various options to control the exporting proces. This includes:
* Automatically anonymizing sensitive data
* Controlling which data will and won't be included in the exportdata
* Controlling which additional data (like theme, category, product, etc.) will be added to the exportdata

These export options will be presented as shown here:

Image

## Automatically anonymizing sensitive data
This should (at least) remove the internal remarks on the form, and all email addresses assigned to the E-mail registration backend (this includes the email addresses for submissions and payment updates, and the email addresses fetched from variables (this includes all places that could set the variable value, like logic rules)).

Ideally this should work based on some model/serializer configuration, that ensures that new features/attributes (that contain sensitive data) are handled as well.

These export options aim to simplify the export proces for form builders, by automating certain actions and allowing the complete form configuration to be exported.

Using these options the municipality and environment specific data can be stripped away, and additional supporting information can be included

_No response_

robinmolen:
Closes #6429

**Changes**

This PR introduces form export options, which are used to provide users control over how a form is exported from Open Forms. These export options can be used to exclude configuration with environment-specific values (like services), include additional configuration (product, tile layers and yivi attributes), and automatically anonymize the form configuration.

### Additional configuration

Product, WMS- and WMTS tile layers and Yivi attribute groups can be exported alongside the main form configuration. These additional pieces of information will be included to the exportdata as separate `.JSON` files.

Theme and category are deliberately not (yet) included in this.
Adding import/export support for theme's would create an easy way to include harmful SVG content to the exportdata.
To add import/export support for categories we need a proper plan on how to deal with the category tree structure when importing a category. Situations where a matching instance is found but the parent category doesn't exist or the instance has a different parent then is expected are things we will have to deal with. Also, in cases were both the instance and parent don't exists, do we then only create the category, or also the parent?

### Anonymize form configuration

The form configuration is anonymized using an export option in combination with a opt-in allowlist. Each exportSerializer must specify which fields don't contain sensitive information. When the 'do anonymize' export option is chosen, only those fields will remain in the exportdata. Otherwise, when the 'don't anonymize' option is chosen, all serializer fields will be included.

The exportSerializers can additionally overrule the `remove_sensitive_content` method, and provide additional anonymize logic. This should only be done for more complex/unconventional anonymization needs.

### Export options UI

User interfaces have been made for singular and bulk export actions. With singular export (exporting from the form detail/edit page) the export options UI only shows the export options that will have effect on that form. For example, the option to include the product data is only visible when the form has a product configured.

**Checklist**

- Impact on features

  - [ ] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/docs/manual/forms/export_import.rst
+++ b/docs/manual/forms/export_import.rst
@@ -16,7 +16,7 @@ U kunt elk formulier eenvoudig exporteren door onderstaande stappen te volgen:
 1. Navigeer naar **Formulieren** > **Formulieren**.
 2. Klik op de titel van het gewenste formulier om het formulier te openen.
 3. Klik onderaan op de knop **Exporteren**
-4. Een ZIP-bestand, met de naam van het *URL-deel* van het formulier, wordt nu 
+4. Een ZIP-bestand, met de naam van het *URL-deel* van het formulier, wordt nu
    gedownload op uw computer.
 
 U hoeft het ZIP-bestand niet uit te pakken.
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
 
     **Technische achtergrond**
 
-    Formulieren worden geëxporteerd als een ZIP-bestand waarin meerdere 
-    JSON-bestanden zitten. Elk JSON-bestand bevat de configuratie van het 
+    Formulieren worden geëxporteerd als een ZIP-bestand waarin meerdere
+    JSON-bestanden zitten. Elk JSON-bestand bevat de configuratie van het
     formulier zelf of een stap binnen het formulier. De JSON-structuur komt
     1-op-1 overeen met de API-specificatie waardoor het export formaat tevens
     open en transparent is.
 
+Formulier exporteer opties
+--------------------------
+
+Bij het exporteren van een formulier heb je enkele opties over hoe het
+ZIP-bestand wordt gemaakt:
+
+* **Anonimiseer formulierinstellingen**: Indien aangevinkt, dan worden de
+  formulierinstellingen automatisch geanonimiseerd. Dit verwijderd alle
+  e-mailadressen die zijn configureerd in de E-mail registratie en de interne
+  opmerkingen op het formulier. Deze optie is standaard aangevinkt.
+* **Formulierinstellingen**: Hiermee kunnen registratie backends,
+  betaalprovider, prefill en inlogmethode eenvoudig bij het exporteren
+  weggelaten worden. Standaard worden alle instellingen mee geëxporteerd.
+* **Aanvullende formulierinstellingen**: Hiermee kunnen aanvullende
+  instellingen van het formulier aan het ZIP-bestand toegevoegd worden.
 
 Formulieren importeren
 ======================
 
-Formulieren die zijn geëxporteerd met Open Formulieren kunnen ook weer 
+Formulieren die zijn geëxporteerd met Open Formulieren kunnen ook weer
 geïmporteerd worden.
 
 1. Navigeer naar **Formulieren** > **Formulieren**.
@@ -44,25 +59,25 @@ geïmporteerd worden.
 4. Klik op **Importeren**
 
 Als het goed is, is het formulier inclusief alle stappen geïmporteerd. Het
-geïmporteerde formulier is standaard niet actief en dus niet direct voor de 
+geïmporteerde formulier is standaard niet actief en dus niet direct voor de
 buitenwereld toegankelijk.
 
 Bijzonderheden
 --------------
 
-Als een formulier wordt geïmporteerd en het bevat een stap waarvan het 
+Als een formulier wordt geïmporteerd en het bevat een stap waarvan het
 *URL-deel* overeenkomt met een bestaande stap, dan controleert het import proces
-of de te importeren stap **exact hetzelfde** is als de bestaande stap. Als dat 
+of de te importeren stap **exact hetzelfde** is als de bestaande stap. Als dat
 zo is, dan wordt de bestaande stap gebruikt voor het geïmporteerde formulier. Er
 wordt dan geen nieuwe stap aangemaakt.
 
 Als de stap **niet exact hetzelfde** is, dan wordt er een nieuwe stap aangemaakt
-en wordt het *URL-deel* van de stap gewijzigd. Bijvoorbeeld van 
-``persoonsgegevens`` naar ``persoonsgegevens-2``. U krijgt hier altijd een 
+en wordt het *URL-deel* van de stap gewijzigd. Bijvoorbeeld van
+``persoonsgegevens`` naar ``persoonsgegevens-2``. U krijgt hier altijd een
 melding van.
 
-Als het *URL-deel* van een te importeren formulier zelf al bestaat, dan kan het 
-formulier niet worden geïmporteerd. U kunt dan het *URL-deel* van het 
+Als het *URL-deel* van een te importeren formulier zelf al bestaat, dan kan het
+formulier niet worden geïmporteerd. U kunt dan het *URL-deel* van het
 conflicterende formulier zelf aanpassen en het nogmaals proberen.
 
 Export en import van :ref:`logica <manual_logic>` regels die gebruik maken van
--- a/src/openforms/conf/locale/nl/LC_MESSAGES/django.po
+++ b/src/openforms/conf/locale/nl/LC_MESSAGES/django.po
@@ -4,8 +4,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Open Forms\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-07-23 13:00+0200\n"
-"PO-Revision-Date: 2026-07-22 14:23+0200\n"
+"POT-Creation-Date: 2026-08-02 13:12+0200\n"
+"PO-Revision-Date: 2026-08-02 13:17+0200\n"
 "Last-Translator:   <admin@admin.com>\n"
 "Language-Team: Dutch <support@maykinmedia.nl>\n"
 "Language: nl\n"
@@ -28,12 +28,13 @@ msgid ""
 "omit this header without losing functionality. We recommend favouring the "
 "nonce mechanism though."
 msgstr ""
-"De waarde van de CSP-nonce gegenereerd door de pagina die de SDK insluit/"
-"embed. Als deze meegestuurd wordt, dan worden velden met rich tekst (uit "
-"\"WYSIWYG\" editors) verwerkt om inline-styles toe te laten met de gegeven "
-"nonce. Indien de insluitende pagina een `style-src` policy heeft die `unsafe-"
-"inline` bevat, dan kan je deze header weglaten zonder functionaliteit te "
-"verliezen. We raden echter aan om het nonce-mechanisme te gebruiken."
+"De waarde van de CSP-nonce gegenereerd door de pagina die de SDK "
+"insluit/embed. Als deze meegestuurd wordt, dan worden velden met rich tekst "
+"(uit \"WYSIWYG\" editors) verwerkt om inline-styles toe te laten met de "
+"gegeven nonce. Indien de insluitende pagina een `style-src` policy heeft die"
+" `unsafe-inline` bevat, dan kan je deze header weglaten zonder "
+"functionaliteit te verliezen. We raden echter aan om het nonce-mechanisme te"
+" gebruiken."
 
 #: openforms/accounts/admin.py:38
 msgid "Groups"
@@ -86,7 +87,7 @@ msgstr "stafstatus"
 msgid "Designates whether the user can log into this admin site."
 msgstr "Bepaalt of de gebruiker zich op deze beheerwebsite kan aanmelden."
 
-#: openforms/accounts/models.py:55 openforms/forms/models/form.py:311
+#: openforms/accounts/models.py:55 openforms/forms/models/form.py:312
 msgid "active"
 msgstr "actief"
 
@@ -114,8 +115,8 @@ msgstr ""
 "De voorkeurstaal voor de (beheer) gebruikersinterface. Wanneer geen waarde "
 "ingesteld is, worden de browserinstellingen gebruikt."
 
-#: openforms/accounts/models.py:85 openforms/forms/models/form.py:841
-#: openforms/forms/models/form_version.py:64 soap/models.py:53
+#: openforms/accounts/models.py:85 openforms/forms/models/form.py:854
+#: openforms/forms/models/form_version.py:86 soap/models.py:53
 msgid "user"
 msgstr "gebruiker"
 
@@ -161,10 +162,8 @@ msgstr ""
 "verzoekenlogging in voor meer gedetailleerde informatie."
 
 #: openforms/analytics_tools/admin.py:15
-#, fuzzy
-#| msgid "General Email settings"
 msgid "General settings"
-msgstr "Algemene e-mailinstellingen"
+msgstr "Algemene instellingen"
 
 #: openforms/analytics_tools/admin.py:25
 msgid "Analytics: Google"
@@ -227,8 +226,8 @@ msgstr "Google Analytics code"
 msgid ""
 "Typically looks like 'UA-XXXXX-Y'. Supplying this installs Google Analytics."
 msgstr ""
-"Heeft typisch het formaat 'UA-XXXXX-Y'. Schakelt Google Analytics in als een "
-"waarde ingevuld is."
+"Heeft typisch het formaat 'UA-XXXXX-Y'. Schakelt Google Analytics in als een"
+" waarde ingevuld is."
 
 #: openforms/analytics_tools/models.py:145
 msgid "enable google analytics"
@@ -245,7 +244,8 @@ msgstr "Matomo server-URL"
 #: openforms/analytics_tools/models.py:156
 msgid "The base URL of your Matomo server, e.g. 'https://matomo.example.com'."
 msgstr ""
-"De basis-URL van uw Matomo server, bijvoorbeeld 'https://matomo.example.com'."
+"De basis-URL van uw Matomo server, bijvoorbeeld "
+"'https://matomo.example.com'."
 
 #: openforms/analytics_tools/models.py:160
 msgid "Matomo site ID"
@@ -306,8 +306,8 @@ msgstr "Piwik PRO site-ID"
 
 #: openforms/analytics_tools/models.py:206
 msgid ""
-"The 'idsite' of the website you're tracking in Piwik PRO. https://"
-"help.piwik.pro/support/questions/find-website-id/"
+"The 'idsite' of the website you're tracking in Piwik PRO. "
+"https://help.piwik.pro/support/questions/find-website-id/"
 msgstr ""
 "De 'idsite'-waarde van de website die je analyseert met Piwik PRO. Zie ook "
 "https://help.piwik.pro/support/questions/find-website-id/"
@@ -335,12 +335,12 @@ msgstr "SiteImprove ID"
 #: openforms/analytics_tools/models.py:226
 msgid ""
 "Your SiteImprove ID - you can find this from the embed snippet example, "
-"which should contain a URL like '//siteimproveanalytics.com/js/"
-"siteanalyze_XXXXX.js'. The XXXXX is your ID."
+"which should contain a URL like "
+"'//siteimproveanalytics.com/js/siteanalyze_XXXXX.js'. The XXXXX is your ID."
 msgstr ""
 "Uw SiteImprove ID - deze waarde kan je terugvinden in het embed-voorbeeld. "
-"Deze bevat normaal een URL zoals '//siteimproveanalytics.com/js/"
-"siteanalyze_XXXXX.js'. De XXXXX is uw ID."
+"Deze bevat normaal een URL zoals "
+"'//siteimproveanalytics.com/js/siteanalyze_XXXXX.js'. De XXXXX is uw ID."
 
 #: openforms/analytics_tools/models.py:232
 msgid "enable siteImprove analytics"
@@ -398,8 +398,8 @@ msgstr "GovMetric secure GUID (inzending voltooid)"
 
 #: openforms/analytics_tools/models.py:269
 msgid ""
-"Your GovMetric secure GUID for when a form is finished - This is an optional "
-"value. It is created by KLANTINFOCUS when a list  of questions is created. "
+"Your GovMetric secure GUID for when a form is finished - This is an optional"
+" value. It is created by KLANTINFOCUS when a list  of questions is created. "
 "It is a string that is unique per set of questions."
 msgstr ""
 "Het GovMetric secure GUID voor voltooide inzendingen. Deze waarde is niet "
@@ -428,8 +428,8 @@ msgid ""
 "The name of your organization as registered in Expoints. This is used to "
 "construct the URL to communicate with Expoints."
 msgstr ""
-"De naam/het label van de organisatie zoals deze bij Expoints bekend is. Deze "
-"waarde wordt gebruikt om de URL op te bouwen om met Expoints te verbinden."
+"De naam/het label van de organisatie zoals deze bij Expoints bekend is. Deze"
+" waarde wordt gebruikt om de URL op te bouwen om met Expoints te verbinden."
 
 #: openforms/analytics_tools/models.py:293
 msgid "Expoints configuration identifier"
@@ -440,8 +440,8 @@ msgid ""
 "The UUID used to retrieve the configuration from Expoints to initialize the "
 "client satisfaction survey."
 msgstr ""
-"Het UUID van de configuratie in Expoints om het tevredenheidsonderzoek op te "
-"halen."
+"Het UUID van de configuratie in Expoints om het tevredenheidsonderzoek op te"
+" halen."
 
 #: openforms/analytics_tools/models.py:301
 msgid "use Expoints test mode"
@@ -449,8 +449,8 @@ msgstr "gebruik Expoints testmode"
 
 #: openforms/analytics_tools/models.py:304
 msgid ""
-"Indicates whether or not the test mode should be enabled. If enabled, filled "
-"out surveys won't actually be sent, to avoid cluttering Expoints while "
+"Indicates whether or not the test mode should be enabled. If enabled, filled"
+" out surveys won't actually be sent, to avoid cluttering Expoints while "
 "testing."
 msgstr ""
 "Indien aangevinkt, dan wordt de testmode van Expoints ingeschakeld. "
@@ -495,8 +495,8 @@ msgstr ""
 #: openforms/analytics_tools/models.py:339
 msgid ""
 "Force the tag manager scripts to run, even if no cookie consent is given or "
-"individual tools are not enabled. You typically need this when you are using "
-"an external Consent Management Platform (CMP) through a tag manager. Be "
+"individual tools are not enabled. You typically need this when you are using"
+" an external Consent Management Platform (CMP) through a tag manager. Be "
 "careful when you enable this, as incorrect configuration may have privacy "
 "and/or legal implications."
 msgstr ""
@@ -525,8 +525,8 @@ msgstr ""
 
 #: openforms/analytics_tools/models.py:495
 msgid ""
-"If you enable GovMetric, you need to provide the source ID for all languages "
-"(the same one can be reused)."
+"If you enable GovMetric, you need to provide the source ID for all languages"
+" (the same one can be reused)."
 msgstr ""
 "Wanneer je GovMetric inschakelt, dan moet je een source ID ingegeven voor "
 "elke taaloptie. Je kan hetzelfde ID hergebruiken voor meerdere talen."
@@ -561,8 +561,8 @@ msgid ""
 "has passed, the session is expired and the user is 'logged out'. Note that "
 "every subsequent API call resets the expiry."
 msgstr ""
-"Aantal seconden waarna een sessie verloopt. Na het verlopen van de sessie is "
-"de gebruiker uitgelogd en kan zijn huidige inzending niet meer afmaken. "
+"Aantal seconden waarna een sessie verloopt. Na het verlopen van de sessie is"
+" de gebruiker uitgelogd en kan zijn huidige inzending niet meer afmaken. "
 "Opmerking: Bij elke interactie met de API wordt de sessie verlengd."
 
 #: openforms/api/drf_spectacular/hooks.py:29
@@ -574,8 +574,8 @@ msgid ""
 "If true, the user is allowed to navigate between submission steps even if "
 "previous submission steps have not been completed yet."
 msgstr ""
-"Indien waar, dan mag de gebruiker vrij binnen de formulierstappen navigeren, "
-"ook als de vorige stappen nog niet voltooid zijn."
+"Indien waar, dan mag de gebruiker vrij binnen de formulierstappen navigeren,"
+" ook als de vorige stappen nog niet voltooid zijn."
 
 #: openforms/api/drf_spectacular/hooks.py:45
 msgid "Language code of the currently activated language."
@@ -686,12 +686,12 @@ msgstr "Ping de API"
 
 #: openforms/api/views/views.py:60
 msgid ""
-"Pinging the API extends the user session. Note that you must be a staff user "
-"or have active submission(s) in your session."
+"Pinging the API extends the user session. Note that you must be a staff user"
+" or have active submission(s) in your session."
 msgstr ""
-"Door de API te pingen wordt de sessie van de gebruiker verlengd. Merk op dat "
-"je een actieve inzending in je sessie dient te hebben of ingelogd moet zijn "
-"in de beheerinterface."
+"Door de API te pingen wordt de sessie van de gebruiker verlengd. Merk op dat"
+" je een actieve inzending in je sessie dient te hebben of ingelogd moet zijn"
+" in de beheerinterface."
 
 #: openforms/appointments/admin.py:43
 msgid "Please configure the plugin first"
@@ -741,10 +741,10 @@ msgstr "Productcode"
 
 #: openforms/appointments/api/serializers.py:49
 #: openforms/appointments/api/serializers.py:84
+#: openforms/config/models/csp.py:76 openforms/config/models/map.py:18
 #: openforms/contrib/customer_interactions/models.py:19
 #: openforms/contrib/objects_api/models.py:21
 #: openforms/registrations/contrib/zgw_apis/models.py:29
-#: openforms/config/models/csp.py:76 openforms/config/models/map.py:18
 msgid "identifier"
 msgstr "Unieke identificatie"
 
@@ -754,17 +754,17 @@ msgstr "Unieke sleutel van het product"
 
 #: openforms/appointments/api/serializers.py:51
 #: openforms/appointments/api/serializers.py:86
+#: openforms/config/models/map.py:70 openforms/config/models/theme.py:21
 #: openforms/contrib/customer_interactions/models.py:14
 #: openforms/contrib/objects_api/models.py:16
-#: openforms/dmn/api/serializers.py:76 openforms/forms/admin/form.py:268
+#: openforms/dmn/api/serializers.py:76 openforms/forms/admin/form.py:271
 #: openforms/forms/admin/form_definition.py:69
-#: openforms/forms/models/category.py:13 openforms/forms/models/form.py:80
+#: openforms/forms/models/category.py:13 openforms/forms/models/form.py:81
 #: openforms/forms/models/form_definition.py:40
 #: openforms/forms/models/form_registration_backend.py:23
 #: openforms/forms/models/form_variable.py:265
 #: openforms/multidomain/models.py:7 openforms/products/models/product.py:24
 #: openforms/registrations/contrib/zgw_apis/models.py:24
-#: openforms/config/models/map.py:70 openforms/config/models/theme.py:21
 msgid "name"
 msgstr "naam"
 
@@ -791,11 +791,12 @@ msgstr "Limiet aantal afspraakaanvragen"
 
 #: openforms/appointments/api/serializers.py:63
 msgid ""
-"The maximum number of times a user can request this appointment. Defaults to "
-"0 (unlimited) if the plugin does not specify this."
+"The maximum number of times a user can request this appointment. Defaults to"
+" 0 (unlimited) if the plugin does not specify this."
 msgstr ""
 "Het maximale aantal keren dat een gebruiker deze afspraak kan aanvragen. De "
-"standaardwaarde 0 (onbeperkt) wordt gebruikt als de plugin dit niet aangeeft."
+"standaardwaarde 0 (onbeperkt) wordt gebruikt als de plugin dit niet "
+"aangeeft."
 
 #: openforms/appointments/api/serializers.py:75
 msgid ""
@@ -972,7 +973,8 @@ msgstr "Het geselecteerde tijdstip is niet (langer) beschikbaar."
 msgid "ID of the product, repeat for multiple products."
 msgstr "Product-ID, herhaal deze parameter voor meerdere producten."
 
-#: openforms/appointments/api/views.py:77 openforms/products/api/viewsets.py:12
+#: openforms/appointments/api/views.py:77
+#: openforms/products/api/viewsets.py:12
 msgid "List available products"
 msgstr "Producten weergeven"
 
@@ -1050,8 +1052,9 @@ msgid "Submission does not contain all the info needed to make an appointment"
 msgstr ""
 "Er ontbreken gegevens in de inzending om een afspraak te kunnen registreren."
 
-#: openforms/appointments/constants.py:11 openforms/submissions/constants.py:15
-#: openforms/submissions/constants.py:68 openforms/translations/constants.py:9
+#: openforms/appointments/constants.py:11
+#: openforms/submissions/constants.py:15 openforms/submissions/constants.py:68
+#: openforms/translations/constants.py:9
 msgid "Failed"
 msgstr "Mislukt"
 
@@ -1110,7 +1113,7 @@ msgid "Hidden"
 msgstr "Verborgen"
 
 #: openforms/appointments/contrib/jcc_rest/constants.py:16
-#: openforms/forms/constants.py:62
+#: openforms/forms/constants.py:60
 msgid "Required"
 msgstr "Verplicht"
 
@@ -1245,6 +1248,10 @@ msgstr "Ongeldige pluginconfiguratie: {exc}"
 
 #: openforms/appointments/contrib/jcc_rest/plugin.py:564
 #: openforms/appointments/contrib/qmatic/plugin.py:408
+#: openforms/config/templates/admin/config/overview.html:4
+#: openforms/config/templates/admin/config/overview.html:9
+#: openforms/config/templates/admin/config/overview.html:15
+#: openforms/config/views.py:49 openforms/config/views.py:65
 #: openforms/contrib/brk/checks.py:42
 #: openforms/contrib/kadaster/config_check.py:30
 #: openforms/contrib/kadaster/config_check.py:61
@@ -1259,10 +1266,6 @@ msgstr "Ongeldige pluginconfiguratie: {exc}"
 #: openforms/registrations/contrib/microsoft_graph/plugin.py:130
 #: openforms/registrations/contrib/stuf_zds/plugin.py:681
 #: openforms/submissions/api/serializers.py:313
-#: openforms/config/templates/admin/config/overview.html:4
-#: openforms/config/templates/admin/config/overview.html:9
-#: openforms/config/templates/admin/config/overview.html:15
-#: openforms/config/views.py:49 openforms/config/views.py:65
 msgid "Configuration"
 msgstr "Configuratie"
 
@@ -1395,7 +1398,7 @@ msgid "The submission that made the appointment"
 msgstr "De inzending bij deze afspraak"
 
 #: openforms/appointments/models.py:68
-#: openforms/forms/models/form_version.py:56
+#: openforms/forms/models/form_version.py:78
 msgid "created"
 msgstr "aangemaakt"
 
@@ -1546,7 +1549,8 @@ msgid "The authentication attribute provided by this plugin."
 msgstr "Het authenticatie attribuut geleverd door deze plugin."
 
 #: openforms/authentication/api/serializers.py:28
-#: openforms/forms/api/viewsets.py:637 openforms/payments/api/serializers.py:18
+#: openforms/forms/api/viewsets.py:660
+#: openforms/payments/api/serializers.py:18
 #: openforms/registrations/api/serializers.py:13
 msgid "JSON schema"
 msgstr "JSON-schema"
@@ -1597,7 +1601,8 @@ msgid "..."
 msgstr "..."
 
 #: openforms/authentication/api/serializers.py:60
-#: openforms/dmn/api/serializers.py:17 openforms/payments/api/serializers.py:25
+#: openforms/dmn/api/serializers.py:17
+#: openforms/payments/api/serializers.py:25
 msgid "Identifier"
 msgstr "Unieke identificatie"
 
@@ -1777,8 +1782,9 @@ msgstr "Claim pad"
 #: openforms/authentication/contrib/yivi_oidc/oidc_plugins/schemas.py:21
 #: openforms/authentication/contrib/yivi_oidc/oidc_plugins/schemas.py:53
 msgid ""
-"Path to the claim value holding the level of assurance. If left empty, it is "
-"assumed there is no LOA claim and the configured fallback value will be used."
+"Path to the claim value holding the level of assurance. If left empty, it is"
+" assumed there is no LOA claim and the configured fallback value will be "
+"used."
 msgstr ""
 "Naam van de claim die het betrouwbaardheidsniveau bevat. Als er geen waarde "
 "wordt opgegeven, wordt verondersteld dat er geen LOA-claim aanwezig is en "
@@ -1793,7 +1799,8 @@ msgstr "Standaardwaarden"
 #: openforms/authentication/contrib/digid_eherkenning_oidc/oidc_plugins/schemas.py:144
 #: openforms/authentication/contrib/yivi_oidc/oidc_plugins/schemas.py:33
 #: openforms/authentication/contrib/yivi_oidc/oidc_plugins/schemas.py:65
-msgid "Fallback level of assurance, in case no claim value could be extracted."
+msgid ""
+"Fallback level of assurance, in case no claim value could be extracted."
 msgstr ""
 "Standaardwaarde voor het betrouwbaarheidsniveau. Wordt gebruikt wanneer er "
 "geen claimwaarde kan worden verkregen."
@@ -1811,7 +1818,8 @@ msgstr "eIDAS identity instellingen."
 
 #: openforms/authentication/contrib/digid_eherkenning_oidc/oidc_plugins/schemas.py:60
 msgid ""
-"Path to the claim value holding the bsn identifier of the authenticated user."
+"Path to the claim value holding the bsn identifier of the authenticated "
+"user."
 msgstr "Naam van de claim die het BSN bevat van de geauthoriseerde gebruiker."
 
 #: openforms/authentication/contrib/digid_eherkenning_oidc/oidc_plugins/schemas.py:69
@@ -1831,8 +1839,8 @@ msgstr "voornaamclaim van de wettelijke vertegenwoordiger"
 
 #: openforms/authentication/contrib/digid_eherkenning_oidc/oidc_plugins/schemas.py:82
 msgid ""
-"Path to the claim value that holds the legal first name of the authenticated "
-"user."
+"Path to the claim value that holds the legal first name of the authenticated"
+" user."
 msgstr ""
 "Claim die de wettelijke achternaam van de geauthenticeerde gebruiker bevat"
 
@@ -1857,7 +1865,8 @@ msgid ""
 "Path to the claim value that holds the legal birthdate of the authenticated "
 "user."
 msgstr ""
-"Claim die de wettelijke geboortedatum van de geauthenticeerde gebruiker bevat"
+"Claim die de wettelijke geboortedatum van de geauthenticeerde gebruiker "
+"bevat"
 
 #: openforms/authentication/contrib/digid_eherkenning_oidc/oidc_plugins/schemas.py:120
 msgid "OIDC eIDAS company configuration options."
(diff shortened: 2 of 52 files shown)


src/openforms/forms/import_export/typing.py
```diff
@@ -0,0 +1,47 @@
...
+
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class FormConfigurationOptions(models.TextChoices):
+    registration_backends = "registrationBackends", _("Registration backends")
+    prefill = "prefill", _("Prefill")
+    payment_backend = "paymentBackend", _("Payment backend")
+    auth_backends = "authBackends", _("Authentication backends")
```
robinmolen, line 15: The `auth_backends` form configuration option is used for by the import options, which will be added in another PR

src/openforms/js/components/admin/form_design/FormSubmit.js
```diff
@@ -28,7 +34,27 @@ const ExportAction = () => {
     defaultMessage: 'Export this form',
     description: 'Export form button title',
   });
-  return <ActionButton name="_export" text={btnText} title={btnTitle} />;
+
+  return (
+    <>
+      <input ref={exportButtonRef} type="submit" name="_export" hidden />
+      <input ref={exportOptionsRef} type="hidden" name="export_options" />
```
robinmolen, line 41: I tried to adjust this implementation and use the API endpoint, but couldn't get it to work. Might give it another try to clean this up

Edit: Haven't tried yet. I think it's alright for now

src/openforms/forms/tests/test_import_export.py
```diff
@@ -56,6 +65,7 @@
 from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
 from openforms.variables.tests.factories import ServiceFetchConfigurationFactory
 
+from ...authentication.constants import AuthAttribute
```
robinmolen: Import path

src/openforms/forms/import_export/resources/yivi_attribute_group.py
```diff
@@ -0,0 +1,21 @@
...
+    def export_for_form(self, form: Form):
+        yivi_auth_backend = form.auth_backends.all().filter(backend="yivi_oidc").first()
+        if yivi_auth_backend is None:
+            return self.export(queryset=[])
+
+        return self.export(
+            queryset=AttributeGroup.objects.filter(
+                uuid__in=yivi_auth_backend.options["additional_attributes_groups"]
+            )
+        )
```
robinmolen, line 24: Missing unit tests for all these `export_for_form` methods

src/openforms/forms/tests/test_api_import_export.py
robinmolen, line 1: Add negative test that breaks the API serializer

src/openforms/forms/utils.py
```diff
@@ -78,12 +88,9 @@ def to_json(obj: Any):
     return json.dumps(obj, cls=DjangoJSONEncoder)
 
 
-def form_to_json(form_id: int) -> dict:
+def form_to_json(form_id: int, export_options: FormExportOptions = None) -> dict:
```
robinmolen: `export_options` should be required

src/openforms/forms/utils.py
```diff
@@ -140,8 +168,10 @@ def form_to_json(form_id: int) -> dict:
     return resources
 
 
-def export_form(form_id, archive_name=None, response=None):
-    resources = form_to_json(form_id)
+def export_form(
+    form_id, archive_name=None, response=None, export_options: FormExportOptions = None
```
robinmolen: I left `export_options` optional to transition slowly, but should now be marked as required

src/openforms/import_export/serializers/base.py
```diff
@@ -0,0 +1,68 @@
...
+    ) -> JSONObject:
+        return representation
+
+    def remove_excluded_form_configuration(
+        self, representation: JSONObject
+    ) -> JSONObject:
+        selected_options = (
+            set(export_options.form_configuration)
+            if (export_options := self.get_export_options()) is not None
+            else []
```
robinmolen:
```suggestion
            else set()
```
Same goes for remove_excluded_additional_form_configuration

src/openforms/forms/import_export/serializers/form.py
```diff
@@ -0,0 +1,88 @@
...
+        return fields
+
+    def remove_sensitive_content(self, instance, representation):
+        representation = super().remove_sensitive_content(instance, representation)
+        representation["internal_remarks"] = ""
+
+        for registration in representation.get("registration_backends", []):
+            if registration["backend"] == "email":
+                registration["options"]["to_emails"] = []
+                registration["options"]["payment_emails"] = []
```
robinmolen: Might be nice/better to move this to a `FormRegistrationExportSerializer`

src/openforms/forms/import_export/serializers/form_definition.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if (form := self.context.get("form", None)) is None:
+            return representation
+
+        sensitive_variables = [
+            registration.options["to_emails_from_variable"]
+            for registration in form.registration_backends.all()
+            if registration.backend == "email"
+            and "to_emails_from_variable" in registration.options
+        ]
```
robinmolen, line 83: This is used for form definitions, logic and variables. Might be good to make to a shared utils method

src/openforms/js/lang/nl.json
```diff
@@ -2632,6 +2713,11 @@
     "isTranslated": true,
     "originalDefault": "Label"
   },
+  "iAFlLD": {
+    "defaultMessage": "Registratie backends",
```
robinmolen, line 2747:
```suggestion
    "defaultMessage": "Registratie plugins",
```

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -6661,28 +6665,68 @@ msgstr "ingezonden tussen"
...
-#: openforms/forms/admin/tasks.py:70
+#: openforms/forms/admin/tasks.py:74
 msgid "Forms export ready"
 msgstr "Export formulieren gereed"
 
-#: openforms/forms/admin/views.py:39
+#: openforms/forms/admin/views.py:43
+#| msgid "Form.io configuration"
+msgid "Anonymize form configuration"
+msgstr "Formulierconfiguratie anonimiseren"
```
robinmolen:
```suggestion
msgstr "Formulierinstellingen anonimiseren"
```

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -6661,28 +6665,68 @@ msgstr "ingezonden tussen"
...
+msgid ""
+"Whether sensative form configuration should be anonymized during exporting."
+msgstr ""
+"Geeft aan of gevoelige informatie bij het exporteren automatisch "
+"geanonimiseerd worden."
+
+#: openforms/forms/admin/views.py:51
+#| msgid "Form.io configuration"
+msgid "Form configuration"
+msgstr "Formulierconfiguratie"
```
robinmolen:
```suggestion
msgstr "Formulierinstellingen"
```

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -6661,28 +6665,68 @@ msgstr "ingezonden tussen"
...
+msgstr "Ondersteunende formulierconfiguratie"
+
+#: openforms/forms/admin/views.py:70
+#: openforms/forms/api/serializers/form.py:680
+msgid ""
+"Which additional form configuration should be included in the export file "
+"content."
+msgstr ""
+"Welke ondersteunende formulierconfiguratie meegenomen worden bij het "
+"exporteren."
```
robinmolen, line 6745: `Formulierinstellingen` ipv `Formulierconfiguratie`

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -8699,18 +8742,76 @@ msgstr "U ziet deze pagina omdat u een ingelogde beheerder bent."
...
+"The component '{key}' (at JSON path '{path}') has template syntax errors in "
+"the field '{field}'."
+msgstr ""
+"De component  '{key}' (op JSON-pad '{path}') heeft sjabloonsyntaxfouten in "
+"het veld '{field}'."
+
+#: openforms/import_export/typing.py:11
+#| msgid "Registration backend"
+msgid "Registration backends"
+msgstr "Registratiebackends"
```
robinmolen:
```suggestion
msgstr "Registratie plugins"
```

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
```diff
@@ -8699,18 +8742,76 @@ msgstr "U ziet deze pagina omdat u een ingelogde beheerder bent."
...
+
+#: openforms/import_export/typing.py:12 openforms/prefill/apps.py:7
+#: openforms/submissions/constants.py:52
+msgid "Prefill"
+msgstr "Prefill"
+
+#: openforms/import_export/typing.py:13 openforms/payments/models.py:105
+msgid "Payment backend"
+msgstr "Betaalprovider backend"
```
robinmolen:
```suggestion
msgstr "Betaalprovider plugin"
```

src/openforms/forms/import_export/typing.py
robinmolen: Discussed with Laurens, `auth_backends` is now also used by the export options

src/openforms/forms/import_export/serializers/form.py
robinmolen: Has been moved

sergei-maertens:
(I just rebased this on latest main before starting the review)

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
-    Formulieren worden geëxporteerd als een ZIP-bestand waarin meerdere 
-    JSON-bestanden zitten. Elk JSON-bestand bevat de configuratie van het 
+    Formulieren worden geëxporteerd als een ZIP-bestand waarin meerdere
+    JSON-bestanden zitten. Elk JSON-bestand bevat de configuratie van het
     formulier zelf of een stap binnen het formulier. De JSON-structuur komt
     1-op-1 overeen met de API-specificatie waardoor het export formaat tevens
     open en transparent is.
 
+Formulier exporteer opties
+--------------------------
```
sergei-maertens, line 35:
```suggestion
Formulierexporteeropties
------------------------
```

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
+
+Bij het exporteren van een formulier heb je enkele opties over hoe het
+ZIP-bestand wordt gemaakt:
```
sergei-maertens, line 38:
```suggestion
Bij het exporteren van een formulier heb je enkele opties die de inhoud van
het exportbestand beïnvloeden:
```

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
+
+* **Anonimiseer formulierinstellingen**: Indien aangevinkt, dan worden de
+  formulierinstellingen automatisch geanonimiseerd. Dit verwijderd alle
```
sergei-maertens, line 41:
```suggestion
  formulierinstellingen automatisch geanonimiseerd. Dit verwijdert alle
```

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
+  e-mailadressen die zijn configureerd in de E-mail registratie en de interne
```
sergei-maertens, line 42:
```suggestion
  e-mailadressen die zijn configureerd in de e-mailregistratie en de interne
```

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
+  opmerkingen op het formulier. Deze optie is standaard aangevinkt.
+* **Formulierinstellingen**: Hiermee kunnen registratie backends,
```
sergei-maertens, line 44:
```suggestion
* **Formulierinstellingen**: Hiermee kunnen registratiebackends,
```

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
+  betaalprovider, prefill en inlogmethode eenvoudig bij het exporteren
+  weggelaten worden. Standaard worden alle instellingen mee geëxporteerd.
```
sergei-maertens, line 46:
```suggestion
  weggelaten worden. Standaard worden alle instellingen mee-geëxporteerd.
```

docs/manual/forms/export_import.rst
```diff
@@ -25,17 +25,32 @@ U hoeft het ZIP-bestand niet uit te pakken.
...
+* **Aanvullende formulierinstellingen**: Hiermee kunnen aanvullende
+  instellingen van het formulier aan het ZIP-bestand toegevoegd worden.
```
sergei-maertens, line 48:
```suggestion
  instellingen van het formulier aan de export toegevoegd worden.
```

I'd reduce the focus on the ZIP file format - it's an implementation detail.

src/openforms/conf/locale/nl/LC_MESSAGES/django.po
sergei-maertens, line 1: 10K lines changed, I can't review this. Why is this included in the diff/PR?

src/openforms/forms/admin/form.py
```diff
@@ -12,13 +14,14 @@
 
 from openforms.api.utils import underscore_to_camel
 from openforms.emails.models import ConfirmationEmailTemplate
+from openforms.forms.import_export.export_form import export_form
+from openforms.forms.import_export.typing import FormExportOptions
```
sergei-maertens, line 18:
```suggestion
from ..import_export.export_form import export_form
from ..import_export.typing import FormExportOptions
```
it might also be elegant to re-export things in the `__init__.py` so you can do:

```py
from ..import_export import FormExportOptions, export_form
```

src/openforms/forms/admin/form.py
```diff
@@ -299,14 +302,30 @@ def response_post_save_change(self, request, obj):
                 reverse("admin:forms_form_change", args=(copied_form.pk,))
             )
         if "_export" in request.POST:
+            export_options = json.loads(request.POST.get("export_options", "{}"))
```
sergei-maertens, line 305: this is quite naive - you're deserializing untrusted user input. Consider:

```py
json.loads("{'broken")
```

at a minimum, this should have a `try: except:` around it that does some basic error handling, but you'll certainly want to perform validation on the `export_options` data as well. It's now extremely easy to spam Sentry with crash information.

Looking at this whole code-path, and the response being a file download... wouldn't it make more sense to make this an (internal) API endpoint where you get the JSON deserialization and validation through DRF serializers "for free"?

src/openforms/forms/admin/form.py
```diff
@@ -299,14 +302,30 @@ def response_post_save_change(self, request, obj):
...
+                export_options=FormExportOptions(
+                    **{
+                        field_name: export_options[field_name]
+                        for field_name in (
+                            "remove_sensitive_content",
+                            "form_configuration",
+                            "additional_form_configuration",
+                        )
+                        if field_name in export_options
+                    },
```
sergei-maertens, line 326: this is essentially `FormExportOptions(**export_options)` but with a tiny guard rail where you ignore unknown keys.

Going the serializer route for validation will definitely make this neater, you can even implement a convenience method on the serializer:

```py
class ExportOptionsSerializer(...):
    def as_export_options(self) -> FormExportOptions:
        return FormExportOptions(
            remove_sensitive_content=self.validated_data["remove_sensitive_content"],
            remove_sensitive_content=self.validated_data["form_configuration"],
            remove_sensitive_content=self.validated_data["additional_form_configuration"],
        )
```

src/openforms/forms/admin/tasks.py
```diff
@@ -17,18 +17,25 @@
 from openforms.accounts.models import User
 from openforms.celery import app
 from openforms.emails.utils import send_mail_html
+from openforms.forms.import_export.export_form import export_form
+from openforms.forms.import_export.typing import (
+    FormExportOptions,
+    FormExportOptionsData,
+)
```
sergei-maertens, line 24: I think the public API can be exposed through `openforms.forms.import_export` as remarked before. Orrrr, as documented in the backend developer documentation, you expose the public API through a `service.py` module: `from ..import_export.service import ...`.

That makes it very clear what's public/private python API.

src/openforms/forms/admin/tasks.py
```diff
@@ -17,18 +17,25 @@
...
 from openforms.logging import audit_logger
 from openforms.utils.urls import build_absolute_uri
 
 from ..models import Form
 from ..models.form import FormsExport
-from ..utils import export_form, import_form
+from ..utils import import_form
```
sergei-maertens, line 30: should this still be in `utils`? shouldn't this also be moved into `openforms.forms.import_export`?

src/openforms/forms/admin/tasks.py
```diff
@@ -17,18 +17,25 @@
...
 
 logger = structlog.stdlib.get_logger(__name__)
 
 
 @app.task(ignore_result=True)
-def process_forms_export(forms_uuids: list, user_id: int) -> None:
+def process_forms_export(
+    forms_uuids: list, user_id: int, export_options: FormExportOptionsData
```
sergei-maertens, line 37:
```suggestion
    forms_uuids: Sequence[str], user_id: int, export_options: FormExportOptionsData
```
I really dislike bare container types :) the existing version was equivalent to `list[Any]`

sergei-maertens, line 37: You're passing the form export options as a dict here rather than the dataclass that you use internally to represent it. I'm sure that's because you realised that the json serializer (used by celery tasks) can only handle the basic JSON serializable types.

There *are* mechanisms in Celery that allow you to provide custom encoders/serializers that would allow passing dataclasses, see https://stackoverflow.com/a/22520957

I think that's currently out of scope though - I've been hitting a similar consideration in my msgspec branch where ideally I'd pass along msgspec structs that get serialized/deserialized properly at the celery framework level, but also opted not to commit to that complexity yet.

At the least I'd expect some code comments / TODO to be around that there may be more elegant patterns for passing along the configuration options.

Note that Celery has built-in support for Pydantic: https://docs.celeryq.dev/en/stable/userguide/tasks.html#task-pydantic though I'm really not much of a Pydantic fan

src/openforms/forms/admin/views.py
```diff
@@ -22,6 +22,10 @@
 from privates.storages import private_media_storage
 from rest_framework.exceptions import ValidationError
 
+from openforms.forms.import_export.typing import (
+    AdditionalFormConfigurationOptions,
+    FormConfigurationOptions,
+)
```
sergei-maertens, line 28:
```suggestion
from ..import_export.typing import (
    AdditionalFormConfigurationOptions,
    FormConfigurationOptions,
)
```

src/openforms/forms/admin/views.py
```diff
@@ -33,6 +37,38 @@
 
 class ExportFormsForm(forms.Form):
     forms_uuids = SimpleArrayField(forms.UUIDField(), widget=forms.HiddenInput)
+    remove_sensitive_content = forms.BooleanField(
+        label=_("Anonymize form configuration"),
+        required=False,
+        initial=True,
+        help_text=_(
+            "Whether sensative form configuration should be anonymized during exporting."
```
sergei-maertens, line 45:
```suggestion
            "Whether sensative/internal form configuration should be anonymized during exporting."
```

src/openforms/forms/api/serializers/form.py
```diff
@@ -735,15 +739,27 @@ def get_submission_limit_reached(self, obj: Form) -> bool:
...
-class FormExportSerializer(FormSerializer):
-    def get_fields(self):
-        fields = super().get_fields()
-        # for export we want to use the list of plugin-id's instead of detailed info objects
-        if "login_options" in fields:
-            del fields["login_options"]
-        if "payment_options" in fields:
-            del fields["payment_options"]
-        return fields
+class FormAPIExportRequestSerializer(serializers.Serializer):
```
sergei-maertens, line 742: Oh, you have the serializer already - why isn't it used to validate the export options?

src/openforms/forms/api/serializers/form.py
```diff
@@ -735,15 +739,27 @@ def get_submission_limit_reached(self, obj: Form) -> bool:
...
+            "Which form configuration should be included in the export file content."
+        ),
+    )
+    additional_form_configuration = serializers.MultipleChoiceField(
+        required=False,
+        choices=AdditionalFormConfigurationOptions.choices,
+        help_text=_(
+            "Which additional form configuration should be included in the export file content."
+        ),
+    )
```
sergei-maertens, line 762: Does it make sense that the defaults are different here from the export form defaults?

src/openforms/forms/api/serializers/form_definition.py
```diff
@@ -121,9 +122,7 @@ def to_representation(self, instance):
         #    for the dynamic formio configuration in the context of a submission.
         # 2. The serializers/API endpoints of :module:`openforms.forms.api` for
         #    'standalone' use/introspection.
-        is_export = self.context.get("is_export", False)
-
-        if not is_export:
+        if not self.is_exporting:
```
sergei-maertens, line 125: what was the reason that you can't use serializer context?

I think this is quite fragile now because you have to remember to mark each serializer instance, whereas the context approach is set-once-configured-always, because serializer context is the same object for the whole serializer/field tree once it's set - it's globally available for any node in the serializer tree(s).

src/openforms/forms/api/viewsets.py
```diff
@@ -21,6 +21,8 @@
 
 from openforms.api.pagination import PageNumberPagination
 from openforms.api.serializers import ExceptionSerializer, ValidationErrorSerializer
+from openforms.forms.import_export.export_form import export_form
+from openforms.forms.import_export.typing import FormExportOptions
```
sergei-maertens, line 25:
```suggestion
from ..import_export.export_form import export_form
from ..import_export.typing import FormExportOptions
```
& of course the other remarks about using `__init__` or `service` module for public API still apply.

src/openforms/forms/api/viewsets.py
```diff
@@ -435,7 +441,24 @@ def export(self, request, *args, **kwargs):
...
+                **{
+                    field_name: serializer.validated_data[field_name]
+                    for field_name in (
+                        "remove_sensitive_content",
+                        "form_configuration",
+                        "additional_form_configuration",
+                    )
+                    if field_name in serializer.validated_data
+                },
+            ),
```
sergei-maertens, line 460: I don't really like this "cleverness", when you can make things properly explicit and type-checkable. See my earlier remark about the serializer convenience method, but at minimum you can make the code a ton more readable *and* shorter:


```suggestion
            export_options=FormExportOptions(
                remove_sensitive_content=serializer.validated_data["remove_sensitive_content"],
                form_configuration=serializer.validated_data["form_configuration"],
                additional_form_configuration=serializer.validated_data["additional_form_configuration"],
            ),
```

set serializer field defaults so that the key is guaranteed to be present in the validated data, and you don't even need to `.get(...)` it.

src/openforms/forms/import_export/resources/__init__.py
sergei-maertens, line 1: this could all have been a single `resources.py`. From the Zen of Python:

> Flat is better than nested

It's 100 lines of code in total, keep it in a single module please.

src/openforms/forms/import_export/resources/base.py
```diff
@@ -0,0 +1,10 @@
+from import_export.resources import ModelResource
+
+from openforms.forms.models import Form
+
+
+class BaseResource(ModelResource):
+    def export_for_form(self, form: Form):
```
sergei-maertens, line 7: a return type hint would be the minimum to know what to expect from subclasses

src/openforms/forms/import_export/resources/wms_tile_layer.py
```diff
@@ -0,0 +1,28 @@
...
+from .base import BaseResource
+
+
+class WMSTileLayerResource(BaseResource):
+    class Meta:
+        model = MapWMSTileLayer
+        fields = ("uuid", "name", "url")
+
+    def export_for_form(self, form: Form):
+        wms_tile_layers = []
```
sergei-maertens, line 13: please add type hints! They help so much while refactoring implementations, as I'm doing in the msgspec rework.

src/openforms/forms/import_export/resources/wms_tile_layer.py
```diff
@@ -0,0 +1,28 @@
...
+                        overlay["uuid"]
+                        for overlay in component.get("overlays", [])
+                        if overlay["uuid"] != ""
+                    )
+
+        if len(wms_tile_layers) == 0:
+            return self.export(queryset=[])
+
+        return self.export(
+            queryset=MapWMSTileLayer.objects.filter(uuid__in=list(set(wms_tile_layers)))
```
sergei-maertens, line 27: why isn't `wms_tile_layers` a set to begin with?

src/openforms/forms/import_export/resources/wms_tile_layer.py
sergei-maertens: Renaming this to `tile_layer_uuids` also properly defines what's the expected value, I was under the impression that the whole layer definition would be included rather than just the ID

src/openforms/forms/import_export/resources/wmts_tile_layer.py
```diff
@@ -0,0 +1,26 @@
...
+from .base import BaseResource
+
+
+class WMTSTileLayerResource(BaseResource):
+    class Meta:
+        model = MapTileLayer
+        fields = ("identifier", "label", "url")
+
+    def export_for_form(self, form: Form):
+        wmts_tile_layers = []
```
sergei-maertens, line 13: this too can be a set from the start

src/openforms/forms/import_export/resources/wmts_tile_layer.py
```diff
@@ -0,0 +1,26 @@
...
+        for step in form.form_step_map.values():
+            for component in step.form_definition.iter_components():
+                if component["type"] == "map":
+                    wmts_tile_layers.append(component.get("tileLayerIdentifier", ""))
+
+        if len(wmts_tile_layers) == 0:
+            return self.export(queryset=[])
```
sergei-maertens, line 20: I believe this branch can be omitted and Django will short-circuit in queries with empty collections, but that would have to be confirmed

src/openforms/forms/import_export/resources/yivi_attribute_group.py
```diff
@@ -0,0 +1,24 @@
...
+from .base import BaseResource
+
+
+class YiviAttributeGroupResource(BaseResource):
+    class Meta:
+        model = AttributeGroup
+        fields = ("uuid", "name", "description", "attributes")
+
+    def export_for_form(self, form: Form):
+        yivi_auth_backend = form.auth_backends.all().filter(backend="yivi_oidc").first()
```
sergei-maertens, line 13:
```suggestion
        yivi_auth_backend = form.auth_backends.filter(backend="yivi_oidc").first()
```
`.all().filter()` doesn't do anything extra, you can always leave out that `.all()` in chains.

sergei-maertens, line 13: This backend ID should be defined in a constant and re-used instead of this magic string. A constant is easier to grep for its usages.

sergei-maertens, line 13: At the DB level there's no guarantee that there will only ever be one yivi backend, right? So I woudn't use `.first()` and instead just loop over the whole collection and collect everything.

src/openforms/forms/import_export/resources/yivi_attribute_group.py
```diff
@@ -0,0 +1,24 @@
...
+        if (
+            yivi_auth_backend is None
+            or "additional_attributes_groups" not in yivi_auth_backend.options
+        ):
+            return self.export(queryset=[])
```
sergei-maertens, line 18: this can probably also be simplified by building up the set of UUIDs and then always doing the filter, django shortcuts it:

```
>>> from django.db import connection
>>> len(connection.queries)
9
>>> Form.objects.filter(uuid__in=())
<MultilingualFormQuerySet []>
>>> len(connection.queries)
9
```

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
+from collections.abc import Sequence
+from typing import ClassVar
+
+from rest_framework import serializers
+
+from openforms.forms.import_export.typing import (
```
sergei-maertens, line 6: should be a relative import

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
...
+from openforms.typing import JSONObject
+
+
+class BaseExportSerializer(serializers.Serializer):
+    excluded_form_configuration_cleanup: ClassVar[
+        Sequence[FormConfigurationCleanup]
+    ] = ()
+    excluded_additional_form_configuration_cleanup: ClassVar[
+        Sequence[AdditionalFormConfigurationCleanup]
+    ] = ()
```
sergei-maertens, line 20: please add docstrings so that I know what this means without seeing how it's used in the implementation

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
...
+    save_export_fields: ClassVar[Sequence[str]] = ()
```
sergei-maertens, line 21:
```suggestion
    safe_export_fields: ClassVar[Sequence[str]] = ()
```

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
...
+            representation
+        )
+
+        return representation
+
+    def remove_sensitive_content(
+        self, instance, representation: JSONObject
+    ) -> JSONObject:
+        """
+        Remove all fields that are not in the save_export_fields list.
```
sergei-maertens, line 48:
```suggestion
        Remove all fields that are not in the safe_export_fields list.
```

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
...
+        representation = self.remove_excluded_form_configuration(representation)
+        representation = self.remove_excluded_additional_form_configuration(
+            representation
+        )
+
+        return representation
+
+    def remove_sensitive_content(
+        self, instance, representation: JSONObject
+    ) -> JSONObject:
```
sergei-maertens, line 46: These type hints are way too generic to capture the various flavours and don't give much benefit. However, since you're using a base serializer class, you can use generic types:

```py
class BaseExportSerializer[MT: type[Model], RT: JSONObject](serializers.Serializer):
    def remove_sensitive_content(self, instance: MT, representation: RT) -> RT: ...
```
with `MT` model type and `RT` the representation type, and that will give more type safety for the cleanup operations etc.

You can make your concrete serializer classes by doing:

```py
class MySerializer(BaseExportSerializer[Form, FormExportRepresentation]):
    ...
```

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
...
+            set(export_options.form_configuration)
+            if (export_options := self.get_export_options()) is not None
+            else set()
+        )
+
+        for config in self.excluded_form_configuration_cleanup:
+            if config.option not in selected_options:
+                config.cleanup(representation)
+
+        return representation
```
sergei-maertens, line 69: This structure looks weird because the export options is pretty much entirely on the `Form`-level, but this method is expected to handle nested resources of form as well?

It seems quite complex and over-engineered, and so generic to the point that the type annotations really don't provide any benefit anymore because everything ends up being a generic `JSONObject` blob and you have no more context of what you're actually receiveing 🤔 

I'm not sure how to fix this though.

src/openforms/forms/import_export/serializers/base.py
```diff
@@ -0,0 +1,87 @@
...
+        return {
+            key: field
+            for key, field in representation.items()
+            if key in self.save_export_fields
+        }
+
+    def remove_excluded_form_configuration(
+        self, representation: JSONObject
+    ) -> JSONObject:
+        selected_options = (
```
sergei-maertens, line 59: I'd rename `selected_options` to something like `options_to_keep`

That an option is selected doesn't tell me anything about the meaning of it being selected. Do you select it to keep it, or to remove it?

src/openforms/forms/import_export/serializers/form.py
```diff
@@ -0,0 +1,154 @@
...
+    representation["registration_backends"] = []
+
+
+def exclude_payment_backend(representation: JSONObject):
+    representation["payment_backend"] = ""
+    representation["payment_backend_options"] = {}
+
+
+def exclude_auth_backends(representation: JSONObject):
+    representation["auth_backends"] = []
```
sergei-maertens, line 34: This really needs proper type annotations support.

What I'm mostly thinking now is:

* why are we stripping out data after serialization (so the result of `to_representation`)?
* can't we instead patch the instances being exported in memory, clear the data, and then run the "normal" export?

Then at least we're operating on proper model instances which are strict, strongly typed Python objects and you won't get f*cked over by a stupid typo.

src/openforms/forms/import_export/serializers/form_definition.py
```diff
@@ -0,0 +1,89 @@
...
+    FormConfigurationOptions,
+)
+from openforms.prefill.constants import IdentifierRoles
+from openforms.typing import JSONObject
+
+from .base import BaseExportSerializer
+
+
+def clear_wms_tile_layers(representation: JSONObject):
+    for component in representation.get("configuration", {}).get("components", []):
```
sergei-maertens, line 16: why is this now doing the manual looping, not taking into account nested layout components?

src/openforms/forms/import_export/serializers/form_logic.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def remove_sensitive_content(self, instance, representation):
+        representation = super().remove_sensitive_content(instance, representation)
+        form = instance.form
+
+        sensitive_variables = (
+            registration.options["to_emails_from_variable"]
+            for registration in form.registration_backends.all()
+            if registration.backend == "email"
+            and "to_emails_from_variable" in registration.options
```
sergei-maertens, line 27: so what happens if that options serializer field gets renamed in the future?

Perhaps architecture wise we need to make sure that every node has some kind of export handler that it must implement?

We already enforce registration backend option serializers for every registration plugin, it's invoked during the registration process and ensures that we're actually working with the proper data. Can't we defer this cleaning to those data structures that properly understand their own context?

src/openforms/forms/import_export/serializers/form_step.py
```diff
@@ -0,0 +1,24 @@
...
+        "name",
+        "internal_name",
+        "url",
+        "is_applicable",
+        "login_required",
+        "is_reusable",
+        "literals",
+        "previous_text",
+        "save_text",
+        "next_text",
```
sergei-maertens, line 22: FOO_text is inside the `literals`, no?

Additionally, for the import/export process, we only need the `translations`, all the rest is derived from that

src/openforms/forms/import_export/serializers/form_variable.py
```diff
@@ -0,0 +1,57 @@
...
+from openforms.prefill.constants import IdentifierRoles
+from openforms.typing import JSONObject
+
+from .base import BaseExportSerializer
+
+
+def remove_prefill_from_variable(representation: JSONObject):
+    representation["prefill_plugin"] = ""
+    representation["prefill_attribute"] = ""
+    representation["prefill_identifier_role"] = IdentifierRoles.main
```
sergei-maertens, line 15: I'd leave this untouched - if you have a mandate form where some fields are prefill with details of the authorizer, while other fields are intended for the authorizee, you just changed the semantic meaning of those fields.

src/openforms/forms/import_export/export_form.py
```diff
@@ -0,0 +1,130 @@
...
+    resources = form_to_json(form_id, export_options)
+
+    outfile = response or archive_name
+    with zipfile.ZipFile(outfile, "w") as zip_file:
+        for name, data in resources.items():
+            zip_file.writestr(f"{name}.json", data)
+    return outfile
+
+
+def form_to_json(form_id: int, export_options: FormExportOptions) -> dict:
```
sergei-maertens, line 49: A bare `dict` type hint really doesn't say anything at all.

src/openforms/forms/import_export/export_form.py
```diff
@@ -0,0 +1,130 @@
...
+    form = Form.objects.get(pk=form_id)
+
+    # Reset the submission counter
+    form.submission_counter = 0
```
sergei-maertens, line 53: A `resource.prepare_for_export(export_options)` wouldn't be a bad pattern I think? We can enforce that interface, while every model type/serializer then knows what it needs to do itself, and this way we build in export support first class instead of it being tacked on and breaking when runtime semantics get updated and people forget about export.

Potentially there could be some kind of intermediate data structure and some developer-time checks (perhaps with typing and pattern matching exhaustiveness?) to not pollute the model classes while still being robust?

src/openforms/forms/import_export/export_form.py
```diff
@@ -0,0 +1,130 @@
...
+
+    form_steps = FormStep.objects.filter(form__pk=form_id).select_related(
+        "form_definition"
+    )
+
+    form_definitions = FormDefinition.objects.filter(
+        pk__in=form_steps.values_list("form_definition", flat=True)
+    )
```
sergei-maertens, line 61: what's the point of using `select_related` if you're going to do a separate query anyway? This now looks like some odd manual prefetch_related thing.

but instead, it could be:

```py
form_definitions = [form_step.form_definition for form_step in form_steps]
```

src/openforms/forms/import_export/export_form.py
```diff
@@ -0,0 +1,130 @@
...
+
+    request = _get_mock_request()
+
+    forms = [
+        FormExportSerializer(
+            instance=form,
+            context={"request": request, "export_options": export_options},
+        ).data
+    ]
+    form_definitions = FormDefinitionExportSerializer(
```
sergei-maertens, line 80: I don't see `is_exporting` being set anywhere?

src/openforms/forms/import_export/export_form.py
```diff
@@ -0,0 +1,130 @@
...
+        instance=form_definitions,
+        many=True,
+        context={
+            "request": request,
+            "export_options": export_options,
+            "form": form,
+        },
```
sergei-maertens, line 87: I would definitely create the context dict only once and pass the same thing to all serializers, which is how DRF-experienced developer expect `serializer.context` to work.

src/openforms/forms/import_export/typing.py
```diff
@@ -0,0 +1,56 @@
...
+from dataclasses import dataclass, field
+from typing import TypedDict
+
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from openforms.typing import JSONObject
+
+
+class FormConfigurationOptions(models.TextChoices):
```
sergei-maertens, line 11: these are not types, they are `constants.py` constants :)

src/openforms/forms/import_export/typing.py
```diff
@@ -0,0 +1,56 @@
...
+class AdditionalFormConfigurationOptions(models.TextChoices):
+    product = "product", _("Product")
+    wms_tile_layers = "wmsTileLayers", _("WMS-tile layers")
+    wmts_tile_layers = "wmtsTileLayers", _("Background tile layers")
+    yivi_attribute_groups = "yiviAttributeGroups", _("Yivi attribute groups")
+
+
+class FormExportOptionsData(TypedDict, total=False):
+    remove_sensitive_content: bool
+    form_configuration: list[FormConfigurationOptions]
```
sergei-maertens, line 27: does it have to be a `list`, or is a `Sequence` sufficient? What's the actual interface you need?

src/openforms/forms/import_export/typing.py
```diff
@@ -0,0 +1,56 @@
...
+    additional_form_configuration: list[AdditionalFormConfigurationOptions]
+
+
+@dataclass(slots=True)
+class FormExportOptions:
```
sergei-maertens, line 32: I'm not sure I see the benefit of a data class if you have to defined and work with `FormExportOptionsData` anyway.

src/openforms/forms/import_export/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    WMTSTileLayerResource,
+    YiviAttributeGroupResource,
+)
+from .typing import AdditionalFormConfigurationOptions, FormExportOptions
+
+
+@dataclass(frozen=True)
+class ExportResourceConfig:
+    resource: type[BaseResource]
+    output_name: str
```
sergei-maertens, line 19: I'd actually be tempted to avoid this dataclass overhead and code complexity and use plain tuples?

```py
type ExportResourceConfig = tuple[type[BaseResource], str]
```

src/openforms/forms/import_export/utils.py
sergei-maertens, line 1: Please do not name this module `utils.py`, it reminds me of `openforms.forms.utils` which _you are fixing_.

It's a module dealing witht he additional resource exports, name it so that is clear.

src/openforms/forms/import_export/utils.py
sergei-maertens: I have to admit, I'm not sure where the crossover point is to use data class vs a plain tuple - I think it mostly depends on how "small" the tuple is and how the calling code uses it - in this case you'd expect this to be immediately destructured:

```py
for option, (resource, output_name) in ADDITIONAL_FORM_CONFIGURATION_RESOURCES.items():
```
which is nice and concise

src/openforms/forms/templates/admin/forms/form/export.html
sergei-maertens, line 1: I was thinking "why is this still a plain django page when we do have the endpoint" but it's probably for the bulk export, right?

src/openforms/forms/tests/admin/test_form.py
sergei-maertens, line 1: Can you rename this test module to make clear it's testing import/export rather than the forms admin?

src/openforms/forms/tests/factories.py
```diff
@@ -3,15 +3,22 @@
...
+    LogicActionTypes,
+    SubmissionAllowedChoices,
+)
+from openforms.forms.import_export.export_form import form_to_json
+from openforms.forms.import_export.typing import (
+    AdditionalFormConfigurationOptions,
+    FormConfigurationOptions,
+    FormExportOptions,
+)
+from openforms.forms.models import Form, FormDefinition, FormStep, FormVariable
```
sergei-maertens, line 17: please use relative imports within the same python package

sergei-maertens (changes requested):
Lot's of code added, but not a single module seems to have been added to the pyright config to actually enable type checking in CI. This way we'll just keep cleaning up type errors in the backend forever :(

I've reviewed most of the backend code so far - I notice I can't focus any more on the remaining tests so that'll happen at a later stage. Same for the JS code.

src/openforms/forms/templates/admin/forms/form/export.html
robinmolen: Yes, thats the bulk-export template

src/openforms/forms/tests/import_export/test_resources.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class ProductResourceTests(TestCase):
+    def test_export_for_form(self):
+        product = ProductFactory.create()
+        form = FormFactory.create(product=product)
+
+        dataset = ProductResource().export_for_form(form).dict
+
+        self.assertEqual(len(dataset), 1)
```
sergei-maertens, line 25: I think you want to add another project in the test setup here and assert that that second unrelated product is not included in the export data.

This pattern should be applied for all resource exports that use `export_for_form`

src/openforms/forms/tests/import_export/test_resources.py
```diff
@@ -0,0 +1,269 @@
...
+                            },
+                        ],
+                    },
+                ],
+            },
+        )
+
+        dataset = WMSTileLayerResource().export_for_form(form).dict
+
+        self.assertEqual(len(dataset), 0)
```
sergei-maertens, line 126: here too you want to make sure the resource itself exists in the database - this export will always be empty even with a "valid" form component configuration, because there simply is nothing to export

src/openforms/forms/tests/import_export/test_resources.py
```diff
@@ -0,0 +1,269 @@
...
+                            "polyline": False,
+                        },
+                    },
+                ],
+            },
+        )
+
+        dataset = WMSTileLayerResource().export_for_form(form).dict
+
+        self.assertEqual(len(dataset), 0)
```
sergei-maertens, line 150: same remark as above here

src/openforms/forms/tests/import_export/test_resources.py
```diff
@@ -0,0 +1,269 @@
...
+            str(product.price).replace(".", ","),
+        )
+        self.assertEqual(dataset[0]["information"], product.information)
+
+    def test_export_for_form_without_product(self):
+        form = FormFactory.create(product=None)
+
+        dataset = ProductResource().export_for_form(form).dict
+
+        self.assertEqual(len(dataset), 0)
```
sergei-maertens, line 39: an export not specific for a form will also pass this assertion - the test data setup is missing some actual records that should be proven to be ignored :)

src/openforms/forms/tests/test_api_import_export.py
```diff
@@ -134,6 +154,744 @@ def test_form_export_staff_required(self):
...
+        url = reverse("api:form-export", args=(form.uuid,))
+        response = self.client.post(
+            url,
+            format="json",
+            HTTP_AUTHORIZATION=f"Token {self.token.key}",
+            # Invalid value for the remove_sensitive_content option
+            data={"remove_sensitive_content": "whatever"},
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
sergei-maertens, line 173: you also want to assert at least on the `invalidParams` `name` and `code` properties, as you may have a 400 for other reasons than you assume in the test (e.g. a new required field is added to the serializer but the test is not updated -> it will pass, but no longer tests the same contract)

src/openforms/forms/tests/test_api_import_export.py
```diff
@@ -134,6 +154,744 @@ def test_form_export_staff_required(self):
...
+                "form_configuration": [
+                    FormConfigurationOptions.registration_backends,
+                ],
+            },
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        zf = ZipFile(BytesIO(response.content))
+        self.assertEqual(
```
sergei-maertens, line 223: I think I'd also like a follow up ticket/issue to improve the DX of testing import/export, because all this manual zip file wrangling in tests distracts so much from what's actually being tested and it's very verbose

src/openforms/forms/tests/test_api_import_export.py
```diff
@@ -134,6 +154,744 @@ def test_form_export_staff_required(self):
...
+        forms = json.loads(zf.read("forms.json"))
+        form_logic = json.loads(zf.read("formLogic.json"))
+        form_variables = json.loads(zf.read("formVariables.json"))
+
+        self.assertEqual(len(forms), 1)
+        self.assertEqual(len(form_logic), 1)
+        self.assertEqual(len(form_variables), 2)
+
+        # Internal remarks should be removed
+        self.assertIsNone(forms[0].get("internal_remarks"))
```
sergei-maertens, line 244: I think I'd assert that the key is not present, now the export data can contain the key with a value of `None` and the test will pass, but it being a text-based field, it should not be nullable at the serializer level, so it will throw validation errors.

src/openforms/js/components/admin/form_design/Context.js
```diff
@@ -9,7 +9,14 @@ const FeatureFlagsContext = React.createContext({
 FeatureFlagsContext.displayName = 'FeatureFlagsContext';
 
 const FormContext = React.createContext({
-  form: {url: '', uuid: '', type: 'regular'},
+  form: {
+    url: '',
+    uuid: '',
+    type: 'regular',
+    authBackends: [],
+    paymentBackend: [],
```
sergei-maertens, line 17: with the v3 endpoint rework, this is now in the state as `payment: {backend: '', options: {}}` so I'd recommend using the same structure if it's not too complicated.

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1267,6 +1267,8 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl, outgoingRequestsUr
             uuid: state.form.uuid,
             type: state.form.type,
             authBackends: state.form.authBackends,
+            paymentBackend: state.form.paymentBackend,
```
sergei-maertens, line 1270:
```suggestion
            paymentBackend: state.form.payment.backend,
```
when you rebase on `main`

src/openforms/js/components/admin/form_design/FormSubmit.js
sergei-maertens: I'm not sure why you're using refs here, couldn't you do:

* keep the export options in plain `useState` and call `setState` on modal confirm
* have the hidden `export_options` input `defaultValue={JSON.stringify(serializeExportOptions(exportOptions)}` that uses the state

I'm not too fond of the `exportButtonRef` either, but the alternatives aren't much better (somehow looking up the `form` object to call `form.submit()` or using a DOM query based on form ID from Django...)

src/openforms/js/components/admin/form_export/AdditionalFormConfigurationField.js
```diff
@@ -0,0 +1,77 @@
...
+    description: 'Label for additionalFormConfiguration yiviAttributeGroups input',
+    defaultMessage: 'Yivi attribute groups',
+  },
+});
+
+const AdditionalFormConfigurationField = () => {
+  const options = useAdditionalFormConfigurationOptions();
+  const [fieldProps, , fieldHelpers] = useField('additionalFormConfiguration');
+  const {value: selected} = fieldProps;
+  const {setValue} = fieldHelpers;
```
sergei-maertens, line 33:
```suggestion
  const [{value: selected}, , {setValue}] = useField('additionalFormConfiguration');
```
for consistency with other places where we use Formik?

src/openforms/js/components/admin/form_export/useExportOptions.js
```diff
@@ -0,0 +1,62 @@
...
+ * Get a list of options that should be included in the form configuration. These options
+ * represent key parts of the form configuration.
+ *
+ * The options are based on the form configuration. For example, if the form doesn't have
+ * any registration backends, then the registration backends option is not included.
+ *
+ * @returns {String[]}
+ */
+export const useFormConfigurationOptions = () => {
+  const {registrationBackends, form, formVariables} = useContext(FormContext);
```
sergei-maertens, line 15: do we really want to use the indirection of context here? As far as I can see it, it's pretty straight forward to pass `state.form` to the `FormSubmit` component as a prop, which can pass it to the hook.

That considerably simplifies reasoning about the implementation and testing in storybook where you can now simply use `args` to pass the necessary props and don't have to mess with context (especially unrelated context).

src/openforms/js/components/admin/form_export/utils.js
sergei-maertens, line 1: this is only used in one place, I think it makes more sense to define that function in the same module where it's called instead of putting it in a generic `utils` module.

src/openforms/js/components/admin/form_export/ExportOptionsModal.js
```diff
@@ -0,0 +1,114 @@
...
+      }
+      closeModal={onCloseModal}
+      extraModifiers={['large']}
+    >
+      <Formik
+        initialValues={{
+          removeSensitiveContent: true,
+          formConfiguration: formConfigurationOptions,
+          additionalFormConfiguration: [],
+        }}
```
sergei-maertens, line 45: do we want to always reset this when the modal closes/opens? Or do we want to keep configuration and feed it in from the state of the parent component (which is currently a ref)?

For consistency I'd say reset every time, as that's what happens now in other modals - changes are only "persisted" once you confirm.

src/openforms/js/lang/nl.json
```diff
@@ -1552,11 +1583,21 @@
...
+    "description": "Label for formConfiguration paymentBackend input",
+    "originalDefault": "Payment provider"
+  },
   "OX8wxE": {
     "defaultMessage": "Uw sessie vervalt {delta}. Verleng uw sessie indien u wenst door te gaan.",
     "description": "Session expiry warning (in modal)",
     "originalDefault": "Your session is about to expire {delta}. Extend your session if you wish to continue."
   },
+  "OaS7Pf": {
+    "defaultMessage": "Formulier exportopties",
```
sergei-maertens, line 1597:
```suggestion
    "defaultMessage": "Formulierexportopties",
```

src/openforms/js/lang/nl.json
```diff
@@ -2084,6 +2130,11 @@
     "description": "Email registration options 'emailContentTemplateHtml' helpText",
     "originalDefault": "Content of the registration email message (as html)."
   },
+  "Z4/E9e": {
+    "defaultMessage": "Authenticatie plugins",
```
sergei-maertens, line 2134:
```suggestion
    "defaultMessage": "Authenticatieplugins",
```

src/openforms/js/lang/nl.json
```diff
@@ -2677,6 +2743,11 @@
     "isTranslated": true,
     "originalDefault": "Label"
   },
+  "iAFlLD": {
+    "defaultMessage": "Registratie backends",
```
sergei-maertens, line 2747: should be a single word :)

src/openforms/js/lang/nl.json
```diff
@@ -2957,6 +3034,11 @@
     "description": "Fixed metadata table name title",
     "originalDefault": "Name"
   },
+  "n55FhL": {
+    "defaultMessage": "Beveiliging",
```
sergei-maertens, line 3038: I'm not sure I like the "beveiliging" or "security" terms here. I think it's more privacy-oriented, but that also doesn't feel right.

Maybe more somsething like "Organisation specific details"?

src/openforms/js/lang/nl.json
```diff
@@ -3469,6 +3551,11 @@
     "description": "JSON editor: object entry has empty key name",
     "originalDefault": "Set a key name before you can configure this variable"
   },
+  "w2UNV6": {
+    "defaultMessage": "Yivi attribuut groepen",
```
sergei-maertens, line 3555:
```suggestion
    "defaultMessage": "Yivi-attribuutgroepen",
```

sergei-maertens (changes requested):

PR #6490, opened 2026-07-24
[#5958] Add support for logic rules in v3 Form serializer

### LINKED ISSUE #5958: Add support for logic rules

vaszig:
Closes #5958

**Changes**

- Add new field `logic_rules` to the new v3 form endpoint
- Add validation concerning the logic rules and actions
- Add tests

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openapi.yaml
+++ b/src/openapi.yaml
@@ -8500,6 +8500,49 @@ components:
       required:
       - actions
       - jsonLogicTrigger
+    FormLogicV3:
+      type: object
+      description: Represent logic rules with their actions as a nested serializer.
+      properties:
+        uuid:
+          type: string
+          format: uuid
+          readOnly: true
+        jsonLogicTrigger:
+          type: object
+          title: JSON logic
+          description: The trigger expression to determine if the actions should execute
+            or not. Note that this must be a valid JsonLogic expression, and the first
+            operand must be a reference to a variable in the form.
+          additionalProperties: true
+          example:
+            ==:
+            - var: checkbox
+            - true
+        description:
+          type: string
+          description: Logic rule description in natural language.
+          maxLength: 100
+        order:
+          type: integer
+          description: Order of the rule relative to the form it belongs to. Logic
+            rules are evaluated in this order. Note that specifying a value for the
+            order here will cause the rules before/after this rule to be re-calculated.
+        actions:
+          type: array
+          items:
+            $ref: '#/components/schemas/LogicComponentActionV3'
+          description: Actions triggered when the trigger expression evaluates to
+            'truthy'.
+        isAdvanced:
+          type: boolean
+          description: Is this an advanced rule (the admin user manually wrote the
+            trigger as JSON)?
+      required:
+      - actions
+      - jsonLogicTrigger
+      - order
+      - uuid
     FormModelTranslations:
       type: object
       description: |-
@@ -8899,6 +8942,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/FormStepV3'
+        logicRules:
+          type: array
+          items:
+            $ref: '#/components/schemas/FormLogicV3'
         showProgressIndicator:
           type: boolean
           description: Whether the step progression should be displayed in the UI
@@ -9539,6 +9586,38 @@ components:
           description: De UUID van de formulierstap waarop de actie van invloed is.
             Dit veld is verplicht voor actietypen `disable-next`, `step-applicable`,
             `step-not-applicable` - anders is het optioneel.
+        action:
+          $ref: '#/components/schemas/LogicActionPolymorphic'
+      required:
+      - action
+      - uuid
+    LogicComponentActionV3:
+      type: object
+      properties:
+        uuid:
+          type: string
+          format: uuid
+          readOnly: true
+        component:
+          type: string
+          title: Form.io component key
+          description: 'Sleutel van de Form.io-component waarop de actie van toepassing
+            is. Dit veld is verplicht voor de actietypes `property` - anders is het
+            optioneel. '
+          pattern: ^(\w|\w[\w.\-]*\w)$
+        variable:
+          type: string
+          title: Key of the target variable
+          description: Sleutel van de variabele die aangepast wordt door de actie.
+            Dit veld is verplicht voor de actietypes `variable` - anders is het optioneel.
+          pattern: ^(\w|\w[\w.\-]*\w)$
+        formStepSlug:
+          type: string
+          description: The slug of the form step that will be affected by the action.
+            This field is required for action types `disable-next`, `step-applicable`,
+            `step-not-applicable` - otherwise it's optional. We deliberately use the
+            slug instead of the uuid because the uuid is not fixed, as we create from
+            scratch the form steps dusring form update.
         action:
           $ref: '#/components/schemas/LogicActionPolymorphic'
       required:
--- a/src/openforms/forms/api/v3/serializers/form.py
+++ b/src/openforms/forms/api/v3/serializers/form.py
@@ -1,4 +1,6 @@
 from collections import Counter, defaultdict
+from collections.abc import Collection, Mapping
+from datetime import date
 from uuid import UUID
 
 from django.db import transaction
@@ -6,6 +8,8 @@ from django.utils.text import get_text_list
 from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.utils import extend_schema_serializer
+from glom import glom
+from json_logic.typing import Primitive
 from rest_framework import serializers
 from rest_framework.exceptions import ErrorDetail, ValidationError
 
@@ -17,6 +21,7 @@ from openforms.formio.service import (
     DuplicateKeyError,
     FormioConfigurationWrapper,
     get_branch_representation,
+    holds_submission_data,
 )
 from openforms.formio.typing.base import Component
 from openforms.prefill.contrib.customer_interactions.constants import (
@@ -25,7 +30,7 @@ from openforms.prefill.contrib.customer_interactions.constants import (
 from openforms.products.models import Product
 from openforms.translations.api.serializers import ModelTranslationsSerializer
 from openforms.typing import StrOrPromise
-from openforms.variables.constants import FormVariableSources
+from openforms.variables.constants import FormVariableDataTypes, FormVariableSources
 from openforms.variables.models import ServiceFetchConfiguration
 from openforms.variables.service import get_static_variables
 
@@ -36,18 +41,27 @@ from ....api.serializers.form import (
     HelpDialogSerializer,
     SubmissionsRemovalOptionsSerializer,
 )
-from ....constants import FormTypeChoices
+from ....constants import (
+    LOGIC_ACTION_TYPES_REQUIRING_COMPONENT,
+    LOGIC_ACTION_TYPES_REQUIRING_FORM_STEP_SLUG,
+    LOGIC_ACTION_TYPES_REQUIRING_VARIABLE,
+    FormTypeChoices,
+    LogicActionTypes,
+)
+from ....logic_analysis import CyclesDetected, analyze_rules
 from ....models import (
     Category,
     Form,
     FormDefinition,
+    FormLogic,
     FormRegistrationBackend,
     FormStep,
     FormVariable,
 )
 from ...validators import RequireAppointmentsPlugin
-from ..typing import FormStepData, FormValidatedData, FormVariableData
+from ..typing import FormLogicData, FormStepData, FormValidatedData, FormVariableData
 from .form_step import FormStepSerializer
+from .logic_rules import FormLogicSerializer
 from .payment import FormPaymentSerializer
 from .variables import FormVariableSerializer
 
@@ -77,6 +91,7 @@ class FormSerializer(serializers.ModelSerializer):
     variables = FormVariableSerializer(
         many=True, source="formvariable_set", required=False
     )
+    logic_rules = FormLogicSerializer(many=True, required=False, source="formlogic_set")
 
     payment = FormPaymentSerializer(required=False, source="*")
 
@@ -109,6 +124,7 @@ class FormSerializer(serializers.ModelSerializer):
         "confirmation_email_template",
         "formstep_set",
         "formvariable_set",
+        "formlogic_set",
         "registration_backends",
     )
 
@@ -145,6 +161,7 @@ class FormSerializer(serializers.ModelSerializer):
             "category",
             "theme",
             "steps",
+            "logic_rules",
             "show_progress_indicator",
             "show_summary_progress",
             "maintenance_mode",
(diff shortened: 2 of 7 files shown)


src/openforms/forms/api/v3/typing.py
```diff
@@ -140,6 +141,16 @@ class FormVariableData(TypedDict):
     service_fetch_configuration: NotRequired[ServiceFetchConfigurationData | None]
 
 
+class FormLogicData(TypedDict):
+    uuid: str
+    description: NotRequired[str]
+    json_logic_trigger: JSONObject
+    actions: JSONObject
```
sergei-maertens:
```suggestion
    actions: list[JSONObject]
```
I'd expect?

Though I believe we should be able to fully define the action typed dicts for each type and end up with a union type, instead of the unspecified `JSONObject`

src/openforms/forms/api/v3/typing.py
```diff
@@ -140,6 +141,16 @@ class FormVariableData(TypedDict):
...
+    is_advanced: bool
+    # DeprecationWarning: remove in OF 4.1
+    trigger_from_step: FormStep
```
sergei-maertens: I think we can already omit this here in the v3 serializer, because it's not used in import/export

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -3294,6 +3316,922 @@ def test_profile_form_variable_unknown_key(self):
...
+        self.assertEqual(variables[1].source, FormVariableSources.user_defined)
+        self.assertIsNone(variables[1].form_definition)
+        self.assertEqual(variables[1].data_type, FormVariableDataTypes.string)
+
+        # 2nd component variable, generated for the form step (based on the form defintion)
+        self.assertEqual(variables[2].name, "TextField")
+        self.assertEqual(variables[2].key, "textField")
+        self.assertEqual(variables[2].source, FormVariableSources.component)
+        self.assertEqual(variables[2].form_definition.uuid, form_definition_uuid)
+        self.assertEqual(variables[2].data_type, FormVariableDataTypes.string)
```
sergei-maertens: the test case is about logic rules - why are you asserting on variables?

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -3294,6 +3316,922 @@ def test_profile_form_variable_unknown_key(self):
...
+                                "internalName": "Form configuratie",
+                            },
+                        },
+                    },
+                },
+            ],
+            "logic_rules": [
+                {
+                    "order": 0,
+                    "jsonLogicTrigger": {"==": [{"var": "checkbox"}, True]},
```
sergei-maertens: I'd expect we can simplify these tests that verify action validation by removing the unnecessary components (like checkbox and textfield in this case)


```suggestion
                    "jsonLogicTrigger": True,
```

a simple trigger like this should make that possible

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -3294,6 +3316,922 @@ def test_profile_form_variable_unknown_key(self):
...
+        }
+        response = self.client.put(url, data=data)
+        response_data = response.json()
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_data["code"], "invalid")
+        self.assertEqual(
+            response_data["invalidParams"][0],
+            {
+                "name": "logicRules.0.actions.0.value",
```
sergei-maertens: shouldn't this have an extra level of nesting?


```suggestion
                "name": "logicRules.0.actions.0.action.value",
```

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -3294,6 +3316,922 @@ def test_profile_form_variable_unknown_key(self):
...
+                        "translations": {
+                            "en": {
+                                "name": "Form configuration",
+                                "internalName": "Form configuration",
+                            },
+                            "nl": {
+                                "name": "Form configuratie",
+                                "internalName": "Form configuratie",
+                            },
+                        },
```
sergei-maertens: translations can also be omitted, it's not relevant for the test

src/openforms/forms/tests/v3/test_endpoints.py
```diff
@@ -3294,6 +3316,922 @@ def test_profile_form_variable_unknown_key(self):
...
+        response_data = response.json()
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response_data["code"], "invalid")
+        self.assertEqual(
+            response_data["invalidParams"][0],
+            {
+                "name": "logicRules.0.actions.0.formStepSlug",
+                "code": "null",
+                "reason": "This field may not be null.",
```
sergei-maertens, line 3929: the code/message is not entirely correct considering that `""` was passed rather than `null`.

src/openforms/forms/constants.py
```diff
@@ -38,6 +38,11 @@ def get_label(cls, value: str) -> str:
     LogicActionTypes.step_not_applicable.value,
     LogicActionTypes.disable_next.value,
 }
+LOGIC_ACTION_TYPES_REQUIRING_FORM_STEP_SLUG: Set[str] = {
```
sergei-maertens:
```suggestion
LOGIC_ACTION_TYPES_REQUIRING_FORM_STEP_SLUG: LOGIC_ACTION_TYPES_REQUIRING_FORM_STEP_UUID
```
or rather swap the variables around, as UUID is the old situation and slug the new one?

They are semantically identical, so if they get edited/updated, both constants should change together

src/openapi.yaml
```diff
@@ -8484,6 +8484,52 @@ components:
...
+      description: |-
+        A ModelSerializer to provide a serializer that can be update and create
+        objects in a specific order.
+
+        Typically a `models.PositiveIntegerField` field called `order` is used to
+        store the order of the Model objects. This field can be customized by setting
+        the `order_field_name` attribute on the Model class.
+
+        This serializer will move the object to the correct
+        order if the ordering field is passed in the validated data.
```
sergei-maertens: Can we do something about this generated description? Define our own docstring, as I believe that should be picked up instead?

src/openapi.yaml
```diff
@@ -8484,6 +8484,52 @@ components:
...
+      properties:
+        uuid:
+          type: string
+          format: uuid
+          readOnly: true
+        jsonLogicTrigger:
+          title: JSON logic
+          description: The trigger expression to determine if the actions should execute
+            or not. Note that this must be a valid JsonLogic expression, and the first
+            operand must be a reference to a variable in the form.
```
sergei-maertens, line 8516: Can we add some extra type information here? I think https://drf-spectacular.readthedocs.io/en/latest/customization.html#step-3-extend-schema-field-and-type-hints will take you a long way

src/openforms/forms/api/v3/serializers/logic_rules.py
```diff
@@ -0,0 +1,109 @@
...
+        required=False,
+        allow_blank=True,
+        label=_("Form.io component"),
+        help_text=_(
+            "Key of the Form.io component that the action applies to. This field is "
+            "required for the action types {action_types} - otherwise it's optional."
+        ).format(
+            action_types=_join_action_types(LOGIC_ACTION_TYPES_REQUIRING_COMPONENT)
+        ),
+    )
```
sergei-maertens: Look, we have a dedicated field for this :-) https://github.com/open-formulieren/open-forms/blob/0bbb02fb79bcae7b84d2796f4b36574c4d47c79b/src/openforms/formio/api/fields.py#L8

src/openforms/forms/api/v3/serializers/logic_rules.py
```diff
@@ -0,0 +1,109 @@
...
+    variable = serializers.CharField(
```
sergei-maertens: this should probably also use `FormioVariableKeyField` as the key naming constraints are the same

src/openforms/forms/api/v3/serializers/logic_rules.py
```diff
@@ -0,0 +1,109 @@
+import uuid
+
+from django.utils.translation import gettext_lazy as _
+
+from drf_spectacular.utils import extend_schema_serializer
+from ordered_model.serializers import OrderedModelSerializer
+from rest_framework import serializers
+
+from ....api.serializers.logic.action_serializers import (
```
sergei-maertens:
```suggestion
from ...serializers.logic.action_serializers import (
```
I think this should be the proper relative path - we're already in the `api` subpackage

src/openforms/forms/api/v3/serializers/logic_rules.py
```diff
@@ -0,0 +1,109 @@
...
+        label=_("Key of the target variable"),
+        help_text=_(
+            "Key of the target variable whose value will be changed. This field is "
+            "required for the action types {action_types} - otherwise it's optional."
+        ).format(
+            action_types=_join_action_types(LOGIC_ACTION_TYPES_REQUIRING_VARIABLE)
+        ),
+    )
+    form_step_slug = serializers.CharField(
+        allow_null=True,
```
sergei-maertens: can we make this `False` and require the frontend to send an empty string instead of `null`?

src/openforms/forms/api/v3/serializers/logic_rules.py
```diff
@@ -0,0 +1,109 @@
...
+        allow_blank=True,
+        required=False,
+        label=_("form step slug"),
+        help_text=_(
+            "The slug of the form step that will be affected by the action. This field "
+            "is required for action types {action_types} - otherwise it's optional. We "
+            "deliberately use the slug instead of the uuid because the uuid is not fixed, "
+            "as we create from scratch the form steps dusring form update."
+        ).format(
+            action_types=_join_action_types(LOGIC_ACTION_TYPES_REQUIRING_FORM_STEP_UUID)
```
sergei-maertens: you probably want to use the new slug-directed constant here instead of `FORM_STEP_UUID`

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -178,32 +422,43 @@ def create(self, validated_data: FormValidatedData) -> Form:
...
+                    **{
+                        **step_data,
+                        "form_definition": form_definition,
+                        "form": instance,
+                        "order": index,
+                    }
+                )
             )
 
+        form_steps = FormStep.objects.bulk_create(form_steps)
```
sergei-maertens, line 342: nice ⚡

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
             "type": {"validators": [RequireAppointmentsPlugin()]},
         }
 
+    def _validate_actions(self, form: Form, temp_rules_instances: dict[FormLogic, int]):
```
sergei-maertens:
```suggestion
    def _validate_actions(self, form: Form, temp_rules: Mapping[FormLogic, int]):
```
the type hint makes it clear that we're working with instances, so we can use a shorter variable

Additionally, using `Mapping` instead of `dict` emphasizes that `temp_rules` is not mutable (we can still mutate the rules themselves though)

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+                form_step_slug = action.get("form_step_slug")
+                variable = action.get("variable")
+
+                # validate component exists in action
+                if (
+                    action_type
+                    and action_type in LOGIC_ACTION_TYPES_REQUIRING_COMPONENT
+                    and not component
+                ):
+                    raise serializers.ValidationError(
```
sergei-maertens: it would be good to gather all the validation errors for all actions / rules and only emit them at the end, because we can validate a lot of things "in parallel" rather than bailing out as soon as we encounter any error.

Probably with a default dict you'll get pretty far:

```py
type ActionErrors = defaultdict[int, list[dict[str, ErrorDetail]]]  # action index to list of property-specific errors

rule_errors: defaultdict[int, ActionErrors]
```

the nice thing about these default dicts is that you can have sparse data structures - only the indices with errors get added (and you can assign them as a mapping), and then at the end convert that into the `ValidationErrors` structure with only the relevant errors without having to deal with lists/empty indices.

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        code="blank",
+                    )
+
+                # validate variable exists in form
+                if action_type == LogicActionTypes.variable:
```
sergei-maertens: should this be limited to only this action type, or target `action_type in LOGIC_ACTION_TYPES_REQUIRING_VARIABLE`? I'm leaning towards the latter, because if you require a variable reference, you want that reference to be valid. So I think this check should be part of the previous `if` block.

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+                    form_var = form_variables.get(variable)
+                    if form_var is None:
```
sergei-maertens:
```suggestion
                    if form_var not in form_variables:
```
no need for the lookup, right?

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+                                            str(action_index): {
+                                                "variable": _(
+                                                    "The variable {varname} does not exist."
+                                                ).format(varname=variable),
+                                            }
+                                        }
+                                    }
+                                }
+                            },
+                            code="blank",
```
sergei-maertens:
```suggestion
                            code="invalid",
```
it's not empty, it's wrong 😄

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+                    and not variable
+                ):
+                    raise serializers.ValidationError(
+                        {
+                            "logic_rules": {
+                                str(rule.order): {
+                                    "actions": {
+                                        str(action_index): {
+                                            "variable": _(
+                                                "This field may not be blank."
```
sergei-maertens:
```suggestion
                                                "You must specify a variable."
```
a more direct error message is definitely more useful

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+                                                )
+                                            }
+                                        }
+                                    }
+                                }
+                            },
+                            code="null",
+                        )
+                    # validate form step slug is valid
+                    if not form.formstep_set.filter(slug=form_step_slug).exists():
```
sergei-maertens: I don't think we need to do this query every time, but can instead build a collection upfront in one query of all the known slugs (similar to the variables)

src/openforms/forms/api/v3/serializers/form.py
sergei-maertens: The implementation is quite visibly overloading and long (making it harder to review). I'd consider breaking out the various action-type specific validations into little helper functions and wire them up in this method with some pattern matching:

```py
match action:
    case {"type": LogicActionTypes.variable}:
        ...
    case {"type": str(action_type)} if action_type in LOGIC_ACTION_TYPES_REQUIRING_COMPONENT:
        errors = _check_component_action_definition(...)
```

together with the collection of all errors, you can cut down on a lot of boilerplate and avoid repeating this block all the time as well:

```

                            "logic_rules": {
                                str(rule.order): {
                                    "actions": {
                                        str(action_index): {
                                            ...
                                        }
                                    }
                                }
                            }
```

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -158,12 +175,239 @@ class Meta:  # pyright: ignore[reportIncompatibleVariableOverride]
...
+
+    def _validate_and_process_logic_rules(
+        self,
+        form: Form,
+        logic_rules_raw: list[FormLogicData],
+        temp_logic_rules: dict[FormLogic, int],
+    ):
+        first_step: FormStep = min(
+            form.form_step_map.values(), key=lambda step: step.order
+        )
```
sergei-maertens: there was a crash recently w/r to this kind of validation running for appointment forms or forms that didn't have any steps at all - this should now not happen anymore because we already have the earlier validation about the number of steps per component type, right?

sergei-maertens (changes requested):
Quite a number of remarks, but in general I think you managed this well - the current codebase doesn't allow for a much cleaner implementation :-)

src/openforms/forms/constants.py
vaszig: I will go with your suggestion and keep the old one too for v2. When we remove v2 we can then clean this up I guess. For now I have added a comment.

src/openforms/forms/api/v3/serializers/form.py
vaszig: This is executed only if there are logic rules. If there are and we have no steps then we get an `AssertionError` from logic analysis (here I should put a default value of `None` which I forgot).

sergei-maertens (approved):

PR #6491, opened 2026-07-24
Concept of Broncodebeheer statement

joeribekker:
This is the approved statement by the G4, Bregje and me.

[skip: e2e]

### DIFF
--- a/PROJECT_GOVERNANCE.md
+++ b/PROJECT_GOVERNANCE.md
@@ -34,7 +34,7 @@ have the following responsibilities (summary):
 * Coordinating security issues and bug report follow-ups
 * Coordinating responses to inquiries
 * In the event of prolonged discussions or disputes, maintainers (among others)
-  may request a decision from the community council.
+  may request a decision from the SC.
 
 The current maintainer of this project is [Maykin](https://www.maykin.nl). For
 more in-depth information about the tasks, activities, responsibilities and 
--- a/PROJECT_GOVERNANCE.md
+++ b/PROJECT_GOVERNANCE.md
@@ -1,0 +1,56 @@
+# Project Governance
+
+The fundamental principle is that decisions are based on consensus within the 
+community. If a decision-making process takes too long or a decision is urgently
+required, the "Steering Committee" has the authority to make 
+decisions.
+
+## Steering Committee
+
+The Steering Committee (SC) is responsible for:
+
+1. The general ambitions and objectives of this project
+2. Guidelines, procedures, and tool selection
+3. Decisions regarding architecture and (development) infrastructure
+4. Raising issues relevant to the direction and development of this project
+
+The SC consists of the following members:
+
+1. Wieteke Kremer
+2. Bregje van der Horst
+3. Raoul Lander
+4. Stef Schots
+5. Agnesa Krasniqi
+6. Joeri Bekker
+
+## Maintainers
+
+Maintainers are responsible for maintaining parts of the codebase. Maintainers 
+have the following responsibilities (summary):
+
+* Coordinating development activities
+* Ensuring code and documentation reviews are performed
+* Coordinating Pull Requests
+* Coordinating security issues and bug report follow-ups
+* Coordinating responses to inquiries
+* In the event of prolonged discussions or disputes, maintainers (among others)
+  may request a decision from the community council.
+
+The current maintainer of this project is [Maykin](https://www.maykin.nl). For
+more in-depth information about the tasks, activities, responsibilities and 
+financing, please see
+[Broncodebeheer](https://maykinmedia.github.io/maykin-plus/broncodebeheer).
+
+Any community member or contributor can ask a question or report an issue by
+opening a GitHub issue (see the [Contribution Guide](CONTRIBUTING.md)).
+
+## Contributors
+
+"Contributors" refers to anyone in the community who contributes code, 
+documentation, or other technical artifacts to the project.
+
+Anyone can become a contributor.
+
+There is no expectation of commitment to the project, no specific skill 
+requirements, and no selection process. To become a contributor, a community 
+member simply needs to perform one or more actions that benefit the project.
--- a/PROJECT_GOVERNANCE.md
+++ b/PROJECT_GOVERNANCE.md
@@ -11,7 +11,7 @@ The Steering Committee (SC) is responsible for:
 
 1. The general ambitions and objectives of this project
 2. Guidelines, procedures, and tool selection
-3. Decisions regarding architecture and (development) infrastructure
+3. Decisions regarding information/enterprise architecture
 4. Raising issues relevant to the direction and development of this project
 
 The SC consists of the following members:
@@ -47,7 +47,7 @@ opening a GitHub issue (see the [Contribution Guide](CONTRIBUTING.md)).
 ## Contributors
 
 "Contributors" refers to anyone in the community who contributes code, 
-documentation, or other technical artifacts to the project.
+documentation, or other useful artifacts to the project.
 
 Anyone can become a contributor.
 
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "url": "git+ssh://git@github.com/open-formulieren/open-forms.git"
   },
   "author": "Maykin",
-  "license": "UNLICENSED",
+  "license": "EUPL-1.2",
```
sergei-maertens, line 28: this *was* deliberate because we don't publish the backend JS as a package to NPM and never will.

PROJECT_GOVERNANCE.md
```diff
@@ -0,0 +1,56 @@
...
+required, the "Steering Committee" has the authority to make 
+decisions.
+
+## Steering Committee
+
+The Steering Committee (SC) is responsible for:
+
+1. The general ambitions and objectives of this project
+2. Guidelines, procedures, and tool selection
+3. Decisions regarding architecture and (development) infrastructure
```
sergei-maertens: I have some concerns about architecture decisions being made by non-technical people. The only current member of the SC with an architect hat (as far as I know) is Joeri (and Wietieke is also quite technical).

The way this governance document describes it, it reads a lot that decisions made by the SC are final and a tech lead (me) has no say in it.

PROJECT_GOVERNANCE.md
```diff
@@ -0,0 +1,56 @@
...
+4. Raising issues relevant to the direction and development of this project
+
+The SC consists of the following members:
```
sergei-maertens, line 17: Maybe I'm complicating things too much, but that's the current composition of the SC. How are decisions made to add or remove people to the SC? How many seats are there? How long can someone be a member?

PROJECT_GOVERNANCE.md
```diff
@@ -0,0 +1,56 @@
...
+Maintainers are responsible for maintaining parts of the codebase. Maintainers 
+have the following responsibilities (summary):
+
+* Coordinating development activities
+* Ensuring code and documentation reviews are performed
+* Coordinating Pull Requests
+* Coordinating security issues and bug report follow-ups
+* Coordinating responses to inquiries
+* In the event of prolonged discussions or disputes, maintainers (among others)
+  may request a decision from the community council.
```
sergei-maertens: this is the first time the term "community council" is mentioned. What is it? How do you contact them? Who are in it?

PROJECT_GOVERNANCE.md
```diff
@@ -0,0 +1,56 @@
...
+financing, please see
+[Broncodebeheer](https://maykinmedia.github.io/maykin-plus/broncodebeheer).
+
+Any community member or contributor can ask a question or report an issue by
+opening a GitHub issue (see the [Contribution Guide](CONTRIBUTING.md)).
+
+## Contributors
+
+"Contributors" refers to anyone in the community who contributes code, 
+documentation, or other technical artifacts to the project.
```
sergei-maertens: I struggle with the notion that a contributor must contribute "technical" artifacts.

Feedback about documentation, language/translation fixes... are also valuable contributions, to name a few.

README.rst
```diff
@@ -43,6 +43,18 @@ both end users and administrators.
...
+Sustainable Management
+======================
+
+This software is open source and free to use under the terms of the EUPL. However, safe and reliable use in a production environment requires structured management, including security updates, dependency and release management, quality assurance, and vulnerability handling. Maintaining public-facing functions associated with the open-source product also requires a sustained commitment.
+
+**Public code calls for public responsibility.**
+
+The `Stakeholders`_ expect public organizations, using this software in production, to make a financial contribution towards its collective maintenance.
+
+Read more about the management organization, contributions, and responsibilities in `PROJECT_GOVERNANCE.md`_.
```
sergei-maertens, line 55: A copy of this in `docs/open_source/index.rst` would be appropriate IMO

package.json
joeribekker: EUPL does not require you to push this package to NPM? Imo 2 separate things.

PROJECT_GOVERNANCE.md
joeribekker: @bregje-hub So, should Sergei be in the SC or should the SC not be concerned with architectural decisions?

PROJECT_GOVERNANCE.md
joeribekker: This document was largely based on https://developer.overheid.nl/kennisbank/open-source/standaarden/project-governance-md where this is also not stated.

I agree it would be nice to have this written out but until someone wants to join, we can figure this out at that time I guess.

PROJECT_GOVERNANCE.md
joeribekker: I agree. Let's just use the TS in this case.

PROJECT_GOVERNANCE.md
joeribekker: Documentation is mentioned. I don't see the need to make the list extensive or complete.

PROJECT_GOVERNANCE.md
bregje-hub: I agree with @sergei-maertens but because there are 2 members in the SC from Maykin (Joeri and me) i thought this was enough. I think at least one person with the role senior architect (or something similar) has to be in the SC.

package.json
sergei-maertens: no, but it's a hint for future developers that this package is not meant to be published. Adding a license creates the impression that it's not published by accident, as this license information is displayed *on* NPM

PROJECT_GOVERNANCE.md
sergei-maertens: I'd favour leaving architecture discussion out of the responsibilities of the SC tbh, especially with my other questions about the population of it - I don't see a mechanism now to guarantee that people with technical knowledge are in the SC at all times.

PROJECT_GOVERNANCE.md
sergei-maertens: Yeah I think this is a big shortcoming from the upstream document tbh

PROJECT_GOVERNANCE.md
sergei-maertens: maybe replace "technical" with "useful"?

package.json
joeribekker: I find UNLICENSED less clear than EUPL and also not a "hint". Let's just do EUPL -or- remove license altogether.

PROJECT_GOVERNANCE.md
joeribekker: I think -but its an assumption- that @sergei-maertens concern is about the software architecture.  Which is not the intended "architecture" here.

If so, this can be resolved by stating the type of architecture and/or the intention. For clarity: The intention is to oversee the broader ICT landscape and/or new laws that should be taken into account. For example: WMEBV (law) or the existence of Open Product or Open Wave and how and -if- Open Forms should do something with this.

So, if we change it to "information and/or enterprise architecture" I think it's ok. I agree that "solution and/or software architecture" is for the tech lead and does not fit in the SC.

PROJECT_GOVERNANCE.md
joeribekker: Fine by me :)

PROJECT_GOVERNANCE.md
joeribekker: We can agree on that but I don't think its worth deciding on a text now.

package.json
sergei-maertens: discussed in the office, let's go with this change indeed (make it EUPL)

PROJECT_GOVERNANCE.md
sergei-maertens: Discussed in the office - indeed, the "kind" of architecture was misleading me. If we can clarify that wording, I'm happy.

joeribekker: Done.

sergei-maertens (approved):

PR #6495, opened 2026-07-28
Add help callout page

### LINKED ISSUE #6317: Create help callout page
Image

# [Refinement](https://github.com/open-formulieren/open-forms/issues/5787#issuecomment-3870621080):

**Callout**

* Click the help button on the callout page -> what happens? a11y? -> aria-disabled
* Callout is only shown when help text on form is configured + display option selected (before/after start page) + text in the global config (disable config option on form if the preconditions are not met)

Aspects:

* global configuration -> (tinymce images: not setup right now, so we go for a simple `models.ImageField` instead): 1D
* only display configuration options for callout display if the global config is done for its content: 2 days
* displaying in the SDK -> (add new page/route(s) conditional before/after start page, preserve query parameters, new component + CSS + design tokens & render image): 5-10days
* help button -> (interactive vs non interactive, navbar separation): 3 days
* add configuration to form detail endpoint -> (global configuration + form-specific configuration): 1 day

# [Acceptatiecriteria](https://github.com/open-formulieren/open-forms/issues/5787#issuecomment-4178223829):

**Story:** Als gebruiker wil ik eenvoudig contact opnemen met de gemeente Den Haag om mij te helpen bij mijn aanvraag, tijdens het invullen van het formulier. 

### Onderdelen in scope
- Hulpknop: het telefoon icoon dat aangeklikt kan worden
- Callout: de introductie van de hulpknop
- De dialog: het schermpje dat getoond wordt door op de hulpknop te klikken

**Acceptatiecriteria**
- Deze feature (hulpknop, callout en dialog) kan optioneel ‘aan’ of ‘uit’ gezet worden per formulier
- De hulpknop en callout staan vast qua tekst, afbeelding en design in alle formulieren
- De dialog kan per formulier qua tekst aangepast worden (wysiwyg editor)
afbeelding tonen is nice-to-have
- De callout kan getoond worden:
optie 1: na inloggen (met DigiD)
optie 2: voor inloggen (met DigiD)
- De hulpknop staat bovenaan de pagina en is niet sticky

# Tasks

- [x] Add fields (content + image) to display on this page to the `GlobalConfiguration` (the image is optional but like the other fields is fixed in position)
- [x] Add a new form configuration option `help_callout_display` with the options `before_startpage`, `after_startpage`, `none`
  - Show a disabled field for this option whenever the content field in the `GlobalConfiguration` is **not** filled in
  - Check what happens to the single-step form configuration when adding this option -> should be disabled
- [x] Expose the new `GlobalConfiguration` fields and form specific fields through the form detail endpoint for the front end
- [x] Include a new route in the SDK to show the page's content (see point above for the when and how)
  -   Based on the `help_callout_display` configuration option, an additional route is added in the SDK, matching the screenshot. If it's before the startpage, it follows after the (optional) introduction page. if after the startpage, it comes directly after the start page 
- [x] Implement additional design token wishes
- [x] Add some documentation

viktorvanwijk:
Partly closes #6317

[skip: e2e]

**Changes**

* Add help callout page display option to the form
* Add help callout page content and image to global configuration

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/config/models/config.py
```diff
@@ -667,6 +667,20 @@ class GlobalConfiguration(SingletonModel):
         ),
     )
 
+    # Help callout page
+    help_callout_page_content = HTMLField(
```
viktorvanwijk: TODO: figure out which exact text formatting options we want to enable for this field

sergei-maertens: This is the config I went with for the help dialog content: https://github.com/open-formulieren/open-forms/blob/3c660179aacff9a4be7ca1134f5b0a1f07b91d16/src/openforms/js/components/admin/form_design/FormConfigurationFields.js#L520

src/openforms/config/models/config.py
```diff
@@ -667,6 +667,20 @@ class GlobalConfiguration(SingletonModel):
...
+        _("help callout page content"),
+        default=_(""),
```
sergei-maertens:
```suggestion
        default="",
```
no need to mark empty strings translatable 😄

src/openforms/forms/api/serializers/form.py
```diff
@@ -132,6 +134,43 @@ def validate(self, attrs):
...
+        config = GlobalConfiguration.get_solo()
+        return field.to_representation(config.help_callout_page_content)
+
+    @extend_schema_field(OpenApiTypes.STR)
+    def get_image(self, *args) -> str:
+        config = GlobalConfiguration.get_solo()
+        field = serializers.ImageField()
+        if uri := field.to_representation(config.help_callout_page_image):
+            return build_absolute_uri(uri)
+        return ""
```
sergei-maertens: This is certainly creative 😅 

It irks me a bit though, because `serializers.FileFIeld` has built-in support for emitting the fully qualified file URL if the `request` is in the context, and you're not getting that now because the field is not bound to its parent.

Additionally, the `field.bind(...)` semantics are low-level private DRF API and you should really only be using those in library-like code (like we do in `openforms.api` for example)

The challenge here is getting access to the `GlobalConfiguration` from a `Form` instance being serialized, but I think an approach like this is neater:

```py
class GlobalConfigurationLookupMixin:
    def get_attribute(self, instance):
        # inject our global configuration instance instead
        instance = GlobalConfiguration.get_solo()
        return super().get_attribute(instance)  # pyright: ignore[reportAttributeAccessIssue]


class GlobalConfigurationCSPPostProcessedHTMLField(
    GlobalConfigurationLookupMixin, CSPPostProcessedHTMLField
):
    pass


class GlobalConfigurationImageField(
    GlobalConfigurationLookupMixin, serializers.ImageField
):
    pass


class HelpCalloutPageSerializer(serializers.Serializer):
    display = serializers.ChoiceField(
        label=_("Help callout page display"),
        help_text=_("When to display the help callout page."),
        choices=HelpCalloutPageDisplayChoices.choices,
        source="help_callout_page_display",
    )
    content = GlobalConfigurationCSPPostProcessedHTMLField(
        source="help_callout_page_content",
        label=_("Help callout page content"),
        help_text=_(
            "Content for the help callout page, fetched from the global configuration."
        ),
        read_only=True,
    )
    image = GlobalConfigurationImageField(
        source="help_callout_page_image",
        label=_("Help callout page image"),
        help_text=_(
            "Image for the help callout page, fetched from the global configuration."
        ),
        read_only=True,
    )
```

src/openforms/forms/tests/test_api_forms.py
```diff
@@ -2125,3 +2200,24 @@ def test_create_with_translations_normalize_values(self):
...
             "{% appointment_information %} {% payment_information %} {% cosign_information %}",
         )
+
+    @patch(
+        "openforms.forms.api.serializers.form.GlobalConfiguration.get_solo",
+        return_value=GlobalConfiguration(
+            help_callout_page_content_nl="Nederlands",
+            help_callout_page_content_en="English",
+        ),
+    )
```
sergei-maertens: these days I prefer not mocking this and instead adding a `self.addCleanup(GlobalConfiguration.clear_cache)` to the test

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -124,6 +124,7 @@ const initialFormState = {
       brpPersonenProcessingHeaderValue: '',
     },
     authBackends: [],
+    helpCalloutPage: {display: 'never', content: '', image: ''},
```
sergei-maertens: I think `content` and `image` can/should be omitted because they're read-only fields not managed through a form anyway?

src/openforms/js/components/admin/form_design/form-creation-form.js
```diff
@@ -1270,6 +1272,11 @@ const FormCreationForm = ({formUuid, formUrl, formHistoryUrl, outgoingRequestsUr
           registrationBackends: state.form.registrationBackends,
           selectedAuthPlugins: state.selectedAuthPlugins,
           updateComponents: updateComponents,
+          helpCalloutPage: {
+            display: state.form.helpCalloutPage.display,
+            content: state.form.helpCalloutPage.content,
+            image: state.form.helpCalloutPage.image,
```
sergei-maertens: these two can be omitted, no?

src/openforms/js/components/admin/form_design/FormConfigurationFields.js
```diff
@@ -514,6 +514,53 @@ const FeatureFields = ({formType, translationEnabled, suspensionAllowed, onChang
...
+          }
+          helpText={
+            <FormattedMessage
+              description="Help callout page display setting help text"
+              defaultMessage={`When to display the help callout page. Note that this field is
+              disabled when the help callout page content in the global configuration is not
+              specified.`}
+            />
+          }
+          disabled={!contentConfigured}
```
sergei-maertens: ah right, you need to see if there's content or not.

I'd just pass the derived boolean as prop instead of passing along the whole content/image

sergei-maertens (changes requested):

src/openforms/js/components/admin/form_design/form-creation-form.js
viktorvanwijk: Yeah I was doubting between leaving the image out and keeping it consistent with what the serializer outputs

src/openforms/js/components/admin/form_design/form-creation-form.js
viktorvanwijk: I have removed it from the context completely, because it's not used anyway actually. The `FormConfigurationFields` works directly with the form from the state

sergei-maertens (approved):

PR #6496, opened 2026-07-29
Prepare for formiojs drop

### LINKED ISSUE #307: Drop dependency Formiojs
Remove the Formiojs dependency from the projects, and replace whatever we still use with local implementations

- [x] Formio-builder #308 
- [x] Backend project

When removing formiojs from the backend, also add a end-to-end to validate that open-formulieren/open-forms#6297 doesn't happen anymore

robinmolen:
Closes open-formulieren/formio-builder#307

**Changes**

Removing various Formiojs dependencies, to make the complete dropping of Formiojs easier.

There are a couple instances left untouched. These instances are required, as long as we don't make the full switch to replace the Formiojs `FormBuilder` with our own formio-builder `FormDefinitionDesigner`:
* The `WebformBuilder` component. This contains numerous function calls to Formiojs utilities.
* The component definitions - these all use/extend Formiojs component classes
* The styling for the `FormBuilder`

None of these instances are used for the new formio-builder `FormDefinitionDesigner`, so replacing these with local alternatives, just to throw it away in a couple weeks/months seems wasteful.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/forms/tests/e2e_tests/helpers.py
```diff
@@ -35,11 +35,13 @@ async def open_component_options_modal(page: Page, label: str, exact: bool = Fal
     # formio doesn't have accessible roles here, so use CSS selector
     await page.locator('css=[ref="editComponent"]').locator("visible=true").last.click()
     # check that the modal is open now
-    await expect(page.locator("css=.formio-dialog-content")).to_be_visible()
+    await expect(
+        page.locator(".react-modal__content, .formio-dialog-content")
+    ).to_be_visible()
```
sergei-maertens: Can we use accessible queries here now? We're now using proper dialog/modal semantics so the CSS selector shouldn't be necessary anymore, right?

src/openforms/forms/tests/e2e_tests/test_form_designer.py
```diff
@@ -1423,9 +1423,11 @@ def setUpTestData():
                             "label": "Field Set 1",
                             "components": [
                                 {
-                                    "type": "datagrid",
-                                    "key": "datagrid",
+                                    "type": "editgrid",
```
sergei-maertens, line 1426: cool, one less case to fix in my msgspec branch 😛

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiSummaryHandler.js
```diff
@@ -89,9 +89,8 @@ const ObjectsApiV2SummaryHandler = ({variable, backendOptions}) => {
 
   if (
     !variableMapping ||
-    (!variableMapping.targetPath &&
-      variableMapping.options &&
-      !Object.keys(variableMapping.options).length > 0)
+    !variableMapping.targetPath ||
+    (variableMapping.options && !Object.keys(variableMapping.options).length > 0)
```
sergei-maertens: what's happening here? why the change from `and` to `or`?

(that's the kind of information I'm looking for in commit messages)

src/openforms/forms/tests/e2e_tests/test_registration_backend_conf.py
```diff
@@ -138,21 +134,30 @@ def collect_requests(request):
...
+            str(zgw_api_1.pk),
+        )
+        self.assertEqual(
+            furl(requests_to_endpoint[3].url).args["zgw_api_group"],
+            str(zgw_api_2.pk),
+        )
+        self.assertEqual(
+            furl(requests_to_endpoint[4].url).args["zgw_api_group"],
             str(zgw_api_2.pk),
         )
```
robinmolen: Like we discussed briefly during standup, this validation might not be very relevant anymore (its now basically a "yeah, the changes to the registration backend have been correctly passed to the frontend components"). And seeing the webkit CI failure, this also seems pretty fragile/irregular results

src/openforms/forms/tests/e2e_tests/test_registration_backend_conf.py
```diff
@@ -92,15 +92,14 @@ def collect_requests(request):
                 return
 
             match = resolve(str(url.path))
-            if match.view_name == "api:zgw_apis:document-type-list":
+            if match.view_name == "api:zgw_apis:catalogue-list":
```
robinmolen: I thought that `document-type` wasn't requested anymore, so i changed it to catalogue, but i see it gets requested when the registration actually configures the document type. So i can revert this, and extend the test if we want to keep it

robinmolen: Same for the Objects api test below

sergei-maertens: This whole test can be deleted (same for objects API) because the file registration tab is gone, and that's what it was testing.

src/openforms/forms/tests/e2e_tests/test_registration_backend_conf.py
sergei-maertens: Yep, nuke it!

src/openforms/forms/tests/e2e_tests/test_registration_backend_conf.py
sergei-maertens: also remove this test :)

src/openforms/js/components/admin/form_design/variables/utils.js
```diff
@@ -19,16 +24,13 @@ const getComponentDatatype = component => {
 };
 
 const isInEditGrid = (component, configuration) => {
-  // Get all edit grids in the configuration
-  let editGrids = [];
-  FormioUtils.eachComponent(configuration.components, configComponent => {
-    if (configComponent.type === 'editgrid') editGrids.push(configComponent);
-  });
+  for (const {component: editgrid} of iterComponents(configuration?.components || [])) {
+    if (editgrid.type !== 'editgrid') continue;
```
sergei-maertens: it's quite confusing to name the variable already editgrid and only later check if it actually is one - I'd prefer a variable reassignment after the check:

```js
const editgrid = component;
```

src/openforms/js/components/admin/form_design/variables/utils.js
```diff
@@ -92,17 +94,20 @@ const shouldNotUpdateVariables = (newComponent, oldComponent, mutationType, step
 const getFormVariables = (formDefinition, configuration) => {
   const newFormVariables = [];
 
-  FormioUtils.eachComponent(configuration.components, component => {
-    if (component.type === 'softRequiredErrors') return;
+  for (const {component} of iterComponents(configuration?.components || [])) {
+    if (component.type === 'softRequiredErrors') continue;
+    // Previous formio implementation skipped layout components that don't hold data
+    // (everything but 'editgrid'), so we do the same now.
+    if (isLayoutComponentWithoutData(component)) continue;
```
sergei-maertens: this function name implies there's something like a layout component with data, but that's a contradiction. A layout component by definition doesn't hold data.

src/openforms/js/components/admin/icons/EditIcon.js
```diff
@@ -4,12 +4,15 @@ import React from 'react';
...
 const EditIcon = ({label, onClick}) => (
-  <FAIcon
-    icon="edit"
-    extraClassname="fa-lg actions__action"
-    accessibleLabel={label}
+  <button
+    type="button"
     onClick={onClick}
-  />
+    className="button button--plain"
```
sergei-maertens: I'd expect there's some layout that depends on the `actions__action` child, so probably that class needs to be moved from the icon into this button.

src/openforms/js/components/admin/modals/Modal.js
```diff
@@ -35,7 +37,14 @@ const Modal = ({
     >
       <header className="react-modal__header">
         {title ? <h2 className="react-modal__title">{title}</h2> : null}
-        <button className="react-modal__close" aria-label="Sluiten" onClick={closeModal}>
+        <button
+          className="react-modal__close"
+          aria-label={intl.formatMessage({
+            description: 'Modal close button aria label',
```
sergei-maertens:
```suggestion
            description: 'Modal close button accessible label',
```

src/openforms/js/components/utils.js
```diff
@@ -1,8 +1,31 @@
...
+import {getRegistryEntry as getRendererRegistryEntry} from '@open-formulieren/formio-renderer';
 
 const getComponentEmptyValue = component => {
-  const componentInstance = Formio.Components.create(component);
-  return componentInstance.emptyValue;
+  const entry = getRendererRegistryEntry(component);
+  if (!entry?.getInitialValues) return undefined;
+
+  const initialValue = entry.getInitialValues(component, getRendererRegistryEntry);
+  return initialValue[component.key];
```
sergei-maertens: empty value != default value

A component definition (for a textfield for example) that has a default set would produce that default as initial value, but the empty value is either the emptry string (`multiple: false`) or empty list (`multiple: true`).

I don't think we have that information currently in either the renderer or the builder, but it exists in the backend: https://github.com/open-formulieren/open-forms/blob/b1f925f27cab857a873297b3e54289e47d5a99c9/src/openforms/formio/registry.py#L115

I'd approach this one pragmatically and have some JSON script output this to the template engine - something like this should cover all variants:

```py
from openforms.formio.registry import register as component_registry

# mapping of tuple (type, multiple) to the empty value for each component type
empty_values: dict[tuple[str, bool], JSONValue] = {}  
for component_plugin in component_registry:
    for multiple in (True, False):
        _component_mock = {"type": component_plugin.identifier, "multiple": multiple, "key": "dummy", "label": "Dummy"}
        empty_value = component_registry.get_empty_value(_component_mock)
```

this it to be used in the admin template https://github.com/open-formulieren/open-forms/blob/main/src/openforms/forms/templates/admin/forms/includes/formio_config.html then with the `json_script`, so if you put the above `empty_values` in the context you should be able to expose it to the JS code.

src/openforms/js/components/utils.js
```diff
@@ -1,8 +1,31 @@
...
+  Array.from(iterComponents(components)).reduce((carry, configuration) => {
+    carry[configuration.component.key] = configuration.component;
+    return carry;
+  }, {});
+
+const isLayoutComponent = component => {
+  const entry = getBuilderRegistryEntry(component['type']);
+  // Return true if the component has children
+  return entry?.getComponentSlots !== undefined;
+};
```
sergei-maertens: an edit grid component has children but it is *NOT* a layout component :)

webpack.config.js
```diff
@@ -62,6 +62,14 @@ module.exports = {
         loader: 'babel-loader',
       },
 
+      // .mjs
+      {
+        test: /\.m?js/,
+        resolve: {
+          fullySpecified: false,
+        },
+      },
```
sergei-maertens, line 74: huh, where do we output/import mjs?

sergei-maertens (changes requested):

src/openforms/forms/tests/e2e_tests/helpers.py
robinmolen: Not yet, the formiojs dialog doesn't implement a dialog role. And as the Formiojs FormBuilder is still (for this PR) being used, we have to support continue with classname selectors.
It could be that `react-modal__content` is only needed for opening the registration configuration in the variables view (the e2e tests that have been removed). In that case i'll revert this back to the singular classname selector.

With the formio-builder FormBuilder implementation we can fully shift to a proper `page.locator("[role=dialog]")` implementation.

src/openforms/js/components/admin/form_design/variables/utils.js
robinmolen: Yeah, good point. I've changed this a bit, renamed the target component variable and just stick to `component` for the iter yields

src/openforms/js/components/utils.js
robinmolen: This naming was perhaps a bit poor. This is a custom implementation of the Formio `isLayoutComponent` util, which does classify every component with children as a layout component.
I'll rename this to `hasChildren`, to better reflect the expected behaviour.

sergei-maertens: Yeah we got bitten by their own lies in the past for this exact thing, I'd rather not repeat that :D

webpack.config.js
robinmolen: The resolve rule needs to set on `mjs` as the formio-renderer is marked and regarded as a module (there is an import there that causes issues).
However, its probably best to fix this directly in the formio-renderer. I'll make a PR there to solve this issue

sergei-maertens: Before you do anything, please first describe what the plan is.

The renderer is indeed intended to be a module, and implicit module resolution (without extension) is not where modern JS is going. Aliases can be set via the [`exports`](https://github.com/open-formulieren/formio-renderer/blob/d6916203ec75f99823c0d2f24ffbf08aa20e1a56/package.json#L6) config, which we already do.

robinmolen: The issue was in "incomplete" import path in the formio-renderer (`@formatjs/intl/src/utils` instead of `@formatjs/intl/src/utils.js`). This been updated in the formio-renderer (https://github.com/open-formulieren/formio-renderer/pull/367), and will be part of the 2.0 release.
This rule will be kept for now, so we can remove the Formiojs utils. In OF 4.0 this rule will be removed, as we will then use formio-renderer 2.0.

src/openforms/js/components/utils.js
robinmolen: Ah, with how the empty values were used (as fallback for missing default values) i thought the formio-renderer initial values would work okay. Anyway, i've adding something similar to your proposal (using simple string keys instead of tuple keys, as JSON doesn't really like those)

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiSummaryHandler.js
robinmolen: I've reverted this, but re-testing it now i see this is indeed a problem. (Mind you, weird edge-case broken registration type stuff)

I noticed this when trying to fix the e2e bugs for FormDesignerRegistrationBackendConfigTests (removed in this PR in 3a7fa3db70a031d3f8a6828aa2b25a29bd71ac12). This hard crashes (i.e. white/blank screen) the admin when you do some weird edge-case broken registration configuration.

So, what you do is:
1. add Objects API registration to an form with a textfield (or, whatever component)
2. configure a API group and save it
3. go to variables tab
4. add Objects API registration for one of the components
5. save the variable registration configuration
6. ✨ crash ✨ 

In code, there will be a `variableMapping` like `{"variableKey": "addressNl", "targetPath": undefined}`. This passes this if statement, as the `variableMapping` isn't undefined/falsy, and there are no `options` (so the second statement also doesn't stop it). The final line of this component tries to render some string using the `targetPath`, which errors as `targetPath` is undefined.

I'll add this as a bug ticket, as this should be covered with a unit-test and doesn't really relate to this PR

robinmolen: https://github.com/open-formulieren/open-forms/issues/6516

src/openforms/forms/tests/e2e_tests/helpers.py
sergei-maertens: > `page.locator("[role=dialog]")` implementation.

I'd expect this to be `page.get_by_role("dialog")` - these CSS selector queries are really a last resort in playwright.

src/openforms/js/components/admin/form_design/registrations/objectsapi/ObjectsApiSummaryHandler.js
sergei-maertens: Check, that's indeed an issue and highlight again why typescript is essential for these codebase sizes/complexities. A separate ticket is indeed the way to go here.

webpack.config.js
sergei-maertens: Created a follow up ticket #6518

src/openforms/forms/admin/mixins.py
```diff
@@ -30,6 +32,31 @@ def get_wms_layers():
...
+                "label": "Dummy",
+                # digitalAddressTypes is needed for customerProfile component
+                "digitalAddressTypes": ["email", "phoneNumber"],
+            }
+            empty_value = component_registry.get_empty_value(_component_mock)
+            json_safe_empty_value = (
+                empty_value if empty_value is not NotImplemented else None
+            )
+
+            key = f"empty_value_{'multiple' if multiple else 'single'}"
```
sergei-maertens: string-encoding like that is really not elegant but it'll do for now I guess.

You *could* also have used a list of lists which encodes just fine in JSON :) And client-side you can chuck it into a `Map` which allows keys of any type.

src/openforms/forms/admin/mixins.py
```diff
@@ -30,6 +32,40 @@ def get_wms_layers():
...
+type EmptyValueOption = tuple[str, bool, JSONValue]
+"""
+Data meaning: (component_type, multiple or not, associated empty value).
+"""
+
+
+def get_component_empty_values():
+    # hack to pass the empty values, as it's not readily available in the
+    # formio-renderer or formio-builder...
+    # FIXME: build a proper solution for this.
```
sergei-maertens, line 44: I cleaned this up in 809917a41f4f4d9b776b53d366d3008fcfcbad0c

src/openforms/js/components/utils.js
```diff
@@ -1,8 +1,41 @@
...
+
+const isLayoutComponent = component => {
+  const entry = getBuilderRegistryEntry(component['type']);
+  // if the component is not marked as "holding data", it is by definition a layout
+  // component that only serves a presentational aspect. Whether the component has
+  // children or not, is not relevant.
+  return !entry.holdsData;
```
sergei-maertens, line 38: The semantics were still wrong here, I cleaned it up in e456ec79cfd2a953fe178ef042415eaf58757e93

sergei-maertens (approved):
I'll merge this if CI is green, had to do some more polishing.

PR #6500, opened 2026-07-29
Prevent hidden required multiple fields from triggering validation

### LINKED ISSUE #6297: hidden fields with "multiple" enabled & not "clear on hide" trigger validation (required)

3.3.x > 3.4.x > latest

### Customer reference

Venray 194

Image

This configuration will add a empty string to the list while loading the step:
Image

Which will block the submission:
Image

Old & new renderer, with and without new logic. 

submissions which cannot be submitted: 
multivalue textfield: [""]
which used to be:
multivalue textfield: [null]

form 1071 on m-test.

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

robinmolen:
Partly closes #6297

**Changes**

This PR fixes the bug where a required hidden `multiple=True` component with `clearOnHide=False` triggers validation with the value `[""]`. The error states that some unknown field (the child of the ListField) is left empty (the blank `""` value).

To prevent hidden fields from triggering validation we apply apply `_remove_validations_from_field` on them, to strip them from their validation rules. This however only strips the direct component/ListField of its validation rules, not its children. By also applying this stripping to the children of hidden `multiple=True` components, we can allow falsy list values like `[""]`.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/tests/test_submission_completion.py
+++ b/src/openforms/submissions/tests/test_submission_completion.py
@@ -875,6 +875,52 @@ class SubmissionCompletionTests(SubmissionsMixin, APITestCase):
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
+    @tag("gh-6297")
+    def test_required_hidden_multiple_field_should_not_fail_validation(self):
+        submission = SubmissionFactory.create(
+            form__generate_minimal_setup=True,
+            form__formstep__form_definition__configuration={
+                "components": [
+                    {
+                        "type": "textfield",
+                        "key": "textfieldVisible",
+                        "label": "Textfield visible",
+                        "hidden": False,
+                    },
+                    {
+                        "type": "textfield",
+                        "key": "hiddenField",
+                        "label": "Hidden field",
+                        "validate": {
+                            "required": True,
+                        },
+                        "multiple": True,
+                        "hidden": False,
+                        "clearOnHide": False,
+                        "conditional": {
+                            "show": False,
+                            "when": "textfieldVisible",
+                            "eq": "hide",
+                        },
+                        "defaultValue": [],
+                    },
+                ]
+            },
+        )
+        SubmissionStepFactory.create(
+            submission=submission,
+            form_step=submission.form.formstep_set.get(),
+            data={
+                "textfieldVisible": "hide",
+                "hiddenField": [""],
+            },
+        )
+        self._add_submission_to_session(submission)
+        endpoint = reverse("api:submission-complete", kwargs={"uuid": submission.uuid})
+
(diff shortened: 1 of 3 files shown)


sergei-maertens:
> we can allow falsy list values like `[""]`

But this shouldn't be allowed - a hidden field with clear on hide should have an empty list `[]` as value...

Effectively with the clear-on-hide rework this entire key will be absent from the submission data, and pre 4.0 the behaviour is indeed that it uses the default value, so that component definition should not have `[""]` set as default value if it can be hidden/requires clearing.

This patch also completely ignores the builder frontend bug I can reproduce reliably: https://github.com/open-formulieren/open-forms/issues/6297#issuecomment-5116611027

robinmolen:
This bug is about when "clear on hide" is set to false..

sergei-maertens:
Ah, I was thrown off by the test that has clear on hide enabled.

That does still leave the broken frontend behaviour in the admin though 😬

robinmolen:
Yeah i wanted to keep those separated, so one PR for the logic validation bug, on for component configuration

src/openforms/submissions/tests/test_submission_completion.py
```diff
@@ -939,6 +939,52 @@ def test_field_required_hidden_depending_on_another_field(self):
...
+                    {
+                        "type": "textfield",
+                        "key": "hiddenField",
+                        "label": "Hidden field",
+                        "validate": {
+                            "required": True,
+                        },
+                        "multiple": True,
+                        "hidden": False,
+                        "clearOnHide": True,
```
sergei-maertens: the end to end test uses `clearOnHide: False`, as the reported bug, but here it's `clearOnHide: True` which is the oppossite

The submitted step data doesn't make sense in this configuration - the key `hiddenField` should not be present at all. Is this deliberate?

sergei-maertens: I think that the test would still properly capture the buggy behaviour if you simply switch `clearOnHide: Frue` to `clearOnHide: False` - as empty items in that data would still be kept in the submission data then.

robinmolen: Mhh yeah, the `clearOnHide: True` doesn't seem right.. Changing it

sergei-maertens (approved):

sergei-maertens:
Backports:

* 3.5.x: e6169eec495badd7e087ff44e34686a98f7c844c
* 3.4.x: 816a3b3eb4b3ebb065127b2a630cadea1ec2fb52

PR #6512, opened 2026-08-03
Fixing default value for multiple field

### LINKED ISSUE #6297: hidden fields with "multiple" enabled & not "clear on hide" trigger validation (required)

robinmolen:
Partly closes #6297

**Changes**

This PR fixes the configuration issue described in #6297.

Short version, Formiojs messes with empty list values in the schema's, replacing them with their default schema values. For textfields (and a bunch of other components) this means that the schema value `defaultValue=[]` gets replaced with `defaultValue=""`.

This fix checks if the component `defaultValue` picked by Formiojs correctly represents the `multiple` property (i.e. if `multiple=true`, then `defaultValue` must be an array value). If the `defaultValue` doesn't match with the `multiple` property, we set the `defaultValue` to an empty array `[]`.

Because this issue is caused by the Formiojs component classes, this won't be a problem when we completely drop Formiojs. So this fix is a temporary one.

### Extra Formiojs context

The Formiojs `Component` base class implements a [`getModifiedSchema`](https://github.com/formio/formio.js/blob/v4.13.13/src/components/_classes/component/Component.js#L685C3-L715) method which returns all the modified schema properties. This is done by checking which component configuration properties are different from the default schema AND, if they are list values, are NOT empty.
The result is: a component with the configuration `{defaultValue: []}` will be returned as `{}` (i greatly simplified the configuration for dramatic effect, but you get the idea)

The missing component properties are later filled-in using the default schema which, in the case of all our components that support `multiple` configuration, uses the configuration of `defaultValue=""`.

Formiojs does contain some code which do correct `multiple` + `defaultValue` + `emptyValue` calculations, but these results are never persisted back to the component schema.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/migration_converters.py
+++ b/src/openforms/formio/migration_converters.py
@@ -343,6 +343,10 @@ def fix_empty_default_value(component: Component) -> bool:
         changed = True
 
     if component.get("multiple", False):
+        if not isinstance(component["defaultValue"], list):
+            component["defaultValue"] = []
+            changed = True
+
         for index, value in enumerate(component["defaultValue"]):
             if value is None:
                 component["defaultValue"][index] = ""
@@ -458,6 +462,7 @@ CONVERTERS: dict[str, dict[str, ComponentConverter]] = {
     },
     "date": {
         "alter_prefill_default_values": alter_prefill_default_values,
+        "fix_empty_default_value": fix_empty_default_value,
         "rename_identifier_role_authorizee": rename_identifier_role_authorizee,
         "remove_empty_conditional_values": remove_empty_conditional_values,
         "replace_empty_datepicker_properties": replace_empty_datepicker_properties,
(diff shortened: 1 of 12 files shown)


robinmolen:
The changes to the component classes will solve any configuration problems when you edit the form definition. Simply go to the form edit page in the admin, opening the effected form step and saving the form.

There have also been changes to the `fix_empty_default_value` migration converter. This will ensure that effected form definitions that are shared using import/export will also be fixed

sergei-maertens (commented):
I'll approve and merge this because this code is soon going to be deleted anyway, but I strongly disagree with all these cross-imports between different component types.

The proper approach would be to put this patch function in a separate module and have all these component types import and call it, instead of treating the textfield module as some kind of core thing.

sergei-maertens:
Backports:

* 3.5.x: f0a4bc8d3436f8a0717fe4785068d49f8876e5e4
* 3.4.x: 06c8812c798eae8af5fdce7f08a4db759e50409d

PR #6525, opened 2026-08-11
Do not crash on empty content field

### LINKED ISSUE #6524: Form summary crashes with empty content field

3.5

### Customer reference

_No response_

The form summary page crashes when there is an empty content field

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

bartjkdp:
Closes #6524

**Changes**

This change makes sure the summary page does not crash when there is an empty content field in there.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/emails/tests/test_utils.py
+++ b/src/openforms/emails/tests/test_utils.py
@@ -4,6 +4,11 @@ from ..utils import strip_tags_plus
 
 
 class StripTagsPlusTests(SimpleTestCase):
+    def test_empty_content(self):
+        output = strip_tags_plus("")
+
+        self.assertEqual(output, "")
+
     def test_strip_tags_no_keep_leading_whitespace(self):
         markup = """
         <p>Some paragraph</p>
--- a/src/openforms/emails/utils.py
+++ b/src/openforms/emails/utils.py
@@ -8,7 +8,7 @@ from django.template.loader import get_template
 
 from mail_cleaner.mail import send_mail_plus
 from mail_cleaner.sanitizer import sanitize_content as _sanitize_content
-from mail_cleaner.text import strip_tags_plus
+from mail_cleaner.text import strip_tags_plus as _strip_tags_plus
 
 from openforms.config.models import GlobalConfiguration, Theme
 from openforms.template import openforms_backend, render_from_string
@@ -19,6 +19,13 @@ from .context import get_wrapper_context
 MESSAGE_SIZE_LIMIT = 2 * 1024 * 1024
 
 RE_NON_WHITESPACE = re.compile(r"\S")
+
+
+def strip_tags_plus(text: str, keep_leading_whitespace: bool = False) -> str:
+    if not text:
+        return ""
+
+    return _strip_tags_plus(text, keep_leading_whitespace=keep_leading_whitespace)
 
 
 def get_system_netloc_allowlist() -> list[str]:


src/openforms/api/views/views.py
```diff
@@ -36,6 +38,7 @@ def exception_handler(exc, context):
         if config("DEBUG", default=False):
             return None
 
+        logger.exception("api.unhandled_exception", exc_info=exc)
```
sergei-maertens, line 41: `HandledException.as_serializer` on L49 already logs: https://github.com/open-formulieren/open-forms/blob/7e5a70a12b1f3e6e202cfa2f06c0972dac736d14/src/openforms/api/exception_handling.py#L138

src/openforms/api/tests/test_error_format.py
```diff
@@ -177,7 +177,8 @@ def test_500_error(self):
 
     @tag("gh-4505")
     @patch.dict("os.environ", {"DEBUG": "no"})
-    def test_assertion_error_crash(self):
+    @patch("openforms.api.views.views.logger.exception")
```
sergei-maertens, line 180: please don't monkeypatch - [`assertLogs`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs) is built into the standard library

src/openforms/formio/rendering/tests/test_vanilla_formio_components.py
```diff
@@ -386,6 +382,35 @@ def test_wysiwyg_component_show_in_summary_enabled(self):
...
+            "showInSummary": True,
+        }
+        submission = SubmissionFactory.create(
+            form__name="public name",
+            form__generate_minimal_setup=True,
+            form__formstep__form_definition__configuration={"components": [component]},
+        )
+        step = SubmissionStepFactory.create(
+            submission=submission,
+            form_step=submission.form.formstep_set.get(form__name="public name"),
```
sergei-maertens, line 402:
```suggestion
            form_step=submission.form.formstep_set.get(),
```
there should only exist one step in the system

src/openforms/formio/rendering/default.py
```diff
@@ -173,6 +176,9 @@ def value(self) -> str | SafeString:
         if self.as_html:
             return mark_safe(content)
 
+        if not content:
+            return ""
```
sergei-maertens, line 180: looks to me that this guard clause should be moved to the `strip_tags_plus` implementation, as the same problem exists there with different inputs

sergei-maertens (changes requested):

src/openforms/api/views/views.py
bartjkdp: I will inspect this further and create a separate PR if required.

src/openforms/api/tests/test_error_format.py
bartjkdp: I will inspect this further and create a separate PR if required.

src/openforms/formio/rendering/default.py
bartjkdp: I agree, fixed this for now in Open Forms, but this could also be fixed upstream.

sergei-maertens (approved):

sergei-maertens:
I created an issue in the mail-cleaner library which should ideally have this patch: https://github.com/maykinmedia/mail-cleaner/issues/7 (for follow-up)

sergei-maertens:
Backports:

* stable/3.5.x: 1756b7356390bf13deab5a06881910793d5f2bf8
* stable/3.4.x: a2a159e803bfa63e8b4fb222244b9858f8335841

PR #6537, opened 2026-08-17
Fix incomplete variable extraction

### LINKED ISSUE #6292: Extracting variables from template expressions is incomplete
Noticed while doing a PR review for #6289 

Specifically for if statements which consist of multiple operators
The following test will not pass
```python
    def test_different_operators_in_if_statement(self):
        source = """
        {% if varA or varB %}
            {{otherVar}}
        {% elif varC != 'asdf' and varD > varE or varF %}
            {{yetAnotherVar}}
        {% else %}
            {{finalVar}}
        {% endif %}
        """

        variable_names = set(extract_variables_used(source))
        self.assertEqual(
            variable_names,
            {
                "varA",
                "varB",
                "otherVar",
                "varC",  # not in variable_names
                "varD",  # not in variable_names
                "varE",  # not in variable_names
                "varF",
                "yetAnotherVar",
                "finalVar",
            },
        )
```

This might lead to edge cases when analyzing/sorting logic rules, because variables are not picked up when they should be. (Variable extraction happens when determining the input variables of a service fetch action)

I think we need some more recursion in https://github.com/open-formulieren/open-forms/blob/669f65f30d2b01ab3941463520b995333079fe1a/src/openforms/template/__init__.py#L102

sergei-maertens:
Closes #6292

Again, thanks to @viktorvanwijk  for doing the hard work! More recursion is indeed the answer.

**Changes**

* Added regression test
* Explicitly target all Django-template supported infix operators

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/template/__init__.py
+++ b/src/openforms/template/__init__.py
@@ -18,6 +18,7 @@ from collections.abc import Iterator, Mapping
 from django.template.backends.django import Template as DjangoTemplate
 from django.template.base import FilterExpression, Node, Variable, VariableNode
 from django.template.defaulttags import ForNode, IfNode, TemplateLiteral
+from django.template.smartif import OPERATORS, TokenBase
 
 from .backends.sandboxed_django import backend as sandbox_backend, openforms_backend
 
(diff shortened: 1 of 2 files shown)


viktorvanwijk (approved):
I remember this being really annoying to write because of all the assertions required for type checking 😅

sergei-maertens:
Backports:

* 3.5.x: f4c781db6a3f8dc0969c7e2781fda0acb63fbdfc
* 3.4.x: not relevant, doesn't have new logic evaluation

PR #6538, opened 2026-08-17
Auto apply HC data patches

sergei-maertens:
Some quality of life improvements if you frequently clean up your docker system (like I do) and wonder every time why those tests are failing on BRP data when re-recording VCR cassettes...

[skip: e2e]


robinmolen (approved):
Cool! I think we can merge it 👍

PR #6543, opened 2026-08-18
Fix stale cache issue in logic evaluation / permission checks

### LINKED ISSUE #6468: Logic evaluation can have different results for Admins

3.3.x, 3.4.x, 3.5.x

### Customer reference

DH 875

under certain conditions the data of a step returns different data.
Resulting in different "results" between a submitter logged in as admin and non-admin.
Like a field(set) being hidden. 
It "feels" like the admin user is 1 "step" ahead and processes a extra round of logic before the step loads, compared to a non-admin user. 

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #6468
[skip: e2e]

**Changes**

...

**TODO**

- [ ] Check if this also happens on 3.5, both new and old logic evaluation
- [ ] Fortward port the dynamic options patch (all versions)
- [ ] Add (updated) regression test to main to ensure this problem doesn't happen (first attempts to reproduce there didn't hint at it, but that may be because of the missing default value in logic)
- [ ] Drop data src commit, the renderer does not assert anything

### DIFF
--- a/src/openforms/formio/dynamic_config/dynamic_options.py
+++ b/src/openforms/formio/dynamic_config/dynamic_options.py
@@ -116,10 +116,13 @@ def add_options_to_config(
                 raise ServiceUnavailable(
                     _("Could not retrieve options from Referentielijsten API."),
                 )
+            del component["openForms"]["service"]
+            del component["openForms"]["code"]
         case DataSrcOptions.variable:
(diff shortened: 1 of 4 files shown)


sergei-maertens:
Closing in favour of #6545

PR #6551, opened 2026-08-20
[#6515] cache certain catalogi client requests

### LINKED ISSUE #6515: A structural solution for uWSGI HTTP timeouts on forms with many backend registrations
### Thema / Theme

Frontend

For forms with many backend registrations, validating the backend plugins can take quite a long time. As a result, the request hits the configured timeout.

We have seen this before at Omgevingsdienst Groningen, where we increased the timeout to 120 seconds (when I tested it it could validate 70 registrations within the timeout - but most of them were the same, so maybe this was cached)   
Today, the issue occurred again. For an OMWB form with 28 registrations, 120 seconds still does not seem to be enough.
Omgevingsdiensten in particular tend to use a relatively large number of backend registrations, so we will probably run into this issue more often.

 I think it would be useful to look for a more structural solution. 

@sergei-maertens suggested caching some parts of the validation chain.

We can accommodate the needs of Omgevingdiensten

_No response_

SonnyBA:
Closes #6515 

**Changes**

These changes will decrease the amount of (external) recurring requests that can be done when validating ZGW form registration backends through caching. Note that the initial requests can still be slow as the cache needs to be populated.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

[skip: e2e]

### DIFF
--- a/src/openforms/contrib/zgw/clients/catalogi.py
+++ b/src/openforms/contrib/zgw/clients/catalogi.py
@@ -4,6 +4,8 @@ from functools import cached_property
 from operator import itemgetter
 from typing import Literal, NotRequired, TypedDict
 
+from django.core.cache import BaseCache, caches
+
 from flags.state import flag_enabled
 from requests import Response
 from zgw_consumers.nlx import NLXClient
@@ -138,16 +140,38 @@ class RoleTypeListParams(TypedDict, total=False):
     page: int
 
 
+CATALOGI_LOOKUP_CACHE_TIMEOUT = 60 * 60 * 12  # 12 hours
+CATALOGI_VERSION_LOOKUP_CACHE_TIMEOUT = 60 * 60 * 12  # 12 hours
+CATALOGI_CASE_TYPE_LOOKUP_CACHE_TIMEOUT = 60 * 60 * 1  # 1 hour
+CATALOGI_IOT_LOOKUP_CACHE_TIMEOUT = 60 * 60 * 1  # 1 hours
+
+
 class CatalogiClient(LoggingMixin, NLXClient):
     _api_version: CatalogiAPIVersion | None = None
 
     @property
+    def cache(self) -> BaseCache:
+        cache = caches["catalogi_client"]
+        assert isinstance(cache, BaseCache)
+        return cache
+
+    @property
     def api_version(self) -> CatalogiAPIVersion:
-        if self._api_version is None:
+        def _lookup_version() -> CatalogiAPIVersion:
             # hit a known endpoint and parse the response headers. VRSN is just a made
             # up domain to limit the result set, hopefully it's even entirely empty.
             response = self.get("catalogussen", params={"domein": "VRSN"})
-            self._api_version = self._determine_api_version(response)
+            return self._determine_api_version(response)
+
+        if self._api_version is None:
+            cache_key = f"ZGW|catalogi|version|{self.base_url}"
+            self._api_version = self.cache.get_or_set(
+                cache_key,
+                default=_lookup_version,
+                timeout=CATALOGI_VERSION_LOOKUP_CACHE_TIMEOUT,
+            )
+
+        assert self._api_version is not None
         return self._api_version
 
     def _determine_api_version(self, response: Response) -> CatalogiAPIVersion:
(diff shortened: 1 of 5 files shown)


src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -138,16 +141,36 @@ class RoleTypeListParams(TypedDict, total=False):
...
-        if self._api_version is None:
-            # hit a known endpoint and parse the response headers. VRSN is just a made
-            # up domain to limit the result set, hopefully it's even entirely empty.
-            response = self.get("catalogussen", params={"domein": "VRSN"})
-            self._api_version = self._determine_api_version(response)
+        cache_key = f"ZGW|catalogi|version|{self.base_url}"
+        if self._api_version is not None:
+            return self._api_version
+        elif (version := self.cache.get(cache_key)) is not None:
+            return version
```
sergei-maertens: I think you still want to set `self._api_version` if you get a cache hit - it saves a lookup in the cache when doing multiple checks on the same client instance


```suggestion
        elif (version := self.cache.get(cache_key)) is not None:
            self._api_version = version
            return version
```

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -138,16 +141,36 @@ class RoleTypeListParams(TypedDict, total=False):
...
+
+        # hit a known endpoint and parse the response headers. VRSN is just a made
+        # up domain to limit the result set, hopefully it's even entirely empty.
+        response = self.get("catalogussen", params={"domein": "VRSN"})
+        version = self._determine_api_version(response)
+        assert version is not None
+        self.cache.set(cache_key, version, CATALOGI_VERSION_LOOKUP_CACHE_TIMEOUT)
+        self._api_version = version
         return self._api_version
```
sergei-maertens: I think you can keep the implementation entirely in the `if self._api_version is None` block with the above suggestion, and you may even make the code a bit shorter by using `cache.get_or_set`, since you can pass a callable `default`

src/openforms/contrib/zgw/tests/test_catalogi_client.py
```diff
@@ -11,16 +11,26 @@
...
 from ..exceptions import StandardViolation
 
+CACHES = settings.CACHES.copy()
+CACHES["catalogi_client"] = {"BACKEND": "openforms.utils.cache.RequestProxyCache"}
+
 
+@override_settings(CACHES=CACHES)
 class CatalogiClientTests(TestCase):
+    def setUp(self) -> None:
+        self.addCleanup(clear_caches)
```
sergei-maertens:
```suggestion
    def setUp(self) -> None:
        super().setUp()
        self.addCleanup(clear_caches)
```
the next person who adds a mixin will thank you

src/openforms/forms/tests/test_import_export.py
```diff
@@ -2160,12 +2165,18 @@ def test_export_form_with_objects_registration_backend(self):
             )
 
 
+@override_settings(CACHES=CACHES)
```
sergei-maertens: what happens if you use the dummy cache instead in the CI/dev settings during testing? Because now it looks like all these tests are updated to use a per-request cache & reset the caches between tests anyway, so if we want a clean slate, why not run with code that always has a cache miss?

sergei-maertens (dismissed):

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -138,17 +140,41 @@ class RoleTypeListParams(TypedDict, total=False):
...
+CATALOGI_IOT_LOOKUP_CACHE_TIMEOUT = 60 * 60 * 1  # 1 hours
+
+
 class CatalogiClient(LoggingMixin, NLXClient):
     _api_version: CatalogiAPIVersion | None = None
 
+    @property
+    def cache(self) -> BaseCache:
+        cache = caches["catalogi_client"]
+        assert isinstance(cache, BaseCache)
```
sergei-maertens, line 155: nice type-checking workaround!

src/openforms/contrib/zgw/clients/catalogi.py
```diff
@@ -138,17 +140,41 @@ class RoleTypeListParams(TypedDict, total=False):
...
+        return cache
+
     @property
     def api_version(self) -> CatalogiAPIVersion:
-        if self._api_version is None:
+        cache_key = f"ZGW|catalogi|version|{self.base_url}"
+        api_version = self.cache.get_or_set(
+            cache_key,
+            default=self._api_version,
```
sergei-maertens: err, that's not entirely what I meant - I'd rather you define a callback/closure here that you can pass as default callback:

```py
@property
def api_version(self):
    def _lookup_version():
        response = self.get("catalogussen", params={"domein": "VRSN"})
        return self._determine_api_version(response)

    if self._api_version is None:
        self._api_version = self.cache.get_or_set(cache_key, default=_lookup_version, timeout=...)
    return self._api_version
```
something like that should do

sergei-maertens:
Has been tested and accepted!

sergei-maertens (approved):

sergei-maertens:
Backports:

* stable/4.0.x: aade82129c82f435312e54f1610a73bcb7b2f98a
* stable/3.5.x: 36fd4da005e5666d4c40c000252c97c555d6ebde

PR #6552, opened 2026-08-20
Fix variable-from-component sync

sergei-maertens:
Components in form steps must be backed by a form variable - this apparently broken with the new form builder integration. The new desired form variables state is computed and assigned in the state reducer for the form.

Additionally, on loading of the form steps and processing the configuration into form variables, this didn't work for edit grids because the edit grid component itself was incorrectly marked as being "inside an edit grid". This is now also patched by only considering the edit grid children during processing.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/js/components/admin/form_design/variables/utils.js
+++ b/src/openforms/js/components/admin/form_design/variables/utils.js
@@ -29,7 +29,7 @@ const isInEditGrid = (targetComponent, configuration) => {
 
     // Check if any of the components in the editgrid has the same key as the component
     // we're looking for.
-    const flatChildren = flattenComponents([component]);
+    const flatChildren = flattenComponents(component.components);
     if (flatChildren[targetComponent.key] !== undefined) return true;
   }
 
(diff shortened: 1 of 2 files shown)


src/openforms/js/components/admin/form_design/variables/utils.js
```diff
@@ -133,34 +131,24 @@ const updateFormVariables = (
...
     } else {
-      // When a new component is created, the callback is called multiple times by Formio. So we need to avoid adding
-      // the variable more than once.
-      const existingIds = updatedFormVariables
-        .filter(variable => !!variable._id)
-        .map(variable => variable._id);
-      if (existingIds.includes(component.id)) return updatedFormVariables;
-
-      updatedFormVariables.push(makeNewVariableFromComponent(component, formDefinition));
+      formVariables.push(makeNewVariableFromComponent(component, formDefinition));
```
sergei-maertens, line 137: simplified because our own builder does not do multiple change event dispatches :)

vaszig (approved):

PR #6567, opened 2026-08-24
Fix values only being set after interaction

### LINKED ISSUE #6507: Values set with back-end logic only appear after interaction with a form/step

3.5.6 (not 4.0!)

### Customer reference

DH 895 

With new renderer and new logic on, and off. the values set by a actions only appear in the form after interaction from the user. 

Image

interaction:
Image

result:
Image

single logic rule:

Image

also tested with number and string triggers. 

the kicker:
This behavior is different for admins, who do see the result on load. 

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #6507

**Changes**

The described problem is merely a side effect of the underlying root cause where some stale cached datastructures resulted in a data updated not being returned in the logic outcome of the submission step detail endpoint, leading to not-initially populated form fields. This didn't happen for admin users, because the logic check is skipped there in the permission backend.

It works in the interaction later on because it fires a check logic call with explicit `field: ""` data in the input, which *does* lead to a data update eventually, but it requires an interaction.

The regression test passes on main as-is, see 778d470969efe1b7c7dfedf181832fc2b3c41b22

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/submissions/api/permissions.py
+++ b/src/openforms/submissions/api/permissions.py
@@ -131,6 +131,7 @@ class CanNavigateBetweenSubmissionStepsPermission(permissions.BasePermission):
         # check, as we don't take unsaved/dirty data into account here.
         obj.form_step.form_definition.configuration = configuration_copy
         del obj.form_step.form_definition.configuration_wrapper
+        del submission.variables_state
 
         if not incomplete_steps:
             return True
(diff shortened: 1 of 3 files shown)


sergei-maertens:
This is already "the backport" - only 3.5.x is affected. Main is okay, and 3.4 doesn't have the new logic evaluation.

SonnyBA (approved):

PR #6569, opened 2026-08-24
[#6497] Report pre-registration failures

### LINKED ISSUE #6497: Registration issues not being reported in daily digest

3.5.x

### Customer reference

_No response_

It seems like registration issues are either not being reported in the daily digest mails, or are "under" reported.

in all cases the forms were in a "failed" state, registration errors in sentry (and the submission). So we expect these failures to be reported in the digest mail. 

let me know what kind of extra info is needed.

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes #6497

**Changes**

- Added logs for pre-registration phase failures.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/registrations/tasks.py
+++ b/src/openforms/registrations/tasks.py
@@ -1,5 +1,4 @@
 import traceback
-from contextlib import contextmanager
 
 from django.db import transaction
 from django.utils import timezone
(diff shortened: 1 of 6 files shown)


src/openforms/registrations/tasks.py
```diff
@@ -30,45 +29,22 @@
...
-
-
 @app.task()
 def pre_registration(submission_id: int, event: PostSubmissionEvents) -> None:
     submission = Submission.objects.get(id=submission_id)
+    form = submission.form
+    structlog.contextvars.bind_contextvars(
+        form=form.admin_name,
+        submission_uuid=str(submission.uuid),
+        public_reference=submission.public_registration_reference,
```
sergei-maertens: it doesn't have a public registration reference yet at the beginning of this task, pre-registration is the mechanism that populates it :)

src/openforms/registrations/tasks.py
```diff
@@ -94,6 +70,7 @@ def pre_registration(submission_id: int, event: PostSubmissionEvents) -> None:
         None if registration_plugin is None else registration_plugin.identifier
     )
     log = log.bind(plugin=plugin_repr)
+    audit_log = audit_log.bind(plugin=registration_plugin)
```
sergei-maertens, line 72:
```suggestion
    audit_log = audit_log.bind(plugin=plugin_repr)
```

we now get some duplicate reporting in Grafana on plugins used because the memory addresses are included in the repr, using this proper `plugin_repr` avoids that

sergei-maertens (changes requested):
I probably would've used the same `registration_failure` event, but it doesn't hurt to be more granular!

src/openforms/registrations/tasks.py
vaszig: I had it that way but I realized I get errors in https://github.com/open-formulieren/open-forms/blob/main/src/openforms/logging/adapter.py#L86 and I saw that everywhere in the code we use the instance for the audit logs.

sergei-maertens: ah check, can you create a backlog issue then to clean this up for the grafana dashboard reporting?

I think that a custom structlog processor to do this transformation would work well enough to solve both issues :)

sergei-maertens (approved):

sergei-maertens:
Backports:

* stable/3.5.x: dbb996f9176e9d659db65f4603c0e7206b08ab43

PR #6572, opened 2026-08-25
Ensure global validate.required default configuration is not ignored

### LINKED ISSUE #6570: General configuration option "fields required" is no longer honored
This parameter is not forwarded to our formio-builder and thus not taken into account any longer.

Requires a patch in the formio-builder + some glue code to forward the setting.

sergei-maertens:
Closes #6570 - depends on https://github.com/open-formulieren/formio-builder/pull/325 being merged and released

**Changes**

* Forward global configuration option as `FormioBuilder` prop
* Fix missing/empty context in appointmentconfig, which would break the JSON boolean parsing for this parameter

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/appointments/contrib/jcc_rest/admin.py
+++ b/src/openforms/appointments/contrib/jcc_rest/admin.py
@@ -10,3 +10,23 @@ from .models import JccRestConfig
 class JccRestConfigAdmin(SingletonModelAdmin):
     form = JccRestConfigForm
     change_form_template = "admin/jcc_rest/jccrestconfig/change_form.html"
+
+    def render_change_form(
+        self, request, context, add=False, change=False, form_url="", obj=None
+    ):
+        # Can't use `openforms.forms.admin.FormioConfigMixin` because the appointment
+        # configuration has different semantics - `required_default` is controlled by
+        # the upstream API, while file uploads/map components are irrelevant.
+        context.update(
+            {
+                "required_default": False,
+                "rich_text_colors": [],
+                "map_tile_layers": [],
+                "wms_layers": [],
+                "upload_filetypes": [],
+                "feature_flags": {},
+                "confidentiality_levels": [],
+                "component_empty_values": [],
+            }
+        )
+        return super().render_change_form(request, context, add, change, form_url, obj)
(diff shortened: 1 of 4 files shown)


src/openforms/appointments/contrib/jcc_rest/admin.py
```diff
@@ -10,3 +10,23 @@
 class JccRestConfigAdmin(SingletonModelAdmin):
     form = JccRestConfigForm
     change_form_template = "admin/jcc_rest/jccrestconfig/change_form.html"
+
+    def render_change_form(
+        self, request, context, add=False, change=False, form_url="", obj=None
+    ):
+        # Can't use `openforms.forms.admin.FormioConfigMixin` because those appointment
+        # configuration has different semantics - `required_default` is controlled by
+        # the upstream API, while file uplodas/map components are irrelevant.
```
robinmolen:
```suggestion
        # Can't use `openforms.forms.admin.FormioConfigMixin` because the appointment
        # configuration has different semantics - `required_default` is controlled by
        # the upstream API, while file uploads/map components are irrelevant.
```
I think?

src/openforms/appointments/contrib/jcc_rest/admin.py
```diff
@@ -10,3 +10,23 @@
...
+        context.update(
+            {
+                "required_default": False,
+                "rich_text_colors": [],
+                "map_tile_layers": [],
+                "wms_layers": [],
+                "upload_filetypes": [],
+                "feature_flags": {},
+                "confidentiality_levels": [],
+                "component_empty_values": [],
```
robinmolen, line 29: Should we not also pass the component empty values? (Genuine question, not sure if its needed)

robinmolen (approved):
Small typo and a question about the render_change_form context.

Besides that its great 👍

src/openforms/appointments/contrib/jcc_rest/admin.py
sergei-maertens: nope - I checked the code path and suspected(hoped) that this getComponentEmptyValue was dead code now with formiojs removal, but unfortunately it's still used when producing variables from components client side. As such, it's only relevant in the full form designer (as it's only related to variables), and not in the JCC config standalone builder.

Mind you - the standalone builder was working fine with broken context/variables, it only breaks with the new config property being passed along.

PR #6580, opened 2026-08-26
[#6575] Add migration for fixing `min` and `max` time values in `Time` component

### LINKED ISSUE #6575: Time component empty value is problematic
https://github.com/open-formulieren/open-forms/issues/6556#issuecomment-5394243797

See also Sentry from August 24th

vaszig:
Closes #6575

**Changes**

- The empty values for the undefined `minTime` and `maxTime` are empty strings so we end up with these wrongly passed to the `fromisoformat` function. The old renderer had `None` instead so the checks were normally passing. This depends on the new version of the form builder which fixes that.

**Checklist**

- [x] Update form builder
- [x] Add migration for wrongly configured components 

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


sergei-maertens:
Ah I see - the problem isn't in the renderer, but in the builder. I suspect this worked fine before because formiojs was still mangling stuff with the component when dragging it, but now that we have our own formiojs builder that layer is gone.

These should probably be `null` instead: https://github.com/open-formulieren/formio-builder/blob/4a2cc694125cebc7c5cd935829f920e4c94069a4/src/registry/time/edit.tsx#L152

Then we'll need a data migration / compont conversion to fix up any misconfigured components

vaszig:
> Ah I see - the problem isn't in the renderer, but in the builder. I suspect this worked fine before because formiojs was still mangling stuff with the component when dragging it, but now that we have our own formiojs builder that layer is gone.
> 
> These should probably be `null` instead: https://github.com/open-formulieren/formio-builder/blob/4a2cc694125cebc7c5cd935829f920e4c94069a4/src/registry/time/edit.tsx#L152
> 
> Then we'll need a data migration / compont conversion to fix up any misconfigured components

Thanks for the hint, I will take a look!
And you are absolutely right!This solves the problem indeed.

sergei-maertens:
Form builder update is happening in #6572

sergei-maertens (approved):

Issue #6210, opened 2026-04-23
Add missing validations for the new form types


vaszig:
After the new filed `type` has been added to the Form model, some validations need to take place in the backend. Check that all the necessary ones are in place (see below some examples).

- [x] a single page form type contains exactly 1 step (implemented in the form serializers (v2 & v3))
- [x] the one step does not require login
- [x] always allow the submission for `type=single_step`
- [x] no login related stuff is configured - This requires authentication details to be implemented for V3, which is covered in https://github.com/open-formulieren/open-forms/pull/6608
- [x] test prefill is not allowed (has been implemented and tested in the submission viewset)
- [x] write tests which prove that registration, logic, confirmation templates are allowed
- [x] tests which prove the logic is only used for the registration - deprecated, DMN, change variable value and registration backend are available actions in a single step form

vaszig:
The frontend (admin) already uses specific/restricted functionality regarding the `single_step` forms. We need these backend validations too, but this should be done after the related API (v3) rework from Sonny (the goal is adding these once and not touching everything twice).

sergei-maertens:
All the v3 API endpoint stuff should be there now :)

PR #6608, opened 2026-08-31
Fix missing endpoint features

### LINKED ISSUE #3941: Rethink admin Javascript to save a form
Currently we perform a bunch of parallel requests in the frontend to save a form after/during editing it: 

https://github.com/open-formulieren/open-forms/blob/ffd32bc0729c80f2a620efb5c4d4dd16c709ff89/src/openforms/js/components/admin/form_design/data/complete-form.js#L394

One big drawback is that the form variables are saved _after_ the form detail call (form create or update), but the registration backends/registration backend options are included in the latter. This means that the validation of registration backend options cannot check for form variables, since they are only saved later, which is not good for our input validation.

There are two options to be explored:

* break the registration backends saving out into its own call
* replace all the individual calls with one big endpoint to save the entire form

**TODO**

- [x] Validation errors on the form tab
- [x] Validation errors on the steps tab
- [x] Validation errors on the registration backends tab
- [x] Validation errors on the literals/translations tab
- [x] Validation errors on the variables tab
- [x] Validation errors on the logic tab
- [x] Validation errors on the confirmation tab
- [x] Validation errors on the submissions tab
- [x] Validation errors on the product tab
- [x] Validation errors on the data pruning tab
- [x] Validation errors on the advanced settings tab
- [x] Ensure auth backends are persisted/submitted
- [x] Ensure payment backend + options are persisted
- [x] Ensure `brpPersonenRequestOptions` advanced options are included in the payload/saved
- [x] Validation errors for service fetch logic

sergei-maertens:
Partly closes #3941
Relevant for #6210

**Changes**

Added missing serializer fields and removed unnecessary ones.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/forms/api/v3/serializers/form.py
+++ b/src/openforms/forms/api/v3/serializers/form.py
@@ -34,7 +34,6 @@ from openforms.variables.service import get_static_variables
 
 from ....api.serializers.form import (
     FormAuthenticationBackendSerializer,
-    FormLiteralsSerializer,
     FormRegistrationBackendSerializer,
     HelpCalloutPageSerializer,
     HelpDialogSerializer,
@@ -115,8 +114,6 @@ class FormSerializer(serializers.ModelSerializer):
         allow_null=True,
     )
 
-    literals = FormLiteralsSerializer(source="*", required=False)
-
     confirmation_email_template = ConfirmationEmailTemplateSerializer(
         required=False, allow_null=True
     )
@@ -181,7 +178,6 @@ class FormSerializer(serializers.ModelSerializer):
             "variables",
             "payment",
             "appointment_options",
-            "literals",
             "product",
             "slug",
             "type",
--- a/src/openforms/forms/api/v3/serializers/form.py
+++ b/src/openforms/forms/api/v3/serializers/form.py
@@ -5,7 +5,7 @@ from uuid import UUID
 
 from django.db import transaction
 from django.utils.text import get_text_list
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext, gettext_lazy as _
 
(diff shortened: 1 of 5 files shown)


sergei-maertens:
Needs some more tests to address the missing coverage.

src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -480,7 +498,16 @@ def update(self, instance: Form, validated_data: FormValidatedData) -> Form:
         for form_definition in form_definitions_created.values():
             FormVariable.objects.synchronize_for(form_definition)
 
-        # 3. registration backends
+        # 3. auth backends
+        auth_backends = validated_data.get("auth_backends", None)
```
vaszig, line 502: Is there a specific reason that we added `None` here?Just to be explicit?

vaszig:
> Needs some more tests to address the missing coverage.

Added the missing test for the auto login backend. I don't know if this is only what's left to review that.
If this is the case I can do it an then merge it since you have finished with release too.

src/openforms/forms/api/v3/serializers/form.py
sergei-maertens: Yeah, I kept it consistent with similar code for the registration backends etc.

sergei-maertens:
> > Needs some more tests to address the missing coverage.
> 
> Added the missing test for the auto login backend. I don't know if this is only what's left to review that.
> If this is the case I can do it an then merge it since you have finished with release too.

Yeah that was all that was left! Feel free to finish it and merge it

vaszig (approved):

Issue #6563, opened 2026-08-22
Fix data type for empty value of date, datetime and time components


sergei-maertens:
Currently, due to historical formio reasons, the empty value for `date`, `datetime` and `time` is an empty string. It's easy to see how this came to be, as these data types are usually serialized as ISO-8601 strings in JSON.

There are a couple of issues with having `""` the empty value, rather than `None` though:

* we require some special treatment now in our value conversion into the python domain
* it's not intuitive
* it makes client-side validation unexpected
* it doesn't play well with msgspec conversion, either requiring clunky wrapper classes/conversion or polluting all dowmstream code
* JSON schema does not treat the empty string as valid for `type: string, format: date-time` types, so current json schema's already need to allow for the empty string as a constant or widen up the schema

I believe this can be fixed structurally by switching to `None` / `null` as empty value. Existing form definitions/components can be patched up with a data migration, while import converters can handle legacy imports.

To be backwards compatible, some extra attention is required:

- [x] Registration backends must get an option to send "empty date, datetime, time as `null`", allowing opt-in for existing registration backends, while using that as default for newly added plugins. This gives downstream systems the time to update their schema's, on a per-form basis.
- [x] Logic that calculates dates/datetimes with a result of `""` should under the hood transform this into `None` - we can't automatically detect all these cases but also don't want to bother people too much. The UI can display warnings if it can be detected client-side.
- [x] Prefill needs to be normalized to not emit empty strings
- [x] Fix logic rules/triggers that compare against a date/datetime/time field (data migration) - this must also happen as an import converter!
- [x] Update renderer
- [x] Update builder
- [x] Update types repository/package
- [x] Find the code that was special casing date/datetime/time w/r to empty values and check if it can be removed (was already done in f799a46ef3dd7f70c3ee0cf0fbead7d4dde37da7)

sergei-maertens:
https://github.com/open-formulieren/open-forms/commit/df081a144a053dd20697e1d43478d13ada6105ef already tackled the datatypes normalization into the Python domain, meaning that our logic evaluation already uses `None` for empty date/datetime/time component values rather than the empty string. That means we don't need to do much additional work for the backwards compatibility - that part is limited to the registration backends.

PR #6615, opened 2026-09-01
Use `None` as empty value data type for date, datetime and time components

sergei-maertens:
Closes #6563 (partly)

For the reviewer(s) - this is a challenging one to review because you need to have an updated SDK with the updated renderer, which are both unreleased/unmerged at this point. All of those (breaking changes) PRs will only be released once the general direction is approved and we'll have to do a coordinated effort across multiple repositories :grimacing: 

I will change this PR in a stacked PR to make reviewing the individual changes easier!

[skip: e2e] (remove this before checking in/reviewing!)

---

Requires:

* an update of the formio builder
* an update of the formio renderer

---

**Changes**

* Updated the backend code that deals with the empty values of these component types - now `None` is used for empty values rather than the empty string `""`
* Updated the form designer JS code to ensure the correct json logic triggers are constructed and variables are initialized correctly.
* ~~Added data migration that automatically fixes existing logic triggers and actions~~ not necessary
* [x] Added option to registration backends to either send empty string or `null`
* [x] Re-record Objects API VCR cassettes

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/js/components/admin/form_design/logic/Trigger.js
sergei-maertens, line 1: This is... hard to comprehend :|

src/openforms/formio/components/custom.py
```diff
@@ -212,11 +224,12 @@ def mutate_config_dynamically(
         mutate_min_max_validation(component, data)
 
         # inject the translated placeholder for the formio DateTimeField component
+        # TODO: this can be removed now that the legacy renderer is gone?
```
sergei-maertens, line 227: follow up on TODO

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -603,18 +620,22 @@ def _value_to_json(
...
-                return ""
+                return "" if use_legacy_mode_for_datelike else None
             return value.isoformat()
 
         if value and data_type == FormVariableDataTypes.partners:
             value["dateOfBirth"] = self._value_to_json(
-                value["dateOfBirth"], FormVariableDataTypes.date
+                value["dateOfBirth"],
+                FormVariableDataTypes.date,
+                use_legacy_mode_for_datelike=True,
```
sergei-maertens, line 630: Need to determine what the impact is to not use legacy mode here, on our end it would simplify processing

src/openforms/submissions/models/submission_value_variable.py
```diff
@@ -603,18 +620,22 @@ def _value_to_json(
...
             )
             return value
 
         if value and data_type == FormVariableDataTypes.children:
             value["dateOfBirth"] = self._value_to_json(
-                value["dateOfBirth"], FormVariableDataTypes.date
+                value["dateOfBirth"],
+                FormVariableDataTypes.date,
+                use_legacy_mode_for_datelike=True,
```
sergei-maertens, line 638: Need to determine what the impact is to not use legacy mode here, on our end it would simplify processing

sergei-maertens:
df081a144a053dd20697e1d43478d13ada6105ef already normalizes (empty) date variable values to be in the python domain as `None`, which means that existing form logic must already compare against `None` for empty values rather than the empty string. I think that effectively removes the need for an import converter and data migration to fix up broken logic rules (as that would be necessary for backwards compatibility).

This has been live since Open Forms 3.4.0 even (!)

src/openforms/formio/components/custom.py
```diff
@@ -119,6 +119,7 @@ def mutate_config_dynamically(
         mutate_min_max_validation(component, data)
 
         # inject the translated placeholder for the formio DateField component
+        # TODO: this can be removed now that the legacy renderer is gone?
```
sergei-maertens, line 122: follow up on TODO

sergei-maertens:
Replaced by a PR stack starting at #6670 for more granular changes/reviewing

PR #6639, opened 2026-09-10
[#6622] apply email verification validation to customerProfile component

### LINKED ISSUE #6622: Update the formio renderer
- [ ] Move the existing code to it's own folder
- [ ] Update the profile component to use the existing verification flow)

Also have a look at https://github.com/open-formulieren/open-forms/issues/6494 (if the warning still shows up after the changes were made for this ticket)

SonnyBA:
Is required for the back-end part of https://github.com/open-formulieren/open-forms/issues/6622

**Changes**

Adds back-end validation for email verification for the customer profile component.

**Checklist**

- Impact on features

  - [ ] Checked copying a form
  - [ ] Checked new model fields are usable in the admin

- Dockerfile/scripts

- Commit hygiene

- Documentation

### DIFF
--- a/src/openforms/formio/tests/validation/test_profile.py
+++ b/src/openforms/formio/tests/validation/test_profile.py
@@ -1,7 +1,10 @@
 from django.test import TestCase, tag
 from django.utils.translation import gettext_lazy as _
 
-from openforms.submissions.tests.factories import SubmissionFactory
+from openforms.submissions.tests.factories import (
+    EmailVerificationFactory,
+    SubmissionFactory,
+)
 
 from ...typing.custom import CustomerProfileComponent
 from .helpers import extract_error, validate_formio_data
@@ -24,6 +27,12 @@ class ProfileValidationTests(TestCase):
         submission = SubmissionFactory.create(
             form__generate_minimal_setup=True,
             form__formstep__form_definition__configuration={"components": [component]},
+        )
+        EmailVerificationFactory.create(
+            submission=submission,
+            component_key="profile",
+            email="john@smith.org",
+            verified=True,
         )
         valid_values = {
             "profile": [
@@ -54,6 +63,9 @@ class ProfileValidationTests(TestCase):
         submission = SubmissionFactory.create(
             form__generate_minimal_setup=True,
             form__formstep__form_definition__configuration={"components": [component]},
+        )
+        EmailVerificationFactory.create(
+            submission=submission, component_key="profile", email="", verified=True
         )
         valid_values = {
             "profile": [
@@ -141,6 +153,12 @@ class ProfileValidationTests(TestCase):
                 ]
             },
         )
+        EmailVerificationFactory.create(
+            submission=submission,
+            component_key="profile",
+            email="some-incorrect-email",
+            verified=True,
+        )
         values = {
             "profile": [
                 {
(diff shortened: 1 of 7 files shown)


src/openforms/formio/validators.py
```diff
@@ -41,3 +46,44 @@ def validate_formio_js_schema(value: dict):
...
+        self.component_type = component_type
+
+    def __call__(self, value: str | DigitalAddress, field: serializers.Field) -> None:
+        from openforms.submissions.models import EmailVerification, Submission
+
+        address: str
+        error_format: str | dict[str, list[str]]
+        if self.component_type == "customerProfile":
+            assert isinstance(value, dict)
+            if value.get("type", "") == "phoneNumber":
```
SonnyBA, line 71:
```suggestion
            if value.get("type", "") != "email":
```
Note to self to update this

src/openforms/formio/tests/validation/test_profile.py
```diff
@@ -55,6 +64,9 @@ def test_validate_profile_empty_valid(self):
             form__generate_minimal_setup=True,
             form__formstep__form_definition__configuration={"components": [component]},
         )
+        EmailVerificationFactory.create(
```
vaszig, line 67: This is a bit weird to me, is it needed?It doesn't make sense to me to have an email verified without an email address.

src/openforms/formio/tests/validation/test_profile.py
```diff
@@ -141,6 +153,12 @@ def test_validate_profile_wrong_address_format_email(self):
                 ]
             },
         )
+        EmailVerificationFactory.create(
```
vaszig, line 156: The same here.

src/openforms/formio/validators.py
```diff
@@ -41,3 +46,44 @@ def validate_formio_js_schema(value: dict):
...
+    component_key: str
+    component_type: EmailVerificationComponentType
+
+    def __init__(
+        self, component_key: str, component_type: EmailVerificationComponentType
+    ) -> None:
+        self.component_key = component_key
+        self.component_type = component_type
+
+    def __call__(self, value: str | DigitalAddress, field: serializers.Field) -> None:
```
vaszig, line 64: The `EmailVerificationValidator` validates the email addresses, maybe we should have it as a validator in the real address CharField and not the whole `DigitalAddress`?I was thinking that it should be more specific to the real field we check.What do you think?

Furthermore, should the `EmailVerificationValidator` know about all these?Maybe it should only accept and validate email addresses?

src/openforms/formio/tests/validation/test_profile.py
SonnyBA: Since there is a separate test to check the validation of empty email addresses I don't think we should change this behavior although I do agree that it seems a bit strange. @sergei-maertens can you recall why empty email addresses are considered valid for the profile component?

sergei-maertens: I think it's valid because the component doesn't have `validate.required: true` (`required` is entirely absent, which implicitly makes it `required=False`)

sergei-maertens: In this case, I would then also not expect any verification to be done and/or created/saved because the field was simply not filled out, so the remark about the weird empty `EmailVerification` stands. I think it's probably even wise to add a constraint to the model that doesn't allow empty email addresses at the database level!

PR #6657, opened 2026-09-14
Prevent whitespace collapse in summary PDF

### LINKED ISSUE #6649: Textfields values are rendered without any linebreaks in the PDF

3.5

### Customer reference

_No response_

When people write something like this in a TextField:

```
1. Foo
2. Bar
This is
a test
```

The PDF that you can download afterwards shows:

```
1. Foo 2. Bar This is a test
```

_No response_

I expect line breaks to be kept in the PDF for TextFields at least. Both for readability and it was what the user actually submitted (so technically, we lost data).

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes #6649

**Changes**

Textarea components allow multi-line content and some basic formatting by users with newlines. Ensure that those newlines are converted to linebreaks when formatting the content as HTML, as the default HTML behaviour is to collapse all whitespace which destroys the careful structuring by the form submitter.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


vaszig (commented):
Seems ok to me according to the test, but while testing that locally I see that it works as expected when I follow the link via admin (`Submission PDF preview`) but it doesn't when I download the report from the confirmation page. Am I forgetting something about these two and does it work properly for both when you test it?

sergei-maertens:
Weird - I seem to remember you having a similar issue last time we had something with the PDFs. I did generate the PDF locally from the CLI for a submission and there it was broken before the patch and fixed after the patch :thinking:

vaszig:
Yes, thanks for making that clear because I remember something similar with PDFs but wasn't sure. It's weird because I saw in the code (with a breakpoint) that the newlines are converted to linebreaks as expected but maybe I have a problem when downloading the PDF. I guess the PDF is generated in the exact way in both ways (CLI and frontend via the link).

vaszig (approved):
Ok, this is now working in both. I tried a fresh incognito browser and I can see the proper results.

sergei-maertens:
Backports:

* stable/4.0.x: 7c9b5cb63afe5627784ada329a3105be5713324f
* stable/3.5.x: c32c602c3fb42031011e57c25de53cad82db9ae4

PR #6659, opened 2026-09-15
[#6625] Update the format of Profile component options

### LINKED ISSUE #6625: Update the Customer Interactions prefill plugin
To retrieve the verification timestamp and de-duplicate the digital addresses. Also double check if the verification status has to be checked in some way if a verification timestamp was already set.

*** Make sure we have backwards compatibility (regarding the available options for emails and phones).

vaszig:
Closes #6625

**Changes**

In this PR I have only implemented what's needed for de-duplicating the addresses (have also updated the format of options to avoid double work). The necessary update concerning the verification date will be in another PR.

 - modified/updated Open Klant container concerning fixtures (added
 2 more identical emails and one phone number for testing the functionality
 of de-duplicating the addresses)
 - updated the serializer and the types for communication channels as we now
 return a list of dicts and the verification date as well
 - updated the relevant/existing tests and added integration test for the whole flow

 We used to return a list of options (strings) for both email and phone
 addresses for the Profile component. This is now changed to a JSON field
 since we need to have the verification date exposed to the frontend too.
 The addresses (both phone and email) are now de-duplicated and we only
 expose the unique ones to the frontend.

 For backwards compatibility we had at the beginning an extra function that
 transformed the lists of strs to a dict and checked if the verication date
 was present in our` EmailVerification` model. This was not ideal so we opted
 for a better approach, run the prefill again and populate the data accordingly.

 Initially, this was implemented by calling the prefill again when `submission_resumed`
 signal was fired. Writing some integration tests showed though that this is not stable
 because several places use this signal so the order of the functions is not guaranteed.
 Django will execute the receivers in the order they are connected internally, but with
 decorators that effectively means import/registration order can depend on how your
 apps/modules are loaded. Having all these in mind the prefill is again called via the
 actual resume view.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +46,156 @@ def transform_digital_addresses(
...
                 None,
             ),
         }
         result.append(group_preferences)
     return result
+
+
+# TODO
+# Remove this on Open Forms v5.0?
+def transform_options(
```
vaszig, line 73: @SonnyBA @sergei-maertens feel free to take a first look at how I implemented this. As we have already discussed we should maybe consider adding a data migration instead for suspended/not completed submissions.

src/openforms/prefill/contrib/customer_interactions/api/serializers.py
```diff
@@ -16,7 +16,7 @@ class CommunicationPreferencesSerializer(serializers.Serializer):
         help_text=_("Communication channel type"),
     )
     options = serializers.ListField(
-        child=serializers.CharField(),
+        child=serializers.JSONField(),
```
sergei-maertens, line 19: at this point you may as well make the entire `options` file a `JSONField` and store a list inside, no?

src/openforms/prefill/contrib/customer_interactions/api/views.py
```diff
@@ -66,4 +70,15 @@ def get_objects(self):
         state = submission.variables_state
         value = state.get_data()[form_variable.key]
         assert isinstance(value, Sequence)
-        return value
+
+        # for backwars compatibility we process and transform the options according to
+        # the new format (JSON field with verification date added)
+        transformed_value = None
```
sergei-maertens, line 76:
```suggestion
        transformed_value = value
```
and then you can drop the if check on L81 and just always used `transformed_value`?

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +46,156 @@ def transform_digital_addresses(
...
+                    "verification_date": address.get("verificatieDatum"),
+                }
+                if address["soortDigitaalAdres"] == "email"
+                else {
+                    "phone_number_address": address["adres"],
+                    "verification_date": address.get("verificatieDatum"),
+                }
+                for address in group
+                if address["soortDigitaalAdres"] in ("email", "telefoonnummer")
+            ],
```
sergei-maertens, line 61: can't we use the generic key `value` or `address` instead of making this polymorphic? I think it will simplify processing and improve code readability

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +46,156 @@ def transform_digital_addresses(
...
+
+        assert isinstance(options, list)
+
+        # new format of options
+        if all(isinstance(option, dict) for option in options):
+            # early return as we know that only the new format contains dicts for
+            # both email and phone addresses
+            return digital_address_data
+
+        # old/backwards compatible format
```
sergei-maertens, line 112: is there any way we could *refetch* the prefill data for this particular field? That way we would avoid making users verify an address that was already verified

I think there could be a generic hook in there when a submission is resumed that it triggers prefill again (but limit it to only some plugins?), just like on submission start, and that should update the relevant user defined variables, including the format. Then we don't need to do "runtime" transformation at all?

vaszig: Yes, I guess we could but as I had mentioned I was trying to avoid re-triggering the whole prefill again. But what you mean is having a generic hook which will call the https://github.com/open-formulieren/open-forms/blob/main/src/openforms/prefill/service.py#L114, so I have to modify the generic code in order to do that for a specific submission value variable, if I understand what you mean.

On the other hand, I believe that this is an edge case and will not affect so much the end user if the prefill plugin is again called (instead of forcing them to re-verify an address), so I can try?

sergei-maertens: yeah indeed - call that again, but perhaps pass it an (optional) list of plugins to limit it to, so that we don't run all the prefill again?

src/openforms/prefill/contrib/customer_interactions/api/serializers.py
vaszig: I do not understand tbh why we would need this change. The `options` will always be a list and the elements a JSON configuration, not the opposite.

sergei-maertens: A list of JSON values is kind of weird because a list is already a JSON value itself. It adds a bunch of indirection without any guaranteeds that the item structure is as expected.

I'd rather go for a top level JSON field and a validator on that field that:

* validates the value is a list
* validates the shape of each item

In particular, these things should even be possible with django-jsonform and a related json schema to describe the structure, but we don't currently do that anywhere else so I'd keep that out of scope.

sergei-maertens: Actually now that I wrote that down, why don't we use a serializer that describes the shape of a single item and do `options = OptionSerializer(many=True)`?

src/openforms/submissions/views.py
```diff
@@ -173,7 +175,24 @@ def get_redirect_url(
...
     token_generator = submission_resume_token_generator
 
+    def _re_run_prefill_plugins(self, submission: Submission) -> None:
+        """
+        Perform the prefill flow again for specific/supported plugins.
+        """
+        # After the signal is done we need a fresh instance of the submission where all
+        # the needed data has been updated (see AuthInfo model for example which is updated
+        # via the ``store_auth_details`` function during the ``submission_resumed`` signal)
+        submission.refresh_from_db()
```
vaszig, line 185: This was necessary because the submission we pass is not the updated (from the signal work) one. After a lot of debugging I saw that the `AuthInfo` instance was updated as expected but the submission itself had the old one attached.

Basically all these are mentioned/updated because we had the hashed values here instead of the real ones.

src/openforms/contrib/customer_interactions/tests/test_client.py
```diff
@@ -29,25 +29,54 @@ def test_list_digital_addresses_for_bsn(self):
...
+                "soortDigitaalAdres": "email",
+                "isStandaardAdres": False,
+                "verificatieDatum": None,
             },
         ]
-        self.assertEqual(len(data), 5)
+        self.assertEqual(len(data), 8)
+
+        remaining_addresses = list(data)
+
```
SonnyBA: I think this cast is not necessary? At line 14 it is also cast to a list.

src/openforms/contrib/customer_interactions/tests/test_client.py
```diff
@@ -80,9 +109,18 @@ def test_list_digital_addresses_for_kvk(self):
             },
         ]
         self.assertEqual(len(data), 2)
+
+        remaining_addresses = list(data)
```
SonnyBA: See comment above

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +45,54 @@ def transform_digital_addresses(
...
                 None,
             ),
         }
         result.append(group_preferences)
     return result
+
+
+# TODO
+# Check if we need to choose which address to keep in the case of duplicates (based on
+# another key like isStandaardAdres for example)
```
SonnyBA, line 91: Was this discussed already? Or will this be discussed at a later point? As a suggestion for this, we could opt for picking the option that has a `verification_date` set.

Edit: on second thought, I think the current implementation skips duplicate options that have the `verification_date` set as it picks up the first found address and skips the rest. This might lead to a situation where the first found address did not have a verification date set while a option later in the list (which is now skipped) could have the verification date set.

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +45,54 @@ def transform_digital_addresses(
...
+def filter_duplicate_addresses(
+    initial_addresses: VariableValue,
+) -> Iterable[VariableValue]:
+    """
+    Helper function to de-duplicate addresses.
+
+    Given a list of digital addresses remove all the duplicates (both emails and phones)
+    and return the updated one.
```
SonnyBA:
```suggestion
    and return the first found address.
```

src/openapi.yaml
```diff
@@ -9900,6 +9899,19 @@ components:
       required:
       - status
       - version
+    Options:
+      type: object
+      properties:
+        address:
+          type: string
+          description: An address value for the supported communication channels.
+        verificationDate:
```
SonnyBA: Since this is going to be a date it would be to have the `format` specified to `date` similar to how it is [described here](https://github.com/open-formulieren/open-forms/blob/eccb866ca1c8ecbaf41d586754fff6d49cae6d41/src/openapi.yaml#L435). For me this won't be a blocking issue though.

src/openforms/submissions/tests/test_resume_form_view.py
```diff
@@ -510,3 +524,204 @@ def test_resume_with_form_max_submissions_limit_not_reached(self):
...
+        # 2. suspend submission
+        # update the session again since the user at this point should be authenticated
+        session = self.client.session
+        session[FORM_AUTH_SESSION_KEY] = {
+            "plugin": "digid",
+            "attribute": AuthAttribute.bsn,
+            "value": "123456782",
+            "loa": DIGID_DEFAULT_LOA,
+        }
+        session.save()
```
SonnyBA: Is this needed (again) because the previous post request alters the session in some way?

src/openforms/submissions/tests/test_resume_form_view.py
```diff
@@ -510,3 +524,204 @@ def test_resume_with_form_max_submissions_limit_not_reached(self):
...
+            "submissions:resume",
+            kwargs={
+                "token": submission_resume_token_generator.make_token(submission),
+                "submission_uuid": submission.uuid,
+            },
+        )
+
+        self.assertNotEqual(submission.auth_info.value, "123456782")
+
+        self.client.get(resume_endpoint)
```
SonnyBA: I think an assertion that the response redirects to the expected form URL would help debugging if this test would fail for some reason in the future.

SonnyBA (changes requested):
I think the de-duplication code might need some changes but the other code looks 👍

src/openforms/contrib/customer_interactions/transform.py
vaszig: This is not discussed, that's why I added that comment. If there is a verification date all of them will be considered as verified. I think the best here is to check if there is indeed a verification date and if it's the case we pick that one instead. But I am not sure if this has to take into account the preferred address too. I think it should. So, we have the possible cases if you agree:

- verification date exists -> pick that one
- no verification date but existing as preferred -> pick that one
- no verification date and no preferred -> pick any of them

src/openforms/submissions/tests/test_resume_form_view.py
vaszig: I will double check because it's maybe a leftover from my endless effort to understand what's missing :)

src/openapi.yaml
vaszig: What you mention is a `serializers.DateField` so yes it makes sense there, but here I have a simple `serializers.CharField` and the Open Klant API expects something like `"verificatieDatum": "2019-08-24"`

src/openforms/contrib/customer_interactions/transform.py
sergei-maertens: I haven't looked at the code, so maybe I'll say something stupid, but I'm not sure why you have to "pick" anything?

If you have the following addresses (limiting it to email in this case):

* `foo@example.com`, referentie empty (`""`), `isStandaardAdres: false`, `verified: true` 
* `foo@example.com`, referentie `"OF-123456"`, `isStandaardAdres: true`, `verified: false`
* `bar@example.com`, referentie `"OF-ABCDEF"`, `isStandaardAdres: false`, `verified: false`

Then:

* `foo@example.com` is verified, because in all the addresses for that value there is at least one with a verification timestamp
* `bar@example.com` is not verified, because none of the addresses (only 1 here, but there may be more) have the verification timestamp

For `foo@example.com`, there is nothing to update to set verification - it already is considered verified. So, you don't have to pick any address there either.

For `bar@example.com`, I would apply the following logic:

* if there's a referentie other than `"portaalvoorkeur"` (set via envvar), do not consider that for patching
* if there's a referentie `"portaalvoorkeur"` (set via envvar and opted in to), prefer that for updating
* if there's a standard address (when not using the portaalvoorkeur mechanism), prefer that for updating
* otherwise create a new address and set the verification timestamp accordingly

src/openapi.yaml
sergei-maertens: only a date and not a full timestamp (datetime)?

src/openforms/contrib/customer_interactions/transform.py
SonnyBA: Yeah I agree with the behavior for the preferred addresses. Do note that:

- verification date exists and is preferred

is also one of the possible situations in the list you mentioned. ~@sergei-maertens how do you feel about the discussed behavior?~

Edit: this was commented before the comment above

vaszig: The code here https://github.com/open-formulieren/open-forms/pull/6659/changes#diff-000637dcdc0956b7b54d0891d9a35c5e3c1f6dab4832baff600aef529f7e2962R96 receives the initial dict of addresses and removes the duplicates (by returning the same structure), so what we return to the frontend will contain the verification date too. If we have 3 same addresses but we send the one that hasn't the verification date we would have to verify it again. Am I correct?

src/openapi.yaml
vaszig: If we want to go with what we already have in the `EmailVerification` model we should have a datetime, Open Klant has the date as I mentioned above, so I thought I shouldn't have a datetime when I receive a date from the API.

https://github.com/maykinmedia/open-klant/blob/master/src/openklant/components/klantinteracties/models/digitaal_adres.py#L79

src/openforms/contrib/customer_interactions/transform.py
SonnyBA: Yes, if there is no verification timestamp present in the data that is send to the front end then it assumes that the user should verify the email address and no verification has taken place yet.

vaszig: Some more context if it can help you.

This is an example of the data that is retrieved from the prefill (after the new structure transformation):

```json
[
    {
        "type": "email",
        "options": [
            {"address": "john.smith@gmail.com", "verification_date": None},
            {"address": "john.smith@gmail.com", "verification_date": "2026-09-09"},
            {"address": "someemail@example.org", "verification_date": None},
            {"address": "devilkiller@example.org", "verification_date": None},
            {"address": "john.smith@gmail.com", "verification_date": None},
        ],
        "preferred": "john.smith@gmail.com",
    },
    {
        "type": "phoneNumber",
        "options": [
            {"address": "0612345678", "verification_date": None},
            {"address": "0687654321", "verification_date": None},
            {"address": "0612345678", "verification_date": None},
        ],
        "preferred": "0612345678",
    },
]
```

So `john.smith@gmail.com` is verified but if we run the filter the address that will remain (according to the existing code) in the deduplicated addresses is the first one. So then the frontend will ask for verification again which is not correct. That's why I had the comment at this part of the code. 

Maybe Sergei had something else in mind concerning the this function.

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -8,7 +9,10 @@
 )
 
 from openforms.formio.typing.custom import SupportedChannels
-from openforms.prefill.contrib.customer_interactions.typing import CommunicationChannel
+from openforms.prefill.contrib.customer_interactions.typing import (
+    CommunicationChannel,
+)
```
sergei-maertens: I know it's just an import update, but it looks like this type should rather be defined in this package and imported by the prefill module, rather than the other way around (now the generic code is using stuff from prefill)

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +45,54 @@ def transform_digital_addresses(
...
         group_preferences: CommunicationChannel = {
             "type": channel_name,
-            "options": [address["adres"] for address in group],
+            "options": [
+                {
+                    "address": address["adres"],
+                    "verification_date": address.get("verificatieDatum"),
+                }
+                for address in group
+                if address["soortDigitaalAdres"] in ("email", "telefoonnummer")
```
sergei-maertens: this shouldn't be necessary because of L42-43, no? We already to the grouping by `soortDigitaalAdres` and only process the known types.

src/openforms/prefill/contrib/customer_interactions/api/serializers.py
```diff
@@ -9,17 +9,25 @@ class CommunicationChannels(TextChoices):
...
+class OptionsSerializer(serializers.Serializer):
+    address = serializers.CharField(
+        label=_("Address"),
+        help_text=_("An address value for the supported communication channels."),
+    )
+    verification_date = serializers.CharField(
+        label=_("Verification date"),
+        allow_null=True,
+        help_text=_("The verification date of an address."),
+    )
```
sergei-maertens: As suggested, this should be a boolean field `isVerified` (read only, all fields should be read only here) which is derived from "any (duplicated) address having a non-empty verification date"

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +45,54 @@ def transform_digital_addresses(
...
         }
         result.append(group_preferences)
     return result
+
+
+# TODO
+# Check if we need to choose which address to keep in the case of duplicates (based on
+# another key like isStandaardAdres for example)
+def filter_duplicate_addresses(
+    initial_addresses: VariableValue,
```
sergei-maertens: I would like to see a more accurate type here for the expected data - it should match the prefilled data shape:

```suggestion
    initial_addresses: Sequence[CommunicationChannel],
```

it's hard for me to reason about the shape of the data being processed - I had to look at the API view to see that it is indeed literally the value of the prefill variable that's extracted. That will give a call-site type checker warning, but in the future we'll be able to make that stricter with msgspec.

src/openforms/contrib/customer_interactions/transform.py
sergei-maertens: Discussed in standup this morning, so my conclusions are:

1. We store the "raw" data with verification date (date or none) as the result of the prefill action, which is covered by `transform_digital_addresses`
2. When emitting the available addresses to the frontend (which calls an API endpoint that calls this `filter_duplicate_addresses` I believe we have no use for the actual verification date in the frontend, just a flag if the address is verified or not

So given the input from above:

```json
[
    {
        "type": "email",
        "options": [
            {"address": "john.smith@gmail.com", "verification_date": None},
            {"address": "john.smith@gmail.com", "verification_date": "2026-09-09"},
            {"address": "someemail@example.org", "verification_date": None},
            {"address": "devilkiller@example.org", "verification_date": None},
            {"address": "john.smith@gmail.com", "verification_date": None},
        ],
        "preferred": "john.smith@gmail.com",
    },
    {
        "type": "phoneNumber",
        "options": [
            {"address": "0612345678", "verification_date": None},
            {"address": "0687654321", "verification_date": None},
            {"address": "0612345678", "verification_date": None},
        ],
        "preferred": "0612345678",
    },
]
```

I'd expect the filtering (and transformation) to lead to:

```json
[
    {
        "type": "email",
        "options": [
            {"address": "john.smith@gmail.com", "isVerified": true},
            {"address": "someemail@example.org", "isVerified": false},
            {"address": "devilkiller@example.org", "isVerified": false},
        ],
        "preferred": "john.smith@gmail.com",
    },
    {
        "type": "phoneNumber",
        "options": [
            {"address": "0612345678", "isVerified": false},
            {"address": "0687654321", "isVerified": false},
        ],
        "preferred": "0612345678",
    },
]
```

sergei-maertens (dismissed):

src/openforms/contrib/customer_interactions/transform.py
SonnyBA: @vaszig should we remove the comments in the code?

src/openforms/contrib/customer_interactions/typing.py
```diff
@@ -0,0 +1,28 @@
...
+    verification_date: str | None
+
+
+class CommunicationChannel(TypedDict):
+    type: SupportedChannels
+    options: Sequence[CommunicationChannelOptions]
+    preferred: str | None  # The preferred address in this channel
+
+
+# this defines the shape of the addresses tha the frontend receives
```
SonnyBA:
```suggestion
# this defines the shape of the addresses that the frontend receives
```

SonnyBA (approved):

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +45,89 @@ def transform_digital_addresses(
...
             "preferred": next(
                 (address["adres"] for address in group if address["isStandaardAdres"]),
                 None,
             ),
         }
         result.append(group_preferences)
     return result
+
+
+def normalized_data_for_frontend(
```
sergei-maertens: I'm not sure if this should be a separate function, will take a look in a bit and see if this `filter_duplicate_addresses` can be rewritten in a shorter version that combines the filtering and transformation. AFAIK this is only called in one place: the view that serializes the prefilled data.

vaszig: Yeah, maybe it's a bit excessive. And indeed, it's called only in the view that sends the data to the frontend.

src/openforms/prefill/constants.py
```diff
@@ -1,6 +1,14 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
+from .contrib.customer_interactions.constants import (
+    PLUGIN_IDENTIFIER as CUSTOMER_INTERACTIONS_PLUGIN,
+)
+
+# list of the supported prefill plugins that allow re-triggering prefill when a submission
+# is resumed and needs to update the prefilled data
+PREFILL_ON_SUBMISSION_RESUME_PLUGINS = [CUSTOMER_INTERACTIONS_PLUGIN]
```
sergei-maertens: This approach is not right architecture wise - it's forcing the generic prefill to be aware of all its plugins. We want to always invert that - generic code should be agnostic to plugin details.

Instead, I would explore if we can define a class attribute on the plugin class to mark if it needs to re-run prefill during submission resumption:

```py
class Plugin:
    rerun_on_submission_resume: ClassVar[bool] = False
```

sergei-maertens (changes requested):

src/openforms/contrib/customer_interactions/transform.py
```diff
@@ -41,11 +45,89 @@ def transform_digital_addresses(
...
+            existing_verified = bool(existing["verification_date"])
+
+            # prioritize verified address
+            if cur_verified and not existing_verified:
+                unique_address_options[cur_address] = option
+
+        address["options"] = list(unique_address_options.values())
+
+    result = normalized_data_for_frontend(deduplicated_addresses)
+    return result
```
sergei-maertens: This seems to pass the tests:

```py
def prepare_addresses_for_frontend(
    initial_addresses: Sequence[CommunicationChannel],
) -> Sequence[CommunicationChannelReturn]:
    """
    De-duplicate addresses and mark their verification status.
    """
    channels: list[CommunicationChannelReturn] = []
    for communication_channel in initial_addresses:
        # map of address to verification status
        addresses = defaultdict[str, bool](lambda: False)
        for option in communication_channel["options"]:
            address = option["address"]
            is_verified = addresses[address] or option["verification_date"] is not None
            addresses[option["address"]] = is_verified

        channels.append({
            "type": communication_channel["type"],
            "options": [
                {"address": address, "is_verified": is_verified}
                for address, is_verified in addresses.items()
            ],
            "preferred": communication_channel["preferred"],
        })

    return channels
```

src/openforms/submissions/views.py
```diff
@@ -173,7 +175,24 @@ def get_redirect_url(
...
+
+        if submission.suspended_on and not submission.completed_on:
+            prefill_variables(
+                submission,
+                plugins_for_submission_resume=PREFILL_ON_SUBMISSION_RESUME_PLUGINS,
```
sergei-maertens: With my earlier remark, I think we need a parameter more like `submission_phase: Literal["start", "resume"] = "start"`

for a generic approach :)

```suggestion
                submission_phase="resume",
```

src/openforms/prefill/base.py
```diff
@@ -51,6 +51,10 @@ class BasePlugin[OptionsT: Options](AbstractBasePlugin):
     """
     for_components: Container[str] = AllComponentTypes()
     options: SerializerCls = EmptyOptions
+    """
+    Indicates whether the prefill flow should re-run when a submission is resumed.
+    """
+    rerun_on_submission_resume: ClassVar[bool] = False
```
sergei-maertens:
```suggestion
    rerun_on_submission_resume: ClassVar[bool] = False
    """
    Indicates whether the prefill flow should re-run when a submission is resumed.
    """
```
docstring comes after the attribute for class/module attributes :)

sergei-maertens (approved):
Can you also squash your PR processing commits into the single PR commit?

vaszig:
Backports

stable/4.0.x: a7b711fe2d
stable/3.5.x: 8aeec9ea0d

PR #6660, opened 2026-09-15
Fix missing logic rule (static) analysis

sergei-maertens:
Was reported in person by Laurens - affects only main/our test env

**Changes**

The logic rule analysis handled the validation part, but didn't fully cover the assingment of each logic rule to the relevant form steps.

Added an assertion to existing tests to reproduce this & patched the code with some big implementation simplifications.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/forms/api/v3/serializers/form.py
```diff
@@ -296,18 +294,15 @@ def _validate_and_process_logic_rules(
...
+            [FormLogic(**rule, form=form) for rule in logic_rules_raw]
         )
 
         try:
-            updated_rules_and_steps = analyze_rules(
-                form,
-                rules=list(temp_logic_rules.keys()),
-                first_step=first_step,
-            )
+            form.apply_logic_analysis()
```
vaszig, line 305: The same we do in the tests right?I missed that method in the model.

vaszig (approved):
Nice cleanup, thanks. 
I saw there are some e2e tests that use Form Logic configuration, maybe we should trigger these too in CI?

src/openforms/forms/api/v3/serializers/form.py
sergei-maertens: indeed - the method was originally made as it's an easy way to "fix" the logic setup in existing tests, but now we can actually use it in "production" mode too :)

PR #6662, opened 2026-09-15
Handle upload race condition in e2e tests

sergei-maertens:
There's a race condition between the upload completing and the browser 'user' clicking the next button. When the button is clicked before the upload is complete, a validation error is displayed and the button is then rendered as disabled, but by then playwright is already looking for the next button to click on the confirmation page which obviously does not exist.

You don't see this in the snapshots of the failing test because when the upload eventually completes, the validation errors are cleaned as well and the button is enabled again.

The patch ensures that we wait until the delete button is rendered for the uploaded file, which replaces the 'loading' spinner while the upload is in progress. It should be a reliable signal that the upload has completed.


vaszig (approved):

sergei-maertens:
Backports:

* stable/4.0.x: a8fcd7a3700ea3ceeb7fa9dda5eac51547186957
* stable/3.5.x: 48dbad75791552676f094977d1a89dc438ee8455

PR #6663, opened 2026-09-15
fix(theme): improve .btn-light contrast in dark mode

### LINKED ISSUE #6661: Copy/paste buttons background color is not compatible with dark mode
See screenshot:

Image

These copy-paste buttons have a `btn-light` class name which is fine in light mode, but in dark mode the contrast with the icon is almost absent and it stands out negatively.

Probably a theme-specific override for `btn-light` is appropriate, but a targeted override is also okay. Eventually we do want to get rid of bootstrap styling though.

nazasnow:
Fixes #6661

### Summary
- Overrides .btn-light in the dark theme mixin to use a dark background and high-contrast text.
- Fixes illegible copy/paste button contrast in dark mode.


sergei-maertens (approved):
Verified - looks much better now! Thanks for the contribution.

For future contributions it would be nice if you can follow the commit message format described in https://github.com/open-formulieren/open-forms/blob/main/CONTRIBUTING.md#making-the-changes, but allt he essential information is there so I'll merge this as-is.

PR #6670, opened 2026-09-16
Date/datetime/time cleanup - remove obsoleted properties & prevent a logic analysis crash

sergei-maertens:
Closes #6563

[skip: e2e] (final PR will not skip)

**Changes**

Initially in #6615 I had a diagnostic tool that was analyzing logic rules for "invalid" date/datetime/time value comparisons. That crashed on a broken logic action configuration in my local dev database, so I fixed it in the code. This fix is circumstantial.

The main body of this PR is removing the formio properties from test code and type annotations because they're no longer relevant - they existed only to accomodate flatpickr used in the date/datetime picker from formio.js, but since we removed that in favour of our own renderer, this configuration is effectively useless.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/formio/tests/test_migration_converters.py
```diff
@@ -380,17 +379,6 @@ def test_empty_errors(self):
 
 
 class DatetimeTests(ParametrizedTestCase, SimpleTestCase):
-    def test_update(self):
```
vaszig, line 383: I don't remember this use case..haven't used that.

src/openforms/forms/models/logic.py
```diff
@@ -10,7 +10,7 @@
 
 from ordered_model.models import OrderedModel
 
-from openforms.forms.models import FormStep
+from openforms.forms.models import Form, FormStep
```
vaszig, line 13: I guess this could be a local import or is it on purpose?

vaszig (approved):

src/openforms/formio/tests/test_migration_converters.py
sergei-maertens: this was for an old formio library thing which is no longer relevant, something Silvia fixed at some point

src/openforms/forms/models/logic.py
sergei-maertens: yeah, looks like it could be not on purpose afaik!

I will not apply it in this PR though, since then I have to rebase 11 times again :sweat_smile: 

Can be picked up at some later point by anyone that stumbles on it :)

vaszig: Of course!Just asked to be sure there is not a specific reason

PR #6675, opened 2026-09-16
Date/datetime/time cleanup - ensure external date/datetime/times from prefill are normalized

sergei-maertens:
Closes #6563

**Changes**

If an external system still communicates empty dates/datetimes/times to us and that's used in prefill (of which we can't but assume that it works as expected right now), we should take care to transform that to `None` / `null` as that's how we model empty values.

I'm not really sure how necessary this is - normally I'd say this would be solved by the submission value variable `to_python` and `to_json` conversions, but the prefill mechanism works by directly assigning the retrieved value to the component `defaultValue` so I suspect those hooks never run.

[skip: e2e] (final PR in the stack will run them)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/formio/tests/test_normalization.py
```diff
@@ -1,10 +1,13 @@
...
 
+from hypothesis import given, strategies as st
+from unittest_parametrize import ParametrizedTestCase, parametrize
+
 from ..service import normalize_value_for_component
 from ..typing import Component
 
 
-class NormalizationTests(SimpleTestCase):
+class NormalizationTests(ParametrizedTestCase, SimpleTestCase):
```
vaszig, line 10: I guess having the `ParametrizedTestCase` (which inherits `TestCase`) and the `SimpleTestCase` make sense somehow?We use it like a mixin?

vaszig (approved):

src/openforms/formio/tests/test_normalization.py
sergei-maertens: yeah it's supposed to be like this - this `ParametrizedTestCase` should probably have `Mixin` in the name but it's an external library...

It does some magic with test generation based on the decorators/parameters.

PR #6678, opened 2026-09-16
Date/datetime/time cleanup - add option for legacy behaviour to Generic JSON registration plugin

sergei-maertens:
Closes #6563

**Changes**

Updated the backend and UI to allow opt-in to the legacy behaviour during registration, for the generic JSON registration plugin.

In generic JSON, you already select which variables to register. Where the date, datetime and time type variables used to have an empty string as empty value, they now have `None`, and the plugin emits the values as they are.

The legacy option applies a post-processing step to convert that back to empty string, when enabled.

Additionally, a data migration enables this option for all existing forms/plugins.

During import, we don't apply this conversion - see commit message for the data migration for details.

[skip: e2e] (final PR in the stack will run them)

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


src/openforms/registrations/contrib/generic_json/config.py
```diff
@@ -78,3 +78,14 @@ class GenericJSONOptionsSerializer(JsonSchemaSerializerMixin, serializers.Serial
...
         ),
     )
+    # DeprecationWarning - remove in Open Forms 5.0, then always send 'null'.
+    use_empty_string_for_empty_datelike_variables = serializers.BooleanField(
+        label=_("Use empty string for empty date, datetime and time values"),
+        required=False,
+        default=False,
+        help_text=_(
+            "Legacy compatibility option - send empty values for date, datetime and "
+            "time components as empty string instead of 'null'. This option be removed "
```
vaszig:
```suggestion
            "time components as empty string instead of 'null'. This option will be removed "
```

src/openforms/registrations/contrib/generic_json/plugin.py
```diff
@@ -231,6 +237,7 @@ def process_component(
     attachments: dict[str, list[SubmissionFileAttachment]],
     current_data_path: str = "",
     transform_to_list: list[str] | None = None,
+    use_empty_string_for_empty_datelike_variables: bool = False,
```
vaszig, line 243: Can we add the variable to the doc string help texts?

vaszig (approved):

PR #6682, opened 2026-09-16
Date/datetime/time cleanup - apply `null` empty value to family members date of birth fields

sergei-maertens:
Closes #6563

**Changes**

While fixing up the variable code I stumbled on the family members components/prefill that also have an embedded date field.

Apparently, this can be empty when dealing with partial dates - but from the code I can only conclude that this doesn't work properly in 4.0 or older either, so the change to `null` actually comes at a good time?

Either way - those empty values are now always normalized to `null` rather than the empty string.

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


vaszig (approved):
After having reviewed the whole stack I am not convinced that this is a really helpful way to review that. A lot of files that have been already part of previous PRs are again present in the following ones. If I remember right when I base a branch on another one it contains only the new changes, anyway I am a bit confused concerning how this is working in github, so we can have a quick call (maybe the whole team) and give us a bit of explanation!

sergei-maertens:
> A lot of files that have been already part of previous PRs are again present in the following ones. 

I think this is just because of my mistake in forgetting to push a branch somewhere in the middle.

The PR from branch X to branch Y is the same if you use stacks or not, so with my mistake the reviewing experience would have been the same even if these PRs were not in a stack.

sergei-maertens:
Ran the failing e2e test on an in-development SDK branch with the latest renderer, it passes locally:

```
(open-forms) ➜  open-forms git:(cleanup/6563-handle-family-members-embedded-date-of-birth) runtests src -k test_optional_fields
Found 1 test(s).
Using existing test database for alias 'default'...
System check identified no issues (0 silenced).
.
Ran 1 test in 5.735s

OK
Preserving test database for alias 'default'...
```

So I expected backend CI to be green once the renderer is released and updated in the main branch of the SDK.

Issue #6692, opened 2026-09-18
JSON editor is very narrow


magdalena-source:
### Product versie / Product version

3.5.8

### Customer reference

SED 144

### Omschrijf het probleem / Describe the bug

The JSON editor for most components is very narrow. This affects for example field groups, text fields and radio components. The only component that displays the JSON interface correctly is the content field.

<img width="1335" height="769" alt="Image" src="https://github.com/user-attachments/assets/a7d947d4-0310-41b6-bf34-c5c04e9b53bf" />

### Verwacht gedrag / Expected behavior

This:
<img width="1335" height="769" alt="Image" src="https://github.com/user-attachments/assets/70af5bf4-a4de-4fb4-be6c-0cd4abad10a7" />

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

sergei-maertens:
Can't reproduce this in 4.0 or newer, so the workaround will be limited to 3.5.x :)

PR #6693, opened 2026-09-18
Workaround monaco height determination

sergei-maertens:
Closes #6692

Monaco seems to look at it's container height to set it's automatically calculated width/height, and something broken between 3.5.7 and 3.5.8.

I suspect it has something to do with the browserslist update, but haven't confirmed that yet, and rolling back to older browser version support is untenable anyway.

As a workaround, we can help in setting up a minimum block size/ height for monaco to operate it - if the JSON mode is activated, we now allow Monaco to take up 75% of the modal height, which keeps the save/cancel/remove buttons in view most of the time, and always in regular preview mode (non-json editor).

This doesn't appear broken on main, but it turns out that we are mixing monaco versions there now - 0.3.0 of our monaco-json-editor for the non formio-builder stuff, and 0.2.0 otherwise. That will be looked into separately.

[skip: e2e]

**Checklist**

- Impact on features

  - [x] Checked copying a form

- Dockerfile/scripts

- Commit hygiene

- Documentation


sergei-maertens:
I have no idea what exactly is causing this regression between 3.5.7 and 3.5.8 - it  works fine on 3.5.7, stable/4.0.x and main. I suspected it was something with postcss/browserlist dependencies that got updated, but 4.0.x has the same versions. It could still be some interaction with the formio css that got removed in 4.0.x that causes the issue not to present itself in later versions.

vaszig (approved):
As mentioned on slack, this is not easily reproduced, but applying that patch works too.

